PR #241, opened 2013-07-12
Date Based File Archiving

mkaltner:
I made a few minor changes that allows archiving based on dates instead of rolling/sequential numbering.
The changes have minimal impact and work with the existing code/parameters as much as possible.
I found many posts/bug reports about deleting archive files with dates in the file name as I was having the same issue.
I've been testing this for a couple days and is working as expected.

Sample output showing archive files with datetime and maxArchiveFiles of 2 (The log file for 4:29 has been deleted):
image

Config used to achieve the results in the previous screenshot (archiveDateFormat is optional and will be generated based on archiveEvery setting if not specified):
image


swapneeld:
:thumbsup:

PR #334, opened 2014-05-01
Fixes empty "properties" collection.

erwinwolff:
When using an event-context variable, the properties collection was
empty, even when property items where explicitly given. This function
fixes this.

For more information, see this stackoverflow question (and answer!) on how and why this small patch was created: 

http://stackoverflow.com/questions/23272439/nlog-event-contextitem-xxxx-not-writing-in-logging-database

### DIFF
--- a/src/NLog/Targets/Target.cs
+++ b/src/NLog/Targets/Target.cs
@@ -435,6 +435,8 @@ namespace NLog.Targets
         {
             try
             {
+                this.MergeEventProperties(logEvent.LogEvent);
+
                 this.Write(logEvent.LogEvent);
                 logEvent.Continuation(null);
             }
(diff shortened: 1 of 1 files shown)


BrewDawg:
We are doing this exact same thing, this bug appears to not be fixed. We are on the latest via NuGet, have disassembled the code using DotNet Peek and the supposed fix is in there, however, none of our properties for these items are being saved to the database, always blank:

<parameter name = "@ErrorMethod" layout = "${event-properties:item=error-method}" />
<parameter name = "@ErrorClass" layout = "${event-properties:item=error-class}" />
<parameter name = "@ErrorSource" layout = "${event-properties:item=error-source}" />
<parameter name = "@ErrorMessage" layout = "${event-properties:item=error-message}" />
<parameter name = "@InnerErrorMessage" layout = "${event-properties:item=internal-error-message}" />

This is just not working and makes it almost useless to try to log information to SQL Server as you get almost no useful information

PR #422, opened 2014-09-19
Run logging code outside of transaction

Giorgi:
If the user tries to log to database inside transaction scope and rollbacks that transaction the logging will be rolled back to0. With this change we never create a transaction when writing to database.

### DIFF
--- a/src/NLog/NLog.doc.csproj
+++ b/src/NLog/NLog.doc.csproj
@@ -52,6 +52,7 @@
     <Reference Include="System.Messaging" />
     <Reference Include="System.Runtime.Serialization" />
     <Reference Include="System.ServiceModel" />
+    <Reference Include="System.Transactions" />
     <Reference Include="System.Web.Services" />
     <Reference Include="System.Windows.Forms" />
     <Reference Include="System.Xml" />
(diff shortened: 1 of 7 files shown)


Giorgi:
@Xharze Was this included in the 4.0 release?

304NotModified:
yes :)

I has been merged and it has the milestone 4.0. 

It was not in the news post maybe, that because it was not labeled as "feature"

304NotModified:
I was not in the NLog team when this was merged. 

As far as I understand, `UseTransactions ` option has been removed?

304NotModified:
It would be great if you would update the wiki. https://github.com/NLog/NLog/wiki/Database-target

Giorgi:
Yes, the option was removed because the logging code always runs outside of transaction to ensure that the log gets written to the database if you rollback the main transaction because of an error and want to log the error.

304NotModified:
It would be great if you would update the wiki. https://github.com/NLog/NLog/wiki/Database-target

Giorgi:
Done.

304NotModified:
Thanks but the wiki is also for older versions. So just removing the option on the docs isn't maybe  the best idea. 

I think It needs be there and explained why it's removed in 4.0. 

Do you agree?

Giorgi:
Done.

304NotModified:
your still editing?

Giorgi:
No

304NotModified:
can you break down to smaller sentences ;) ?

Giorgi:
Better now?

PR #451, opened 2014-11-26
added FastColoredConsoleTarget

grehovm:
I wrote a new target. It is faster than ColoredConsole, but has fewer features. It is necessary for debugging when you want to display a lot of information to the console without slowing down the program.

In my test, it is 2-3 times faster
```C#
        public async Task LoggerTest(ILogger logger)
        {
            var sw = new Stopwatch();
            Console.WriteLine("Start Logger Test");
            sw.Start();
            var list = new List<Task>();
            for (int i = 0; i < 30; i++)
                list.Add(Task.Factory.StartNew(() =>
                {
                    for (int j = 0; j < 1000; j++)
                        logger.Debug("Thread {0} - ping {1}", Thread.CurrentThread.ManagedThreadId, j);
                    logger.Error("Thread {0} - end", Thread.CurrentThread.ManagedThreadId);
                }));
            await Task.WhenAll(list);
            sw.Stop();
            //Console.ReadKey();
            Console.WriteLine("Elapsed time {0}", sw.ElapsedMilliseconds);
            Console.WriteLine("Stop Logger Test");
        }
```

I do not know what projects to add target, so only add .net4.5. I hope you will take and do the rest.
Sorry for google translate.

### DIFF
--- a/src/NLog/Targets/FastColoredConsoleTarget.cs
+++ b/src/NLog/Targets/FastColoredConsoleTarget.cs
@@ -1,0 +1,174 @@
+﻿using System;
+
+namespace NLog.Targets
+{
+    /// <summary>
+    /// Writes log messages to the console with customizable coloring
+    /// The maximum decrease in the number of switching colors console
+    /// x2-3 faster than ColoredConsole, but not so advanced 
+    /// </summary>
+    [Target("FastColoredConsole")]
+    public class FastColoredConsoleTarget : TargetWithLayoutHeaderAndFooter
+    {
+        /// <summary>
+        /// Gets or sets a value indicating whether to send the log messages to the standard error instead of the standard output.
+        /// </summary>
+        public bool Error { get; set; }
+        /// <summary>
+        /// Gets or sets a value indicating whether to use foreground color.
+        /// </summary>
+        public bool UseForegroundColor { get; set; }
+        /// <summary>
+        /// Gets or sets a value indicating whether to use background color.
+        /// </summary>
+        public bool UseBackgroundColor { get; set; }
+        /// <summary>
+        /// Gets or sets a value indicating whether to reset color after write.
+        /// </summary>
+        public bool UseResetColor { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value for Trace foreground color.
+        /// </summary>
+        public ConsoleColor TraceForegroundColor { get; set; }
+        /// <summary>
+        /// Gets or sets a value for Trace background color.
+        /// </summary>
+        public ConsoleColor TraceBackgroundColor { get; set; }
+        /// <summary>
+        /// Gets or sets a value for Debug foreground color.
+        /// </summary>
+        public ConsoleColor DebugForegroundColor { get; set; }
+        /// <summary>
+        /// Gets or sets a value for Debug background color.
+        /// </summary>
+        public ConsoleColor DebugBackgroundColor { get; set; }
+        /// <summary>
+        /// Gets or sets a value for Info foreground color.
+        /// </summary>
+        public ConsoleColor InfoForegroundColor { get; set; }
+        /// <summary>
+        /// Gets or sets a value for Info background color.
+        /// </summary>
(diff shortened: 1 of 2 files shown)


304NotModified:
@Xharze  We need also unit tests for `ColoredConsoleTarget`?

src/NLog/Targets/FastColoredConsoleTarget.cs
```diff
@@ -0,0 +1,174 @@
...
+                else if (level == LogLevel.Debug)
+                    SetColor(DebugForegroundColor, DebugBackgroundColor);
+                else if (level == LogLevel.Info)
+                    SetColor(InfoForegroundColor, InfoBackgroundColor);
+                else if (level == LogLevel.Warn)
+                    SetColor(WarningForegroundColor, WarningBackgroundColor);
+                else if (level == LogLevel.Error)
+                    SetColor(ErrorForegroundColor, ErrorBackgroundColor);
+                else if (level == LogLevel.Fatal)
+                    SetColor(FatalForegroundColor, FatalBackgroundColor);
```
304NotModified: Please refactor this to a switch which sets two vars. (and the setColor call below the switch)

304NotModified:
I'm willing to write the unit tests if:

- Refactor the code a bit (see comment)
- We get some documentation for this (on the wiki). Most important: what are the "fewer features", what is different to the `ColoredConsole` and why and how should be use this.

grehovm:
Hello, LogLevel is a class, not enum (switch can not be used or I just do not understand you). What's wrong with the code that is already written?
As for the tests - I think should go test for coloredconsole (but it is not present).
Standard coloredconsole have an extra rules and often set color that reduces performance

304NotModified:
Hi,

Yes you're right about `LogLevel` and the `switch`. I was sleepy ;)

There is nothing wrong, although I think the maintainability of the code could be better.

304NotModified:
Not sure what to do with this. Close or fix it. If we add docs to the wiki and make ik mergable, we could include it.

304NotModified:
If someone need like this.  Up vote this.

ie-zero:
When we identify the features included in the FastColouredConsoleTarget I think we should merge it.

304NotModified:
Agree, and also fixing the conflicts :P

grehovm:
it's do only this(minimization of switching colors)
```C#
        private void Output(string s, LogLevel level)
        {
            if (UseForegroundColor || UseBackgroundColor)
                if (level == LogLevel.Trace)
                    SetColor(TraceForegroundColor, TraceBackgroundColor);
                else if (level == LogLevel.Debug)
                    SetColor(DebugForegroundColor, DebugBackgroundColor);
                else if (level == LogLevel.Info)
                    SetColor(InfoForegroundColor, InfoBackgroundColor);
                else if (level == LogLevel.Warn)
                    SetColor(WarningForegroundColor, WarningBackgroundColor);
                else if (level == LogLevel.Error)
                    SetColor(ErrorForegroundColor, ErrorBackgroundColor);
                else if (level == LogLevel.Fatal)
                    SetColor(FatalForegroundColor, FatalBackgroundColor);
            (Error ? Console.Error : Console.Out).WriteLine(s);
            if(UseResetColor)
                Console.ResetColor();
        }

        private void SetColor(ConsoleColor foreground, ConsoleColor background)
        {
            if (UseForegroundColor && (UseResetColor || _lastForegroundColor != foreground))
            {
                _lastForegroundColor = foreground;
                Console.ForegroundColor = foreground; 
            }
            if (UseBackgroundColor && (UseResetColor || _lastBackgroundColor != background))
            {
                _lastBackgroundColor = background;
                Console.BackgroundColor = background;
            }
        }
```
if you look at ColoredConsole it does not care about it.
a year has passed, to me don't need it more.
Sorry for google translate.

304NotModified:
Well we can also add a "fastmodus" to the current consoleTarget and close this PR?

304NotModified:
OK close this. Maybe in the future a performance switch consoletarget.

PR #461, opened 2014-12-08
added support for platform: unity3d webplayer

AqlaSolutions:
Based on Silverlight 4 but without file access. See http://docs.unity3d.com/410/Documentation/ScriptReference/MonoCompatibility.html


304NotModified:
We like to merge this PR, but we need some advise from experienced users of Unity3D. Do you have experience and some time to help? Let us know, so we can fix this PR, merge and release it!

AqlaSolutions:
Hi again, just want to mention that Unity3D now have a new WebGL technology which is recommended to be used instead of WebPlayer. Considering that my PR was made exactly for Unity3D WebPlayer platform (yes, they have different API set for different "subplatforms", not just "one big Unity") I don't think anymore that it's necessary for you to accept this.

304NotModified:
OK Good to know. 

I Know NPAPI - and that's not a good thing ;)

I will close this PR then.

PR #464, opened 2014-12-11
Added 'enabled' attribute to the logging rule element.

### LINKED ISSUE #13: Add an Enabled attribute to the logger rules
I'd like a quick way to enable/disable specific loggers. Yes, I can comment them out but I'd prefer an additional attribute that tells NLog whether the logger should be active. See the example below.

```
<rules>
<logger name="*" minlevel="Error" writeTo="console" enabled="true"/>
<logger name="MyProblemNameSpace.*" minlevel="Trace" writeTo="f1" enabled="false" />
</rules>
```

Moved from Codeplex http://nlog.codeplex.com/workitem/5138

ilya-g:
Enabled attribute was added in revision 9a069ef3121b7d87edaaf1c40a7b5f6bc57886e2. Now it's supported in NLog.xsd schema too.
closes #13

### DIFF
--- a/tools/MakeNLogXSD/TemplateXSD.xml
+++ b/tools/MakeNLogXSD/TemplateXSD.xml
@@ -143,6 +143,12 @@
         <xs:documentation>Ignore further rules if this one matches.</xs:documentation>
       </xs:annotation>
     </xs:attribute>
+
+    <xs:attribute name="enabled" type="xs:boolean" default="true">
+      <xs:annotation>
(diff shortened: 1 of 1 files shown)


304NotModified:
@Xharze why is this milestone 4.0? The `enabled` attribute is already in the 3.x releases?

Xharze:
It believe this fix was introduced after the v3.2 release. But I'll try and
create an update for xsd file which can be released separately.

Issue #423, opened 2014-09-23
Level attribute not used in matching


pawelpabich:
Hi,

My understanding of https://github.com/nlog/NLog/wiki/Configuration-file#rules is that both logger name and log level are used during the matching phase. If not and only the name is used then please disregard this issue. 

I'm trying to disregard NSB debug output.  My expectation was that it is enough to add one more rule to the config file:

```
<logger name="NServiceBus.*" level="Debug"  final=true  />
```

Unfortunately that's not true and all log levels were captured by that rule. To work around this problem I had to basically copy and past all generic rules I had and scope them to NServiceBus.*.

```
  <logger name="NServiceBus.*" level="Debug"  />
    <logger name="NServiceBus.*" minlevel="Info" writeTo="Web"/>
    <logger name="NServiceBus.*" minlevel="Warn" writeTo="Database" final="true"/>
    
    <logger name="*" minlevel="Debug" writeTo="Web"/>
    <logger name="*" minlevel="Warn" writeTo="Database"/>
```

Am I doing something wrong here?

304NotModified:
The pull request has been merged. Thus I assume the issue has been fixed. If not, then just re-open the issue.

pawelpabich:
Thanks a lot! Are you planning to release a new version of nlog on nuget?

ilya-g:
It will be released in v4.0. @304NotModified, can you tag this issue with v4.0 milestone?

PR #467, opened 2014-12-12
LoggingRule.Final only suppresses matching levels. 

ilya-g:
Fix for #423

Added a test that reproduces the bug and passes after the change.

### DIFF
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -592,7 +592,7 @@ namespace NLog
             }
         }
 
-        internal void GetTargetsByLevelForLogger(string name, IList<LoggingRule> rules, TargetWithFilterChain[] targetsByLevel, TargetWithFilterChain[] lastTargetsByLevel)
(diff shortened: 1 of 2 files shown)


snakefoot:
I have an application with different 3rd Party Dependencies. I have created some very specific logger-name-star-filters with final for these dependencies (Info-level and above), and then everything else (my stuff) goes to the catch-all-star-logger.

Does this change mean that I have specify these 3rd Party dependencies logger-filters for each log-level, so they don't get caught by my catch-all-star-loggger?.

Is there a setting so I can specify "FinalAll" and not just "FinalLevel" for my 3rd Party Dependency logger-filters?

304NotModified:
Hi @snakefoot , please open a new issue for this (reference to the PR would be nice) and please post an example config.

snakefoot:
Will minLevel + Final for the 3rd Party Dependency loggers (Instead of just Level), make sure that nothing reaches my catch-all-start-logger?

304NotModified:
@snakefoot you can use `minlevel` or just remove the whole `level` attributes.

PR #471, opened 2014-12-19
Fix use of single archive in file target

### LINKED ISSUE #470: using a single archive file may not continue to rollover properly
My file target looks like this:

 ```xml
<target name="LogFile" xsi:type="File" 
   FileName="${specialfolder:CommonApplicationData}\AppName\logfilename.log"
   Layout="${longdate} | ${threadid} | ${level:uppercase=true} | ${logger} | ${message}" 
   archiveAboveSize="5242880"
   maxArchiveFiles="1"
   archiveFileName="${specialfolder:CommonApplicationData}\AppName\logfilename.lg2"
/>
```

The `logfilename.log` fills up and rolls over to `logfilename.lg2` but then logging stops.

`FileTarget.DynamicArchiveFileHandlerClass.AddToArchive(string archiveFileName, string fileName, bool createDirectoryIfNotExists)` had an empty `archiveFileEntryQueue`.  And so `File.Delete` wasn't being called on the archive file which caused the `File.Move` to throw an exception.

I've modified `AddToArchive` to delete the archive file if only one file is to be kept, and can submit a PR if that solution is good.

Is there a different way to configure the logging that I need?

jltrem:
delete the single file if only one archive file is used
fixes #470

### DIFF
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -105,6 +105,22 @@ namespace NLog.Targets
                     return false;
                 }
 
+                // Delete a single file if there is only one archive kept.
+                // This allows using a single archive file with a non-autonumbered name.
+                if (MaxArchiveFileToKeep == 1)
+                {
+                   archiveFileEntryQueue.Clear();
+
+                   try
+                   {
+                      File.Delete(archiveFileName);
+                   }
+                   catch (Exception exceptionThrown)
+                   {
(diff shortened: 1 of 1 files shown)


jltrem:
@Xharze ...a lot of water under the bridge since my PR.  I'll try to take a look this week to see if the issue may have been resolved by @ie-zero's FileTarget changes.

Xharze:
@jltrem No problem

304NotModified:
@Xharze unit tests for a (small) bug fix is a bit overkill?

PR #481, opened 2014-12-28
JsonLayout

vladikk:
A new layout that renders log events as structured JSON documents.
The implementation and setup are similar to the CsvLayout:

``` XML
<target name="jsonFile" xsi:type="File" fileName="${logFileNamePrefix}.json">
              <attribute name="time" layout="${longdate}" />
              <attribute name="level" layout="${level:upperCase=true}"/>
              <attribute name="callsite" layout="${callsite:includeSourcePath=true}" />
              <attribute name="stacktrace" layout="${stacktrace:topFrames=10}" />
              <attribute name="exception" layout="${exception:format=ToString}"/>
       </layout>
</target>
```

### DIFF
--- a/src/NLog/Layouts/JsonAttribute.cs
+++ b/src/NLog/Layouts/JsonAttribute.cs
@@ -1,7 +1,40 @@
-using NLog.Config;
+// 
+// Copyright (c) 2004-2011 Jaroslaw Kowalski <jaak@jkowalski.net>
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without 
+// modification, are permitted provided that the following conditions 
+// are met:
+// 
+// * Redistributions of source code must retain the above copyright notice, 
+//   this list of conditions and the following disclaimer. 
+// 
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution. 
+// 
+// * Neither the name of Jaroslaw Kowalski nor the names of its 
+//   contributors may be used to endorse or promote products derived from this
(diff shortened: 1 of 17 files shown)


grbinho:
Hi.
First of all I think this is a great addition to the library.

I'm trying to use it, but i have a problem with nested objects. If you want to add something like a property bag to event-context that is already serialized Json, this layout will escape it. 

Were nested objects a consideration for this feature or just plain key value pairs?

Ivan

grbinho:
Hi,

First of all, I think this is a great addition to NLog. 

I have a question about nested JSON objects.

Is this intended only of key value pairs? Can we nest our own JSON? Something like a property bag that we serialize and add to event-context. 

As far as i can see all attributes are not json-encoded. That means that attributes that are already valid JSON will get escaped where that is not needed or am i missing something?
It would be beneficial if encoding is a property on the attribute.

Best,
Ivan

304NotModified:
Hi,

We do you mean with 
> As far as i can see all attributes are not json-encoded.

So far I know, everything is encoded with the `JsonEncodeLayoutRendererWrapper`.

I think other objects besides key-value pairs would be nice. Please create a new issue for this feature request if you like. An (config) example would be nice.

grbinho:
Sorry, tipo :)

I have forked the project and implemented a small change.

I have added encode attribute on JsonAttribute so that user can control if output should be encoded. Setting this to false will enable user to add arbitrary JSON into context.

I'm still missing some tests, but can eventually create a pull request if you find it interesting.

304NotModified:
I find it interesting :+1: 

PR is welcome, but I does need some unit tests (and examples for the documentation)

grbinho:
Ok. Great, I will try to get that done as soon as possible.

PR #487, opened 2015-01-04
LogLevel Implements IEquatable<LogLevel>

ie-zero:
* LogLevel class now implements IEquatable<LogLevel>. This can
potentially help slightly the performance is certain cases as no object
boxing is required.
* Private default constructor was removed as it was obsolete.
* Unit tests added to cover Equals(Object) and Equals(LogLevel) methods.


Xharze:
Merged manually in fdfa3337749f529ea62f83d789187f9c64b4aa00

PR #489, opened 2015-01-05
Add .editorconfig

damageboy:
This allows developers who install the editor config vs extension:
https://visualstudiogallery.msdn.microsoft.com/c8bccfe2-650c-4b42-bc5c-845e21f96328

To automagically switch their tab/newline settings to match NLog's when opening the NLog project/solution

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+; Top-most EditorConfig file
+root = true
+
+; Unix-style newlines
+[*]
+end_of_line = LF
+
+; 4-column space indentation
+[*.cs]
+indent_style = space
+indent_size = 4


damageboy:
This is just a small, almost meaningless commit, but I have two other ones that are waiting for this to be merged, then I can rebase on it.

Xharze:
What do you plan to use this commit for? I'm wondering if NLog should
support all kinds of different extensions, so far we have ReSharper, and
with this commit, if accepted, also editorconfig. What could be next? I'm
not saying this won't be accepted, just curious about where you are going
with it.
Den 05/01/2015 18.52 skrev "damageboy" <notifications@github.com>:

damageboy:
I'm sorry, I didn't explain myself properly.
The other commits are completely unrelated to such things.
They are related to mono build + NuGet.
I just use editorconfig all around places so I don't mess up othert peoples projects when it comes to eol/tab/space configurations.
I wish VS / R# would support this on their own, but since they don't editorconfig is ok.

My other commits are about trying to change the #if MONO conditional compilations into runtime checks, so that the NuGet build would actually behave more like the Mono build when using NLog in Mono with NuGet (i.e. the official build).

Xharze:
The pull request has been merged

PR #490, opened 2015-01-08
AllEventProperties Layout Renderer

vladikk:
A new layout renderer that outputs all of the event's properties. Format and separator can be manually configured.
It allows more structured logging, without knowing all the possible property names in advance.

Usage examples:
* ${all-event-properties}
* ${all-event-properties:Separator=|}
* ${all-event-properties:Separator= | :Format=[key] is [value]}


304NotModified:
Looks nice! Can you fix the merge errors?

PR #496, opened 2015-01-14
Fix monodevelop build

dmitry-shechtman:


Xharze:
Thanks for the pull request it has ben merged

PR #497, opened 2015-01-14
Remove Windows Forms targets

### LINKED ISSUE #455: Move WCF and Winforms targets to separate assemblies
I am using NLog in my web app that runs on mono on Debian Linux. Because NLog references System.ServiceModel, System.Windows.Forms, and System.Drawing, I have to install the libmono-system-servicemodel4.0a-cil, libmono-system-windows-forms4.0-cil, and libmono-system-drawing4.0-cil packages, which themselves bring in other packages, including things like Cairo, GDI+, X, and Mesa, resulting in a couple dozen extra packages for functionality I don't even use. I think the main NLog package should only reference "core" assemblies to minimize dependencies.

This may become more relevant with ASP.NET vNext and .NET Core, which will not have System.Windows.Forms (and maybe not WCF?)

Obviously this is a breaking change, so maybe for 4.0?

Xharze:
These targets have been moved to seperate repository, https://github.com/NLog/NLog.Windows.Forms, and will be released in a seperate package.
Should not be merged until NuGet package exists for new package.

closes #455


ilya-g:
Since `RichTextBoxTarget` has been moved away, maybe some related stuff should also go with it, namely `RichTextBoxWordColoringRule` and `RichTextBoxWordColoringRule`?

Xharze:
Missed those, thanks.

Issue #274, opened 2013-11-21
UTC based archiving


adamralph:
When using UTC universally (:wink:) including timestamps on log messages, it would make more sense if time based archiving could also be set to use UTC. E.g.:

```xml
<target
  archiveEvery="Day:universalTime=true" />
```

304NotModified:
@mattAtCSL PR Would be great.

mattAtCSL:
Pull request #504.

adamralph:
@mattAtCSL great!

adamralph:
I've gotten completely lost in the various PR's which may or may not add this feature and have/have not been merged. I think if/when this feature is released the most important thing is that the wiki is updated so that it's clear how to switch on UTC based archiving.

mattAtCSL:
If they go for the fix where the same date time source is used for archiving as the time stamps on log entries you'll not need another setting, using a UTC time type will mean the archiving period matches the log entry times.

adamralph:
As long as the final solution is clearly communicated, I'll be happy :smile:

304NotModified:
I think the lasted PR (#512), should include this feature. Too bad the're a lot of failing unit tests with PR #512.

PR #504, opened 2015-01-19
Added UTC options to ArchiveEvery.

mattAtCSL:
Added support for archiving on UTC time, currently NLog lets you use UTC date and times in log entries but archiving files operates in whatever time zone the system is set to. By amending additional UTC versions to the end of the archive periods this should be a backwards compatible fix for adding UTC support to creating archives.   

Added UTCYear, UTCMonth, UTCDay, UTCHour, UTCMinute to the FileArchivePeriod enum.
In  FileTargetcs added check for UTC version of the enums and if so DateTime.UTCNow is used. Added UTC versions of the enum to the switch statements that get the date and formatted date string.

### DIFF
--- a/src/NLog/Targets/FileArchivePeriod.cs
+++ b/src/NLog/Targets/FileArchivePeriod.cs
@@ -66,6 +66,31 @@ namespace NLog.Targets
         /// <summary>
         /// Archive every minute.
         /// </summary>
-        Minute
+        Minute,
+
+        /// <summary>
+        /// Archive every UTC year.
+        /// </summary>
+        UTCYear,
+
+        /// <summary>
+        /// Archive every UTC month.
+        /// </summary>
+        UTCMonth,
+
+        /// <summary>
+        /// Archive UTC daily.
+        /// </summary>
+        UTCDay,
+
+        /// <summary>
+        /// Archive every UTC hour.
+        /// </summary>
+        UTCHour,
+
+        /// <summary>
+        /// Archive every UTC minute.
+        /// </summary>
+        UTCMinute
     }
 }
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -1024,10 +1024,12 @@ namespace NLog.Targets
                 switch (this.ArchiveEvery)
                 {
                     case FileArchivePeriod.Year:
+                    case FileArchivePeriod.UTCYear:
                         formatString = "yyyy";
                         break;
 
                     case FileArchivePeriod.Month:
+                    case FileArchivePeriod.UTCMonth:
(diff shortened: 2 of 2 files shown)


ilya-g:
Maybe we should use time from the current time source in `GetArchiveDate` function?
Then one can specify `<time xsi:type='FastUTC' />` in configuration and have both timestamps and date-based archiving refer to UTC time.

```C#

         private DateTime GetArchiveDate()
         {
             DateTime archiveDate = TimeSource.Current.Time;
             ...

304NotModified:
@mattAtCSL thanks for the PR.

Maybe I don't understand the impact well, but I expected a boolean for this and not altering the `FileArchivePeriod`? Something like 'UseUTC'. I'm not sure if this is related to @ilya-g

mattAtCSL:
@ilya-g that's sort of how I expected archiving would work when setting NLog to use UTC, only when British summer time started did I notice that the archiving was occurring an hour early due to GetArchiveDate always using DateTime.Now.

adamralph:
This is awesome, but it looks like you've committed against an incredibly old version of master, which is I guess why it has conflicts.

304NotModified:
Unit tests are really appreciated. Also I'm pretty sure that `FileArchivePeriod.UTCMinute` (compared to `FileArchivePeriod.Minute:`) doesn't do anything in the submitted PR

adamralph:
@304NotModified the condition is `>= FileArchivePeriod.UTCYear` but, given that the enum members are not explicitly numbered, it's relying on source declaration order, so perhaps a little fragile.

Your suggestion of a Boolean rather than new enum members may be a better approach.

304NotModified:
Ow yes, missed that spot. Didn't expected ´>=´ without explicit numbering indeed.

mattAtCSL:
Yes, using a bool would be a much simpler solution, you'd only need to switch between using DateTime.Now and UTCNow in GetArchiveDate.

ilya-g:
Proposed an alternate pull request #512 where the time is obtained from the current time source, which can be set to UTC.

Xharze:
I think the solution provided by @ilya-g is better, as it doesn't introduce more options to the target. I've actually merge the other pull request, but input is appreciated

mattAtCSL:
Feel free to close this pull request, it was a quick rush fix to get me out of a bind a while back, I'll move my project to the version with the changes from @ilya-g

adamralph:
I saw that #517 got merged, but #512 was supposedly the fix for #274 and is still open. Or do I misunderstand?

Xharze:
I'm a bit behind on the pull requests, because of problems with the build
server, this has now been fixed and #512 will be merged.

PR #505, opened 2015-01-19
Fix #492: RichTextBox target breaks row coloring

### LINKED ISSUE #492: RichTextBox target breaks row coloring when using MaxLines
Using NLog 3.2.0.0 and .Net Framework 4.0 (C#) 

I experience Issues using the MaxLines feature on the RichTextBoxTarget.

It looks like the code breaks the Rtf behaviour when removing parts with
`rtbx.Text = rtbx.Text.Remove(...);`
while the old implementation
`rtbx.SelectedText = string.Empty;`
does not work when the RichTextBox is ReadOnly

I suggest to use:
`rtbx.SelectedRtf = "{\\rtf1\\ansi}";`
which works on ReadOnly RichTextBox and does not break the coloring

See [stackoverflow remove-text-from-a-readonly-richtextbox](http://stackoverflow.com/questions/13318238/how-can-i-programmatically-remove-text-from-a-readonly-richtextbox)

I've tested the change with C# visualstudio 2013 .NET 4.0 and it works for that use-case.

related to #17 #324 #362 "RichTextBox target doesn't respect maxlines(keeps adding more and more lines)"

ehoeks:
Fix #492: RichTextBox target breaks row coloring when using MaxLines.
Using method decribed in "How can I programmatically remove text from a
ReadOnly RichTextBox?"
http://stackoverflow.com/questions/13318238/

### DIFF
--- a/src/NLog/Targets/RichTextBoxTarget.cs
+++ b/src/NLog/Targets/RichTextBoxTarget.cs
@@ -355,7 +355,7 @@ namespace NLog.Targets
                 {
                     rtbx.SelectionStart = 0;
                     rtbx.SelectionLength = rtbx.GetFirstCharIndexFromLine(1);
-                    rtbx.Text = rtbx.Text.Remove(rtbx.SelectionStart, rtbx.SelectionLength);
+                    rtbx.SelectedRtf = "{\\rtf1\\ansi}";
                     this.lineCount--;
                 }
             }


Xharze:
Thanks for the pull request, we're currently in the process of moving Windows.Forms specific targets into a separate assembly, see #497, I'll make sure this change is this change is introduced in the other assembly. Until then I'll keep this pull request open.

Xharze:
Fix has been introduced in NLog.Windows.Forms repository

Issue #307, opened 2014-02-07
Many 1kb log files generated when change system date for FileTarget


tanguohao:
1. Config the Nlog to daily archive and target is a FileTarget
2. While the system is running and many processes will write log to NLog, CHANGE THE SYSTEM DATE.
3. You will see there are a lot of 1KB log files.

Detailed description in the forum:
https://groups.google.com/forum/#!topic/nlog-users/-V4b0z4QQTc

ilya-g:
> `if (long.Parse(ts) < long.Parse(ts2))`

Date format string may be MMDDYYYY for example, or contain some non-digit characters, so parsing it to long and then comparing the resulting values may fail or give the incorrect result.

304NotModified:
We know the date format, so we can parse to datetime?

ilya-g:
Not sure if we can parse back date formatted with any possible date format, but with DateTime.TryParseExact and a reasonable fallback behavior this solution can be given a chance.

304NotModified:
It would be nice if someone could retest this with NLog 4.3 (RC)

304NotModified:
We think this is a non-issue. Don't change your time of your machine. So closing this.

Please let us know if some disagrees

PR #512, opened 2015-01-22
FileTarget uses time from the current TimeSource for date-based archiving

### LINKED ISSUE #274: UTC based archiving

ilya-g:
`FileTarget` uses the same time source which is used for timestamping log events. 
Therefore when one specifies `<time xsi:type='AccurateUTC' />` in configuration, he will have both timestamps and date-based archiving rely on UTC time.

Fixes #274.


adamralph:
Where exactly in the config file would this go?

ilya-g:
@adamralph, `time` element can be placed as a direct child element of `nlog` configuration element, for example:

```xml

  <time xsi:type="AccurateUTC" />

  <targets>
  </targets>

  <rules>
  </rules>
</nlog>
```

adamralph:
@ilya-g thanks. So do I understand correctly that, when `<time xsi:type="AccurateUTC" />` is present, it changes the default for https://github.com/NLog/NLog/wiki/Time-Layout-Renderer from local to UTC?

ilya-g:
@adamralph No it doesn't change properties of layout renderers, it provides date and time for log event timestamp from `DateTime.UtcNow`. This timestamp value is already converted to  `DateTimeKind.Utc`, so no matter what is the value of `universalTime` property of date and time layout renderers, the timestamp will be written in UTC format. This occurs due to these layout renderers only convert time from local to utc when `universalTime=true` is specified, and do not convert back to local, when it is not.

adamralph:
OK, I'm thoroughly confused :stuck_out_tongue_closed_eyes: 

I'll leave it the project owners to decide whether this works or not :wink:.

Has anything about this already been written up on the wiki? I could't find anything.

304NotModified:
I have to admit i'm also a bit confused @ilya-g.  This PR substitutes https://github.com/NLog/NLog/pull/504?

Xharze:
It seems everybody is a bit confused at the moment, including me. Could you create some tests for this to make sure it works as expected?

304NotModified:
And maybe explain why it is `AccurateUTC` and no `UTC`. I was somewhere covered in another PR, but I can't find that one. 

From a technical point the PR looks nice. 

@Xharze there are unit tests :)

Xharze:
Sorry I've missed the test.
At least we're sure we are missing some documentation on time sources.

ilya-g:
@304NotModified, @Xharze, in this commit I've created new unit test named `DeleteArchiveFilesByDateFromTimeSource` in `FileTargetTests.cs`.
It verifies that every time the day changes in the artificial UTC time source used in test, the log file is being archived. I don't expect it work somehow different with the real UTC time source, but to verify it, it would require whole day to run the test.
I think I can improve this test to make it clear, that log is archived before the first write on the new day.

ilya-g:
@304NotModified I'm also wondering why there is nothing about time sources in the documentation.
Robert Važan, the author of this feature, has covered this topic in three articles in his blog:

 - [NLog timestamps with millisecond accuracy](http://blog.angeloflogic.com/2013/10/nlog-timestamps-with-millisecond.html)
 - [How to configure NLog time source](http://blog.angeloflogic.com/2013/10/how-to-configure-nlog-time-source.html)
 - [How to write custom NLog time source](http://blog.angeloflogic.com/2013/10/how-to-write-custom-nlog-time-source.html)

I think they may be reworked and incorporated in the documentation.

ilya-g:
I've improved the unit test to cover several different parameter combinations of FileTarget. Unfortunately, it shows that neither my proposed change, nor the one from PR #504 is not sufficient to get this feature working (.

The FileTarget decides when to stop writing to the file and move it to the archive, based on the function `ShouldAutoArchiveBasedOnTime` result. This function compares `lastWriteTime` (last time the file was written) with the timestamp of the `logEvent` to write, and decides to archive if these values are different when formatted with `FileTarget.ArchiveDateFormat`.

ilya-g:
By the way, why AppVeyor says that "All is well", though I've added 15 failing unit tests?

Xharze:
It seems like the reporting of failed tests have stopped working, I'll look into this tomorrow

304NotModified:
@ilya-g strange indeed.  AppVeyor does report  the failing tests in the log

Xharze:
Appveyor build status now works. Could you fix the failing unit tests?

ilya-g:
I've managed to get tests working but it cost a kinda breaking change to the `TimeSource`: its implementations now have to provide the method `FromSystemTime`, which converts the system-originated time to the same form as that time source time values have.

This method now is used in the  `FileTarget.ShouldAutoArchiveBasedOnTime` function to convert file last write time before comparing it with the logEvent.
Also now all the implementations of BaseFileAppender rely on the system UTC time when determining last write time.

Please review the changes and especially the test to be correct.

304NotModified:
@ilya-g any ideas if #307 got fixed with this PR?

I think it's in `ShouldAutoArchiveBasedOnTime`

ilya-g:
@304NotModified, no, it doesn't. The condition of archiving is still the same.

Issue #119, opened 2012-10-07
Logger to extend an ILogger interface to ease unit-testing and mocking of loggers


Xharze:
<p>If I had one wish for NLog 2 it would be for an ILogger interface that Logger would implement. This would help greatly with testing, mocking, stubbing and wrapping of Logger objects.</p>
<p>&nbsp;</p>
<p>Coming from the Java world where every method is virtual by default I find trying to write tests in C# when using classes that aren&#39;t completely virtual or don&#39;t implement an interface quite frustrating.</p>

ilya-g:
@Xharze, could you tag this with v4.0 milestone, so we don't forget to get this done before releasing v4.0?

Xharze:
Done

304NotModified:
Done so close

PR #516, opened 2015-01-22
Removed non-generic ILogger interface methods

ilya-g:
Justification: methods defined in `ILogger-v1` and `ILoggerBase-v1` partial classes are redudant since for each of them there does exist a generic counterpart.

Consider for example `void Debug(string message, decimal argument)` or `void Debug(string message, long argument)` or `void Debug(string message, sbyte argument)`: there is one `void Debug<TArgument>([Localizable(false)] string message, TArgument argument)` which covers them all.

Why these methods are initially in the `Logger` class and why they shouldn't be introduced to the `ILogger` interface is explained in detail in this comment: https://github.com/NLog/NLog/issues/119#issuecomment-68323685


304NotModified:
Nice. Less is more. But i'm curious if this is has impact on the performance.

ilya-g:
@304NotModified Good catch, I must admit that interface generic method call is almost 5 times slower than interface non-generic method call with the same signature.

I've created microbenchmark to measure method call times. The results show no difference for concrete class instance method calls, but for interface method calls there is noticeable slowdown.
 
```
Instance Generic: 3,28 ns/call
Instance NonGeneric: 3,62 ns/call
Interface Generic: 14,58 ns/call
Interface NonGeneric: 3,78 ns/call
```

304NotModified:
Thanks for the benchmark.  Could you post the benchmark code for future reference?

With thousands of log statements,  5 times slower isn't good idea. 

I'm curious what problem were trying to solve. Improve maintainability?

PR #517, opened 2015-01-22
Fluent API uses the same time source for timestamping as the Logger.

ilya-g:
When new `LogEventInfo` is created by the `Logger` it uses `TimeSource.Current.Time` value as it's timestamp.
The `LogBuilder` does the same now.


Xharze:
Thank you for the pull request, it has been merged

PR #519, opened 2015-01-23
Fluent API available for ILogger interface

ilya-g:
Now fluent API extension methods can be applied to the `ILogger` interface.

I have to attribute every method receiving `ILogger` paramater with CLSCompliant(false), because the `ILogger` itself is marked as non-CLSCompliant.


304NotModified:
Thanks for explaining the `CLSCompliant(false)` :+1:

Issue #222, opened 2013-05-21
Feature Request: source code line number as first class Layout Renderer variable


beatgates:
Escalate from my comment in #128. Please add a Layout Renderer ${srcline} to refer to the source code line where the logging takes place.

I believe this is a very basic and helpful function in logging and debuging that most log4x framework has. So it deserves a first class variable rather than unintuitive ${callsite:lineno}.

I've been using log4j, log4net for a long time but new to NLog. So please correct me if I missed out and this feature does exist. Thanks.

304NotModified:
PR merged

PR #521, opened 2015-01-24
Added CallSiteLineNumber layout renderer

### LINKED ISSUE #128: Add $callsite option to output the line number only
>28.08.2009 09:20:44,926| Worker.DoSomething@84 | Working at: 28.08.2009

Please add $callsite option to output the line number only (without file name).

`${callsite:lineno=true}`

We prefer a compact output format like the following:

304NotModified:
Added separate renderer.

Adding `bool linenumber` at `CallSiteRenderer` would be not backwardscompatible or strange (true, but not rendered when `Filename` is false)

usage: `${callsite-linenumber}`

Closes #128 and #222.

### DIFF
--- a/src/NLog/LayoutRenderers/CallSiteLineNumberLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/CallSiteLineNumberLayoutRenderer.cs
@@ -1,0 +1,85 @@
+// 
+// Copyright (c) 2004-2011 Jaroslaw Kowalski <jaak@jkowalski.net>
+// 
+// All rights reserved.
+// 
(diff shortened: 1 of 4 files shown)


304NotModified:
Only added at `NLog.netfx45`. Not sure if this should be added to all solution files.

Xharze:
It should but I can do that when merging, I've got a tool for it

304NotModified:
:+1: nice tool!

304NotModified:
accidentally reverted by  945d35a23fcf2753e3ec6e81a332cbec830a94f2

PR #534, opened 2015-01-27
Auto load extensions to allow easier integration with extensions

kichristensen:
Because we have started moving a lot of targets into separate repositories and assemblies, I think it would be nice to allow using extensions more easily.
Auto loading assemblies following the naming format, NLog*.dll, could be a way. This is almost the same solution, used to load NLog.Extended automatically.
Please comment on this solution, this is not necessarily the final solution, or the solution at all.


Xharze:
Works locally but not on appveyor, and also wrong github account closing until I figure out why appveyor tests fails.

304NotModified:
Great idea! :+1:

PR #543, opened 2015-01-31
Auto load extensions to allow easier integration with extensions

Xharze:
Because we have started moving a lot of targets into separate repositories and assemblies, I think it would be nice to allow using extensions more easily.
Auto loading assemblies following the naming format, NLog*.dll, could be a way. This is almost the same solution, used to load NLog.Extended automatically.
Please comment on this solution, this is not necessarily the final solution, or the solution at all.

### DIFF
--- a/src/InstallNLogConfig/InstallNLogConfig.netfx35.csproj
+++ b/src/InstallNLogConfig/InstallNLogConfig.netfx35.csproj
@@ -71,9 +71,9 @@
     <Compile Include="Properties\AssemblyInfo.cs" />
   </ItemGroup>
   <ItemGroup>
-    <ProjectReference Include="..\..\src\NLog\NLog.netfx35.csproj">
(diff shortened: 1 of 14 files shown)


304NotModified:
:+1:

304NotModified:
With this feature, do we still need `<extensions />` in the config?

ilya-g:
@304NotModified, third-party extensions can be located not only in NLog.* assemblies, so yes we do.

304NotModified:
@ilya-g  Well that is something you can require ;)

Xharze:
We still need the extension element, mainly because this feature only works
if the extension assemblies are located in the same folder as NLog
assembly, meaning this won't work with extensions fx placed in the GAC

304NotModified:
Thanks for explaining about the extension element.

PR #547, opened 2015-01-31
Fix use of single archive in file target

### LINKED ISSUE #470: using a single archive file may not continue to rollover properly

Xharze:
Manual fix of merge conflicts. Code taken from #471
closes #471
fixes #470


304NotModified:
Can we unit test this?

Xharze:
It should be possible to unit test, I'll give it a shot

304NotModified:
@Xharze I think this can be merged?

Xharze:
It has been merged

PR #560, opened 2015-02-05
Archive file zip compression

aelij:
Adds the `EnableArchiveFileCompression` property to `FileTarget`, which zips the archive files to save space. Supported only in .NET 4.5 (using System.IO.Compression).
Includes unit tests.

### DIFF
--- a/src/NLog/NLog.netfx45.csproj
+++ b/src/NLog/NLog.netfx45.csproj
@@ -57,7 +57,6 @@
     <Reference Include="System.Data" />
     <Reference Include="System.Drawing" />
     <Reference Include="System.IO.Compression" />
-    <Reference Include="System.IO.Compression.FileSystem" />
     <Reference Include="System.Messaging" />
     <Reference Include="System.Runtime.Serialization" />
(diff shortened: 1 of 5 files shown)


304NotModified:
@aelij Any idea why the build fails? Can you commit something to restart AppVeyor?

aelij:
According to the AppVeyor log there was a problem retrieving NuGet packages. I don't think it had anything to do with my changes.

I've removed the reference to `System.IO.Compression.FileSystem` because it's not supported on all platforms (such as PCL), and I could manage without it.

304NotModified:
Yes, Nuget was down. So we needed a restart. 

With the new commit it's restarted and I builds! Thanks!

aelij:
Any idea when this will be released? :)
Thanks!

304NotModified:
In 4.0, but there are some [open issues](https://github.com/NLog/NLog/milestones/4.0) for that release.

304NotModified:
4.0 will be released on 2 June 2015.

Issue #561, opened 2015-02-05
Multicast and LogReceiverTarget


MartinTherriault:
I want to use the LogReceiverTarget with a WCF router service to be able to dynamically multicast the logs message.

One of the requirement for multicast is that the ServiceContract must only allow one way messages.
Right now the ILogReceiverServer contract does not meets the requirement because its only method (ProcessLogMessages) is not one way, but it already has a void return type.

Would it be possible to change the OperationContract attribute of said method to add "IsOneWay=true"? I already have tested this change and it works fine in my multicast scenario.

Otherwise I'll just re-implement it all in my own custom target if there would be any possible issue with this change.

Thanks

304NotModified:
A Pull Request will be accepted for this!

304NotModified:
PR is merged so closing this.

304NotModified:
Hi @MartinTherriault

We are reopening this issue because we having some problems with the merged PR.

- It's not backwards-compatible - but we thought it was. (see #565)
- We are missing a lot of properties in the `WcfLogReceiverClientFacade`
- We have a lot of code duplication.

I'm ATM busy writing some good unit tests for this (with really starting a WCF server). After that, I think we should refactor a lot to fix the issues. Can you help with that?

MartinTherriault:
Of course I will help. I looked at the issues and now believe the facade may not be the right approach. I think there would be two easy ways to fix it. In both case we would revert the changes.
The first way would be to add the new one way method to the service contract (I think it's not breaking, but would have to confirm) and send the log message on the right contract method specified in the target (one or two way).
The second way would be to create a new target for this and reuse most of the code already in place.


Date: Thu, 2 Jul 2015 14:31:23 -0700

304NotModified:
Thanks for the response!

About the possible solutions:

new one way method to current service contract
==
I was also thinking about this. But if I understand correct: we use another methodname for the oneway contract, and we have 2 interfaces? Because, if we have 1 interface, then this change would we a breaking change.

new target
==
I think we should look a this solution if the other isn't working (because of breaking changes for example) - because both targets would have almost the same functionality and should be the same class for an architectural point of view. 

I did also checked if we could change the WCF properties with an API instead of attributes, but that seems not be possible to bad.

MartinTherriault:
Something just hit me this morning, I've been thinking of this backward. NLog is a WCF client, so we cannot just add a method to the contract since the server side would no longer compile without implementing the new method.

[...]

Issue #106, opened 2012-10-07
ILogReceiverServer ProcessLogMessages


Xharze:
<p>Under the OperationContract attribute definition, IsOneWay = false (the default). I don&#39;t know if this is the intended behavior, because it prevents the ability to use MSMQ.</p>
<p>&nbsp;</p>
<p>&quot;If the IsOneWay property is set to false (the default), even methods that return void result in a reply message. In this case, the infrastructure creates and sends an empty message to indicate to the caller that the method has returned. (Using this approach enables the infrastructure to send SOAP faults back to the client.) Setting IsOneWay to true is the only way to cancel the creation and dispatch of a response message.&quot;</p>

304NotModified:
Closed by #565

PR #565, opened 2015-02-09
Set the service contract for LogReceiverTarget as one way

MartinTherriault:
Changed the service contract for the LogReceiverTarget to be one way. This will allow the messages to be multicast through a WCF routing service.

See issue #561 for more details.


Xharze:
Thanks for the pull request. Will this change be breaking?

MartinTherriault:
Yes you are right. I don't know what I was thinking. I will make some changes to allow the selection of the service contract through an optional property of the LogReceiverTarget, with the default value being the current two-way contract. This way it will be non breaking and will support the feature I need for multicast.

Xharze:
Thanks for the pull request, it looks good.

304NotModified:
Is #106 related?

MartinTherriault:
Yes #106 is about the same thing (IsOneWay property) in a different usage scenario (MSMQ). The changes I made should allow it to works with MSMQ when the new property of the target is set.

304NotModified:
Thanks @MartinTherriault

Issue #431, opened 2014-10-20
"Right" way to log exceptions


MgSam:
The onexception and exception layout renderers seem to only do anything if one of the logger methods with "exception" in the name are called. However, those logger methods with "exception" are now deprecated. 

So what is the "right" way to log exceptions now? Passing the exception into one of the normal log methods (Warn(), Error(), Fatal()) does not include the stacktrace as part of the message.

Maybe there should be an onerrororabove layout renderer?

fhoeben:
Having the exception as first parameter would work, but seems to break symmetry with other overloads for logging an exception that take it as last parameter. Would you mark all those as obsolete?

To me giving special meaning to the last argument, in case it is an exception, seems the most logical approach. Asking around in teams I'm involved in confirms that this is what people expect, as do some of the bugs I encounter :-). (I also noticed this is also the approach adopted by the Java logging framework slf4j, http://slf4j.org/faq.html#paramException).

The fact that Resharper (currently) would warn about an argument without placeholder in the message does not mean it should not be done. Resharper is a very good tool, but I believe its warnings/suggestions need to be evaluated by the programmer, not blindly followed. If the logging methods have such meaning for their parameters this might even trigger the Resharper team to create a new annotation or rule to help programmers use it correctly. They aim to assist programmers, not define what can and cannot be done in APIs.

304NotModified:
Reopen as inspired by #568.  I think this should be solved. I have also seen that some less experienced NLog users have problems to correctly log exceptions.  

I think the solution by @ilya-g (11 dec) Is the best solution.

304NotModified:
Reshaper is always a guideline, but IMO we have to think twice when Resharper complains.

304NotModified:
I will create a PR for this.

Xharze:
@304NotModified Could this issue be closed, now that #602 have been merged?

304NotModified:
Nope. #602 was only a structural change.

304NotModified:
@ilya-g , @Xharze 

I was working on this, but there is a problem:

```c#
public void Error(Exception exception, [Localizable(false)] string message, params object[] args)`
```

Looks similar to

```c#
  public void Error(IFormatProvider formatProvider, [Localizable(false)] string message, params object[] args)
```
       
because the call 

```
Error(null, "message {0}", 1)
```

now will give a `ambiguous invocation`. 

Any ideas on this?

ilya-g:
[...]

Issue #63, opened 2012-10-06
Add args overload to the Exception methods


Xharze:
All of the logger methods have an overload for (string message, params object[] args) EXCEPT the various Exception methods. That means I have to code something like this:

```
logger.FatalException(string.format("asfaf{0} {1}", var1, var2), exceptionVariable)
```

I'd prefer to have an overload of (Exception exception, string message, params object[] args). Then I could write my exception logging like this:

```
logger.FatalException(exceptionVariable, "asfaf{0} {1}", var1, var2)
```

I assume this overload is not present to avoid possible FormatException errors in the Exception logging process. That is to prevent problems in case the developer doesn't have the correct number of args like this:

```
logger.FatalException(exceptionVariable, "asfaf{0} {1} {2}", var1, var2)
```

or typos a brace like this:

```
logger.FatalException(exceptionVariable, "asfaf{0} {1]", var1, var2)
```

Those issues could be handled with a wrapper similar to the code below.

```c#
//wrapper to prevent string format mask/argument errors
private static string SafeStringFormat(string formatMask, params object[] formatArgs)
{
    string toReturn = string.Empty;
    try
    {
        toReturn = string.Format(formatMask, formatArgs);
    }
    catch (FormatException ex)
    {
        int index = 0;
        StringBuilder sb = new StringBuilder();
        sb.AppendFormat("Format Mask: {0}\n", formatMask);
        foreach (object arg in formatArgs)
        {
            sb.AppendFormat("{0}: {1}\n", index, arg.ToString());
            index += 1;
        }
        toReturn = sb.ToString();
        logger.Error("FormatMask or args are incorrect (check for matching braces)\n {0}", toReturn);
    }
    return toReturn;
}
```

304NotModified:
This is implemented by #691

PR #568, opened 2015-02-11
Ensure exception as last formatting parameter is picked up as exception

fhoeben:
It would be really nice if the string-formatting overload of Logger's methods pick up an exception as last parameter and allow it to be used as a 'real exception' that will be logged according the the formatting for an exception.

This will allow a usage such as:

        public void TestExceptionLog()
        {
            string name = "my name";
            try
            {

                throw new Exception("Broken");
            }
            catch (Exception e)
            {
                Logger.Error("Unable to work for {0}", name, e);
            }
        }

Where e will be picked up as an actual exception and not ignored. The only way to do this currently is as below, but I would prefer to use a formatting overload:

        Logger.Error("Unable to work for " + name, e);

I added unit test statements to validate which exception was captured (I did not see such test statements in the tests so far, but maybe I overlooked them). Furthermore I was unfortunately not able to build the solution directly in visual studio (and did not spend enough time figuring out how to do a correct build and full unit test run) so my changes might not be completely correct, but I believe the basic working should be ok.

### DIFF
--- a/src/NLog/LogEventInfo.cs
+++ b/src/NLog/LogEventInfo.cs
@@ -91,7 +91,9 @@ namespace NLog
         /// <param name="message">Log message including parameter placeholders.</param>
         /// <param name="parameters">Parameter array.</param>
         public LogEventInfo(LogLevel level, string loggerName, IFormatProvider formatProvider, [Localizable(false)] string message, object[] parameters) 
-            : this(level, loggerName, formatProvider, message, parameters, null)
+            : this(level, loggerName, formatProvider, message, parameters, parameters == null 
+                                                                            ? null
+                                                                            : parameters[parameters.Length - 1] as Exception)
         {
         }
 
--- a/src/NLog/Targets/DebugTarget.cs
+++ b/src/NLog/Targets/DebugTarget.cs
@@ -82,12 +82,19 @@ namespace NLog.Targets
         public string LastMessage { get; private set; }
 
         /// <summary>
+        /// Gets the last event rendered by this target.
+        /// </summary>
+        /// <docgen category='Debugging Options' order='10' />
+        public LogEventInfo LastEvent { get; private set; }
+
+        /// <summary>
         /// Increases the number of messages.
         /// </summary>
         /// <param name="logEvent">The logging event.</param>
         protected override void Write(LogEventInfo logEvent)
         {
             this.Counter++;
+            this.LastEvent = logEvent;
             this.LastMessage = this.Layout.Render(logEvent);
         }
     }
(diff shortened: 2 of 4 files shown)


ilya-g:
````C#
Logger.Error("Unable to work for {0}", name, e)
````
This overload would make resharper complain that `e` argument is not used in format string, so it's generally not a good idea to give special meaning to some of string.Format arguments.

ilya-g:
There were already some discussions, related to exception-taking overloads with format string arguments: see #63, https://github.com/NLog/NLog/issues/431#issuecomment-66605694.

304NotModified:
You can also use the fluent interface!

fhoeben:
How would one do this with the fluent interface?

fhoeben:
I added some thoughts to #431 comments.

DavidDeSloovere:
I'm not sure if this would work, but could you use <T> with type inference?
just typing some things here
public void Info<T>(string message, args, T exception) where T : Exception
and call it with
log.Info("this was not for working for {0}", name, ex)

304NotModified:
Thanks for the PR but I don't think this PR is a good solution. I'm now typing on a mobile so I can't easily explain this with many lines but I think the solution proposed in #431 (exception as first parameter) Is much more clear and less "magic" . Also is easier to retain backwards compatible

Xharze:
I agree, this may be too magic

304NotModified:
Closing this PR but the discussion is still open on #431

fhoeben:
The good things about this 'magic' approach:
- very minor change in the codebase (only 1 line updated in LogEventInfo, other changes are just unit tests)
- very unlikely that this breaks anyones code (even if last argument was intended as an exception to be included in the formatted message that still occurs, only change is that the exception will also be caught as a proper exception), and it might even fix some bugs (where people were expecting the exception to be properly caught)
- no 'new' API, just added feature of existing methods (actually I believe you could get rid of a few overloads of info(), etc since now one method definition captures all variations: `void info(string message, params object[] arguments)`
- consistent API (Exception always sent as last parameter). To achieve this with the Exception as first parameter (as suggested) requires new overloads on all methods taking an Exception so it may be passed as first parameter and marking all current ones as obsolete. That's quite a big change for all existing users...

304NotModified:
IMO code/API's should be transparent. Magic is the opposite of  transparent.

The issues I see:
- It's nontransparent (as said before)
- It's is non-backwards compatible change (same API, different behavior)
- I can affect the performance (10K log calls =  10K checks for datatype)

IMO the non-backwards compatible change is a big change for existing users.

PR #575, opened 2015-02-14
Event Log Target unit tests improvement

ilya-g:
I've noticed that EventLogTarget tests run relatively slow on my machine: approximately 5 sec per every test in the `EventLogTargetTests` fixture.

When I digged into the tests, I've found 2 issues:

1. The `EventLogTarget.Source` property is not specified explicitly, and that results in AppDomain name being used as the source name. When the unit test runner executes tests, it creates an app domain with a random name (namely, random Guid). So on every test run a new source is  created, thus polluting system registry with source names, that will never be used again.
To prevent this I've set fixed source name `NLog.UnitTests`.

2. Every test in this fixture required to scan the entire Application log to find events it had written before. I've found a different approach to verify that an event is created. The `EventLog.EntryWritten` event have a handler attached, which collects all events written to this log. Then these collected entries are used to find the event entry being tested.

### DIFF
--- a/tests/NLog.UnitTests/Targets/EventLogTargetTests.cs
+++ b/tests/NLog.UnitTests/Targets/EventLogTargetTests.cs
@@ -31,6 +31,8 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
+using System.Collections.Generic;
+using System.Threading;
 using NLog.Layouts;
 
 #if !SILVERLIGHT && !MONO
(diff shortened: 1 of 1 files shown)


ilya-g:
Seems that an approach with watching for `EventLog.EntryWritten` event is not suitable for AppVeyor environment, so I switched back to querying the eventlog, but now with the `System.Diagnostics.Eventing.Reader.EventLogReader` class.

Now these tests run almost instantaneously in Visual Studio and 2 times faster in AppVeyor.

304NotModified:
:+1:

PR #576, opened 2015-02-15
Instance property XmlLoggingConfiguration.DefaultCultureInfo should not change global state

ilya-g:
When running all tests at once I have `NLogTraceListenerTests` constantly failing with the following output:

````
NLog.UnitTests.NLogTraceListenerTests.TraceDataTests: Assert.Equal() Failure [d:\NLog\src\NLog.proj]                                                                   
Position: First difference is at position 21 [d:\NLog\src\NLog.proj]                                                                                                   
Expected: MySource1 Fatal 42, 3,14, foo 145 [d:\NLog\src\NLog.proj]                                                                                                    
Actual:   MySource1 Fatal 42, 3.14, foo 145 [d:\NLog\src\NLog.proj]                                                                                                    
   at NLog.UnitTests.NLogTestBase.AssertDebugLastMessage(String targetName, String msg) in d:\NLog\tests\NLog.UnitTests\NLogTestBase.cs:line 74 [d:\NLog\src\NLog.proj]
   at NLog.UnitTests.NLogTraceListenerTests.TraceDataTests() in d:\NLog\tests\NLog.UnitTests\NLogTraceListenerTests.cs:line 187 [d:\NLog\src\NLog.proj]                
````

The reason why it passes in AppVeyor is because the CurrentCulture has the same decimal separator as the InvariantCulture there and this difference is not seen.

I have examined this test thoroughly, but lauched alone it passes! So the reason why it fails may be in some interference with other tests run before it. That interfering test appeared to be the `CultureInfoTests`.

This seems strange: at first sight `CultureInfoTests.WhenInvariantCultureDefinedThenDefaultCultureIsInvariantCulture` test doesn't change any global state in the LogManager, but in reality it is so, because the `XmlLoggingConfiguration.DefaultCultureInfo` actually changes `LogManager.DefaultCultureInfo` delegate.

### DIFF
--- a/src/NLog/Config/LoggingConfiguration.cs
+++ b/src/NLog/Config/LoggingConfiguration.cs
@@ -41,6 +41,8 @@ namespace NLog.Config
     using System.Collections.ObjectModel;
     using System.Reflection;
 
+    using JetBrains.Annotations;
+
     using NLog.Common;
     using NLog.Internal;
     using NLog.Targets;
@@ -92,13 +94,13 @@ namespace NLog.Config
         public IList<LoggingRule> LoggingRules { get; private set; }
 
         /// <summary>
-        /// Gets or sets the default culture info use.
+        /// Gets or sets the default culture info to use as <see cref="LogEventInfo.FormatProvider"/>.
         /// </summary>
-        public CultureInfo DefaultCultureInfo
-        {
-            get { return LogManager.DefaultCultureInfo(); }
-            set { LogManager.DefaultCultureInfo = () => value; }
-        }
+        /// <value>
+        /// Specific culture info or null to use <see cref="CultureInfo.CurrentCulture"/>
+        /// </value>
+        [CanBeNull]
+        public CultureInfo DefaultCultureInfo { get; set; }
 
         /// <summary>
         /// Gets all targets.
--- a/src/NLog/LogEventInfo.cs
+++ b/src/NLog/LogEventInfo.cs
@@ -439,7 +439,7 @@ namespace NLog
             {
                 try
                 {
-                    this.formattedMessage = string.Format(this.FormatProvider ?? LogManager.DefaultCultureInfo(), this.Message, this.Parameters);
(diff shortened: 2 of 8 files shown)


304NotModified:
Note:  I had also some troubles before with cultures.  Most of the bugs where in the unit tests, not the code. 

It would be great to run AppVeyor (temp or beside) under a different culture. Any idea if that is possible @Xharze?

Xharze:
@304Modified I'll try ask the appveyor support team if this is possible.

304NotModified:
I think AppVeyor uses the culture set in the .config

304NotModified:
Note the trunk won't build now

304NotModified:
@ilya-g should the cleanup be done in `dispose`?

ilya-g:
> Note the trunk won't build now? Don't get it, it isn't merged yet?

@304NotModified technically, it is merged, but not to the master branch. Github creates 2 special branches for every pull request: refs/pull/*NNN*/head and refs/pull/*NNN*/merge. The latter is fetched and checked out on the build server.

This branch builds, but its tests do not pass: it is so intentionally, because I've yet committed only tests, that reproduce the problem, not the solution to the problem itself.

ilya-g:
The proposed solution is:

1. Make `LoggerConfiguration.DefaultCultureInfo` a plain property with an instance backing storage.
2. Expose this value via new `LogFactory.DefaultCultureInfo` readonly property.
2. Obsolete `LogManager.DefaultCultureInfo` property.
4. In the `Logger.WriteToTargets` modify the `LogEventInfo` before it is to be written: set the `FormatProvider` to factory's `DefaultCultureInfo` if it is not set before.
4. In the `LogEventInfo.CalcFormattedMessage` default to `CurrentCulture` if `FormatProvider` is not set.

I'll update this PR soon by pushing new commits with this solution.

304NotModified:
@ilya-g Thanks for explaining about the build status. I fixed the badge on the readme.md

304NotModified:
The proposed solution sounds solid. :+1:

304NotModified:
`Breaking change` because different (but correct) behavior after fix

PR #583, opened 2015-02-18
.gitattributes specifies which files should be considered as text

### LINKED ISSUE #572: Normalize line endings across the code base
I've noticed that some of the source files use CRLF as the line delimiter, some LF only and some mix LF with CRLF.
What is the recommened line delimiter for NLog repository?

ilya-g:
Specified in .gitattributes which files should be considered as text and have their line endings normalized.
closes #572

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,43 @@
+# All files: detect if file is text automatically
+
+* text=auto
+
+# Text files that should be normalized to LF
+
+*.cs text diff=csharp
+*.config text
+*.sln text
+*.csproj text
+*.md text
+*.sh text
+*.ps1 text
+*.cmd text
+*.bat text
+*.markdown text
+*.msbuild text
+
+# Binary files that should not be normalized or diffed
+
+*.exe binary
+*.dll binary
+*.pdb binary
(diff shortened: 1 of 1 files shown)


ilya-g:
After merging this change all files with inconsistent line endings would be subject to renormalization upon next time they will be commited.

If you decide to perform this renormalization right after the merge, you can use the following commands:

````
del .git/index
git reset
git add -u
````

The first one deletes the index, the second – restores that index from the repository, now applying eol-settings in .gitattributes, the third — stages all files with inconsistent line endings, now being marked as modified.

304NotModified:
So it's LF in the future? Need that on the docs.

Xharze:
Good idea, and thanks for the pull request. It will be merged just before
the release of 4.0
Den 18/02/2015 02.50 skrev "ilya-g" <notifications@github.com>:

ilya-g:
@304NotModified Text files would be stored with LF in repository, but on checkout they would have their line endings transformed accoding to `core.eol` setting value — by default it is CRLF for Windows and LF for Mac/Unix.
So generally LF in repository, `core.eol` in your working copy.

PR #597, opened 2015-02-21
Added DEBUG-Conditional trace and debug methods

### LINKED ISSUE #408: Provide DEBUG-Conditional trace and debug methods
A while ago I submitted PR #286, which was to add [Conditional("DEBUG")] on the Trace overloads. The rationale was for extremely latency- or CPU-sensitive applications, the NLog logic that checks whether TRACE is enabled is sometimes too expensive and should not be compiled into release builds. That PR was (understandably) rejected because not all users are latency-sensitive and it's a pretty big change for people already using trace.

Here's a better alternative... What about adding ConditionalTrace() and ConditionalDebug() versions on Logger, with [CONDITIONAL("DEBUG")] on them? This wouldn't break backwards-compatibility in any way, while providing this feature to latency-sensitive users. We currently have this implemented internally as extension methods to Logger, if you think this is a good idea I'll submit a PR to integrate this logic into Logger.

304NotModified:
In separate file. 

Fixed #408

Notes: Not unit tested. No changes to current classes (only additions)


304NotModified:
`SC : error CS1606: Assembly signing failed; output may not be signed -- Error signing assembly -- The system cannot find the file specified. [C:\projects\nlog-134\src\NLog\NLog.netfx40.csproj` 

don't get this

304NotModified:
Would it be wise if I add those to the `Ilogger` interface of a new interface?

304NotModified:
This is ready for merge.

304NotModified:
It was mergeable before, but now I can't get it fixed. Newlines issue I guess, locally there are no changes when I merge with the master. 

I will create a new PR

PR #600, opened 2015-02-21
Add variables property

### LINKED ISSUE #518: LogManager.Configuration.Variables not available in 3.2.0
Hey.

I was very happy when I read about this feature in the 3.2.0 announcement. But it seems it is not available in the package. (nuget)

Could someone update the packages if it is indeed available in the code base?

Thanks

Xharze:
Otherwise LogManager.Configuration cannot be used to read variables
defined in XML

closes #518

### DIFF
--- a/src/NLog/Config/LoggingConfiguration.cs
+++ b/src/NLog/Config/LoggingConfiguration.cs
@@ -65,6 +65,20 @@ namespace NLog.Config
         }
 
         /// <summary>
+        /// Gets the variables defined in the configuration.
+        /// </summary>
+        /// <remarks>
+        /// Returns null if not configured using XML configuration.
+        /// </remarks>
+        public virtual Dictionary<string, string> Variables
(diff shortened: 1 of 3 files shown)


304NotModified:
I find the `return null` a bit unexpected.  

I'm also curious if we could change the variables programmatically.  The `return null` Will make that difficult?

Xharze:
The variables property is only available when NLog is configured using XML.
Configuring programmatically has never supported variables, therefore I
think it is important to show variables won't work when configured
programmatically.

304NotModified:
What do you think about a notsupportedException? 

Otherwise it looks like a bug.

Xharze:
That would be a good idea, I'll update the pull request

304NotModified:
bump @Xharze

304NotModified:
We need this also in 4.0 :)

PR #601, opened 2015-02-21
Mailtarget allow empty 'To' and various code improvements

### LINKED ISSUE #599: MailTarget: remove mandatory To requirement
In the code we get an error when 'To' is empty when sending a mail.

But why? I think is also acceptable to send an mail with only a CC or BCC?

See `MailTarget.SetupMailMessage`

304NotModified:
- Allow empty 'To' (but at least To, CC or Bcc) should be set. Fixes #599
- Created a single method to create the message (was before scatterd)
- Removed some code duplication on adding the addresses (parsing and
rendering)
- Always log the error in the log
- added some comments


304NotModified:
Ready for merge

304NotModified:
Added some more code improvements and refactoring

PR #602, opened 2015-02-21
Logger overloads generated by T4

304NotModified:
Great improvement in code quality, the logger methods for the six levels (Trace, Debug, Info etc) are now generated with a T4 template. (`.tt` file)

This will reduce the lines of code that should be maintained and guarantees the consistency between the overloads. The lines of code in the `Logger` class reduced from ~2000 to ~500 + ~300 (T4 template)

T4 templates are run at build time and the T4 generator is built in VIsual Studio 2012+ 

The output is compared with a text diff and nothing has changed in the signature of the methods (= the same methods, the same parameter names, the same attributes). So **nothing** functional has changed!

PS: there was only 1 change, beside some spaces, the code of `Error<TArgument>` was inconsistent. This is now fixed. (see screenshot, left=old, right=new)

image


304NotModified:
Note did this change to easily implement #431 in the near future.

304NotModified:
Probably the T4 is too fancy to handle. 

Issue 1
---
The "sync projects" corrupts the `.csproj` files. (`NLog.proj /t:SyncProjectItems`)

before

```xml
    <Compile Include="Logger1.cs">
      <AutoGen>True</AutoGen>
      <DesignTime>True</DesignTime>
      <DependentUpon>Logger.tt</DependentUpon>
    </Compile>
```
after

```xml
    <Compile Include="Logger1.cs" />
```

Issue 2
---
The unit test `SourceCodeTests.VerifyNamespacesAndClassNames` cannot handle this. It's convenient to only add the template code in the template file. So another file is generated with the `Logger` class. This is no problem, because for this case Microsoft gave us partial classes.
To bad this unit test is so picky.

304NotModified:
PR is ready for merge

DavidDeSloovere:
When can we expect a new release with the exception logging fix?

304NotModified:
2 June 

See https://github.com/NLog/NLog/milestones

Issue #390, opened 2014-07-18
delete old log file with date being part of log file name


slav:
From docs on File-Target in One log file per day section:
```
        <target name="file" xsi:type="File"
            layout="${longdate} ${logger} ${message}" 
            fileName="${basedir}/${shortdate}.log" />
```
This is exactly what I want to do. BUT, I want to only keep last 30 files, and delete file from 31 days ago.

I tried to test it with this setup
```
		<target name="fileAll" xsi:type="File" fileName="${basedir}/${date:format=yyyyMMdd-HHmm}.log"
		layout="${defaultLayout}" archiveEvery="minute" maxArchiveFiles="2" archiveFileName="${basedir}/{#}.log" archiveDateFormat="yyyyMMdd-HHmm" archiveNumbering="Date"/>
```

But files didn't get deleted. It's as if archive isn't working in this specific case. Using NLog 3.1.

Any advice on how to set it up to work?

The reason for this setup is because I want the logfile to be shipped. For log file shipped to have less problems it's ideal for the file to never be renamed after it's created. So, I don't want to have log.txt that later becomes log.{date}.txt. I want it to be log.{date}.txt from the beginning and only to be deleted in 30 days.

Thank you.

rbouny:
+1

testzcrypto:
+1

304NotModified:
moved to 4.2 as it's not clear how to fix this (we need to investigate further)

304NotModified:
We fixed some code related to this in 4.1. We should test after release.

304NotModified:
I'm working on this. Thanks for the +1. 

We are now using http://nlog.userecho.com/ :)

304NotModified:
I can confirm this bug has been fixed in 4.1. See unit tests in https://github.com/NLog/NLog/pull/986 (Tested in 4.1 and 4.0, and some relevant tests are failing in 4.0 and succeeding in 4.1)

PR #608, opened 2015-02-26
Archiving system runs when new log file is created #390

awardle:
If archiving mode is set to date when a new log file is created, due to naming rules, the program will delete any
out of date archives.

Summary of changes
--------------------------
* Moved code to delete old archives in date mode to separate function 
* Moved code to get file name pattern into a separate function
* Added code to write function that runs the delete archive function when a new log file is started
* Added unit test for bug 390

Unit test for bug 390 is commented out as it takes 5 minutes to run it. This is due to the fact the minimum time period for archiving is a minute.

Fixes bug 390

### DIFF
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -630,15 +630,9 @@ namespace NLog.Targets
 #endif
             byte[] bytes = this.GetBytesToWrite(logEvent);
 
-            if (this.ShouldAutoArchive(fileName, logEvent, bytes.Length))
-            {
-                this.InvalidateCacheItem(fileName);
-                this.DoAutoArchive(fileName, logEvent);
-            }
-
             // Clean up old archives if this is the first time a log record has been written to
             // this log file and the archiving system is date/time based.
-            if (this.ArchiveNumbering == ArchiveNumberingMode.Date  && this.ArchiveEvery != FileArchivePeriod.None)
+            if (this.ArchiveNumbering == ArchiveNumberingMode.Date && this.ArchiveEvery != FileArchivePeriod.None)
             {
                 FileInfo fileInfo = new FileInfo(fileName);
                 if (!fileInfo.Exists)
@@ -647,7 +641,12 @@ namespace NLog.Targets
                     this.DeleteOldDateArchive(fileNamePattern);
                 }
             }
-            
+
+            if (this.ShouldAutoArchive(fileName, logEvent, bytes.Length))
+            {
+                this.InvalidateCacheItem(fileName);
+                this.DoAutoArchive(fileName, logEvent);
+            }
 
             this.WriteToFile(fileName, bytes, false);
         }
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -641,8 +641,7 @@ namespace NLog.Targets
             {
                 if (this.previousFileName != fileName)
                 {
-                    FileInfo fileInfo = new FileInfo(fileName);
-                    string fileNamePattern = this.GetFileNamePattern(fileName, logEvent, fileInfo);
+                    string fileNamePattern = this.GetFileNamePattern(fileName, logEvent);
                     this.DeleteOldDateArchive(fileNamePattern);
                     this.previousFileName = fileName;
                 }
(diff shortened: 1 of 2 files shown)


awardle:
There is currently a bug in the patch which is causing 1 too many files to be deleted by the archiver sometimes.

tests/NLog.UnitTests/Targets/FileTargetTests.cs
```diff
@@ -562,6 +562,71 @@ public void DeleteArchiveFilesByDate()
             }
         }
 
+        // Test commented out as it takes 5 minutes to run
+        /*
+        [Fact]
+        public void DeleteArchiveFilesByDateWithDateName()
+        {
+            var tempPath = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString());
+            var tempFile = Path.Combine(tempPath, "${date:format=yyyyMMddHHmm}.txt");
```
ilya-g: You can use the format `yyyMMddHHmmss` to have new file created every second, so this test would require 5 seconds to run instead of 5 minutes.

src/NLog/Targets/FileTarget.cs
```diff
@@ -630,6 +630,18 @@ protected override void Write(LogEventInfo logEvent)
 #endif
             byte[] bytes = this.GetBytesToWrite(logEvent);
 
+            // Clean up old archives if this is the first time a log record has been written to
+            // this log file and the archiving system is date/time based.
+            if (this.ArchiveNumbering == ArchiveNumberingMode.Date && this.ArchiveEvery != FileArchivePeriod.None)
+            {
+                FileInfo fileInfo = new FileInfo(fileName);
+                if (!fileInfo.Exists)
```
ilya-g: Isn't it quite expensive to check whether the file exists on every write?

304NotModified: Good catch @ilya-g

awardle:
Function now checks if the filename of the log file has changed since the last write to check if it is a new log file instead checking if the file exists. I think this should be less expensive than checking if the file exists.

src/NLog/Targets/FileTarget.cs
```diff
@@ -1160,20 +1188,7 @@ private void DoAutoArchive(string fileName, LogEventInfo eventInfo)
...
-                fileNamePattern = Path.ChangeExtension(fi.FullName, ".{#}" + ext);
-            }
-            else
-            {
-                //The archive file name is given. There are two possibiliy 
-                //(1) User supplied the Filename with pattern
-                //(2) User supplied the normal filename
-                fileNamePattern = this.ArchiveFileName.Render(eventInfo);
-            }
+            string fileNamePattern = GetFileNamePattern(fileName, eventInfo, fi);
```
304NotModified: Please no 'fi' as name

src/NLog/Targets/FileTarget.cs
```diff
@@ -1210,6 +1225,25 @@ private void DoAutoArchive(string fileName, LogEventInfo eventInfo)
             }
         }
 
+        private string GetFileNamePattern(string fileName, LogEventInfo eventInfo, FileInfo fi)
```
304NotModified: Why do we have both filename and fileinfo as input?

awardle:
I've improved the documentation of the functions that I created and renamed some variables. In addition I have removed the fileInfo parameter from the GetFileNamePattern method.

304NotModified:
Looks good

src/NLog/Targets/FileTarget.cs
```diff
@@ -630,6 +635,18 @@ protected override void Write(LogEventInfo logEvent)
 #endif
             byte[] bytes = this.GetBytesToWrite(logEvent);
 
+            // Clean up old archives if this is the first time a log record has been written to
+            // this log file and the archiving system is date/time based.
+            if (this.ArchiveNumbering == ArchiveNumberingMode.Date && this.ArchiveEvery != FileArchivePeriod.None)
+            {
+                if (this.previousFileName != fileName)
```
ilya-g: How would this work when the `FileName` contains not only the date layout but some other dynamic layout renderer, for example some client id or logger? Then this condition could be `true` almost on every write.

awardle: Would it be better then if it stored in some sort of collection the previous log filenames instead of only the last log filename? Then the archiver cleaner (`DeleteOldDateArchive`) runs if the log file that is being written to is not in the collection.

awardle:
It now stores by default the last 20 file names used, although this can be changed in the configuration file, instead of just the last one. This will cause the `DeleteOldDateArchive` to run less often if the log file is changing between several different names.

Issue #212, opened 2013-04-24
WebServiceTarget writes encoding preamble (e.g. UTF BOMs)


fryburger:
When using WebServiceTarget with Protocol HttpPost, default encoding is UTF-8. Internally the requests body is written by a StreamWriter, that always prepends some encoding mark bytes (e.g. FE FF for UTF-16). This unfortunately may result in odd behavior, when using standard serializers in receiving ASP.NET services, that ain't aware of such encoder preambles and sometimes deserialize models not completely or even wrong.

My current workaround is, after I configured NLog, parse through the 'AllTargets'-property for a WebServiceTarget and set the Encoding manually like this:

		private static void patchWebServiceTargetEncodingPreambleIssue()
		{
			LogManager.Configuration.AllTargets
				.OfType<global::NLog.Targets.WebServiceTarget>()
				.ToList()
				.ForEach(p => p.Encoding = new System.Text.UTF8Encoding(false));
		}

This creates a UTF-8 encoding for every WebServiceTarget, without written preamble.

Maybe it'd be possible to change that behavior of WebServiceTarget?

304NotModified:
Needs probably http://msdn.microsoft.com/en-us/library/s064f8w2.aspx

304NotModified:
3 reports of the same issue. Well this should be fixed in 4.0

304NotModified:
Note: there are multiple workarrounds:

- See Issue start post
- #250 has another workarround

#380 has a nice example

Issue #380, opened 2014-06-27
Url encoding with WebserviceTarget and HttpPost


yetanotherchris:
I've created a WebService target and am calling an MVC controller using it, but it looks like NLog is incorrectly URL encoding a space (0xFEFF). Here's an example of the form that gets sent across:

	{%ufeffempty=&guid=336cec87129942eeabab3d8babceead7&m=Debg&
	date=2014-06-26+23%3a15%3a14.6348&logger=TestClient.Program&
	level=Debug&machinename=DELL}

The config file for the logger client is:

	<targets>
		<target xsi:type="WebService"
				name="webservice"
				url="http://localhost:57953/Home/Foo2"
				protocol="HttpPost"
				encoding="UTF-8"
				methodName="Foo">
			<parameter name="empty" type="System.String" layout=""/> 
			<parameter name="guid" type="System.String" layout="${guid}"/>
			<parameter name="m" type="System.String" layout="${message}"/>
			<parameter name="date" type="System.String" layout="${longdate}"/>
			<parameter name="logger" type="System.String" layout="${logger}"/>
			<parameter name="level" type="System.String" layout="${level}"/>
			<parameter name="machinename" type="System.String" layout="${machinename}"/>
		</target>
	</targets>

304NotModified:
This is BOM issue (FEFF = UTF-8 BOM). And so this is a duplicate of #212. 

This hasn't been fixed yet. An unit test will boost the progress ;)

Issue #250, opened 2013-07-31
WebService Target httpPost body starts with BOM in case of UTF-8


skironDotNet:
Hi Jarek (co slychac)

Could you fix stream to not include BOM, it messes up first parameter of httpPost body

here is more details about the problem (from someone else):

http://nlog-forum.1685105.n2.nabble.com/WebService-target-HttpPost-and-HttpGet-not-working-td7577418.html

304NotModified:
Duplicate of #212

PR #629, opened 2015-03-05
BOM option/fix for WebserviceTarget + code improvements

304NotModified:
Fixes: #212, #250 (duplicate), #380 (duplicate)

- option to include BOM in `WebserviceTarget ` - UTF-8 only. Doens't do anything with a different encoding. BOM is disabled by default. (breaking change). Property `IncludeBOM`. Possible values; 
 - `null` (dont change BOM), 
 - `true` (always include UTF-8 bom on UTF-8 encodings), 
 - `false` (**default**, always skip bom on UTF-8 encodings)
- no `Stream` to `byte[]` and back to `Stream`. Just write direct from `Stream` to `Stream`
- removed code duplication `WebserviceTarget `
- documented code `WebserviceTarget `
- made `WebserviceTarget ` mockable with internal method. See unit tests.
- unit tests for `WebserviceTarget ` including async testing


304NotModified:
@ilya-g short code review would be appreciated :)

ilya-g:
Nice refactoring in `WebServiceTarget`.

It was suprise for me that you have abandoned your idea about 'utf-8-no-bom' special encoding name. I've made checkout at the commit https://github.com/304NotModified/NLog/commit/81cb6721f4dc67f4a96435b8d0f482dc43867fd1 and have all its tests passing. Why do you feel that solution was wrong?

tests/NLog.UnitTests/Targets/WebServiceTargetTests.cs
```diff
@@ -0,0 +1,244 @@
...
+            var streamMock = new StreamMock();
+
+            //event for async testing
+            var counterEvent = new CountdownEvent(1);
+
+            var parameterValues = new object[] {"", "336cec87129942eeabab3d8babceead7", "Debg", "2014-06-26 23:15:14.6348", "TestClient.Program", "Debug", "DELL"};
+            target.DoInvoke(parameterValues, c => counterEvent.Signal(), request,
+                callback =>
+                {
+                    var t = new Task(() => { });
```
ilya-g: It's very strange that this line was compiled successfully in NLog.UnitTests.*netfx35* project, since tasks were introduced to BCL starting from .net 4.0.

304NotModified: image


....

304NotModified: @ilya-g About the "utf8-nobom"-solution, it wasn't good enough because:

- The unit tests were sometimes wrong, it was testing the preamble (BOM) of the 'Encoding', but it doesn't tell if we are really writing the preamble. 
- The default of the `WebserviceTarget` should be no-BOM. (because of the issues). This is problematic because some users configured `UTF-8` as encoding.
- The `FileTarget` was already writing without BOM, if `utf-8` was configured. I didn't want to change it.

ilya-g:
> The default of the WebserviceTarget should be no-BOM. (because of the issues). This is problematic because some users configured UTF-8 as encoding.

I don't think there are users who configured UTF-8 for WebServiceTarget in NLog.config: just because of the issues they all are using some workarounds to get the request formatted without BOM. Or in case if they nevertheless have used UTF-8 in config, they must be happy with BOM now.

> The FileTarget was already writing without BOM, if utf-8 was configured. I didn't want to change it.

Yes, it's a breaking change, but it makes some sense: it would be not bad to have byte order mark in the beginning of the file to help the readers to identify the file encoding. And again this behavior could be opted out.

304NotModified:
> > The default of the WebserviceTarget should be no-BOM. (because of the issues). This is problematic because some users configured UTF-8 as encoding.

>I don't think there are users who configured UTF-8 for WebServiceTarget in NLog.config: just because of the issues they all are using some workarounds to get the request formatted without BOM. Or in case if they nevertheless have used UTF-8 in config, they must be happy with BOM now.

Well I see them in the bug reports. But it's now more consistent: setting UTF-8 in `FileTarget` and `WebserviceTarget` is now UTF-8, without BOM.

>> The FileTarget was already writing without BOM, if utf-8 was configured. I didn't want to change it.

>>Yes, it's a breaking change, but it makes some sense: it would be not bad to have byte order mark in the beginning of the file to help the readers to identify the file encoding. And again this behavior could be opted out.

I didn't changed anything now with the `Filetarget`. So it won't write BOM now (also not possible to configure yet)

304NotModified:
I think this PR is good enough.

Issue #532, opened 2015-01-26
Callsite Layout Renderer and Caller Attributes


304NotModified:
The current `CallSiteLayoutRenderer` uses the stacktrace for the method name, line number and file name.

This could be done with the [Caller attributes](https://msdn.microsoft.com/en-us/library/hh534540.aspx). In theory this would be a performance improvement. 

This should be tested in a prototype.

Keep in mind that the current classname isn's supported with the 'Caller Attributes'.

From a quick scan, this should be done:
- Extend **all** the methods in the `Logger` with the following parameters:

```
 [CallerMemberName] string memberName = "",   [CallerFilePath] string sourceFilePath = "",
        [CallerLineNumber] int sourceLineNumber = 0
```

Extend the `LogEventInfo` with the three properties so that the `CallSiteLayoutRenderer` could use this information. 

Inspired by #509

sorin-bolos:
I gave the Caller attributes a try, using the approach sketched above. 

Even after obfuscating with .NET Reactor the method name, file name and line line number look ok, but the class name is still obfuscated because there is no attribute for that.

When using 
`layout="${longdate:universalTime=true} | ${uppercase:${level}} | ${callsite} -> ${message}"`
 I get:

*2015-03-08 03:16:41.5883 | ERROR | At09ooH3UChoDESHRc.OWt56Q4H8GFKUfs6n9.CallLogger -> my error log*

When using 
`Layout = "${longdate:universalTime=true} | ${uppercase:${level}} | ${callsite:className=false:fileName=true:includeSourcePath=false:methodName=false} -> ${message}" `
(like in  #509)

i get 

*2015-03-08 03:12:09.1576 | ERROR | (Program.cs:49) -> my error log*

304NotModified:
There is a problem with binary incompatibility with this. This will force users to convert ALL there code in one solution to NLog5 (if they move from NLog 4).

This is bad.

FYI:

if we have both in the interface, like

```c#
public interface ILogger2
{
  void Info(string message);
  void Info(string message, [CallerMemberName] string memberName = "");
}
```

Then the call to `Info("test")` will invoke the one without the caller attributes (also in .Net 4.5 / 4.6)

snakefoot:
Partially resolved with #2386. Now the callsite-layout-renderer will attempt to use  Caller Attributes (when available)

Still missing the optimization of not capturing StackTrace when actually having the needed information from Caller-Attributes.

snakefoot:
[...]

PR #632, opened 2015-03-09
Added caller attributes for a single log method (WIP)

sorin-bolos:
Applies to #532.

I added Caller attributes to a single method from the Llogger class:
 `public void Log(LogEventInfo logEvent, [CallerMemberName] string memberName = "", [CallerFilePath] string sourceFilePath = "", [CallerLineNumber] int sourceLineNumber = 0)`

I also made the necessary changes to `LogEventInfo` and `CallSiteLayoutRenderer`.

I can extend the changes to all Logger public methods but I wanted to try with a single method firs.

I think the existing unit tests cover the functionality well enough, but I can add tests if needed.

### DIFF
--- a/src/NLog/ILoggerBase.cs
+++ b/src/NLog/ILoggerBase.cs
@@ -36,6 +36,9 @@ namespace NLog
     using System;
     using System.ComponentModel;
     using JetBrains.Annotations;
+#if NET4_5
+    using System.Runtime.CompilerServices;
+#endif
 
     public partial interface ILoggerBase
     {
@@ -69,11 +72,27 @@ namespace NLog
 
         #region Log() overloads
 
+#if NET4_5
         /// <summary>
         /// Writes the specified diagnostic message.
         /// </summary>
         /// <param name="logEvent">Log event.</param>
-        void Log(LogEventInfo logEvent);
+        /// <param name="memberName">Method or property name of the caller.</param>
+        /// <param name="sourceFilePath">Full path of the source file that contains the caller.</param>
+        /// <param name="sourceLineNumber">Line number in the source file at which the method is called.</param>
+#else
+        /// <summary>
+        /// Writes the specified diagnostic message.
+        /// </summary>
+        /// <param name="logEvent">Log event.</param>
+#endif
+        void Log(LogEventInfo logEvent
+#if NET4_5
+           , [CallerMemberName] string memberName = ""
+           , [CallerFilePath] string sourceFilePath = ""
+           , [CallerLineNumber] int sourceLineNumber = 0
+#endif
+            );
 
         /// <summary>
         /// Writes the specified diagnostic message.
--- a/src/NLog/LayoutRenderers/CallSiteLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/CallSiteLayoutRenderer.cs
@@ -130,89 +130,79 @@ namespace NLog.LayoutRenderers
         /// <param name="logEvent">Logging event.</param>
         protected override void Append(StringBuilder builder, LogEventInfo logEvent)
         {
-            StackFrame frame = logEvent.StackTrace != null ? logEvent.StackTrace.GetFrame(logEvent.UserStackFrameNumber + SkipFrames) : null;
-            if (frame != null)
+            if (this.ClassName)
             {
-                MethodBase method = frame.GetMethod();
-                if (this.ClassName)
-                {
-                    if (method.DeclaringType != null)
-                    {
-                        string className = method.DeclaringType.FullName;
-
-                        if (this.CleanNamesOfAnonymousDelegates)
-                        {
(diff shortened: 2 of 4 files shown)


src/NLog/Logger.cs
```diff
@@ -95,14 +96,35 @@ public bool IsEnabled(LogLevel level)
...
+#endif
+            )
         {
             if (this.IsEnabled(logEvent.Level))
             {
+#if NET4_5
+                logEvent.CallerMethodName = memberName;
+                logEvent.CallerSourceFilePath = sourceFilePath;
+                logEvent.CallerLineNumber = sourceLineNumber;
+#endif
```
304NotModified: U'm not sure if we should capture this for **every** LogEvent.  The `sourceFilePath` can be pretty long. For example, this can be easily 100 chars. If we printing just a small message (say 10 chars)  without the need of the caller info, than the logevent is 10 times bigger then expected.  

What do you think?

src/NLog/LayoutRenderers/CallSiteLayoutRenderer.cs
```diff
@@ -130,89 +130,79 @@ StackTraceUsage IUsesStackTrace.StackTraceUsage
...
-                            // <Main>b__2
-                            if (methodName.Contains("__") == true && methodName.StartsWith("<") == true && methodName.Contains(">") == true)
-                            {
-                                int startIndex = methodName.IndexOf('<') + 1;
-                                int endIndex = methodName.IndexOf('>');
 
-                                methodName = methodName.Substring(startIndex, endIndex - startIndex);
-                            }
-                        }
+        private void AppendFormatedCalssName(StringBuilder builder, LogEventInfo logEvent)
```
304NotModified: small typo ;)

src/NLog/LayoutRenderers/CallSiteLayoutRenderer.cs
```diff
@@ -130,89 +130,79 @@ StackTraceUsage IUsesStackTrace.StackTraceUsage
...
+            {
+                AppendFormatedMethodName(builder, logEvent);
+            }
+#if !SILVERLIGHT
+            if (this.FileName)
+            {
+                AppendFormatedFileName(builder, logEvent);
+            }
+#endif
+        }
```
304NotModified: nice refactor work :+1:

src/NLog/ILoggerBase.cs
```diff
@@ -69,11 +72,27 @@ LogFactory Factory
 
         #region Log() overloads
 
+#if NET4_5
         /// <summary>
         /// Writes the specified diagnostic message.
         /// </summary>
         /// <param name="logEvent">Log event.</param>
-        void Log(LogEventInfo logEvent);
+        /// <param name="memberName">Method or property name of the caller.</param>
```
304NotModified: small mistake: The summary is no also .net 4.5 only

src/NLog/Logger.cs
```diff
@@ -95,14 +96,35 @@ public bool IsEnabled(LogLevel level)
...
+        /// Writes the specified diagnostic message.
+        /// </summary>
+        /// <param name="logEvent">Log event.</param>
+#endif
+        public void Log(LogEventInfo logEvent
+#if NET4_5
+           , [CallerMemberName] string memberName = ""
+           , [CallerFilePath] string sourceFilePath = ""
+           , [CallerLineNumber] int sourceLineNumber = 0
+#endif
```
ilya-g: If you remove `Log(LogEventInfo)` and replace it with `Log(LogEventInfo, string, string, int)` method, it will cause binary incompatibility.
Consider, for example, a library compiled against NLog for .NET 4.0 that is being redirected to NLog for .NET 4.5 upon load — it will fail to find the method `Log(LogEventInfo)`

sorin-bolos: I will do a test to confirm this, but I think you are right. That's why I wanted to start by modifying a single log method to discover possible issues.

src/NLog/Logger.cs
sorin-bolos: Is there an alternative to capture this info only on some events? 
We should also consider not breaking compatibility

As ilya-g said:
>Consider, for example, an assembly compiled against NLog for .NET 4.0 that is being redirected to NLog for .NET 4.5 upon load — it will fail to find the method Log(LogEventInfo)

src/NLog/Logger.cs
304NotModified: I'm not sure what the compiler does with the parameters (Because of the caller attibutes) . The proposed test would really be helpfull :+1:

304NotModified:
Also
https://visualstudio.uservoice.com/forums/121579-visual-studio/suggestions/2762025-caller-membername-filepath-linenumber-of-net-4-5

304NotModified:
Closing this because of the lack of activity and because we are closing stuff for the upcoming release. Feel free to reopen when it's active again.

PR #638, opened 2015-03-10
Added "CustomActionTarget"

sorin-bolos:
Added a new target that uses an implementation of an interface to run a coded action. 
By implementing IActionProvider interface and registering the instance in the CustomActionTarget you can perform any action when a LogEvent is logged.

# Example

## Logging to console using a custom action

The configuration file looks like this.

```xml 
<?xml version="1.0" ?>
<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <targets>
    <target name="myTarget" xsi:type="CustomAction" layout="${date:format=HH\:MM\:ss} ${logger} ${message}" />
  </targets>

  <rules>
    <logger name="*" minlevel="Info" writeTo="myTarget" />
  </rules>
</nlog>
 ```

To write the code for the action we want to perform we must build a class that implements the interface **IActionProvider** (located in **NLog.Targets** namespace).

```c#
using System;
using NLog;
using NLog.Targets;

public class ConsoleActionProvider : IActionProvider
{
    public void Action(TargetWithLayout target, LogEventInfo logEvent)
    {
        Console.WriteLine(" Action Povider: {0}; Target {1}; Event: {2}",
             this.GetType(), target.Name, target.Layout.Render(logEvent));
    }
}
```

After that we must create an instance of this class and register it on the CustomActionTarget static action set.

```c#
class Example
{
    {
        ConsoleActionProvider provider = new ConsoleActionProvider();
        CustomActionTarget.Register(provider);

        Logger logger = LogManager.GetLogger("Example");
        logger.Info("log message");
        logger.Error("error message");
        Console.ReadKey();
    }
}
```
The result is: 
```
 Action Povider: ConsoleActionProvider; Target myTarget; Event: 18:03:10 Example log messagege
 Action Povider: ConsoleActionProvider; Target myTarget; Event: 18:03:08 Example error message
```

It is possible to unregister an action provider by using `CustomActionTarget.Unregister(provider)`

P.S. I tried to add the new files to all the projects, but in doing so I may have added some unwanted changes to the solution files ( src/NLog.monodevelop.sln; All_Targets.sln and NLog.wp7.csproj) (because I only have visual studio 2012 and 2013). If I did please ignore those changes.

I also added unit tests and examples


304NotModified:
Looks cool! I will check the code in detail this week 

About the project files: Did you check https://github.com/NLog/NLog/blob/master/CONTRIBUTING.md#sync-projects?

sorin-bolos:
Yes, I ran the msbuild script. The result was no errors, no warnings.

304NotModified:
Well your unit tests are failing ;)

See AppVeyor details

sorin-bolos:
From AppVeyor I see that the tests fail when checking the InternalLogger after an error.

But when I run these tests locally on my PC they don't fail. I also ran AppVeyor on the forked project on my account and these 3 tests were successful.

Do you have any idea on how to solve this?

304NotModified:
Did you try the failing unit tests on all 3 .net solutions? (3.5, 4.0, 4.5)

sorin-bolos:
The test run successfully on all 3 solutions.

I have another 3 tests that fail, but I don't think they have any connection with the new functionality. 
The test that fail are 
 - ImpersonatingWrapperTest, RevertToSelfTest - becuse I do not have the user NLogTestUser
 - Extension_should_be_auto_loaded_when_following_NLog_dll_format - because the test are run by xunit in a temporary folder where I do not have the extension dlls.

ilya-g:
I don't like the idea that custom actions providers are registered statically. What if I need to have two `CustomActionTarget`s configured and have first to invoke one provider, and second to invoke another one?

304NotModified:
I agree. Multiple custom actions providers should be possible

sorin-bolos:
This is very useful input. I agree with you, but I wanted the action providers to be developed as part of the consumer code. Given that the targets can be configured at run time I don't think there is any way to make the action provider register to a specific target. 

Another approach wold be to let the target specify the provider action to run but this has several problems. You wold have to use reflection, you wold not be able to specify a specific instance of the provider and you wold not be able to register/unregister in the consumer code.

What do you think?

304NotModified:
I propose:


- extend the `<target` with an extra attribute, "action"
- extend `CustomActionTarget.Register` with an extra parameter, "action"

Then we can register multiple actions. This can be done with a (case intensive) dictionary's and no usage of reflection is needed. ;)

sorin-bolos:
Ok. I'll try and do that.

304NotModified:
:+1:

304NotModified:
Do you need help on this @sorin-bolos ?

304NotModified:
Closing this because of the lack of activity and because we are closing stuff for the upcoming release. Feel free to reopen when it's active again.

PR #650, opened 2015-03-26
fix default value of Commandtype

304NotModified:
fixes #571 

The default value was not set. The default of enums are always 0. (http://stackoverflow.com/questions/4967656/what-is-the-default-value-for-enum-variable)

This seems the same bug as #555 ! The `[DefaultValue]` attribute was set, but it wasn't set from the ctor.

PR #654, opened 2015-03-27
Added unit test to validate the [DefaultValue] attribute values + update DefaultAttributes

### LINKED ISSUE #652: Need unit test: test [DefaultValue] is really the default
There were, (and maybe are), multiple issues with the `[DefaultValue]` attribute. Some think this will set the default value, but it doesn't.

This will lead to bugs like #555 and #650.

We need a unit test for this. It probably use "reflection" and should be located in `SourceCodeTests`

304NotModified:
Added unit test to validate the `[DefaultValue]` attribute values because of the recurring bugs.

Closes #652

Note: the unit tests are failing for 3 properties. We have to fix those 3. (also I would like to see the error message in AppVeyor.

message:
----

Additional information: 3 errors: 

 -------- 

- NLog.Targets.DatabaseTarget.KeepConnection has a wrong value for [DefaultValueAttribute] compared to the default ctor. DefaultValueAttribute says = 'True' and ctor tells = 'False'

- NLog.Targets.DatabaseTarget.CommandType has a wrong value for [DefaultValueAttribute] compared to the default ctor. DefaultValueAttribute says = 'Text' and ctor tells = '0'

- NLog.Targets.MailTarget.Body has a wrong value for [DefaultValueAttribute] compared to the default ctor. DefaultValueAttribute says = '${message}' and ctor tells = ''${message}${newline}''

### DIFF
--- a/src/NLog/Targets/DatabaseTarget.cs
+++ b/src/NLog/Targets/DatabaseTarget.cs
@@ -162,7 +162,7 @@ namespace NLog.Targets
         /// database connection open between the log events.
         /// </summary>
         /// <docgen category='Connection Options' order='10' />
-        [DefaultValue(true)]
+        [DefaultValue(false)]
         public bool KeepConnection { get; set; }
 
         /// <summary>
--- a/src/NLog/Targets/MailTarget.cs
+++ b/src/NLog/Targets/MailTarget.cs
@@ -148,7 +148,7 @@ namespace NLog.Targets
         /// </summary>
         /// <remarks>Alias for the <c>Layout</c> property.</remarks>
         /// <docgen category='Message Options' order='6' />
-        [DefaultValue("${message}")]
+        [DefaultValue("${message}${newline}")]
         public Layout Body
         {
             get { return this.Layout; }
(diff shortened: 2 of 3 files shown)


304NotModified:
- `NLog.Targets.DatabaseTarget.KeepConnection `: wiki says `true` but the default ctor says `false`. I think the ` [DefaultValueAttribute] ` and wiki are wrong?
- `NLog.Targets.DatabaseTarget.CommandType` is fixed in #650

Xharze:
I would say that we should change it to match the wiki. This is what everyone thinks it should be anyway

304NotModified:
Yes, but It's unclear for the mailtarget. Dunno what currently the default is.

304NotModified:
I think we should match the documentation on the code. Otherwise we get unneeded backwards comp problems. I think this is no problem because:

- If the user sets the property, there is no problem.
- If the user use the defaults, I would have noticed that the documentation is wrong, or he doesn't care ;)

@Xharze What doe you think?

@Xharze How can be a property required and also have a default value?

304NotModified:
SO proposal: change [defaultAttributes] and wiki to:

- `MailTarget.Body` and `MailTarget.Layout` to `${message}${newline}`
- `DatabaseTarget.KeepConnection` to `false`

Xharze:
If it is unclear I think it should be changed to match the code as you suggest

304NotModified:
wikipages also updated!

304NotModified:
This code is nowadays in tools\CheckSourceCode

Issue #669, opened 2015-04-16
EventLogTarget does not seem to expand layout renderers for the 'source' attribute


oniony:
We are using 2.0.0.0 but I have confirmed this problem still exists in 3.2.0.0. When I have a target of type `EventLog` and use layout renderers such as `${processname}` in the `source` attribute, the raised events have a source of `${processname}` (literally) rather than the process's name.

This made me first think that the layout renderers are not expanded for this attribute but for two reasons:

1. The example on the [Eventlog target](https://github.com/nlog/nlog/wiki/EventLog-target) page actually uses one (`${appname}`) in the example.
2. If I specify a variable instead of one of the predefined layout renderers it then includes the literal value of that variable (with no further layout renderer resolution). For example, if I define a variable `${processname2}` that is defined as `${replace:searchFor=e:replaceWith=E:inner=${processname}}` and then specify the `source` in the target as `${processname2}` then the items logged have a source of the literal text `${replace:searchFor=e:replaceWith=E:inner=${processname}}` and not the modified process name 'chEEsE', for example.

304NotModified:
The example is wrong. The source is a string. Not a layout.

304NotModified:
Duplicate of https://github.com/NLog/NLog/issues/508

304NotModified:
I have updated the wiki.

oniony:
Apologies: I did actually search for somebody having reported this previously but obviously failed.

Just to be clear: for fields of type `string`, is it correct that variables are expanded?

304NotModified:
Only `Layouts` are expanded. `Strings` should remain the same.

oniony:
In which case I think my point 2 in the ticket suggests there is another bug: the variable I was using was expanded to the literal value of that variable.

E.g.

    <variable name="test" value="hello"/>
    ...
    <target xsi:type="EventLog"
            ...
            source="${test}"/>

The events raised have a source of `hello` not `${test}`.

304NotModified:
That would be strange as https://github.com/NLog/NLog/blob/master/src/NLog/Targets/EventLogTarget.cs

wont change the `Source`

304NotModified:
it seems indeed that `ExpandVariables` is ran on all strings (see 
[XmlLoggingConfiguration.cs](https://github.com/NLog/NLog/blob/master/src/NLog/Config/XmlLoggingConfiguration.cs)

oniony:
[...]

PR #670, opened 2015-04-16
added unit test: string with variable get expanded

304NotModified:
Skipped as it's unclear if its a bug

### DIFF
--- a/tests/NLog.UnitTests/Config/VariableTests.cs
+++ b/tests/NLog.UnitTests/Config/VariableTests.cs
@@ -67,5 +67,23 @@ namespace NLog.UnitTests.Config
             Assert.Equal("[[", lr1.Text);
             Assert.Equal("]]", lr3.Text);
         }
+
+        [Fact(Skip = "Bug? Dunno, but it's bad")]
+        public void VariablesTest_string_should_not_exand()
+        {
+            var configuration = CreateConfigurationFromString(@"
+<nlog throwExceptions='true'>
+  <variable name='test' value='hello'/>
+  <targets>
+    <target type='EventLog'  name='test'  source='${test}'/>
+  </targets>
+</nlog>");
+
+            var target = configuration.FindTargetByName("test") as EventLogTarget;
+            Assert.NotNull(target);
+            //dont change the ${test} as it isn't a Layout
+            Assert.Equal("${test}", target.Source);
+
+        }
     }
 }


Xharze:
I believe this is a bug, shouldn't we try to fix in this pull request too?

304NotModified:
If it's a bug, yes. Although I could be handy to expand also strings? For configuring passwords globally etc.

Xharze:
It could have use cases but I think it is a bit counter intuitive

304NotModified:
Agree. But I'm always reticent (not sure if this is the correct English word) with changing things non-backwards compatible.

Xharze:
I'll merged the test for now, when we figure out what to do a new pull request can be added

304NotModified:
:+1:

ArchonMegalon:
It seems to me that, there is a (new) bug originated in the change of the source Eventlog from the original Post: https://github.com/NLog/NLog/issues/669

I'm new to github, so - if this is not the place to say this please tell me so.

What I noticed: The EventLog-Target has as Source a Layout
https://github.com/nlog/nlog/wiki/EventLog-target

But: trying to set it like this failed:
```C#
logEvent.Properties["MySource"] = useSource;
logEvent.Properties["EventID"] = dicId.EventId;
```
```XML
   <target xsi:type="EventLog" name="ELNoFileName" machineName="."
            source="${event-properties:item=MySource}" log="Application"
            eventId="${event-properties:EventID}"
            layout="ELNoFileName (${event-properties:item=MySource}): ${message}"/>
```

It DOES set the EventID, it DOES set the source, but it seems to me it KEEPS the source the same for all logevents to the target after the first loggin-action.

Now - as you can see - it's pretty mystic for me. How to report such a bug or how to give you guys the possibility to re-create this problem?

Maybe it's also about the useSource?`Basically i fill it like this: 
```C#
 var eventLogSource = !string.IsNullOrEmpty(userMethod?.DeclaringType?.Assembly.GetName().Name)
                ? userMethod.DeclaringType.Assembly.GetName().Name
                : "Application Error";
```

304NotModified:
@ArchonMegalon is this the same issue as https://github.com/NLog/NLog/issues/1542?

ArchonMegalon:
Hi! Yes! I was not able to find my own post again, so I opened a new issue. Sorry for that!

PR #671, opened 2015-04-17
Bugfix: Broken xml stops logging

### LINKED ISSUE #273: NLog stops logging when the config file is broken, and then config is repaired (autoreload)
Reproducting the issue
1. Use autoreload="True" parameter in config file.
2. Start application - NLog working fine.
3. Invalidate XML structure in NLog.config file (i.e. delete some closing tag), save the file, then NLog will stop logging.
4. Change NLog.config to the previous valid state, save it - NLog still not logging until the application is restarted.

304NotModified:
fixes #273 

problem: `XmlLoggingConfiguration.Initialize` eats exception with invalid
XML. ALso `XmlLoggingConfiguration.Reload` never returns `null`.
So the `null` check is not enough

- added `InitializeSucceeded` (nullable `bool`) in `XmlLoggingConfiguration`
- checks also `InitializeSucceeded`. If` false`, DON't apply the config.

Result: invalidating the XML won't stop logging, the old config will be used.
Also, the auto reload won't turned to `false` (the default)

The unit tests proves the solution and also the previous problem.

### DIFF
--- a/src/NLog/Config/XmlLoggingConfiguration.cs
+++ b/src/NLog/Config/XmlLoggingConfiguration.cs
@@ -62,6 +62,8 @@ namespace NLog.Config
 
         private string originalFileName;
 
+
+
         /// <summary>
         /// Initializes a new instance of the <see cref="XmlLoggingConfiguration" /> class.
         /// </summary>
@@ -155,6 +157,11 @@ namespace NLog.Config
             }
         }
 #endif
+
+        /// <summary>
+        /// Did the <see cref="Initialize"/> Succeeded? <c>true</c>= success, <c>false</c>= error, <c>null</c> = initialize not started yet.
+        /// </summary>
+        public bool? InitializeSucceeded { get; private set; }
 
         /// <summary>
         /// Gets the variables defined in the configuration.
@@ -259,6 +266,7 @@ namespace NLog.Config
         {
             try
             {
+                InitializeSucceeded = null;
                 reader.MoveToContent();
                 var content = new NLogXmlElement(reader);
                 if (fileName != null)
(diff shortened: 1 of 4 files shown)


304NotModified:
The changes are small, but a review would be wise because there is a new public property added.

304NotModified:
Unit tests are working local. Dunno why AppVeyor is complaining

304NotModified:
Fixed: )

Xharze:
Looks good to me. Just wondering about all the linebreaks between the tests? Any reason for this?

304NotModified:
The line breaks is just bad coding style of me ;) - I like white space when programming.

Most of the time I let Resharper fix this, but too bad the NLog code style if really different then my Resharper profile (I like `var` for example)

I will fix this code styling thing with a commit.

304NotModified:
wtf. Removed newlines and now are the unit tests failing?

304NotModified:
Unit tests are fixed and PR is ready for merge after review.

src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -62,6 +62,8 @@ public class XmlLoggingConfiguration : LoggingConfiguration
 
         private string originalFileName;
 
+
```
Xharze: Could you delete these extra newlines?

tests/NLog.UnitTests/Config/IncludeTests.cs
```diff
@@ -87,6 +92,8 @@ public void IncludeTest()
             }
         }
 
+
```
Xharze: Please delete extra linebreaks

Xharze:
@304NotModified Good work, please see my comments, otherwise it looks great

304NotModified:
Thanks. Unneeded newlines removed!

PR #681, opened 2015-04-23
Fixed an issue where log files couldn't use the same name as archive file...

BrandonLegault:
Fixed an issue where log files couldn't use the same name as archive files when archiving with DateSequence when the files were in the same directory. Didn't affect other archive types.

### DIFF
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -1006,9 +1006,12 @@ namespace NLog.Targets
                 var filesByDate = new List<string>();
                 for (int index = 0; index < files.Count; index++)
                 {
-                    string archiveFileName = Path.GetFileName(files[index]);
+                    //Get the archive file name or empty string if it's null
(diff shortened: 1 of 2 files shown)


304NotModified:
Thanks

Xharze:
Could you add a unit test for this?

BrandonLegault:
I added the unit test for the fix. I should mention that out of the gate (before I made changes) I had 13 failing unit tests. I assume I'm just missing something locally, but just to mention.

304NotModified:
Appveyor isn't complaining.  So the unit tests are OK :)

304NotModified:
Thanks for the unit test!

fritzcoder:
When will this fix be merged and deployed to nuget package? We are using date sequence and hitting this bug. Thanx!

304NotModified:
Probably yes. But there are some more fixes with dates in filename and archiving. We have to double check that.

PR #689, opened 2015-05-01
Make alignment stay consistent when fixed-length truncation occurs.

logiclrd:
The `PaddingLayoutRenderWrapper` class implements left and right padding, based on the sign of the `Padding` parameter. However, something unexpected happens when `FixedLength` is turned on, if the string is too long to fit:
```
"        T" <= "T"
"       Th" <= "Th"
"      Thi" <= "Thi"
"     This" <= "This"
"    This " <= "This "
"   This m" <= "This m"
"  This me" <= "This me"
" This mes" <= "This mes"
"This mess" <= "This mess"
"This mess" <= "This messa"
"This mess" <= "This messag"
"This mess" <= "This message"
"This mess" <= "This message is"
"This mess" <= "This message is too"
"This mess" <= "This message is too long"
```
What's happening, in effect, is that we're getting right-alignment of text shorter than the pad length, but then left-alignment of text that is too long. Intuitively, this sequence should end like this:
```
" This mes" <= "This mes"
"This mess" <= "This mess"
"his messa" <= "This messa"
"is messag" <= "This messag"
"s message" <= "This message"
"essage is" <= "This message is"
"ge is too" <= "This message is too"
" too long" <= "This message is too long"
```
When `Padding` is negative, the current behaviour is what you'd expect:
```
"T        " <= "T"
"Th       " <= "Th"
"Thi      " <= "Thi"
"This     " <= "This"
"This     " <= "This "
"This m   " <= "This m"
"This me  " <= "This me"
"This mes " <= "This mes"
"This mess" <= "This mess"
"This mess" <= "This messa"
"This mess" <= "This messag"
"This mess" <= "This message"
"This mess" <= "This message is"
"This mess" <= "This message is too"
"This mess" <= "This message is too long"
```
This pull request updates `PaddingLayoutRendererWrapper.cs` to remove characters from the start of the string instead of the end when `Padding` is positive. Negative values continue to remove characters from the end of the string. In effect, this means that the padding action, whether it is to add or remove characters, always modifies the same end of the string.

The pull request also adds unit tests that verify this behaviour.

### DIFF
--- a/tests/NLog.UnitTests/LayoutRenderers/MessageTests.cs
+++ b/tests/NLog.UnitTests/LayoutRenderers/MessageTests.cs
@@ -155,7 +155,7 @@ namespace NLog.UnitTests.LayoutRenderers
         {
             LogManager.Configuration = CreateConfigurationFromString(@"
             <nlog>
-                <targets><target name='debug' type='Debug' layout='${message:padding=-3:padcharacter=x:fixedlength=true:alignmentOnTruncation=left}' /></targets>
+                <targets><target name='debug' type='Debug' layout='${message:padding=-3:padcharacter=x:fixedlength=true}' /></targets>
                 <rules>
                     <logger name='*' minlevel='Debug' writeTo='debug' />
                 </rules>
--- a/tests/NLog.UnitTests/LayoutRenderers/MessageTests.cs
+++ b/tests/NLog.UnitTests/LayoutRenderers/MessageTests.cs
@@ -85,11 +85,11 @@ namespace NLog.UnitTests.LayoutRenderers
 
 
         [Fact]
-        public void MessageFixedLengthRightPaddingTest()
+        public void MessageFixedLengthRightPaddingLeftAlignmentTest()
         {
             LogManager.Configuration = CreateConfigurationFromString(@"
             <nlog>
-                <targets><target name='debug' type='Debug' layout='${message:padding=-3:fixedlength=true}' /></targets>
+                <targets><target name='debug' type='Debug' layout='${message:padding=3:fixedlength=true}' /></targets>
                 <rules>
                     <logger name='*' minlevel='Debug' writeTo='debug' />
                 </rules>
@@ -97,13 +97,35 @@ namespace NLog.UnitTests.LayoutRenderers
 
             ILogger logger = LogManager.GetLogger("A");
             logger.Debug("a");
-            AssertDebugLastMessage("debug", "a  ");
+            AssertDebugLastMessage("debug", "  a");
             logger.Debug("a{0}", 1);
-            AssertDebugLastMessage("debug", "a1 ");
+            AssertDebugLastMessage("debug", " a1");
             logger.Debug("a{0}{1}", 1, "2");
             AssertDebugLastMessage("debug", "a12");
             logger.Debug(CultureInfo.InvariantCulture, "a{0}", new DateTime(2005, 1, 1));
             AssertDebugLastMessage("debug", "a01");
+        }
+
+        [Fact]
+        public void MessageFixedLengthRightPaddingRightAlignmentTest()
+        {
+            LogManager.Configuration = CreateConfigurationFromString(@"
+            <nlog>
+                <targets><target name='debug' type='Debug' layout='${message:padding=3:fixedlength=true:alignmentOnTruncation=right}' /></targets>
+                <rules>
+                    <logger name='*' minlevel='Debug' writeTo='debug' />
+                </rules>
+            </nlog>");
+
+            ILogger logger = LogManager.GetLogger("A");
+            logger.Debug("a");
+            AssertDebugLastMessage("debug", "  a");
+            logger.Debug("a{0}", 1);
+            AssertDebugLastMessage("debug", " a1");
+            logger.Debug("a{0}{1}", 1, "2");
+            AssertDebugLastMessage("debug", "a12");
+            logger.Debug(CultureInfo.InvariantCulture, "a{0}", new DateTime(2005, 1, 1));
+            AssertDebugLastMessage("debug", ":00");
         }
 
         [Fact]
@@ -129,11 +151,11 @@ namespace NLog.UnitTests.LayoutRenderers
         }
 
         [Fact]
-        public void MessageFixedLengthLeftPaddingTest()
+        public void MessageFixedLengthLeftPaddingLeftAlignmentTest()
         {
             LogManager.Configuration = CreateConfigurationFromString(@"
             <nlog>
-                <targets><target name='debug' type='Debug' layout='${message:padding=-3:padcharacter=x:fixedlength=true}' /></targets>
+                <targets><target name='debug' type='Debug' layout='${message:padding=-3:padcharacter=x:fixedlength=true:alignmentOnTruncation=left}' /></targets>
                 <rules>
                     <logger name='*' minlevel='Debug' writeTo='debug' />
                 </rules>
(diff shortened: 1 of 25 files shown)


304NotModified:
Thanks for this PR. 

I have to admit that I have difficulties to understand your examples. What is the `<=`? Can you post a config example? 
Thanks!

304NotModified:
And it seems that other padding unit tests are failing now. Could you look at this?

logiclrd:
Ah, my apologies, I didn't see a tests file for padding in `LayoutRenderers/Wrappers` so I made a new one, and it didn't occur to me that tests of the padding wrapper might exist elsewhere. Somewhat silly of me not to have run all the unit tests after my change. I'll look into that right away.

As for my examples, I put the input message on the right because the length of the sample input messages is variable while the expected output, with `FixedLength = true`, is not. I wanted the expected output to all line up. So, when I wrote this:
```
"This mess" <= "This message is"
```
What I meant to indicate was that if you format the message `"This message is"` with a layout that specifies `Padding=9:FixedLength=true`, then you get the resulting output string `"This mess"` (instead of the more logical `"essage is"`).

logiclrd:
There we go. :-)

logiclrd:
Well, rather than fixing the bug, it would be possible to add a new property to `PaddingLayoutRendererWrapper.cs`. The purpose of the property would essentially be to say "stop being buggy" -- this could be written as `public bool UseConsistentAlignment`. I don't think it would make sense to add a property that allows left- or right-alignment to be explicitly specified, because that then competes with the already-documented method of specifying alignment through the sign of the `Padding` value -- negative for left-aligned, positive for right-aligned.

The only situation where this change as-written would break things for existing users would be where they are putting class names into a right-aligned field that is fixed length and not adequately wide enough for the name of the class.

If *all* of the class names are wider than the field name, then they can get back their original behaviour by making the `Padding` value negative. The only situation where they can't just alter settings to get back the original behaviour would be where some of the class names are shorter and some are longer. These people currently get this kind of behaviour:
```
Layout string: "[${className:padding=15}]"

Class Name            Output
----------------------------------------
Short                 [          Short]
StillShort            [     StillShort]
GettingLonger         [  GettingLonger]
TooLongForTheField    [TooLongForTheFi]
```

With these changes, they can either preserve `[TooLongForTheFi]` but left-align, by using a negative `Padding`, or they can preserve right-alignment but change `[TooLongForTheFi]` to `[LongForTheField]`.

Would that be significantly problematic for existing users, do you think?

304NotModified:
@Xharze I need your opinion on this. I don't know what is expected on this, but changing the current unit tests for this seems wrong.

304NotModified:
@logiclrd thanks for the detailed post

logiclrd:
Just to be clear, my opinion on it is that the current behaviour is a bug, and the existing unit tests are simply tailored to the bug's behaviour. If you fix a bug and unit tests start failing, then the unit tests are wrong, right? :-)

304NotModified:
I think we need `alignment` (left or right)  property

logiclrd:
There's a problem with that, though. What to do if the user specifies `padding=10:alignment=left`? Positive padding is explicitly documented to mean right alignment (padding characters are added on the left). Similarly, what if the user specifies `padding=-10:alignment=right`?

Xharze:
I think the alignment property is the right way to go, this will ensure this isn't a breaking change, as long as the default value is consistent with the current behaviour. @304NotModified

304NotModified:
it's maybe 'alignment on truncation'. I think also that this is sometimes desired:

```
"        T" <= "T"
"       Th" <= "Th"
"      Thi" <= "Thi"
"     This" <= "This"
"    This " <= "This "
"   This m" <= "This m"
"  This me" <= "This me"
" This mes" <= "This mes"
"This mess" <= "This mess"
"This mess" <= "This messa"
"This mess" <= "This messag"
"This mess" <= "This message"
"This mess" <= "This message is"
"This mess" <= "This message is too"
"This mess" <= "This message is too long"
```

logiclrd:
Okeydoke, I'll rework the PR. :-)

304NotModified:
:+1:

logiclrd:
Where is the optimal balance between brevity self-descriptivity? Would it be too long to name it `alignmentOnTruncation`, or too short to name it `trAlign`? My personal feeling is that naming it simply `align` or `alignment` may lead to confusion. When people encounter it without the context of the documentation, e.g. in Stack Overflow answers, they'll either make incorrect assumptions about its behaviour or they'll say, "It's not doing what I'd expect it to do, I'd better steer clear of it."

304NotModified:
alignmentOnTruncation sounds fine to me.

304NotModified:
Good question by the way. IMO as short as possible but understanding the purpose with only the class name as context is a must.

logiclrd:
Thanks. :-)

I guess it would be horribad to reuse `System.Windows.Forms.HorizontalAlignment` for this new property. ;-)

I'll add a new enum.

304NotModified:
Yes indeed! A new enum is much better

logiclrd:
I know you don't like the idea of changing tests to make things work, but in this case I bumped into the `ThreadAgnosticAttributeTest`, which is based on the assumption that all concrete types in the `NLog.LayoutRenderers.Wrappers` namespace will have the `[ThreadAgnostic]` attribute. While `PaddingLayoutRendererWrapper.cs` already has this attribute, the new `enum` of course does not.

I noticed that `ThreadAgnosticAttributeTest` already checks for & skips abstract types, so I updated it to also skip enumerations.

If the tests pass on this commit, then I think the rework is ready for review. :-)

304NotModified:
OK thanks for the info. Changing unit tests should be avoided, but in this case it's a very good change. :+1:

304NotModified:
Is this still work in progress?

logiclrd:
Nope, I think it's good to go. Let me know if you find anything in review, though. :-)

304NotModified:
I only checked the tests in short. This PR is changing existing unit tests and from the view of backwards compatibility, we don't prefer that.

E.g. why is `ClassNameWithPaddingTestTest` changed? Maybe reverting this and creating a new unit test for your case?

logiclrd:
These unrelated tests were affected by the first revision of this change. Before any change, the part of the tests where the message gets truncated by argument `fixedLength=true` worked whether `padding` was positive or negative. The first revision made the truncation contingent on whether `padding` was positive or negative, so suddenly it mattered whether a positive or negative value was used in the tests. The changes to existing tests consisted of changing the `padding` value from a positive value to a negative value.

The second revision to the change reverts this dependency on the sign of `padding`, so that it no longer matters whether the unit tests use a positive or a negative value. I could revert the tests to the way they were before, but I reviewed them and they're basically still testing what they were before. Let me know if you'd like me to do that reversion, but if you look at what's actually changed I think you'll see that the underlying semantics are unchanged. :-)

304NotModified:
First, thanks for explaining everything in great detail. :+1: 

Second, the unit tests. I understand the progress and how these unit tests changed (in time). But we're trying not break backwards compatibility, and so, it's very important to keep the current unit tests intact. That's also the great advance in using unit tests :)
Also that gives us confidence to include this change in the new release without breaking things.

I would suggest:
- copy the changed unit test and rename them.
- revert the changes in the original unit tests.

Best of both worlds I think. To be clear, the change in `ThreadAgnosticTests` don't have to be reverted (or new added unit tests).

304NotModified:
One note: we will probably create tomorrow a release candidate for 4.0. After that, the code is feature frozen until the (RTM) release of 4.0.

logiclrd:
Okay, that ought to do it. :-)

In each case, I reverted the test, but then split it out into different tests for different signs of `Padding` and different values of `AlignmentOnTruncation`. The base case for each is back to its exact definition before 006ab17.

I'm writing this comment before AppVeyor finishes its work, but on my machine, all of these tests pass.

Does that sound good?

304NotModified:
Yes sounds good. :+1: 

I'm always lazy. If I think it should work (the unit tests), I just push it to GitHub and let AppVeyor do the boring stuff ;)

tests/NLog.UnitTests/LayoutRenderers/MessageTests.cs
```diff
@@ -129,11 +151,11 @@ public void MessageLeftPaddingTest()
         }
 
         [Fact]
-        public void MessageFixedLengthLeftPaddingTest()
+        public void MessageFixedLengthLeftPaddingLeftAlignmentTest()
         {
             LogManager.Configuration = CreateConfigurationFromString(@"
             <nlog>
-                <targets><target name='debug' type='Debug' layout='${message:padding=-3:padcharacter=x:fixedlength=true}' /></targets>
+                <targets><target name='debug' type='Debug' layout='${message:padding=-3:padcharacter=x:fixedlength=true:alignmentOnTruncation=left}' /></targets>
```
304NotModified: Please revert this one also

304NotModified:
You forgot one I think. See line comment

logiclrd:
Ah, oops -- that was actually a test that I hadn't modified in the first pass, but because I was doing the splitting up on the "RightPadding" version, I figured I'd do the same split on the "LeftPadding" version. Without thinking, I added `alignmentOnTruncation=left` to one and `alignmentOnTruncation=right` to the other. I had, fact, added `alignmentOnTruncation` in all instances that were split, and then subsequently realized that if I was going to keep the original tests unchanged, then the original incarnations would need to simply rely on `left` being the default value for `alignmentOnTruncation`, rather than explicitly setting it.

I overlooked removing that `alignmentOnTruncation=left` from the "LeftPadding" version of the `MessageFixedLength` tests before committing, and ended up causing a new change (though technically a non-semantic one) from the pre-#689 tests. :-)

Reverting now.

logiclrd:
I think we're all good now! :-)

304NotModified:
Great. We only need docs on the wiki now :)

logiclrd:
Change applied :-) How does that look?

304NotModified:
FYI: I removed the `goto` because we think  it's not optimal.

PR #691, opened 2015-05-05
Right way to log exceptions

### LINKED ISSUE #431: "Right" way to log exceptions

304NotModified:
Closes #431

Changes:

- All "exception" methods are starting with exception. E.g `Error(Exception exception, [Localizable(false)] string message, params object[] args);`
- All "exception" methods has 'args' as parameters
- All "exception" methods has an overload with an `IFormatProvider` as parameter.
- Also changed the "conditional" methods from #616 . This wasn't released yet, so no need for Obsoletes. 

Backwardscomp changes:
- removed "exceptionCandidate" hack: Log<Exception>(string message, Exception ex) would write to exception property instead of message. This is non-backwards compatible in behaviour! 
- all other "exception methods": Eg. `ErrorException` and 'Error(string message, Exception exception)` marked as Obsolete, also in the interfaces (`ILogger, ILoggerBase). No removals, only additions or adding Obsolete attributes. 

Note: You can ignore changes in logger1.cs, this file is generated with logger.tt

### DIFF
--- a/src/NLog/ILogger.cs
+++ b/src/NLog/ILogger.cs
@@ -135,6 +135,23 @@ namespace NLog
         void TraceException([Localizable(false)] string message, Exception exception);
 
         /// <summary>
+        /// Writes the diagnostic message and exception at the <c>Trace</c> level.
+        /// </summary>
+        /// <param name="message">A <see langword="string" /> to be written.</param>
+        /// <param name="exception">An exception to be logged.</param>
+        /// <param name="args">Arguments to format.</param>
+        void Trace(Exception exception, [Localizable(false)] string message, params object[] args);
+
+        /// <summary>
+        /// Writes the diagnostic message and exception at the <c>Trace</c> level.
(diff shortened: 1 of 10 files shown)


304NotModified:
This is done. 'm doubting about removing the "exceptionCandidate hack", because I had to change some unit tests for that. 

e.g. had to change `Log("message", ex)` to `Log(ex, "message")` (`ex` is of type `Exception`)

Xharze:
Looks good, I'll merge it

304NotModified:
also closes #63

PR #692, opened 2015-05-05
added Encoding property for consoleTarget + ColorConsoleTarget

### LINKED ISSUE #676: Ability to change Encoding in ConsoleTarget and ColoredConsoleTarget
It's rather good for debugging to support more languages which based on encoding. MSDN recommend to use Unicode encoding (not Default), but it's nice when you can choose it for your-self, as in FileTarget.cs (https://msdn.microsoft.com/en-us/library/system.text.encoding.default%28v=vs.110%29.aspx). It's frustrating when some messages become "question mark" array. So please, fix it.

304NotModified:
Closes #676

No unit tests added as it is a small addition.

The property has a side effect (static property on `Console`), but I think we can live with that.

### DIFF
--- a/src/NLog/Targets/ColoredConsoleTarget.cs
+++ b/src/NLog/Targets/ColoredConsoleTarget.cs
@@ -31,6 +31,8 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
+using System.Text;
+
 #if !SILVERLIGHT
 
 namespace NLog.Targets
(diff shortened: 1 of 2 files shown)


Xharze:
Looks good, but I think the using should be moved inside the `!IF SILVERLIGHT` block like the rest of them

304NotModified:
I will update my PR.

304NotModified:
@Xharze I don't understand your comment I think:

- `ColoredConsoleTarget` is already in `#if !SILVERLIGHT`
- `Encoding` is available in Silverlight? (see `FileTarget`) - for `ConsoleTarget`

src/NLog/Targets/ColoredConsoleTarget.cs
```diff
@@ -31,6 +31,8 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
+using System.Text;
```
Xharze: Move inside `#if !SILVERLIGHT`

Xharze:
I've tried to add as a line comment :)

304NotModified:
Thanks, I fixed it :) 

Including also some doc/newlines stuff

PR #698, opened 2015-05-10
Fixed where log files couldn't use the same name as archive file

Xharze:
Manual merge of #681

closes #681


304NotModified:
Just renamed it, because I did not understood it ;)

Issue #381, opened 2014-06-27
Lost messages using a Web service


Iehbr89:
When you send a message to a [WebService](https://github.com/nlog/NLog/wiki/WebService-target) or [LogReceiverService](https://github.com/NLog/NLog/wiki/LogReceiverService-target) lost some messages.
Demo apps:
- [Sender](https://yadi.sk/d/8RJw-GkYV7H9t) (`AfcNLogTest`).
- [WebServiceReceiver](https://yadi.sk/d/Fwt01R27W87Jq) (`AfcNLogWebService`);
- [WCFReceiver](https://yadi.sk/d/xJN99C6JW87E9) (`AfcNLogWCF`);

You can see this by comparing three files:
1 - `c:\logs\AfcNLogTest.log`:

```
2014-07-07 17:27:56.9919|TRACE|AfcNLogTest.Program|Log message 0
2014-07-07 17:27:57.6755|DEBUG|AfcNLogTest.Program|Log message 0
2014-07-07 17:27:57.6823|INFO|AfcNLogTest.Program|Log message 0
2014-07-07 17:27:57.6823|WARN|AfcNLogTest.Program|Log message 0
2014-07-07 17:27:57.6823|ERROR|AfcNLogTest.Program|Log message 0
2014-07-07 17:27:57.6979|FATAL|AfcNLogTest.Program|Log message 0
```

2 - `c:\logs\AfcNLogWebService.log`:

```
2014-07-07 17:27:56.9919|TRACE|Client.AfcNLogTest.Program|Log message 0
```

3 - `c:\logs\AfcNLogWCF.log`:

```
2014-07-07 17:27:57.6940|TRACE|AfcNLogTest.Program|Log message 0
2014-07-07 17:27:57.7663|WARN|AfcNLogTest.Program|Log message 0
2014-07-07 17:27:57.7663|FATAL|AfcNLogTest.Program|Log message 0
2014-07-07 17:27:57.7663|INFO|AfcNLogTest.Program|Log message 0
2014-07-07 17:27:57.6940|DEBUG|AfcNLogTest.Program|Log message 0
```

Ideally, they should be identical!

I will be grateful if you can help solve this problem.

For new results do the following:
1. Clean `c:\logs\AfcNLogWebService.log` and `c:\logs\AfcNLogWCF.log` and `c:\logs\AfcNLogWCF.log`;
2. Open `AfcNLogWebService` solution in Visual Studio and Run it;
[...]

PR #703, opened 2015-05-13
Added 'lost messages' Webservice unittest

304NotModified:
- 1000 messages.
-.net 4.5 only
- documented and renamed some unclear names.

Tested: 10, 100, 1000 message.
1000 message took ~1 SEC

conclusion: cannot reproduce #381

### DIFF
--- a/tests/NLog.UnitTests/Targets/WebServiceTargetTests.cs
+++ b/tests/NLog.UnitTests/Targets/WebServiceTargetTests.cs
@@ -32,6 +32,7 @@
 // 
 
 using System;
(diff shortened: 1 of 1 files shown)


304NotModified:
Note: locally tested with the versions: 3.2.1, 3.0.0 and 2.0.1 - the same results - in this case succesfull

(for example the code of [WebserviceTarget of 3.2.1](https://raw.githubusercontent.com/NLog/NLog/v3.2.1/src/NLog/Targets/WebServiceTarget.cs))

PR #705, opened 2015-05-15
EventLogTarget.Source layoutable & code improvements to EventLogTarget

### LINKED ISSUE #508: EventLogTarget.Source should be of type `Layout`, not `string` - and therefore will be layout-able
I have a windows service which is installed from the same assembly path as multiple instances (differentiated only by service account and service name).  I use the FileTarget log and archive paths with `${windows-identity}` successfully but `EventLogTarget.Source` member does not get expanded in the same way and so the source's Layout string shows up literally, assuming the user account has rights to create new Event Sources in Windows.

(Haters note: I'm not particularly happy with the choice of installing the same service multiple times, but that wasn't up to me.  But as I must stick to that, installing the service multiple times from the same binary is certainly preferable to having to create different builds for each one!)

304NotModified:
- making event source layoutable as requested. Closes #508.
- removed code duplication
- improved docs.
- improved logging to internal logger.

Note: on install/uninstall, the event source is only created / removed when the Source doesn't contain layout renderers. This is needed because we need a `LogEventInfo` when rendering the Layout, and there isn't a `LogEventInfo` when installing/uninstalling.

### DIFF
--- a/src/NLog/Layouts/SimpleLayout.cs
+++ b/src/NLog/Layouts/SimpleLayout.cs
@@ -119,6 +119,21 @@ namespace NLog.Layouts
                 this.SetRenderers(renderers, txt);
             }
         }
+        /// <summary>
+        /// Is the message fixed? (no Layout renderers used)
+        /// </summary>
+        public bool IsFixedText
+        {
+            get { return this.fixedText != null; }
+        }
+
+        /// <summary>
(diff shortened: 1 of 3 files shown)


304NotModified:
Implemented this, because it's a breaking change (in behaviour) and the 4.0 release is the most suitable for that.

PR #706, opened 2015-05-16
Introduce auto flush behaviour again

### LINKED ISSUE #656: Intentional AutoFlush change in FileTarget?
We recently migrated from v2.1 to v3.2 and did some focused stress testing to understand lower level behavioral differences. We found a slight regression in performance using `AutoFlush=true` within `FileTarget`. To understand the change a little better, we profiled a small console application using dotTrace performance profiler and found that [this commit](https://github.com/NLog/NLog/pull/415) was a performance hot spot.

It looks like the latest version of `FileTarget.cs` in master no longer contains the logic from PR #415; [this commit](https://github.com/NLog/NLog/commit/00d53bc3b234917517e2fdc1735f8b4af2d5d8aa#diff-ed0a3fc25ad7408a5b63501fa77dbe69L1523) removed it. I wasn't able to identify the source of the change by reading through the issue list and I also saw that there was a merge conflict on `FileTarget.cs` when merging in v4. The only thing I can find in the repo history is a comment from @Xharze on issue #414, which may suggest that `FileTarget` was undergoing a much larger transformation:
>This is an ongoing improvement I'm making, the File target should be much simpler than currently

I figured it was worth mentioning to understand if this change was intentional, possibly preventing a functional regression (original change was without supporting tests).

kichristensen:
It was be mistake removed in a merge commit 00d53bc.
closes #656

### DIFF
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -1542,6 +1542,11 @@ namespace NLog.Targets
                 this.WriteHeader(appender);
             }
 
+            if (this.AutoFlush)
+            {
+                appender.Flush();
+            }
+
             appender.Write(bytes);
         }
 


Xharze:
New pull request will be create with correct author :)

PR #711, opened 2015-05-19
new SHFB config [wip]

304NotModified:
- Created new SHFB config for NLog 4
- Removed old en unused:
 - tools
 - configs
 - scripts

SHFB.exe is no longer in this repos. I could not get it working without installing it locally, hence the .exe in the repos is unused.


304NotModified:
The SHFB is working, but not with the mbuild target. Maybe we should fix that before merge of this one?

Xharze:
What is the problem when running with MSBuild?

304NotModified:
This config is separated developed, without any msbuild  / script integration. 

I cannot integrate in the MSbuild targets, as it's fully undocumented.

Xharze:
Could you create a script file or similar generating the documentation? Then this can be called from MSBuild

304NotModified:
Yes I will. I will let you know when it's done

304NotModified:
I will re-open this in the future

Issue #730, opened 2015-06-08
BaseFileAppender.WindowsCreateFile randomly causes OverflowExceptions in 64-bit runtime


nernst:
It happens randomly when the handle returned by Win32FileNativeMethods.CreateFile returns a value that is not convertible to an Int32.

I started seeing this after upgrading to 3.2.1 from 2.0.1.2.

NetFX runtime version is 4.5.1.
OS is 64-bit Windows 8.1.

Call stack:
>	mscorlib.dll!System.IntPtr.ToInt32()	Unknown
 	NLog.dll!NLog.Internal.FileAppenders.BaseFileAppender.WindowsCreateFile(string fileName, bool allowConcurrentWrite)	Unknown
 	NLog.dll!NLog.Internal.FileAppenders.BaseFileAppender.TryCreateFileStream(bool allowConcurrentWrite)	Unknown
 	NLog.dll!NLog.Internal.FileAppenders.BaseFileAppender.CreateFileStream(bool allowConcurrentWrite)	Unknown
 	NLog.dll!NLog.Internal.FileAppenders.RetryingMultiProcessFileAppender.Write(byte[] bytes)	Unknown
 	NLog.dll!NLog.Targets.FileTarget.WriteToFile(string fileName, byte[] bytes, bool justData)	Unknown
 	NLog.dll!NLog.Targets.FileTarget.FlushCurrentFileWrites(string currentFileName, NLog.LogEventInfo firstLogEvent, System.IO.MemoryStream ms, System.Collections.Generic.List<NLog.Common.AsyncContinuation> pendingContinuations)	Unknown
 	NLog.dll!NLog.Targets.FileTarget.Write(NLog.Common.AsyncLogEventInfo[] logEvents)	Unknown
 	NLog.dll!NLog.Targets.Target.WriteAsyncLogEvents(NLog.Common.AsyncLogEventInfo[] logEvents)	Unknown
 	NLog.dll!NLog.Targets.Wrappers.AsyncTargetWrapper.ProcessPendingEvents(object state)	Unknown
 	mscorlib.dll!System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state, bool preserveSyncCtx)	Unknown
 	mscorlib.dll!System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state, bool preserveSyncCtx)	Unknown
[...]

PR #732, opened 2015-06-08
Issue 730: Correct detection of an invalid file handle returned from …

nernst:
…Win32 CreateFile API to be compatible with 32 & 64 bit systems.

SafeFileHandle was already being used to manage the lifetime of the created file. It also has a IsInvalid property that handles detection of an invalid handle via an IntPtr for both 32 & 64-bit systems.


304NotModified:
Thanks for the PR. 

We are doubting to include this in 4.0 (release, today). I dunno the possible impact on this issue.

304NotModified:
Related #730

304NotModified:
Again thanks for this PR, but it seems that #744 is a safer version (regarding leaks) of this fix.

nernst:
No problem at all. Just happy it's being resolved. Thanks, everyone!

PR #734, opened 2015-06-09
Fix autoflush

### LINKED ISSUE #656: Intentional AutoFlush change in FileTarget?

304NotModified:
Fixes #656 

- Flush was on wrong place
- also flush after replaceFileContent

### DIFF
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -1544,12 +1544,12 @@ namespace NLog.Targets
                 this.WriteHeader(appender);
             }
 
+            appender.Write(bytes);
+
             if (this.AutoFlush)
             {
                 appender.Flush();
             }
-
-            appender.Write(bytes);
         }
 
         private bool InitializeFile(string fileName, bool justData)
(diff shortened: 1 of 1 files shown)


304NotModified:
This was the original code (which was broken by regression bug)

https://github.com/NLog/NLog/pull/415/files

304NotModified:
Merge https://github.com/NLog/NLog.github.io/pull/25 if we merge this

304NotModified:
replaced with #753, because the flush in replaceFileContent wasn't needed

PR #744, opened 2015-06-11
Alternate fix for #730

williamb1024:
Modify Win32FileNativeMethods.CreateFile to return SafeFileHandle rather than IntPtr and adjust callsite to ensure the returned handle is not leaked.

Issue #730

### DIFF
--- a/src/NLog/Internal/FileAppenders/BaseFileAppender.cs
+++ b/src/NLog/Internal/FileAppenders/BaseFileAppender.cs
@@ -213,24 +213,39 @@ namespace NLog.Internal.FileAppenders
                 fileShare |= Win32FileNativeMethods.FILE_SHARE_DELETE;
             }
 
-            IntPtr handle = Win32FileNativeMethods.CreateFile(
+            Microsoft.Win32.SafeHandles.SafeFileHandle handle = null;
+            FileStream fileStream = null;
+
+            try
+            {
+                handle = Win32FileNativeMethods.CreateFile(
                 fileName,
                 Win32FileNativeMethods.FileAccess.GenericWrite,
                 fileShare,
                 IntPtr.Zero,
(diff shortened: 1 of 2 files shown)


304NotModified:
Thanks! 

Can you check the AppVeyor error? 

> NLog.UnitTests.Targets.Wrappers.AsyncTargetWrapperTests.AsyncTargetWrapperExceptionTest [FAIL]

williamb1024:
The failing test isn't related to the changes. It looks to be a race condition within the test with the processing timer. The test does not fail when with the changes locally (so far anyway). I'm still playing with the test a bit to see if I can figure out exactly where the race is.

304NotModified:
Related #747

304NotModified:
This really looks nice :+1: 

But I have to admit I cannot oversee the impact of this change. What's your opinion? Is it safe to include in 4.0.1?

williamb1024:
It's really just a "safer" version of the previous code, there should be no functionality change other than eliminating the overflow exception and ensuring that the native file handle is always closed in the case of an error. In the previous code it was possible that the handle would be leaked until the process terminated. 

There were mutliple possible leaks previously:

1. If an exception occurred after the native CreateFile completed and before SafeFileHandle took ownership of the handle. This could be caused by aborting the thread or an out of memory error (amongst other things). This would leak the native handle until the process terminated.
2. After SafeFileHandle took ownership but before FileStream took ownership of the SafeFileHandle. In this case, the leak would last until SafeFileHandle was finalized (which may or may not occur).

So, to summarize, the only visible difference will be that a native handle is not leaked in error cases.

304NotModified:
Thanks for the explanation. It's clear you have great knowledge regarding these handler stuff. :+1: :smiley: 

I'm more a MVC/Web developer ;)

PR #747, opened 2015-06-11
Correct race condition in AsyncTargetWrapperExceptionTest

williamb1024:
Modify AsyncTargetWrapperExceptionTest to avoid race condition with AsyncWrapper thread timer.


304NotModified:
Thanks! 

Can you explain why the `sleep` isn't needed?

williamb1024:
Sure. The `Thread.Sleep()` call doesn't guarantee that the `AsyncTargetWrapper` actually processes its queue. The timer that is used may fire, but never actually be scheduled to run. If that happens the test will fail and seems to eventually lead to an exception in xUnit. Calling `targetWrapper.Close` will force the queue to be processed and stop the timer, which will ensure that the internal logger is written to.

304NotModified:
Thanks for the explanation :+1: :+1:

Issue #26, opened 2012-10-06
UTF8 should write BOM (optional) (FileTarget)


ghost:
> When you add encoding="utf-8" on a file target, encoding of the result file is a UTF-8 encoding but the preamble (Binary Object Mark) is not present at the start of the file then some tools that use this BOM not recognize the encoding as UTF-8...
> 
> In https://github.com/jkowalski/NLog/blob/master/src/NLog/Targets/FileTarget.cs source file
> 
> I suggest to add preamble bytes as below:
> 
> Method "WriteToFile", line 964

``` c#
 private void WriteToFile(string fileName, byte[] bytes, bool justData)
 {
     if (this.ReplaceFileContentsOnEachWrite)
     {
         using (FileStream fs = File.Create(fileName))
         {
             //get preamble bytes
             byte[] preambleBytes = this.Encoding.GetPreamble();
             byte[] headerBytes = this.GetHeaderBytes();
             byte[] footerBytes = this.GetFooterBytes();
             //write preamble bytes
             if (preambleBytes != null)
             {
                 fs.Write(preambleBytes, 0, preambleBytes.Length);
             }
             if (headerBytes != null)
             {
                 fs.Write(headerBytes, 0, headerBytes.Length);
             }
             fs.Write(bytes, 0, bytes.Length);
             if (footerBytes != null)
             {
                 fs.Write(footerBytes, 0, footerBytes.Length);
             }
         }
         return;
     }

     /* some source code ... */
     if (writeHeader && !justData)
     {
         //get preamble bytes
         byte[] preambleBytes = this.Encoding.GetPreamble();
         //write preamble bytes
         if (preambleBytes != null)
         {
             appenderToWrite.Write(preambleBytes);
         }
         byte[] headerBytes = this.GetHeaderBytes();
         if (headerBytes != null)
         {
             appenderToWrite.Write(headerBytes);
         }
     }
     /* end of method */
 }

```

Moved from http://nlog.codeplex.com/workitem/6831

304NotModified:
BOM is no bug, as it is valid to the specs of UTF-8

Xharze:
Have been fixed #629

304NotModified:
No it isn't. The PR is changed for only Webservice, because it seemed the same problem, but it wasn't.

Xharze:
Sorry, you're of course correct

304NotModified:
NP! I understand the confusion :)

npelletm:
[...]

PR #749, opened 2015-06-11
Add support for byte order mark suffix to unicode transformation encodings.

williamb1024:
Add support for byte order marks to `FileTarget` (Issue #26).

* Adds "-bom" and "-no-bom" suffix for UTF8, UTF16, and UTF32 encodings.
* Adds support for writing the Encoding's preamble to the `FileTarget` class.
* Removes "IncludeBOM" property from `WebServiceTarget` as it is no longer needed.
* Adds `DefaultByteOrderMarkAttribute` which when applied to a property will determine whether a BOM is included when specifying a UTF encoding that does not specifically indicate the inclusion or exclusion of the BOM.

When specifying UTF-8 encoding, the default state will be to exclude the BOM. UTF-16 and UTF-32 encodings, by default include the BOM. The DefaultByteOrderMarkAttribute can be used to override this on a per-property basis. The `FileTarget.Encoding` property is an example of this. The FileTarget's Encoding property is attributed to ensure the BOM is excluded by default, this ensure that existing configurations with UTF-16 or UTF-32 encodings will not begin including a BOM automagically.


304NotModified:
Thanks for the PR. But this is a breaking change  and we need to discuss this first.

Some things:

- we have tested before with "UTF-8-bom" encoding name, but this was leading to backwardscomp problems: what is UTF-8? BOM or no BOM? In FileTarget and WebserviceTarget, it had different behavior 
- BOM is required for UTF16 and UTF32 AFAIK. (because it uses multiple bytes, and order does matter)

williamb1024:
I understand completely. I just wanted to get the code up here so you guys could decide if you wanted it or not. I have attempted to address the concerns that you raised in your comment. 

* The code understands a number of different names for the various UTF encodings, rather than list them here, I will just refer you to the code which is in EncodingHelper.cs. As far as I am aware, all of the names supported by the framework are covered. 
* Each name has three possible states: "name", "name-bom", and "name-no-bom". For example, UTF-32BE may be referred to as "utf-32be", "utf-32be-bom" and "utf-32be-no-bom". If the name is suffixed with "-bom" or "-no-bom" then the resulting encoding will always follow that suffix. For example, "utf-16-bom" will always return an Encoding instance that will return an array of bytes from `Encoding.GetPreamble()`. This leaves the encoding name that does not contain a suffix, for example: "utf-16le", "utf-32" and "utf-8". The default for a non-suffixed encoding name depends on the encoding, UTF-8 defaults to no BOM (which follows the Unicode spec that states it is not recommended, but differs from Microsoft's default stance which includes it). UTF-16 and UTF-32 (regardless of BE/LE) default to including a BOM, which passes the Unicode spec which states the BOM "may" be included.
* As you point out, `FileTarget` and `WebServiceTarget` have different behavior as to whether or not a BOM is included. `FileTarget` did not include a BOM, regardless of the selected encoding. Prior to this point, `FileTarget` did not contain any code to write the BOM. Where as `WebSeriviceTarget` would include the BOM, as it ultimately used a `StreamWriter` which respected the contents of `Encoding.GetPreamble()`. This is why I included the `DefaultByteOrderMarkAttribute`. This attribute can be applied to Property members and indicates to the configuration system whether encoding names without a suffix should or should not include a BOM. Without this attribute, the code falls back to the rules defined in the previous bullet point: UTF-8 has no BOM, UTF-16 and UTF-32 have a BOM (unless the encoding name specifies a BOM or no-BOM specifically).

So, without changes, a FileTarget configured with any of the UTF encodings will continue to operate as it did before this commit is applied. 

A `WebServiceTarget` configured for UTF-8 will not include a BOM, but will include a BOM for UTF-16 and UTF-32. This differs from pre-4.0 code where UTF-8, UTF-16 and UTF-32 would all include a BOM when used with a `WebServiceTarget`.

I believe the only breaking change involved is that of `IncludeBOM` being removed from the `WebServiceTarget`. I'm sure with a little bit of thought, I could come up with a way to include it while still allowing for the newly suffixed encoding names.

304NotModified:
With this change, how do we handle the if a user uses `UTF8Encoding` in the API? Is it with or without BOM? 

I was wrong with the required BOM with UTF-16  / UTF-32.
>Where the precise type of the data stream is known (e.g. Unicode big-endian or Unicode little-endian), the BOM should not be used. 

(from: http://unicode.org/faq/utf_bom.html)

williamb1024:
Programmatic configuration will always follow the Encoding instance's preamble configuration. This does have the side effect that existing programmatic configuration will probably include a BOM where it previously did not.

I believe this is proper result for the following reasons:
1. If the caller specifies an encoding configured to include a preamble, then the code should output the preamble in the appropriate locations.
2. The fact that `FileTarget` is not including a preamble when the encoding specifies a preamble is a bug.

Basically, my view is that non-programmatic configuration should default to following the appropriate protocol for what is being configured (ie: text files have a BOM for UTF-16/UTF-32, but not UTF-8). Where as programmatic configuration should do exactly as the caller specifies (even if he or she doesn't realize what is being specified).

With that stated, my code isn't quite following those guidelines. I need to modify the DefaultByteOrderMarkAttribute to allow different defaults for each of the different transform encodings.

williamb1024:
I've updated the code to adjust the defaults, the results are now:

For `FileTarget` non-programmatic configuration will default to a BOM for UTF-16 and UTF-32 encodings and default to no-BOM for UTF-8 encodings, if not specified.

Programmatic configuration will continue to follow the encoding's preamble setting.

304NotModified:
I was thinking about this change:

- The 'IncludeBOM' solution was chosen,inter alia, because it would have the same solution from the config as from the API. WIth this solution it hasn't?
- We just changed to 'IncludeBOM' in 4.0 with a breaking change. The next breaking change will be at 5.0, but I think 5.0 will take at least a year.

Concluding, I can understand you don't find the 'IncludeBOM' an optimal solution, but we're looking for a backwards-compatible change for the moment.

williamb1024:
OK, before writing off this change completely, consider the following:

The specific problem that I'm attempting to solve is the FileTarget not supporting encoding preambles. My solution burrowed it's way into other targets, specifically the WebServiceTarget. 

Outstanding problems are:

* Programmatic configuration operates differently than non-programmatic configuration.
* Removal of the "IncludeBOM" property is a breaking change

Solutions:

* Change non-programmatic configuration to follow `Encoding.UTF8`, `Encoding.Unicode` and `Encoding.UTF32` defaults. This ensures that the programmatic and non-programmatic configuration will result in the same output.
* Restore the `IncludeBOM` property, making it a `Nullable<bool>` type. If specified (ie, non-null) it will forcibly override the current encoding's (after determining if it's UTF8, UTF16 or UTF32) preamble setting. If the `IncludeBOM` property is not configured, the `WebServiceTarget` target would follow the encoding's preamble setting.

Issues that would arise from these changes:

* FileTarget will begin including BOM's for UTF-8 files by, where as currently it does not. This would bring FileTarget into alignment with Microsoft practices, but differ from Unicode suggested practices.

Do you believe this would meet the requirements?

304NotModified:
I'm sorry but I'm a bit lost. What are the differences compared to the 4.0 release?

304NotModified:
Closing this one due to inactivity.

PR #750, opened 2015-06-11
Optional encoding for JsonAttribute

grbinho:
In case we want to log already valid JSON message, we need to be able to disable additional encoding JsonAttribute does.

Here we can decide to add additional object into the context that is already JSON formatted. To be able to write correct JSON it the log, we need to disable additional encoding that JsonAttribute does, so we just set encode="false"

        <attribute name="Timestamp" layout="${longdate}"/>
        <attribute name="LogLevel" layout="${level}"/>
        <attribute name="MachineName" layout="${machinename}"/>
        <attribute name="Logger" layout="${logger}"/>
        <attribute name="Message" layout="${message}"/>
        <attribute name="Exception" layout="${exception:format=ToString}"/>
        <attribute name="Details" layout="${event-context:item=Details}" encode="false" />
    </layout>


304NotModified:
Thanks!

We have some line endings issue, so I will compare this on my PC this weekend. 

There are some unit tests failing:

 - NLog.UnitTests.SourceCodeTests.VerifyFileHeader: the file must start with the header (just copy from another file)
- NLog.UnitTests.SourceCodeTests.VerifyNamespacesAndClassNames: the class name should be the same as the file and the namespace and the folders should also by in-sync.

Could you fix that?

grbinho:
Ok. Will check it out. I think this is related to my VS extension (Trailing Whitespace Visualizer). I think it removes all trailing whitespace on save.

304NotModified:
We fixed the EOL in master. Can you revert the changes in `MessageQueueTargetTests`?

grbinho:
Hi. Hope I did it :)

304NotModified:
it's still there and I our fault ;)

I had the same issues local. I followed: https://help.github.com/articles/dealing-with-line-endings/#refreshing-a-repository-after-changing-line-endings

304NotModified:
So 
> git rm --cached -r .
> git reset --hard
> git add .

grbinho:
Oh yes. I will do it now.

grbinho:
I did it, no changes to commit. I'm syncing master now..

304NotModified:
Ow yes, you need to revert "MessageQueueTargetTests".

304NotModified:
I learned a trick:

No white-space comparison:

https://github.com/NLog/NLog/pull/750/files?w=1

304NotModified:
So it still only a white space issue. We really don't want to change this file in this PR. So please try a revert on that file.

grbinho:
I will revert to come commit before my changes.

grbinho:
I reverted the file and did the cache+reset..

304NotModified:
Nice! Thanks! The lne ending issue has been resolved as you can see

grbinho:
Great! Sorry for the noobnes on my side. Still getting accustomed to git :)

304NotModified:
No problem at all. This was advanced git stuff, I needed also some tutorials.

src/NLog/Layouts/JsonAttribute.cs
```diff
@@ -45,17 +45,19 @@ public class JsonAttribute
...
         /// </summary>
         /// <param name="name">The name of the attribute.</param>
         /// <param name="layout">The layout of the attribute's value.</param>
-        public JsonAttribute(string name, Layout layout)
+        /// <param name="encode">Encode value with json-encode, true by default</param>
+        public JsonAttribute(string name, Layout layout, bool? encode)
         {
             this.Name = name;
             this.Layout = layout;
+            this.Encode = encode ?? true;
```
304NotModified: What's the benefit of `bool? ` here? 

This is also not backwards compatible!

src/NLog/Layouts/JsonLayout.cs
```diff
@@ -76,18 +76,26 @@ protected override string GetFormattedMessage(LogEventInfo logEvent)
...
                     {
                         sb.Append(", ");
                     }
 
                     first = false;
 
-                    sb.AppendFormat("\"{0}\": \"{1}\"", col.Name, text);
+                    if(col.Encode)
+                    {
+                        sb.AppendFormat("\"{0}\": \"{1}\"", col.Name, text);
```
304NotModified: Please explain this (comments) and less code duplication is preferred

tests/NLog.UnitTests/Layouts/JsonLayoutTests.cs
```diff
@@ -70,9 +70,9 @@ public void JsonLayoutRenderingAndEncodingSpecialCharacters()
             {
                 Attributes =
                     {
-                        new JsonAttribute("date", "${longdate}"),
-                        new JsonAttribute("level", "${level}"),
-                        new JsonAttribute("message", "${message}"),
+                        new JsonAttribute("date", "${longdate}", true),
+                        new JsonAttribute("level", "${level}", true),
+                        new JsonAttribute("message", "${message}", true),
```
304NotModified: Please don't change existing unit tests, without good reason. (this is also a hint that it isn't backwards compatible)

src/NLog/Layouts/JsonAttribute.cs
grbinho: Actually, none. What I wanted to do is optional parameter, but that will not binary compatible. I will change it using multiple constructors. What is the preference?

tests/NLog.UnitTests/Layouts/JsonLayoutTests.cs
grbinho: Ok. Will return to previous state.

src/NLog/Layouts/JsonLayout.cs
grbinho: This will enable user to store already formatted JSON in one of the attributes. This code does not encode or additionaly escape attributes that have col.Encode == false. I will refactor for duplication

src/NLog/Layouts/JsonAttribute.cs
304NotModified: Adding a new ctor has the preference.

src/NLog/Layouts/JsonAttribute.cs
```diff
@@ -56,6 +56,18 @@ public JsonAttribute(string name, Layout layout)
         {
             this.Name = name;
             this.Layout = layout;
+            this.Encode = true;
```
304NotModified: This is a bit inconsistent. The first ctor calls the third, but the second is fully separated. I would suggest that the second ctor also calls the third one (with `this(name, layout, true)

src/NLog/Layouts/JsonLayout.cs
```diff
@@ -76,18 +76,29 @@ protected override string GetFormattedMessage(LogEventInfo logEvent)
...
 
                     first = false;
 
-                    sb.AppendFormat("\"{0}\": \"{1}\"", col.Name, text);
+                    var format = "\"{0}\": \"{1}\"";
+
+                    //If encoding was disabled for current attribute,
+                    //Do not escape the value of the attribute.
+                    //This enables user to write already formated JSON.
+                    if(col.Encode == false)
```
304NotModified: I suggest: `if(!col.Encode)`

src/NLog/Layouts/JsonLayout.cs
```diff
@@ -76,18 +76,29 @@ protected override string GetFormattedMessage(LogEventInfo logEvent)
...
                     }
 
                     first = false;
 
-                    sb.AppendFormat("\"{0}\": \"{1}\"", col.Name, text);
+                    var format = "\"{0}\": \"{1}\"";
+
+                    //If encoding was disabled for current attribute,
+                    //Do not escape the value of the attribute.
+                    //This enables user to write already formated JSON.
```
304NotModified: The really important difference is here that the escaped value is always a string? (the quotes).

src/NLog/Layouts/JsonLayout.cs
```diff
@@ -76,18 +76,29 @@ protected override string GetFormattedMessage(LogEventInfo logEvent)
...
+                    if (!first)
                     {
                         sb.Append(", ");
                     }
 
                     first = false;
 
-                    sb.AppendFormat("\"{0}\": \"{1}\"", col.Name, text);
+                    var format = "\"{0}\": \"{1}\"";
+
```
304NotModified: if-else is preferred. The current contruct I would use if I change  (not overwrite) the `format` in the `if`

tests/NLog.UnitTests/Layouts/JsonLayoutTests.cs
```diff
@@ -104,5 +104,47 @@ public void JsonLayoutRenderingAndEncodingLineBreaks()
...
+            ev.TimeStamp = new DateTime(2010, 01, 01, 12, 34, 56);
+            ev.Level = LogLevel.Info;
+            ev.Message = "{ \"hello\" : \"world\" }";
+
+            Assert.Equal("{ \"date\": \"2010-01-01 12:34:56.0000\", \"level\": \"Info\", \"message\": { \"hello\" : \"world\" } }", jsonLayout.Render(ev));
+        }
+
+        [Fact]
+        public void JsonLayoutRenderingAndEncodingMessageAttribute()
+        {
```
304NotModified: good test cases :+1:

304NotModified:
The code looks good, but I think we could improve it a bit more (see comments)

src/NLog/Layouts/JsonAttribute.cs
grbinho: Makes sense :)

304NotModified:
This PR is ready to merge for 4.1

304NotModified:
We release this in 4.1. :) 

Can you add the option (with example)  on the wiki?

grbinho:
Sure. Will do this weekend.

304NotModified:
:+1:

grbinho:
Updated wiki page
https://github.com/NLog/NLog/wiki/JsonLayout

304NotModified:
Looks good.  Thanks!

PR #751, opened 2015-06-12
Fix incorrect loglevel obsolete message

### LINKED ISSUE #745: Questionable obsolete message
So I have `logger.Debug( message, exception);`

And I get 

```
Error: 'NLog.Logger.Debug(string, System.Exception)' is obsolete: 'Use Debug(LogLevel level, Exception exception, string message, params object[] args) method instead.'	
```

But seems i should be using `logger.Debug(exception, message);` and not `Debug(LogLevel level, Exception exception, string message, params object[] args)`.

So am I mistaken or is the obsolete wrong?

SimonCropp:
fixes #745


304NotModified:
Thanks! :+1: :+1:

PR #756, opened 2015-06-14
Move MDLC and TraceActivityId from Contrib

Xharze:
MDLC can only be used in .NET 4.5

### DIFF
--- a/src/NLog/LayoutRenderers/MdlcLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/MdlcLayoutRenderer.cs
@@ -1,0 +1,66 @@
+// 
+// Copyright (c) 2004-2011 Jaroslaw Kowalski <jaak@jkowalski.net>
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without 
+// modification, are permitted provided that the following conditions 
+// are met:
+// 
+// * Redistributions of source code must retain the above copyright notice, 
+//   this list of conditions and the following disclaimer. 
+// 
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution. 
+// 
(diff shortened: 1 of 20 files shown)


304NotModified:
Not sure if we should include this in 4.0.1 or 4.1

But if we choose for 4.1, then it's not possible to use it with NLog 4.0

Xharze:
Merged in 15af3148dfb32b22d00abc8c654a5ff1ea1306b3 on the 4.1 branch

BrutalCode:
@Xharze Why MDLC can only be used in .Net 4.5?

[CallContext.LogicalGetData] (https://msdn.microsoft.com/en-US/library/system.runtime.remoting.messaging.callcontext.logicalgetdata(v=vs.100).aspx) , [CallContext.LogicalSetData] (https://msdn.microsoft.com/en-us/library/system.runtime.remoting.messaging.callcontext.logicalsetdata(v=vs.100).aspx) , [ConcurrentDictionary<TKey, TValue>] (https://msdn.microsoft.com/en-US/library/Dd287191(v=vs.100).aspx) exist in .Net 4.0 also.

And there is no thing used for MDLC that only exist in .Net 4.5. Am I wrong?

304NotModified:
If it compiles for 4.0, then make a PR please.!

BrutalCode:
They are supported in .Net 4.0 according to MSDN. But not in lower version of .Net. 

OK I created a branch for this from my local master branch.

Also, I noticed that MDLC is supports only string vals.
https://github.com/NLog/NLog-Contrib/pull/7#issuecomment-136138373

So I want to add to support for object vals. OK?

304NotModified:
The object vales is already a ticket and we should not fix that now. (release to soon) .

BrutalCode:
But I am about to complete it :(

BrutalCode:
Here is my PR to support .Net 4.0
https://github.com/NLog/NLog/pull/884

PS: I discarded the object val changes.

PR #758, opened 2015-06-15
Correct obsolete text

### LINKED ISSUE #757: Incorrect overload reference in ObsoleteAttribute after upgrading from 3.2 to 4.0
Hi.
I've upgraded NLog nuget-package from v. 3.2 to 4.0. As a result, this method call:

    Logger.Error(ex.Message, ex)

generates a warning:

> NLog.Logger.Error(string, System.Exception)' is obsolete: 'Use Error(LogLevel level, Exception 
> exception, string message, params object[] args) method instead.

While the warning text is absurd itself (I can't imagine `Error(LogLevel.Information, ...)` call), there's no such overload, which accepts `LogLevel level` as a first argument.

Xharze:
The correct overload is Error(Exception exception, string message, params
object[] args)

closes #757

### DIFF
--- a/src/NLog/Logger1.cs
+++ b/src/NLog/Logger1.cs
@@ -1207,7 +1207,7 @@ namespace NLog
         /// </summary>
         /// <param name="message">A <see langword="string" /> to be written.</param>
         /// <param name="exception">An exception to be logged.</param>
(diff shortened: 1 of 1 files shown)


304NotModified:
This is the same as #751?

304NotModified:
yes it is, but changing the generated file (generated by logger.tt), wont help much ;)

PR #760, opened 2015-06-15
Autoload fix for ASP.net + better autoloading logging

304NotModified:
Tested in MVC application. With `nlogAssembly.Location` we get a path some where in

```
C:\Windows\Microsoft.NET\Framework\v4.0.30319\Temporary ASP.NET Files\....

```
as starting path. Because every `.dll` is in a separate folder in `Temporary ASP.NET Files`, the auto-load didn't work.

This has been resolved with this PR.

### DIFF
--- a/src/NLog/Config/ConfigurationItemFactory.cs
+++ b/src/NLog/Config/ConfigurationItemFactory.cs
@@ -239,6 +239,7 @@ namespace NLog.Config
             var assemblyLocation = GetNLogAssemblyLocation(nlogAssembly);
             if (assemblyLocation == null)
             {
+                InternalLogger.Warn("No auto loading because Nlog.dll location is unknown");
                 return factory;
             }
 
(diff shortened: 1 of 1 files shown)


304NotModified:
Closes https://github.com/NLog/NLog.Web/issues/4

304NotModified:
This is done, but maybe we should check if this also works under differrent situations (Windows.forms, unit test, class lib etc)

304NotModified:
I think we should this this with a local nuget package (not on nuget.org yet).

If someone can try test this also, that would be great

304NotModified:
Still not fully working. Assembly path is correct, but target not found

Xharze:
@304NotModified what is left in this, I'm totally confused about tie problems right know as part of it is here and some in NLog/Nlog.Web#4

304NotModified:
I could image it was confusing ;)

One thread was about loading nlog.web, and the other about auto loading. Those are separate things. But I think after all the NLog.web loading issue has been caused by the not-working auto loading, so this fix would solve it.

But I'm not sure of this works on every situation. We need to test this on other machines and under different situations (Windows forms, class library etc)

304NotModified:
Status: done but needs some more testing.

304NotModified:
Tested by @Xharze and me :)

PR #769, opened 2015-06-19
FIxed ArchiveOldFileOnStartup with DateAndSequential archive numbering mode

remye06:
A nullref in the FileTarget.DateAndSequentialArchive method would consistently prevent ArchiveOldFileOnStartup from working when used along with this archive numbering mode.
Thus, the behavior was exactly the same as if ArchiveOldFileOnStartup were disabled.


304NotModified:
Thanks. The breaking unit test seams not to be related to this change. (more a race condition)

304NotModified:
Closing:

From the creator: 
> #774 PR obsoletes PR #769

PR #781, opened 2015-06-28
All events layout renderer: added `IncludeCallerInformation` option. 

### LINKED ISSUE #768: Fluent api Write() forces you to get file/line in event properties in .NET 4.5
I'd like to use the new ${all-event-properties} to print custom properties to my log, but under .NET 4.5 it's kind of useless since calling the Write method in the Fluent API systematically adds unwanted [CallerMemberName], [CallerFilePath] and [CallerLineNumber] to the properties. Calling Write(null, null, 0) everytime isn't exactly handy.

Importantly, this means that the AllEventProperties example on http://nlog-project.org/2015/06/09/nlog-4-has-been-released.html doesn't work as advertised on .NET 4.5 .

Solving this (besides the "solution" of dropping the feature) is tricky, since those attributes have to be on optional parameters, so you can't really overload the Write() function. I see several possible solutions:

1. One option is adding those properties to the log builder through another method (or rather 4 new methods, to add them individually or all at once). So you'd do Log.Message("bla").Callsite().Write() . Since this is the fluent API you're normally gonna call the Callsite() method on the same line as the Write(), so the caller information should be accurate.

2. If the call site information absolutely have to be retrieved by Write(), then Callsite() could simply set a boolean in the builder indicating whether to use or drop the caller information.

3. Possibly in addition to 1. or 2., have a static option you can set somewhere that tells the builder whether it includes caller info by default.

4. Split properties internally into two dictionaries, Default and Custom. Call info would to into Default. User-added properties go in Custom, and ${all-event-properties} only prints out the contents of the Custom dictionary.

5. Add a WriteSimple() function that doesn't add caller info. Bleh. The Horror.

As a side remark, it's a bit weird that the fluent API includes caller info like this but the regular one doesn't. You might want to think about how to harmonize this behavior across the board

304NotModified:
- added IncludeCallerInformation property (default `false` to be consistent with docs and news post)
- fixes #768 

ready for merge

### DIFF
--- a/src/NLog/LayoutRenderers/AllEventPropertiesLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/AllEventPropertiesLayoutRenderer.cs
@@ -31,6 +31,11 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
+using System.Runtime.CompilerServices;
+
 namespace NLog.LayoutRenderers
 {
     using System;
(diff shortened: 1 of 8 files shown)


304NotModified:
- [ ] add IncludeCallerInformation property on wiki (default false)

BrutalCode:
Wiki page updated: https://github.com/NLog/NLog/wiki/All-Event-Properties-Layout-Renderer

304NotModified:
Thanks!

Issue #787, opened 2015-07-02
Exception is not correctly logged when calling without message


MichaelLogutov:
For NLog 4.0.1

Code
```c#
LogManager.GetCurrentClassLogger ().Error (new InvalidOperationException ("Test"), "test");
LogManager.GetCurrentClassLogger ().Error (new InvalidOperationException ("Test"));
```

With NLog.config
```xml
<?xml version="1.0" encoding="utf-8"?>

<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
      autoReload="true" throwExceptions="true">

    <targets async="true">

        <target name="Console"
                xsi:type="Console"
                layout="[${date:format=dd.MM.yyyy HH\:mm\:ss}] ${exception:format=tostring}" />

    </targets>

    <rules>
        <logger name="*" writeTo="Console" />
    </rules>

</nlog>
```

Outputs
```
[02.07.2015 15:45:56] System.InvalidOperationException: Test
[02.07.2015 15:45:56]
```

304NotModified:
That's a serious bug. We will add those to the unit tests (there are many unit tests for error logging already)

304NotModified:
I found the issue. Without the message argument it's calling the generic version: 

See #788 for details.

304NotModified:
PR merged and so this is fixed in 4.1

304NotModified:
Closing as this had been merged

PR #788, opened 2015-07-02
Fix:  exception is not correctly logged when calling without message [WIP]

304NotModified:
Added unit test for bug report:  #787.

 Without the message argument it's calling the generic version: 

```c#
error(new Exception("test")
```

will call:

```c#
        public void Error<T>(T value)
        {
            if (this.IsErrorEnabled)
            {
                //LogLevel level, IFormatProvider formatProvider, T value
                this.WriteToTargets(LogLevel.Error, null, value);
            }
        }
```

This method is interpreting the argument as `message`. (which is correct). I'm not sure how to fix this neat - the `Exception` is (again) a special case.

This is not a bug since NLog 4, but we said that since NLog 4 the exception parameter is always the first one. 

N.B.: creating a special method for `Exception`, like `Error(Exception ex)` won't solve it, as this works for `Exeption` but not for the sub classes of `Exception` - those will call the generic version. 

Dunno how to solve this neat. Only thing that come to mind: re-introduce the is Exception quirk which just was removed in NLog 4 (another method).

Also  what should be the result of `Error(ex)` when the `${message}` layout renderer is used?

### DIFF
--- a/src/NLog/Logger.cs
+++ b/src/NLog/Logger.cs
@@ -527,7 +527,15 @@ namespace NLog
 
         internal void WriteToTargets<T>(LogLevel level, IFormatProvider formatProvider, T value)
         {
-            LoggerImpl.Write(this.loggerType, this.GetTargetsForLevel(level), PrepareLogEventInfo(LogEventInfo.Create(level, this.Name, formatProvider, value)), this.Factory);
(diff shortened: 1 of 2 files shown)


304NotModified:
Dunno how to solve this neat. Only thing that come to mind: re-introduce the `is Exception` quirk which just was removed in NLog 4 (another method)

304NotModified:
Leave the PR like this. it fixed it and improvements can be done in future versions

PR #789, opened 2015-07-02
added config option for breaking change (Exceptions logging) in NLog 4.0 [WIP]

304NotModified:
Created option
 ```xml
 <nlog exceptionLoggingOldStyle='true'>
```
API: `LoggingConfiguration.ExceptionLoggingOldStyle` so that we still can use 

```c#
Error("test", new Exception());
```

The preferred, and NLog 4 way:

```c#
Error( new Exception(), "test");
```

and that the `Exception` gets rendered by `${exception}`.

This for updating (partially) to NLog 4.0.

This is forwardcompatible, as you can use the new exception logger methods with this option on.

flow:

1. enabled this option for old code
1. upgrade to nlog 4 step-by-step
1. when done, disable this option

### DIFF
--- a/src/NLog/Config/LoggingConfiguration.cs
+++ b/src/NLog/Config/LoggingConfiguration.cs
@@ -64,6 +64,10 @@ namespace NLog.Config
             this.LoggingRules = new List<LoggingRule>();
         }
 
+        /// <summary>
+        /// Use the old exception log handling of NLog 3.0? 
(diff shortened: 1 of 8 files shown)


304NotModified:
proposal name: exceptionLoggingV3Style

304NotModified:
maybe making this forward compatible

Issue #694, opened 2015-05-08
How can I get NLog to reevaluate the config with variable values set from code?


rubenwe:
Hi.

I tried setting a variable named `logpath` from code. This path is used in the fileName attribute of some of my targets. I would love to change this path during the initialization of my application.

I can set the value of the variable but I couldn't get NLog to reevaluate the fileNames. What would be the propper way of doing this?

Thanks, Ruben.

304NotModified:
Is everything configured programmatically or with an config file?

Check also #663

304NotModified:
Is your question answered?

winterlude:
I have exactly the same problem. My console application has to write into different log files on the basis of the input argument. I decided to use a variable, in which I set the value of the input parameter.

This is an excerpt of my NLog config file, where I defined the variable "currentDataTransfer":

    <variable name="currentDataTransfer" value="" />
    <variable name="logDir" value="${basedir}/logs/${processname}"/>
    <targets>
        <target xsi:type="File" name="FileLog" fileName="${logDir}/${currentDataTransfer}_${shortdate}.log"

Then, in the C# main program I set the value of that variable:

    static int Main(string[] args)
    {
        LogManager.Configuration.Variables["currentDataTransfer"] = args[0];

but none of the following instructions works:

        LogManager.ReconfigExistingLoggers();
        LogManager.Configuration.Reload();

304NotModified:
Hi,

only `ReconfigExistingLoggers()` should work. `LogManager.Configuration.Reload();` will reload the full file and so also the new variable value is lost.

If not, this is probably a bug. Another way to change a property:

```C#
   var target = configuration.FindTargetByName("FileLog") as FileTarget;
   target.FileName = "new_path_to_filename";
```

winterlude:
It doesn't work. 
It seems that the variables I defined into the NLog config file are replaced with their values *before* the application runs, instead of keeping a reference to them. Thus, changing the value of one of them has no effects on the targets because they do not contain references to variables anymore.

I have defined a target named `FileLog` in which I used my custom variables `logDir` and `currentDataTransfer`:

    fileName="${logDir}/${currentDataTransfer}_${shortdate}.log"

If I change programmatically the value of the variables like this:

    LogManager.Configuration.Variables["currentDataTransfer"] = args[0];

and then call

    LogManager.ReconfigExistingLoggers();

nothing happens.

[...]

PR #790, opened 2015-07-04
added unit test: Re-evaluate config  - variables uses old values [WIP]

304NotModified:
WIP.

Unit test added. Fails and confirms bug in #694
.
The problem: 

- When reading config from XML, we do `ExpandVariables()`, which replaces `"${filename}"` with `"initVal"`. It then makes a `SimpleLayout` with "initVal" (and not aware of "${filename}"`)
-  `LogManager.ReconfigExistingLoggers()` won't do `ExpandVariables()` and have also not the old layout renderer (`${filename}`)

How to solve? Possible solutions:

1. Re-read the XML config. (but already not possible with load config from `string`)
1. keep the layout renderer somewhere. 
1. make variables `ISupportsInitialize` ?  Create `SimpleLayout` aware of layout renderer?

note: we lose information. Also, we lose information when we expand an variable. (same problem, Layout not stored) - also the order of the vars (because they can be depenend of each other), is not stored.

### DIFF
--- a/tests/NLog.UnitTests/Config/VariableTests.cs
+++ b/tests/NLog.UnitTests/Config/VariableTests.cs
@@ -99,8 +99,8 @@ namespace NLog.UnitTests.Config
         {
             var configuration = new LoggingConfiguration();
             LogManager.Configuration = configuration;
-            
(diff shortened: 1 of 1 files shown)


304NotModified:
not really need. Closing

PR #797, opened 2015-07-13
added switch to JsonLayout to suppress the extra spaces

tmusico:
the extra spaces can now be suppressed for the JsonLayout.  I found this modification to be useful when using a network target sending via UDP ... every little byte matters.

also added was a unit test and I tested the nlog file configuration (with the Nlog.test project) to ensure the configuration attribute was working correctly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,27 +36,28 @@ TOOLS=build/bin/Tools
 OUTPUT_DIR=build/bin/$(CONFIG)/Mono\ 2.x
 MONO_LIB_DIR=$(shell pkg-config --variable=libdir mono)/mono/4.0
 XBUILD=xbuild /nologo
+SOLUTION=monodevelop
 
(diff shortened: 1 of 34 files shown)


304NotModified:
Thanks 

It would be nice if we create one (append)  method which add the optional spacing. Do you agree?

tmusico:
That would work too. I was also thinking an extension method, that was the
way I first wrote it. But then when I looked through the rest of the nlog
source code I realized there wasn't any other extension methods that I
could see.

304NotModified:
I would like to merge this. But there is a conflict.  Can you fix this?

304NotModified:
We need also docs on the wiki. Can you add them (with an 4.1 note)

304NotModified:
Thanks! It has been merged!

Issue #86, opened 2012-10-07
Option to strip newlines from layout-renderers


Xharze:
> I find it quite hard to read (as in, visually parse) log-files where the events span more than one line (because, for example, they include a stacktrace or a multi-line message).
  It would be great if there were an option to strip newlines from layout-renderers and instead render them as either nothing or some other character.

Moved from http://nlog.codeplex.com/workitem/6531

Comments:
It is very simple to implement as a wrapper layout renderer, possibly decorated with AmbientPropertyAttribute. See https://github.com/jkowalski/NLog/blob/master/src/NLog/LayoutRenderers/Wrappers/TrimWhiteSpaceLayoutRendererWrapper.cs for an example.

304NotModified:
developers hint: https://github.com/NLog/NLog/blob/master/src/NLog/LayoutRenderers/Wrappers/TrimWhiteSpaceLayoutRendererWrapper.cs

304NotModified:
merged so closing this.

PR #800, opened 2015-07-14
Replace NewLines Layout Renderer Wrapper

flower189:
Added a layout renderer wrapper to replace newline characters with spaces as per issue https://github.com/NLog/NLog/issues/86

### DIFF
--- a/src/NLog/LayoutRenderers/Wrappers/ReplaceNewLinesLayoutRendererWrapper.cs
+++ b/src/NLog/LayoutRenderers/Wrappers/ReplaceNewLinesLayoutRendererWrapper.cs
@@ -1,10 +1,6 @@
 ﻿using NLog.Config;
 using System;
-using System.Collections.Generic;
 using System.ComponentModel;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
 
 namespace NLog.LayoutRenderers.Wrappers
 {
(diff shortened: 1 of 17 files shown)


304NotModified:
Thanks!

It seems that to unused `using` are used? 

What's the benefit of the config property?

flower189:
I thought I had removed the unused `using`s and that property. It is redundant, I added it before I realised initially.

304NotModified:
OK. You need to copy the file header of another file to it to fix the failing unit test.

304NotModified:
I looks nice! short and clean :+1: 

Maybe we could also make it configurable how to replace the newline? Because I was expecting "" (no chars), but a space is also an option. I propose to add the property: `public string replacement`. The default for space is then fine.

What do you think?

flower189:
I like the idea. I was thinking it needed an option to control the separation character/string. I figured having no separator would make it hard to read.

304NotModified:
hint: the `[DefaultValue]` is only a marker and won't set the value. (it's a common mistake and therefor I have created an unit test for it.)

304NotModified:
You can ignore the failing unit test. That one is failing really sometimes (so there is really something wrong but not in this code)

flower189:
Sorry, I really should have checked how the other properties had been implemented.

304NotModified:
No problem.  It's just a matter of style.

304NotModified:
We will include this in 4.1 :) 

One last thing: can you document this on the wiki? (new page) 

Maybe also reference from https://github.com/NLog/NLog/wiki/Replace-Layout-Renderer

flower189:
Added a documentation page to the wiki here: https://github.com/NLog/NLog/wiki/Replace-NewLines-Layout-Renderer

304NotModified:
OK thanks,

I added also your examples and a 4.1 notice.

But I missed something. How does the test `ReplaceNewLineOneLineTest` work? And can't we replace with empty string now? It seems that `ReplaceNewLineOneLineTest` is wrong.

flower189:
The ReplaceNewLineOneLineTest was added for a check that it wasn't applying when it wasn't meant to.

The default could be replaced with an empty string. My preference is to have a space to help keep it a little bit more readable by default.

304NotModified:
The default of space is good.

But replace with a empty value is a corner case. We need an unit test for that. Can you add this unit test?

304NotModified:
Thanks! It has been merged :)

PR #801, opened 2015-07-17
Added the functionality to notify user code when log files are archived

BrandonLegault:
I created a Set-Only property that takes a delegate to be fired when NLog archives its log files. The delegate is called in a thread so as not to block the logger, and is passed the name of the created archive file.

Also added unit tests.


304NotModified:
Nice idea. This is of course only available from the API.

I think it would be also nice if we have one for removal of the file?

Also, can you post some examples how to use it? I will use them in the new post.

BrandonLegault:
Removal of the file at what time? Do you mean on archive file deletion when we're > MaxArchiveFiles?

I will post some examples...

BrandonLegault:
It's pretty simple to use. Here are the 4 ways to use the `OnFileArchival` property:

```C#
public class Foo
{
	public void ArchiveNotificationWithClassMethod()
	{
		var ft = new FileTarget
		{
			FileName = "FileName.log",
			ArchiveFileName = "ArchiveFileName.log",
			ArchiveAboveSize = 1000,
			Layout = "${message}",
			OnFileArchived = this.HandleFileArchivalClassMethod
		};
	}

	private void HandleFileArchivalClassMethod(string newlyCreatedArchiveFileName)
	{
		//Notify the user that an archive file was created.
		// or do some other archive-file-related tasks.
	}

	public void DontHandleArchiveNotification()
	{
		//Redundant because null is the default. Nothing will be done when the file is archived.
		var ft = new FileTarget
		{
			FileName = "FileName.log",
			ArchiveFileName = "ArchiveFileName.log",
			ArchiveAboveSize = 1000,
			Layout = "${message}",
			OnFileArchived = null 
		};
	}

	private FileTarget.FileArchivedHandler _myArchiveHandler;
	public void ArchiveNotificationWithClassField()
	{
		this._myArchiveHandler = newlyCreatedArchiveFileName =>
		{
			//Notify the user that an archive file was created.
			// or do some other archive-file-related tasks.
		};

		var ft = new FileTarget
		{
			FileName = "FileName.log",
			ArchiveFileName = "ArchiveFileName.log",
			ArchiveAboveSize = 1000,
			Layout = "${message}",
			OnFileArchived = this._myArchiveHandler
		};
	}

	public void ArchiveNotificationWithInlineMethod()
	{
		var ft = new FileTarget
		{
			FileName = "FileName.log",
			ArchiveFileName = "ArchiveFileName.log",
			ArchiveAboveSize = 1000,
			Layout = "${message}",
			OnFileArchived = newlyCreatedArchiveFileName =>
			{
				//Notify the user that an archive file was created.
				// or do some other archive-file-related tasks.
			}
		};
	}
}
```

Hope that helps.

304NotModified:
Yes the delete when max. There are multiple `file. Delete`  I propose to replace them with a function (which also have an optional event handler)

304NotModified:
As I see correctly (on mobile now).  You use lamdas en not events? 

We should use events (to subscribe handlers) 
https://msdn.microsoft.com/en-us/library/edzehd2t(v=vs.110).aspx

BrandonLegault:
I used lambdas for this, yes. The reason being if we raise an event, (this is my understanding, I may be wrong) the user code is called in the same thread as the event raiser. This means that if the user's code takes a long time to execute, the archive operation will block. I'm not familiar enough with the inner workings of NLog to know if blocking this operation will throw off the timing of the logging.

An alternative solution is to raise the events in a thread in the same way that I have it now. I try to avoid events in .NET as much as is possible to be honest.

304NotModified:
as was under the impression that `delegate` also runs in the current thread. It's just a "function pointer"?

304NotModified:
Some info:

-  http://stackoverflow.com/questions/5600930/what-to-use-delegate-event-or-funct
- http://www.codeproject.com/Articles/741064/Delegates-its-Modern-Flavors-Func-Action-Predicate

I'm not sure if `delegate` is legacy

BrandonLegault:
Yes you're correct it is just a function pointer.

If you have a look at my implementation, I fire the function in a thread
(or task for .NET 4.5) to remedy that. I could do the same with an event If
that's preferred.

304NotModified:
I'm not sure if creating a thread is a good idea. That can also be done in the referenced function?

BrandonLegault:
Well again, I only created the thread to prevent the file archival operation from blocking and waiting on hostile code. Whether we do this with an event or a delegate, if it isn't in a task/thread I think we'll still have that problem, no?

304NotModified:
Yes I understand that.  But is sounds like premature optimalization. And it's even doable if it's needed - if we just do a console write,  we have created an unneeded (expensive) thread.

304NotModified:
If you you need a thread (or task of whatever), you can still create them in the func /delegate. 

I propose to remove this thread creation in this PR.

BrandonLegault:
Okay, I will remove the threading/task and leave it as a delegate.

304NotModified:
Did something went wrong? I see a lot of unrelated changes (of other PRs)

BrandonLegault:
I'm not sure what happened there. I had the pull request set, and then GH
said I was behind NLog:master so I re-based my pull request. On commit,
that's what happened.

On Thu, Jul 23, 2015 at 1:34 PM, Julian Verdurmen <notifications@github.com>
wrote:

> Did something went wrong? I see a lot of unrelated changes (of other PRs)
>
> —
> <https://github.com/NLog/NLog/pull/801#issuecomment-124176111>.
>



-- 
Brandon Legault

304NotModified:
How did you rebase?

BrandonLegault:
git pull upstream master

where upstream is NLog and master is my local master, then

git rebase master myBranch

then

git push origin master

and

git push origin myBranch


On Thu, Jul 23, 2015 at 1:43 PM, Julian Verdurmen <notifications@github.com>
wrote:

> How did you rebase?
>
> —
> <https://github.com/NLog/NLog/pull/801#issuecomment-124178113>.
>



-- 
Brandon Legault

BrandonLegault:
As I was typing that out I realized the problem. I think (I'm more of a SVN guy so Git is not my forte).

When I did the rebase from the upstream pull, I overwrote the changes in my feature branch locally, which is where the other commits from the NLog:master came from. If my pull request is merged, the changes should go through seemlessly, as NLog:master already has them.

My bad.

304NotModified:
I understand you problem (I'm also more experienced with SVN). Any way, I also do this to merge with the master:

```
git fetch upstream
git checkout <youbranch>
git merge upstream/master
```

The changes in this PR a issue, because we cannot merge it now easily. Maybe this could help? http://sethrobertson.github.io/GitFixUm/fixup.html

BrandonLegault:
I don't know what the problem is and how to fix it. I've downloaded two Visual Git applications and read multiple pieces of help documentations. I'm going to close this pull request and re-branch to make the change again.

304NotModified:
Good luck.  If you need some help, let us know

Issue #784, opened 2015-07-01
refactor WcfLogReceiverClientFacade, fix backwardscomp


304NotModified:
Fix backwardscompatiblity with 3.2
---

The change in NLog 4.0 was unwanted breaking. We need to fix this:

The problem:
- WcfLogReceiverClientFacade won't inherit from `ClientBase<...>`. But in 3.2.0 `WcfLogReceiverClient` was returned, and this class did implement  `ClientBase<...>` (which contains 'Endpoint` and `ClientCredentails` properties). Thus: WcfLogReceiverClientFacade  should inherit from `ClientBase<..>`. But this is difficult to fix. We have to find a good solution for this. 

See: #783 and #561

Refactor or remove WcfLogReceiverClientFacade
---
The class `WcfLogReceiverClientFacade` introduced in 4.0 (#565), contains unwanted stuff:

- there is a lot of code duplication in the constructors
- every property is implemented with the same if-else constructions
- there are two properties, `m_twoWayClient` and `m_oneWayClient` and there is always only one used, hence, it should be one property.
- there are some Hungarian names. e.g. `m_oneWayClient` should be without the prefix `m_`.
- `ILogReceiverClient` and `ILogReceiverOneWayClient` are almost identical
- `WcfLogReceiverClient` and `WcfLogReceiverOneWayClient` are code duplication?

Solutions:
- We should clean this up and create **one** factory method with do the   `if (useOneWay)` check. It should return an `interface` and only that interface should be used in the rest of the class.
- Clean up code duplication as much as possible. 
 
code: https://github.com/NLog/NLog/blob/master/src/NLog/LogReceiverService/WcfLogReceiverClientFacade.cs

304NotModified:
code introduced by  #565

MartinTherriault:
I just finished doing it in my fork. Can you take a look? I'm not sure how things should be named. I just used "V2" but I'll change it to something else if you want to.

I restarted from the original code and added the method to the client interface. For the server interface, I inherited the original interface and added the new method. The target kept its parameter to indicate which method to use.

There is however one possibility in which the changes could break old code, it is if someone was inheriting directly the client interface, they would have to implement the new method for the code to compile. I tried to make it work with two client interfaces, but I get errors with the .config file that I have not managed to fix yet.

304NotModified:
Great!! Just make a PR please. That's easier to discuss.

304NotModified:
fixed by #874

Issue #783, opened 2015-07-01
LogReceiverWebServiceTarget backwards compatible problem


tryhp3:
After the update (to NLog 4.0) I have a serious problem:
1. 'NLog.CustomExtendedService.LogReceiverSecureService.CreateWcfLogReceiverClient()': return type must be 'NLog.LogReceiverService.WcfLogReceiverClientFacade' to match overridden member 'NLog.Targets.LogReceiverWebServiceTarget.CreateWcfLogReceiverClient()'	
2. New class WcfLogReceiverClientFacade does not include: ClientCredentials and Endpoint.
![przechwytywanie](https://cloud.githubusercontent.com/assets/4246990/8451869/c678f2cc-1fe6-11e5-9056-e5c8f9726a2b.PNG)

304NotModified:
Hi,


- I don't fully understand the problem of point 1. Yes it is a breaking change, but what is the serious problem?
-  The `ClientCredentials` was also not in 3.2.0 and before? (in don't see them in `WcfLogReceiverClient` (https://github.com/NLog/NLog/blob/v3.2.0.0/src/NLog/LogReceiverService/WcfLogReceiverClient.cs)
- There is an endpoint address in the `LogReceiverWebServiceTarget`. Could that help?

304NotModified:
PS: I noticed that `WcfLogReceiverClientFacade ` is a mess and should be refactored. See #784

304NotModified:
I see the problem: 

- WcfLogReceiverClientFacade won't inherit from `ClientBase<...>`. But in 3.2.0 `WcfLogReceiverClient` was returned, and this class did implement  `ClientBase<...>` (which contains 'Endpoint` and `ClientCredentails` properties). Thus: WcfLogReceiverClientFacade  should inherit from `ClientBase<..>`

304NotModified:
fixed by #874

PR #803, opened 2015-07-19
Fix issues with WcfLogReceiverClient

MartinTherriault:
(see #784)
I restarted from the original code and added the method to the client interface. For the server interface, I inherited the original interface and added the new method. The target kept its parameter to indicate which method to use.

There is however one possibility in which the changes could break old code, it is if someone was inheriting directly the client interface, they would have to implement the new method for the code to compile.

I do not think we can do otherwise:
- If we use two client interfaces (V1 and V2) which are exclusives, we end up with the same issues as before because we must inherit from ClientBase< T >
- If we use two client interfaces which V2 inherit from V1, it is fine until you try to configure a V1 endpoint, which will fail (throws) since we used a ClientBase< V2 > and it does not find any ClientBase< V1 >


src/NLog/LogReceiverService/ILogReceiverServerV2.cs
```diff
@@ -41,14 +41,14 @@ namespace NLog.LogReceiverService
     /// Service contract for Log Receiver server.
     /// </summary>
     [ServiceContract(Namespace = LogReceiverServiceConfig.WebServiceNamespace)]
-    public interface ILogReceiverOneWayServer
```
304NotModified: We lost #if WCF_SUPPORTED

src/NLog/LogReceiverService/WcfLogReceiverClient.cs
```diff
@@ -237,12 +265,22 @@ protected override ILogReceiverClient CreateChannel()
...
+            }
+            else {
+                return ((ILogReceiverClient)this).BeginProcessLogMessages(events, callback, asyncState);
+            }
         }
 
         private object[] OnEndProcessLogMessages(IAsyncResult result)
         {
-            ((ILogReceiverClient)this).EndProcessLogMessages(result);
+            if (UseOneWayCallsToServer) {
```
304NotModified: We don't need to copy the `if (UseOneWayCallsToServer)` (code duplication). I advise a helper method who's returning a `ILogReceiverClient`

304NotModified:
I'm not sure if this is the best solution. I will also do a try, and then make a PR. Let see if we can fix all the issues.

304NotModified:
Hi, it tried also something. See #821. I tried also removing the duplicate code in the clients, but failed

MartinTherriault:
Ok, so do you want me to do now?

304NotModified:
Maybe check the code and give your opinion? :)  which approach is the best? Would it work?

MartinTherriault:
I'm going to try to give a neutral opinion on both approach, but keep in mind I'm biased toward the #803 one. I believe both would work but one would require a higher maintenance effort, hence why I favor #803.

**Facade (#821)**
Pros
 * Ensure the WCF contract in non breaking, since we do not modify the original one
 
Cons
 * `ClientBase` inheritance force to have code duplication
 * Since the facade itself (`WcfLogReceiverClient`) cannot inherit from client base, it will break old code where someone is using a `ClientBase` property we did not implement. Using the proxy instead would likely fix the issue, otherwise a new method/property would be needed on `WcfLogReceiverClient`. (#783)
 * A server would need two endpoints to support both method since it is different contracts
 
**WCF versionning (#803)**
Pros
 * Adding a method to the contract is non breaking and inheriting the server contract ensure a server will not need any changes unless he wants to support the new method
 * A server can support both methods with a single endpoint
 * Client class (`WcfLogReceiverClient`) inherit `ClientBase`
 
Cons
 * Will break old code if someone was inheriting `ILogReceiverClient`. Implementing the new methods in the old code would fix the issue.
 
 
In both case, to support the new one way method, a server will have to change code, be it inherit another interface, implement it, add or not a new endpoint, ...

304NotModified:
Thanks for the analysis! 

I don't use WCF much. But I think that most of the time people only would implement one of the server interfaces?

MartinTherriault:
Yes most of the time.

I do not think having to implement both methods is a burden though. Keep in mind if you only want to use the old two way method, the server can use the old interface and do not have to implement the other method. Also both method have the same signature so they can call the same processing method. The difference (one way vs two way) is done behind the curtains by the WCF layer.

kevindaub:
This might work for the code duplication and might combine both ideas.  This will also bring in the ClientBase.

```csharp
public abstract class WcfLogReceiverClientBase<TInterface> : ClientBase<TInterface>
        where TInterface : ILogReceiverClient
{
}

public sealed class WcfLogReceiverOneWayClient : WcfLogReceiverClientBase<ILogReceiverOneWayClient>, ILogReceiverOneWayClient
{
}
```

I'll take #821 and test out the above idea and I'll get something together this weekend.

304NotModified:
superseded by #874

Issue #709, opened 2015-05-19
FileTarget.FileName.ToString() returns value enclosed in single quotes


grbinho:
Hi,

as mentioned in the title, FileTarget.FileName.ToString() returns value enclosed in single quotes.
It's an easy Trim fix, but not expected.

Ivan

304NotModified:
Hi,

That's because [ FileTarget.FileName](http://nlog-project.org/documentation/v3.2.1/html/P_NLog_Targets_FileTarget_FileName.htm) is a [Layout](http://nlog-project.org/documentation/v3.2.1/html/T_NLog_Layouts_Layout.htm), and the ToString of that quotes a fixed value (no layout renderers used)

PR #815, opened 2015-07-25
Overhaul variables [wip]

### LINKED ISSUE #694: How can I get NLog to reevaluate the config with variable values set from code?

304NotModified:
In the past we could not change the value of a variable, because the variable was rendered while parsing the XML - the old value was lost. (issue #694.). Also, we could not use variables from the API, which was a bit strange.

Both are now fixed, by doing things differently:
- variables can we now  rendered by a layout renderer (`${var:username} `instead of `${username}`)
- the variables can we still set from the XML config, with the same syntax. (e.g.    `<variable name='username' value='admin' />`
- the "old" variables 

Please not:
- The XML is fully backwards-compatible.
- The API isn't, the Variables in config has been changed from `Dictionary<string, string>` to `IDictionary<string, Layout>`. This can not be prevented, without tricks or re-parse every Layout every time.
- This new style won't replace fully the old variables, as I can only be used in Layouts.
- Also the `ToString()` of `Layout`/`SimpleLayout` has been changed. It now returns the original value (before parse). Before it printed the Text in quotes. This was also odd behaviour (because `(Layout) (layout.ToString) == layout`)  (#709)
 

 

- variables are now Layouts
- included new layout rendererer: VariableRenderer
- this will make it easy to change the value of variable on runtime or even remove it
- VariableRenderer has name and default (if not set)

fixed #694.

### DIFF
--- a/src/NLog/Config/LoggingConfiguration.cs
+++ b/src/NLog/Config/LoggingConfiguration.cs
@@ -33,6 +33,7 @@
 
 using System.Globalization;
 using System.Linq;
+using NLog.Layouts;
 
 namespace NLog.Config
 {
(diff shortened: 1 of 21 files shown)


304NotModified:
Too bad the SimpleLayout.toString has too much impact. I will revert that (new PR)

Issue #528, opened 2015-01-25
Feature request: define/set variable inside a logger


304NotModified:
I would like to defined (or change) a variable for a particular logger.

For example:

``` xml
<?xml version="1.0" encoding="utf-8"?>

<nlog>
  <targets>
    <target name="db" xsi:type="Database" keepConnection="true" useTransactions="false"
            connectionString="server='...';user id='...';password='...';database='...'"
            commandText="">
      <parameter name="@message" layout="${message}" />
      <parameter name="@category" layout="${category}" />
    </target>
  </targets>
  <rules>
    <logger name="*.sqlExecutor" minlevel="Debug" writeTo="db">
      <!--new, variable for this logger-->
      <variable name="category" value="queries" />
    </logger>
  </rules>
</nlog>
```

snakefoot:
Alternative it should be possible to specify on a Logger, that it should append metadata as LogEvent properties. Then these can be reference elsewhere (Ex. CorrelationId or other context dependent data).

PR #816, opened 2015-07-25
Overhaul NLog variables

### LINKED ISSUE #694: How can I get NLog to reevaluate the config with variable values set from code?

304NotModified:
In the past we could not change the value of a variable, because the variable was rendered while parsing the XML - the old value was lost. (see issue #694 and test #790). Also, we could not use variables from the API, which was a bit strange.

Both are now fixed, by doing things differently:
- variables are now rendered by a layout renderer (`${var:username} `instead of `${username}`) - the `VariableRenderer`
- the variables can we still set from the XML config, with the same syntax. (e.g.    `<variable name='username' value='admin' />`
- we can now set, add and remove variables from the API.
- the "old" variables can be still used (same syntax, same cannot-change behavior)
- this will prepares us for the feature: #528

Please note:
- The XML is fully backwards-compatible.
- We can now render the variables in two ways (old and new)
- The API has a small breaking change, the `Variables` in config has been changed from `Dictionary<string, string>` to `IDictionary<string, SimpleLayout>`. This can not be prevented, without tricks or re-parse every `Layout` every time. 
- We needed `SimpleLayout` (instead of `Layout`), because it should be easy to get the OrginalText. The OrginalText` property is also added to `SimpleLayout` in the PR. (and `.ToString()` has a strange behaviour, fixing that had to much impact, see #815 )
- This new style won't replace fully the old variables, as It can only be used in Layouts.
- VariableRenderer has name and default (if not set).

fixed #694.

### DIFF
--- a/src/NLog/Config/LoggingConfiguration.cs
+++ b/src/NLog/Config/LoggingConfiguration.cs
@@ -73,7 +73,7 @@ namespace NLog.Config
         /// <summary>
         /// Gets the variables defined in the configuration.
         /// </summary>
-        public Dictionary<string, SimpleLayout> Variables
+        public IDictionary<string, SimpleLayout> Variables
         {
             get
             {
--- a/src/NLog/Config/LoggingConfiguration.cs
+++ b/src/NLog/Config/LoggingConfiguration.cs
@@ -33,6 +33,7 @@
 
 using System.Globalization;
 using System.Linq;
+using NLog.Layouts;
 
 namespace NLog.Config
 {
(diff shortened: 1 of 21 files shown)


304NotModified:
Status done. Feedback appreciated :)

304NotModified:
Status missing lasr push....

PR #818, opened 2015-07-26
Updated InternalLogger to Create Directories If Needed

### LINKED ISSUE #802: Internal Logger should create directories when needed
Hi,

I'm using Internal Logging with following NLog.config.

	      internalLogFile="c:\Logs\InternalLog.log" 
	      internalLogLevel="Trace">
	...
	</nlog>

If the folder "c:\Logs" already exists, Internal Logging is appending.
But, if the folder doesn't exist, it not appending.

I hope insert this code before 
https://github.com/NLog/NLog/blob/4.0.1/src/NLog/Common/InternalLogger.cs#L323

`Directory.CreateDirectory(Path.GetDirectoryName(filename));`

The current NLog version: 4.0.1

Thanks,

kevindaub:
Updated the InternalLogger to create directories if they are needed.

Resolves #802

### DIFF
--- a/src/NLog/Common/InternalLogger.cs
+++ b/src/NLog/Common/InternalLogger.cs
@@ -63,6 +63,11 @@ namespace NLog.Common
             LogToConsoleError = GetSetting("nlog.internalLogToConsoleError", "NLOG_INTERNAL_LOG_TO_CONSOLE_ERROR", false);
             LogLevel = GetSetting("nlog.internalLogLevel", "NLOG_INTERNAL_LOG_LEVEL", LogLevel.Info);
             LogFile = GetSetting("nlog.internalLogFile", "NLOG_INTERNAL_LOG_FILE", string.Empty);
+            if (!string.IsNullOrEmpty(LogFile))
+            {
+				CreateDirectoriesIfNeeded(LogFile);
+			}
+			
             Info("NLog internal logger initialized.");
 #else
             LogLevel = LogLevel.Info;
@@ -320,7 +325,6 @@ namespace NLog.Common
                 var logFile = LogFile;
                 if (!string.IsNullOrEmpty(logFile))
                 {
-                    CreateDirectoriesIfNeeded(logFile);
                     using (var textWriter = File.AppendText(logFile))
                     {
                         textWriter.WriteLine(msg);
--- a/src/NLog/Common/InternalLogger.cs
+++ b/src/NLog/Common/InternalLogger.cs
@@ -320,6 +320,7 @@ namespace NLog.Common
                 var logFile = LogFile;
                 if (!string.IsNullOrEmpty(logFile))
                 {
+                    CreateDirectoriesIfNeeded(logFile);
                     using (var textWriter = File.AppendText(logFile))
                     {
                         textWriter.WriteLine(msg);
@@ -424,6 +425,21 @@ namespace NLog.Common
                 }
 
                 return defaultValue;
+            }
+        }
+        
+        private static void CreateDirectoriesIfNeeded(string filename)
+        {
+            try
+            {
+                Directory.CreateDirectory(Path.GetDirectoryName(filename));
+            }
+            catch (Exception exception)
+            {
+                if (exception.MustBeRethrown())
+                {
+                    throw;
+                }
             }
         }
 #endif
--- a/src/NLog/Common/InternalLogger.cs
+++ b/src/NLog/Common/InternalLogger.cs
@@ -101,10 +101,13 @@ namespace NLog.Common
             set
             {
                 _logFile = value;
+
+#if !SILVERLIGHT
                 if (!string.IsNullOrEmpty(_logFile))
                 {
                     CreateDirectoriesIfNeeded(_logFile);
                 }
+#endif
             }
         }
 
--- a/src/NLog/Common/InternalLogger.cs
+++ b/src/NLog/Common/InternalLogger.cs
@@ -88,7 +88,7 @@ namespace NLog.Common
         public static bool LogToConsoleError { get; set; }
 
         /// <summary>
-        /// Gets or sets the name of the internal log file.
+        /// Gets or sets the file path of the internal log file.
         /// </summary>
         /// <remarks>A value of <see langword="null" /> value disables internal logging to a file.</remarks>
         public static string LogFile
--- a/src/NLog/Common/InternalLogger.cs
+++ b/src/NLog/Common/InternalLogger.cs
@@ -456,6 +456,10 @@ namespace NLog.Common
                 {
                     throw;
                 }
+                else
+                {
+                    Error("Cannot create needed directories to {0}. {1}", filename, exception.Message);
+                }
             }
         }
 #endif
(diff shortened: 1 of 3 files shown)


src/NLog/Common/InternalLogger.cs
```diff
@@ -320,6 +320,7 @@ private static void Write(LogLevel level, string message, object[] args)
                 var logFile = LogFile;
                 if (!string.IsNullOrEmpty(logFile))
                 {
+                    CreateDirectoriesIfNeeded(logFile);
```
304NotModified: Please move to constructor (performance)

tests/NLog.UnitTests/Common/InternalLoggerTests.cs
```diff
@@ -336,6 +336,58 @@ public void TimestampTests()
                 Assert.Contains(expectedDateTime + ".", str);
             }
         }
+        
+        [Fact]
+        public void CreateDirectoriesIfNeeded()
```
304NotModified: Nice test :+1:

304NotModified:
Thanks! Please check the first note.

src/NLog/Common/InternalLogger.cs
kevindaub: Unfortunately, that will make my unit test moot, but agree with the performance.

304NotModified:
That's a lot of failed tests :/

kevindaub:
That's a lot better.  I forgot to remove out the unit test.  Thoughts on moving all the static logic to an instance class that is initialized as a singleton.  I think that will give you the performance and the unit testibility.

304NotModified:
I really liked the unit test. I forgot that calling the method in a static ctor is difficult to mock. But we can fix this :)

New proposal:

- move the code to the setter (`InternalLogger.LogFile`)
- restore unit test

304NotModified:
I see the failing test. Dunno why they fail. Maybe you can change ` Assert.True` with `Assert.Equal` in those cases. (which is better)

kevindaub:
The failed unit test was not my unit test.  However, I do notice that the tests in there seem to randomly fail.  It varies on my computer which one fails.  I take a closer to look to see if my change caused that.

kevindaub:
My changes seems to have made the tests unstable.  It might be threading since it's a static class.  I'll dig in a bit more.

304NotModified:
yes, there are some tests which has race conditions. But that aren't those two failing tests. Do those tests successfully run on your local computer?

kevindaub:
My test seems to make the race conditions worse. As long as my test doesn't
get executed, things are fine.  Perhaps making my unit test a skip?

On Monday, July 27, 2015, Julian Verdurmen <notifications@github.com> wrote:

> yes, there are some tests which has race conditions. But that aren't those
> two failing tests. Do those tests successfully run on your local computer?
>
> —
> <https://github.com/NLog/NLog/pull/818#issuecomment-125111745>.
>


-- 
Sent from Gmail Mobile

304NotModified:
I think you should 'revert' the value of `InternalLogger.LogFile` at then end of your unit test.

kevindaub:
Resetting the LogFile back to the previous value resolved the issue.  It worked repeatedly on my machine, which it was not doing before.  Thanks for the help.

kevindaub:
Are we good with this pull request?

src/NLog/Common/InternalLogger.cs
```diff
@@ -426,6 +443,21 @@ private static T GetSetting<T>(string configName, string envName, T defaultValue
...
+        
+        private static void CreateDirectoriesIfNeeded(string filename)
+        {
+            try
+            {
+                Directory.CreateDirectory(Path.GetDirectoryName(filename));
+            }
+            catch (Exception exception)
+            {
+                if (exception.MustBeRethrown())
```
304NotModified: We need an log to the internallog here.

304NotModified:
Looks good! one small thing (because 'empty' catches are confusing)  - see the code note. Can you fix that?

src/NLog/Common/InternalLogger.cs
kevindaub: Do you mean something like this?

```csharp
            try
            {
                Directory.CreateDirectory(Path.GetDirectoryName(filename));
            }
            catch (Exception exception)
            {
                InternalLogger.Error("Cannot create needed directories {0}, Exception : {1}", Path.GetDirectoryName(filename), exception);
                {
                    throw;
                }
            }
```

kevindaub:
Added a log if we are not rethrowing the exception.  But I think this actually might cause issues unless we log only to the console.  Whether we rethrow or not, the Write method uses the LogFile, which will probably not let that method finish.  Thoughts?

304NotModified:
You're worried about infinite loops? I don't think this is a problem, but maybe we should unit test it to be sure.

tests/NLog.UnitTests/Common/InternalLoggerTests.cs
```diff
@@ -336,6 +336,67 @@ public void TimestampTests()
...
+                InternalLogger.Log(LogLevel.Info, "III");
+
+                AssertFileContents(tempFile, expected, Encoding.UTF8);
+                Assert.True(File.Exists(tempFile));
+            }
+            finally
+            {
+                // Reset LogFile to the previous value
+                InternalLogger.LogFile = previousLogFile;
+
```
304NotModified: please revert all the InternalLogger properties, to be sure. Maybe we should create a helper for that.

kevindaub:
I did a small test myself and my fear was lacking any solid basis.  I did a small update.  Besides your last comment.  I think it's good.

kevindaub:
Added an additional check for the failing test.  It looked like there was inner type that was auto generated by the compiler for the ReplaceLayoutRendererWrapper.  Running the test again with the additional check of t.IsPrivateNested showed that this was the only generated class.

304NotModified:
I will fix #839 soon and then we can use it in this PR.

304NotModified:
I recently enabled the Silverlight builds on appveyor.  Probably this will cause the build errors

304NotModified:
@kevindaub 

I created a separate PR for your fix. (https://github.com/NLog/NLog/pull/844). Thanks for the fix!

It should not lead to merge conflicts in this PR.

kevindaub:
Great. Let me know if you need anything after the other pull request is
merged.

On Wednesday, August 12, 2015, Julian Verdurmen <notifications@github.com>
wrote:

> @kevindaub <https://github.com/kevindaub>
>
> I created a separate PR for your fix. (#844
> <https://github.com/NLog/NLog/pull/844>). Thanks for the fix!
>
> It should not lead to merge conflicts in this PR.
>
> —
> <https://github.com/NLog/NLog/pull/818#issuecomment-130457656>.
>


-- 
Sent from Gmail Mobile

304NotModified:
Yes I let you know. :)

it's almost fixed :)

304NotModified:
#844 isn't working as expected. So we don;t have to change something in this PR for now. .

PR #821, opened 2015-07-26
Fix issues with WcfLogReceiverClient - alternative

304NotModified:
More backwardscompatible to v4. Both one and two way are now unit tested with a `ServiceHost`

Keep the facade idea, but a bit different
- `WcfLogReceiverClient` is renamed to `WcfLogReceiverOneWayClient`. We have now `WcfLogReceiverOneWayClient` and `WcfLogReceiverTwoWayClient`. Both will implement the same (new) interface.
- `WcfLogReceiverClientFacade` is renamed to `WcfLogReceiverClient`
- `WcfLogReceiverClient` (the facade) now delegates (almost) everything of `WcfLogReceiverOneWayClient`/`WcfLogReceiverTwoWayClient`.
- `WcfLogReceiverClient`  has a `Proxy` property, if still something is missing (cast is probably required for the end-user)

Things to think about:
- `ILogReceiverClient` is renamed to `ILogReceiverOneWayClient`. This is not needed, maybe revert it. (there are already some inconsistent namings to bad)
- The `WcfLogReceiverOneWayClient` and `WcfLogReceiverTwoWayClient` does contain code duplication due to the ClientBase inheritance. Difficult to solve with delegates. Maybe fix with T4.

TODO
---
 - [x] re-enable the warnings as error, and fix the xmldocs
 - [x] enable cookiecontainer


304NotModified:
- [] revert ILogReceiverClient is renamed to ILogReceiverOneWayClient ?

kevindaub:
ILogReceiverClient.cs - Breaks compatibility because it doesn't have any of the methods from 3.2.1
WcfLogReceiverClient.cs - No longer inherits from Clientbase<ILogReceiverClient>.
WcfLogReceiverOneWayClient.cs - is equivalent to WcfLogReceiver in 3.2.1

For better backwards compatibility to 3.2.1:
-Keep methods and attributes 3.2.1 in ILogReceiverClient
-Keep WcfLogReceiverClient using ILogReceiverClient, but mark as deprecated, so that we can clean up and move people to ILogReceiverOneWayClient/ILogReceiverTwoWayClient

For better backwards compatibility to 4.0.1:
-Keep methods and attributes in ILogReceiverOneWayClient 

For TwoWay Client:
-Duplicate ILogReceiverOneWayClient, but make it replace OneWay with TwoWay

Long Term Suggestions:
-As already suggested, use T4 template to generate ILogReceiverClient, ILogReceiverOneWayClient, and ILogReceiverTwoWayClient
-As already suggested, use T4 template to generate LogReceiverClients

304NotModified:
fixed by #874

PR #836, opened 2015-08-09
Appveyor Build SL 4

304NotModified:


304NotModified:
appveyor.yml is ignored, so this is not needed

PR #839, opened 2015-08-10
Unittest internallogger side effects [WIP]

304NotModified:
fixed with retrieve/restore with replace.

Call `RestoreStaticPropertyValues()`

We need only a teardown in xUnit...


304NotModified:
This new tick doesn't seem to work correctly... (10 failed unit tests)

Issue #837, opened 2015-08-09
Fix Mono unit tests


304NotModified:
The unit tests are not working for Mono with xbuild/travis:

- the solution build with xbuild if the unit test project is excluded 
- If we build the solution with the unit test project included, we got the error from xbuild: : "Input string was not in a correct format.")
- building the csproj for the unit test also fails: "The type or namespace name `Xunit' could not be found. Are you missing an assembly reference?".

See https://travis-ci.org/304NotModified/NLog/builds/ and #742

We should get this working, otherwise I think we should drop mono as it's impossible to validate if it's working...

304NotModified:
See https://travis-ci.org/304NotModified/NLog/builds/75989279

304NotModified:
Fixed by #860

PR #858, opened 2015-08-16
Fixed travis build with unit tests

kevindaub:
The issue was that the SamplesExtension project was not referenced correctly.

This also pulls in changes from #837.


304NotModified:
Great! 

We have to find out where monodevelop is referenced (appveyor error)

kevindaub:
I don't see any errors in the log that would indicate monodevelop.  Everything compiled and all the tests passed.

304NotModified:
nice, we will fix the unit tests at a later stage.

Issue #857, opened 2015-08-15
Fix logReceiverServiceTests one-way unit test


304NotModified:
https://github.com/NLog/NLog/pull/856

304NotModified:
Done

PR #861, opened 2015-08-17
LogReceiverServiceTests: Added one-way unit test (retry)

304NotModified:
real unit test with servicehost.

merged unit test from #821. 

more robust test case (sleep between messages for ordering)

supersedes reverted #854


304NotModified:
Closes #857

PR #874, opened 2015-08-24
Wcf receiver client

kevindaub:
* Added back compatibility with 3.2.1 for ILogReceiverClient (missing methods).
* The base class allows for the code duplication to removed.  Also, it allows for the inheriting class to invoke the correct channel from the ClientBase<TService>, so WCF should be happy at runtime, too.
* Added deprecation to ILogReceiverClient and its implementation in favor of the other two interfaces.
* Added deprecation to the LogReceiverWebServiceTarget's protected method that returned a concrete class in favor of a new method that returns an interface.

This should supersede #821.

- [x] Verify runtime with WCF
- [X] <s>Add some unit tests</s> (or run the ones available)
- [x] Documentation cleanup
- [x] Check compatibility again
- [x] Other framework cleanup


304NotModified:
I'm pretty sure that the generic clientbase won't work. :( See unit tests

kevindaub:
I saw that error. I don't think that's an issue. I've seen errors like that
before. I'll debug it tonight. It looked like we just need a config entry
or modify the custom binding.

On Monday, August 24, 2015, Julian Verdurmen <notifications@github.com>
wrote:

> I'm pretty sure that the generic clientbase won't work. :( See unit tests
>
> —
> <https://github.com/NLog/NLog/pull/874#issuecomment-134258792>.
>


-- 
Sent from Gmail Mobile

304NotModified:
Well then it Isn't backwards compatible?

304NotModified:
http://weblog.west-wind.com/posts/2007/Dec/14/ClientBaseT-and-a-Common-Base-Class

kevindaub:
That's not the same error. I'll investigate tonight.

On Monday, August 24, 2015, Julian Verdurmen <notifications@github.com>
wrote:

>
> http://weblog.west-wind.com/posts/2007/Dec/14/ClientBaseT-and-a-Common-Base-Class
>
> —
> <https://github.com/NLog/NLog/pull/874#issuecomment-134275606>.
>


-- 
Sent from Gmail Mobile

kevindaub:
Hooray!  The issue was the call to the base constructor.  The unit tests passed.  I was going to look through tonight to see if there are any compatibility issues, but it looks hopeful.

304NotModified:
Good work!

kevindaub:
The documentation looks good to me.  Also, the following might be compatibility concerns:
* WcfLogReceiverClient implements just IWcfLogReceiverClient vs ILogReceiverClient and ClientBase<ILogReceiverClient>
* LogReceiverWebServiceTarget uses WcfLogReceiverClient as return in the protected method vs the WcfLogReceiverClientFacade
* Overall, I think compatibility hits both 3.2.1 and 4.0.1

Lastly, there is a small amount of code duplication for the SilverLight in each of the WcfLogReceiverClient implementations.  I think this can be resolved by implementing a similar base class like the WcfLogReceiverClient.

Let me know if there's anything else you need.

304NotModified:
To be sure:

> Also, the following might be compatibility concerns:
> * WcfLogReceiverClient implements just IWcfLogReceiverClient vs ILogReceiverClient and ClientBase

This is a compatibility issue vs (only)  NLog 3.2?
> * LogReceiverWebServiceTarget uses WcfLogReceiverClient as return in the protected method vs the WcfLogReceiverClientFacade

This is a compatibility issue vs (only) NLog 4.0?

> Overall, I think compatibility hits both 3.2.1 and 4.0.1
Yes, too bad. But I think we improved the compatibly from  3.2 to 4.1 (4.0 was a mistake)

kevindaub:
* Both: WcfLogReceiverClient implements just IWcfLogReceiverClient vs ILogReceiverClient and ClientBase
* Removed the facade and reverted to the 3.2.1. As long as they didn't cast the return value to Clientbase<ILogReceiverClient>, the compatibility should be pretty close.  
 * 4.0.1: Returned WcfLogReceiverClientFacade
 * 3.2.1: Returned WcfLogReceiverClient
 * 4.1.0: Returned WcfLogReceiverClient

304NotModified:
Can't we fix the first one with a rename?

kevindaub:
Rename client to facade client?

On Thursday, August 27, 2015, Julian Verdurmen <notifications@github.com>
wrote:

> Can't we fix the first one with a rename?
>
> —
> <https://github.com/NLog/NLog/pull/874#issuecomment-135288750>.
>


-- 
Sent from Gmail Mobile

304NotModified:
rename `IWcfLogReceiverClient` to `ILogReceiverClient` ?

kevindaub:
That won't fix it. It's a binary incompatibility because we changed the
inheritance. But as long as most of the methods of client base are
available. It should be okay.

On Thursday, August 27, 2015, Julian Verdurmen <notifications@github.com>
wrote:

> rename IWcfLogReceiverClient to ILogReceiverClient ?
>
> —
> <https://github.com/NLog/NLog/pull/874#issuecomment-135413771>.
>


-- 
Sent from Gmail Mobile

MartinTherriault:
Finally managed to take a look at this. I tried it in my test scenarios and found no issues.

PR #884, opened 2015-08-30
Changes at MDLC to support .Net 4.0 and .Net 4.5

BrutalCode:
Changed compiler #if directive to support .Net 4.0 and 4.5

### DIFF
--- a/src/NLog/LayoutRenderers/MdlcLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/MdlcLayoutRenderer.cs
@@ -33,7 +33,7 @@
 
 namespace NLog.LayoutRenderers
 {
-#if NET4_5
+#if NET4_0 || NET4_5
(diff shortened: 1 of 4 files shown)


304NotModified:
And the unit tests?

BrutalCode:
I just changed the #if directive and compile it in 4.0. What kind of unit test can be written for this change? For references?

304NotModified:
There should be already unit tests for this. Those need also enabled for. Net 4

It should be in the test project and the class name is almost the same.

BrutalCode:
You are right! Sorry for oversight. I changed and pushed it too.

304NotModified:
Yes, those where all of them (I used "find all references" in the .Net 4.5 solution, and I checked if there were more `#if` statement in 1 file )

304NotModified:
So good catch!  :+1:

PR #894, opened 2015-09-03
Obsolete attribute doesn't specify the correct replacement

vbfox:
Calls to `Logger.Error(string, Exception)` were displaying a warning of :

    Use Error(LogLevel level, Exception exception, string message, params object[] args) method instead.

But the correct version to call doesn't contains the LogLevel


304NotModified:
Thanks, but you should change this in the Logger.tt.

Can you update the PR?

vbfox:
Yep, will do

304NotModified:
Thanks

PR #898, opened 2015-09-07
Resolves Internal Logging With Just Filename

kevindaub:
When setting the internal logging to just a filename, the create directories method would throw an exception because there was not a parent directory.

To resolve, we needed to check if there was any parent directories to create.  If yes, then create, otherwise, continue on.


304NotModified:
Thanks!

PR #915, opened 2015-09-16
Improve exception handling (make it consistent and clear) 

tetrodoxin:
and refactored respective occurrences.

It's the first shot and definitely object to discuss, but I had to start somewhere.


tetrodoxin:
How can I actually start the tests from command line? Since the VS test runner creates bin/ and obj/ directories which bother source code tests.

tetrodoxin:
Well, I nearly expected something like this, when I left out the configuration exception. So... back to scratch.

304NotModified:
Looks really good! 2 things I'm thinking we maybe should handle :

- `throw new Exception`  maybe a NLog specific one? 
- `throw` if throw exceptions is false.

304NotModified:
Msbuild NLog.proj /t:runtests? (now on mobile)

304NotModified:
Maybe you can pass the expected exception as type argument? 

Eg

`HandleException<TException > where TException = class,  new() `

PS still on mobile. Ignore the syntax errors :p

tetrodoxin:
So, I'm sort of back after some time at work. 
I also thought about using generics with new(), but this way you have absolutely no chance to provide any error message text or inner exception, so the exception thrown would have the correct type, but no further content, which I considered a bad deal.
A factory delegate would be possible too, but I'm not sure if
`ex.HandleException(x => new NLogConfigurationException("Error in...", x))`
really would be a real benefit.

tetrodoxin:
An idea I head was something like a wrapper `NLogWrapperException<TException>` which possibly also could be casted to its inner exception (TException) and then extend XUnit's `Assert.Throws<TException>()` to check for `TException` and for `NLogWrapperException<TException>`, so testing would not be a real additional effort.

304NotModified:
> I also thought about using generics with new(), but this way you have absolutely no chance to provide any error message text or inner exception, so the exception thrown would have the correct type, but no further content, which I considered a bad deal.

Yes you're right! I forgot that `Message` property is readonly for an `Exception`. 



> An idea I head was something like a wrapper NLogWrapperException<TException> which possibly also could be casted to its inner exception (TException)

Not sure if this is nice, but...

> XUnit's Assert.Throws<TException>() to check for TException and for NLogWrapperException<TException>, so testing would not be a real additional effort.

This is a nice idea! No need for a cast I think, we can create a ShouldThrows<TException>() in `NLogTestBase`.

304NotModified:
> A factory delegate would be possible too, but I'm not sure if
> ex.HandleException(x => new NLogConfigurationException("Error in...", x))
> really would be a real benefit.

Or maybe something like:` ex.HandException<TException> where TException : IExceptionFactory`


and 

```
class ExceptionFactory
{
 static TException  Create<TException > (string message, Exception innermessage)
....

}
```

304NotModified:
Thinking of this from scratch: 

```c#
bool throw = ex.LogException();
if(throw) throw;
```

but then we're almost back... (`MustBeRethrown` is only replaced by `LogException`)

tetrodoxin:
It seems it would be a breaking change in any way we consider it useful. So spare it for later? Just fix the problematic occurences?

304NotModified:
> It seems it would be a breaking change in any way we consider it useful.

Well not if we extend/rename the `MustBeRethrown `? 

We get really soon complains if introduce a breaking change ;)

tetrodoxin:
If we just extend it the way I tried, we will break some tests that rely on specific exceptions to be thrown, which were the reason for me thinking about adjusting/extending the tests by accepting this wrapper exception. Since we cannot properly rethrow a catched exception outside the catch-block, we're stuck between redesign (breaking change) and very little changes.

304NotModified:
> We're stuck between redesign (breaking change) and very little changes.

Agree with this. I'm doubting if there will be a huge benefit for the breaking change - yes it looks better, but every change in an unit test will result in changes needed by our users. So if we can fix it without breaking change, that's preferred, despite I really liked our approach.

tetrodoxin:
To summarize, by now we are at something like this:
 ```csharp
if(ex.LogAndCheckRethrow()) throw;
```
aren't we? `LogAndCheckRethrow()` extension method would do right this, the logging and the check, if rethrowing is necessary. This would at least not really break any current behavior but ensure the way it was originally designed, rethrow the exception when `rethrowExceptions=true` and therefore satisfy issue   #277 
Tenminste iets, hè?

**Comment by 304NotModified** [comment]
Yes,  I think that is the best option. 

Sehr schön! :p

**Comment by 304NotModified** [comment]
Need some help on this @tetrodoxin ?

**Comment by tetrodoxin** [comment]
I'm abroad these days and seldom online, will be back next week, at the latest.


-------- Original Message --------

**Comment by 304NotModified** [comment]
OK thanks for the info!

**Comment by tetrodoxin** [comment]
I had to adjust one test class, namely VariableLayoutRendererTests, because it did not fit into the 'expected exception behavior'. It wasn't prepared for exceptions being rethrown, although it was activated in config.

**Comment by 304NotModified** [comment]
Thanks! I checked the code and have some questions:

> I had to adjust one test class, namely VariableLayoutRendererTests, because it did not fit into the 'expected exception behavior'. It wasn't prepared for exceptions being rethrown, although it was activated in config.

I don't fully understand this. Should we change the unit tests for this? 

- In which cases we don't like to log to the internal log? (`exception.MustBeRethrown(false)`)? I prefer to always log to the internal log? 
- In which case we like to call `MustRethrowSevere`? I think that one should be private? It's a bit unclear when we should use MustRethrowSevere and when MustRethrow. I think I would be nice if it was one method.

**Comment by tetrodoxin** [comment]
I used these versions of measures, just because I actually didn't dare to change the existing unit tests in their behavior. 
`MustRethrowSevere` is effectively the same as the former `MustRethrowSevere()`, it just checks for that list of serious exceptions, that will be rethrown always, regardless of `LogManager.ThrowExceptions`, since some classes seem to follow their own plans for exception handling with `LogManager.ThrowExceptions`. At these occurences I used `MustRethrowSevere()`. Do you think we should adjust the code there and log always as a rule?
I just didn't want to do that without asking, you know.

**Comment by 304NotModified** [comment]
> Do you think we should adjust the code there and log always as a rule?

Yes please :) Internal logging is never a problem, maybe in some test cases, but we need to change them.

**Comment by tetrodoxin** [comment]
Ok, so I'll go for it.
No more `MustRethrowSevere()` and no `MustBeRethrown(false)`

Also: I do the logging only for 'non severe' exceptions. Shall I change this behavior as well? I thought  after an `OutOfMemoryException` or `StackOverflowException` we should possibly not try to do more memory and resource consumpting stuff.

**Comment by 304NotModified** [comment]
Did you pushed everything?

**Comment by tetrodoxin** [comment]
I'm currently seeing, that `MustBeRethrown(false)` is solely used in `InternalLogger`. Is it still a good idea to log into internal logger when exceptions right in that internal logger occur?

**Comment by tetrodoxin** [comment]
Did not touch `MustBeRethrown(false)` yet, waiting for your thought about possible infinite recursion.

**Comment by 304NotModified** [comment]
> I'm currently seeing, that MustBeRethrown(false) is solely used in InternalLogger. Is it still a good idea to log into internal logger when exceptions right in that internal logger occur (exception loops) ?

We have to really prevent loops. So we have to create multiple functions IMO (private). I was thinking (don't mind the names):

- internallogger.Write (with catch, logs into internallogger.WriteInternal()
- internallogger.WriteInternal( empty catch)

**Comment by tetrodoxin** [comment]
An we would use `WriteInternal()` from within `MustBeRethrown()`, correct?

**Comment by tetrodoxin** [comment]
Another issue: `ignoreErrors` attribute when including configs definitely wants to suppress exceptions, that aren't severe. See `XmlLoggingConfigurationParseIncludeElement()`.

**Comment by 304NotModified** [comment]
> An we would use WriteInternal() from within MustBeRethrown(), correct?

Yes indeed.



> Another issue: ignoreErrors attribute when including configs definitely wants to suppress exceptions, that aren't severe. See XmlLoggingConfiguration.ParseIncludeElement().

I think we need an extra try-catch for that in XmlLoggingConfiguration.ParseIncludeElement().

**Comment by tetrodoxin** [comment]
> I think we need an extra try-catch for that in XmlLoggingConfiguration.ParseIncludeElement().

But then we again cannot distinguish, if we should instantly rethrow a serious exception, that should always be rethrown, or use `NLogConfigurationException`, because we catch any exception.

**Comment by tetrodoxin** [comment]
This is kind of a mashup of the recent versions, to solve our issues

**Comment by 304NotModified** [comment]
It's looks really sweet :)

File: src/NLog/Common/InternalLogger.cs [hunk 1]
```diff
@@ -181,7 +181,20 @@ public static bool IsFatalEnabled
...
+
+        /// <summary>
+        /// Logs the specified message at the specified level 
+        /// and catches any exceptions without rethrowing them.
+        /// </summary>
+        /// <param name="level">Log level.</param>
+        /// <param name="message">Message which may include positional parameters.</param>
+        /// <param name="args">Arguments to the message.</param>
+        [StringFormatMethod("message")]
+        internal static void LogSecure(LogLevel level, string message, params object[] args)
```
304NotModified: proposal: LogSwallowException.

src/NLog/Internal/ExceptionHelper.cs
```diff
@@ -63,17 +155,35 @@ public static bool MustBeRethrown(this Exception exception)
                 return true;
             }
 
-            if (exception is NLogConfigurationException)
+            // only log after the 'serious' exceptions above
+
+            var shallRethrow = false;
+
+            if (!shallRethrow)
```
304NotModified: proposal: remove unneeded `if` and combine asignment & declaration of "shallRethrow"

src/NLog/Common/InternalLogger.cs
```diff
@@ -311,10 +335,10 @@ public static void Fatal([Localizable(false)] string message, params object[] ar
         /// <param name="message">Log message.</param>
         public static void Fatal([Localizable(false)] string message)
         {
-            Write(LogLevel.Fatal, message, null);
+            Write(LogLevel.Fatal, message, true, null);
         }
 
-        private static void Write(LogLevel level, string message, object[] args)
+        private static void Write(LogLevel level, string message, bool respectExceptions, object[] args)
```
304NotModified: please remove "respectExceptions" (not used)

src/NLog/Internal/ExceptionHelper.cs
```diff
@@ -63,17 +155,35 @@ public static bool MustBeRethrown(this Exception exception)
...
+                level = shallRethrow ? LogLevel.Error : LogLevel.Warn;
+
             }
 
-            return false;
+            if (level != LogLevel.Off)
+            {
+                if (string.IsNullOrEmpty(logMessage))
+                {
+                    InternalLogger.LogSecure(level, exception.Message);
```
304NotModified: I think it's nice to have whole exception details in the log, so exception.toString()?

src/NLog/Conditions/ConditionParser.cs
```diff
@@ -136,7 +136,7 @@ private ConditionMethodExpression ParsePredicate(string functionName)
             }
             catch (Exception exception)
             {
-                if (exception.MustBeRethrown())
+                if (exception.MustBeRethrown("While resolving function '{0}' follwing exception occured: {1}", functionName, exception))
```
304NotModified: proposal: **always** add the exception to the end of the message. So no need to to pass exception twice (implicit and explicit). Need a lot of changes.

src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -319,28 +321,18 @@ private void Initialize(XmlReader reader, string fileName, bool ignoreErrors)
...
-                        throw ConfigException;
-                    }
-                    else
-                    {
-                        InternalLogger.Error("Error in Parsing Configuration File. Exception : {0}", ConfigException);
-                    }
-                }
-                else
+                // we already logged the exception, so merely rethrow if necessary, without logging
+                if (exception.MustBeRethrown(ignoreErrors, LogLevel.Off, null))
```
304NotModified: is this not "logsafe"?

src/NLog/LogManager.cs
```diff
@@ -362,12 +362,10 @@ private static void SetupTerminationEvents()
             }
             catch (Exception exception)
             {
-                if (exception.MustBeRethrown())
+                if (exception.MustBeRethrown("Error setting up termination events: {0}", exception))
```
304NotModified: loglevel should be warn?

src/NLog/LoggerImpl.cs
```diff
@@ -230,12 +230,11 @@ private static FilterResult GetFilterResult(IList<Filter> filterChain, LogEventI
             }
             catch (Exception exception)
             {
-                if (exception.MustBeRethrown())
+                if (exception.MustBeRethrown("Exception during filter evaluation: {0}", exception))
```
304NotModified: loglevel should be warn?

src/NLog/Config/LoggingConfiguration.cs
```diff
@@ -311,12 +311,10 @@ internal void Close()
                 }
                 catch (Exception exception)
                 {
-                    if (exception.MustBeRethrown())
+                    if (exception.MustBeRethrown("Exception while closing {0}", exception))
```
304NotModified: loglevel should be warn?

tests/NLog.UnitTests/LayoutRenderers/VariableLayoutRendererTests.cs
```diff
@@ -247,6 +247,7 @@ private void CreateConfigFromXml()
                     <logger name='*' minlevel='Debug' writeTo='debug' />
                 </rules>
             </nlog>");
+            LogManager.ThrowExceptions = false;     // since none of these test are prepared for exceptions
```
304NotModified: I will check this change

304NotModified:
Hey Andreas,

I reviewed your code. I think this is really great improvement for NLog. Good work!

Some last things:

- Some messages are changed from warning to error. Is this a deliberated choice - not sure if we discussed that before. (I didn't marked all of them)
- I think that it would be nice to always add the exception to the message. (see code comment)
- Other small things, see code comments

I will check the issue with the unit test ATM

tests/NLog.UnitTests/LayoutRenderers/VariableLayoutRendererTests.cs
304NotModified: This needs to be reverted. The issue is somewhere else (post it in the PR)

304NotModified:
I see the problem with the "LogManager.ThrowExceptions"

The internallogger should only throw exceptions severe exceptions. (internallogger line 405).

Proposal:

- revert change in unit test
- create a IsServereException on exceptionhelpers (internal)
- use that one in the internallogger

Also "Var_default_after_set_null" fails, that's a bug in the variabeleRenderers.

- [ ] fix Var_default_after_set_null

tetrodoxin:
>proposal: LogSwallowException.

N.P. Changed that.

>proposal: remove unneeded if and combine asignment & declaration of "shallRethrow" 
`src/NLog/Internal/ExceptionHelper.cs (160)`

Oh, I just missed that one. Was a relict of several changes in the code/tries around it.

>`src/NLog/Common/InternalLogger.cs (341)`
please remove "respectExceptions" (not used)

Actually, the mistake was, that `respectExceptions` wasn't used. I've corrected that.

tetrodoxin:
>`src/NLog/Internal/ExceptionHelper.cs`
I think it's nice to have whole exception details in the log, so exception.toString()?

>proposal: always add the exception to the end of the message. So no need to to pass exception twice (implicit and explicit). Need a lot of changes.

Well, I should have hit on it myself. I changed `MustBeRethrown()` that way, it always appends `exception.ToString()` and will fold that behavior into the respective method calls in code by refactoring string formatting.

tetrodoxin:
>loglevel should be warn? (several places)

Yes and no. `MustBeRethrown()` uses warn level, if the exception is not to be rethrown, otherwise error level. I read the occurences you pointed out so, that they would produce the same behavior, meaning a warning would have been written, if the rethrow hasn't happened.Since no tests were broken, I considered that behavior plausible.

tetrodoxin:
>`src/NLog/Config/XmlLoggingConfiguration.cs (331)`
is this the same as calling "logsafe"?

I guess you mean `LogSwallowExceptions()`?
Actually no. Granted, these lines are sort of a bit hacky, but to follow the expected (and test covered) behavior I had to use this combination of calls to `MustBeRethrown()`. The first one is nearly an equivalent to the former `MustRethrowSevere()` as it only throws that serious exceptions and does the logging to error level.
The second call is just for encapsulate the check to `LogManager.ThrowExceptions` to throw a custom exception in that case.

tetrodoxin:
I must admit, I have no idea why `Var_default_after_set_null` ever passed without setting `ThowExceptions` to false, because an `NullReferenceException` is obvious with the current code in `VariableLayoutRenderer.Append()`

tetrodoxin:
Ok, I can see you changed right this code :-)

304NotModified:
let me know when I can review it again :) Thanks!

304NotModified:
> Yes and no. MustBeRethrown() uses warn level, if the exception is not to be rethrown, otherwise error level.

I was thinking about this. Is Error level not the best way for exceptions? (for all)?

304NotModified:
I think #1040 broke this PR :(

304NotModified:
If you added as me collaborator, I can help fixing the last stuff. Just let me know.

304NotModified:
bit busy on other stuff, will look at this next week

304NotModified:
superseded by #1117

304NotModified:
@tetrodoxin I like to have this in 4.3, so I'm busy fixing the conflict and some loose ends.

Thanks for the PR!

tetrodoxin:
I am really sorry for being absent that last couple of months and I hope, you at least could make some use of the code snippets, we worked out last year.
Cheers!

304NotModified:
Yes, we did :) This feature has been merged in 4.3 

Thanks for the contributions!

Issue #444, opened 2014-11-18
Text cut off when using When layout renderer combined with a colon


tino1986:
In a File target, for example, I'm using a **When** layout renderer like this (simplified example):

    layout="${when:when=1 == 1:inner=Test: Hello}"

I would expect the result to be **Test: Hello**. However, only **Hello** is logged, without **Test:**.

I've already figured out that it has something to do with the colon, and I've tried solving the issue using several methods:

 - Escaping it with \
 - Wrapping it in a Literal layout renderer
 - Replacing it with its ASCII code &#58;
 - Putting it in a separate variable

But none of them seem to work.

ie-zero:
Wiki page (https://github.com/NLog/NLog/wiki/When-Layout-Renderer) updated with the workaround until this issue is resolved.

304NotModified:
Changing the too short before a release isn't a good idea, so moving to 4.1

304NotModified:
Also unit test #914 with this case.

PR #925, opened 2015-09-23
RegistryLayoutRenderer: Support for layouts, RegistryView (32, 64 bit) and all root key names (HKCU/HKLM etc)

Niklas-Peter:
- Layouts for Key, Value, DefaultValue
- Added View
- Root key names like "HKLM" are now allowed

Please check the code, especially your conventions for exception handling. ParseKey() and this.DefaultValue.Render() could throw an exception. I do not know, how you would like to handle it.

Obviously Microsoft.Win32 is a problem. Maybe you can add appropriate pre compiler conditions to only activate it on supporting platforms.


304NotModified:
What do you mean with

>Obviously Microsoft.Win32 is a problem

I'm on mobile now, so I cannot easily see the code.  But did you excluded the code for Silverlight? (see appveyor error)

Niklas-Peter:
Silverlight has already been excluded in the old version and still is. So it should not be the problem. However the ```enum RegistryView``` is not supported in .NET 3.5, so I think, that it is the problem.
Furthermore I had accidentally removed ```using NLog.Internal```, so that the extension metod for ```Exception``` was missing. I fixed it.

However some unit tests fail, but I am not quite sure, whether the unit tests themselves are correct. Why do you use ```@``` and additionally escape with ```\\``` (see ```RegistryNamedValueTest``` and ```RegistryUnnamedValueTest```)? Why do you expect, that  ```defaultValue=C\\:temp``` results in ```C:temp``` (see ```RegistyDefaultValueTest_with_colon```)?

304NotModified:
> Why do you use @ and additionally escape with \\

I think because `Layout` is interpreted as  `string` and not an verbatim string. 

> Why do you expect, that defaultValue=C\\:temp results in C:temp

We need to escape the `:` (because separator) and also we need to escape the `\` (because of `\n` etc) in C#, so `\\:`.

src/NLog/LayoutRenderers/RegistryLayoutRenderer.cs
```diff
@@ -74,79 +85,122 @@ public class RegistryLayoutRenderer : LayoutRenderer
...
+                            {
+                                registryValue = registryKey.GetValue(this.Value.Render(logEvent));
+                            }
+                        }
+                        else
+                        {
+                            registryValue = rootKey.GetValue(this.Value.Render(logEvent));
+                        }
 
-                    this.subKey = this.key.Substring(pos + 1).Replace('/', '\\');
```
304NotModified: I think this removal makes it non-backwards-compatible

Niklas-Peter: Why? I just moved it in ParseKey().

Niklas-Peter:
The reason, why the 3 tests fail, is, that I changed "string" to Layout for Key, Value and DefaultValue. Obviously ```\\``` is treated different, when it is a Layout. I do not know enough about the inner workings of NLog, so I can not fix it. Either you know what to do or we have to change it back to a string, but actually the layouts were the main reason for creating this pull request.

304NotModified:
We can fix it :) 

But not now. On holiday with limited Internet / pc. I will look at it this weekend.

304NotModified:
I'm looking at this now :)

304NotModified:
Looked at this:

- note: `registryValue ` can be `null` in registryValue = registryKey.GetValue(renderedValue);` 
- code for handling the Exception should be
```
       catch (Exception ex)
            {
        
                if (ex.MustBeRethrown())
                {
                    throw;
                }
               InternalLogger.Error(ex.ToString());
            }
```

- I changed the parsekey for the slashes to:
```c#
            if (pos >= 0)
            {
                hiveName = key.Substring(0, pos);

                var subKey1 = key.Substring(pos + 1).Replace('/', '\\');
              
                //remove duplicates
                subKey1 = subKey1.Replace(@"\\", @"\");
                subKey1 = subKey1.TrimStart('\\');
                this.subKey = subKey1;
            }
```

- RegistyDefaultValueTest_with_colon is still failing. I have to think how to fix this. This has to do with https://github.com/NLog/NLog/issues/444 I guess...

304NotModified:
I will add some tests to the RegistryTests in master, because i'm curious how `key` was working with forwardsslashes (and that we don't break it)

Niklas-Peter:
> note: registryValue can be null in registryValue = registryKey.GetValue(renderedValue); 

From MS documentation:
> Retrieves the value associated with the specified name. Returns null if the name/value pair does not exist in the registry.

I do not see the problem.

304NotModified:
> Retrieves the value associated with the specified name. Returns null if the name/value pair does not exist in the registry

It's a false positive of Resharper :) Can you add this a comment in the code?

Niklas-Peter:
Actually I had already done so, see line 142.

304NotModified:
I see the confusion. It about openSubKey.

https://msdn.microsoft.com/en-us/library/z9f66s0a(v=vs.110).aspx

> The subkey requested, or null if the operation failed.

So Reshaper was right :)

304NotModified:
I added some more tests in #937. Can you merge them in your PR?

Git commands:

```
git remote add upstream https://github.com/NLog/NLog.git
git fetch upstream
git checkout patch-1
git merge upstream/master
```

304NotModified:
TODO:

- [ ] fix possible nullref
- [ ] merge with master
- [ ] fix slash issue (@me)

304NotModified:
Hi @Niklas-Peter ,

The slash isssue has been resolved, see https://github.com/NLog/NLog/commit/118b8a5ae767afd06b7582365142d31d7fcaaf28

The parse of the inner layout fixed is fixed in #975 

Can you merge with master and fix the last issues?

Niklas-Peter:
I will do that anywhen, but currently I do not even have time to fix these small issues ...

304NotModified:
No problem. Any idea when you have time?

The release of 4.2 is nearby, but we can postpone this feature to 4.3

304NotModified:
If you like help this, just add me as a contributor and let me know. I like to have the feature implemented :)

304NotModified:
We need unit test for this.

Niklas-Peter:
I added you as a collaborator to Niklas-Peter/NLog. Is that, what you intended?

304NotModified:
Yes, collaborator I ment :)

304NotModified:
I added and option `RequireEscapingSlashesInDefaultValue` for backwards-compatibility. 

testcases:
      
```c#     

AssertLayoutRendererOutput("${registry:value=NOT_EXISTENT:key=HKLM/NOT_EXISTENT:defaultValue=C\\\\temp}", "C\\temp");
AssertLayoutRendererOutput("${registry:value=NOT_EXISTENT:key=HKLM/NOT_EXISTENT:defaultValue=C\\temp:requireEscapingSlashesInDefaultValue=false}","C\\temp");
```

304NotModified:
let's wait for codecov if we need more unit tests

304NotModified:
Dunno why it doens't work in Mono

```
NLog.UnitTests.LayoutRenderers.RegistryTests.RegistryNamedValueTest [FAIL]
   Assert.Equal() Failure
   Position: First difference is at position 0
   Expected: FooValue
   Actual:
   Stack Trace:
     at NLog.UnitTests.NLogTestBase.AssertDebugLastMessage (System.String targetName, System.String msg) [0x00000] in <filename unknown>:0 
     at NLog.UnitTests.LayoutRenderers.RegistryTests.RegistryNamedValueTest () [0x00000] in <filename unknown>:0 
     at (wrapper managed-to-native) System.Reflection.MonoMethod:InternalInvoke (System.Reflection.MonoMethod,object,object[],System.Exception&)
     at System.Reflection.MonoMethod.Invoke (System.Object obj, BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x00000] in <filename unknown>:0
NLog.UnitTests.LayoutRenderers.RegistryTests.RegistryUnnamedValueTest [FAIL]
   Assert.Equal() Failure
   Position: First difference is at position 0
   Expected: UnnamedValue
   Actual:
   Stack Trace:
     at NLog.UnitTests.NLogTestBase.AssertDebugLastMessage (System.String targetName, System.String msg) [0x00000] in <filename unknown>:0 
     at NLog.UnitTests.LayoutRenderers.RegistryTests.RegistryUnnamedValueTest () [0x00000] in <filename unknown>:0 
     at (wrapper managed-to-native) System.Reflection.MonoMethod:InternalInvoke (System.Reflection.MonoMethod,object,object[],System.Exception&)
     at System.Reflection.MonoMethod.Invoke (System.Object obj, BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x00000] in <filename unknown>:0
```

304NotModified:
Todo: 

Test for no subkey
Test for null defaultvalue 
Test for invalid hive

No exception when invalid config (only when needed)

304NotModified:
Maybe:

Fix backwards compatibility another way (rename defaultvalue,  keep both, old on as string and obsolete)

304NotModified:
RegistryView is never used?

src/NLog/LayoutRenderers/RegistryLayoutRenderer.cs
```diff
@@ -126,13 +117,33 @@ public string Key
...
+            Object registryValue = null;
+            // Value = null is necessary for querying "unnamed values"
+            string renderedValue = this.Value != null ? this.Value.Render(logEvent) : null;
 
+            var parseResult = ParseKey(this.Key.Render(logEvent));
             try
             {
-                using (RegistryKey registryKey = this.rootKey.OpenSubKey(this.subKey))
+#if !NET3_5
+                using (RegistryKey rootKey = RegistryKey.OpenBaseKey(parseResult.Hive, View))
```
Niklas-Peter: > RegistryView is never used?
Here it is used.

(I assume, that the other messages are just for yourself?)

304NotModified: Thanks! Searching on mobile wasn't working well ;)

304NotModified:
@Niklas-Peter 

it's almost done,but these things should be done:

Todo: 

- [ ] Test for no subkey
- [ ] Test for null defaultvalue 
- [ ] Test for invalid hive
- [ ] Test for view
- [ ] No exception when invalid config (only when needed) 

@Niklas-Peter: can you write some tests? At least for the 'view' ?

Maybe:

@Niklas-Peter 
I don't really like my own solution with `RequireEscapingSlashesInDefaultValue`. Another way is to create a new property and keep the old (string) one. What do you think? Which name should be suitable ?

- [ ] Fix backwards compatibility another way

Niklas-Peter:
> can you write some tests? At least for the 'view' ?
Yes, but probably not until christmas ...

 >  Another way is to create a new property and keep the old (string) 
one. What do you think? Which name should be suitable ?
````RequireEscapingSlashesInDefaultValue```` is really not that good, 
because it addresses just an implementation detail. The actual change 
is, that I switched from a string to a Layout and the problem with the 
strings is just an unwanted side effect.
Maybe sth. like ````LayoutableDefaultValue````, but I am really bad at 
finding names.

**Comment by 304NotModified** [comment]
Can you explain me in short where the "RegistryView" is useful for? As I never used it.

**Comment by Niklas-Peter** [comment]
I think the MS documentation explains it pretty well:

 > On the 64-bit version of Windows, portions of the registry are stored 
separately for 32-bit and 64-bit applications. There is a 32-bit view 
for 32-bit applications and a 64-bit view for 64-bit applications.
(https://msdn.microsoft.com/en-us/library/microsoft.win32.registryview%28v=vs.110%29.aspx)

My specific use case was a 32 bit COM application, that accessed a 64 
bit .NET library. With default settings a registry key set by a 64 bit 
application can not be read by a 32 bit app. So you must make sure to 
always use the same registry view by explicitly setting it.

**Comment by 304NotModified** [comment]
I will continue on this one when https://github.com/NLog/NLog/pull/915 has been merged. 

The code is almost ready.

**Comment by 304NotModified** [comment]
Full change:

- Support for all hives:
  - HKEY_LOCAL_MACHINE            
  - HKLM                          
  - HKEY_CURRENT_USER             
  - HKCU                          
  - HKEY_CLASSES_ROOT             
  - HKEY_USERS                    
  - HKEY_CURRENT_CONFIG           
  - HKEY_DYN_DATA                 
  - HKEY_PERFORMANCE_DATA

**Comment by 304NotModified** [comment]
@bhaeussermann / @Page-Not-Found / @UgurAldanmaz can someone review this one?

I like to include this in NLog 4.3 

Thanks!

File: tests/NLog.UnitTests/LayoutRenderers/RegistryTests.cs [hunk 3]
```diff
@@ -70,6 +103,40 @@ public void RegistryNamedValueTest()
...
+            LogManager.Configuration = CreateConfigurationFromString(@"
+            <nlog>
+                <targets><target name='debug' type='Debug' layout='${registry:key=HKCU\\Software\\NLogTest:value=view32:view=Registry32}' /></targets>
+                <rules>
+                    <logger name='*' minlevel='Debug' writeTo='debug' />
+                </rules>
+            </nlog>");
+
+            LogManager.GetLogger("d").Debug("zzz");
+            AssertDebugLastMessage("debug", "reg32");
```
bhaeussermann: Can these tests not use `AssertLayoutRendererOutput()` like the other tests?

tests/NLog.UnitTests/LayoutRenderers/RegistryTests.cs
```diff
@@ -196,6 +263,67 @@ public void RegistyDefaultValueTest_with_foward_slash()
             AssertLayoutRendererOutput("${registry:value=NOT_EXISTENT:key=HKLM/NOT_EXISTENT:defaultValue=C\\\\temp}",
                 "C\\temp");
         }
+
+
+        [Fact]
+        public void RegistyDefaultValueTest_with_foward_slash2()
```
bhaeussermann: Many test method names (including some old ones) start with `Registy` instead of `Registry`.

bhaeussermann:
I checked the code. I cannot find any problems. Just two minor remarks on the tests.

tests/NLog.UnitTests/LayoutRenderers/RegistryTests.cs
bhaeussermann: I see there are old tests which also use this approach; can be changed to use `AssertLayoutRendererOutput()`?

304NotModified:
Thanks @bhaeussermann!

tests/NLog.UnitTests/LayoutRenderers/RegistryTests.cs
304NotModified: Good idea :)

tests/NLog.UnitTests/LayoutRenderers/RegistryTests.cs
304NotModified: Will replace-all them

tests/NLog.UnitTests/LayoutRenderers/RegistryTests.cs
304NotModified: I found a difference. `AssertLayoutRendererOutput` only renders the layout renderer, while these test cases really writes to the registry. So keep it like this

304NotModified:
Merge after successful builds

304NotModified:
docs: https://github.com/NLog/NLog/wiki/Registry-Layout-Renderer

PR #926, opened 2015-09-24
Assure automatic re-configuration after configuration change

Niklas-Peter:
Currently it is just an implementation detail, but if you add this note to the specification, there is no need to call unnecessarily ReconfigExistingLoggers() after updating the configuration.


304NotModified:
Thanks!

Issue #612, opened 2015-02-27
Cached Layout Renderer is reevaluated when LoggingConfiguration is changed


Quergo:
Hi,

I use Cached Layout Renderer to generate a separate log file each time my application starts. This is done by caching the date of the first logging call. 

The target definition i use is:
```xml
<target name="default_logfile" xsi:type="File" fileName="${basedir}\${date:cached=True:format=yyyy-MM-dd_HH-mm-ss}_Log.txt" />
```

This works fine until i start to add new logging targets at runtime. I make changes to the LoggingConfiguration like this:

```C#
LoggingConfiguration config = LogManager.Configuration;

var logFile = new FileTarget();
config.AddTarget(dynamicTargetName, logFile);

logFile.FileName = dynamicFilePath + "_Ext.log";
logFile.Layout = @"|${level:uppercase=true} @ ${logger:shortName=true}|:${date:format=yyyy-MM-dd HH\:mm\:ss} ${message} ${exception:stacktrace:separator=*}";

var rule = new LoggingRule("*", LogLevel.Trace, logFile);
config.LoggingRules.Add(rule);

LogManager.Configuration = config;
```

On the one hand it correctly generates my dynamic logging target. But on the other, this seems to reevaluate the ${date:cached=True:format=yyyy-MM-dd_HH-mm-ss} expression which in consequence will generate a complete new logfile for my "default_logfile" target. The date used for the new logfile is the one of the first logging call after LoggingConfiguration reassignment.

I also tested this:
```C#
LogManager.Configuration = LogManager.Configuration
```
which also triggers the reevaluation of cached layout render.

I don't know if this is a bug or I am just doing something wrong with the dynamic logfile generation.

304NotModified:
I don't think this is bug

Quergo:
In the class CachedLayoutRendererWrapper.cs if found these two methods getting called if LoggingConfiguration is reassigned:
```C#
        /// <summary>
        /// Initializes the layout renderer.
        /// </summary>
        protected override void InitializeLayoutRenderer()
        {
            base.InitializeLayoutRenderer();
            this.cachedValue = null;
        }

        /// <summary>
        /// Closes the layout renderer.
        /// </summary>
        protected override void CloseLayoutRenderer()
        {
            base.CloseLayoutRenderer();
            this.cachedValue = null;
        }
```
Both times the cached value is cleared and will be reevaluated when the wrapped content is rendered again.

[...]

PR #933, opened 2015-09-28
Issue #612: Cached Layout Renderer is reevaluated when LoggingConfiguration is changed

bhaeussermann:
Fixes Issue #612

### DIFF
--- a/src/NLog/LayoutRenderers/Wrappers/CachedLayoutRendererWrapper.cs
+++ b/src/NLog/LayoutRenderers/Wrappers/CachedLayoutRendererWrapper.cs
@@ -47,7 +47,18 @@ namespace NLog.LayoutRenderers.Wrappers
     [ThreadAgnostic]
     public sealed class CachedLayoutRendererWrapper : WrapperLayoutRendererBase
     {
-        private string cachedValue;
+        /// <summary>
+        /// A value indicating when the cache is cleared.
+        /// </summary>
+        public enum ClearCacheOption 
+        { 
+            /// <summary>Never clear the cache.</summary>
+            Never = 0,
+            /// <summary>Clear the cache whenever the <see cref="CachedLayoutRendererWrapper"/> is initialized.</summary>
+            OnInitialize = 1
+        }
+
+        private string cachedValue = null;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="CachedLayoutRendererWrapper"/> class.
@@ -55,6 +66,7 @@ namespace NLog.LayoutRenderers.Wrappers
         public CachedLayoutRendererWrapper()
         {
             this.Cached = true;
+            this.ClearCache = ClearCacheOption.OnInitialize;
         }
 
         /// <summary>
@@ -65,12 +77,18 @@ namespace NLog.LayoutRenderers.Wrappers
         public bool Cached { get; set; }
 
         /// <summary>
+        /// Gets or sets a value indicating when the cache is cleared.
+        /// </summary>
+        public ClearCacheOption ClearCache { get; set; }
+
+        /// <summary>
         /// Initializes the layout renderer.
         /// </summary>
         protected override void InitializeLayoutRenderer()
         {
             base.InitializeLayoutRenderer();
-            this.cachedValue = null;
+            if (ClearCache == ClearCacheOption.OnInitialize)
+                this.cachedValue = null;
         }
 
         /// <summary>
@@ -79,7 +97,6 @@ namespace NLog.LayoutRenderers.Wrappers
         protected override void CloseLayoutRenderer()
         {
             base.CloseLayoutRenderer();
-            this.cachedValue = null;
         }
 
         /// <summary>
--- a/src/NLog/LayoutRenderers/Wrappers/CachedLayoutRendererWrapper.cs
+++ b/src/NLog/LayoutRenderers/Wrappers/CachedLayoutRendererWrapper.cs
@@ -33,8 +33,9 @@
 
 namespace NLog.LayoutRenderers.Wrappers
 {
-    using System.ComponentModel;
     using NLog.Config;
+    using System;
+    using System.ComponentModel;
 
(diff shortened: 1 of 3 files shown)


bhaeussermann:
How will NLog.xsd get updated? Is there a build-process that will do this automatically?

304NotModified:
> How will NLog.xsd get updated? Is there a build-process that will do this automatically?

semi-automatic  - it's a msbuild script. But we do have to remember to run and upload it ;)

tests/NLog.UnitTests/SourceCodeTests.cs
```diff
@@ -183,10 +183,10 @@ private int CompareDirectoryWithProjects(List<string> filesToCompile, params str
             var missingFiles = filesToCompile.Except(filesInProject).ToList();
             if (missingFiles.Count > 0)
             {
-                Console.WriteLine("The following files must be added to {0}", string.Join(";", projectFiles));
+                System.Diagnostics.Debugger.Log(0, "Tests", string.Format("The following files must be added to {0}{1}", string.Join(";", projectFiles), Environment.NewLine));
```
304NotModified: thanks for this improvements :)

304NotModified:
We don;t need to change `CloseLayoutRenderer` as suggested in #612?

src/NLog/LayoutRenderers/Wrappers/CachedLayoutRendererWrapper.cs
```diff
@@ -47,7 +47,18 @@ namespace NLog.LayoutRenderers.Wrappers
     [ThreadAgnostic]
     public sealed class CachedLayoutRendererWrapper : WrapperLayoutRendererBase
     {
-        private string cachedValue;
+        /// <summary>
+        /// A value indicating when the cache is cleared.
+        /// </summary>
+        public enum ClearCacheOption
```
304NotModified: Look good, but I think a `[flags]` enum would be more future proof? Although i'm not sure if a flags-enum is working from the XML config. 

What do you think?

bhaeussermann:
> We don;t need to change CloseLayoutRenderer as suggested in #612?

We do. The change removes the statement `this.cachedValue = null;` from CloseLayoutRenderer().

src/NLog/LayoutRenderers/Wrappers/CachedLayoutRendererWrapper.cs
bhaeussermann: I agree with making this a `[Flags]` enum. I will change this and check to see if flags-enums work from the XML config.

304NotModified:
> We do. The change removes the statement this.cachedValue = null; from CloseLayoutRenderer().

Ow yes, missed that.

I see that the `[flags]` enum should work from the XML config, see [this unit test](https://github.com/NLog/NLog/blob/b518e925b40dc0e2e8bd5f97493831cb02a0c1bd/tests/NLog.UnitTests/Config/TargetConfigurationTests.cs#L422-L472)

src/NLog/LayoutRenderers/Wrappers/CachedLayoutRendererWrapper.cs
```diff
@@ -47,7 +48,21 @@ namespace NLog.LayoutRenderers.Wrappers
...
     {
-        private string cachedValue;
+        /// <summary>
+        /// A value indicating when the cache is cleared.
+        /// </summary>
+        [Flags]
+        public enum ClearCacheOption 
+        { 
+            /// <summary>Never clear the cache.</summary>
+            None = 0,
```
304NotModified: I think 'None' is a bit tricky, because what will this do? ` var option = ClearCacheOption.None | ClearCacheOption .OnClose` ?

bhaeussermann: This value was renamed from 'Never'. What do you propose?

bhaeussermann: What about "Off"?

304NotModified:
After thinking about this, I think it's OK now.  I will do a final check in the upcoming days and then merge it. 

You didn't like the onInit as name? In the ASP.NET Page Life its also called "init"

bhaeussermann:
> You didn't like the onInit as name? In the ASP.NET Page Life its also called "init" 

Well, I don't like abbreviations in code, and the method that the value applies to is called `InitializeLayoutRenderer`, so `OnInitialize` seems more consistent with that.

304NotModified:
It's also the value in the xml.  For code you have autocompletion, but for xml not.

304NotModified:
OK :)

It looks good! It's done for merging? 

I only need an (XML) example for the wiki/news post. Can you post one?

bhaeussermann:
It is done for merging.

bhaeussermann:
I'm looking at the Wiki...

304NotModified:
I has been merged! Thanks :)

bhaeussermann:
I updated the [Wiki page](https://github.com/NLog/NLog/wiki/Cached-Layout-Renderer).

bhaeussermann:
Is there anything else I must do?

304NotModified:
Thanks! I added an 4.2 note and the possible options. Can you add also an example to the wiki? That would be great! 

And that would be the last thing :)

dnlgmzddr:
@304NotModified I was reviewing the issue, and we will need to add a new setting to specified the "PickupDirectoryLocation" override, but also this setting open a flow for the creation of the SMTP client.

With a pickup directory, we no longer require a server and the presence of both we raise a new exception. Should we use the same logic used by [SmptClient](http://referencesource.microsoft.com/#System/net/System/Net/mail/SmtpClient.cs,ebaa2d61cfed8051) which is: 
* if there's a server use it if not look for the pickup directory and if is not present throw an exception.

bhaeussermann:
I added the example.

Issue #913, opened 2015-09-15
Log NLog version to internal log. 


304NotModified:
Log Assembly version and file version to internal log (info) . also of loaded assemblies (NLog.web etc)

bhaeussermann:
What are these loaded assemblies and NLog.web?

304NotModified:
Those assemblies are installed by the user (eg nuget package) 

When we do the auto load, we like to log the versions.

304NotModified:
No auto loading is done here: https://github.com/NLog/NLog/blob/eebb37d20904dbf70a919cd6fbc4cdcae7f488d7/src/NLog/Config/ConfigurationItemFactory.cs

PR #938, opened 2015-10-04
Issue #913: Log NLog version to internal log.

bhaeussermann:


src/NLog/Common/InternalLoggerHelpers.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// </summary>
+    public static class InternalLoggerHelpers
+    {
+        /// <summary>
+        /// Logs the assembly version and file version of the given Assembly.
+        /// </summary>
+        /// <param name="assembly">The assembly to log.</param>
+        public static void LogAssemblyVersion(Assembly assembly)
+        {
+            InternalLogger.Info("{0} - Assembly version: {1}; File version: {2}.",
```
304NotModified: 1. can you add a try-catch around this? To be sure. 
2. we also like the AssemblyInformationalVersion  (System.Diagnostics.FileVersionInfo.ProductVersion) 

thanks!

src/NLog/Config/ConfigurationItemFactory.cs
```diff
@@ -104,7 +98,16 @@ public ConfigurationItemFactory(params Assembly[] assemblies)
         /// <summary>
         /// Gets or sets default singleton instance of <see cref="ConfigurationItemFactory"/>.
         /// </summary>
-        public static ConfigurationItemFactory Default { get; set; }
+        public static ConfigurationItemFactory Default
+        {
+            get
+            {
+                if (defaultInstance == null)
```
304NotModified: why this change?

src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -65,10 +65,14 @@ public class XmlLoggingConfiguration : LoggingConfiguration
     {
         #region private fields
 
-        private readonly ConfigurationItemFactory configurationItemFactory = ConfigurationItemFactory.Default;
         private readonly Dictionary<string, bool> visitedFile = new Dictionary<string, bool>(StringComparer.OrdinalIgnoreCase);
 
         private string originalFileName;
+        
+        private ConfigurationItemFactory ConfigurationItemFactory
```
304NotModified: why this change?

src/NLog/Common/InternalLoggerHelpers.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.Reflection;
+
+namespace NLog.Common
+{
+    /// <summary>
+    /// Helper methods for internal logger.
+    /// </summary>
+    public static class InternalLoggerHelpers
```
304NotModified: I'm curious why this isn't added to the `InternalLogger` class?

304NotModified:
Thanks for this PR :+1: 

I reviewed your code. I looks good, but I saw some unexpected changes. See the code notes :)

304NotModified:
I think `FileVersionInfo` isn't in Silverlight (see AppVeyor message). 

Please add for silverlight the log, but not the fileversion. You can use `#if SILVERLIGHT` , `#else`, and `#endif`

See https://github.com/NLog/NLog/blob/master/CONTRIBUTING.md#multiple-net-versions

src/NLog/Config/ConfigurationItemFactory.cs
bhaeussermann: The first commit of the pull request ("Fixed Auto-loading logs did not render because extension DLL's were loaded before InternalLogger config was read") contains all the changes that were needed to fix the auto-loading logs that did not appear.

This part, in particular, implements lazy instantiation on the ConfigurationItemFactory, because it was created too early. Now the ConfigurationItemFactory is created only after the internal logger had been configured, so that the auto-loading logs actually appear in the internal log file.

src/NLog/Config/XmlLoggingConfiguration.cs
bhaeussermann: This is connected to the other change you queried. It facilitates the lazy instantiation of the `ConfigurationItemFactory`.

src/NLog/Common/InternalLoggerHelpers.cs
bhaeussermann: The logic seemed too specific to add to `InternalLogger`.  
Would you think this to fit better in `InternalLogger`?

304NotModified: I prefer combining with InternalLogger with these reasons:

- It's highly correlated to the InternalLogger (25%)
- The method is not used somewhere else
- The method can be created `private` after combine, which has preference (= less maintenance)
- The InternalLogger has now only one method.

src/NLog/Config/ConfigurationItemFactory.cs
304NotModified: OK thanks. Can you add this (summary is OK), to the `<remarks>` ?

bhaeussermann:
We now write `assembly.FullName` to the log, which includes the assembly version. The log now looks like this:
`2015-10-05 20:24:08.3910 Info NLog.Web, Version=2.0.0.0, Culture=neutral, PublicKeyToken=null. File version: 2.0.14. Product version: 2.0.14.`

304NotModified:
Good idea!

bhaeussermann:
The pull request is ready to my knowledge. Do you notice anything that's outstanding?

304NotModified:
I looks good. I will check the PR in detal this week

304NotModified:
Thanks for this nice PR! :+1:

Issue #690, opened 2015-05-04
useSystemNetMailSettings=false still uses .config (partially) 


yang7749:
MailTarget's useSystemNetMailSettings parameter doesn't seem to work.

NLog.config:
```
<target xsi:type="Mail" name="Email" from="test@test.com" to="test@test.com" useSystemNetMailSettings="false"
            smtpServer="localhost" smtpPort="25" timeout="30000"
            subject="test" 
            body="test" />
```

web.config:
```
<system.net>
        <mailSettings>
            <smtp deliveryMethod="SpecifiedPickupDirectory">
                <specifiedPickupDirectory pickupDirectoryLocation="C:\TEMP\email_out" />
            </smtp>
        </mailSettings>
    </system.net>
```

The logger still add files to "C:\TEMP\email_out" whenever it tries to send email

304NotModified:
Please post the internal log file. Thanks!

https://github.com/NLog/NLog/blob/master/CONTRIBUTING.md#bug-reports

yang7749:
There aren't any message in the internal log related to it. Here is all I see in the internal log after my code logs an error that should have send the email:

```
2015-05-04 16:22:01.3895 Info Adding target Mail Target[Email]
2015-05-04 16:22:01.4085 Info Configured from an XML element in C:\SVN\TestWebsite\bin\NLog.config...
2015-05-04 16:22:01.4085 Info Watching path 'C:\SVN\TestWebsite\bin' filter 'NLog.config' for changes.
2015-05-04 16:22:01.4085 Info Watching path 'c:\' filter 'NLog.config' for changes.
2015-05-04 16:22:01.4245 Info Found 12 configuration items
```

304NotModified:
Also not in `debug` level?

304NotModified:
Or even `trace`.  It would be odd if it doesn't work because we use the default .net class for this. https://msdn.microsoft.com/en-us/library/k1c4h6e2(v=vs.110).aspx

yang7749:
Here is my internal log with trace:
```
2015-05-06 10:42:48.3925 Trace FindReachableObject<NLog.Internal.IRenderable>:
2015-05-06 10:42:48.3985 Trace Scanning LongDateLayoutRenderer 'Layout Renderer: ${longdate}'
2015-05-06 10:42:48.3985 Debug Setting 'UppercaseLayoutRendererWrapper.uppercase' to 'true'
[...]

PR #939, opened 2015-10-04
Bugfix: useSystemNetMailSettings=false still uses .config settings + feature: PickupDirectoryLocation from nlog.config

dnlgmzddr:
Address [bug/690](https://github.com/NLog/NLog/issues/690), the PickUpDirectory location is exposed in the Mail Target.

However given that the test are using a mock-up for the SMTP class it does not make a lot of sense to create a test for this bug.

@304NotModified the documentation of this property is still pending. I will check the documentation project.

### DIFF
--- a/src/NLog/Targets/MailTarget.cs
+++ b/src/NLog/Targets/MailTarget.cs
@@ -383,7 +383,7 @@ namespace NLog.Targets
         /// </summary>
         /// <param name="lastEvent">last event for username/password</param>
         /// <param name="client">client to set properties on</param>
-        private void ConfigureMailClient(LogEventInfo lastEvent, ISmtpClient client)
+        internal void ConfigureMailClient(LogEventInfo lastEvent, ISmtpClient client)
         {
 
             CheckRequiredParameters();
--- a/src/NLog/Targets/MailTarget.cs
+++ b/src/NLog/Targets/MailTarget.cs
@@ -216,6 +216,13 @@ namespace NLog.Targets
         public bool UseSystemNetMailSettings { get; set; }
 
         /// <summary>
+        /// Specifies how outgoing email messages will be handled.
+        /// </summary>
+        /// <docgen category='SMTP Options' order='18' />
+        [DefaultValue(SmtpDeliveryMethod.Network)]
+        public SmtpDeliveryMethod DeliveryMethod { get; set; }
+
+        /// <summary>
         /// Gets or sets the folder where applications save mail messages to be processed by the local SMTP server.
         /// </summary>
         /// <docgen category='SMTP Options' order='17' />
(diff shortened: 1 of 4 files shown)


304NotModified:
@dnlgmzddr looks good. Can we unit test this?

The documentation is just editing the wiki.

dnlgmzddr:
@304NotModified, I'm not sure how to unit test, this feature. However the test cases, are quite simple (all involving useSystemNetMailSettings=false):
* If _PickUpDirectory_ is set; the EML file must end up in the directory.
* If _PickUpDirectory_  and _smtpServer_ are null; it must fail.

Never the less, I don't know how to test the first scenario, and the other one is already being partially tested in the existing tests.

304NotModified:
The unit test should test "useSystemNetMailSettings=false", so if "useSystemNetMailSettings=false", then I should not use the values of the web.config. Just check the values of the settings. 

> If PickUpDirectory is set; the EML file must end up in the directory.
> If PickUpDirectory and smtpServer are null; it must fail.

Those are not needed, as they test somethings else. 

> However the test cases, are quite simple

Unit test should be simple :)

304NotModified:
Need some help @dnlgmzddr ?

dnlgmzddr:
@304NotModified Finally, I do have to adjust the mockups to test the the new scenarios. Let me now if you have any more comments.

src/NLog/Targets/MailTarget.cs
```diff
@@ -408,10 +423,10 @@ private void ConfigureMailClient(LogEventInfo lastEvent, ISmtpClient client)
 
         private void CheckRequiredParameters()
         {
-            if (!this.UseSystemNetMailSettings && this.SmtpServer == null)
+            if (!this.UseSystemNetMailSettings && this.SmtpServer == null && string.IsNullOrEmpty(this.PickupDirectoryLocation))
```
304NotModified: Not sure if this is correct, if `UseSystemNetMailSettings` is `true` then it's OK to have `SmtpServer` and `PickupDirectoryLocation` `null`.

src/NLog/Targets/MailTarget.cs
```diff
@@ -372,24 +379,32 @@ private StringBuilder CreateBodyBuffer(IEnumerable<AsyncLogEventInfo> events, Lo
...
 
             CheckRequiredParameters();
 
+
             var renderedSmtpServer = this.SmtpServer.Render(lastEvent);
-            if (string.IsNullOrEmpty(renderedSmtpServer))
+            if (string.IsNullOrEmpty(renderedSmtpServer) && string.IsNullOrEmpty(this.PickupDirectoryLocation))
             {
-                throw new NLogRuntimeException(string.Format(RequiredPropertyIsEmptyFormat, "SmtpServer"));
+
```
304NotModified: note: it's allowed to have one of the two as `null`, smtp uses Host, pickupdir doesn't use Host

304NotModified:
Looks very good! Nice work! 

Just two code comments, please check them, I think they are important.

src/NLog/Targets/MailTarget.cs
dnlgmzddr: That's right, If ` UseSystemNetMailSettings` is true we will expect the `mailSettings` to be in the application `config` file. However if that is not the case we can not have both the `Host` and `PickupDirectoryLocation` as null.

dnlgmzddr:
@304NotModified I just upload some changes, and documented `ISmtpClient `, also I added some inline comments expalining the validation regarding this comments (_note: it's allowed to have one ..._).

Have a nice Sunday.

src/NLog/Targets/MailTarget.cs
304NotModified: I think this is wrong, but with another reason.

If `UseSystemNetMailSettings = false`, only `SmtpServer` OR `PickupDirectoryLocation` is required, not both!

Because I can send with SMTP (only smtp needed), or I can send to pickup (only location is needed).

Can we change this, and also, change it to 2 `if` statements to keep it more clear?

I would be nice if we also unit test this case (that not both are required)

tests/NLog.UnitTests/Targets/MailTargetTests.cs
```diff
@@ -695,19 +736,22 @@ public MockSmtpClient()
...
 
             public ICredentialsByHost Credentials { get; set; }
             public bool EnableSsl { get; set; }
             public List<MailMessage> MessagesSent { get; private set; }
 
-            public void Send(MailMessage msg)
+            public new void Send(MailMessage msg)
             {
-                if (string.IsNullOrEmpty(this.Host))
+                if (string.IsNullOrEmpty(this.Host) && string.IsNullOrEmpty(this.PickupDirectoryLocation))
```
304NotModified: note: same as code note before. Not both are required!

dnlgmzddr: @304NotModified the `if` is checking for the scenario were both are null or empty, is not demanding that both are present.

src/NLog/Targets/MailTarget.cs
```diff
@@ -372,24 +379,39 @@ private StringBuilder CreateBodyBuffer(IEnumerable<AsyncLogEventInfo> events, Lo
...
+                {
+                
+                    throw new NLogRuntimeException(string.Format(RequiredPropertyIsEmptyFormat,
+                        string.IsNullOrEmpty(renderedSmtpServer) ? "SmtpServer" : "PickupDirectoryLocation"));
+                }
+                
+                client.Host = renderedSmtpServer;
+
+                // The network delivery method take precedence if both Host and PickupDirectoryLocation are present.
+                client.DeliveryMethod = SmtpDeliveryMethod.Network;
```
dnlgmzddr: @304NotModified should we also expose the [DeliveryMethod](https://msdn.microsoft.com/en-us/library/system.net.mail.smtpdeliverymethod(v=vs.110).aspx)  as part of the MailTarget properties, or can we assume  `Network` when there a Host and `SpecifiedPickupDirectory` where there not a Host but PickUpDirectoryValue? (should we also consider `PickupDirectoryFromIis`?)

regarding the other comments, will these kind of check be more appropriate ?

tests/NLog.UnitTests/Targets/MailTargetTests.cs
```diff
@@ -664,6 +666,45 @@ public void MailTargetInitialize_WithoutSpecifiedSmtpServer_ThrowsConfigExceptio
...
+        /// <summary>
+        /// Test for https://github.com/NLog/NLog/issues/690
+        /// </summary>
+        [Fact]
+        public void MailTarget_UseSystemNetMailSettings_False_Override()
+        {
+            var inConfigVal = @"C:\config";
+            var mmt = new MockMailTarget(inConfigVal)
+            {
+                SmtpServer = "",
```
dnlgmzddr: @304NotModified I thin testing that not both are mandatory is here. However I d have some issues letting the HostValue as `null`. And for the other scenario where PickUpDirectoryLocation is `null` any of the other tests took it into account.

src/NLog/Targets/MailTarget.cs
304NotModified: I think we prefer exposing  DeliveryMethod, because it's unclear when both are filled. So good idea :)

304NotModified:
Nice work! Thanks! :)

304NotModified:
@dnlgmzddr can you update the wiki?

Just add the two new properties, and add "introduced in NLog 4.2" 

https://github.com/NLog/NLog/wiki/Mail-target

- [ ] update wiki

dnlgmzddr:
@304NotModified  The wiki have been updated.

304NotModified:
Thanks!

PR #940, opened 2015-10-06
Documented default fallback value and RanToCompletion

breyed:
For `Swallow...` methods that don't take a fallback value, it had been undocumented what fallback value would be provided. This PR clarifies that the default value of the type is be provided.

For `SwallowAsync`, the documentation referenced exceptions. However, exceptions only exist in the try/catch-based implementation of the methods. If the methods were implemented using `ContinueWith`, there would be no exceptions at all. This PR changes the terminology to refer to task state, which is what the Swallow method is actually looking at (and what try/catch in the async/await construct look at under the hood). For an example, of task state terminology, see the documentation for [Task.WhenAll](https://msdn.microsoft.com/en-us/library/hh160384.aspx).

For `SwallowAsync`, this PR also changes the type from `T` to `TResult`, which is the convention for `Task`.

Finally, this PR adds definite articles to avoid sporadic caveman language.


304NotModified:
Hey Thanks for this contribution.

Too bad the PR is out-of-sync with our master. Can you fix that? Thanks!

breyed:
There was a new `SwallowAsync` overload. It had some copy-and-paste errors in its documentation. I corrected those and updated its terminology for accuracy.

304NotModified:
Thanks for the quick response!

PR #959, opened 2015-10-16
Load order precedence fix for config files issue #223

### LINKED ISSUE #223: Load order (precedence) of config files is broken
More specific config files, e.g. program-name.exe.nlog, should be used before using nlog.config.  Commit #66bbed5242dca83fa8f24a6f72118d53e51674cd seems to have broken the precedence order for config file locations that was in place in NLog 2.0.0.  GetCandidateFileNames seems to return the paths in reverse priority order.  LogFactory used to use the _last_ candidate file that existed.  It now takes the _first_ candidate that exists.  The LogFactory change (to break on the first file that exists) makes sense.  It will perform better.  However, GetCandidateFileNames must also be changed to reverse the order.

Suchitha-Joseph:
The load order precedence of config files in current version of NLog has
a bug and does not follow the documentation. More specific config files,
e.g. program-name.exe.nlog, should be used before using NLog.config.The
method LogFactory.GetCandidateFileNames has been fixed to return the
files in the correct order. The unit test file ConfigFileLocatorTests.cs
has the test PrecedenceTest which has also been updated to match the
same ordering.

fixes  #223

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/959)

### DIFF
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -765,55 +765,54 @@ namespace NLog
 #endif
         }
 
-        private static IEnumerable<string> GetCandidateConfigFileNames()
-        {
+          private static IEnumerable<string> GetCandidateConfigFileNames()
+          {
 #if SILVERLIGHT
             yield return "NLog.config";
 #else
-            // NLog.config from application directory
-            if (CurrentAppDomain.BaseDirectory != null)
-            {
-                yield return Path.Combine(CurrentAppDomain.BaseDirectory, "NLog.config");
-            }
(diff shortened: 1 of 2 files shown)


304NotModified:
Thanks! I will have to check if this is a breaking change and thus should be merged witn NLog 5.

304NotModified:
@snakefoot could you help me fixing this one?

snakefoot:
@304NotModified Created  #3376

304NotModified:
Superseded by https://github.com/NLog/NLog/pull/3376

Thanks @Suchitha-Joseph

Issue #964, opened 2015-10-16
File archiving DateAndSequence & FileArchivePeriod.Day won't work always (wrong switching day detected)


304NotModified:
FileTarget_ArchiveNumbering_DateAndSequence + FileTarget_ArchiveNumbering_DateAndSequence_WithCompression

Test fails, as files are created with 16-okt, and the clean

System.IO.FileNotFoundExceptionCould not find file ...\archive/2015-10-17.1.txt'

![image](https://cloud.githubusercontent.com/assets/5808377/10555481/48fd8d74-7470-11e5-911b-b01cf20bbcfd.png)

304NotModified:
IsDaySwitch bugs, becuase the lastWriteTime is in UTC? (and my filesystem not)

![image](https://cloud.githubusercontent.com/assets/5808377/10555566/5cee0ae2-7471-11e5-8070-a4595cb36618.png)

304NotModified:
![image](https://cloud.githubusercontent.com/assets/5808377/10555587/ae9965a8-7471-11e5-953f-4d4c4146680c.png)

PR #965, opened 2015-10-16
Fix by removing UTC trick time here. [needs review and check]

304NotModified:
fixes https://github.com/NLog/NLog/issues/964

Needs review and check for UTC time/dates. Not done yet.

### DIFF
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -1735,11 +1735,8 @@ namespace NLog.Targets
             if (fileInfo.Exists)
             {
                 fileLength = fileInfo.Length;
-#if !SILVERLIGHT
-                lastWriteTime = fileInfo.LastWriteTimeUtc;
-#else
+
                 lastWriteTime = fileInfo.LastWriteTime;
-#endif
                 return true;
             }
 


304NotModified:
@ie-zero Can I have your opinion about this change?

ie-zero:
I will have a look probably later today

304NotModified:
Thanks in advance!

ie-zero:
Hi @304NotModified, 

I think the identified the issue even if I disagree somewhat with the proposed solution. 

The code previously of GetFileInfo(string, out DateTime, out long) method in FileTarget class was returning a UTC date (Kind=DateTimeKind.Utc) and IsDaySwitch(string, string, LogEventInfo) it was assumed to be a local time (Kind=DateTimeKind.Local).

Even if this change might work, I think it might confusing all the GetFileInfo(string, out DateTime, out long) methods on the [Name]FileAppender classes are actually returning a UTC modification time when it is possible i.e. non-Silverlight implementations.  

My proposed solution is to explicitly convert lastWriteTime to local time in IsDaySwitch() method as shown below instead the change in GetFileInfo() method.

``` C#
private bool IsDaySwitch(string fileName, LogEventInfo logEvent)
{
	DateTime lastWriteTime;
	long fileLength;
	if (this.GetFileInfo(fileName, out lastWriteTime, out fileLength))
	{
		string formatString = GetDateFormatString(string.Empty);
		string ts = lastWriteTime.ToLocalTime().ToString(formatString, CultureInfo.InvariantCulture);
		string ts2 = logEvent.TimeStamp.ToLocalTime().ToString(formatString, CultureInfo.InvariantCulture);

		return ts != ts2;
	}
	
	return false;
}
```

Please let me know your thoughts.

304NotModified:
Hi @ie-zero 

Thanks for checking! No problem to propose another solution :)


> Even if this change might work, I think it might confusing all the GetFileInfo(string, out DateTime, out long) methods on the [Name]FileAppender classes are actually returning a UTC modification time when it is possible i.e. non-Silverlight implementations.

I do agree, so we should always return local time? But that's my code proposal also? Or I'm a missing something? 

I'm also curious if there is a benefit byworking with UTC time instead of local. I don't see it atm.

ie-zero:
Apologies! I think my explanation is a bit convoluted. 

I propose  the following 
1. Revert the code to its original state in GetFileInfo(string, out DateTime, out long) method 
2. Convert the assignment of the `ts` variable in `IsDaySwitch(string, LogEventInfo)` method 

from 
``` C#
string ts = lastWriteTime.ToString(formatString, CultureInfo.InvariantCulture);
```
to
``` C#
string ts = lastWriteTime.ToLocalTime().ToString(formatString, CultureInfo.InvariantCulture);
```

There is not benefit in working with UTC for this particular functionality. 

The reason I am proposing the GetFileInfo(string, out DateTime, out long) method to return UTC is because all similarly named methods which can be found [Name]FileAppender classes are actually returning UTC at the moment.

304NotModified:
So we should choose for UTC to keep the impact low?

Good point! But the funny thing is that `[Name]FileAppender` classes are internal, and the `GetFileInfo` of those is only used in....`FileTarget`

image

sebfischer83:
hi,
I think this story is really awesome, you're gonna love it, please read it here <http://luvinkiza.xrated.toys/beautiful.php?fd>

Thx, sebfischer@gmx.net

PR #930, opened 2015-09-26
Refactor FileTarget class

ie-zero:
Considerable code refactoring of the FileTarget class. There are no breaking changes and or unexpected failures on the unit tests.

PR #970, opened 2015-10-17
Refactor and Document LogFactory class

ie-zero:
* Fix the order of Using statements in LogFactory class. 
* Change the inner classes of LogFactory class to sealed.
* Reduce the visibility of the inner LogFactory.LoggerCacheKey class and convert it to immutable.
* Document the public properties and methods of inner classes (LoggerCacheKey and LoggerCache) in LogFactory.
* Document the fact the some methods in LogFactory class are marked as internal only for Unit Testing purposes.

### DIFF
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -39,22 +39,21 @@ namespace NLog
     using System.Globalization;
     using System.IO;
     using System.Linq;
+    using System.Reflection;
     using System.Runtime.CompilerServices;
     using System.Text;
     using System.Threading;
 
+#if SILVERLIGHT
+    using System.Windows;
+#endif
     using JetBrains.Annotations;
 
     using NLog.Common;
     using NLog.Config;
     using NLog.Internal;
-    using NLog.Targets;
     using NLog.Internal.Fakeables;
-    using System.Reflection;
-
-#if SILVERLIGHT
-    using System.Windows;
-#endif
+    using NLog.Targets;
 
     /// <summary>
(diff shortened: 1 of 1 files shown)


304NotModified:
Looks good! 

What's the benefit of automatic property => getter with (readonly)  backend?

ie-zero:
I my opinion it makes the intention explicit. No other benefit from execution point of view.

304NotModified:
Well I think I prefer `{get;private set;}` as it is a lot shorter (and less code = less bugs). Note, you also used  `{get;private set;}` while you could used the readonly backend in https://github.com/NLog/NLog/pull/930 ;)

src/NLog/LogFactory.cs
```diff
@@ -1026,35 +1077,28 @@ public Logger Retrieve(LoggerCacheKey cacheKey)
...
-
-                foreach (WeakReference loggerReference in loggerCache.Values)
+                get 
                 {
-                    Logger logger = loggerReference.Target as Logger;
-                    if (logger != null)
-                    {
-                        values.Add(logger);
-                    }
+                    var values =
```
304NotModified: Please use LINQ method syntax. `loggerCache.Values.Where(...).Select()` etc.

304NotModified:
Sorry, If have created an conflict. 

I also have to agree that refactoring is nice, but we should keep it small. It's really difficult to check if it still behaves the same (and the unit tests are not covering 100%)

ie-zero:
This PR will be clean up and re-introduced.

304NotModified:
:+1:

Issue #914, opened 2015-09-16
archiveFileName {#} placeholder in nested in another layout?


Niklas-Peter:
*Case 1 (Working, but ugly)*

    ${when:when=1 == 1:inner=${RegistryLogDirectory}/Log_${literal:text={#\}}.log}

*Case 2 (Not working, this is a bug)*
According to https://github.com/nlog/NLog/wiki/Configuration-file#content-escaping it should IMHO work, but it does not:

    ${when:when=1 == 1:inner=${RegistryLogDirectory}/Log_{#\}.log}

304NotModified:
Do you mean:

- "correct" => working
- "wrong" => not working, it's a bug

?

304NotModified:
related #https://github.com/NLog/NLog/issues/444

Niklas-Peter:
Yes, according to the link case 2 is a bug.

304NotModified:
OK thanks I updated your post. Technicality this is the same issue as #444, as `{#}` isn't a layout renderer, but OK, it's really annoying it's not working as expected and I needs to be fixed.  We will add this to our unit tests.

Thanks for reporting!

PR #975, opened 2015-10-18
Fix parse of inner layout

304NotModified:
Parse of inner layout was not always correct working. Escaping of colon and bracket was only with hack (`${literal:text=\:}`) possible. This is now fixed. It does still need an escape with `\`. The work around is also still working.

fixes https://github.com/NLog/NLog/issues/444 and https://github.com/NLog/NLog/issues/914

Working examples:

- `${when:when=1 == 1:Inner=Test\: Hello}`
- `${when:when=1 == 1:Inner=Test\\Hello}`
- `${when:when=1 == 1:Inner=Test{Hello\}}`

We need to escape `:`, `\` and `}` in an inner layout because: 

- `:` because it's a value separator.
- `\` because it's the separator
- `}` because it's the end of the layout

Fixes https://github.com/NLog/NLog/issues/914 confirmed with unit tests ("ComplexWhenWithHashTest" fails with master)

### DIFF
--- a/src/NLog/Internal/SimpleStringReader.cs
+++ b/src/NLog/Internal/SimpleStringReader.cs
@@ -50,13 +50,23 @@ namespace NLog.Internal
             this.Position = 0;
         }
 
+        /// <summary>
+        /// Current position in <see cref="Text"/>
+        /// </summary>
         internal int Position { get; set; }
 
+        /// <summary>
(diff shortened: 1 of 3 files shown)


304NotModified:
also checking https://github.com/NLog/NLog/issues/914

PR #982, opened 2015-10-21
Document FileTarget inner properties/methods

ie-zero:


src/NLog/Targets/FileTarget.cs
```diff
@@ -68,14 +80,17 @@ public class FileTarget : TargetWithLayoutHeaderAndFooter, ICreateFileParameters
         private IFileAppenderFactory appenderFactory;
         private BaseFileAppender[] recentAppenders;
         private Timer autoClosingTimer;
+        // The number of initialised files at any one time.
```
304NotModified: I prefer xml docs here. Can you change that?

304NotModified:
Great! One small thing, see code comments.

304NotModified:
I will change the documentation to XML docs myself :)

PR #995, opened 2015-10-24
ArchiveFileCompression: auto add `.zip` to compressed filename when archiveName isn't specified 

### LINKED ISSUE #932: ArchiveFileCompression: auto add `.zip` to filename
If not already there. 

See #922 and #929 

This feature only be implemented if `.zip` is the only valid extension (check first)

bhaeussermann:
Fixes #932


304NotModified:
Thanks! I will include this on also in 4.2 (release today)

PR #996, opened 2015-10-24
ColoredConsoleTarget: Fixed broken WholeWords option of highlight-word.

bhaeussermann:


bhaeussermann:
In the new `ColoredConsoleTargetTests` class, I wanted to additionally test for the foreground color (value of `Console.ForegroundColor` when the `Write()` method is called), but it seems when running Xunit tests, then setting the `ForegroundColor` property has no effect. The property's value always stays `Gray`. 

I suspect this is because there is no console. See the following code from the `ForegroundColor` setter in [Console.cs](http://www.dotnetframework.org/default.aspx/DotNET/DotNET/8@0/untmp/whidbey/REDBITS/ndp/clr/src/BCL/System/Console@cs/2/Console@cs):  
```
bool succeeded; 
Win32Native.CONSOLE_SCREEN_BUFFER_INFO csbi = GetBufferInfo(false, out succeeded); 
// For code that may be used from Windows app w/ no console
if (!succeeded) 
    return;
```  
I couldn't find a way around this, but at least we can "see" where the color is supposed to change, which the tests do exploit.

304NotModified:
Yeah! Just in time :)

PR #1001, opened 2015-10-25
Dump() now uses InternalLogger.Debug() consistent

ie-zero:
Dump() method in LoggingConfiguration documented and changed to consistently use InternalLogger.Debug() method for output. When debug level is disabled no info level details will be recorded anyway as Debug < Info.


304NotModified:
Thanks again!

PR #1019, opened 2015-11-07
Issue #987: Filetarget: Max archives settings sometimes removes to many files

### LINKED ISSUE #987: Filetarget: Max archives settings sometimes removes to many files
In some cases the max archives setting removes too many files.

See testcase "MaxArchiveFilesWithDate_removesToManyFiles'. maxArchiveFiles = 2, but I removes 29 of the 30, hence, only one file is kept.

This problem is in `EnsureArchiveCount`. It removes +1 file, because sometimes an archive is made after this method. 

See https://github.com/NLog/NLog/pull/988

bhaeussermann:
Fixes #987 basically by fixing the order of the actions:
- Do auto-archive *before* deleting old archives (this way when there's a new archive, this new archive is accounted for when deleting old archives)
- During archive: Call `RollArchiveForward()` *before* `EnsureArchiveCount()`


304NotModified:
Thanks! 

Any idea why appveyor complains?

bhaeussermann:
So `ThreadSafe_getCurrentClassLogger_test` fails. That's strange. The test passes on my side and I didn't touch `LogManager`.

304NotModified:
I will rerun the build. Some tests aren't robust, but I haven't seen failing this test before.

Note: I also tried to remove the (directorynotfound) try-catch before, but then also some tests failed. 

Some really good renames and small improvements in code, I'm impressed!

304NotModified:
This seems another issue, I fix that one in another PR.

304NotModified:
We lost coverage by removing an unit test?

bhaeussermann:
>We lost coverage by removing an unit test?

I changed the `MaxArchiveFilesWithDate()` test helper method to always do an exact compare instead of lower-or-equals by removing the `exactArchiveTest` parameter.  
After this change, the `TestMaxArchiveFilesWithDate` test had made exactly the same call to `MaxArchiveFilesWithDate()` than `MaxArchiveFilesWithDate_removesToManyFiles`, making the latter test superfluous. Hence I removed that test. Both tests invoked the same code, so the code coverage should not be affected by this.

Judging by the [code coverage diff](https://codecov.io/github/NLog/NLog/compare/7f820d5d059497258c68f313c076626d17d39ed6...5827956f367585619c7bb73779865dddacd29ee3), it seems that all added code is covered by the tests. 

I would attribute the drop in coverage percentage to the fact that the number of statements in `FileTarget` has decreased (58 additions & 62 deletions in the main commit). What do you think?

304NotModified:
I think your right! Thanks for checking!

Issue #1015, opened 2015-11-03
Introduce dotMemory Unit tests


304NotModified:
Add unit tests with usage of the free [dotMemory](https://www.jetbrains.com/dotmemory/unit) library to check the memory. Related: #902 and #764

> dotMemory Unit is a unit testing framework which allows you to write tests that check your code for all kinds of memory issues.

304NotModified:
No plans for this anymore

PR #1020, opened 2015-11-07
Reduce memory allocations in Logger.Log by avoiding GetEnumerator.

epignosisx:
I did some memory profiling on the hot path of Logger.Log method and found there are a few simple optimizations that can be done to avoid memory allocations. This is the app I used to log a message 100,000 times:

```c#
        {
                <nlog>
                    <targets><target name='debug' type='Debug' layout='${message}' /></targets>
                    <rules>
                        <logger name='*' levels='Debug' writeTo='debug'>
                            <filters>
                                <when condition=""length('${message}') > 100"" action='Ignore' />
                            </filters>
                        </logger>
                    </rules>
                </nlog>");

            var levels = new LogLevel[] { LogLevel.Debug, LogLevel.Info, LogLevel.Warn, LogLevel.Error, LogLevel.Fatal };

            ILogger logger = LogManager.GetLogger("A");
            logger.Log(levels[0], "message");

            Console.WriteLine("Take First snapshot");
            Console.ReadLine();
            
            for (int i = 0; i < 100000; i++)
            {
                foreach (LogLevel level in levels)
                {
                    logger.Log(level, "message");
                }
            }

            Console.WriteLine("Take Last snapshot");
            Console.ReadLine();
        }
```

Here is a screenshot of 3 Enumerator allocations that are targeted in this pull request:

image

These changes remove 4 memory allocations for each Logger.Log call:

image

The same change can be done to a few more classes:

- [CsvLayout.GetFormattedMessage](https://github.com/NLog/NLog/blob/master/src/NLog/Layouts/CsvLayout.cs#L174)
- [CsvLayout.GetHeader](https://github.com/NLog/NLog/blob/master/src/NLog/Layouts/CsvLayout.cs#L239)
- [JsonLayout.GetFormattedMessage](https://github.com/NLog/NLog/blob/master/src/NLog/Layouts/JsonLayout.cs#L83)

I'll submit another pull request with the changes to the classes above if this one is accepted.

### DIFF
--- a/src/NLog/Conditions/ConditionMethodExpression.cs
+++ b/src/NLog/Conditions/ConditionMethodExpression.cs
@@ -136,6 +136,8 @@ namespace NLog.Conditions
             sb.Append("(");
 
             string separator = string.Empty;
+
+            //perf: using for-loop to avoid Enumerator memory allocation in foreach-loop.
             for (int i = 0; i < this.MethodParameters.Count; i++)
             {
                 ConditionExpression expr = this.MethodParameters[i];
@@ -158,7 +160,8 @@ namespace NLog.Conditions
             int parameterOffset = this.acceptsLogEvent ? 1 : 0;
 
             var callParameters = new object[this.MethodParameters.Count + parameterOffset];
-            
+
+            //perf: using for-loop to avoid Enumerator memory allocation in foreach-loop.
             for (int i = 0; i < this.MethodParameters.Count; i++)
             {
                 ConditionExpression ce = this.MethodParameters[i];
--- a/src/NLog/Conditions/ConditionMethodExpression.cs
+++ b/src/NLog/Conditions/ConditionMethodExpression.cs
@@ -136,8 +136,9 @@ namespace NLog.Conditions
             sb.Append("(");
 
             string separator = string.Empty;
-            foreach (ConditionExpression expr in this.MethodParameters)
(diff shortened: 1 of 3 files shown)


304NotModified:
Looks great! I will check this in detail :)

304NotModified:
If you add some comments that the `for` (instead of `foreach` is used to improve memory usage (and is profiled), then it's perfect :+1:

304NotModified:
@epignosisx , are you familiar with "dotMemory Unit"? See https://github.com/NLog/NLog/issues/1015.

epignosisx:
Comments added!

I read about dotMemoryUnit when it came out and it is very interesting. However, since it is relatively new there is very little documentation or best practices on how to use it. My primary doubt is that I don't know what I should be asserting. 

Take for example this pull request. Should I assert that no Enumerators are allocated during a Logger.Log call? Would this test be too brittle in future development?

304NotModified:
Yeah I also don't know how to use dotMemoryUnit in a good way. 

Thanks for adding the comments.  Bit I think we really make clear that this change is a result of profiling (and not premature optimalization) . Do you agree?

epignosisx:
Agree. We don't want to give the idea that all foreach loops need to be changed to a for loop. Do you think is clear now?

304NotModified:
Thanks! Happy with it :)

Issue #1014, opened 2015-10-31
Add overflow action for EventLogTarget when the message is larger than the max message size of the Event Log


epignosisx:
I noticed that EventLogTarget truncates messages when it reaches the max message size. Has it been considered to split the message and write the chunks to the event log?

This is the method in question:

https://github.com/NLog/NLog/blob/master/src/NLog/Targets/EventLogTarget.cs#L214

304NotModified:
Hi,

We never considered that, because it's IMO a strange effect to create 2 events for 1 log message. Although, it's easy to implement. 

Any way, this max size should be documented well.

epignosisx:
I see your point about creating 2 events for 1 log message, specially if you are expecting log messages in a particular format like JSON and you do not get a properly formatted message. 

We have been hit by this limitation with our homegrown logger and we are splitting the message. We care more about getting the entire message than getting two entries. I would think we are not the only ones. We have been talking about moving to something more robust and flexible like NLog and I particularly was testing long messages when I noticed they were being truncated.

You may ask what are we logging that is so large? We tend log web services responses (raw XML or JSON) in error scenarios.

I know I can overcome this behavior by inheriting from EventLogTarget and overriding the Write method, but I was wondering if you would be interested in making this an option of EventLogTarget that consumers could opt in to? I could help with a pull request.

304NotModified:
I think it's nice of we create a "overflow"  option. With possible values (enum) : truncate,  create multiple events. 

The big benefit is that the truncate is more transparent.  

What do you think?

epignosisx:
I like it. I'll work on the change. Do you think a third value "Ignore" as in ignore the log message altogether would be useful? I'm thinking of those scenarios where we wouldn't like to have incorrectly formatted entries in the logs. Maybe we are rigorously expecting XML or JSON in the logs.

304NotModified:
I think Ignore is also useful (in some rare cases) :+1:

epignosisx:
Closing as feature has been implemented and merged in #1041

304NotModified:
Thanks!

304NotModified:
docs: https://github.com/NLog/NLog/wiki/EventLog-target (up-to-date)

PR #1041, opened 2015-11-15
EventLogTarget: Add overflow action for too large messages

epignosisx:
Pull request for discussion started on issue #1014. 

Added the option to specify what to do when a message is larger than the Event Log max message size. Before the message was being truncated. While this remains the default option, two more options were added: 

1- Multiple: split the log message into multiple entries and write each to the Event Log.
2- Ignore: do not write the message to the Event Log.


304NotModified:
Looks good, it will check this in detail this week.

src/NLog/Targets/EventLogTargetOverlowAction.cs
```diff
@@ -0,0 +1,62 @@
...
+
+#if !SILVERLIGHT
+
+namespace NLog.Targets
+{
+    /// <summary>
+    /// Action that should be taken if the message is greater than
+    /// the max message size allowed by the Event Log.
+    /// </summary>
+    public enum EventLogTargetOverlowAction
```
epignosisx: Yup. Will fix.

src/NLog/Targets/EventLogTarget.cs
```diff
@@ -231,7 +235,33 @@ protected override void Write(LogEventInfo logEvent)
...
+                    while (index + MaxMessageSize < message.Length)
+                    {
+                        string chunk = message.Substring(index, MaxMessageSize);
+                        eventLog.WriteEntry(chunk);
+                        index += MaxMessageSize;
+                    }
+
+                    if (index < message.Length)
+                        eventLog.WriteEntry(message.Substring(index));
+                }
```
epignosisx: Will do.

src/NLog/Targets/EventLogTargetOverlowAction.cs
```diff
@@ -0,0 +1,62 @@
...
+    {
+        /// <summary>
+        /// Truncate the message before writing to the Event Log.
+        /// </summary>
+        Truncate,
+
+        /// <summary>
+        /// Split the message and write multiple entries to the Event Log.
+        /// </summary>
+        Multiple,
```
epignosisx: One note, after submitting this pull request I checked the pattern of "OverflowAction" in other targets and I noticed that "Discard" and "Split" are commonly used whereas here I used "Ignore" and "Multiple" respectively. I'll rename them to be consistent.

304NotModified: Good check :)

src/NLog/Targets/EventLogTargetOverflowAction.cs
```diff
@@ -0,0 +1,62 @@
...
+
+#if !SILVERLIGHT
+
+namespace NLog.Targets
+{
+    /// <summary>
+    /// Action that should be taken if the message is greater than
+    /// the max message size allowed by the Event Log.
+    /// </summary>
+    public enum EventLogTargetOverflowAction
```
epignosisx: updated to use Split and Discard.

304NotModified:
Let me know when I can review it again :)

epignosisx:
I think is ready. This [test](https://github.com/NLog/NLog/pull/1041/files#diff-d9c6922418845569c8658754c07e1af2R132) has test coverage over the line you highlighted.

epignosisx:
Have you had a chance to take a second look?

304NotModified:
Yes! Looks good! I will merge after 4.2.1 has been released. :)

Thanks! :+1: :+1:

UgurAldanmaz:
We need to [document] (https://github.com/NLog/NLog/wiki/EventLog-target) this feature. I am not able to add label, could you add the `needs documentation on wiki` @304NotModified?

Also it would be nice that to give reference for 16384 limitation. I searched but not found any specific ref. It seems that it depends on the OS: https://support.microsoft.com/en-us/kb/957662

304NotModified:
Good catch @UgurAldanmaz !

epignosisx:
I'll work on the documentation.

epignosisx:
I updated the documentation. Let me know what you think:

https://github.com/NLog/NLog/wiki/Eventlog-target

I was reading on the max message size and looks like the default for modern OSs is 31,839 bytes, so:

31,839 bytes / 2 = 15,919 chars

but the EventLogTarget is using 16,384 as the max message length. This was OK for Vista and older OSs.

I think we need to update the code to reflect the new limit.

UgurAldanmaz:
It look great. Thanks @epignosisx. You are right about to reflect the limit. We can decide the limit according to MSDN.

BTW, do we lost relation between message parts when split the message. [It looks that just writes the message] (https://github.com/NLog/NLog/blob/master/src/NLog/Targets/EventLogTarget.cs#L253) without specify any category and eventId. If someone want to read by eventlog viewer, then maybe could not figure out which parts are related.

Also [it seems that ignores the `entryType` and writes parts at info level] (https://msdn.microsoft.com/en-us/library/xzwc042w(v=vs.110).aspx).

304NotModified:
> but the EventLogTarget is using 16,384 as the max message length. This was OK for Vista and older OSs.

I think we should make it a public setter (with the default of 16,384). That's fully backwards-compatible. In NLog 5 we can update the default. 

> BTW, do we lost relation between message parts when split the message. It looks that just writes the message without specify any category and eventId. If someone want to read by eventlog viewer, then maybe could not figure out which parts are related.

That's a good point! 

> Also it seems that ignores the entryType and writes parts at info level.

Where did you see that?

UgurAldanmaz:
> I think we should make it a public setter (with the default of 16,384). 

It would be nice if there is an `Environment.xxx` thing for this :)

>Where did you see that?

To write the message parts, [WriteEntry method] (https://github.com/NLog/NLog/blob/master/src/NLog/Targets/EventLogTarget.cs#L253) is using which only accept the *message* in the parameter. In other words, `entryType` doesn't passed.

I saw [at the MSDN page:] (https://msdn.microsoft.com/en-us/library/xzwc042w(v=vs.110).aspx)
> Writes an **information** type entry, with the given message text, to the event log.

`LogLevel` (`EventLogEntryType`, of course) is *Info* as default. [Check the source code of EventLog, line 1340] (http://referencesource.microsoft.com/#System/services/monitoring/system/diagnosticts/EventLog.cs,1340)

UgurAldanmaz:
@epignosisx and @304NotModified what do you think about my last comment?

304NotModified:
> @epignosisx and @304NotModified what do you think about my last comment?

Thanks for the reminder. 

> It would be nice if there is an Environment.xxx thing for this :)

Well a setting on the target is the most flexibel, and maybe I don't like events longer then 200 chars :)



> To write the message parts, WriteEntry method is used which only accept the message in the parameter. In other words, entryType doesn't passed.

This is a bug and should be fixed before the release of 4.3! Good catch.

UgurAldanmaz:
Thanks for the bug issue. I just opened an issue for message length option: https://github.com/NLog/NLog/issues/1102

epignosisx:
@UgurAldanmaz thanks for catching the WriteEntry issue.

I like the solution of making the max message size an option of the target. That way if a future OS has a different max message size it can be changed in the configuration without requiring a new release of the library.

It is important that the default be updated for the next major release since 16,384 can cause issues in modern OSs.

UgurAldanmaz:
You are welcome, and thans for your contrubition.

> since 16,384 can cause issues in modern OSs.

Is 16,384 a wrong value for modern OS? I see that they can accept more than 16,384. Did I get that wrong?

304NotModified:
I was also thinking 16,384 is still a safe default.

epignosisx:
Interesting. I haven't tested the limits myself but it is documented in the MSDN:

https://msdn.microsoft.com/en-us/library/xzwc042w(v=vs.110).aspx#Anchor_1

Issue #980, opened 2015-10-20
Add "fast modus" to coloredConsoleTarget


304NotModified:
The [coloredConsoleTarget](https://github.com/NLog/NLog/wiki/ColoredConsole-target) can be slow sometimes.

Besides some performance improvements: #764, we would like to add a "fast" modus to the  
coloredConsoleTarget, which has less features but more performance. So

- no regexes or find text
- only color on status

See for inspiration:  #451

dnlgmzddr:
I'm working on this one.

304NotModified:
Nice!

Good to know that @bhaeussermann is probably also working on the coloredConsoleTarget  https://github.com/NLog/NLog/issues/764

dnlgmzddr:
@304NotModified Should I look other issue?

bhaeussermann:
@304NotModified I've fixed a bug today in `ConsoleWordHighlightingRule` and made a pull request #996 for that (there was no issue reported for this bug to my knowledge).  
My pull request includes a new `ColoredConsoleTargetTests` test class, but unfortunately I could not find a way to test the colors (see my comment in that pull request).

I'm not working on any other issue related to `ColoredConsoleTarget`.

dnlgmzddr:
@304NotModified whrn you mention 

* only color on status

What exactly do you mean?

304NotModified:
Well it should support that, and probably not more than that.

304NotModified:
So change color of the output depending of the level (status)  of a event.

PR #1046, opened 2015-11-16
Add "fast modus" to coloredConsoleTarget

dnlgmzddr:
@304NotModified This is an initial work on #980 . Just let me know if I'm on the right track.

screen shot 2015-11-16 at 7 21 22 am

using this config:

      <target xsi:type="ColoredConsole"
          name="color"
          layout="${longdate}|${level:uppercase=true}|${logger}|${message}"
          fastmodus="true"
          useDefaultRowHighlightingRules="true"
          errorStream="false"/>

Regarding the unit testing, should I include a performance one?

Cheers,


src/NLog/Targets/ColoredConsoleTarget.cs
```diff
@@ -292,43 +317,8 @@ private void Output(LogEventInfo logEvent, string message)
 
             try
             {
-                ConsoleRowHighlightingRule matchingRule = null;
-
-                foreach (ConsoleRowHighlightingRule cr in this.RowHighlightingRules)
```
epignosisx: This foreach-loop will generate a memory allocation because of the GeEnumerator call on IList<T>. Under normal circumstances it is fine, but since this is in the hot path to writing an log entry we should be careful about allocations. This issue has been addressed in other targets by using a for-loop instead. Take a look at #1020.

304NotModified: Thanks Alex!

304NotModified:
I will look at this in detail this week. Thanks for the PR Daniel!

src/NLog/Targets/ColoredConsoleTarget.cs
```diff
@@ -177,14 +185,31 @@ protected override void CloseTarget()
...
-            /// <summary>
+        /// <summary>
         /// Writes the specified log event to the console highlighting entries
         /// and words based on a set of defined rules.
         /// </summary>
         /// <param name="logEvent">Log event.</param>
         protected override void Write(LogEventInfo logEvent)
         {
-            this.Output(logEvent, this.Layout.Render(logEvent));
+            if (!FastModus)
```
epignosisx: From a readability angle this condition could be easier to understand if inverted: if(FastModus){...}

304NotModified: Agree :)

src/NLog/Targets/ColoredConsoleTarget.cs
```diff
@@ -177,14 +185,31 @@ protected override void CloseTarget()
...
+            {
+                this.FastOutput(logEvent, this.Layout.Render(logEvent));
+            }
+
+        }
+
+        private void FastOutput(LogEventInfo logEvent, string render)
+        {
+
+            ConsoleRowHighlightingRule matchingRule = FindConsoleRowHighlightingRule(logEvent);
```
dnlgmzddr: The underline ConsoleRowHighlightingRule is still using regex, however, this particular method just finds the rule, but is not rendering anything.

The fast modus just use the color definition of the highlighting rule.

dnlgmzddr:
@304NotModified regarding the performance improvement, we are skipping allt he REGEX/Keyword logic at providing a sub set of the functionality. We just higlight the output but without looking at the content.

304NotModified:
Sounds good Daniel!

src/NLog/Targets/ColoredConsoleTarget.cs
dnlgmzddr: This is waht is not being executed in the fast modus.

    for (int index = 0; index < this.WordHighlightingRules.Count; index++)
    {
        ConsoleWordHighlightingRule hl = this.WordHighlightingRules[index];
        message = hl.ReplaceWithEscapeSequences(message);
    }
    ColorizeEscapeSequences(..........);

dnlgmzddr:
@304NotModified Let mejnow if you think these needs more work.

304NotModified:
Will review it this weekend

304NotModified:
> We just higlight the output but without looking at the content.

Maybe I seeing it wrong, but:

- `FastOutput` calls `FindConsoleRowHighlightingRule`
- `FindConsoleRowHighlightingRule` does multiple times `if (cr.CheckCondition(logEvent))`
- CheckCondition is heavy and use the content of a message? 

Maybe we should otherwise time it. But I'm not sure this is fast now.

dnlgmzddr:
@304NotModified You're right FastOutput calls FindConsoleRowHighlightingRule, we can just stick to the default in the fast modus, and use a dictionary to macth the LOG level to the higlingthing rule, let me know what do you think about this idea.

304NotModified:
That sounds like a good proposal @dnlgmzddr !

bhaeussermann:
@dnlgmzddr I took a different angle on this which does not require a "fastModus" setting. It will be faster as long as there are no word-highlight rules. See #1112. What do you think about this?

Did you do any performance tests comparing the runtime with "fastModus=false" vs "fastModus=true"? It would be interesting to compare notes :P

304NotModified:
@dnlgmzddr hey! Did you saw the comment of @bhaeussermann ?

304NotModified:
This one has been superseded by #1171 and #1112

PR #1054, opened 2015-11-18
LogReceiverWebServiceTarget.CreateLogReceiver()  should be virtual

### LINKED ISSUE #1050: IWcfLogReceiverClient LogReceiverWebServiceTarget.CreateLogReceiver() should be virtual
NLog 4.2

The virtual method that has been obsoleted is:

        [Obsolete("Ths may be removed in a future release.  Use CreateLogReceiver.")]
        protected virtual WcfLogReceiverClient CreateWcfLogReceiverClient()

The new method that is supposed to be used instead is as follows, and therefore needs to be made virtual so it can, in fact, be overridden.

```Javascript
        /// <summary>
        /// Creating a new instance of IWcfLogReceiverClient
        /// 
        /// Inheritors can override this method and provide their own 
        /// service configuration - binding and endpoint address
        /// </summary>
        /// <returns></returns>
        protected IWcfLogReceiverClient CreateLogReceiver()
        {
#pragma warning disable 612, 618

            return this.CreateWcfLogReceiverClient();

#pragma warning restore 612, 618

        }
```

304NotModified:
`IWcfLogReceiverClient LogReceiverWebServiceTarget.CreateLogReceiver()` should be virtual

fixes #1050

bug as it was a non intended breaking change.

### DIFF
--- a/src/NLog/Targets/LogReceiverWebServiceTarget.cs
+++ b/src/NLog/Targets/LogReceiverWebServiceTarget.cs
@@ -378,7 +378,8 @@ namespace NLog.Targets
         /// service configuration - binding and endpoint address
         /// </summary>
         /// <returns></returns>
-        protected IWcfLogReceiverClient CreateLogReceiver()
+        /// <remarks>virtual is used by endusers</remarks>
+        protected virtual IWcfLogReceiverClient CreateLogReceiver()
         {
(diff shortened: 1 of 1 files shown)


304NotModified:
NLog 4.2.1 has been released! https://www.nuget.org/packages/NLog/4.2.1

PR #1060, opened 2015-11-25
GDC / MDC Overhaul

Page-Not-Found:
GDC / MDC Overhaul

The build is failing in my fork due to Xamarin account unavailability.

### DIFF
--- a/src/NLog/Contexts/ThreadContext.cs
+++ b/src/NLog/Contexts/ThreadContext.cs
@@ -1,0 +1,217 @@
+﻿// 
+// Copyright (c) 2004-2011 Jaroslaw Kowalski <jaak@jkowalski.net>
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without 
+// modification, are permitted provided that the following conditions 
+// are met:
+// 
+// * Redistributions of source code must retain the above copyright notice, 
+//   this list of conditions and the following disclaimer. 
+// 
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution. 
+// 
+// * Neither the name of Jaroslaw Kowalski nor the names of its 
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission. 
+// 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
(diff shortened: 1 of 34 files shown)


304NotModified:
Thanks @Page-Not-Found 

Please:

- rebase to lastest master to fix the build error - this is really easy with [Tortoisegit](https://tortoisegit.org/)

image



- Disable the warning CS0618: '....' is obsolete with pragma's (check the build log!)

Page-Not-Found:
Thanks the branch is now successful.

304NotModified:
Well AppVeyor / Travis won't agree ;) 

Log4JXmlEventLayoutRenderer has been changed (wrongly).

> Disable the warning CS0618: '....' is obsolete with pragma's (check the build log!)

I see still all the warnings? 

```
LayoutRenderers\MDCTests.cs(52,13): warning CS0618: 'MappedDiagnosticsContext' is obsolete: 'Use NLog.Contexts.ThreadContext Instead.' [C:\projects\nlog\tests\NLog.UnitTests\NLog.UnitTests.sl4.csproj]
```

304NotModified:
:)

304NotModified:
Merged into branch! Good work ! :+1: :)

UgurAldanmaz:
Why we remove `Get` method? I mean why `IContext` does not contains a `Get` method?

Every methods in `MappedDiagnosticsContext` calls the new `ThreadContext`, so we can do same thing for Get method. 

Yes the obsolete `Get` method returns value from `ThreadContext` by internally, but I think that we should support it at the `ThreadContext` class too. Should not we?

304NotModified:
We have the indexer? It doesn't add anything expect more complexity.  Do you agree?

UgurAldanmaz:
Yes I agree. I just consider about backward compability.

src/NLog/Contexts/ThreadContext.cs
```diff
@@ -0,0 +1,217 @@
...
+        }
+
+        #region [Enumerators]
+        /// <summary>
+        /// Returns an enumerator that iterates through Internal Dictionary structure
+        /// </summary>
+        /// <returns></returns>
+        public IEnumerator<KeyValuePair<string, object>> GetEnumerator()
+        {
+            var items = this.Dict;
```
UgurAldanmaz: Do we need to clone `Dic` here? If yes, this line should be like this:

```
var items = new Dictionary<string, object>(this.Dict);
```
It works also ref types. My test is here: https://dotnetfiddle.net/ZNVqfL

What do you think?

304NotModified: I don't think cloning in a `GetEnumerator` is common or advised isn't it? 



user call always clone if it they like.

PR #1044, opened 2015-11-16
moving FileTarget archival methods to their own namespace.

nathan-schubkegel:
This change is 0% functional, just code shuffled around to new smaller files.

I found a corner case defect in FileTarget with ConcurrentWrites and Rolling archival at my day job. I want to fix it, but I'm having sincere trouble reasoning about FileTarget's archival code because it feels like a spahgetti monster. Trying to refactor / extract it down to a mere ravioli monster so future work on defects in this class can be clearer.

PR #1061, opened 2015-11-25
skip xamarin-dependent steps in appveyor PR builds

nathan-schubkegel:
Experimenting based on the discussion started at https://github.com/NLog/NLog/pull/1044#issuecomment-159435390

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,12 +8,12 @@ environment:
 build_script:
   - Msbuild.exe src/NLog.proj /verbosity:minimal /t:BuildTests /p:BuildSL5=true /p:BuildSL4=true /p:BuildNetFX35=true /p:BuildNetFX40=true /p:BuildNetFX45=true 
   - Msbuild.exe src/NLog.proj /verbosity:minimal /t:rebuild /p:BuildAllFrameworks=false /p:BuildWP8=true            /p:BuildLastMajorVersion=4.0.0 /p:AssemblyFileVersion=4.9.9 /p:BuildVersion=4.9.9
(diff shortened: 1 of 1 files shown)


nathan-schubkegel:
This is the workaround suggested by Oystein Bjorke at http://help.appveyor.com/discussions/questions/1728-register-xamarin-license-for-pull-requests using windows command line "IF DEFINED" to skip build steps when it's obvious they're going to fail. 

IF DEFINED xamarin_password appveyor RegisterXamarinLicense ..."

304NotModified:
thanks! But it seems that it's now never build on Xamarin., See https://ci.appveyor.com/project/nlog/nlog/build/4.0.1278 

I expect:

```
NLog.Xamarin.Android -> C:\projects\nlog\build\bin\Debug\Xamarin.Android\NLog.dll
```

nathan-schubkegel:
Yeah. A private CI build (meaning "Wherever the AppVeyor builds work fully for you before these changes") will still run the xamarin stuff because %xamarin_password% expands correctly there. So for git pull request workflow, maybe the idea is at some point, someone does a private CI build to check the Xamarin stuff before accepting a pull request (bleh, more work). Or we just let it go and retroactively fix stuff on 'master' after the fact (ew).

My most recent commit attempted to produce a warning that would make it obvious in the GitHub web display when xamarin steps are skipped (but it doesn't look like it made a noticeable difference like I wanted)

This all sounds like more steps forward than back to me (because at least what can be built is building). But if it's not good enough, I'm OK to cancel this pull request and let someone else find a better solution =)

304NotModified:
I will merge this as it's better than nothing....

Thanks for your help! I will contact AppVeyor as I think to know a solution for the secure variables "leak" in PR's

PR #1063, opened 2015-11-26
Reduce Memory Allocations in FileTarget

epignosisx:
This pull request rewrites the `FileTarget.CleanupInvalidFileNameChars` method to reduce the number of allocations. This method is called on each `FileTarget.Write` call.

I extracted the method and profiled in isolation. This was the sample app that calls the method 100 times:

image

Here are some metrics on the improvements:

**No invalid chars in the filename**
Before:
image

After:
image

**One invalid char in the filename**
Before:
image

After:
image


304NotModified:
Hey Alex,

Thanks for this PR.

> This method is called on each FileTarget.Write call.

I think we should improve this instead of optimizing the method? 

Also, I think we could optimize the Filetarget, but the bottlenecks are I/O bound I think.

epignosisx:
Can you expand on what you mean by improve? Do you refer to improve the entire FileTarget?

304NotModified:
> Do you refer to improve the entire FileTarget?

Yes, I meant that. But maybe not now a good moment as we are busy with porting to Xamarin/UWP/CoreCLR

Anyway, I will test this and probably merge it, as `CleanupInvalidFileNameChars` is really a performance eater (in CPU terms), that's what you already noticed I think. See https://github.com/NLog/NLog/issues/1071

So good work :)

Page-Not-Found:
@Alex just curious mind. What is the tool you are using?

epignosisx:
I hadn't seen that issue and although my change was mainly focused on reducing memory allocations, I expect better CPU utilization.

@Page-Not-Found I'm using JetBrains dotMemory

PR #1064, opened 2015-11-26
Reduce memory allocations in short date layout renderer

epignosisx:
Applying the same optimization used in `LongDateLayoutRenderer` to reduce the number of allocations in `ShortDateLayoutRenderer`.

Here are the numbers where it can be seen how the number of strings and char[] are reduced.

Before:
image

After:
image


304NotModified:
Thanks for this PR.

But I doubting if this is a good idea. The usage of `datetime.ToString` is of course very know and clear. It's also more flexible and less code = less bugs.

We can merge this, but only if:

- We unit test this new method. 
- Make it more readable: 
```
   builder.Append((char)((number / 10) + '0'));
   builder.Append((char)((number % 10) + '0'));
```
is hard to read (compared to `datetime.ToString`)
- The memory usage optimize is significant. (I'm doubting on this one also)

What do you think?

epignosisx:
I think every bit of performance improvements we can make is good for the library. Small wins here and there add up to something good.

It is true that datetime.ToString() is more readable, but considering:
1- The desired format is pretty straightforward to achieve in a more efficient way.
2- These allocations occur on each Log call
3- The optimization was already used before in the codebase, therefore it has been battle-tested.

I'll work on the unit tests and making those lines you mentioned more readable if you think it is worth the change.

304NotModified:
@epignosisx

I like the proposal. We have to make sure that we aren't doing premature optimizations, as they hurt more then benefit.

FTY, the code of `append` is straightforward, the code of `Append2DigitsZeroPadded`  should get some love. (maybe checking string.length instead of modulo?)

epignosisx:
I added the unit tests to verify the format is correct for 1-digit and 2-digit months and days.

I did some additional profiling on the int.ToString("D2") you suggested but it is not as efficient as the module approach. Here are the numbers:

D2 code:

image

D2 results:

image


-----------------------------

module results:

image

The D2 approach generates 200 more allocations.

Sample application used for profiling:

image

304NotModified:
Thanks for testing! 

I have difficulties with the code of `Append2DigitsZeroPadded` . Can we try: tostring, prepent 0 if length=1 (no string concst but two appends) ? That's at least a simple (and very imperative) way to code it.

epignosisx:
I tried what you suggested:

304NotModified:
And?

epignosisx:
I tried what you suggested but:

	private static void Append2DigitsZeroPadded(StringBuilder builder, int number)
	{
		if(number.ToString().Length == 1) //ToString allocates a string which takes a step backwards
			builder.Append('0');
		builder.Append(number);
	}

Then I tried this variation:

	private static void Append2DigitsZeroPadded(StringBuilder builder, int number)
	{
		if (number < 10)
			builder.Append('0');
		builder.Append(number);
	}

but it turned out to have the same number of allocations as the .ToString("D2") variation. The problem is that the line:

    builder.Append(number);

ends up calling ToString() on the number incurring in the additional memory allocation. Here is the source code for StringBuilder.Append(int value):

http://referencesource.microsoft.com/#mscorlib/system/text/stringbuilder.cs,945

In conclusion none of these variations have the same level of performance of the module approach.

304NotModified:
Thanks for testing! Those cases are also the ones I was thinking of. Great work! 

Well then the only solution to difficult code is to document it very well. I will merge it if we document the method and both lines. 

I also propose the rename the method to :` appendAsTwoDigits`,  because 2 is also used sometimes for "to", like nl2br.

304NotModified:
One last thing,  the short date won't change a lot (once a day) . What would happen if we cache the outcome?

epignosisx:
That would definitely make it allocation free most of the time!

    private DateTime cachedDate = DateTime.MinValue;
    private string cachedDateFormatted;

    {
        var ts = logEvent.TimeStamp;
        if (this.UniversalTime)
        {
            ts = ts.ToUniversalTime();
        }

        if (cachedDateFormatted == null || cachedDate.Year != ts.Year || cachedDate.Month != ts.Month || cachedDate.Day != ts.Day)
        {
            cachedDate = ts;
            cachedDateFormatted = ts.ToString("yyyy-MM-dd", CultureInfo.InvariantCulture);
        }

        builder.Append(cachedDateFormatted);
    }

Is it worth making it thread safe? That is, making safe reads and writes from/to the cached value. It could actually slow things down. The value only changes once a day, so it might be OK to not do any locking. What do you think?

304NotModified:
Well I'm not sure about making it thread safe. What ga possibility go wrong if we won't make it threadsafe? Only two things I think:

- we won't use in race conditions the cached value
- or we overwrite in race conditions unneeded the value. 

If those are the two, there is no need to make it threadsafe :)  (locks are indeed expensive) 

We need of course to check all the options (only universal time?) for the cache.

epignosisx:
I started working on it but I think we are making the code very difficult to read for such a simple renderer  from what it was originally. Caching, locking? I think the extra overhead will hurt performance more than the memory allocation.

I prefer to just leave it as is if you do not feel comfortable with the modulo solution.

304NotModified:
Well it should save some time. But It's fine to me to do (to quote myself):



> Well then the only solution to difficult code is to document it very well. I will merge it if we document the method and both lines.
> 
> I also propose the rename the method to :appendAsTwoDigits, because 2 is also used sometimes for "to", like nl2br.


In the future we can look after caching. I like the caching because (if no locking is used), it saves a lot of CPU (and some memory)

epignosisx:
Sorry, I was so excited about the caching idea I forgot your comment about the modulo solution. Since you think there is value in the extra complexity of the caching idea to save resources I'll give it a shot to the caching solution and we can refine it through the pull request.

304NotModified:
OK, I don't mind the extra complexitiy as I think it's limited. 

If we do the caching thing, it would be nice to benchmark it. I used before a small testprogram for that, see https://github.com/NLog/NLog/tree/1071-test-program-diag-tryouts (the test project is in NLog.netfx45.sln)

epignosisx:
I did the changes to cache the local and UTC dates as strings. Let me know what you think of the implementation. Here are some numbers I ran.

The number of allocations went down by 100 compared with the module solution:

image

Then I wrote this small program to check the performance of each solution:

image

Here are the numbers for each solution we have tried:

image

The cache solution is a clear winner.

I wanted to run your test program but the link you left takes me to the root of the code.

304NotModified:
Wohoo! Great results!

If you can remove the code duplication and add some docs WHY why are caching (including the measurements would be nice), I will merge it :)

304NotModified:
I think a git squach would be also nice. What do you think?

epignosisx:
I moved to a separate method the duplicate logic and added the docs.

In regard to the git squash, I haven't done it before and I was trying to do it but I kept getting stuck. Is this something you can do from your side?

304NotModified:
> In regard to the git squash, I haven't done it before and I was trying to do it but I kept getting stuck. Is this something you can do from your side?

Don't think so. But it's pretty easy with tortiosegit: http://stackoverflow.com/a/21306306/201303

epignosisx:
Couldn't figure out the git squash. I was getting merge conflicts and a bunch of other stuff. I'm going to create a fresh pull request with a clean history.

304NotModified:
superseded by https://github.com/NLog/NLog/pull/1131

PR #1066, opened 2015-11-30
Add AllLevels and AllLoggingLevels to LogLevel.cs.

rellis-of-rhindleton:
This is mostly for presentation purposes, which might be out of scope.  I just find it convenient with type safe enums like LogLevel to have a built-in enumerable of all possible values.  Makes it easier to create query filters etc. without having to hardcode anything.


304NotModified:
Nice idea! 

Could it be a `IList` instead of an array?

rellis-of-rhindleton:
For the private fields?  Anything that can be enumerated works for me.

304NotModified:
OK, I prefer (generic) `IList`. Arrays are legacy IMO

rellis-of-rhindleton:
Updated, if I followed you correctly.

304NotModified:
Nice, it has been merged :)

304NotModified:
Thanks!

PR #1068, opened 2015-11-30
Overhaul ExceptionLayoutRenderer

### LINKED ISSUE #1016: Overhaul ExceptionLayoutRenderer Methods
*Modify the Delegates to be func.

Delegate Definition / Signature:

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
private delegate void ExceptionDataTarget(StringBuilder sb, Exception ex);
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Methods which will be affected by this

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
AppendMessage
AppendMethod
AppendStackTrace
AppendToString
AppendType
AppendShortType
AppendData
CompileFormat
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Page-Not-Found:
* Added enums so that the API can set with enums instead of strings.
* Converted all delegates to action delegates.

fixes #1016

### DIFF
--- a/src/NLog/Config/ExceptionRenderingFormat.cs
+++ b/src/NLog/Config/ExceptionRenderingFormat.cs
@@ -45,14 +45,9 @@ namespace NLog.Config
     public enum ExceptionRenderingFormat
     {
         /// <summary>
-        /// This is for internal NLog Library usage. This is required we would a default value in switch cases. 
-        /// </summary>
-        NotSet = -1,
-        /// <summary>
         /// Appends the Message of an Exception to the specified target.
         /// </summary>
         Message = 0,
-
         /// <summary>
         /// Appends the type of an Exception to the specified target.
         /// </summary>
--- a/src/NLog/Config/ExceptionRenderingFormat.cs
+++ b/src/NLog/Config/ExceptionRenderingFormat.cs
@@ -1,4 +1,38 @@
-﻿using System;
+﻿// 
+// Copyright (c) 2004-2011 Jaroslaw Kowalski <jaak@jkowalski.net>
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without 
+// modification, are permitted provided that the following conditions 
+// are met:
+// 
+// * Redistributions of source code must retain the above copyright notice, 
+//   this list of conditions and the following disclaimer. 
+// 
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution. 
+// 
+// * Neither the name of Jaroslaw Kowalski nor the names of its 
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission. 
+// 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE 
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE 
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE 
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR 
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS 
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN 
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) 
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF 
+// THE POSSIBILITY OF SUCH DAMAGE.
+// 
+
+
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
--- a/src/NLog/Config/ExceptionRenderingFormat.cs
+++ b/src/NLog/Config/ExceptionRenderingFormat.cs
@@ -1,0 +1,48 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace NLog.Config
+{
+    /// <summary>
+    /// Format of the excpetion output to the specific target.
+    /// </summary>
+    public enum ExceptionRenderingFormat
+    {
+        /// <summary>
+        /// This is for internal NLog Library usage. This is required we would a default value in switch cases. 
+        /// </summary>
+        NotSet = -1,
+        /// <summary>
+        /// Appends the Message of an Exception to the specified target.
+        /// </summary>
+        Message = 0,
+
+        /// <summary>
+        /// Appends the type of an Exception to the specified target.
+        /// </summary>
+        Type = 1,
+        /// <summary>
+        /// Appends the short type of an Exception to the specified target.
+        /// </summary>
+        ShortType = 2,
+        /// <summary>
+        /// Appends the result of calling ToString() on an Exception to the specified target.
+        /// </summary>
+        ToString = 3,
+        /// <summary>
+        /// Appends the method name from Exception's stack trace to the specified target.
+        /// </summary>
+        Method = 4,
+        /// <summary>
+        /// Appends the stack trace from an Exception to the specified target.
+        /// </summary>
+        StackTrace = 5,
+        /// <summary>
+        /// Appends the contents of an Exception's Data property to the specified target.
+        /// </summary>
+        Data = 6
+    }
+}
--- a/src/NLog/Config/ExceptionRenderingFormat.cs
+++ b/src/NLog/Config/ExceptionRenderingFormat.cs
@@ -2,7 +2,6 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
-using System.Threading.Tasks;
 
 namespace NLog.Config
 {
--- a/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
@@ -52,7 +52,8 @@ namespace NLog.LayoutRenderers
     {
         private string format;
         private string innerFormat = string.Empty;
-        private static readonly Dictionary<ExceptionRenderingFormat, Action<StringBuilder, Exception>> _functions = new Dictionary<ExceptionRenderingFormat, Action<StringBuilder, Exception>>() {
+        private static readonly Dictionary<ExceptionRenderingFormat, Action<StringBuilder, Exception>> _renderingfunctions = new Dictionary<ExceptionRenderingFormat, Action<StringBuilder, Exception>>() 
+                                                                                                    {
                                                                                                         {ExceptionRenderingFormat.Message, AppendMessage}, 
                                                                                                         {ExceptionRenderingFormat.Type, AppendType},
                                                                                                         {ExceptionRenderingFormat.ShortType, AppendShortType},
@@ -60,6 +61,17 @@ namespace NLog.LayoutRenderers
                                                                                                         {ExceptionRenderingFormat.Method, AppendMethod},
                                                                                                         {ExceptionRenderingFormat.StackTrace, AppendStackTrace},
                                                                                                         {ExceptionRenderingFormat.Data, AppendData}
+                                                                                                    };
+
+        private static readonly Dictionary<String, ExceptionRenderingFormat> _formatsMapping = new Dictionary<string, ExceptionRenderingFormat>(StringComparer.OrdinalIgnoreCase)
+                                                                                                    {
+                                                                                                        {"MESSAGE",ExceptionRenderingFormat.Message}, 
+                                                                                                        {"TYPE", ExceptionRenderingFormat.Type}, 
+                                                                                                        {"SHORTTYPE",ExceptionRenderingFormat.ShortType}, 
+                                                                                                        {"TOSTRING",ExceptionRenderingFormat.ToString}, 
+                                                                                                        {"METHOD",ExceptionRenderingFormat.Method}, 
+                                                                                                        {"STACKTRACE", ExceptionRenderingFormat.StackTrace}, 
+                                                                                                        {"DATA",ExceptionRenderingFormat.Data}, 
                                                                                                     };
 
         /// <summary>
@@ -175,7 +187,7 @@ namespace NLog.LayoutRenderers
                 {
                     sb2.Append(separator);
 
-                    var currentRenderFunction = _functions[renderingFormat];
+                    var currentRenderFunction = _renderingfunctions[renderingFormat];
 
                     currentRenderFunction(sb2, logEvent.Exception);
 
@@ -193,7 +205,7 @@ namespace NLog.LayoutRenderers
                     {
                         sb2.Append(separator);
 
-                        var currentRenderFunction = _functions[renderingFormat];
+                        var currentRenderFunction = _renderingfunctions[renderingFormat];
 
                         currentRenderFunction(sb2, currentException);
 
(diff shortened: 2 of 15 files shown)


304NotModified:
This weekend I will review it.

304NotModified:
fixes https://github.com/NLog/NLog/issues/1016

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -142,13 +167,13 @@ protected override void Append(StringBuilder builder, LogEventInfo logEvent)
...
                 string separator = string.Empty;
 
-                foreach (ExceptionDataTarget targetRenderFunc in this.exceptionDataTargets)
+                foreach (ExceptionRenderingFormat renderingFormat in this.Formats)
                 {
                     sb2.Append(separator);
-                    targetRenderFunc(sb2, logEvent.Exception);
+                    _functions[renderingFormat](sb2, logEvent.Exception);
+
                     separator = this.Separator;
```
304NotModified: this is a bit strange, do you agree? (assignment can be outside the loop)

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -142,13 +167,13 @@ protected override void Append(StringBuilder builder, LogEventInfo logEvent)
                 var sb2 = new StringBuilder(128);
                 string separator = string.Empty;
 
-                foreach (ExceptionDataTarget targetRenderFunc in this.exceptionDataTargets)
+                foreach (ExceptionRenderingFormat renderingFormat in this.Formats)
                 {
                     sb2.Append(separator);
-                    targetRenderFunc(sb2, logEvent.Exception);
+                    _functions[renderingFormat](sb2, logEvent.Exception);
```
304NotModified: please create a variable of `  _functions[renderingFormat]` (just for readability).

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -157,10 +182,10 @@ protected override void Append(StringBuilder builder, LogEventInfo logEvent)
                     sb2.Append(this.InnerExceptionSeparator);
 
                     separator = string.Empty;
-                    foreach (ExceptionDataTarget targetRenderFunc in this.innerExceptionDataTargets ?? this.exceptionDataTargets)
+                    foreach (ExceptionRenderingFormat renderingFormat in this.InnerFormats ?? this.Formats)
                     {
                         sb2.Append(separator);
-                        targetRenderFunc(sb2, currentException);
+                        _functions[renderingFormat](sb2, currentException);
```
304NotModified: idem

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -157,10 +182,10 @@ protected override void Append(StringBuilder builder, LogEventInfo logEvent)
...
                         separator = this.Separator;
```
304NotModified: idem

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -52,9 +52,16 @@ public class ExceptionLayoutRenderer : LayoutRenderer
     {
         private string format;
         private string innerFormat = string.Empty;
-        private ExceptionDataTarget[] exceptionDataTargets;
-        private ExceptionDataTarget[] innerExceptionDataTargets;
-        
+        private Dictionary<ExceptionRenderingFormat, Action<StringBuilder, Exception>> _functions = new Dictionary<ExceptionRenderingFormat, Action<StringBuilder, Exception>>() {
```
304NotModified: please add `readonly`

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -268,52 +293,54 @@ protected virtual void AppendData(StringBuilder sb, Exception ex)
...
         {
+            List<ExceptionRenderingFormat> formats = new List<ExceptionRenderingFormat>();
             string[] parts = formatSpecifier.Replace(" ", string.Empty).Split(',');
-            var dataTargets = new List<ExceptionDataTarget>();
 
             foreach (string s in parts)
             {
+                ExceptionRenderingFormat renderingFormat;
+
                 switch (s.ToUpper(CultureInfo.InvariantCulture))
```
304NotModified: I think it's still nice to rewrite the `swich` to a static dictionary. We have discussed this before I think, but we can write an overload. 

the dictionary is better for the code complexity (arcording to Cyclomatic complexity) and so easier to mainain.

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -268,52 +293,54 @@ protected virtual void AppendData(StringBuilder sb, Exception ex)
...
                         break;
-
                     case "DATA":
-                        dataTargets.Add(AppendData);
+                        renderingFormat = ExceptionRenderingFormat.Data;
                         break;
-
                     default:
                         InternalLogger.Warn("Unknown exception data target: {0}", s);
+                        renderingFormat = ExceptionRenderingFormat.NotSet;
```
304NotModified: Proposal: remove `NotSet` and change `ExceptionRenderingFormat renderingFormat` to `ExceptionRenderingFormat? renderingFormat`

src/NLog/Config/ExceptionRenderingFormat.cs
```diff
@@ -0,0 +1,81 @@
...
+{
+    /// <summary>
+    /// Format of the excpetion output to the specific target.
+    /// </summary>
+    public enum ExceptionRenderingFormat
+    {
+        /// <summary>
+        /// This is for internal NLog Library usage. This is required we would a default value in switch cases. 
+        /// </summary>
+        NotSet = -1,
```
304NotModified: This is not needed :)

See notes at the usage of `NotSet`. 

This one should be removed.

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -268,52 +293,54 @@ protected virtual void AppendData(StringBuilder sb, Exception ex)
...
                         break;
                 }
-            }
+                if (renderingFormat != ExceptionRenderingFormat.NotSet)
```
304NotModified: this will be a null check

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -268,52 +293,54 @@ protected virtual void AppendData(StringBuilder sb, Exception ex)
             }
         }
 
-        private ExceptionDataTarget[] CompileFormat(string formatSpecifier)
+        /// <summary>
+        /// Split the string and then compile into list of Rendering formats.
+        /// </summary>
+        /// <param name="formatSpecifier"></param>
+        /// <returns></returns>
+        private List<ExceptionRenderingFormat> CompileFormat(string formatSpecifier)
```
304NotModified: please add `static`

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -66,8 +73,6 @@ public ExceptionLayoutRenderer()
             this.MaxInnerExceptionLevel = 0;
         }
 
-        private delegate void ExceptionDataTarget(StringBuilder sb, Exception ex);
-
         /// <summary>
         /// Gets or sets the format of the output. Must be a comma-separated list of exception
```
304NotModified: please refer to Formats + ExceptionRenderingFormat

with
`<see cref="Formats"/>` and `<see cref="ExceptionRenderingFormat"/>`

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -85,7 +90,7 @@ public string Format
             set
```
304NotModified: same, refer to InnerFormats + ExceptionRenderingFormat

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -131,6 +136,26 @@ public string InnerFormat
         public string InnerExceptionSeparator { get; set; }
 
         /// <summary>
+        ///  Gets the formats of the output of inner exceptions to be rendered in target.
```
304NotModified: please ref to `Format` with the `<see` tag.

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -131,6 +136,26 @@ public string InnerFormat
...
+        {
+            get;
+            private set;
+        }
+
+        /// <summary>
+        ///  Gets the formats of the output to be rendered in target.
+        /// </summary>
+        /// <docgen category='Rendering Options' order='10' />
+        public List<ExceptionRenderingFormat> InnerFormats
```
304NotModified: please ref to `InnerFormat` with the `<see` tag.

304NotModified:
hey @Page-Not-Found 

I reviewed the code. It looks really good, I added some notes for docs and some code improvements.
I'm happy we can remove the `NotSet` enum value. :)

Let me know what you think. I still think the switch to dictionary is a great improvement (I use it always for "conversion", but let me know if you think otherwise.

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
Page-Not-Found: For the first level the separator is not required and so the separator is assigned after the first loop.

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
Page-Not-Found: I am not sure I understand you. But the current code is with the Dictionary.

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
Page-Not-Found: What does this mean?

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
Page-Not-Found: Is this one you are looking for?

~~~~~~~~~~~~~~~~~~~~~
/// <summary>
        /// Gets or sets the format of the output. Must be a comma-separated list of exception
        /// properties: Message, Type, ShortType, ToString, Method, StackTrace.
        /// This parameter value is case-insensitive.
        /// </summary>
        /// <see cref="Formats"/>
        /// <see cref="ExceptionRenderingFormat"/>
        [DefaultParameter]
        public string Format
        {
            get
            {
                return this.format;
            }

            set
            {
                this.format = value;
                Formats = CompileFormat(value);
            }
        }

Page-Not-Found:
>Let me know what you think. I still think the switch to dictionary is a great improvement (I use it always for conversions) but let me know if you think otherwise.

I think the initial format to list conversion is currently in the switch. Do you mean that needs to be changed?

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
private static List<ExceptionRenderingFormat> CompileFormat(string formatSpecifier)
        {
            List<ExceptionRenderingFormat> formats = new List<ExceptionRenderingFormat>();
            string[] parts = formatSpecifier.Replace(" ", string.Empty).Split(',');

            foreach (string s in parts)
            {
                ExceptionRenderingFormat? renderingFormat = null;

                switch (s.ToUpper(CultureInfo.InvariantCulture))
                {
                    case "MESSAGE":
                        renderingFormat = ExceptionRenderingFormat.Message;
                        break;
                    case "TYPE":
                        renderingFormat = ExceptionRenderingFormat.Type;
                        break;
                    case "SHORTTYPE":
                        renderingFormat = ExceptionRenderingFormat.ShortType;
                        break;
                    case "TOSTRING":
                        renderingFormat = ExceptionRenderingFormat.ToString;
                        break;
                    case "METHOD":
                        renderingFormat = ExceptionRenderingFormat.Method;
                        break;
                    case "STACKTRACE":
                        renderingFormat = ExceptionRenderingFormat.StackTrace;
                        break;
                    case "DATA":
                        renderingFormat = ExceptionRenderingFormat.Data;
                        break;
                    default:
                        InternalLogger.Warn("Unknown exception data target: {0}", s);                        
                        break;
                }
                if (renderingFormat != null)
                    formats.Add(renderingFormat.Value);

            }
            return formats;
        }
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

304NotModified:
Yes,  that would be a great improvement!

Page-Not-Found:
I will write up a proposed code change and put up here again.

Page-Not-Found:
Here you go with a sample code
~~~~~~~~~~~~~~~~~~~~~~~~~~~~
private static readonly Dictionary<String, ExceptionRenderingFormat> _formatsMapping = new Dictionary<string, ExceptionRenderingFormat>
{
	{"MESSAGE",ExceptionRenderingFormat.Message}, 
	{"TYPE", ExceptionRenderingFormat.Type}, 
	{"SHORTTYPE",ExceptionRenderingFormat.ShortType}, 
	{"TOSTRING",ExceptionRenderingFormat.ToString}, 
	{"METHOD",ExceptionRenderingFormat.Method}, 
	{"STACKTRACE", ExceptionRenderingFormat.StackTrace}, 
	{"DATA",ExceptionRenderingFormat.Data}, 
};

/// <summary>
        /// Split the string and then compile into list of Rendering formats.
        /// </summary>
        /// <param name="formatSpecifier"></param>
        /// <returns></returns>
        private static List<ExceptionRenderingFormat> CompileFormat(string formatSpecifier)
        {
            List<ExceptionRenderingFormat> formats = new List<ExceptionRenderingFormat>();
            string[] parts = formatSpecifier.Replace(" ", string.Empty).Split(',');

            foreach (string s in parts)
            {
                ExceptionRenderingFormat? renderingFormat = null;
                if (_formatsMapping.ContainsKey(s))
                {
                    renderingFormat = _formatsMapping[s];
                    formats.Add(renderingFormat.Value);
                }
                else
                {
                    InternalLogger.Warn("Unknown exception data target: {0}", s);
                }

            }
            return formats;
        }

304NotModified:
Yes! that was what I really like.

some last improvements:

1. pass `StringComparer.OrdinalIgnoreCase` to the contructor of dictionary:
 `_formatsMapping = new Dictionary<string, ExceptionRenderingFormat>(StringComparer.OrdinalIgnoreCase)`

2. `formatSpecifier.Replace(" ", string.Empty).Split(',')`, is this `  formatSpecifier.Split('.', StringSplitOptions.RemoveEmptyEntries)` ?

3. use ` dict.TryGetValue()` instead if `Contains` and `[..]` - that one is 2 times faster and more logical here.

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -176,8 +224,8 @@ protected override void Append(StringBuilder builder, LogEventInfo logEvent)
         /// Appends the Message of an Exception to the specified <see cref="StringBuilder" />.
         /// </summary>
         /// <param name="sb">The <see cref="StringBuilder"/> to append the rendered data to.</param>
-        /// <param name="ex">The exception containing the Message to append.</param>
-        protected virtual void AppendMessage(StringBuilder sb, Exception ex)
+        /// <param name="ex">The exception containing the Message to append.</param>        
+        private static void AppendMessage(StringBuilder sb, Exception ex)
```
304NotModified: Why move to private? This isnt backwards-compatible now?

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -268,52 +316,30 @@ protected virtual void AppendData(StringBuilder sb, Exception ex)
...
         {
-            string[] parts = formatSpecifier.Replace(" ", string.Empty).Split(',');
-            var dataTargets = new List<ExceptionDataTarget>();
+            List<ExceptionRenderingFormat> formats = new List<ExceptionRenderingFormat>();
+            string[] parts = formatSpecifier.Replace(" ", string.Empty).Split(new string[] { "," }, StringSplitOptions.RemoveEmptyEntries);
 
             foreach (string s in parts)
             {
-                switch (s.ToUpper(CultureInfo.InvariantCulture))
+                ExceptionRenderingFormat renderingFormat;
```
304NotModified: Really nice :)

304NotModified:
I think this is ready to merge? 

I added one backwards-compatibly question. That's the last thing.

Page-Not-Found:
I am with limited internet access. Will get back to this once I am back.

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
Page-Not-Found: For Adding this into the Dictionary we have to move this to Private. We cannot maintain both the instance and static methods also. 

The other solution which I could think of is making a internal static class and mapping them together.

304NotModified: I get your point. With the virtual protected method (e.g. AppendMessage), you can do two things:

1. replace behavior(virtual).
2. call the method myself. This is the private/protected. 

I don't know how an `Action` will work with `virtual`, but I think this doens't work? So 
1 and 2 are now breaking changes. 2 is easy to fix, but 1 is tricky. I think you warned me before about the breaking change of 1, but I missed to see it :|

Any thoughts? Breaking changes aren't nice.

Page-Not-Found: As I said earlier 
>The other solution which I could think of is making a internal static class and having the new methods in them and call them from the virtual methods. What do you say?

304NotModified: Oops. Missed that one on mobile. 

Do you think it's a good solution?

Page-Not-Found: >Do you think it's a good solution?
Atleast for now and them remove the virtual methods in upcoming releases.

304NotModified: > Atleast for now and them remove the virtual methods in upcoming releases.

Doesn't sound future proof to me. Well, mabye was the dictionary (my) mistake and not a good step?

Page-Not-Found: I will see if I can fine someother solution.

Page-Not-Found: There are two options now

1. Remove the Dict (`_renderingfunctions`) which will fall back to the loop and the switch case.
2. Change all the methods back to virtual method and make the dictionary as INSTANCE type rather than a static variable. 

When I try to change the method to public virtual, the dictionary isn't allowing. Which is understandable. I see the below error

`A field initializer cannot reference the non-static field, method, or property 'NLog.LayoutRenderers.ExceptionLayoutRenderer.AppendMessage(System.Text.StringBuilder, System.Exception)'`

304NotModified: I think 2 won't work with virtual? Action and virtual won't work together? (method override in sub class, maybe unit test it?)  If it does, I prefer 2 and otherwise we should do 1.

Page-Not-Found:
@304NotModified  I was able to change the methods back to virtual and override them as well. Added one more commit to the PR. Let me know what you feel.

tests/NLog.UnitTests/LayoutRenderers/ExceptionTests.cs
```diff
@@ -562,5 +563,135 @@ private class GenericClass<TA, TB, TC>
...
+
+
+            Assert.Equal(ExceptionRenderingFormat.ShortType, elr.Formats[0]);
+            Assert.Equal(ExceptionRenderingFormat.Message, elr.Formats[1]);
+
+            Assert.Equal(ExceptionRenderingFormat.Message, elr.InnerFormats[0]);
+        }
+
+        [Fact]
+        public void CustomExceptionLayoutRendrerInnerExceptionTest()
```
304NotModified: Thanks for this test :D

304NotModified:
Cool! Merged it finally! 

Very happy with the end result. It's a big improvement :)

PR #1084, opened 2015-12-09
Log unused targets to internal logger

### LINKED ISSUE #1074: Log unused targets to internal logger
When using xml configuration, (maybe also non xml),  log a warning or info for targets that aren't used in the rules.  

This is useful for detecting miscongurations, like forgetting the asyncwrapper. See https://github.com/NLog/NLog/issues/1073, https://github.com/NLog/NLog/issues/1071 and https://github.com/NLog/NLog/issues/865

We should do this efficient, so use a dictionary or hashset.

UgurAldanmaz:
Closes #1074


src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -344,6 +348,55 @@ private void Initialize(XmlReader reader, string fileName, bool ignoreErrors)
...
+                return;
+            }
+
+            var swatch = Stopwatch.StartNew();
+
+            ReadOnlyCollection<Target> configuredNamedTargets = this.ConfiguredNamedTargets; //assign to variable. because it is returning new ReadOnlyCollection everytime.
+            InternalLogger.Warn("Unused target checking is started... Rule Count: {0}, Target Count: {1}", this.LoggingRules.Count, configuredNamedTargets.Count);
+
+            HashSet<string> targetNamesAtRules = new HashSet<string>();
+            for (int ri = 0; ri < this.LoggingRules.Count; ri++)
```
304NotModified: Looks good, but we can also use easily LINQ here?

Something like (code not checked)

```c#
 HashSet<string> targetNamesAtRules = new HashSet<string>(this.LoggingRules.SelectMany(r => r..Target).Select(t => t.Name));
```

What do you think?

src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -629,7 +682,7 @@ private void ParseTargetsElement(NLogXmlElement targetsElement)
             foreach (var targetElement in targetsElement.Children)
             {
                 string name = targetElement.LocalName;
-                string type = StripOptionalNamespacePrefix(targetElement.GetOptionalAttribute("type", null));
+                string typeAttributeVal = StripOptionalNamespacePrefix(targetElement.GetOptionalAttribute("type", null));
```
304NotModified: good renames :+1:

304NotModified:
Thanks! Looks really nice!

PS: AppVeyor complains about stopwatch. As the current datetime is logged with the internallogger, I think stopwatch isn't needed.

src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -344,6 +348,55 @@ private void Initialize(XmlReader reader, string fileName, bool ignoreErrors)
...
+        /// <remarks>If initializing not started or failed, then checking process will be canceled</remarks>
+        /// </summary>
+        private void CheckUnusedTargets()
+        {
+            if (this.InitializeSucceeded == null)
+            {
+                InternalLogger.Warn("Unused target checking is canceled -> initialize not started yet.");
+                return;
+            }
+            if (this.InitializeSucceeded.Value == false)
```
304NotModified: We prefer `if (!this.InitializeSucceeded.Value)`. (just code style)

src/NLog/Config/XmlLoggingConfiguration.cs
UgurAldanmaz: It is possible. Maybe I was a bit obsessive about performance, therefore just use classic ways.

src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -344,6 +348,55 @@ private void Initialize(XmlReader reader, string fileName, bool ignoreErrors)
...
+                Target target = configuredNamedTargets[i];
+                if (!targetNamesAtRules.Contains(target.Name))
+                {
+                    InternalLogger.Warn("Unused target detected. Add a rule for this target to the configuration. TargetName: {0}", target.Name);
+                    unusedCount++;
+                }
+            }
+
+            swatch.Stop();
+            InternalLogger.Warn("Unused target checking is completed ({0}). Total Rule Count: {1}, Total Target Count: {2}, UnusedTarget Count: {3}", swatch.Elapsed, this.LoggingRules.Count, configuredNamedTargets.Count, unusedCount);
```
304NotModified: This should be "debug" instead of "warn".

src/NLog/Config/XmlLoggingConfiguration.cs
304NotModified: Yes it's true that the `for` is somewhat better in performance. But luckily this code is only run when the config changes.

304NotModified: You can measure the difference with your stopwatch ;)

UgurAldanmaz: ahaha. Unfortunately I removed my stopwatch :P

OK. Linq way would be more readable.

UgurAldanmaz: To support under version of .Net 3.5, we should not use LINQ. Also [Mono] (http://www.mono-project.com/docs/about-mono/compatibility/).

304NotModified: Hey! Linq is supported, we already use it (and I cannot live without it, I like functional programming) 

Are you confused with linq-to-sql?

UgurAldanmaz: No, I am not confused with linq-to-sql.

To be sure whether Linq will broke something, I searched which version has support. Maybe my conclusion is wrong. But I could not found any result for supporting older than .Net 3.5 ([without extra effort] (http://stackoverflow.com/questions/3348348/is-there-a-way-to-use-linq-query-syntax-with-net-3-0-projects))
[At least Linq MSDN page has not older version] (https://msdn.microsoft.com/en-us/library/system.linq(v=vs.90).aspx).

304NotModified:
Merged! Thanks! :+1:

PR #1089, opened 2015-12-12
Support For Relative Paths in the File Targets

### LINKED ISSUE #1045: Handle relative path, or at least warn when detecting one
When using a relative path in the config as follow : 

    <target xsi:type="File" fileName="Logs\myapp.${shortdate}.log" name="file"
            layout="${longdate}|${level:uppercase=true}|${logger}|${message}" />

The expected behavior from the user point of view is always to log in a folder relative to where the application executes.

This can be easily detected and either pro-actively understand that it is a relative path
or emit a warning in the internal log file.

Page-Not-Found:
Support For Relative Paths in the File Targets

Fixes #1045

- [x] fileName
- [x] archiveFileName


tests/NLog.UnitTests/Targets/FileTargetTests.cs
```diff
@@ -52,6 +52,7 @@ namespace NLog.UnitTests.Targets
     using NLog.Internal;
     using NLog.LayoutRenderers;
 
+    [System.Runtime.InteropServices.GuidAttribute("FF35CCD9-C7C7-4E3F-9194-3C3076DA3B7C")]
```
304NotModified: Why is this needed?

304NotModified:
Thanks! 

One small comment. Also some questions:

- are we sure these are the only usages of the filename? It was thinking there were more cases in the filetarget 
- is this 100% backwards compatible? 
- is the `${basedir}` renderer now useless with this change.

tests/NLog.UnitTests/Targets/FileTargetTests.cs
Page-Not-Found: I guess this one got added automatically. Will Remove that.

Page-Not-Found:
One small comment. Also some questions:

>are we sure these are the only usages of the filename? It was thinking there were more cases in the filetarget

I searched for the references of the "FileName" property and found only these two places. If you feel I might be overseeing things. Let me know we can work through that as well.

>is this 100% backwards compatible?

Not Sure though, I ran all the tests and passed successfully

>is the ${basedir} renderer now useless with this change?

Not really, you can have **${basedir}** as a relative path still.

304NotModified:
> Not really, you can have ${basedir} as a relative path still.

Do you can think of a case (in the future) when we still need the `${basedir}`?

Page-Not-Found:
>Do you can think of a case (in the future) when we still need the ${basedir}?

Not Really, but that can be possible in future.

304NotModified:
I cannot image that usage of `${basedir}`  is not unneeded, as we would never made it in the first case.

I will think about this, and will check git history. 

If someone has a case, let me know :)

304NotModified:
`${basedir}` = The current application domain's base and uses `IAppDomain` and `AppDomainWrapper.CurrentDomain`

304NotModified:
We can of course use `${basedir}` outside the filename

bhaeussermann:
Does this cater for relative paths in `ArchiveFileName` as well? I think it should.

Page-Not-Found:
No as of now it does not cover the archive file name. If needed I can take
a look at that as well.

304NotModified:
I think that's agood idea @bhaeussermann ! 

I added a checklist to the issue. 

If we add that, I will merge this one :)

Page-Not-Found:
@304NotModified I add more Commits.

304NotModified:
Thanks! Looks good! If you could fix the conflict, I will merge it. 

More unit tests added then changed lines. I like that :)

Page-Not-Found:
Fixes #1045 and #1120

304NotModified:
To it has been merged! Thanks!

Issue #550, opened 2015-02-02
Archive files are never created when there are lot of write in log file


jforestier:
When i write informations in logfile with interval of 10ms, the archive files are never created.

I've configure my nlog to archive every minute. So, 1 archive for each minute.
If i write with 100ms interval no problem.

I've test this for 20 minutes with 10ms and 100ms interval.

Below informations of my configuration :

ArchiveEvery = FileArchivePeriod.Minute;
ArchiveFileName = Path.Combine(logDirectory, ArchiveLogFilename);
ArchiveNumbering = ArchiveNumberingMode.Date;
ArchiveDateFormat = "yyyy-MM-dd hh-mm";
MaxArchiveFiles = 90;
KeepFileOpen = false;
Encoding = System.Text.Encoding.UTF8;

Xharze:
@jforestier Could you provide a sample project for this issue?

ilya-g:
NLog relies on file last write time to determine whether it should archive the file: if the logEvent's timestamp is from next period (i.e. minute) and file last write time is from previous period, it decides to archive file.

Technically, it is possible that a file write begin in one period and might have completed in next period. In this case NLog, writing event from the next period, would not catch the difference between event's timestamp and last write time, and, as a consequence, would not archive file.

304NotModified:
@ilya-g do you think that we can fix this?

madsmg:
We have been experiencing this issue quite a lot (log writes that overlaps into the next period), resulting in missed archiving. Currently we have implemented a workaround using a custom target.

One possible solution (however I have a hard time predicting all the consequences) could be to call File.SetCreationTimeUtc when creating the filestream in the base appender (to ensure that the creation time is reliable), and instead of LastWriteTimeUtc, use the CreationTimeUtc to determine when it is time for archival.

I am a complete newcomer in the open source world, but if you think the outlined solution is valid I would be happy to create a pull request.

304NotModified:
That's sounds like a good idea! I think it would be great if we can fix this issue! 

PR is accepted on this one.  If you need some insight info, just let us know!

PR #1090, opened 2015-12-14
Issue #550 - Use file creation time instead of last write time

madsmg:
FileTarget modified to use file creation time instead of last write time for archival. This is a proposed solution for Issue #550. 

I am not sure if the solution proposed in BaseFileAppender (UpdateCreationTime method) is the most elegant solution, however if the file doesn't exist, I have to create it to be able to set the creation time. If the filestream is opened by TryCreateFileStream, it's not always possible to set the creation time due to FileShare settings.

### DIFF
--- a/src/NLog/Internal/FileAppenders/BaseFileAppender.cs
+++ b/src/NLog/Internal/FileAppenders/BaseFileAppender.cs
@@ -61,7 +61,7 @@ namespace NLog.Internal.FileAppenders
             this.CreateFileParameters = createParameters;
             this.FileName = fileName;
             this.OpenTime = DateTime.UtcNow; // to be consistent with timeToKill in FileTarget.AutoClosingTimerCallback
-            this.LastWriteTime = DateTime.MinValue;
+            this.CreationTime = DateTime.MinValue;
         }
 
         /// <summary>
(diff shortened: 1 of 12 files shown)


src/NLog/Internal/FileAppenders/BaseFileAppender.cs
```diff
@@ -284,5 +271,22 @@ private FileStream TryCreateFileStream(bool allowFileSharedWriting)
...
+        /// <summary>
+        /// If the file exists retrieve the creation time, otherwise create a new log file and set the creation time.
+        /// </summary>
+        private void UpdateCreationTime()
+        {
+           if( !File.Exists(this.FileName) )
+           {
+              this.CreationTime = DateTime.UtcNow;
+              File.Create(this.FileName).Dispose();
+              File.SetCreationTimeUtc(this.FileName, this.CreationTime);
```
UgurAldanmaz: #if directive for Silverlight is required at this line.

304NotModified:
Thanks! We will look at this in more detail soon!

304NotModified:
Hey! Thanks for helping us on this one. This PR got superseded by #1100, but it helped is in fixing the issue. 

We will release it in 4.3.

PR #1100, opened 2015-12-19
FileTarget: Fix archive based on time does not always archive.

### LINKED ISSUE #550: Archive files are never created when there are lot of write in log file

### LINKED ISSUE #859: File ArchiveEvery Day/Minute... is ignored when using rolling numbering
Noticed another possible bug when trying to get around the issue with the wrong archive files getting deleted.

With version 4.0.1 using the following code the log archiving function ignores the rollover from one day to another and continues to write the new log messages into the file started the other day.

    fileTarget.FileName = updateLogDirectory + logName + ".log";
    fileTarget.ArchiveFileName = updateLogDirectory + logName + ".{##}.log";
    fileTarget.KeepFileOpen = false;
    fileTarget.ArchiveNumbering = ArchiveNumberingMode.Rolling;
    fileTarget.ArchiveEvery = FileArchivePeriod.Day;
    fileTarget.MaxArchiveFiles = 30; 
    fileTarget.ArchiveAboveSize = 1048576; // 1 MB
    fileTarget.Layout = "${longdate}|${message}";

### LINKED ISSUE #1022: Improve mocking of file target (current date)
There are/were some bugs with file archiving which are difficult to (unit)  test. Most of the time this has to do with the usage of current date(times) . 

Proposal : inject "current"  date into file target for better and more robust testing. 

Related: #859, #888, #964.

bhaeussermann:
Fixes #1022,fix #859,  fix #550

### DIFF
--- a/src/NLog/Internal/FileAppenders/FileAppenderCache.cs
+++ b/src/NLog/Internal/FileAppenders/FileAppenderCache.cs
@@ -239,26 +239,16 @@ namespace NLog.Internal.FileAppenders
         /// Gets the file info for a particular appender.
         /// </summary>
         /// <param name="fileName">The file name associated with a particular appender.</param>
-        /// <param name="fileCharacteristics">The file characteristics, if the file information was retrieved successfully.</param>
-        /// <returns><see langword="true"/> when the operation succeeded; <see langword="false"/> otherwise.</returns>
-        public bool GetFileCharacteristics(string fileName, out FileCharacteristics fileCharacteristics)
+        /// <returns>The file characteristics, if the file information was retrieved successfully, otherwise null.</returns>
+        public FileCharacteristics GetFileCharacteristics(string fileName)
         {
             foreach (BaseFileAppender appender in appenders)
             {
-                if (appender == null)
-                {
-                    break;
-                }
-
-                if (appender.FileName == fileName)
-                {
-                    appender.GetFileCharacteristics(out fileCharacteristics);
-                    return true;
-                }
+                if ((appender != null) && (appender.FileName == fileName))
+                    return appender.GetFileCharacteristics();
             }
-            
-            fileCharacteristics = null;
-            return false;
+
+            return null;
         }
 
         /// <summary>
--- a/src/NLog/Internal/FileAppenders/FileAppenderCache.cs
+++ b/src/NLog/Internal/FileAppenders/FileAppenderCache.cs
@@ -244,7 +244,10 @@ namespace NLog.Internal.FileAppenders
         {
             foreach (BaseFileAppender appender in appenders)
             {
-                if ((appender != null) && (appender.FileName == fileName))
+                if (appender == null)
+                    break;
+
+                if (appender.FileName == fileName)
                     return appender.GetFileCharacteristics();
             }
 
--- a/src/NLog/Internal/FileAppenders/FileAppenderCache.cs
+++ b/src/NLog/Internal/FileAppenders/FileAppenderCache.cs
@@ -239,9 +239,9 @@ namespace NLog.Internal.FileAppenders
         /// Gets the file info for a particular appender.
         /// </summary>
         /// <param name="fileName">The file name associated with a particular appender.</param>
-        /// <param name="fileInfo">The file info, if the file information was retrieved successfully.</param>
+        /// <param name="fileCharacteristics">The file characteristics, if the file information was retrieved successfully.</param>
         /// <returns><see langword="true"/> when the operation succeeded; <see langword="false"/> otherwise.</returns>
-        public bool GetFileInfo(string fileName, out FileInfo fileInfo)
+        public bool GetFileCharacteristics(string fileName, out FileCharacteristics fileCharacteristics)
         {
             foreach (BaseFileAppender appender in appenders)
             {
@@ -252,12 +252,12 @@ namespace NLog.Internal.FileAppenders
 
                 if (appender.FileName == fileName)
                 {
-                    appender.GetFileInfo(out fileInfo);
+                    appender.GetFileCharacteristics(out fileCharacteristics);
                     return true;
                 }
             }
             
-            fileInfo = null;
+            fileCharacteristics = null;
             return false;
         }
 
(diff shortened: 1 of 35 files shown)


304NotModified:
Thanks! Is this the same as https://github.com/NLog/NLog/pull/1090

bhaeussermann:
> Thanks! Is this (partly) the same as #1090? 

Yes. It's basically the same, except I retained the `lastWriteTime` parameter of the `GetFileInfo()` method. I'm not convinced that it is safe to use `creationTime` instead of `lastWriteTime` in `IsDaySwitch()`, but I will investigate.

304NotModified:
Well we can now unit test it,  isn't it? 

Thanks for checking :)

304NotModified:
> I'm not convinced that it is correct to use creationTime instead of lastWriteTime in IsDaySwitch() (as in #1090), but I will investigate.

I think creationTime is better, but please check. 

Can you also check the mono error? Thanks!

bhaeussermann:
> I think creationTime is better, but please check. 

It seems you are right, and I supplemented the tests to enforce this.

And with that, it seems that we never use the `lastWriteTime` parameter of the `GetFileInfo()` method. Do you think we should remove this parameter (as in #1090), or keep it for now in case we need it again?

I'm also thinking about refactoring the `GetFileInfo()` method by replacing all the `out` parameters with a single `out FileInfo` parameter (`FileInfo` being a new class). This should decrease the number of changes when we need to change what file info the `GetFileInfo()` method returns. What do you think?

304NotModified:
> It seems you are right, and I supplemented the tests to enforce this.

Thanks!

> And with that, it seems that we never use the lastWriteTime parameter of the GetFileInfo() method. Do you think we should remove this parameter (as in #1090), or keep it for now in case we need it again?

Remove it. Less is more (and dead code which isn't clearly dead, is a source of confusion). 

> I'm also thinking about refactoring the GetFileInfo() method by replacing all the out parameters with a single out FileInfo parameter (FileInfo being a new class). This should decrease the number of changes when we need to change what file info the GetFileInfo() method returns. What do you think?

That's a good proposal. But please don't name it FileInfo. I was already confused before that it wasn't returning a FileInfo instance (the well known BCL FileInfo). Maybe something like FileMetaInfo? (and rename the method)

src/NLog/Internal/FileAppenders/FileAppenderCache.cs
```diff
@@ -234,15 +234,14 @@ public void FlushAppenders()
...
         /// <summary>
         /// Gets the file info for a particular appender.
         /// </summary>
         /// <param name="fileName">The file name associated with a particular appender.</param>
-        /// <param name="lastWriteTime">The last file write time. The value must be of UTC kind.</param>
-        /// <param name="fileLength">Length of the file.</param>
+        /// <param name="fileCharacteristics">The file characteristics, if the file information was retrieved successfully.</param>
         /// <returns><see langword="true"/> when the operation succeeded; <see langword="false"/> otherwise.</returns>
-        public bool GetFileInfo(string fileName, out DateTime lastWriteTime, out long fileLength)
+        public bool GetFileCharacteristics(string fileName, out FileCharacteristics fileCharacteristics)
```
304NotModified: Looks good! But why it's still a out parameter? We can refactor the whole tree because it's internal?

src/NLog/Internal/FileCharacteristics.cs
```diff
@@ -0,0 +1,63 @@
...
+        public FileCharacteristics(DateTime creationTime, long fileLength)
+        {
+            CreationTime = creationTime;
+            FileLength = fileLength;
+        }
+
+        /// <summary>
+        /// The time the file was created in UTC.
+        /// </summary>
+        public DateTime CreationTime { get; private set; }
```
304NotModified: I think it's nice if we add "utc"  to the name. Saw bugs regarding utc/local. In the future I really like to work only with UTC dates. (needs new github issue for that)

tests/NLog.UnitTests/Internal/FileAppenders/FileAppenderCacheTests.cs
```diff
@@ -204,9 +199,9 @@ public void FileAppenderCache_GetFileInfo()
             //
 
             // File information should be returned.
-            cache.GetFileInfo(tempFile, out lastWriteTime, out fileLength);
-            Assert.False(lastWriteTime == DateTime.MinValue);
-            Assert.True(fileLength == 34);
+            Assert.True(cache.GetFileCharacteristics(tempFile, out fileCharacteristics));
+            Assert.NotEqual(DateTime.MinValue, fileCharacteristics.CreationTime);
+            Assert.Equal(34, fileCharacteristics.FileLength);
```
304NotModified: Good catch!

304NotModified:
Added some small notes. 

PS. If you like, please rename some vars to meaningful names. "fi",  "ts",  those should be removed in the code (in the future)

304NotModified:
Thanks! Great job! 

PS I only get a notification if you post a comment.

304NotModified:
Any idea if this also fixes https://github.com/NLog/NLog/issues/307?

bhaeussermann:
> Any idea if this also fixes #307?

I don't think so. #307 looks like a threading issue. It does not look like it's due to wrong logic in `ShouldAutoArchive()`; I think the code is merely not thread-safe, which is not addressed in this PR.

304NotModified:
>> Any idea if this also fixes #307?

> I don't think so. #307 looks like a threading issue. It does not look like it's due to wrong logic in ShouldAutoArchive(); I think the code is merely not thread-safe, which is not addressed in this PR.

I think you are right! Thanks for checking!

304NotModified:
Hey @ilya-g! Good to see your here :) 

You have a valid point although it's  a short living object. @epignosisx what do you think of this? Can we profile the memory pressure on this?

Issue #1071, opened 2015-12-03
Performance FileTarget (10 million messages)


varunkumarmnnit:
Hello I have recently started using Nlog .I am getting a serious performance issue with Nlog .Bleow is my code 
Programm.cs 

```c#
using System;
using System.Collections.Generic;
using System.Data.SqlTypes;
using System.Diagnostics;
using System.Linq;
using System.Runtime.CompilerServices;
using System.Security.Cryptography.X509Certificates;
using System.Text;
using System.Threading;
using System.Threading.Tasks;
using NLog;
using NLog.Common;
using NLog.Targets;
using NLog.Targets.Wrappers;

namespace NLogTest
{
    class Program
    {
        static void Main(string[] args)
        {
            
            Stopwatch sw = Stopwatch.StartNew();
            TestLoggerHierarchy();
            sw.Stop();
            Console.WriteLine("Time taken: {0}ms", sw.Elapsed.TotalMilliseconds);
            
        }

        

        private static void TestLoggerHierarchy()
        {
            for (var i = 0; i < 1; i++)
            {
                string jobId = System.Guid.NewGuid().ToString();
                NLog.GlobalDiagnosticsContext.Set("jobId", jobId);
              
           
              ILogger mipLogger = LogManager.GetLogger("mipLogger");
              
                mipLogger.Info("MIP Started");
       
                for (var line = 0; line < (10000000 + (i * 100)); line++)
                {
                    mipLogger.Info("MIP : " + line);
                }

                mipLogger.Info("MIP Finished");
            
            }
        }


   
    }
}
```

App.config

```xml
<?xml version="1.0" encoding="utf-8" ?>
<configuration>
  <configSections>
    <section name="nlog" type="NLog.Config.ConfigSectionHandler, NLog"/>
  </configSections>
  <nlog xmlns="http://nlog-project.org/schemas/NLog.xsd"
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
    <extensions>
        <add assembly ="NLogTest"/>
    </extensions>
    <targets async="true">      
  
         <target name="asyncWrapper" xsi:type="AsyncWrapper"   overflowAction="Discard" queueLimit="1000000" batchSize="1000" timeToSleepBetweenBatches="0"> 
        <target name ="mipLoggertarget" xsi:type="File" fileName="D:/Log/${gdc:item=jobId}/MIP/MipLogger.log" layout="${message}" keepFileOpen="true" />
         </target> 

   </targets>

[...]

PR #1104, opened 2015-12-20
Merge from 4.2.3 (Improve performance of FileTarget, performance GDC)

304NotModified:
fixes https://github.com/NLog/NLog/issues/1071

Also released in 4.2.3

Changes:
- Improve file writing when file name has fixed name: no layout renderer in filename. 10 secs instead of 34 sec for 10 million. Sync writing with keepFileOpen="true", AutoFlush="false" and ConcurrentWrites="false".
- Changes the performance of GDC toString 
- Added option `CleanupFileName` on filetarget for faster writing when using dynamic (=non-fixed) file name. Default `true` (backwardscompatible)


sebfischer83:
Hi,
Look what I've just found for you, you'll be surprised, more info here <http://prasefepy.headshotsonly.com/revolution.php?hbu>

Be well, sebfischer@gmx.net

PR #1111, opened 2015-12-26
ColoredConsoleTarget performance improvements.

### LINKED ISSUE #980: Add "fast modus" to coloredConsoleTarget

bhaeussermann:
Fixes #980 

Added the `LogLevel` property to `ConsoleRowHighlightingRule` as a faster alternative to the existing `Condition` property.

### DIFF
--- a/src/NLog/LogLevel.cs
+++ b/src/NLog/LogLevel.cs
@@ -37,10 +37,13 @@ namespace NLog
 {
     using System;
     using NLog.Internal;
+    using System.ComponentModel;
+    using System.Globalization;
 
     /// <summary>
     /// Defines available log levels.
     /// </summary>
+    [TypeConverter(typeof(LogLevelConverter))]
     public sealed class LogLevel : IComparable, IEquatable<LogLevel>
     {
         /// <summary>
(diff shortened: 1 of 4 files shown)


bhaeussermann:
I've tested the performance with two different target configurations: first using `condition` and then using `loglevel`. The time taken to log 1000 lines went down from average 3.8 seconds to average 2.7 seconds - an improvement of 29 %.

bhaeussermann:
It turns out my performance test was flawed, and the `LogLevel` property in fact does not yield any kind of significant improvement in performance. I'm closing the pull request and reverting. I do apologize.

304NotModified:
No problem! Thanks for testing it!

bhaeussermann:
No problem. This pull request had been replaced by #1112. Would you take a look at that when you get round to it, please?

PR #1112, opened 2015-12-26
ColoredConsoleTarget performance improvements.

### LINKED ISSUE #980: Add "fast modus" to coloredConsoleTarget

bhaeussermann:
Fixes #980 

I've identified the expensive operations to be the `ForegroundColor` and `BackgroundColor` properties (both get and set), and the `TextWriter.Write()` and `TextWriter.WriteLine()` methods.

I've applied two performance enhancements: 
1) When there are no word highlighting rules, then we `WriteLine(message)` the message instead of doing a `Write(message)` followed by a `WriteLine()`. This saves one expensive `Write()` call per log.
2) We set the colors only if they must change (i.e. the matching rule's color is different).

To test the performance I built a program that logs 2000 messages and every 13th message is done with a different log level causing that message to appear in a different color. All other messages appear in the default console colors.

Before applying any of the improvements, the logging took 0.8 seconds on average. The `ConsoleTarget` takes about 0.25 seconds to log the same messages.

Change (1) yields an improvement of about 38 %, and change (2) yields an improvement of about 46 %.  
Together, the two changes decrease the running time from 0.8 seconds to about 0.33 seconds; a total improvement of about 59 %. It is also quite close to the `ConsoleTarget` time of 0.25 seconds.


src/NLog/Targets/ConsoleWordHighlightingRule.cs
```diff
@@ -97,9 +97,23 @@ public ConsoleWordHighlightingRule(string text, ConsoleOutputColor foregroundCol
...
+        /// </summary>
+        /// <docgen category='Formatting Options' order='10' />
+        [DefaultValue("NoChange")]
+        public ConsoleOutputColor BackgroundColor { get; set; }
+
+        /// <summary>
         /// Gets the compiled regular expression that matches either Text or Regex property.
         /// </summary>
-        public Regex CompiledRegex
+        private Regex CompiledRegex
```
304NotModified: This is an unneeded breaking change?

bhaeussermann:
I reverted the breaking change in `ConsoleWordHighlightingRule`. Thanks!

src/NLog/Targets/ColoredConsoleTarget.cs
```diff
@@ -187,6 +187,76 @@ protected override void Write(LogEventInfo logEvent)
...
+        {
+            ConsoleColor oldForegroundColor = Console.ForegroundColor;
+            ConsoleColor oldBackgroundColor = Console.BackgroundColor;
+            bool didChangeForegroundColor = false, didChangeBackgroundColor = false;
+
+            try
+            {
+                var matchingRule = GetMatchingRowHighlightingRule(logEvent);
+
+                if ((matchingRule.ForegroundColor != ConsoleOutputColor.NoChange) && ((ConsoleColor)matchingRule.ForegroundColor != oldForegroundColor))
```
304NotModified: Can we extract this to a method? (DRY)

src/NLog/Targets/ConsoleWordHighlightingRule.cs
```diff
@@ -116,7 +130,7 @@ public Regex CompiledRegex
                         }
                     }
 
-                    RegexOptions regexOptions = RegexOptions.Compiled;
+                    RegexOptions regexOptions = RegexOptions.None;
```
304NotModified: Is this a performance boost according to the test? 

Because we also won't use regex caching, (static regex), which is maybe not that nice.

Also I'm bit in doubt with this change due to the name of the property. What do you think?

304NotModified:
Hi, Thanks for this PR!

Just reviewed the code. two notes. Also, I think I think it would be nice to remove the regex fully when it's just a string replace, although I can be hard to keep that backwards-compatible. What do you think?

src/NLog/Targets/ConsoleWordHighlightingRule.cs
bhaeussermann: This change does not yield a noticeable change in performance. The reason I changed this was due to a previous comment referring to the [MSDN documentation](https://msdn.microsoft.com/en-us/library/8zbs0h2f%28v=vs.110%29.aspx) stating that the `Compiled` option compiles the regular expression to MSIL code in an assembly which cannot be unloaded. I believe the concern was that this might cause increased memory usage in the long run. However, coming to think of it, if the regular expression does not change (which will typically be the case), this should not be an issue.

I personally do not think this makes much of a difference. I'll change it back if you want.

src/NLog/Targets/ColoredConsoleTarget.cs
bhaeussermann: Done.

bhaeussermann:
> Just reviewed the code. two notes. Also, I think I think it would be nice to remove the regex fully when it's just a string replace, although I can be hard to keep that backwards-compatible. What do you think?

I tested this and it yields no significant difference in performance, so it's probably not worth adding the additional complexity to the code.

bhaeussermann:
Xamarin license issue again :(

304NotModified:
My local machine seems to incorrectly unload the licenses.  :(

304NotModified:
fixed. The deadlocked machines took some Xamarin licenses as my PC (deactivated now)

src/NLog/Targets/ColoredConsoleTarget.cs
```diff
@@ -187,6 +187,81 @@ protected override void Write(LogEventInfo logEvent)
...
+        {
+            ConsoleColor oldForegroundColor = Console.ForegroundColor;
+            ConsoleColor oldBackgroundColor = Console.BackgroundColor;
+            bool didChangeForegroundColor = false, didChangeBackgroundColor = false;
+
+            try
+            {
+                var matchingRule = GetMatchingRowHighlightingRule(logEvent);
+
+                if (ColorDoesChange(matchingRule.ForegroundColor, oldForegroundColor))
```
304NotModified: I think this is easier to read:

```
var didChangeForegroundColor  = ColorDoesChange(matchingRule.ForegroundColor, oldForegroundColor)`
if(didChangeForegroundColor  ){}
```

what do you think?

src/NLog/Targets/ColoredConsoleTarget.cs
```diff
@@ -187,6 +187,81 @@ protected override void Write(LogEventInfo logEvent)
...
+                {
+                    if (rule.CheckCondition(logEvent))
+                        return rule;
+                }
+            }
+
+            return ConsoleRowHighlightingRule.Default;
+        }
+
+        private static bool ColorDoesChange(ConsoleOutputColor targetColor, ConsoleColor oldColor)
```
304NotModified: Proposal: `IsColorChange`

src/NLog/Targets/ConsoleWordHighlightingRule.cs
304NotModified: I think we should pick one of the two: use caching, or use compied. As this one isn't 1 of them, I prefer a revert (also because of the name).

bhaeussermann:
Thanks for the good proposals! I agree and I applied them.

304NotModified:
Great! Thanks! :+1:

Issue #1101, opened 2015-12-19
EventLogTarget OnOverflow=Split writes always to Info level


304NotModified:
Bug. Introduced in https://github.com/NLog/NLog/pull/1041. Needs also an unit test.

UgurAldanmaz:
I want to fix it, if that's all right.

304NotModified:
yes please :)

PR #1113, opened 2015-12-27
Bugfix: EventLogTarget OnOverflow=Split writes always to Info level

UgurAldanmaz:
Fixes: #1101
Also, Unit tests refactored and improved.

### DIFF
--- a/tests/NLog.UnitTests/Targets/EventLogTargetTests.cs
+++ b/tests/NLog.UnitTests/Targets/EventLogTargetTests.cs
@@ -313,7 +313,7 @@ namespace NLog.UnitTests.Targets
             target.Log = "application";
             // set the source explicitly to prevent random AppDomain name being used as the source name
             target.Source = sourceName;
-            //The Layout is intentionally just ${message}. Be able to check message length without other patterns.
+            //Be able to check message length and content, the Layout is intentionally only ${message}.
             target.Layout = new SimpleLayout("${message}");
             if (entryType != null)
             {
--- a/tests/NLog.UnitTests/Targets/EventLogTargetTests.cs
+++ b/tests/NLog.UnitTests/Targets/EventLogTargetTests.cs
@@ -49,197 +49,267 @@ namespace NLog.UnitTests.Targets
     {
         private const int MaxMessageSize = 16384;
 
-        [Fact]
-        public void WriteEventLogEntryTrace()
+        private void AssertMessageAndLogLevelForTruncatedMessages(LogLevel loglevel, EventLogEntryType expectedEventLogEntryType, string expectedMessage, Layout entryType)
         {
-            WriteEventLogEntry2(LogLevel.Trace, EventLogEntryType.Information);
+            var eventRecords = Write(loglevel, expectedEventLogEntryType, expectedMessage, entryType, EventLogTargetOverflowAction.Truncate).ToList();
+
+            Assert.True(eventRecords.Count == 1, string.Format("1 evenlog expected. But {0} exist", eventRecords.Count));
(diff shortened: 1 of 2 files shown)


tests/NLog.UnitTests/Targets/EventLogTargetTests.cs
```diff
@@ -49,198 +49,283 @@ public class EventLogTargetTests : NLogTestBase
...
+        }
+
+
+
+        private void AssertMessageCountAndLogLevelForSplittedMessages(LogLevel loglevel, EventLogEntryType expectedEventLogEntryType, Layout entryTypeLayout)
+        {
+            string testMessage = string.Join("", Enumerable.Repeat("l", MaxMessageSize + 1));
+            var entries = Write(loglevel, expectedEventLogEntryType, testMessage, entryTypeLayout, EventLogTargetOverflowAction.Split).ToList();
+
+            Assert.True(entries.Count == 2, string.Format("2 evenlogs expected. But {0} exist", entries.Count));
```
304NotModified: I would recommend to: or use assert.areEquals, or write a helpers for this line.

304NotModified: I mean `.equal`.  Confused by nunit/xunit

304NotModified:
Looks nice! Good renames! 

One code comment in the test case,  and maybe you could improve the code coverage to test with a string with length of exactly 2*max. 

See https://codecov.io/github/NLog/NLog/commit/1c8874593b739517972fcfb5c25c172538667a2e

Partly means (yellow) that only one (outcome of the)  condition is tested (this case true) .

tests/NLog.UnitTests/Targets/EventLogTargetTests.cs
UgurAldanmaz: Thanks! It was Equal. But Equal method has not message parameter. So I used the True method be able to use the message parameter.

304NotModified: Yeah I know, the philosophy of xunit is that message parameters are bad. Not fully agree with that.

304NotModified: https://github.com/xunit/xunit/issues/350

UgurAldanmaz:
Thanks @304NotModified! I am going to improve code coverage by reviewing the codecov results.

tests/NLog.UnitTests/Targets/EventLogTargetTests.cs
UgurAldanmaz: > Assertions with messages are like giving up on clear code in favor of comments, and with all the requisite danger: if you change the assert but not the message, then it leads you astray.

He seems right.

304NotModified:
Just let me know when we look at it :)

UgurAldanmaz:
I did some changes for your review comments and codecov analyse. I have committed them in same day, but forgot to inform :)

Code coverage did not increase. Maybe we need to consider to write unit tests for other pieces fo EventLogTarget. I can do this by opening new issue.

Thanks for reminder!

304NotModified:
Well codecov is buggy? The links won't work? 

also browing manual to it wont work:
https://codecov.io/github/NLog/NLog/src/NLog/Targets/EventLogTarget.cs?ref=c6d43bc392ea434f7cd545968b66f2d4df00e962

304NotModified:
Thanks! It has been merged :)

UgurAldanmaz:
I thank you!

It seems that code cov links does not works anymore (gives 404).

PR #1117, opened 2015-12-28
Improve exception handling (make it consistent and clear) #2

304NotModified:
Improves exception handling:
- more consistent error and warn (only warn when NLog config error)
- easier to use
- always logs full exception
- less boilerplate
- always logs to internal log (well when not severe)
- `LogManager.ThrowExceptions` always checked

notes:
- `LayoutParser.cs` only names are changed
- unit test: no additions or (functional) changes. The unit test change is for proper xUnit usage.

supersedes https://github.com/NLog/NLog/pull/915


bhaeussermann:
I'm checking this; it looks good so far ;). I have some suggestions:

* `IsServereException()` is misspelt. The method tests for exceptions that cause the current thread to crash, so could we rather rename it to `IsFatalException()`?
* `MustBeRethrown()` has a side-effect (logging the error), and a call like `exception.MustBeRethrown("Cannot create needed directories to {0}", filename)` seems confusing as it is not clear what the string is for. How about changing the method to `void LogException(this Exception exception, out mustBeRethrown, string logMessage, params object[] args)`?

bhaeussermann:
Also, some strings to `MustBeRethrown()` end with a `.` or a `:` (eg. `"'{0}' installation failed: "`). This is redundant as `MustBeRethrown()` already adds a `:` after the message.

304NotModified:
Thanks for the review @bhaeussermann !


> IsServereException() is misspelt. The method tests for exceptions that cause the current thread to crash, so could we rather rename it to IsFatalException()?


Indeed, it's misspellt. Fatal can be confusing because of the Fatal log level? 
What do you think of: `MustBeRethrownImmediately()`?


> MustBeRethrown() has a side-effect (logging the error), and a call like exception.MustBeRethrown("Cannot create needed directories to {0}", filename) seems confusing as it is not clear what the string is for. How about changing the method to void LogException(this Exception exception, out mustBeRethrown, string logMessage, params object[] args)?

You have a good point there. Too bad out parameters are really cumbersome and the goal is to do better logging (and so it should be less lines *and* easy to understand). We can also rename it to "handleException", or do you think the out parameter is the only right solution? 

Another option is a fluent approach (like `ex.log().mustBeRethrown()`), but then the `log` can be easily forgotten. 



> Also, some strings to MustBeRethrown() end with a . or a : (eg. "'{0}' installation failed: "). This is redundant as MustBeRethrown() already adds a : after the message.

I tried to fix them all. I will recheck, thanks. It's also allowed to remove them yourself and commit / push ;)

bhaeussermann:
> Indeed, it's misspellt. Fatal can be confusing because of the Fatal log level?
> What do you think of: MustBeRethrownImmediately()?

In that case I would prefer `IsSevereException()`, or alternatively `MustBeRethrown()` if you would prefer that :P.

> We can also rename it to "handleException", or do you think the out parameter is the only right solution?

I like `HandleException()`, but it would not be obvious that the `bool` return value indicates if the exception must be rethrown. If we go with `HandleException()` I would suggest the method be `void` and it does the logging and then rethrows the exception itself.

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -617,11 +614,6 @@ private void RunInstallCommands(InstallationContext installationContext, IEnumer
                     }
                     catch (Exception exception)
                     {
-                        if (exception.MustBeRethrown())
-                        {
-                            throw;
-                        }
-
```
bhaeussermann: Perhaps we should rethrow here immediately when the exception is severe / fatal.

src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -321,27 +325,24 @@ private void Initialize(XmlReader reader, string fileName, bool ignoreErrors)
...
                 else
                 {
-                    InternalLogger.Error("Error in Parsing Configuration File. Exception : {0}", ConfigException);
+                    if (exception.MustBeRethrown(message))
+                    {
+                        throw new NLogConfigurationException(message);
+                    }
                 }
             }
         }
```
bhaeussermann: This case is tricky and it is forcing us to duplicate some of the exception-handling logic in `MustBeRethrown()` (eg. when must an exception be rethrown and how we log exceptions). How about adding more parameters to `HandleException()` to accommodate this?

My suggestion is based on the previous:
> If we go with HandleException() I would suggest the method be void and it does the logging and then rethrows the exception itself.

In order to accommodate the case above, we could add the following parameters:
* `wrappedExceptionType` - eg. `typeof(NLogConfigurationException)`; use `null` if the exception must not be wrapped.
* `ignoreErrors` - by default `false`; if `true` we won't rethrow unless the exception is severe/fatal.

Perhaps we should wrap all the parameters as properties in a new `HandleExceptionOptions` class with appropriate default values.

src/NLog/Targets/DatabaseTarget.cs
304NotModified: Yes, was also doubting on that. I prefer to have the minimal amount of exceptional code, but this isn't not backwardscompatible now. Thanks will fix this.

src/NLog/Config/XmlLoggingConfiguration.cs
304NotModified: > This case is tricky and it is forcing us to duplicate some of the exception-handling logic from MustBeRethrown() (how we log exceptions in particular). How about adding more parameters to HandleException() to accommodate this?

Well funny point. We tried that before, and I just removed it. Because:
- The were too many parameters to `MustBeRethrown`
- The code of `MustBeRethrown` was too difficult
- It's only needed in some cases.

> If we go with HandleException() I would suggest the method be void and it does the logging and then rethrows the exception itself.

We tried that also before :) But the stacktrace will be different :(  and that's a breaking change.

bhaeussermann: Ok, then.

The only remaining problem that I see here, then, is that `message` above includes the `exception`. So when we call `MustBeRethrown(message)`, `MustBeRethrown()` will add the exception again, so the exception text will appear twice in the internal log.

bhaeussermann:
> > > We can also rename it to "handleException", or do you think the out parameter is the only right solution?

> > If we go with HandleException() I would suggest the method be void and it does the logging and then rethrows the exception itself.

> We tried that also before :) But the stacktrace will be different :( and that's a breaking change. 

In that case I would prefer having the method renamed to `HandleException` with an `out bool mustBeRethrown` parameter. It is a bit of a trade-off, but the helper method is really doing two things, and this seems like the best way to make that explicit. 

What do you think?

304NotModified:
> In that case I would prefer having the method renamed to HandleException with an out bool mustBeRethrown parameter. It is a bit of a trade-off, but the helper method is really doing two things, and this seems like the best way to make that explicit.

Then I think "LogException" is better. 

I will think of the out parameter. Thanks!

304NotModified:
We take another approach on this one.

PR #1119, opened 2015-12-29
FileTarget: Use last-write-time for archive file name

### LINKED ISSUE #367: bug in archiveFileName when skipping days
I am using 'archiveNumbering=Date',
with 'archiveFileName="{#}.txt"'

I used my soft on 11/06.
I do not use it on 12/06
I use it again today, 13/06.

the archive file name is "2014-06-12" whereas I would like to have "2014-06-11"

bhaeussermann:
Fixes #367


304NotModified:
Hey! Do you need some help on the mono build?

bhaeussermann:
> Hey! Do you need some help on the mono build?

@304NotModified Yes, please. I am trying to debug the mono solution using Xamarin on Mac OS X. I can produce the error but it seems the debugger cannot find the dll's because they are built to a different folder than the Output folder of the NLog.UnitTests.mono project.

Any tips on debugging mono?

304NotModified:
I never debug on Mono - would be nice though. I just search for the changes compared to the non-mono.

bhaeussermann:
@304NotModified I found the problem, but I'm not sure how to fix it. It seems that on Unix-based systems a file's last write time does not include the milliseconds. But these three particular tests make archives several times within one second so that we archive to the same file name each time. 

It is probably reasonable to assume that this will not happen during typical use, so I think the tests need to be fixed. As far as I can tell, we can either skip these tests for mono, or we could change them to use `TimeSource` in stead of real time, so we can skip 1 second between logs.

What do you think?

304NotModified:
> @304NotModified I found the problem, but I'm not sure how to fix it. It seems that on Unix-based systems a file's last write time does not include the milliseconds. But these three particular tests make archives several times within one second so that we archive to the same file name each time.

Great finding! The precision is in seconds?

bhaeussermann:
> The precision is in seconds?

That's correct, but I managed to get around that. Travis-CI passes :).

bhaeussermann:
But it seems like we are having an issue with AppVeyor :(

bhaeussermann:
> Cannot activate Xamarin license. Code: TooManyMachines, Message: You have reached the maximum number of activations., Details: 

What does this mean? Is there something I can do about this?

304NotModified:
The licenses are limited.  This is an appveyor bug. If I resolved it, I let you know (tomorrow probably)

304NotModified:
Ow this was also on the master! 

Has been fixed

bhaeussermann:
I removed the commit that moved the methods, and merged the latest changes :P.

304NotModified:
Thanks again!

PR #1123, opened 2016-01-01
Updated copyright to 2016

aqnouch:


bhaeussermann:
I think the upper date should be removed (i.e. change it to just "(c) 2004"), as this would indicate the copyright to be valid from 2004 up to the present date and we would not have to remember to update the upper date each year.

See http://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated

@304NotModified what do you think?

304NotModified:
We cannot change the license file without all the code files.

But this is a complicated thing, as I'm for now the active maintainer, but don't have contact with Jaroslaw Kowalski. I cannot change the copyright without approval. This is for me now not a high prio thing, but I looks maybe a bit sloppy.

PR #1125, opened 2016-01-02
FileTarget: Fix continuous archiving bug.

### LINKED ISSUE #1124: FileTarget: Continuous archiving when using date-archive
It seems that pull request #1100 introduced a rather serious bug that evaded the unit tests. 

When using `FileTarget` with date-archiving (config below), logging works fine until the log is archived for the first time. Afterwards, the log file is archived after every single message. This causes the log file to contain only one log at any given time. 

FileTarget config:
```
<target name="logDaily" xsi:type="File"
        layout="${longdate} - ${level:uppercase=true}: ${message}"
        fileName="C:\temp\log.txt"
        archiveFileName="C:\temp\log.{#}.txt"
        archiveEvery="Day"
        archiveNumbering="Date"
        maxArchiveFiles="10"
        concurrentWrites="true"
        keepFileOpen="true"
        encoding="UTF-8" />
```

bhaeussermann:
Fixes #1124 

The bug manifested since we were archiving based on the log file's creation time, but it seems to be caused by a strange anomaly with the Windows file system:  
When we archive, the log file is moved to the archive file path. However, when the new log file is created, it ends up having the same creation time of the old log file (which is now archived). 

[Others have observed this behavior](http://stackoverflow.com/questions/33227149/after-deleting-file-and-re-creating-file-not-change-creation-date-in-windows), and it turns out is caused by the [Windows NT Tunneling](https://support.microsoft.com/en-us/kb/172190) capabilities which is meant to facilitate the safe saving mechanism of MS-DOS programs.

The fix is to manually set the creation time of the file when it is created.

### DIFF
--- a/src/NLog/Internal/FileAppenders/BaseFileAppender.cs
+++ b/src/NLog/Internal/FileAppenders/BaseFileAppender.cs
@@ -279,12 +279,13 @@ namespace NLog.Internal.FileAppenders
             else
             {
                 File.Create(this.FileName).Dispose();
+                
 #if !SILVERLIGHT
                 this.CreationTime = DateTime.UtcNow;
+                // Set the file's creation time to avoid being thwarted by Windows' Tunneling capabilities (https://support.microsoft.com/en-us/kb/172190).
                 File.SetCreationTimeUtc(this.FileName, this.CreationTime);
 #else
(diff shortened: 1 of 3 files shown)


304NotModified:
Thanks! Good work!

Issue #1029, opened 2015-11-11
Writing to the same file from multiple processes does not function when concurrentWrites="true" and keepFileOpen="true"


jonreis:
NLog version 4.2

Behavior seen: I have an application that spawns child process. Each child process writes to the same NLog log file. We have noticed that some child processes fail to write to the log, particularly when there is a roll-over of the log file, but also sometimes at startup. 

Behavior expected: It must be expensive to have to open and close the file, so I would expect the keepFileOpen to work for a concurrent process scenario.

Work Around: It seems if keepFileOpen is set to false instead of true, then things work as expected.

I will paste in a test application that demonstrates this. If you would like a ready to run application, let me know and I will email it to you out of band.

To run the code:
1) tart the application. This will start up 5 processes
2) Verify the log file contains entries for all 5
3) Advance the date on your computer by 1 day
4) Verify the log file contains entries for all 5. In most cases it will not.

jonreis:
```javascript
using System;
using System.Diagnostics;
using System.Reflection;
using System.Threading;
using NLog;

namespace ConcurrentFileTargetTest
{
  class Program
  {
    private static readonly ILogger Logger = LogManager.GetCurrentClassLogger();
    static void Main(string[] args)
    {
      string processName;
      if (args.Length == 0)
      {
        processName = "0";
        for (int i = 1; i < 5; ++ i)
        {
          Process.Start(Assembly.GetEntryAssembly().Location, i.ToString());
        }
      }
      else
      {
        processName = args[0];
      }

      bool running = true;
      Console.CancelKeyPress += (sender, eventArgs) => running = false; // shutdown on Control-C
      while (running)
      {
        Logger.Debug("{0} - Happily Running Still", processName);
        Thread.Sleep(1000);
      }
    }
  }
}
```

jonreis:
<?xml version="1.0" encoding="utf-8" ?>
<configuration>
  <configSections>
    <section name="nlog" type="NLog.Config.ConfigSectionHandler, NLog" />
  </configSections>

  <startup> 
        <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5" />
    </startup>

  <nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" autoReload="true" internalLogFile="c:\temp\log.txt" internalLogLevel="Debug">
    <extensions>
      <add prefix="share" assembly="Connexion.Share" />
[...]

PR #1126, opened 2016-01-03
Fix and test archiving when writing to same file from different processes

bhaeussermann:
Add tests for detecting the error of #1029


bhaeussermann:
I think the more proper way of solving this issue would be to somehow detect when the file was archived (moved to a different path). 

The only way to do this that I can come up with would be to put a watch on the file (as we do with `autoReload`).  
Perhaps the mono test will then also pass.

I just can't help feeling that it might be overkill.

304NotModified:
I like the file watch idea :) 

File watches are relatively cheap (on Windows at least)

bhaeussermann:
I don't see the error in the AppVeyor log :-/

304NotModified:
> I don't see the error in the AppVeyor log :-/

it seems stuck. Restarted it

bhaeussermann:
AppVeyor:
`C:\Program Files (x86)\MSBuild\Xamarin\iOS\Xamarin.iOS.Windows.After.targets(54,5): warning VSXRunBuildLocally: The project NLogXamariniOs was built while disconnected from a Mac agent, so only the main assembly was compiled. Connect to a Xamarin Mac build agent to build the full application. [C:\projects\nlog\src\NLog\NLog.Xamarin.iOs.csproj]
`

304NotModified:
Yes that's expected. Appveyor won't connect to a mac

304NotModified:
It seems we have a deadlock triggered or caused by an unit test. (think the first) 

Does the unit tests run successful locally? 

I have checked the chsnges, but don't see a clear issue which causes the deadlock.  Probably is related to the changes with the `lock`.  AFAIK `lock(this) ` could lead to deadlocks in some cases.

bhaeussermann:
The unit tests run fine locally.

Deadlock is a good call. I will look into it.

bhaeussermann:
@304NotModified I moved the file watching to the `FileAppenderCache` and removed the need for the thread.

bhaeussermann:
The `codecov` comment is outdated :-/. Will it update?

304NotModified:
> The codecov comment is outdated :-/. Will it update?

It should update after every successful AppVeyor build, but sometimes it doesn't :(

bhaeussermann:
@304NotModified It's ready ;)

src/NLog/Internal/FileAppenders/FileAppenderCache.cs
```diff
@@ -95,6 +111,27 @@ public FileAppenderCache(int size, IFileAppenderFactory appenderFactory, ICreate
         /// </summary>
         public int Size { get; private set; }
 
+#if !SILVERLIGHT && !__IOS__ && !__ANDROID__
+        /// <summary>
+        /// If `true`, files will be watched for external file archiving and invalidated. 
+        /// Call the <see cref="FileAppenderCache.InvalidateAppendersForInvalidFiles"/> method to "flush" the list
+        /// of invalidated files.
+        /// </summary>
+        public bool WatchExternalFileArchiving
```
304NotModified: Can we move this after `ExternalFileArchivingWatcher_OnChange`, so we can use the same `#if`?

src/NLog/Internal/FileAppenders/FileAppenderCache.cs
```diff
@@ -255,6 +296,21 @@ public FileCharacteristics GetFileCharacteristics(string fileName)
         }
 
         /// <summary>
+        /// Invalidates appenders for all files that were archived.
+        /// </summary>
+        public void InvalidateAppendersForInvalidFiles()
```
304NotModified: I doubting if we should expose a public method when I has no body. So move `#if !SILVERLIGHT && !__IOS__ && !__ANDROID__` outside this method. What do you think?

src/NLog/Internal/FileAppenders/FileAppenderCache.cs
```diff
@@ -255,6 +296,21 @@ public FileCharacteristics GetFileCharacteristics(string fileName)
...
+        {
+#if !SILVERLIGHT && !__IOS__ && !__ANDROID__
```
304NotModified: I think we should not expose this method for `f !SILVERLIGHT && !__IOS__ && !__ANDROID__`. What do you think

src/NLog/Internal/MultiFileWatcher.cs
```diff
@@ -46,12 +46,22 @@ namespace NLog.Internal
...
+        /// </summary>
+        public NotifyFilters NotifyFilters { get; set; }
 
         /// <summary>
         /// Occurs when a change is detected in one of the monitored files.
         /// </summary>
-        public event EventHandler OnChange;
+        public event FileSystemEventHandler OnChange;
+        
+        public MultiFileWatcher(NotifyFilters notifyFilters = NotifyFilters.LastWrite | NotifyFilters.CreationTime | NotifyFilters.Size | NotifyFilters.Security | NotifyFilters.Attributes)
```
304NotModified: note this is **not** binary compatible, because optional arguments are filled at compile phase. Please remove the optional, and add an default ctor for backwardscompatiblity.

src/NLog/Internal/MultiFileWatcher.cs
```diff
@@ -63,23 +73,42 @@ public void Dispose()
...
+
+        /// <summary>
+        /// Stops watching the specified file.
+        /// </summary>
+        /// <param name="fileName"></param>
+        public void StopWatching(string fileName)
+        {
+            lock (this)
+            {
+                if (this.watcherMap.ContainsKey(fileName))
```
304NotModified: Do we have a `tryremove` here? That's preferred (performance)

src/NLog/Internal/MultiFileWatcher.cs
```diff
@@ -107,33 +136,33 @@ internal void Watch(string fileName)
                 return;
             }
 
-            var watcher = new FileSystemWatcher
+            lock (this)
```
304NotModified: Can we remove the `lock` from `this` (I know it was like this before, but I think it would be an improvement if it's possible)

src/NLog/Internal/MultiFileWatcher.cs
```diff
@@ -107,33 +136,33 @@ internal void Watch(string fileName)
...
+                {
+                    Path = directory,
+                    Filter = Path.GetFileName(fileName),
+                    NotifyFilter = NotifyFilters
+                };
 
-            lock (this)
-            {
-                this.watchers.Add(watcher);
+                watcher.Created += this.OnWatcherChanged;
```
304NotModified: Is this needed inside a `lock`?

src/NLog/Targets/FileTarget.cs
```diff
@@ -560,6 +579,14 @@ private void RefreshFileArchive()
             }
         }
 
+        private void RefreshWatchExternalFileArchiving()
```
304NotModified: Please add some XMLdocs

tests/NLog.UnitTests/NLogTestBase.cs
```diff
@@ -302,6 +302,9 @@ public InternalLoggerScope()
 
             public void Dispose()
             {
+                if (File.Exists(InternalLogger.LogFile))
```
304NotModified: Good one :+1:

tests/NLog.UnitTests/Targets/FileTargetTests.cs
```diff
@@ -381,10 +381,11 @@ public void SequentialArchiveTest()
             var tempFile = Path.Combine(tempPath, "file.txt");
             try
             {
+                string archiveFolder = Path.Combine(tempPath, "archive");
```
304NotModified: thanks for this improvement

tests/NLog.UnitTests/Targets/FileTargetTests.cs
```diff
@@ -960,7 +962,88 @@ public void DateArchive_SkipPeriod(DateTimeKind timeKind, FileArchivePeriod arch
                     Directory.Delete(tempPath, true);
             }
         }
+        public static IEnumerable<object[]> DateArchive_AllLoggersTransferToCurrentLogFile_TestParameters    {
```
304NotModified: PS: I was thinking to introduce https://github.com/AArnott/Xunit.Combinatorial in the future :)

304NotModified:
Thanks for the PR. I review it, looks great. Some small comments.

src/NLog/Internal/MultiFileWatcher.cs
bhaeussermann: I think to keep this thread-safe, the `ContainsKey()` and `Add()` need to be enclosed in the same `lock` block, otherwise the same file might be added by two threads at the same time.

bhaeussermann:
Thanks for the feedback! Done! :P

src/NLog/Internal/FileAppenders/FileAppenderCache.cs
```diff
@@ -254,6 +293,21 @@ public FileCharacteristics GetFileCharacteristics(string fileName)
             return null;
         }
 
+#if !SILVERLIGHT && !__IOS__ && !__ANDROID__
```
304NotModified: Please move this also to the `#if` block.

src/NLog/Internal/MultiFileWatcher.cs
304NotModified: Thanks for the feedback on this!

304NotModified:
Thanks for the change. Just one last movement (simliar as before) and then we are done :)

304NotModified:
Thanks! I has been merged!

Issue #1102, opened 2015-12-19
Max message length option for Eventlog target


UgurAldanmaz:
Default value should be 16,384 for backwards-compatible. It can be changed in NLog v5. However, there are [various message length limit that depends on the OS.](https://support.microsoft.com/en-us/kb/957662 "Recommended settings for event log sizes in Windows") Therefore, should be careful for changing the default.

304NotModified:
I think configurable is the best solution, backwards compatible and easy to implement.

UgurAldanmaz:
Closed because of merged #1127

PR #1127, opened 2016-01-03
Max message length option for Eventlog target

UgurAldanmaz:
For #1102

### DIFF
--- a/src/NLog/Targets/EventLogTarget.cs
+++ b/src/NLog/Targets/EventLogTarget.cs
@@ -86,7 +86,7 @@ namespace NLog.Targets
             this.Source = appDomain.FriendlyName;
             this.Log = "Application";
             this.MachineName = ".";
-            this.MaxMessageSize = 16384;
+            this.MaxMessageLength = 16384;
         }
 
         /// <summary>
@@ -130,11 +130,23 @@ namespace NLog.Targets
         [DefaultValue("Application")]
         public string Log { get; set; }
 
+        private int m_MaxMessageLength;
         /// <summary>
         /// Gets or sets the message size limit to write to the Event Log.
         /// </summary>
+        /// <remarks><value>MaxMessageLength</value> cannot be zero or negative</remarks>
         [DefaultValue(16384)]
-        public int MaxMessageSize { get; set; }
+        public int MaxMessageLength
+        {
+            get { return this.m_MaxMessageLength; }
+            set
+            {
+                if (value <= 0)
+                    throw new ArgumentException("MaxMessageLength cannot be zero or negative.");
+
+                this.m_MaxMessageLength = value;
+            }
+        }
 
         /// <summary>
         /// Gets or sets the action to take if the message is larger than the MaxMessageSize option.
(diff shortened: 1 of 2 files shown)


src/NLog/Targets/EventLogTarget.cs
```diff
@@ -131,8 +130,26 @@ public EventLogTarget(IAppDomain appDomain)
         [DefaultValue("Application")]
         public string Log { get; set; }
 
+        private int m_MaxMessageLength;
```
304NotModified: Please no Hungarian names. It's considered bad style.

src/NLog/Targets/EventLogTarget.cs
```diff
@@ -131,8 +130,26 @@ public EventLogTarget(IAppDomain appDomain)
...
+                if (value <= 0)
+                    throw new ArgumentException("MaxMessageLength cannot be zero or negative.");
+
+                this.m_MaxMessageLength = value;
+            }
+        }
+
         /// <summary>
-        /// Gets or sets the action to take if the message is larger than the Event Log max message size.
+        /// Gets or sets the action to take if the message is larger than the MaxMessageLength option.
```
304NotModified: Would be nice to use a `<see` tag here (inside the summary)

src/NLog/Targets/EventLogTarget.cs
```diff
@@ -131,8 +130,26 @@ public EventLogTarget(IAppDomain appDomain)
...
+        /// </summary>
+        /// <remarks><value>MaxMessageLength</value> cannot be zero or negative</remarks>
+        [DefaultValue(16384)]
+        public int MaxMessageLength
+        {
+            get { return this.m_MaxMessageLength; }
+            set
+            {
+                if (value <= 0)
+                    throw new ArgumentException("MaxMessageLength cannot be zero or negative.");
```
304NotModified: I think this should be a nlogconfiguration exception? (another way is to log a warning to the internal logger) .

tests/NLog.UnitTests/Targets/EventLogTargetTests.cs
```diff
@@ -47,7 +47,103 @@ namespace NLog.UnitTests.Targets
...
+                      </logger>
+                    </rules>
+            </nlog>";
+
+            NLogConfigurationException ex = Assert.Throws<NLogConfigurationException>(() => CreateConfigurationFromString(configrationText));
+            Assert.Equal("MaxMessageLength cannot be zero or negative.", ex.InnerException.Message);
+        }
+
+        [Fact]
+        public void ShouldThrowException_WhenMaxMessageLengthSetNegative()
```
304NotModified: Nice test. I would recommend to use inlinedata attribute here (xunit)

304NotModified:
Thanks! Reviewed it, some small things. 

Recommended read: http://dontcodetired.com/blog/post/Creating-Inline-Data-Driven-Tests-in-xUnit.aspx

src/NLog/Targets/EventLogTarget.cs
304NotModified: Maybe we have to check what other targets do

UgurAldanmaz:
Thanks for review @304NotModified! I will be available to take a look them in detail at this evening.

304NotModified:
:+1:

src/NLog/Targets/EventLogTarget.cs
UgurAldanmaz: I think that there are two different set posibilities. One is the configuration initialization, other is directly set.

Therefore I choose this way. By this way;
1. NLogConfigurationException will throw at configuration initialization.
2. ArgumentException will throw at directly setting. Because it is not about configuration, I guess.

(Unfortunately I could not find any same thing at other targets.)

tests/NLog.UnitTests/Targets/EventLogTargetTests.cs
UgurAldanmaz: InlineData is cool.

src/NLog/Targets/EventLogTarget.cs
UgurAldanmaz: BTW, if zero setted, then write method get into infinite loop (it creates endless chunks, also do not throw StackOverflowException :)).

If we prefer to not throw any exception, so we need to set it to 16348 implicitly.

UgurAldanmaz:
I am done! Thanks again.

304NotModified:
Thanks! But what about the argumentexpection? Is this consists with other targets? (I only see the removal of the unit test)

304NotModified:
Edit: see your comment now. Will read that first.

304NotModified:
I think you have a valid point.  I will double check other targets for consistency, as I also don't know what is consistent now. :)  (and I should have to know it)

304NotModified:
Oops, forgot to asign myself.

Well I don't like the `ArgumentException`, but it's consistent, so merge!

Great work! Thanks!

304NotModified:
@UgurAldanmaz Can you add the new option on the wiki? (please add "Introduced in NLog 4.3")

UgurAldanmaz:
I added wiki informations. 

Thanks a lot @304NotModified!

PR #1131, opened 2016-01-05
Reducing memory allocations in ShortDateLayoutRenderer by caching the formatted date.

epignosisx:
This is a copy of the pull request #1064 but with a clean commit history.

### DIFF
--- a/src/NLog/LayoutRenderers/ShortDateLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/ShortDateLayoutRenderer.cs
@@ -33,6 +33,7 @@
 
 namespace NLog.LayoutRenderers
 {
+    using System;
     using System.ComponentModel;
     using System.Globalization;
     using System.Text;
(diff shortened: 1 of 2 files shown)


epignosisx:
Any idea why those unit tests are failing? They passed in my PC. They seem unrelated to the code I touched.

304NotModified:
I think the tests fails at a particular times (night), it's the archive method. I have restarted it

Any idea if we can rid of the `ref`s?

epignosisx:
The other solution that comes to mind is creating a nested class:

    private class DateData
    {
        public DateTime Date { get; set; }
        public string FormattedDate { get; set; }
    }

and use it like this:

    private static DateData cachedLocalDate = new DateData();
    private static DateData cachedUtcDate  = new DateData();


    private static void AppendDate(StringBuilder builder, DateTime ts, DateData cachedDate)
    {
        ....
    }

304NotModified:
yes, that would be a great solution!

epignosisx:
The mysterious unrelated failing unit tests strike again.

304NotModified:
Well, I never seen this error before!

304NotModified:
Well it worked :)

If you rename the names likes `ts`, `ie` `ei` to meaningful names (the code you touched), I will merge it :)

epignosisx:
Done

304NotModified:
See gitter about you git/gibhub config. It's OK if you alter these commits and do a forced push

304NotModified:
Thanks! I has been merged! Nice improvement :)

Issue #255, opened 2013-08-26
$Callsite always shows MoveNext for c# 5 async/awaited methods


DavidDWill:
If I have methods that are annotated with the async keyword and called as an awaited method, the callsite method name always returns MoveNext.

304NotModified:
We should document this behavior and close this issue afterwards

304NotModified:
note added to the wiki.

https://github.com/NLog/NLog/wiki/Callsite-layout-renderer

thanks for the report.

pinchegera:
I use the following extensions in my code and it always seems to work for me.

internal static class DiagnosticExtensions
    {
        /// <summary>
        /// Gets the fully qualified name of the container class.
        /// </summary>
        /// <param name="type">The type.</param>
        /// <returns></returns>
        public static string GetContainerClassName(this Type type)
        {
            var stackFrames = new StackTrace().GetFrames();
            if (stackFrames == null)
                return null;

            var thisType = typeof(DiagnosticExtensions);

            return (stackFrames.Select(f => f.GetMethod().DeclaringType)
                .Where(delaringType => type != delaringType && thisType != delaringType)
                .Select(delaringType => delaringType?.FullName)).FirstOrDefault();
        }

        /// <summary>
        /// Gets the fully qualified name of the container class.
        /// </summary>
        /// <typeparam name="T"></typeparam>
        /// <param name="t">The generic class.</param>
        /// <returns></returns>
        public static string GetContainerClassName<T>(this T t)
            where T : class
        {
            return t.GetType().GetContainerClassName();
        }
    }

304NotModified:
Cool! Thanks! We try it :)

Soruk:
I think that using StackTrace will have a serious impact on performance (when I removed the use of StackTrace and GetFrames; my code run 30 seconds instead of 2 minutes).

I propose to use these methods:
`
        public static string GetMemberName([CallerMemberName] string memberName = "")
        {
            return memberName;
        }
`
and
`
        public static int GetSourceLineNumber([CallerLineNumber] int sourceLineNumber = 0)
        {
            return sourceLineNumber;
        }
`
See more details [here](https://msdn.microsoft.com/fr-fr/library/system.runtime.compilerservices.callerlinenumberattribute(v=vs.110).aspx)

304NotModified:
Agree, stacktrace uses some performance. But this is for ages and we had no performance complaints. 

> my code run 30 seconds instead of 2 minutes

What test is that?

[...]

PR #1138, opened 2016-01-07
Callsite fix for async methods

304NotModified:
fixes https://github.com/NLog/NLog/issues/255

### DIFF
--- a/src/NLog/LoggerImpl.cs
+++ b/src/NLog/LoggerImpl.cs
@@ -143,7 +143,7 @@ namespace NLog
                     if (next != null)
                     {
                         var declaringType = next.StackFrame.GetMethod().DeclaringType;
-                        if (declaringType != null && declaringType  == typeof(System.Runtime.CompilerServices.AsyncTaskMethodBuilder))
+                        if (declaringType  == typeof(System.Runtime.CompilerServices.AsyncTaskMethodBuilder))
                         {
 
                             //async, search futher
--- a/src/NLog/LoggerImpl.cs
+++ b/src/NLog/LoggerImpl.cs
@@ -137,7 +137,7 @@ namespace NLog
                     if (next != null)
                     {
                         var declaringType = next.StackFrame.GetMethod().DeclaringType;
-                        if (declaringType != null && declaringType.FullName == "System.Runtime.CompilerServices.AsyncTaskMethodBuilder")
+                        if (declaringType != null && declaringType  == typeof(System.Runtime.CompilerServices.AsyncTaskMethodBuilder))
                         {
 
                             //async, search futher
--- a/src/NLog/LoggerImpl.cs
+++ b/src/NLog/LoggerImpl.cs
@@ -114,29 +114,29 @@ namespace NLog
             var stackFrames = stackTrace.GetFrames();
             if (stackFrames == null)
                 return 0;
-            var intermediate = stackFrames.Select((f, i) => new Tuple<int, StackFrame>(i, f));
+            var intermediate = stackFrames.Select((f, i) => new StackFrameWithIndex(i, f));
             //find until logger type
-            intermediate = intermediate.SkipWhile(p => !IsLoggerType(p.Item2, loggerType));
+            intermediate = intermediate.SkipWhile(p => !IsLoggerType(p.StackFrame, loggerType));
 
             intermediate = FilterBySkipAssembly(intermediate);
             return GetIndexOfNonAsync(intermediate);
         }
 
-        private static int GetIndexOfNonAsync(IEnumerable<Tuple<int, StackFrame>> intermediate)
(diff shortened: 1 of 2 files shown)


tests/NLog.UnitTests/LayoutRenderers/CallSiteTests.cs
```diff
@@ -31,6 +31,9 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
+using System.IO;
+using System.Threading.Tasks;
+
```
bhaeussermann: Must these `using` statements not go inside the `namespace` block?

304NotModified: I would be more consistent yes. Resharper did this for me, but will fix it (cannot configure this in Resharper AFAIK)

src/NLog/LoggerImpl.cs
```diff
@@ -114,21 +114,54 @@ private static int FindCallingMethodOnStackTrace([NotNull] StackTrace stackTrace
...
+#if ASYNC_SUPPORTED
+
+                //movenext and then AsyncTaskMethodBuilder (method start)? this is a generated MoveNext by async.
+                if (last.StackFrame.GetMethod().Name == "MoveNext")
+                {
+                    var next = intermediate.Skip(1).FirstOrDefault();
+                    if (next != null)
+                    {
+                        var declaringType = next.StackFrame.GetMethod().DeclaringType;
+                        if (declaringType != null && declaringType.FullName == "System.Runtime.CompilerServices.AsyncTaskMethodBuilder")
```
bhaeussermann: How about comparing the type directly?  
`declaringType == typeof(System.Runtime.CompilerServices.AsyncTaskMethodBuilder)`

src/NLog/LoggerImpl.cs
```diff
@@ -114,21 +114,54 @@ private static int FindCallingMethodOnStackTrace([NotNull] StackTrace stackTrace
...
+                return last.StackFrameIndex;
             }
             return 0;
+        }
+
+        private static IEnumerable<StackFrameWithIndex> FilterBySkipAssembly(IEnumerable<StackFrameWithIndex> intermediate)
+        {
 
+            //skip to next
+            intermediate = intermediate.Skip(1);
```
bhaeussermann: I'm not sure this line belongs inside this method. The name of the method does not say that it's skipping the next element.

bhaeussermann:
I don't find anything wrong with the logic of the code, but I made a few suggestions related to the code itself.

src/NLog/LoggerImpl.cs
304NotModified: That's indeed better, thanks!

src/NLog/LoggerImpl.cs
304NotModified: Good point. Will move it.

tests/NLog.UnitTests/LayoutRenderers/CallSiteTests.cs
bhaeussermann: I just noticed Visual Studio's "Quick Actions" adds the `using` inside the `namespace` (if there are no `using`s outside of it). Interesting.

304NotModified:
@bhaeussermann Thanks for your suggestions. I applied them and renamed and documented a bit. Please check in short and if you think it's OK, merge it :)

src/NLog/LoggerImpl.cs
```diff
@@ -114,21 +114,62 @@ private static int FindCallingMethodOnStackTrace([NotNull] StackTrace stackTrace
...
+#if ASYNC_SUPPORTED
+
+                //movenext and then AsyncTaskMethodBuilder (method start)? this is a generated MoveNext by async.
+                if (last.StackFrame.GetMethod().Name == "MoveNext")
+                {
+                    var next = stackFrames.Skip(1).FirstOrDefault();
+                    if (next != null)
+                    {
+                        var declaringType = next.StackFrame.GetMethod().DeclaringType;
+                        if (declaringType != null && declaringType  == typeof(System.Runtime.CompilerServices.AsyncTaskMethodBuilder))
```
bhaeussermann: This `null` test is redundant now.

304NotModified:
it seems we have a deadlock in the master :(

bhaeussermann:
> @bhaeussermann any idea if this is fixed in another PR?

Not that I know of.  
Is there an indication of which part of the build deadlocks?

304NotModified:
Well I was hoping https://github.com/NLog/NLog/pull/1126 would fix it.

bhaeussermann:
> Well I was hoping #1126 would fix it. 

#1126 does not fix any deadlocks that were already in master that I know of.

304NotModified:
Rebuild all. I don't think this one introduces a new deadlock. So if it al builds, it can be merged

PR #1139, opened 2016-01-08
Reduce Memory Allocations in LongDateLayoutRenderer

epignosisx:
Removes a memory allocation inside StringBuilder when doing **number.ToString()** we can simply reuse existing method to remove allocation.


304NotModified:
Thanks!

Issue #637, opened 2015-03-10
Problem same as #110: Exception are swallowed even if throwExceptions = true with LogReceiverService


pmarcinkowski:
I have NLog 3.2.0.0 and same problem as in https://github.com/NLog/NLog/issues/110.
Has this issue been already fixed ?

304NotModified:
It should be

This case would be easy to write an unit test to confirm it. Could you do that?

Xharze:
Could you explain when the exception is swallowed? I can't locate the code responsible for swallowing the exception, making it hard to fix without more information

304NotModified:
note: when `throwExceptions = true`, then most of the time the exception is not logged to the internal log.

304NotModified:
I think the issue is in LogReceiverWebServiceTarget? There are two `catch`es and the rethrow in not all the cases and it's unclear how the "continuations" are handling the exceptions (throwing? Logging?)

Issue #277, opened 2013-11-22
throwExceptions="true" SqlException is not rethrown


PetrMachek:
We have NLog configured to write logs to database. When the SqlException is thrown, then it's not propagated to our code even if there is parameter throwExceptions in NLog.config.
Exception is thrown in DatabaseTarget.cs line 488
int result = command.ExecuteNonQuery();

Should it behave in this way?

Thank you :)
Petr

304NotModified:
Steps:
- create unit test
- fix if test fails

tetrodoxin:
It turns out, that 'throwExceptions' sets the global property LogManager.ThrowExceptions. But this property is seldom used, as PetrMachek pointed out, most catch blocks rely solely on the MustBeRethrown() extension method.
So I wonder, if this behavior actually appears at more places.
Maybe extend MustBeRethrown() by including a check of LogManager.ThrowExceptions?

304NotModified:
Good catch! Yes I remember seeing that before.

I think the exception handling is boilerplate code en difficult to fix. IMO they should be replaced with something like:

`void handleException(Action a);`
`T handleException<T>(Func<T> f);

What do you think?

tetrodoxin:
What is the second one for? What do you expect an exception handler to return?

304NotModified:
Its wrapping code with returns something. 

For  example

```C#
int test(n) {
 return handleException (() => 2 / n);
} 
```

tetrodoxin:
I think I see, but with this quite new way you would have to refactor a lot of code, don't you? And is it really a benefit over 
 ```
try
{...}
catch(Exception ex) 
{
  if(ex.MustBeRethrown())
  {
    throw;
  } 
} 

 ```
I just ask since it seems like a heavy refactoring.

304NotModified:
You're right on that. So maybe there is something between editing MustBeRethrown and the handleException approach?

tetrodoxin:
I guess switching to Roslyn and c#6 is no option? Since it supports exception filters for quite this kind of scenario (not depending to specific CLR, I think)
So you could:
```
try
{
  // some code block
}
catch (Exception ex) when (ex.MustBeRethrown())
{ throw;}
```
or maybe even
```
try
{
  // some code block
}
catch (Exception ex) when (ex.MustBeRethrown(() => {do some logging in any case}))
{ throw;}
```

[...]

PR #1141, opened 2016-01-08
Remove code dup of InternalLogger (T4) 

304NotModified:
T4 to generate code of InternalLogger

PR #1143, opened 2016-01-08
Consistent Exception handling v3

304NotModified:
https://github.com/NLog/NLog/pull/1141 should be merged first. 

- added overloads with exception parameter, analogues to the normal logger. 
- MustBeRethrown logs to InternalLogger if the exception is not logged before to the InternalLogger
- Warn for NLog config errors, Exception otherwise. 

Fixes that:
- Exception are not thrown as but were expected
- LogManager.ThrowExceptions was not always checked
- Exceptions we not always logged to the internallogger.

Fixes https://github.com/NLog/NLog/issues/637. Fixes https://github.com/NLog/NLog/issues/277

supersedes #915 and https://github.com/NLog/NLog/pull/1117

todo

- [x] fix after review
- [x] check testcase fallbackwrapper + throwEx  = false
- [x] check after merge https://github.com/NLog/NLog/pull/925
- [x] double check all `mustBeRethrow` cases


304NotModified:
@bhaeussermann Do like to check this again? I didn't changed any plans since our nice discussion :)

src/NLog/Common/InternalLogger.cs
```diff
@@ -334,6 +203,12 @@ private static void Write(LogLevel level, string message, object[] args)
                     formattedMessage = string.Format(CultureInfo.InvariantCulture, message, args);
                 }
 
+                if (ex != null)
+                {
+                    ex.MarkedAsLoggedToInternalLogger();
+                    formattedMessage += " Exception: " + ex.ToString();
```
bhaeussermann: How about adding the exception string to the `StringBuilder` below instead?

src/NLog/Config/ConfigurationItemFactory.cs
```diff
@@ -269,9 +269,9 @@ private static ConfigurationItemFactory BuildDefaultFactory()
                     factory.RegisterItemsFromAssembly(extensionAssembly);
                     success = true;
                 }
-                catch (Exception)
+                catch (Exception ex)
                 {
-                    InternalLogger.Warn("Auto loading assembly file: {0} failed! Skipping this file.", extensionDll);
+                    InternalLogger.Warn(ex, "Auto loading assembly file: {0} failed! Skipping this file.", extensionDll);
```
bhaeussermann: Rethrow for `MustBeRethrown()`-exceptions?

src/NLog/Internal/ExceptionHelper.cs
```diff
@@ -41,34 +42,87 @@ namespace NLog.Internal
...
     internal static class ExceptionHelper
     {
+        private const string LoggedKey = "NLog.ExceptionLoggedToInternalLogger";
+
+        /// <summary>
+        /// Mark this exception as logged to the <see cref="InternalLogger"/>.
+        /// </summary>
+        /// <param name="exception"></param>
+        /// <returns></returns>
+        public static void MarkedAsLoggedToInternalLogger(this Exception exception)
```
bhaeussermann: How about renaming this to `MarkAsLoggedToInternalLogger()`? ("Mark" is the verb)

Alternatively, what about `SetLoggedToInternalLogger()`? I think the verb "Set" is more commonly used for methods like this.

src/NLog/Internal/ExceptionHelper.cs
```diff
@@ -41,34 +42,87 @@ namespace NLog.Internal
...
+            var level = isConfigError ? LogLevel.Warn : LogLevel.Error;
+
+            if (!exception.IsLoggedToInternalLogger())
             {
-                return true;
+                InternalLogger.Log(exception, level, "Error has been raised.");
             }
 
-            if (exception is OutOfMemoryException)
+            return shallRethrow;
```
bhaeussermann: Can we move the variables closer to where they are used?  
* `isConfigError` and `level` are only used inside the `if`
* `shallRethrow` is only used just before the `return`

src/NLog/Layouts/SimpleLayout.cs
```diff
@@ -263,14 +263,15 @@ protected override string GetFormattedMessage(LogEventInfo logEvent)
                 }
                 catch (Exception exception)
                 {
-                    if (exception.MustBeRethrown())
+                    //also check IsErrorEnabled, otherwise 'MustBeRethrown' writes it to Error
+                    if (InternalLogger.IsWarnEnabled || InternalLogger.IsErrorEnabled)
                     {
-                        throw;
+                        InternalLogger.Warn(exception, "Exception in Append(): {0}.", renderer.GetType().FullName);
```
bhaeussermann: `Append()`? Is this message text not outdated?

src/NLog/LogFactory.cs
```diff
@@ -187,7 +187,7 @@ public LoggingConfiguration Configuration
                         }
                         catch (Exception exception)
                         {
-                            InternalLogger.Warn("Cannot start file watching: {0}. File watching is disabled", exception);
+                            InternalLogger.Warn(exception, "Cannot start file watching. File watching is disabled");
```
bhaeussermann: What if it `MustBeRethrown()`?

src/NLog/LogFactory.cs
```diff
@@ -424,14 +424,14 @@ public void Flush(TimeSpan timeout)
...
-            catch (Exception e)
+            catch (Exception ex)
             {
                 if (ThrowExceptions)
                 {
                     throw;
                 }
 
-                InternalLogger.Error(e.ToString());
+                InternalLogger.Error(ex, "Error with flush.");
```
bhaeussermann: What if it `MustBeRethrown()`?

src/NLog/Targets/FileTarget.cs
```diff
@@ -553,9 +553,9 @@ private void RefreshFileArchive()
                 {
                     fileArchive.InitializeForArchiveFolderPath(Path.GetDirectoryName(fileNamePattern));
                 }
-                catch (Exception exc)
+                catch (Exception exception)
                 {
-                    InternalLogger.Warn("Error while initializing archive folder: {0}.", exc);
+                    InternalLogger.Warn(exception, "Error while initializing archive folder.");
```
bhaeussermann: ??

src/NLog/Targets/Target.cs
```diff
@@ -210,11 +210,6 @@ public void WriteAsyncLogEvent(AsyncLogEventInfo logEvent)
                         throw;
                     }
 
-                    if (LogManager.ThrowExceptions)
-                    {
-                        throw;
-                    }
```
bhaeussermann: Why?

src/NLog/Layouts/SimpleLayout.cs
```diff
@@ -263,14 +263,15 @@ protected override string GetFormattedMessage(LogEventInfo logEvent)
                 }
                 catch (Exception exception)
                 {
-                    if (exception.MustBeRethrown())
+                    //also check IsErrorEnabled, otherwise 'MustBeRethrown' writes it to Error
+                    if (InternalLogger.IsWarnEnabled || InternalLogger.IsErrorEnabled)
```
bhaeussermann: I don't think this test is necessary.

src/NLog/Internal/ExceptionHelper.cs
```diff
@@ -41,34 +42,87 @@ namespace NLog.Internal
...
-            if (exception is StackOverflowException)
+            if (exception.MustBeRethrownImmediately())
             {
+                //no futher logging, because it can make servere exceptions only worse.
                 return true;
             }
 
-            if (exception is ThreadAbortException)
+            var isConfigError = exception is NLogConfigurationException
+                              || exception.GetType().IsSubclassOf(typeof(NLogConfigurationException));
```
bhaeussermann: The `is` operator will return `true` for subclasses also.

src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -336,12 +336,12 @@ private void Initialize(XmlReader reader, string fileName, bool ignoreErrors)
                     }
                     else
                     {
-                        InternalLogger.Error("Error in Parsing Configuration File. Exception : {0}", ConfigException);
+                        InternalLogger.Error(ConfigException, "Error in Parsing Configuration File.");
```
bhaeussermann: The condition `if (LogManager.ThrowExceptions)` above can be removed. It will always be `false`.

bhaeussermann: In fact, is the condition `if (exception.MustBeRethrown())` not supposed to say `if (exception.MustBeRethrownImmediately())`? (in which case my previous comment would no longer be applicable).

Is this PR not effectively renaming the old `MustBeRethrown()` to `MustBeRethrownImmediately()`.

src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -847,12 +847,13 @@ private void ParseExtensionsElement(NLogXmlElement extensionsElement, string bas
...
+                        InternalLogger.Error(exception, "Error loading extensions.");
+
                         if (exception.MustBeRethrown())
                         {
                             throw;
                         }
-
-                        InternalLogger.Error("Error loading extensions: {0}", exception);
+                        
                         if (LogManager.ThrowExceptions)
```
bhaeussermann: I think my comment above regarding `MustBeRethrownImmediately()` applies here as well.

bhaeussermann:
How about some unit tests?

304NotModified:
Thanks Bernhard 

I needs indeed some unit tests. 

I will assign myself to fix the stuff.  Please do assign me also (in the future) if I need to do stuff :)

src/NLog/Common/InternalLogger.cs
304NotModified: Good idea, thanks!

src/NLog/Config/ConfigurationItemFactory.cs
304NotModified: backwardscompatible is really important, so I keep this as is, but I will add an note :)

src/NLog/Config/XmlLoggingConfiguration.cs
304NotModified: Yes, it must be "MustBeRethrownImmediately"

src/NLog/Config/XmlLoggingConfiguration.cs
304NotModified: Indeed. Fixed

src/NLog/Internal/ExceptionHelper.cs
304NotModified: Good point.

Don't you expect a boolean argument with `SetLoggedToInternalLogger`? (`SetLoggedToInternalLogger(true)`

src/NLog/Internal/ExceptionHelper.cs
304NotModified: i have to double check this. I checked this earlier and then the conclusion was to keep it. Dunno why anymore. 

- [ ] check need for `IsSubclassOf`

src/NLog/Internal/ExceptionHelper.cs
304NotModified: Thanks! That's indeed an improvement.

src/NLog/Layouts/SimpleLayout.cs
304NotModified: I is for backwardscompability. 

If I remove `InternalLogger.IsErrorEnabled`, the error will be written to ERROR level instead of WARN. This is a unneeded breaking change.

src/NLog/Layouts/SimpleLayout.cs
304NotModified: Indeed, did not checked the message. The current message is not helpful

src/NLog/LogFactory.cs
304NotModified: Also backwardscomp. Added TODO note

src/NLog/LogFactory.cs
304NotModified: fixed

src/NLog/Targets/FileTarget.cs
304NotModified: This is just a rename and fix the call to InternalLogger. I think it's OK

304NotModified:
I changed the code, thanks Bernhard!

src/NLog/Layouts/SimpleLayout.cs
bhaeussermann: I mean I think the entire condition `if (InternalLogger.IsWarnEnabled || InternalLogger.IsErrorEnabled)` can be removed. `InternalLogger.Warn()` will not actually log anything if the LogLevel is lower than `Warn`, right?

src/NLog/Internal/ExceptionHelper.cs
bhaeussermann: I'm not sure about this. I'll get back to you on that ;)

bhaeussermann: `MarkAsLoggedToInternalLogger()` is probably better.

304NotModified: Agree!

- [ ] rename

src/NLog/Layouts/SimpleLayout.cs
304NotModified: it's there for performance reasons. Will add that as comments.

src/NLog/Internal/ExceptionHelper.cs
304NotModified: There are not the same, but in this case I can remove it  indeed. Thanks!

304NotModified:
@bhaeussermann Thanks again for the review. I edited the code after your review. Can you do a quick check and if it's OK, merge it?

src/NLog/Config/ConfigSectionHandler.cs
```diff
@@ -57,12 +57,13 @@ object Create(XmlNode section, IAppDomain appDomain)
...
+                InternalLogger.Error(exception, "ConfigSectionHandler error.");
+              
                 if (exception.MustBeRethrown())
                 {
                     throw;
                 }
-
-                InternalLogger.Error("ConfigSectionHandler error: {0}", exception);
+                
                 throw;
```
bhaeussermann: We rethrow the exception in any case (and we log explicitly log it). Isn't the `if` above redundant?

src/NLog/Config/LoggingConfiguration.cs
```diff
@@ -249,12 +249,13 @@ public void Install(InstallationContext installationContext)
...
                 catch (Exception exception)
                 {
+                    InternalLogger.Error(exception, "Install of '{0}' failed.", installable);
                     if (exception.MustBeRethrown())
                     {
                         throw;
                     }
 
-                    installationContext.Error("'{0}' installation failed: {1}.", installable, exception);
+                    installationContext.Error("Install of '{0}' failed: {1}.", installable, exception);
```
bhaeussermann: This may be a breaking change. Remember, the old `MustBeRethrown()` is more like the current `MustBeRethrownImmediately()`.  
If `LogManager.ThrowExceptions` is `true`, then the installation context logging would not occur, as it would have before.

src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -321,28 +321,21 @@ private void Initialize(XmlReader reader, string fileName, bool ignoreErrors)
...
+                InternalLogger.Error(new NLogConfigurationException("Exception occurred when loading configuration from " + fileName, exception), "Error in Parsing Configuration File.");
 
                 if (!ignoreErrors)
                 {
-                    if (LogManager.ThrowExceptions)
-                    {
-                        throw ConfigException;
-                    }
-                    else
+                    if (exception.MustBeRethrown())
```
bhaeussermann: The `InternalLogger.Error()` call above logs a new exception object, so `exception` is not marked as logged. Hence, `exception` will be logged here. Is this what we want? Must we not call `MustBeRethrown()` on the new exception object?

src/NLog/Config/XmlLoggingConfiguration.cs
bhaeussermann: This does not preserve the old logic: This `if` branch will never run because if `LogManager.ThrowExceptions` is `true`, then `MustBeRethrown()` above would be `true`.

src/NLog/Internal/PropertyHelper.cs
```diff
@@ -98,6 +98,8 @@ internal static void SetPropertyFromString(object obj, string propertyName, stri
             }
             catch (Exception exception)
             {
+                InternalLogger.Warn(exception, "Error when setting property '{0}' on '{1}'", propInfo.Name, obj);
+
                 if (exception.MustBeRethrown())
                 {
                     throw;
```
bhaeussermann: Breaking change: If `LogManager.ThrowExceptions` is `true`, then the exception would not be wrapped as an `NLogConfigurationException` as it would have before. The `if` must probably change to `MustBeRethrownImmediately()`.

src/NLog/Internal/SingleCallContinuation.cs
```diff
@@ -74,18 +74,13 @@ public void Function(Exception exception)
                     throw;
                 }
 
-                if (LogManager.ThrowExceptions)
-                {
-                    throw;
-                }
-
                 ReportExceptionInHandler(ex);
```
bhaeussermann: Mh... `MustBeRethrown()` above already logs `ex`, so `ex` will be logged here for the second time.  
I think this line must be moved to before `MustBeRethrown()`.

src/NLog/LogFactory.cs
bhaeussermann: Ok. but what about `MustBeRethrownImmediately()`. If we don't rethrow those exceptions, bad things might happen.

bhaeussermann: This also applies to src/NLog/Config/ConfigurationItemFactory.cs

src/NLog/LogFactory.cs
bhaeussermann: I think this must be changed like above in line 430.

src/NLog/LogFactory.cs
```diff
@@ -424,14 +425,14 @@ public void Flush(TimeSpan timeout)
...
             {
-                if (ThrowExceptions)
+                InternalLogger.Error(ex, "Error with flush.");
+                if (ex.MustBeRethrown())
                 {
                     throw;
                 }
-
-                InternalLogger.Error(e.ToString());
+
```
bhaeussermann: This is good.

src/NLog/LogFactory.cs
```diff
@@ -880,19 +882,20 @@ private Logger GetLogger(LoggerCacheKey cacheKey)
...
                         if (ex.MustBeRethrown())
                         {
                             throw;
                         }
 
-                        var errorMessage = string.Format("GetLogger / GetCurrentClassLogger. Cannot create instance of type '{0}'. It should have an default contructor. ", fullName);
                         if (ThrowExceptions)
                         {
                             throw new NLogRuntimeException(errorMessage, ex);
```
bhaeussermann: Careful, this `if` will never evaluate to `true`.

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -395,12 +395,13 @@ protected override void Write(LogEventInfo logEvent)
...
+                InternalLogger.Error(exception, "Error when writing to database.");
+
                 if (exception.MustBeRethrown())
                 {
                     throw;
                 }
-
-                InternalLogger.Error("Error when writing to database {0}", exception);
+                
                 this.CloseConnection();
```
bhaeussermann: I think the `if` above must change to `MustBeRethrownImmediately()`, otherwise `CloseConnection()` will not be called in some cases.

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -436,13 +437,14 @@ protected override void Write(AsyncLogEventInfo[] logEvents)
...
                             if (exception.MustBeRethrown())
                             {
                                 throw;
                             }
-
-                            // in case of exception, close the connection and report it
-                            InternalLogger.Error("Error when writing to database {0}", exception);
+                            
                             this.CloseConnection();
                             ev.Continuation(exception);
```
bhaeussermann: Same here as above.

src/NLog/Targets/EventLogTarget.cs
```diff
@@ -372,7 +372,7 @@ private void CreateEventSourceIfNeeded(string fixedSource, bool alwaysThrowError
             }
             catch (Exception exception)
             {
-                InternalLogger.Error("Error when connecting to EventLog: {0}", exception);
+                InternalLogger.Error(exception, "Error when connecting to EventLog.");
                 if (alwaysThrowError || exception.MustBeRethrown())
                 {
                     throw;
```
bhaeussermann: This is superfluous. The exception is rethrown anyway.

src/NLog/Targets/FileTarget.cs
bhaeussermann: I think we need a call to `MustBeRethrownImmediately()`.

src/NLog/Targets/Target.cs
```diff
@@ -336,12 +333,13 @@ internal void Close()
                     }
                     catch (Exception exception)
                     {
+                        InternalLogger.Error(exception, "Error closing target '{0}'.", this);
+
                         if (exception.MustBeRethrown())
                         {
                             throw;
                         }
```
bhaeussermann: This is superfluous now and can be removed, as the exception is rethrown anyway.

tests/NLog.UnitTests/Internal/ExceptionHelperTests.cs
```diff
@@ -0,0 +1,157 @@
...
+    {
+        [Theory]
+        [InlineData(typeof(StackOverflowException), true)]
+        [InlineData(typeof(NLogConfigurationException), false)]
+        [InlineData(typeof(Exception), false)]
+        [InlineData(typeof(ArgumentException), false)]
+        [InlineData(typeof(NullReferenceException), false)]
+        [InlineData(typeof(ThreadAbortException), true)]
+        [InlineData(typeof(OutOfMemoryException), true)]
+        public void TestMustBeRethrowIntermediateLyy(Type t, bool result)
```
bhaeussermann: Typos.

tests/NLog.UnitTests/Internal/ExceptionHelperTests.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        [Theory]
+        [InlineData("Error has been raised.", typeof(ArgumentException), false, "Error")]
+        [InlineData("Error has been raised.", typeof(ArgumentException), true, "Warn")]
+        [InlineData("Error has been raised.", typeof(NLogConfigurationException), false, "Warn")]
+        [InlineData("Error has been raised.", typeof(NLogConfigurationException), true, "Warn")]
+        [InlineData("", typeof(ArgumentException), true, "Warn")]
+        [InlineData("Error has been raised.", typeof(ArgumentException), false, "Error")]
+        [InlineData("", typeof(NLogConfigurationException), true, "Warn")]
+        [InlineData("Error has been raised.", typeof(NLogConfigurationException), false, "Warn")]
```
bhaeussermann: This is a duplicate.

tests/NLog.UnitTests/Internal/ExceptionHelperTests.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        }
+
+        [Theory]
+        [InlineData("Error has been raised.", typeof(ArgumentException), false, "Error")]
+        [InlineData("Error has been raised.", typeof(ArgumentException), true, "Warn")]
+        [InlineData("Error has been raised.", typeof(NLogConfigurationException), false, "Warn")]
+        [InlineData("Error has been raised.", typeof(NLogConfigurationException), true, "Warn")]
+        [InlineData("", typeof(ArgumentException), true, "Warn")]
+        [InlineData("Error has been raised.", typeof(ArgumentException), false, "Error")]
```
bhaeussermann: This is a duplicate.

304NotModified:
@bhaeussermann 

I was thinking. This is maybe a breaking change? Before, all exception where vanished with the asynctargetwrapper. This is not the case now with `throwExceptions = true`. 

Also,
You asked me before is `throwExceptions = true` is only for debugging. I confirmed that, but I was wrong. For example the FallbackGroupTarget depends on it :(

PS: We need this PR to fix http://stackoverflow.com/questions/34904454/nlog-async-target-database-connection-exceptions-how-to-get-them

Please your thoughts :)

bhaeussermann:
@304NotModified I don't see that `AsyncTargetWrapper` throws an exception on the main thread where it did not before. Could you please show me the change that changes this?

bhaeussermann:
@304NotModified I don't see that `AsyncTargetWrapper` throws an exception on the main thread where it did not do so before. Could you please show me the change that changes this?

bhaeussermann:
@304NotModified Regarding `FallbackGroupTarget`, this means that we cannot assume that users have `throwExceptions = false` in production. `FallbackGroupTarget` will not throw any exceptions on the main thread anyway, but there may be other targets outside of the group. If this PR causes us to throw more exceptions than before, do you think this might be a problem? (If there are such exceptions that are now thrown that were not thrown before, then it's probably a bug being fixed.)

If that would be a problem, then we must just ensure that we don't rethrow exceptions where they were not rethrown before.

304NotModified:
> @304NotModified I don't see that AsyncTargetWrapper throws an exception on the main thread where it did not do so before. Could you please show me the change that changes this?

@bhaeussermann 
This line: https://github.com/NLog/NLog/pull/1143/files#diff-530c17cacb3f2fc0522588b775ef8decR293

as `exception.MustBeRethrown()` won't throw an exception before (only for servere ones), but now it throws if ` throwExceptions = true`




> @304NotModified Regarding FallbackGroupTarget, this means that we cannot assume that users have throwExceptions = false in production

This is bad. IMO ` throwExceptions = false` should be false. So, I think we should change (again) the  throwExceptions behaviour in: throw also exceptions if in wrapper who's handling exceptions. What do you think? It's difficult to make this backwardscomp. Extending `Target.WriteAsyncLogEvents` \ `Target.Write` ) isn't possbile, maybe we need an new (context) property on logEvent?

bhaeussermann:
> This line: https://github.com/NLog/NLog/pull/1143/files#diff-530c17cacb3f2fc0522588b775ef8decR293

I somehow missed that there was a change in `AsyncTargetWrapper` :P.  
If we want to throw exactly the same exceptions as before, we will have to replace all old occurrences of `if (exc.MustBeRethrown())` with `if ((exc.MustBeRethrownImmediately()) || (exc is NLogConfigurationException))`. Perhaps create a new method for this in `ExceptionHelper`.

bhaeussermann:
> throw also exceptions if in wrapper who's handling exceptions. What do you think?

Perhaps exception-handling wrappers like `AsyncTargetWrapper` and `FallbackGroupTarget` should get a new property `rethrowExceptions` with a default value of `false` for backwards compatibility. This would also address the issue in the StackOverflow post.

304NotModified:
I was thinking to include a throwExceptions at target level (so also wrappers) . Not sure about the default value, as it up for discussion if this breaking change or just a bug fix. 

We need at least the rethrow property on event level, otherwise it's required to enable throwExceptions to get some (wrapper) targets working. Do you agree? (maybe it can be also implemented with the asyncContinution, have to check that.)

bhaeussermann:
> We need at least the rethrow property on event level

Ah yes, I see that now. So, if I understand correctly, the log event's new `ThrowException` property would be set to the config's file `RethrowExceptions` by default, but wrapper targets would set this value to `true` for all events passing through them.  
Then, before rethrowing exceptions, we check this value instead of the config's `RethrowExceptions`.

bhaeussermann:
> I was thinking to include a throwExceptions at target level

I agree. And its value could default to the config `ThrowExceptions` value. Do you agree?

304NotModified:
This is more work then expected so will move it to 4.4

304NotModified:
Move back to 4.3 because:

- the innerException of `TypeInitializationException` is not always available (dunno why) and masks the true error: see [stackoverflow](http://stackoverflow.com/questions/34994470/send-info-with-typeinitializationexception-innerexception-is-also-empty).
- Today I got two (separate) issues with NLog, and both would be easy to fix if the exception wasn't hidden. So this problem is bigger than expected.

tests/NLog.UnitTests/Internal/ExceptionHelperTests.cs
304NotModified: fixed

tests/NLog.UnitTests/Internal/ExceptionHelperTests.cs
304NotModified: good catch! fixed

tests/NLog.UnitTests/Internal/ExceptionHelperTests.cs
304NotModified: good catch! fixed

src/NLog/LogFactory.cs
304NotModified: Good suggestion. Fixed

src/NLog/Config/ConfigurationItemFactory.cs
304NotModified: Fixed with suggestion below

src/NLog/LogFactory.cs
304NotModified: I think I fixed this one earlier

src/NLog/Targets/FileTarget.cs
304NotModified: fixed

src/NLog/Config/XmlLoggingConfiguration.cs
304NotModified: indeed, fixed!

src/NLog/Config/ConfigSectionHandler.cs
304NotModified: Thanks! Fixed

src/NLog/Config/LoggingConfiguration.cs
304NotModified: fixed

src/NLog/Config/XmlLoggingConfiguration.cs
304NotModified: Good catch! 

Fixed

src/NLog/Internal/PropertyHelper.cs
304NotModified: Indeed. Fixed

src/NLog/Internal/SingleCallContinuation.cs
304NotModified: Correct again :) 

fixed

src/NLog/LogFactory.cs
304NotModified: Fixed with `MustBeRethrownImmediately` instead of `MustBeRethrown`

src/NLog/Targets/DatabaseTarget.cs
304NotModified: Well, it's is in the finally block. But this just lucky. So I applied your proposal :)

src/NLog/Targets/DatabaseTarget.cs
304NotModified: fixed

src/NLog/Targets/EventLogTarget.cs
304NotModified: lol what happend here.... end call to "throw" is removed

src/NLog/Targets/Target.cs
304NotModified: this is a bug. Fixed

304NotModified:
> Also,
> You asked me before is throwExceptions = true is only for debugging. I confirmed that, but I was wrong. For example the FallbackGroupTarget depends on it :(

I was wrong :) The `AsyncContinuation` is used for that, not a try-catch, And it seems that the exceptions are always written to `AsyncContinuation`

304NotModified:
@bhaeussermann 

If you like to check this again, thanks! I rechecked every call to `MustBeRethrown`. I changed some inconsistent test cases as I see them as bugfixes.

src/NLog/Config/Factory.cs
```diff
@@ -72,12 +72,13 @@ public void ScanTypes(Type[] types, string prefix)
                 }
                 catch (Exception exception)
                 {
+                    InternalLogger.Error(exception, "Failed to add type '{0}'.", t.FullName);
+                    
                     if (exception.MustBeRethrown())
                     {
                         throw;
                     }
```
bhaeussermann: Must this not change to `MustBeRethrownImmediately`?

src/NLog/Config/LoggingConfiguration.cs
```diff
@@ -316,12 +318,14 @@ internal void Close()
                 }
                 catch (Exception exception)
                 {
+                    InternalLogger.Warn(exception, "Exception while closing.");
+
                     if (exception.MustBeRethrown())
                     {
                         throw;
                     }
```
bhaeussermann: `MustBeRethrownImmediately()`?

src/NLog/Config/XmlLoggingConfiguration.cs
bhaeussermann: Because `exception` is an inner-exception of `configurationException` it will be written to the log with the `InternalLogger.Error()` call above, but not be marked as logged. Therefore `exception.MustBeRethrown()` will log it for the second time. Perhaps `MarkAsLoggedToInternalLogger()` must be changed to recursively mark all inner exceptions as well.

src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -936,13 +933,13 @@ private void ParseIncludeElement(NLogXmlElement includeElement, string baseDirec
             }
             catch (Exception exception)
             {
+                InternalLogger.Error(exception, "Error when including '{0}'.", newFileName);
+
                 if (exception.MustBeRethrown())
                 {
                     throw;
                 }
```
bhaeussermann: Not `MustBeRethrownImmediately()`?

src/NLog/Config/Factory.cs
304NotModified: Why?

bhaeussermann:
> >    You asked me before is throwExceptions = true is only for debugging. I confirmed that, but I was wrong. For example the FallbackGroupTarget depends on it :(

> I was wrong :) The AsyncContinuation is used for that, not a try-catch, And it seems that the exceptions are always written to AsyncContinuation

Ok, so I assume that it is then safe to throw an exception where we did not throw an exception before if `ThrowExceptions` is set to `true`. Is that correct?

src/NLog/LogFactory.cs
bhaeussermann: This must probably look like line 438 above.

tests/NLog.UnitTests/GetLoggerTests.cs
```diff
@@ -121,43 +121,32 @@ public void GenericGetCurrentClassLoggerTest()
...
-	       }
+            private InvalidLogger()
+            {
+            }
         }
-        
+
+
         [Fact]
         public void InvalidLoggerConfiguration_ThrowsConfigurationException_IfThrowExceptionsFlagIsSet()
```
bhaeussermann: The name should say `IsNotSet`.

tests/NLog.UnitTests/GetLoggerTests.cs
```diff
@@ -121,43 +121,32 @@ public void GenericGetCurrentClassLoggerTest()
...
-                ExceptionThrown = true;
-            }
-            Assert.True(ExceptionThrown);
+            });
+
         }
-        
+
         [Fact]
         public void InvalidLoggerConfiguration_ThrowsConfigurationException_IfThrowExceptionsFlagIsNotSet()
```
bhaeussermann: The name should say `IsSet()`.

tests/NLog.UnitTests/GetLoggerTests.cs
```diff
@@ -121,43 +121,32 @@ public void GenericGetCurrentClassLoggerTest()
...
         {
-            bool ExceptionThrown = false;
-            try 
+            Assert.Throws<NLogConfigurationException>(() =>
             {
                 LogManager.ThrowExceptions = true;
```
bhaeussermann: I see many tests restore this value in a `finally` block, and I believe it is good practice.

304NotModified: :)

src/NLog/Config/Factory.cs
304NotModified: I think the only reason would be backwardscomp. I think it's bug that the exception was not thrown when `throwExceptions=true`

PR #1146, opened 2016-01-09
ASP.NET 5 integration (NLog.Framework.logging)

sirentek:
Usage:

Add namespace NLog.Asp to Startup class.
Add these lines to Configure method of Startup class:

        public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory)
        {
            loggerFactory.MinimumLevel = LogLevel.Verbose;
            loggerFactory.AddConsole();
            loggerFactory.AddDebug();
            loggerFactory.AddNLog();
            ....
            ....
       }

NLog automatically finds the config file. Also the user can define it himself with a relative path:
env.ConfigureNLog("NLog.config"); (env =====>IHostEnvironment)
or
appEnv.ConfigureNLog("NLog.config"); (appEnv =====>IApplicationEnvironment)

### DIFF
--- a/NLogAdapter.cs
+++ b/NLogAdapter.cs
@@ -1,0 +1,87 @@
+using Microsoft.Extensions.Logging;
+using NLog;
+using System;
+
+using ILogger = Microsoft.Extensions.Logging.ILogger;
+using LogLevel = Microsoft.Extensions.Logging.LogLevel;
+
+namespace NLog.Asp
+{
+    public class NLogAdapter : ILogger
+    {
+        private NLog.ILogger logger;
+
+        public NLogAdapter(string loggerName)
+        {
+            logger = LogManager.GetLogger(loggerName);
+        }
+
+        public IDisposable BeginScopeImpl(object state)
+        {
+            return null;
+        }
+
+        public bool IsEnabled(LogLevel logLevel)
+        {
+            switch (logLevel)
+            {
+                case LogLevel.Verbose:
+                    return logger.IsTraceEnabled;
+                case LogLevel.Debug:
+                    return logger.IsDebugEnabled;
+                case LogLevel.Information:
+                    return logger.IsInfoEnabled;
+                case LogLevel.Warning:
+                    return logger.IsWarnEnabled;
+                case LogLevel.Error:
+                    return logger.IsErrorEnabled;
+                case LogLevel.Critical:
+                    return logger.IsFatalEnabled;
+                default:
+                    throw new ArgumentException($"Unknown log level {logLevel}.", nameof(logLevel));
+            }
+        }
+
+        public void Log(LogLevel logLevel, int eventId, object state, Exception exception, Func<object, Exception, string> formatter)
+        {
+            if (!IsEnabled(logLevel))
+            {
+                return;
+            }
+            string message = null;
+            if (null != formatter)
+            {
+                message = formatter(state, exception);
+            }
+            else
+            {
+                message = LogFormatter.Formatter(state, exception);
+            }
+            switch (logLevel)
+            {
+                case LogLevel.Verbose:
+                    logger.Trace(exception, message);
+                    break;
+                case LogLevel.Debug:
+                    logger.Debug(exception, message);
+                    break;
+                case LogLevel.Information:
+                    logger.Info(exception, message);
+                    break;
+                case LogLevel.Warning:
+                    logger.Warn(exception, message);
+                    break;
+                case LogLevel.Error:
+                    logger.Error(exception, message);
+                    break;
+                case LogLevel.Critical:
+                    logger.Fatal(exception, message);
+                    break;
+                default:
+                    logger.Warn($"Encountered unknown log level {logLevel}, writing out as Info.");
+                    logger.Info(exception, message);
+                    break;
+            }
+        }
+    }
+}
--- a/NLogProvider.cs
+++ b/NLogProvider.cs
@@ -1,0 +1,32 @@
+using Microsoft.Extensions.Logging;
+using System.Collections.Generic;
+
+namespace NLog.Asp
+{
+    public class NLogProvider : ILoggerProvider
+    {
+        private IDictionary<string, ILogger> loggers = new Dictionary<string, ILogger>();
+
+        public ILogger CreateLogger(string name)
+        {
+            if (!loggers.ContainsKey(name))
+            {
+                lock (loggers)
+                {
+                    // Have to check again since another thread may have gotten the lock first.
+                    if (!loggers.ContainsKey(name))
+                    {
+                        loggers[name] = new NLogAdapter(name);
+                    }
+                }
+            }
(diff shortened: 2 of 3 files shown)


304NotModified:
Hi! Thanks for this PR. I guess this is for asp.net 5 support?

phenixdotnet:
Yes looks like for asp.net 5.

The question is do we include our own provider or do we use the Microsoft one ?
If we use our own provider do we include it in the NLog lib or provide a separated package (Somethings like NLog.Asp ? )

sirentek:
Hi!
Yes it is for asp.net 5.
Microsoft has an interface for LogProviders. We use this interface.

I have seen similar usages here which is not for coreclr. (8 months old)
https://github.com/aspnet/Logging/blob/d874c5726e713d3eb34938f85faf7be61aae0f2a/src/Microsoft.Framework.Logging.NLog/NLogLoggerProvider.cs

and here
https://github.com/aspnet/Logging/blob/d874c5726e713d3eb34938f85faf7be61aae0f2a/src/Microsoft.Framework.Logging.Serilog/SerilogLoggerProvider.cs

and a sample for log4net. 
http://dotnetliberty.com/index.php/2015/11/09/asp-net-5-logging-with-log4net/

About your last question.. I don't have an idea.

304NotModified:
We will create a new nuget package for this,  similar to others,  see https://github.com/aspnet/Logging#providers

304NotModified:
One thing that isn't clear to me,  do we need coreclr support for ASP.NET 5?

304NotModified:
To let appveyor succeed, we need at least:

- sync the projects. See appveyor log how to fix this easily 
- write xml docs for all public stuff (methods/classes) . 

Maybe I'm old (31 ;), but "asp" refers to "classic asp" to me. So I prefer another namespace for this.  Any idea for a good candidate?

304NotModified:
PS. I will create a new repository for this, analogues to serilog and Elmah.io.

phenixdotnet:
@sirentek : Microsoft has the interface and the implementation for nLog see this : https://github.com/aspnet/Logging/tree/e7b25ca256c57d98634f9f53ecbca677cb2e0f26/src/Microsoft.Extensions.Logging.NLog

@304NotModified : I didn't get your point about naming. Currently the MS packages are named like this: Microsoft.Extensions.Logging.ProviderName. The problem is: MS already provide a package Microsoft.Extensions.NLog

Actually we can release a package with the support of logging provider without the support of coreclr. The asp.net 5 framework run on framework 4.5 and up and coreclr.

sirentek:
Microsoft's implementation is not compatible with coreclr at the moment. I think it will work with Mono framework. As I know, mono runtime is a workaround solution, coreclr is a supported framework by Microsoft.

phenixdotnet:
@sirentek : yes actually isn't compatible with coreclr but NLog isn't too ;).
Once coreclr support will be supported (or in an alpha state) we can make a PR to add support of coreclr in the MS lib. We can also work with MS / coreclr team to improve the existing package and keep it sync with the NLog package status.
IMO, having 2 packages which do the same things isn't the best solution.

sirentek:
@phenixdotnet 
Having 2 similar packages isn't good idea as  you mentioned.
Microsoft will make a much better one.
On the other hand, we don't know about when will Microsoft make it.

phenixdotnet:
@sirentek : They already have and thy are very quick to review PR.
What I suggest is:
- [ ] Release an alpha version with CoreCLR support
- [ ] Make a PR on MS repo [Logging](https://github.com/aspnet/Logging) to add support of coreclr and include your changes (I really like how you manage the config file)

sirentek:
@phenixdotnet 
How about a sample project named as AspNet5SampleApp or DnxCore50AspNetSampleApp.
This will help someone who uses AspNet and Coreclr runtime until Microsoft makes it.

phenixdotnet:
@sirentek :+1: Good idea !

304NotModified:
Great ideas guys! 

Indeed, we don't need two packages for the same. I was not aware there was already a package. 

Maybe we can write a short post how to use NLog with ASP.NET 5 on the website?

phenixdotnet:
@304NotModified : May be we need to release an alpha package before or the setup will be too complicated.

304NotModified:
Yes,  alpha package seems fine. I think that will be 4.4 as I would like to release 4.3 soon.

phenixdotnet:
@304NotModified: What do you think about a 5.0 release ? This version has major changes in source code. We don't break anythings (I hope ^^) but we introduce new platform and rework the solution organization.

304NotModified:
If there are breaking changes,  and we can't prevent them, it will be 5.0 indeed. (semver is very important) 

If it will be 5.0, then we have far more things we like to do (with breaking changes) . See milestone 5.0. I prefer to delay 5.0 (well delaying breaking changes) , because 4.0 is only 6 months ago. On the other hand support for coreclr a really important change, and the sooner the better.  

What are the breaking changes we're dealing with?

phenixdotnet:
There is no breaking changes in the code itself but the build process is completely reworked and I think we should be careful when releasing the nuget package.
The coreclr branch could live 1 or 2 months in parallel of the main branch and we can take time to release alpha, beta and other tests versions.

304NotModified:
> There is no breaking changes in the code itself but the build process is completely reworked and I think we should be careful when releasing the nuget package.
> The coreclr branch could live 1 or 2 months in parallel of the main branch and we can take time to release alpha, beta and other tests versions.

OK, then 4.4 is nice, as I try to release every 6-10 weeks a new minor release.

304NotModified:
I think this PR could be closed, isn't it?

phenixdotnet:
Yes I think too.

304NotModified:
Will close this one then

304NotModified:
Hi @phenixdotnet and @sirentek 

> @sirentek : Microsoft has the interface and the implementation for nLog see this : https://github.com/aspnet/Logging/tree/e7b25ca256c57d98634f9f53ecbca677cb2e0f26/src/Microsoft.Extensions.Logging.NLog
> 
> @304NotModified : I didn't get your point about naming. Currently the MS packages are named like this: Microsoft.Extensions.Logging.ProviderName. The problem is: MS already provide a package Microsoft.Extensions.NLog

It seems that the "Microsoft.Extensions.Logging.NLog" is [never will be released](https://github.com/aspnet/Logging/issues/292#issuecomment-160706959)! I'm asking a confirmation of this on https://github.com/aspnet/Logging/issues/339.

So proposal (dependent of Microsoft answer of course):

1. I will create a new repository, like Serilog: https://github.com/serilog/serilog-extensions-logging. 
2. You code a PR would be great (but first without CoreCLR)
3. I will release the NLog.Framework.logging

Do you also think this is a good idea?

304NotModified:
Update:

Got already response of MS :) 

We need to fix our own package.

PR #1149, opened 2016-01-10
Fix crash during delete of old archives & archive delete optimization

### LINKED ISSUE #1147: Don't try to delete/process date archives if settings prevent it
I'm getting the following exception when using archive date patterns such as `"_yyyyMMdd_HH"`

```
NLog.NLogRuntimeException: Exception occurred in NLog ---> System.ArgumentOutOfRangeException: Index and length must refer to a location within the string.
Parameter name: length
   at System.String.Substring(Int32 startIndex, Int32 length)
   at NLog.Targets.FileTarget.DeleteOldDateArchives(String pattern)
   at NLog.Targets.FileTarget.Write(LogEventInfo logEvent)
   at NLog.Targets.Target.Write(AsyncLogEventInfo logEvent)`
```

This goes through a significant amount of code starting at line [739 of FileTarget.cs](https://github.com/NLog/NLog/blob/master/src/NLog/Targets/FileTarget.cs#L739) which breaks before reaching 'EnsureArchiveCount' which is what actually deletes the file.

'EnsureArchiveCount' short-circuits immediately with `if (this.MaxArchiveFiles <= 0) return;`

I believe that it should not go into the conditional statement in line [739](https://github.com/NLog/NLog/blob/master/src/NLog/Targets/FileTarget.cs#L739) if `MaxArchiveFiles <= 0` to avoid trying to delete anything when you clearly don't need to.

brutaldev:
Fix and an optimization for  #1147

When files that match the archive format pattern are already in the logging directory, then the call to `DeleteOldDateArchives` can break trying to fetch the date part.

This only happens when files will match the layout mask, but have the matching portion shorter than the archive format. So if the string length of the archive format is 10 and the part of the match (including the extension) is less than 10 then the `SubString` on [line 1132](https://github.com/NLog/NLog/blob/master/src/NLog/Targets/FileTarget.cs#L1332) throws an exception.

fixes #1147


tests/NLog.UnitTests/Targets/FileTargetTests.cs
```diff
@@ -1723,7 +1723,54 @@ public void FileTarget_LogAndArchiveFilesWithSameName_ShouldArchive()
...
 
+                SimpleConfigurator.ConfigureForTargetLogging(ft, LogLevel.Debug);
+
+                // Create a file that will match the archive mask pattern but it's not the data format (needs to be shorter that the date format to break).
+                Directory.CreateDirectory(Path.GetDirectoryName(logFile));
+                File.Create(archiveFileLayout.Replace("{#}", "notadate")).Close();
+
+                logger.Debug("test");
+
+                AssertFileContents(logFile, "test" + ft.NewLineChars, Encoding.UTF8);
```
bhaeussermann: How about you add another assert to test that the "notadate" file still exists?

src/NLog/Targets/FileTarget.cs
```diff
@@ -1329,11 +1329,17 @@ private void DeleteOldDateArchives(string pattern)
                 foreach (string nextFile in files)
                 {
                     string archiveFileName = Path.GetFileName(nextFile);
-                    string datePart = archiveFileName.Substring(fileNameMask.LastIndexOf('*'), dateFormat.Length);
-                    DateTime fileDate = DateTime.MinValue;
-                    if (DateTime.TryParseExact(datePart, dateFormat, CultureInfo.InvariantCulture, DateTimeStyles.None, out fileDate))
+                    int lastIndexOfStar = fileNameMask.LastIndexOf('*');
+
+                    // Safety check for issue #1147.
```
bhaeussermann: It seems pretty clear that this is a safety check. Rather remove the comment.

bhaeussermann:
Thanks for this! It looks good :)  
I'm just discussing something with @304NotModified before we merge.

304NotModified:
@brutaldev 

I renamed this PR. I think this would describe the PR better?

bhaeussermann:
That PR name was not quite right. Renamed it again.

bhaeussermann:
@brutaldev `DeleteOldDateArchives()` is called from `ArchiveByDate()` as well, so as an extra optimization, could you add a check on `MaxArchiveFiles` to the beginning of `DeleteOldDateArchives()` as well?

bhaeussermann:
@brutaldev Also, to enhance readability, perhaps we should add a `private` get-only property `ShouldDeleteOldArchives` (or method) to use instead of comparing against `MaxArchiveFiles`.

brutaldev:
Made the additional changes as suggested. I made `ShouldDeleteOldArchives` a method to keep it consistent with other "Should" methods in the class even though it does not take any parameters and could be a property.

src/NLog/Targets/FileTarget.cs
```diff
@@ -1164,7 +1164,10 @@ private void ArchiveByDateAndSequence(string fileName, string pattern, LogEventI
         /// </remarks>
         private void EnsureArchiveCount(List<string> oldArchiveFileNames)
         {
-            if (this.MaxArchiveFiles <= 0) return;
+            if (this.MaxArchiveFiles <= 0)
+            {
+                return;
+            }
```
bhaeussermann: You can replace this with `if (!ShouldDeleteOldArchives())` as well.

bhaeussermann:
> I made ShouldDeleteOldArchives a method to keep it consistent with other "Should" methods in the class even though it does not take any parameters and could be a property.

Thanks! There are a few other lines that compare `MaxArchiveFiles` with 0 that can also be replaced with this method.

bhaeussermann:
@304NotModified I cannot merge the branch :(. Would you do it, please?

304NotModified:
@bhaeussermann why can't you merge? I just double checked your permissions, it should be OK. (any message visible?)

bhaeussermann:
No message. See [screenshot](http://bit.ly/1TT2WZN)

Page-Not-Found:
@304NotModified looks to be the branch is out of date and I hope that is the reason we are seeing the button disabled. I will try to take look from home today. 
capture 1

brutaldev:
@Page-Not-Found @304NotModified 
I've updated the branch, you should be able to merge now without issue.

Page-Not-Found:
@UgurAldanmaz Thanks. I was getting ready to merge this.

304NotModified:
Merge war! :P

UgurAldanmaz:
I am sorry @Page-Not-Found :( I could not see your comment.

Issue #1152, opened 2016-01-11
Cannot install NLog 4.3.0-alpha3 in Xamarin iOS Projects


SMavr:
I am trying to add the package in my Xamarin Solution using Xamarin Studio. There is not a problem adding the library to a Xamarin Android Project. But it seems that NLog can not be added to my Xamarin iOS Project. 

Package Console:

Adding NLog...
Adding 'NLog 4.3.0-alpha3' to TestApp.iOS.
Could not install package 'NLog 4.3.0-alpha3'. You are trying to install this package into a project that targets 'Xamarin.iOS,Version=v1.0', but the package does not contain any assembly references or content files that are compatible with that framework. For more information, contact the package author.

304NotModified:
can you try alpha-2 ?

304NotModified:
The nuget package looks good to me, but I did not tested it:

![image](https://cloud.githubusercontent.com/assets/5808377/12231525/993b3076-b858-11e5-8c60-7edd53aca37b.png)

SMavr:
The problem is in Xamarin Studio. I don't think that is possible to install the alpha2 version by Xamarin Studio via Nuget.

SMavr:
Shouldn't be Xamarin.iOS10 instead of MonoTouch10
http://stackoverflow.com/questions/34718144/could-not-install-package-nlog-4-3-0-alpha3-in-xamarin-ios-project?noredirect=1#comment57183286_34718144

304NotModified:
I think you're right.

I will try to create a new package today

SMavr:
Thanks!

304NotModified:
It's online! https://www.nuget.org/packages/NLog/4.3.0-alpha4

SMavr:
Thanks it works! ... I am trying to test it but I cannot I see something at the Application Output either in iOS or Android Projects. 

Inside Main:
LogManager.GetCurrentClassLogger().Info("Hello from iOS NLog");

Do I need to set any configuration?
Is the NLog.config file working for the Android and iOS projects?

304NotModified:
Yes, you need some configuration. XML config is supported for Xamarin.

Some example XML to get started with:

```xml
<?xml version="1.0" encoding="utf-8" ?>
<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd"
      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">

    <targets>
[...]

PR #1154, opened 2016-01-11
Fix nuget for Xamarin.iOs

304NotModified:
Fixes https://github.com/NLog/NLog/issues/1152

Already published as test https://www.nuget.org/packages/NLog/4.3.0-alpha4

Merge after confirmation in https://github.com/NLog/NLog/issues/1152


304NotModified:
Confirmed in #1152

Issue #1153, opened 2016-01-11
Some static methods are not thread-safe


brutaldev:
When configuring loggers in a parallel loop for example, internal collections being used by the static methods are not actually thread-safe.
```
System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
   at NLog.Config.LoggingConfiguration.ValidateConfig()
   at NLog.Config.LoggingConfiguration.InitializeAll()
   at NLog.LogFactory.ReconfigExistingLoggers()
```
In some cases internal objects can be modified which breaks other threads during configuration.
```
System.NullReferenceException: Object reference not set to an instance of an object.
   at NLog.LogFactory.GetTargetsByLevelForLogger(String name, IEnumerable`1 rules, TargetWithFilterChain[] targetsByLevel, TargetWithFilterChain[] lastTargetsByLevel, Boolean[] suppressedLevels)
   at NLog.LogFactory.GetConfigurationForLogger(String name, LoggingConfiguration configuration)
   at NLog.LogFactory.ReconfigExistingLoggers()
```
Static methods should be made thread-safe or at least advertised as not being safe to call from multiple threads.

304NotModified:
Indeed, this should be fixed

brutaldev:
Any idea how to work around this? I have this problem in a production application now and trying to figure out how to add configuration dynamically in code and call `ReconfigExistingLoggers` safely without crashing the application...

304NotModified:
If you post some current c# code, I can try to find a work around.

brutaldev:
I would if it wasn't so complicated :(

Using MEF plugins on which multiple instances are created. A parallel loop runs requesting each instance to configure it's internal settings, each of these instances configures a unique file target for logging and without the call to `ReconfigExistingLoggers` nothing is written to disk.

For now I've wrapped all of these sections in a `lock` statement with a static variable in hopes that is a temporary solution to modifying configuration and calling `ReconfigExistingLoggers` from multiple threads.

If I get some time I will add a unit test for this. Should be simple enough to fire off multiple threads which all make a modification to the rules collection and calls `ReconfigExistingLoggers`. At the moment I'm dealing with the chaos of this breaking in production.

304NotModified:
I think this can be easily fixed. I'm thinking about a patch release.

304NotModified:
This should be fixed in beta2

[...]

PR #1157, opened 2016-01-11
FIx (xml-) config classes for thread-safe issues

304NotModified:
Replacing foreach with for, for thread-safe usage

`For` forces uses to use lists.

(also `for` is most of the time more performant)

fixes https://github.com/NLog/NLog/issues/1153

### DIFF
--- a/src/NLog/Config/LoggingConfiguration.cs
+++ b/src/NLog/Config/LoggingConfiguration.cs
@@ -50,6 +50,7 @@ namespace NLog.Config
     /// Keeps logging configuration and provides simple API
     /// to modify it.
     /// </summary>
+    ///<remarks>This class is thread-safe.<c>.ToList()</c> is used for that purpose.</remarks>
     public class LoggingConfiguration
     {
         private readonly IDictionary<string, Target> targets =
@@ -238,7 +239,8 @@ namespace NLog.Config
             }
 
             this.InitializeAll();
-            foreach (IInstallable installable in this.configItems.OfType<IInstallable>())
+            var configItemsList = GetInstallableItems();
+            foreach (IInstallable installable in configItemsList)
             {
                 installationContext.Info("Installing '{0}'", installable);
 
@@ -259,6 +261,7 @@ namespace NLog.Config
             }
         }
 
+
         /// <summary>
         /// Uninstalls target-specific objects from current system.
         /// </summary>
@@ -275,7 +278,8 @@ namespace NLog.Config
 
             this.InitializeAll();
 
-            foreach (IInstallable installable in this.configItems.OfType<IInstallable>())
+            var configItemsList = GetInstallableItems();
+            foreach (IInstallable installable in configItemsList)
             {
                 installationContext.Info("Uninstalling '{0}'", installable);
 
@@ -302,7 +306,8 @@ namespace NLog.Config
         internal void Close()
         {
             InternalLogger.Debug("Closing logging configuration...");
-            foreach (ISupportsInitialize initialize in this.configItems.OfType<ISupportsInitialize>())
+            var supportsInitializesList = GetSupportsInitializes();
+            foreach (ISupportsInitialize initialize in supportsInitializesList)
             {
                 InternalLogger.Trace("Closing {0}", initialize);
                 try
@@ -340,13 +345,15 @@ namespace NLog.Config
 
             InternalLogger.Debug("--- NLog configuration dump ---");
             InternalLogger.Debug("Targets:");
-            foreach (Target target in this.targets.Values)
+            var targetList = this.targets.Values.ToList();
+            foreach (Target target in targetList)
             {
                 InternalLogger.Debug("{0}", target);
             }
 
             InternalLogger.Debug("Rules:");
-            foreach (LoggingRule rule in this.LoggingRules)
+            var loggingRules = this.LoggingRules.ToList();
+            foreach (LoggingRule rule in loggingRules)
             {
                 InternalLogger.Debug("{0}", rule);
             }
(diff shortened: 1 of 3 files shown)


Page-Not-Found:
Wouldn't it be better if you expose a lock variable and use the same during modification of the list?

304NotModified:
Good question.  Well locks are more expensive then creating objects. (CPU, lock is some kind of thread sleep)  And when the collection is changed while processing, it's isn't wrong to continue progressing the old collection.  

Or do you see it otherwise?

Page-Not-Found:
@304NotModified I don't have any other thoughts.

src/NLog/Config/LoggingConfiguration.cs
```diff
@@ -302,8 +308,10 @@ public void Uninstall(InstallationContext installationContext)
         internal void Close()
         {
             InternalLogger.Debug("Closing logging configuration...");
-            foreach (ISupportsInitialize initialize in this.configItems.OfType<ISupportsInitialize>())
+            var supportsInitializesList = this.configItems.OfType<ISupportsInitialize>().ToList();
```
Page-Not-Found: Can you create method similar to `GetInstallableItems`

src/NLog/Config/LoggingConfiguration.cs
```diff
@@ -361,10 +369,12 @@ internal void Dump()
         internal void FlushAllTargets(AsyncContinuation asyncContinuation)
         {
             var uniqueTargets = new List<Target>();
-            foreach (var rule in this.LoggingRules)
+            for (int i = 0; i < this.LoggingRules.Count; i++)
```
Page-Not-Found: Wouldn't we need to take a copy into a variable like we did in other places?

src/NLog/Config/LoggingConfiguration.cs
```diff
@@ -397,8 +411,10 @@ internal void ValidateConfig()
             // so that whenever the container is initialized its children have already been
             InternalLogger.Info("Found {0} configuration items", this.configItems.Length);
 
-            foreach (object o in this.configItems)
+            var configItemList = this.configItems.ToList();
```
Page-Not-Found: `InternalLogger.Info("Found {0} configuration items", this.configItems.Length);`
Might need to move this here. 

Why would you want to convert a array to a list? Since the for can act on a array.

src/NLog/Config/LoggingConfiguration.cs
304NotModified: Because it's already a `List` and no foreach. (so safe)

src/NLog/Config/LoggingConfiguration.cs
304NotModified: > Why would you want to convert a array to a list? Since the for can act on a array.

This is unnecessary indeed. Thanks!

src/NLog/Config/LoggingConfiguration.cs
304NotModified: Thanks. I needed a parameters for the `reverse()` in this case (performance)

src/NLog/Config/LoggingConfiguration.cs
Page-Not-Found: I was meaning to say When another thread is updating the list and you are trying to make a `for` loop using the `this.LoggingRules.Count` wouldn't it cause the same issue which we have currently?

304NotModified:
Thanks for the review @Page-Not-Found .

I applied most of your suggestions, see code comment . Please double check. If you think it's OK, just merge it :)

Page-Not-Found:
Sure. I just now added some more comments. Can you also check to see there
is no merge conflicts?

src/NLog/Config/LoggingConfiguration.cs
304NotModified: You have a good point here. We don't get the "System.InvalidOperationExceptionCollection was modified; enumeration operation may not execute.", but it's still not thread-safe

Page-Not-Found: Do you think this needs to be fixed? If not can you resolve the conflicts, I can merge this.

304NotModified: I will fix it. (assigned myself also) 

Funny,  do your see an conflict? A here it says there are no conflicts in this PR.

Page-Not-Found: Let me check later today.

Page-Not-Found:
I don't see the Merge pull request button. it is disabled.

304NotModified:
(Travis has some issues, e.g. timeouts when restoring packages, in multiple branches seen)

304NotModified:
@Page-Not-Found if you check this again, please do. 

I reverted (and squashed) some previous commutes. I use now every time a variable with `.ToList()`  (which makes a new list always, see reference source)

304NotModified:
@Page-Not-Found, you're now a gladiator? ;)

Page-Not-Found:
I am good with all of the code now. But I still prefer to use the `for` construct just because it does not create a enumerator in run time. 

One small suggestion, when you have just one line with in a `if` or `for` block. Don't need `{}`

bhaeussermann:
> One small suggestion, when you have just one line with in a if or for block. Don't need {}

That's a matter of style ;). I personally prefer omitting `{}` when not needed (to keep the code more concise), but others like to include them always. I guess it makes single-line code blocks more visually explicit and consistent with other blocks.

304NotModified:
```c
if ((err = SSLHashSHA1.update(&hashCtx, &serverRandom)) != 0)
    goto fail;
if ((err = SSLHashSHA1.update(&hashCtx, &signedParams)) != 0)
    goto fail;
    goto fail;  
if ((err = SSLHashSHA1.final(&hashCtx, &hashOut)) != 0)
    goto fail;

err = sslRawVerify(...);
```

The enormous SSL fail of Apple (2014)

Page-Not-Found:
I agree will you both. We are good to merge.

304NotModified:
Thanks for checking @Page-Not-Found!

PR #1096, opened 2015-12-18
Add support for XmlLayout 

pwelter34:
This is based on the https://github.com/loresoft/NLog.Xml project.

PR #1167, opened 2016-01-15
Add support for XmlLayout

304NotModified:
supersedes https://github.com/NLog/NLog/pull/1096

todo
- [ ] test for multiple events
- [ ] test from XML config
- [ ] non-indent test
- [ ] encoding test


Page-Not-Found:
@304NotModified Do you need any help with this?

304NotModified:
@Page-Not-Found Thanks! Well I think that's not needed (the crash was an unstable test). 

But we should discuss this feature in gitter before merging it.

Page-Not-Found:
@304NotModified IM me in gitter when your available.

304NotModified:
@Page-Not-Found I will, maybe tonight or tomorrow

Page-Not-Found:
@304NotModified are you looking or any help in this?

304NotModified:
@Page-Not-Found 

Thanks for the offer.

I had to think about this one. I think logging to XML is in some cases useful, but not with this implementation - the layout of the XML is only fixed and it's not a standard. 

We can easily create a configurable XMLLayout, like JSONLayout, and a PR like that I will accept. Maybe we can use parts of this PR for this.

Also good to know, I don't think there is a large request for this and I prefer fixing other stuff before this.

snakefoot:
Resolved with #2670 in NLog 4.6

Issue #764, opened 2015-06-18
Possible OutOfMemoryException in ColoredConsole (fix proposed)


lesserhippias:
NLog  [ NLog.3.2.0.0 ] 

When we try to make  hourly log based on this config,

![cfull](https://cloud.githubusercontent.com/assets/12625093/8232963/1367a7c8-15dc-11e5-881d-11296b63d273.png)

**After sometime we get System.OutOfMemoryException from NLog**

```
System.OutOfMemoryException: Exception of type 'System.OutOfMemoryException' was thrown.
   at System.Text.StringBuilder..ctor(String value, Int32 startIndex, Int32 length, Int32 capacity)
   at NLog.Layouts.SimpleLayout.GetFormattedMessage(LogEventInfo logEvent)
   at NLog.Targets.ColoredConsoleTarget.Write(LogEventInfo logEvent)
   at NLog.Targets.Target.Write(AsyncLogEventInfo logEvent)
   at NLog.Targets.Target.WriteAsyncLogEvent(AsyncLogEventInfo logEvent)
   at NLog.LoggerImpl.WriteToTargetWithFilterChain(TargetWithFilterChain targetListHead, LogEventInfo logEvent, AsyncContinuation onException)
   at NLog.LoggerImpl.Write(Type loggerType, TargetWithFilterChain targets, LogEventInfo logEvent, LogFactory factory)

```

304NotModified:
This is in the `ColoredConsoleTarget`

Please post the full config.

lesserhippias:
When error accored, , targets at rules are  logHourly and coloredConsole targets

![cfull](https://cloud.githubusercontent.com/assets/12625093/8232963/1367a7c8-15dc-11e5-881d-11296b63d273.png)

304NotModified:
Can you try without the `<hightight-row>`? 

The code seems inefficient over there, but I'm not sure it leaks.

304NotModified:
Is this the full config? I'm missing the `WriteToTargetWithFilterChain` from the stracktrace...?

lesserhippias:
**This is the full config.**  In the stackstrace " WriteToTargetWithFilterChain" exist but I do not know, if this exist some other things should be also in config.But stacktrace is the result of this **full config**

Actually not just remove  "highlight-row condition", remove "coloredConsole" target all. Unfortunately error can not be easily re-produced. So will start a test service will check if it again failed by System.OutOfMemoryException again....

[...]

PR #1171, opened 2016-01-17
Coloredconsole not compiled regex by default

304NotModified:
- Added `CompileRegex` option, `false` by default. If `false` the Regex Cache is used.
- Some small performance improvements

fixes https://github.com/NLog/NLog/issues/764

### DIFF
--- a/src/NLog/Targets/ConsoleWordHighlightingRule.cs
+++ b/src/NLog/Targets/ConsoleWordHighlightingRule.cs
@@ -200,7 +200,14 @@ namespace NLog.Targets
         {
             if (CompileRegex)
             {
-                return this.CompiledRegex.Replace(message, this.MatchEvaluator);
+                var regex = this.CompiledRegex;
+                if (regex == null)
+                {
+                    //empty regex so nothing todo
+                    return message;
+                }
+
+                return regex.Replace(message, this.MatchEvaluator);
             }
             //use regex cache
             var expression = GetRegexExpression();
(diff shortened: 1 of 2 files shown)


304NotModified:
Is someone would review this, that would be great :)

bhaeussermann:
How about supplementing one of the tests to test with both `true` and `false` values for this property?

src/NLog/Targets/ConsoleWordHighlightingRule.cs
```diff
@@ -157,9 +195,22 @@ private string MatchEvaluator(Match m)
             return result.ToString();
         }
 
+
         internal string ReplaceWithEscapeSequences(string message)
         {
-            return this.CompiledRegex.Replace(message, new MatchEvaluator(this.MatchEvaluator));
+            if (CompileRegex)
+            {
+                return this.CompiledRegex.Replace(message, this.MatchEvaluator);
```
bhaeussermann: Perhaps we should check if `CompiledRegex` is null here and then return just `message`.

304NotModified:
Thanks @bhaeussermann !

304NotModified:
Review notes applied!

bhaeussermann:
It looks good to me! Will you update the Wiki? Then I'll merge.

304NotModified:
done https://github.com/NLog/NLog/wiki/ColoredConsole-target

PR #1179, opened 2016-01-20
Sample app based on the .NET Core

lukasz-pyrzyk:


304NotModified:
Thanks but it should be send to the coreclr branch as the master don't have any coreclr knowledge (yet)

lukasz-pyrzyk:
Sure, i missed it :)

PR #1183, opened 2016-01-22
FileTarget: Fix compress archive file not working when using concurrentWrites="True" and keepFileOpen="True"

### LINKED ISSUE #1028: File target with  concurrentWrites="True" and enableArchiveFileCompression="True" does not function
Here is my File target specification:

     <target name="FileLogger" xsi:type="File" header = "------ Connexion v${assembly-version} ------" layout="${longdate} (${threadid}) ${level} ${event-context:eventId} [${logger}] ${share.TruncateLayout:${message}} ${exception:format=tostring}" fileName="${specialFolder:folder=CommonApplicationData}/ConnexionV14/logs/Connexion.txt" archiveFileName="${specialFolder:folder=CommonApplicationData}/ConnexionV14/logs/Connexion.{#}.zip" archiveNumbering="Date" archiveEvery="Day" archiveDateFormat="yyyyMMdd" maxArchiveFiles="10" enableArchiveFileCompression="true" concurrentWrites="true" keepFileOpen="true" encoding="iso-8859-2" />

The offending line in void ArchiveFile(string fileName, string archiveFileName, bool enableCompression) appears to here

                using (var originalFileStream = new FileStream(fileName, FileMode.Open, FileAccess.Read))

An exception occurs because another file is using it.

If I change this to

               using (var originalFileStream = new FileStream(fileName, FileMode.Open, FileAccess.Read, FileShare.ReadWrite | FileShare.Delete))
 
Which matches the original sharing specification when the file was opened. My understanding, and I know I have found documentation on this before, is that the FileShare specification needs to match the currently opened file exactly, or the share will fail.

At any rate, the exception is thrown and I am left with a bunch of 0  length zip files. The main log file never rotates.

If I remove the enableArchiveFileCompression flag, everything works, but there is no compression.

Also, do you think it would be better to perform the compression after the file has been rotated? This could be done in a background thread.

bhaeussermann:
Fixes #1028

### DIFF
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -141,7 +141,9 @@ namespace NLog.Targets
 
         private FileArchivePeriod archiveEvery;
         private long archiveAboveSize;
+#if NET4_5
         private bool enableArchiveFileCompression;
+#endif
 
         /// <summary>
         /// The filename if <see cref="FileName"/> is a fixed string
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -109,6 +109,8 @@ namespace NLog.Targets
 
         private Timer autoClosingTimer;
 
+        private Thread appenderInvalidatorThread = null;
+
         /// <summary>
         /// The number of initialised files at any one time.
         /// </summary>
@@ -137,6 +139,10 @@ namespace NLog.Targets
         /// </summary>
         private Layout archiveFileName;
 
+        private FileArchivePeriod archiveEvery;
+        private long archiveAboveSize;
+        private bool enableArchiveFileCompression;
+
         /// <summary>
         /// The filename if <see cref="FileName"/> is a fixed string
         /// </summary>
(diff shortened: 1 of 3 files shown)


bhaeussermann:
I couldn't find a way to unit-test this as the log file needs to be locked by an external process :(.

src/NLog/Targets/FileTarget.cs
```diff
@@ -1087,7 +1087,16 @@ private static void ArchiveFile(string fileName, string archiveFileName, bool en
                     }
                 }
 
-                File.Delete(fileName);
+                if (ConcurrentWrites && KeepFileOpen)
+                {
+                    // Moving the file will cause all other FileTarget's to release the file, unlocking it for deleting.
+                    string tempFileName = Path.GetTempFileName();
```
304NotModified: I think this is a hack?

bhaeussermann:
> I couldn't find a way to unit-test this as the log file needs to be locked by an external process :(.

It turns out the other `FileTarget` need not run in a separate process for the file locking to occur :). I supplemented the existing `DateArchive_AllLoggersTransferToCurrentLogFile()` test with a new `enableArchiveCompression` parameter and confirmed that it exposes the bug.

304NotModified:
Nice that you could reproduce it with an unit test :)

304NotModified:
Thank! It has been merged :)

PR #1187, opened 2016-01-22
MethodCallTarget: allow optional parameters, no nullref exceptions. +unit tests

304NotModified:


Page-Not-Found:
what does this fix?

304NotModified:
I wrote some unit tests (it was missing some), and now you can call a method with optional parameters, without send the values of the optional parameters.

304NotModified:
docs: 

- https://github.com/NLog/NLog/wiki/MethodCall-target

PR #1191, opened 2016-01-25
Throw exception when base.InitializeTarget() is not called + inline GetAllLayouts()

304NotModified:
- Target.GetAllLayouts() was useless a method (once called, private). So inlined it.
- If you write a target and forgot to call `base.InitializeTarget()`, you can get a nullRefException in `Target.PrecalculateVolatileLayouts` (called in `AsyncWrapper/BufferWrapper). The nullref is now fixed, but you still get odd behaviour, so we will check if `base.InitializeTarget()` has been called.


bhaeussermann:
Whoops! It seems after merging the test now throws `NLogRuntimeException` instead of `NLogConfigurationException`. The original exception is no longer caught and rethrown as an `NLogConfigurationException`. This is probably fine as the exception does not occur due to a configuration error. What do you think?

304NotModified:
> Whoops! It seems after merging the test now throws NLogRuntimeException instead of NLogConfigurationException. The original exception is no longer caught and rethrown as an NLogConfigurationException. This is probably fine as the exception does not occur due to a configuration error. What do you think?

I think `NLogRuntimeException` is more suitable?

304NotModified:
Changed to `NLogRuntimeException` is the unit test

PR #1194, opened 2016-01-25
Allow configuring type of parameters for the database target (including fix for Oracle)

sorvis:
Fixed issue with inserting into Oracle clobs that are larger than varchar2.

The issue stems with the Microsoft not properly mapping DbType.String into OracleDbType.Clob. NLog fails to insert records into the database that try to insert too large of an amount of data into a clob field with the following error message:

```
Error when writing to database Oracle.DataAccess.Client.OracleException ORA-01460: unimplemented or unreasonable conversion requested    at Oracle.DataAccess.Client.OracleException.HandleErrorHelper(Int32 errCode, OracleConnection conn, IntPtr opsErrCtx, OpoSqlValCtx* pOpoSqlValCtx, Object src, String procedure, Boolean bCheck)
```

Others have hit this issue before me: https://groups.google.com/forum/#!topic/nlog-users/NPtPAx43uw8

I've made the assumption that we want to assume clob after 4000, but in reality we might want to go with 32767 since that is the max in PL/SQL. Alternatively, we might want to consider an optional parameter type that can be configured in the config file so that the user gets to choose if they need something other then the default.


304NotModified:
Hi,

Thanks for this PR. We cannot merge this as this can have conflicts on other users who has a parameter with the name "OracleParameter".

I think the code can be generalized by adding a "dbtype" property to the `DatabaseParameterInfo`, something like 

 ```
 <parameter name="@message" layout="${message}" dbtype="Oracle.DataAccess.Client.OracleDbType.Clob, Oracle.DataAccess"  />
```

What do you think?

**Comment by sorvis** [comment]
Hmm, you are right if we do something like that it will be a more generic solution that can easily used for other databases. I added a DbTypePropertyName parameter as well since different databases might have a different property that needs set. I'm thinking long term if we want to merge this code it might be nice to break it out into an extension method then add some unit tests for it. What do you think?

Below is the example configuration I am now using for this parameter:
...
<parameter name="@message" layout="${message}"
                     DbType="Oracle.DataAccess.Client.OracleDbType.Clob" DbTypePropertyName="OracleDbType"/>
...

**Comment by 304NotModified** [comment]
Nice,  We are going to the right direction.  We definitely need unit tests. If you make the method internal, you can call from the unit tests. 

We are still using c#5, so please don't use c#6

The appveyor errors are net3.5 issues. See the log at the end.

**Comment by sorvis** [comment]
Ok, I wrote some tests for it and it looks like the .net 3.5 build is working now as well.

**Comment by 304NotModified** [comment]
Cool, I will review it

File: src/NLog/Targets/DatabaseParameterInfo.cs [hunk 1]
```diff
@@ -98,6 +98,20 @@ public DatabaseParameterInfo(string parameterName, Layout parameterLayout)
...
+        /// <docgen category='Parameter Options' order='10' />
+        [DefaultValue(null)]
+        public string DbType { get; set; }
+
+        /// <summary>
+        /// Gets or sets the name of the database column type property.
+        /// </summary>
+        /// <docgen category='Parameter Options' order='10' />
+        [DefaultValue(null)]
+        public string DbTypePropertyName { get; set; }
```
304NotModified: I'm not fully understand why we need this?

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -512,6 +515,36 @@ private void WriteEventToDatabase(LogEventInfo logEvent)
             }
         }
 
+        /// <summary>
+        /// Allows the database type to be customized if we
+        /// are not able to properly infer the database type.
+        /// </summary>
+        /// <param name="dbDataParameter">The parameter.</param>
+        /// <param name="parameterInfo"></param>
+        internal void CustomizeParameterType(IDbDataParameter dbDataParameter, DatabaseParameterInfo parameterInfo)
```
304NotModified: Proposal: rename to "SetParamValue" and add `p.Value = stringValue` (line 504) in this method

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -512,6 +515,36 @@ private void WriteEventToDatabase(LogEventInfo logEvent)
...
+        {
+            if (string.IsNullOrEmpty(parameterInfo.DbType) ||
+                string.IsNullOrEmpty(parameterInfo.DbTypePropertyName))
+                return;
+            var fullEnumNamespace = parameterInfo.DbType.Split('.');
+            var justEnumName = string.Join(".", fullEnumNamespace.Take(fullEnumNamespace.Length - 1).ToArray());
+            var justEnumValue = fullEnumNamespace.Last();
```
304NotModified: are we trying to get the value of the after the last `.` ? e.g. x.y.z.GETTHISVALUE ? 

We can write this less complicated and more performant, use get `.LastIndexOf `

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -512,6 +515,36 @@ private void WriteEventToDatabase(LogEventInfo logEvent)
...
+            var property = paramType.GetProperty(parameterInfo.DbTypePropertyName);
+            if (!string.IsNullOrEmpty(justEnumName))
+            {
+                var oracleType = paramType.Assembly.GetType(justEnumName);
+                var oracleTypeField = oracleType != null ? oracleType.GetField(justEnumValue) : null;
+                oracleTypeFieldRawConstantValue = oracleTypeField != null ? oracleTypeField.GetRawConstantValue() : null;
+            }
+
+            if (oracleTypeFieldRawConstantValue == null || property == null)
+                throw  new Exception($"Unable set the database type from the database parameter for DbType: {parameterInfo.DbType} {Environment.NewLine} and DbTypePropertyName of: {parameterInfo.DbTypePropertyName}");
```
304NotModified: please no c#6

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -512,6 +515,36 @@ private void WriteEventToDatabase(LogEventInfo logEvent)
...
+        internal void CustomizeParameterType(IDbDataParameter dbDataParameter, DatabaseParameterInfo parameterInfo)
+        {
+            if (string.IsNullOrEmpty(parameterInfo.DbType) ||
+                string.IsNullOrEmpty(parameterInfo.DbTypePropertyName))
+                return;
+            var fullEnumNamespace = parameterInfo.DbType.Split('.');
+            var justEnumName = string.Join(".", fullEnumNamespace.Take(fullEnumNamespace.Length - 1).ToArray());
+            var justEnumValue = fullEnumNamespace.Last();
+
+            object oracleTypeFieldRawConstantValue = null;
```
304NotModified: remove the Oracle names (multiple)

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -512,6 +515,36 @@ private void WriteEventToDatabase(LogEventInfo logEvent)
...
+            Type paramType = dbDataParameter.GetType();
+            var property = paramType.GetProperty(parameterInfo.DbTypePropertyName);
+            if (!string.IsNullOrEmpty(justEnumName))
+            {
```
304NotModified: can you add some comments what we are trying to archive here?

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -31,6 +31,8 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
+using System.Linq;
```
304NotModified: move to other `usings`, thanks!

tests/NLog.UnitTests/Targets/DatabaseTargetTests.cs
```diff
@@ -758,6 +759,76 @@ public void OdbcShorthandNotationTest()
             Assert.Equal(typeof(System.Data.Odbc.OdbcConnection), dt.ConnectionType);
         }
 
+        [Fact]
```
304NotModified: Looks good. Can you add a testcase with XML config?

tests/NLog.UnitTests/Targets/DatabaseTargetTests.cs
```diff
@@ -758,6 +759,76 @@ public void OdbcShorthandNotationTest()
...
+        public void CustomizeParameterTypeShouldSetCustomEnumValueTest()
+        {
+            var expected = MockDbParameter.CustomDbType.Clob;
+            var dt = new DatabaseTarget
```
304NotModified: please us readable names, so no "dt", 'ie", "ei" etc

tests/NLog.UnitTests/Targets/DatabaseTargetTests.cs
```diff
@@ -1005,6 +1077,19 @@ public MockDbParameter(MockDbCommand mockDbCommand, int paramId)
                 this.paramId = paramId;
             }
 
+            public CustomDbType CustomDbColumnType
```
304NotModified: this could an auto property?

304NotModified:
Added some notes!

src/NLog/Targets/DatabaseParameterInfo.cs
sorvis: Oracle uses a different property on its object for its custom database type then the property implemented by IDbDataParameter. Perhaps there is a better way to figure out the property we need to set. One thought I have is we could use reflection to find a property that matches what ever DbType is set to then set the value to that property. Do you have any suggestions?

src/NLog/Targets/DatabaseTarget.cs
sorvis: Hmm, I'm not quite understanding this one. We are working with the database column type not the value here. What do you think about "SetParamDbType" for the method name?

sorvis:
OK, I've gone through the code suggestions you made and made updates. I also left a couple comments above on the comments you made.

Also I see the Travis build is failing with the following error:
```
NLog.NLogConfigurationException : Error loading extensions: /home/travis/build/NLog/NLog/SampleExtensions.dll
```

I'm not exactly sure what I did to produce that error message. Do you have any suggestions?

304NotModified:
Travis is sometimes a bit unstable.  

I'm off for a few days (holiday), so I will continue on this one next week

src/NLog/Targets/DatabaseParameterInfo.cs
304NotModified: I think we should create this case in the c# code, not config. e.g.

pseudocode:
```
if (oracle) 
  set property X
else 
  set ...
```

It's maybe a bit less extensible, but it's far more easier to configure. What do you think?

sorvis: Now that we have it generic it bothers me a bit to go back to having it hard coded for Oracle. What do you think about using reflection to find a public writable property of the given enum type then setting that property equal to the enum? That way we could eliminate the DbTypePropertyName configuration making configuration simpler. I'm thinking of doing it similar to how they set properties in this stack overflow answer: http://stackoverflow.com/a/824854/870152

304NotModified: I get your point. But it's still more generic as I can set the parameter type for every db vendor from the config, and it's not limited to clobs. Also, I think most of the db vendors uses the 
IDbDataParameter and not a special property.  

The propery name from the config is a bit difficult too understand, so I prefer for that a special case. 

Scanning the properties sounds a bit magic and performance problematic.  Do you agree?

sorvis: True it is a bit magic and reflection does impact performance. I can change it up to just hard code the needed value for just Oracle for now like you did in your pseudo code, Then in the future a code change will be required to support any other special cases like this. I'll update my code in a day or two when I next get the chance.

304NotModified: OK great!

sorvis:
Ok, I've updated the code so that it hard codes the property it will set for oracle. One problem I see is I created a enum with the same namespace as the real Oracle enum to use in testing, but the AppVeyor build failed because of that. Ironically, the Travis build passed. Do you have any suggestions of how we can make the build work for AppVeyor? Or, is there perhaps a better way to write a test for this?

304NotModified:
> C:\projects\nlog\tests\SampleExtensions\SampleExtensions.sl4.csproj]
  SampleExtensions.sl4 -> C:\projects\nlog\build\bin\Debug\Silverlight 4.0\SampleExtensions.dll
Targets\DatabaseTargetTests.cs(34,7): error CS0246: The type or namespace name 'Oracle' could not be found (are you missing a using directive or an assembly reference?) [C:\projects\nlog\tests\NLog.UnitTests\NLog.UnitTests.sl4.csproj]


it's building the unit test for silverlight (that one won't run on Linux ;))

I would recommend to exclude the unit test for silverlight. (`#if !SILVERLIGHT`)

304NotModified:
Any idea if I can include this in 4.3? (freeze this week)?

sorvis:
OK, after a bit of wrestling I got the build working. We should be good to go!

304NotModified:
Well almost, as this only works for Oracle now. For other types, we should fallback to `DataParameter.DbType` ?

And I think we need some extra checking and some last renames. Will add those to the code comments

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -522,6 +524,55 @@ private void WriteEventToDatabase(LogEventInfo logEvent)
...
+        {
+            if (string.IsNullOrEmpty(parameterInfo.DbType))
+                return;
+            var lastIndexOfDot = parameterInfo.DbType.LastIndexOf('.');
+
+            // Get just the fully qualified name of the enum but not the value
+            var justEnumName = lastIndexOfDot > 0 ? 
+                parameterInfo.DbType.Substring(0, lastIndexOfDot) : string.Empty;
+
+            object rawConstantValue = null;
```
304NotModified: I'm not sure what `rawConstantValue ` is now. Can give this another name?

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -522,6 +524,55 @@ private void WriteEventToDatabase(LogEventInfo logEvent)
...
+        /// <param name="dbDataParameter">The parameter.</param>
+        /// <param name="parameterInfo"></param>
+        internal void SetParamType(IDbDataParameter dbDataParameter, DatabaseParameterInfo parameterInfo)
+        {
+            if (string.IsNullOrEmpty(parameterInfo.DbType))
+                return;
+            var lastIndexOfDot = parameterInfo.DbType.LastIndexOf('.');
+
+            // Get just the fully qualified name of the enum but not the value
+            var justEnumName = lastIndexOfDot > 0 ?
```
304NotModified: proposal: `justEnumName ` =>  `enumName`

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -522,6 +524,55 @@ private void WriteEventToDatabase(LogEventInfo logEvent)
...
+            // Get just the fully qualified name of the enum but not the value
+            var justEnumName = lastIndexOfDot > 0 ? 
+                parameterInfo.DbType.Substring(0, lastIndexOfDot) : string.Empty;
+
+            object rawConstantValue = null;
+            Type paramType = dbDataParameter.GetType();
+            if (!string.IsNullOrEmpty(justEnumName))
+            {
+                // get the enum type from the database parameter we are using
+                var databaseType = paramType.Assembly.GetType(justEnumName);
```
304NotModified: proposal: `databaseType  ` =>  `enumType` or `enumDatabaseType`

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -522,6 +524,55 @@ private void WriteEventToDatabase(LogEventInfo logEvent)
...
+
+                // get the enum value for the database type
+                var justEnumValue = parameterInfo.DbType.Substring(lastIndexOfDot + 1,
```
304NotModified: proposal: `justEnumValue  ` =>  `enumValue`

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -522,6 +524,55 @@ private void WriteEventToDatabase(LogEventInfo logEvent)
...
+
+            // Each custom database type that we want to support will need added here on as needed basis
+            string propertyName = string.Empty;
+            if (parameterInfo.DbType.Contains("OracleDbType"))
+            {
+                propertyName = "OracleDbType";
+            }
+
+            // find the property on the database specific parameter
+            var property = paramType.GetProperty(propertyName);
```
304NotModified: this will fail with `propertyName == ""`. Please check first

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -522,6 +524,55 @@ private void WriteEventToDatabase(LogEventInfo logEvent)
...
+
+            // if something went wrong with the reflection a bad value was probably set and we can't fix that
+            if (rawConstantValue == null || property == null)
+                throw new Exception("Unable set the database type from the database parameter for DbType: "
```
304NotModified: please replace with `NLogConfigurationException`

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -522,6 +524,55 @@ private void WriteEventToDatabase(LogEventInfo logEvent)
...
+
+                // get the enum value for the database type
+                var justEnumValue = parameterInfo.DbType.Substring(lastIndexOfDot + 1,
+                    parameterInfo.DbType.Length - lastIndexOfDot - 1);
+                var typeField = databaseType != null ? databaseType.GetField(justEnumValue) : null;
+                rawConstantValue = typeField != null ? typeField.GetRawConstantValue() : null;
+            }
+
+            // Each custom database type that we want to support will need added here on as needed basis
+            string propertyName = string.Empty;
```
304NotModified: init with `null` is I preferred

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -522,6 +524,55 @@ private void WriteEventToDatabase(LogEventInfo logEvent)
             }
         }
 
+        /// <summary>
+        /// Allows the database type to be customized if we
```
304NotModified: Proposal: set database type on <paramref=dbDataParameter>

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -509,6 +510,7 @@ private void WriteEventToDatabase(LogEventInfo logEvent)
                     string stringValue = par.Layout.Render(logEvent);
 
                     p.Value = stringValue;
+                    SetParamType(p, par);
```
304NotModified: We can pass here `par.DbType` instead of `par`?

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -522,6 +524,55 @@ private void WriteEventToDatabase(LogEventInfo logEvent)
...
+                    parameterInfo.DbType.Length - lastIndexOfDot - 1);
+                var typeField = databaseType != null ? databaseType.GetField(justEnumValue) : null;
+                rawConstantValue = typeField != null ? typeField.GetRawConstantValue() : null;
+            }
+
+            // Each custom database type that we want to support will need added here on as needed basis
+            string propertyName = string.Empty;
+            if (parameterInfo.DbType.Contains("OracleDbType"))
+            {
+                propertyName = "OracleDbType";
```
304NotModified: Add comments; for Oracle we have to set the `OracleDbType` property on the parameter

tests/NLog.UnitTests/Targets/DatabaseTargetTests.cs
```diff
@@ -758,6 +759,98 @@ public void OdbcShorthandNotationTest()
             Assert.Equal(typeof(System.Data.Odbc.OdbcConnection), dt.ConnectionType);
         }
 
+        [Fact]
+        public void Oracle_SetParamTypeShouldSetCustomEnumValueTest()
```
304NotModified: Can you clone this test case and edit it for SQL server? Use [SqlParameter](https://msdn.microsoft.com/en-us/library/system.data.sqlclient.sqlparameter(v=vs.110).aspx)

tests/NLog.UnitTests/Targets/DatabaseTargetTests.cs
```diff
@@ -1005,11 +1108,9 @@ public MockDbParameter(MockDbCommand mockDbCommand, int paramId)
                 this.paramId = paramId;
             }
 
-            public DbType DbType
-            {
-                get { return this.parameterType; }
-                set { this.parameterType = value; }
-            }
+            public OracleDbType OracleDbType { get; set; }
```
304NotModified: Add comments; property for Oracle as we don't have the Oracle types here. 

(or something like that)

src/NLog/Targets/DatabaseParameterInfo.cs
```diff
@@ -98,6 +98,14 @@ public DatabaseParameterInfo(string parameterName, Layout parameterLayout)
         /// <docgen category='Parameter Options' order='10' />
         [DefaultValue(0)]
         public byte Scale { get; set; }
+
+        /// <summary>
+        /// Gets or sets the column type of the database.
```
304NotModified: Add : `if <c>null</c> then string/varchar type is used`

304NotModified:
Added some comments. Mostly names and docs.

src/NLog/Targets/DatabaseTarget.cs
sorvis: I have an exception being thrown for that case with a unit test checking for it to do that. If we fallback to IDataParameter.DbType won't it just cover up a miss configuration instead of helping the user find the issue?

sorvis:
I've updated the code with your suggestions. I left a comment on your comment about falling back to IDataParameter.DbType. I think I miss-understood you then, but I think I see now you meant to just let them set the DbType field through the interface. Is that what you had in mind? Is there a case when a non-custom datatype would need to be manually set?

304NotModified:
> I've updated the code with your suggestions. I left a comment on your comment about falling back to IDataParameter.DbType. I think I miss-understood you then, but I think I see now you meant to just let them set the DbType field through the interface. Is that what you had in mind? Is there a case when a non-custom datatype would need to be manually set?

What I mean is that the submitting code will work for Oracle, but not for other DB-types like SQL server / mysql / mariaDB. SQL server is using `IDataParameter.DbType` (and mysql/mariaDB also I guess).

So if use the property for Oracle, and if not Oracle ("the fallback"),  `DataParameter.DbType`, then we it works for all db-types!

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -522,6 +524,68 @@ private void WriteEventToDatabase(LogEventInfo logEvent)
...
+            string propertyName = null;
+            if (dbType.Contains("OracleDbType")) 
+            {
+                // for Oracle we have to set the OracleDbType property on the parameter
+                propertyName = "OracleDbType";
+            }
+            else if(dbType.Contains("SqlDbType"))
+            {
+                // for Sql Server we have to set the SqlDbType property on the parameter
+                propertyName = "SqlDbType";
```
304NotModified: All others use DbType (the interface). Are you sure Oracle can't use DbType?

sorvis:
Yes, setting DbType property on the Oracle parameter will set the OracleDbType property. The problem is however is SqlDbType does not contain all the values OracleDbType contains. So if you take a look at this [mapping chart](https://msdn.microsoft.com/en-us/library/yk72thhd%28v=vs.110%29.aspx). There is no mapping from DbType into an Oracle clob.	

> Oracle only allows binding a CLOB as a CLOB parameter. The .NET Data Provider for Oracle does not automatically construct one for you if you attempt to bind a non-CLOB value, such as System.String or OracleString.

For SqlDbType it appears that Microsoft has covered the entire [mapping though](https://msdn.microsoft.com/en-us/library/cc716729%28v=vs.110%29.aspx). I'm thinking we need a mix of both setting DbType without reflection and still supporting reflection for OracleDbType. What do you think?

304NotModified:
> The problem is however is SqlDbType does not contain all the values OracleDbType contains.

You're right on this

> For SqlDbType it appears that Microsoft has covered the entire mapping though. I'm thinking we need a mix of both setting DbType without reflection and still supporting reflection for OracleDbType. What do you think?

Yes then we need both, two things important:

- Can we detect with the type with one to set? (PrvDbType vs DbType )?
- I prefer to make the PrvDbType  more generic: detect if db (sql server, oracle) and then get prefix. it that something we can do?

sorvis:
> Can we detect with the type with one to set? (PrvDbType vs DbType )?
Since we are making the DbType configuration be the full namespace of the enum along with its value then we should be able to know if its a Microsoft DbType.

> I prefer to make the PrvDbType more generic: detect if db (sql server, oracle) and then get prefix. it that something we can do?
One thing I did notice was at least in the case of Oracle the property name was the same as the enum type. So the enum is named OracleDbType and the property on the parameter object is also named that. We could make that the default behavior if the type is not DbType. Then later if another provider does not follow the same pattern we would need a code change, but in the mean time it would have a decent chance of working.

304NotModified:
> One thing I did notice was at least in the case of Oracle the property name was the same as the enum type. So the enum is named OracleDbType and the property on the parameter object is also named that. We could make that the default behavior if the type is not DbType. Then later if another provider does not follow the same pattern we would need a code change, but in the mean time it would have a decent chance of working.

Can you give an example for this? I don't fully understand it. Thanks!

sorvis:
Roughly we would do something like this. Assuming dbType is set as follows we could set the propertyName in some fashion like this:

```
dbType =  "System.Data.OracleClient.OracleType.Clob";
propertyName = dbType.Split['.'][3];
```

This will work for oracle because on the OracleParameter class the OracleType property is named OracleType. SqlParameter works the same way in that it has a property called SqlDbType and the type is SqlDbType--even though as we found out this doesn't matter for Sql Server since Microsoft already mapped everything. I'm not sure if other database providers follow the same pattern, but if they do this could be a generic solution.

304NotModified:
I understand your proposal.

I think it's too trick - we don't know how this behaves in all situations. 

My proposal (pseudo code)

```c#

function FindDatabaseTypePropertyPrefix(fullType)
{
   if(is oracle type) 
     return "Oracle"
   if( is SQL type)
     return "Sql"
   return null
 }


function SetParamType
{
  var databaseTypePropertyPrefix = FindDatabaseTypePropertyPrefix(fullnameOfType)

  if(databaseTypePropertyPrefix  != null)
     set property with databaseTypePropertyPrefix 
  else
    set interface  IDataParameter.DbType
}

**Comment by sorvis** [comment]
That seems like it should work. One thing we need to figure out though is if we are allowing them to configure a value directly on the IDataParameter.DbType property how will they specify it in the configuration? Will we make them directly spell out the full namespace of the enum and value like how we made the configuration work for the others? Or, should we make it so they can just give the value such as "DateTime" and then we would just assume its a value from the DbType enum.

**Comment by 304NotModified** [comment]
I would try first the dbtype and otherwise the specialised.  If the full name is used, then no fallback. 

I have to agree that is some code is needed for that. We should create a static method for that and unit test it. 

We could also start with only allowing the full name and allow short names in the future.

**Comment by sorvis** [comment]
OK, I gave that a try setting the DbType first then going into the custom logic for setting the property on Oracle. If you like what I've done I can go further and make it work for short names for DbType.

**Comment by 304NotModified** [comment]
:+1:

**Comment by 304NotModified** [comment]
Can I review/test this one?

**Comment by sorvis** [comment]
Yes, please review this and test it. Let me know if we need any more changes and I'll try to work them in next week.

File: src/NLog/Targets/DatabaseTarget.cs [hunk 25]
```diff
@@ -522,6 +524,82 @@ private void WriteEventToDatabase(LogEventInfo logEvent)
...
+
+                    // get the enum value for the database type
+                    var enumValueString = dbType.Substring(lastIndexOfDot + 1,
+                        dbType.Length - lastIndexOfDot - 1);
+                    var typeField = enumDatabaseType != null ? enumDatabaseType.GetField(enumValueString) : null;
+                    enumValueObject = typeField != null ? typeField.GetRawConstantValue() : null;
+                }
+                catch (Exception exception)
+                {
+                    InternalLogger.Trace(exception, "DatabaseTarget: unable to use " + enumName + " to set database column type.");
```
304NotModified: We have new exception handling pattern in NLog 4.3 Please check: https://github.com/NLog/NLog/pull/1143#issue-125708489

I think this is a `warn`, also you can use `params `, so
`"DatabaseTarget: unable to use {0}" to set database column type.", enumName `

thanks!

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -522,6 +524,82 @@ private void WriteEventToDatabase(LogEventInfo logEvent)
...
+            //if its a DbType then we can just set it
+            if (enumFullName == typeof(DbType).FullName)
+            {
+                dbDataParameter.DbType = (DbType)enumValueObject;
+            }
+            else
+            {
+                // Each custom database type that we want to support will need added here on as needed basis
+                string propertyName = null;
+                if (enumFullName.Contains("OracleDbType"))
```
304NotModified: Can we search for `.OracleDbType.` (included two dots)?

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -522,6 +524,82 @@ private void WriteEventToDatabase(LogEventInfo logEvent)
...
+                {
+                    // for Oracle we have to set the OracleDbType property on the parameter
+                    propertyName = "OracleDbType";
+                }
+                else 
+                {
+                    // we only support additional database types by custom implementation
+                    throw new NLogConfigurationException("Custom setting of the database type is not supported for: " +
```
304NotModified: I think this should be a warn also (no exception throwing), OR error

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -522,6 +524,82 @@ private void WriteEventToDatabase(LogEventInfo logEvent)
...
+                                                     dbType);
+                }
+
+                // find the property on the database specific parameter
+                var property = dbDataParameter.GetType().GetProperty(propertyName);
+
+                // if something went wrong with the reflection a bad value was probably set and we can't fix that
+                if (enumValueObject == null || property == null)
+                    throw new NLogConfigurationException("Unable set the database type from the database parameter for DbType: "
```
304NotModified: I think this should be a warn also (no exception throwing), OR error

tests/NLog.UnitTests/Targets/DatabaseTargetTests.cs
```diff
@@ -759,6 +761,120 @@ public void OdbcShorthandNotationTest()
...
+            var target = new DatabaseTarget
+            {
+                Name = "notimportant",
+                DBProvider = "notimportant",
+                ConnectionString = "notimportant",
+                CommandText = "notimportant",
+            };
+            var parameterInfo = new DatabaseParameterInfo
+            {
+                DbType = typeof(DbType).FullName + "." + expected,
```
304NotModified: I prefer to write the full string out, so we the unit test is kinda docs. Agree?

304NotModified:
Hi @sorvis !

Looks good! There is some new exception handling in NLog 4.3, we need to adopt this in this PR (see my comments).

The code looks good! 

And I think it would be nice to add one more unit test, when `DbType ` isn't set. Backwards-compatibility is very very important . If we done that, I will merge it :)

Thanks in advance!

sorvis:
Hi @304NotModified 

I've updated the code to use the new method for exception handling.

One concern I have. When I do this:
```
InternalLogger.Error(exception, "Custom setting of the database type is not supported for: {0}", dbType);
{
    throw exception;
}
```
I'm not sure exactly how the test for Oracle_SetParamTypeShouldThrowExceptionsOrEatThemTest should be written. If I use an Assert for expected exception sometimes it does not throw, but other times it does.

I ended up just doing this, but it really proves nothing:
```
try
{
    target.SetParamType(parameter, parameterInfo.DbType);
}
catch (Exception exception)
{
    Assert.Contains(dbType, exception.Message);
}
```

304NotModified:
> I'm not sure exactly how the test for Oracle_SetParamTypeShouldThrowExceptionsOrEatThemTest should be written. If I use an Assert for expected exception sometimes it does not throw, but other times it does.

You should use `Assert.ThrowException<..>` and `LogManager.ThrowExceptions = false|true`

sorvis:
Perfect _LogManager.ThrowExceptions_ was exactly what I needed. I fixed my unit test. I think we should be good to go.

304NotModified:
Hi,

was testing this and there were some small problems (beside that I think this PR should be more complete - so support all database types).

I made some changes. Will send them to you, but for now I'm missing the needed time for this. So will move this to 4.3.1

sorvis:
Hi,

Ok, thanks for the update. Let me know when you are ready to start working on it again and I'll help where I can.

304NotModified:
I gonna work on this now.

304NotModified:
superseded by https://github.com/NLog/NLog/pull/1435 (but not done yet)

PR #1207, opened 2016-01-30
FileTarget: Fix Footer for archiving.

bhaeussermann:
The Footer property did not add the footer to archive files. Header is added to each archive, so one expects the footer to work the same way.

### DIFF
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -188,7 +188,7 @@ namespace NLog.Targets
             this.OpenFileCacheTimeout = -1;
             this.OpenFileCacheSize = 5;
             this.CreateDirs = true;
-            this.fileArchive = new DynamicFileArchive(MaxArchiveFiles);
+            this.fileArchive = new DynamicFileArchive(this, MaxArchiveFiles);
             this.ForceManaged = false;
             this.ArchiveDateFormat = string.Empty;
 
@@ -711,7 +711,7 @@ namespace NLog.Targets
             // Uninitialize the files.
             foreach (string fileName in filesToUninitialize)
             {
-                this.WriteFooterAndUninitialize(fileName);
+                this.UninitializeFile(fileName);
             }
         }
 
@@ -820,7 +820,7 @@ namespace NLog.Targets
 
             foreach (string fileName in new List<string>(this.initializedFiles.Keys))
             {
-                this.WriteFooterAndUninitialize(fileName);
+                this.UninitializeFile(fileName);
             }
 
             if (this.autoClosingTimer != null)
@@ -855,10 +855,7 @@ namespace NLog.Targets
             byte[] bytes = this.GetBytesToWrite(logEvent);
 
             if (this.ShouldAutoArchive(fileName, logEvent, bytes.Length))
-            {
-                this.fileAppenderCache.InvalidateAppender(fileName);
                 this.DoAutoArchive(fileName, logEvent);
-            }
 
             // Clean up old archives if this is the first time a log record is being written to
             // this log file and the archiving system is date/time based.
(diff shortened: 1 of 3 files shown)


Page-Not-Found:
@bhaeussermann would you mind adding some more details on the PR so that I can get some context to do the review.

bhaeussermann:
> @bhaeussermann would you mind adding some more details on the PR so that I can get some context to do the review. 

Done.

304NotModified:
Are there 2 fixes (footer + header) is this PR? Or do I read it wrong?

bhaeussermann:
> Are there 2 fixes (footer + header) is this PR? Or do I read it wrong?

This PR fixes footer only; the header was already working as it should.

304NotModified:
Thanks for the info @bhaeussermann. 

I was confused by

http://nlog.userecho.com/topic/1047997-add-headersfooters-to-archive-files/

I think we should create a separate github issue for the userecho post.

bhaeussermann:
> I think we should create a separate github issue for the userecho post. 

There's something fishy about that post. We already add the header to each archive file, and it looks like its been that way for quite a while. This is shown by RepeatingHeaderTest() in [FileTargetTests.cs](https://github.com/NLog/NLog/blob/master/tests/NLog.UnitTests/Targets/FileTargetTests.cs) .

304NotModified:
> We already add the header to each archive file, and it looks like its been that way for quite a while. This is shown by RepeatingHeaderTest() in FileTargetTests.cs .

Indeed. https://github.com/NLog/NLog/blob/119cd8045f6500671e35e05996aaa4e7165f8302/tests/NLog.UnitTests/Targets/FileTargetTests.cs#L1250-L1296 is a good example

Page-Not-Found:
I don't really any comments for this. I guess this can be merged.

src/NLog/Targets/FileTarget.cs
```diff
@@ -1058,22 +1052,21 @@ private void RollArchivesForward(string fileName, string pattern, int archiveNum
             string newFileName = ReplaceNumberPattern(pattern, archiveNumber);
             RollArchivesForward(newFileName, pattern, archiveNumber + 1);
 
-            var shouldCompress = archiveNumber == 0;
-            try
-            {
-                RollArchiveForward(fileName, newFileName, shouldCompress);
-            }
-            catch (IOException)
```
304NotModified: I have one question: why has this changed to two method and why is the `catch` removed?

304NotModified:
It shows merge conflicts here. @Page-Not-Found also with you?

image
image

Page-Not-Found:
@304NotModified it does.

src/NLog/Targets/FileTarget.cs
bhaeussermann: > why has this changed to two method

This is a slight refactoring. See the change in the `RollArchiveForward()` method. This method was previously called to "move archive files forward" (when `shouldCompress` is `false`), but also to archive log files (when `shouldCompress` is `true`). A method must have only one clear responsibility and I want us to make a clear distinction between "rolling forward" and archiving. Therefore I specialized this method to cater for moving archive files forward only. When archiving a log file, we should call `ArchiveFile()` instead.

As a side-effect of this change, the two method calls here make it explicit that we archive the first file, but "roll" all of the rest.

bhaeussermann: > why is the catch removed?

I believe the intention of the catch is to create the directory if it does not exist and then retry.

There are two cases: 1) calling `ArchiveFile()` when `archiveNumber == 0` and 2) calling `RollArchiveForward()` otherwise:  
1) `ArchiveFile()` already creates the directory if it does not exist. 
2) `RollArchiveForward()` is called for archive files only. Here `fileName` is an existing archive file, and `newFileName` is its new file path. But all archive files are in the same directory, so the new file path's directory already exists.

bhaeussermann:
Xamarin license error :(

304NotModified:
> Xamarin license error :(

nooos!

reset and restarted

src/NLog/Targets/FileTarget.cs
304NotModified: > RollArchiveForward() is called for archive files only. Here fileName is an existing archive file, and newFileName is its new file path. But all archive files are in the same directory, so the new file path's directory already exists.

but the folder/file can be deleted already?

bhaeussermann: > but the folder/file can be deleted already?

We test that the file exists on line 1047.

304NotModified: OK thanks!

PR #1208, opened 2016-01-30
FileTargetTests: Supplemented ReplaceFileContentsOnEachWriteTest() to test with and without header and footer

bhaeussermann:


304NotModified:
Thanks! 74%!! :)

PR #1222, opened 2016-02-05
internalLogger, write to System.Diagnostics.Debug / System.Diagnostics.Trace #1217

### LINKED ISSUE #1217: internalLogger, write to System.Diagnostics.Debug / System.Diagnostics.Trace
Allow InternalLogger write to `System.Diagnostics.Debug` / `System.Diagnostics.Trace`

It should be one option (WriteToDiagnostics?). When enabled:
- trace messages to `System.Diagnostics.Trace`
- others to `System.Diagnostics.Debug`

bryjamus:
fixes #1217


304NotModified:
Thanks for this PR! Will review it!

src/NLog/Common/InternalLogger.cs
```diff
@@ -198,7 +203,7 @@ public static void Log(Exception ex, LogLevel level, [Localizable(false)] string
         /// <param name="args">optional args for <paramref name="message"/></param>
         private static void Write([CanBeNull]Exception ex, LogLevel level, string message, [CanBeNull]object[] args)
         {
-            if (ex != null && ex.MustBeRethrownImmediately())
+            if (IsSeriousException(ex))
```
304NotModified: I think this is a bit overkill, a new method, as it used only once and it's only 2 statements

src/NLog/Common/InternalLogger.cs
```diff
@@ -209,14 +214,14 @@ private static void Write([CanBeNull]Exception ex, LogLevel level, string messag
                 return;
             }
 
-            if (string.IsNullOrEmpty(LogFile) && !LogToConsole && !LogToConsoleError && LogWriter == null)
+            if (NoLogger())
```
304NotModified: Although the method is also used once, It has a lot of statements. It's OK to create a new method, but please document the method (xmldocs)

Proposal for the name: "loggingEnabled" (and maybe a property is nicer)?

src/NLog/Common/InternalLogger.cs
```diff
@@ -284,6 +291,36 @@ private static void Write([CanBeNull]Exception ex, LogLevel level, string messag
...
+            }
+
+            System.Diagnostics.Debug.WriteLine(msg);
+
+            if (level != LogLevel.Trace)
+            {
+                return;
+            }
+
+            System.Diagnostics.Trace.WriteLine(msg);
```
304NotModified: This is a bug? We will write to both Debug and Trace. 

Also Loglevel.Off seems to write to trace. (although I can be checked somewhere else)

tests/NLog.UnitTests/Config/InternalLoggingTests.cs
```diff
@@ -44,7 +44,7 @@ public void InternalLoggingConfigTest1()
             using (var scope = new InternalLoggerScope())
             {
                 CreateConfigurationFromString(@"
-<nlog internalLogFile='c:\file.txt' internalLogLevel='Trace' internalLogToConsole='true' internalLogToConsoleError='true' globalThreshold='Warn' throwExceptions='true'>
+<nlog internalLogFile='c:\file.txt' internalLogLevel='Trace' internalLogToConsole='true' internalLogToConsoleError='true' globalThreshold='Warn' throwExceptions='true' internalWriteToDiagnostics='true'>
```
304NotModified: I'm aware that the current tests are not that good. Still I like to ask to write some more tests as we need more tests in the end (code coverage of 73-74% is not that good)

You can use https://github.com/NLog/NLog/blob/master/tests/NLog.UnitTests/NLogTraceListenerTests.cs#L64-L307 to check how-to read the trace.

src/NLog/Common/InternalLogger.cs
```diff
@@ -284,6 +291,36 @@ private static void Write([CanBeNull]Exception ex, LogLevel level, string messag
...
+
+        private static bool NoLogger()
+        {
+            return string.IsNullOrEmpty(LogFile) && !LogToConsole && !LogToConsoleError && LogWriter == null && !WriteToDiagnostics;
+        }
+
+        private static void LogToDiagnostics(LogLevel level, string msg)
+        {
+#if !SILVERLIGHT && !__IOS__ && !__ANDROID__
+            // log to System.Diagnostics.Debug / System.Diagnostics.Trace
```
304NotModified: This is a good start for the XML docs

304NotModified:
Thanks, added some notes. I think this should be easily tested, so we should do that.

src/NLog/Common/InternalLogger.cs
bryjamus: A little. Code should be readable (like a book) so properly called method tald us  exactly what we check. That's why I extractd method. But this is my proposition so you decide.

src/NLog/Common/InternalLogger.cs
bryjamus: I think your name is better. I will change it.

src/NLog/Common/InternalLogger.cs
bryjamus: > It should be one option (WriteToDiagnostics?). When enabled:

> trace messages to System.Diagnostics.Trace others to System.Diagnostics.Debug

I tought it should work like this. Please describe how should it work so I will change it and add some tests.

src/NLog/Common/InternalLogger.cs
bryjamus: Please tell me what you mean?

src/NLog/Common/InternalLogger.cs
304NotModified: I think it should be:

If LogLevel trace then trace.writeline, else and not if LogLevel off, then debug.writeline.

src/NLog/Common/InternalLogger.cs
304NotModified: Please add the line above the comment box to the xml docs of the method (`<summary>`) 

Every method shoukd have a `<summary>`

src/NLog/Common/InternalLogger.cs
```diff
@@ -284,6 +291,36 @@ private static void Write([CanBeNull]Exception ex, LogLevel level, string messag
...
+            if (!WriteToDiagnostics)
+            {
+                return;
+            }
+
+            System.Diagnostics.Debug.WriteLine(msg);
```
304NotModified: I think it's also nice to send a category to this method? Proposal "NLog"

src/NLog/Common/InternalLogger.cs
```diff
@@ -79,6 +77,7 @@ public static void Reset()
             LogToConsoleError = GetSetting("nlog.internalLogToConsoleError", "NLOG_INTERNAL_LOG_TO_CONSOLE_ERROR", false);
             LogLevel = GetSetting("nlog.internalLogLevel", "NLOG_INTERNAL_LOG_LEVEL", LogLevel.Info);
             LogFile = GetSetting("nlog.internalLogFile", "NLOG_INTERNAL_LOG_FILE", string.Empty);
+            WriteToDiagnostics = GetSetting("nlog.internalWriteToDiagnostics", "NLOG_INTERNAL_WRITE_TO_DIAGNOSTICS", false);
```
304NotModified: I think this should be LogToDiagnostics (also in xml config etc), to be consistent.  Do you agree?

304NotModified:
@bryjamus please rebase (preferred) or merge with the new master where the failing unit has been fixed :)

src/NLog/Common/InternalLogger.cs
bryjamus: You are right - rest properties are called Log... so it will be ok not to change.

304NotModified:
> To prevent NuGet from downloading packages during build, open the Visual Studio Options dialog, click on the Package Manager node and uncheck 'Allow NuGet to download missing packages'.
Common\InternalLogger.cs(87,13): error CS0103: The name 'WriteToDiagnostics' does not exist in the current context

Think we are missing a rename somewhere

src/NLog/Common/InternalLogger.cs
```diff
@@ -290,6 +296,50 @@ private static void Write([CanBeNull]Exception ex, LogLevel level, string messag
...
+            return exception != null && exception.MustBeRethrownImmediately();
+        }
+
+        /// <summary>
+        /// Determine if logging is enabled.
+        /// </summary>
+        /// <returns><c>true</c> if logging is enabled; otherwise, <c>false</c>.</returns>
+        private static bool LoggingEnabled()
+        {
+            return string.IsNullOrEmpty(LogFile) && !LogToConsole && !LogToConsoleError && LogWriter == null &&
```
304NotModified: maybe we should check loglevel also?

src/NLog/Common/InternalLogger.cs
```diff
@@ -290,6 +296,50 @@ private static void Write([CanBeNull]Exception ex, LogLevel level, string messag
...
+            if (!LogToDiagnostics)
+            {
+                return;
+            }
+
+            if (logLevel != LogLevel.Trace)
+            {
+                System.Diagnostics.Debug.WriteLine(message, "NLog");
+            }
+
```
304NotModified: this should be `else`.

We need unit tests for this.

304NotModified:
Not sure if you are done. But looks good. one bug and one comment. 

We definitely need unit tests.

bryjamus:
I'm working on tests.

304NotModified:
:+1:

src/NLog/Common/InternalLogger.cs
bryjamus: You are absolutely right. My mistake.

src/NLog/Common/InternalLogger.cs
bryjamus: I think that this is checked in line 211.

304NotModified: Maybe, but the method name suggests it.

bryjamus:
@304NotModified do you know what can be wrong? I try to synchronize repo and get this error:

```
{ NLog } master » git remote -v                                                                                                                                                                                      /d/GitHubSources/NLog
NLog    https://github.com/NLog/NLog.git (fetch)
NLog    https://github.com/NLog/NLog.git (push)
origin  https://github.com/bryjamus/NLog.git (fetch)
origin  https://github.com/bryjamus/NLog.git (push)
{ NLog } master » git status                                                                                                                                                                                         /d/GitHubSources/NLog
On branch master
Your branch is up-to-date with 'origin/master'.
nothing to commit, working directory clean
{ NLog } master » git pull --rebase NLog origin/master                                                                                                                                                             /d/GitHubSources/NLog
From https://github.com/NLog/NLog
 * remote-tracking branch origin/master -> FETCH_HEAD
First, rewinding head to replay your work on top of it...
error: The following untracked working tree files would be overwritten by checkout:
        tools/DeployNetCF
        tools/DeployNetCF
        tools/DeployNetCF
Please move or remove them before you can switch branches.
Aborting
could not detach HEAD
{ NLog } master »      
```

My second question is - when I run test for the first time 32 test failed. Is it normal? Some examples:

```
Test Name:	ImpersonatingWrapperTest
Test FullName:	NLog.UnitTests.Targets.Wrappers.ImpersonatingTargetWrapperTests.ImpersonatingWrapperTest
Test Source:	D:\GitHubSources\NLog\tests\NLog.UnitTests\Targets\Wrappers\ImpersonatingTargetWrapperTests.cs : line 54
Test Outcome:	Failed
Test Duration:	0:00:00,005

Result StackTrace:	
in Xunit.Assert.Null(Object object)
   in NLog.UnitTests.Targets.Wrappers.ImpersonatingTargetWrapperTests.ImpersonatingWrapperTest() in D:\GitHubSources\NLog\tests\NLog.UnitTests\Targets\Wrappers\ImpersonatingTargetWrapperTests.cs:line 83
Result Message:	
Assert.Null() Failure
Expected: (null)
Actual:   NLog.NLogRuntimeException: Target ImpersonatingWrapper Target[(unnamed)](MyTarget) failed to initialize. ---> System.Runtime.InteropServices.COMException: User name or password are wrong. (Exception HRESULT: 0x8007052E)
             in NLog.Targets.Wrappers.ImpersonatingTargetWrapper.CreateWindowsIdentity(IntPtr& handle) in D:\GitHubSources\NLog\src\NLog\Targets\Wrappers\ImpersonatingTargetWrapper.cs:line 226
             in NLog.Targets.Wrappers.ImpersonatingTargetWrapper.InitializeTarget() in D:\GitHubSources\NLog\src\NLog\Targets\Wrappers\ImpersonatingTargetWrapper.cs:line 129
             in NLog.Targets.Target.Initialize(LoggingConfiguration configuration) in D:\GitHubSources\NLog\src\NLog\Targets\Target.cs:line 296
--------------------------------------------------------------------------------------------------------------
Test Name:	WriteToFileTests
Test FullName:	NLog.UnitTests.Common.InternalLoggerTests.WriteToFileTests
Test Source:	D:\GitHubSources\NLog\tests\NLog.UnitTests\Common\InternalLoggerTests.cs : line 275
Test Outcome:	Failed
Test Duration:	0:00:00,013

Result StackTrace:	
in Xunit.Assert.Equal[T](T expected, T actual, IEqualityComparer`1 comparer)
   in Xunit.Assert.Equal[T](T expected, T actual)
   in NLog.UnitTests.NLogTestBase.AssertFileContents(String fileName, String contents, Encoding encoding) in D:\GitHubSources\NLog\tests\NLog.UnitTests\NLogTestBase.cs:line 168
   in NLog.UnitTests.Common.InternalLoggerTests.WriteToFileTests() in D:\GitHubSources\NLog\tests\NLog.UnitTests\Common\InternalLoggerTests.cs:line 300
Result Message:	
Assert.Equal() Failure
Expected: 64
Actual:   358
--------------------------------------------------------------------------------------------------------------
Test Name:	WriteEventLogEntryEqualToMaxMessageLengthWithOverflowTruncate_TheMessageIsNotTruncated
Test FullName:	NLog.UnitTests.Targets.EventLogTargetTests.WriteEventLogEntryEqualToMaxMessageLengthWithOverflowTruncate_TheMessageIsNotTruncated
Test Source:	D:\GitHubSources\NLog\tests\NLog.UnitTests\Targets\EventLogTargetTests.cs : line 292
Test Outcome:	Failed
Test Duration:	0:00:00,47

Result StackTrace:	
in Xunit.Assert.True(Boolean condition, String userMessage)
   in NLog.UnitTests.Targets.EventLogTargetTests.Write(LogLevel logLevel, EventLogEntryType expectedEventLogEntryType, String logMessage, Layout entryType, EventLogTargetOverflowAction overflowAction, Int32 maxMessageLength) in D:\GitHubSources\NLog\tests\NLog.UnitTests\Targets\EventLogTargetTests.cs:line 403
   in NLog.UnitTests.Targets.EventLogTargetTests.WriteEventLogEntryEqualToMaxMessageLengthWithOverflowTruncate_TheMessageIsNotTruncated() in D:\GitHubSources\NLog\tests\NLog.UnitTests\Targets\EventLogTargetTests.cs:line 296
Result Message:	Failed to find entry of type 'Information' from source 'NLog.UnitTestsaa6129cf11364526bb6132e6c86709ba'
```

304NotModified:
> I try to synchronize repo and get this error

I think you have to remove ALL unstaged files before rebase. (e.g. tools/DeployNetCF is not in https://github.com/NLog/NLog/tree/master/tools).

You can also do a merge instead of a rebase, but the latter is preferred.



> My second question is - when I run test for the first time 32 test failed. Is it normal? Some examples:

Yes, some will not succeed locally, if they pass on AppVeyor, then it's fine :)

Those who are listed are know to be difficult locally, so I don't see a problem.

304NotModified:
Any luck on this?

bryjamus:
Great that you asked me. I want to write today with some questions because I have problem with tests. When I turned on logging to diagnostic like this:
```
  <nlog internalLogToDiagnostics='true' internalLogLevel='Trace'>
    <targets>
      <target name='debug' type='Debug' layout='${logger} ${level} ${message}'/>
    </targets>
    <rules>
      <logger name='*' level='Trace' writeTo='debug'/>
    </rules>
  </nlog>"); 
```
I got overflow exception because I listened for trace and internal logging in this case logged using Trace.WriteLine(). I thought to turn off temporarily logging (LogLevel = LogLevel.Off) but in that case I couldn't get any information from internal logger (neither VS nor using NLogTraceListener). I will be grateful for suggestions. I'm goint to work on this today and tomorrow.

304NotModified:
If you can add this unit test, I can look at it in depth from here :)

bryjamus:
This is my test:
[https://github.com/bryjamus/NLog/blob/master/tests/NLog.UnitTests/LogToDiagnosticsTests.cs](https://github.com/bryjamus/NLog/blob/master/tests/NLog.UnitTests/LogToDiagnosticsTests.cs)

304NotModified:
Your test didn't gave me a overflow? 

Any way, I was more thinking about this:


```c#

        [Fact]
        public void ShouldLogToDiagnosticsWhenInternalLogToDiagnosticsIsOnAndLogLevelIsTrace()
        {
                <nlog internalLogToDiagnostics='true' internalLogLevel='Trace'>
                    <targets>
                        <target name='debug' type='Debug' layout='${logger} ${level} ${message}'/>
                    </targets>
                    <rules>
                        <logger name='*' level='Trace' writeTo='debug'/>
                    </rules>
                </nlog>");

            InternalLogger.IncludeTimestamp = false;
            
            var mockTraceListener = new MockTraceListener();

            Trace.Listeners.Clear();

            Trace.Listeners.Add(mockTraceListener);

            InternalLogger.Trace("Logger1 Hello");

            Assert.Equal(1, mockTraceListener.Messages.Count);
            Assert.Equal("NLog: Trace Logger1 Hello" + Environment.NewLine, mockTraceListener.Messages.First());

        }
     

        private class MockTraceListener : TraceListener
        {

            public readonly List<string> Messages = new List<string>();

            /// <summary>
            /// When overridden in a derived class, writes the specified message to the listener you create in the derived class.
            /// </summary>
            /// <param name="message">A message to write. </param>
            public override void Write(string message)
            {
                Messages.Add(message);

            }

            /// <summary>
            /// When overridden in a derived class, writes a message to the listener you create in the derived class, followed by a line terminator.
            /// </summary>
            /// <param name="message">A message to write. </param>
            public override void WriteLine(string message)
            {
                Messages.Add(message + Environment.NewLine);
            }
        }

```

another test

```c#
  [Fact]
        public void SkipTraceLogsIfLeveIsDebug()
        {
                <nlog internalLogToDiagnostics='true' internalLogLevel='Debug'>
                    <targets>
                        <target name='debug' type='Debug' layout='${logger} ${level} ${message}'/>
                    </targets>
                    <rules>
                        <logger name='*' level='Trace' writeTo='debug'/>
                    </rules>
                </nlog>");

            InternalLogger.IncludeTimestamp = false;

            var mockTraceListener = new MockTraceListener();

            Trace.Listeners.Clear();

            Trace.Listeners.Add(mockTraceListener);

            InternalLogger.Trace("Logger1 Hello");
            InternalLogger.Debug("Logger2 Hello");
            InternalLogger.Debug("Logger3 Hello");
            InternalLogger.Trace("Logger4 Hello");

            Assert.Equal(2, mockTraceListener.Messages.Count);
            Assert.Equal("NLog: Debug Logger2 Hello" + Environment.NewLine, mockTraceListener.Messages[0]);
            Assert.Equal("NLog: Debug Logger3 Hello" + Environment.NewLine, mockTraceListener.Messages[1]);

        }
```

bryjamus:
Thank you for hints. I think about trace listener mock and I had done it but then removed it. I'm goid to look at test today or tomorrow.
StackOverFlowException is thrown when this line:
[https://github.com/bryjamus/NLog/blob/master/src/NLog/Common/InternalLogger.cs#L335](https://github.com/bryjamus/NLog/blob/master/src/NLog/Common/InternalLogger.cs#L335)
is commented:
`
//LogLevel = LogLevel.Off; // without this stackoverflowexception occured
`

bryjamus:
I don't know why some of test failed. In AppVeyor all test passed. Can you please help me to check what went wrong?
I add test which force StackOverflowException (I comment it, test method name is ShouldThrowStackOverFlowExceptionWhenUsingNLogTraceListener):
[https://github.com/bryjamus/NLog/commit/89ca7b1aa73a86ef7761e651ddab7f5f4ead8505#diff-7ac58b5b43eb228929eeb6168cec02fdR215](https://github.com/bryjamus/NLog/commit/89ca7b1aa73a86ef7761e651ddab7f5f4ead8505#diff-7ac58b5b43eb228929eeb6168cec02fdR215)

304NotModified:
Sure, will have a look :)

304NotModified:
I think the solution is here: http://stackoverflow.com/questions/8245314/debug-assert-seems-not-to-work-in-mono

Will test it

tests/NLog.UnitTests/LogToDiagnosticsTests.cs
```diff
@@ -0,0 +1,260 @@
...
+        /// </summary>
+        /// <param name="logLevel">The <see cref="LogLevel"/> for the log event.</param>
+        /// <param name="createTraceListenerFunc">Function which creates <see cref="TraceListener"/> instance.</param>
+        /// <param name="internalLogToDiagnostic">internalLogToDiagnostics XML attribute value. If <c>null</c> attribute is omitted.</param>
+        /// <param name="logToDiagnostic">Value of <see cref="InternalLogger.LogToDiagnostics"/> property. If <c>null</c> property is not set.</param>
+        /// <returns><see cref="TraceListener"/> instance.</returns>
+        private T SetupTestConfiguration<T>(LogLevel logLevel, Func<T> createTraceListenerFunc, bool? internalLogToDiagnostic, bool? logToDiagnostic) where T : TraceListener
+        {
+            var sb = new StringBuilder("<nlog ");
+            sb.AppendFormat("internalLogLevel='{0}'", logLevel);
```
304NotModified: Using a stringbuilder is a good thing when combining strings, but in this case I would recommend 1 / multiple string.formats - it's more readable and the little performance  gain won't matter in unit tests

tests/NLog.UnitTests/LogToDiagnosticsTests.cs
```diff
@@ -0,0 +1,260 @@
...
+
+            Assert.Equal(0, mockTraceListener.Messages.Count);
+        }
+
+        [Theory]
+        [InlineData(true, null)]
+        [InlineData(null, true)]
+        [InlineData(true, true)]
+        public void ShouldLogToDiagnosticsWhenInternalLogToDiagnosticsIsOnAndLogLevelIsTrace(bool? internalLogToDiagnostics, bool? logToDiagnostics)
+        {
```
304NotModified: I understand that we have to copy-paste in the current way the tests because we would otherwise write down all combinations. 

Still, I would like to remove this code duplication. 

I like to tryout the following lib: https://github.com/AArnott/Xunit.Combinatorial

I propose:

1. Create a new branch (because we are gonna try-out, so a new branch is a nice place for it)
2. Install https://github.com/AArnott/Xunit.Combinatorial and try it out
3. Fix that everything builds.
4 If you think it's OK, we don't need a new PR, but like to rebase the new branch into this branch. (or merge). If we got it so far, let's try the rebase.

tests/NLog.UnitTests/LogToDiagnosticsTests.cs
```diff
@@ -0,0 +1,260 @@
...
+            var mockTraceListener = SetupTestConfiguration(LogLevel.Fatal, CreateMockTraceListener, internalLogToDiagnostics, logToDiagnostics);
+
+            InternalLogger.Fatal("Logger1 Hello");
+
+            Assert.Equal(1, mockTraceListener.Messages.Count);
+            Assert.Equal("NLog: Fatal Logger1 Hello" + Environment.NewLine, mockTraceListener.Messages.First());
+        }
+
+        /*[Fact]
+        public void ShouldThrowStackOverFlowExceptionWhenUsingNLogTraceListener()
```
304NotModified: Please add the failing unit test with `[Ignore]`. I guess we won't expect a `StackOverflowException`? (so we need to alter this unit test)

src/NLog/Common/InternalLogger.cs
```diff
@@ -276,6 +270,8 @@ private static void Write([CanBeNull]Exception ex, LogLevel level, string messag
                 {
                     Console.Error.WriteLine(msg);
                 }
+
+                PerformLogToDiagnostics(level, msg);
```
304NotModified: Proposal: rename to "WriteToDiagnostic" - it's more in line with the current naming

src/NLog/Common/InternalLogger.cs
```diff
@@ -290,6 +286,61 @@ private static void Write([CanBeNull]Exception ex, LogLevel level, string messag
...
+        /// <summary>
+        /// Write internal messages to the System.Diagnostics.Debug / System.Diagnostics.Trace.
+        /// </summary>
+        /// <param name="logLevel">The <see cref="LogLevel"/> for the log event.</param>
+        /// <param name="message">A message to write.</param>
+        /// <remarks>Works when property <see cref="LogToDiagnostics"/> set to true.</remarks>
+        private static void PerformLogToDiagnostics(LogLevel logLevel, string message)
+        {
+#if !SILVERLIGHT && !__IOS__ && !__ANDROID__
+            // log to System.Diagnostics.Debug / System.Diagnostics.Trace
```
304NotModified: This comment can we removed. It's already there in the XMLdoc.

304NotModified:
Added some final comments. 

If we check these and the mono builds is succeeding, then we are done :)

tests/NLog.UnitTests/LogToDiagnosticsTests.cs
```diff
@@ -0,0 +1,260 @@
...
+
+        /// <summary>
+        /// Helper method to setup tests configuration
+        /// </summary>
+        /// <param name="logLevel">The <see cref="LogLevel"/> for the log event.</param>
+        /// <param name="createTraceListenerFunc">Function which creates <see cref="TraceListener"/> instance.</param>
+        /// <param name="internalLogToDiagnostic">internalLogToDiagnostics XML attribute value. If <c>null</c> attribute is omitted.</param>
+        /// <param name="logToDiagnostic">Value of <see cref="InternalLogger.LogToDiagnostics"/> property. If <c>null</c> property is not set.</param>
+        /// <returns><see cref="TraceListener"/> instance.</returns>
+        private T SetupTestConfiguration<T>(LogLevel logLevel, Func<T> createTraceListenerFunc, bool? internalLogToDiagnostic, bool? logToDiagnostic) where T : TraceListener
```
304NotModified: Please move below public's

304NotModified: Is the Func really needed? I think we can call `CreateMockTraceListener` in this method?

tests/NLog.UnitTests/LogToDiagnosticsTests.cs
```diff
@@ -0,0 +1,260 @@
+//
```
304NotModified: Please move these tests to `LogToDiagnosticsTests.cs`. Otherwise, it's difficult to find them.

304NotModified:
Update: tried fixing mono but failed. Not sure why I doens't work.

Checked also mono source, but not clear which one we are using now (they seems to move a lot now)

info:

- test: https://github.com/mono/mono/blob/master/mcs/class/System/Test/System.Diagnostics/TraceTest.cs
- source: https://github.com/mono/mono/blob/master/mcs/class/System/System.Diagnostics/TraceImpl.cs
- source has been moved? https://github.com/mono/mono/commit/d6bd25591bf2c50bee29f0c66579d2d81795bf33

304NotModified:
@bryjamus

Any ideas on this?

This week I will create an NLog 4.3 RC. If we can't get it fixed before then, I will move this feature to NLog 4.4

bryjamus:
I introduced almost all of your suggestios but still have problem with tests. I'm going to work on it today so maybe I will find solution.

I think that using this:
[https://github.com/AArnott/Xunit.Combinatorial](https://github.com/AArnott/Xunit.Combinatorial)
is not necessary for this RC so I'm going to deal with this later (I think I will create new PR for this).

304NotModified:
> I think that using this:
> https://github.com/AArnott/Xunit.Combinatorial
> is not necessary for this RC so I'm going to deal with this later (I think I will create new PR for this).

Good idea! It's always nice to have separate PR's for refactor things.

bryjamus:
This is your suggestions changes. Now I can get tests failed on mono so I will try to find why are they crashed.

bryjamus:
I've just found why tests which use Diagnostics.Debug not works. This is because NLog.UnitTests.dll is compiled in **Relese** (on travis). When I run unit test with dll compiled in **Debug** all my tests run properly. It looks like mono ignores compiler directive **#define DEBUG**. However I can't understand why NLogTraceListenerTests works? Do you have any idea why?

What can we do with those tests?

bryjamus:
I think that maybe `WriteToDiagnostics` should work like this (if mono use always `Trace.WriteLine` insted of `Debug.WriteLine`):

```
    private static void WriteToDiagnostics(LogLevel logLevel, string message)
    {
#if !SILVERLIGHT && !__IOS__ && !__ANDROID__
        if (!LogToDiagnostics)
        {
            return;
        }

#if MONO
        System.Diagnostics.Trace.WriteLine(message, "NLog");
#else
        if (logLevel == LogLevel.Trace)
        {
            System.Diagnostics.Trace.WriteLine(message, "NLog");
        }
        else
        {
            System.Diagnostics.Debug.WriteLine(message, "NLog");
        }
#endif
#endif
    }
```

304NotModified:
> However I can't understand why NLogTraceListenerTests works? Do you have any idea why?

Maybe we have to set also TRACE? 
If that won't work, I'm fine with your https://github.com/NLog/NLog/pull/1222#issuecomment-192703963 proposal

304NotModified:
Any luck on this?

bryjamus:
Not yet.

bryjamus:
I fixed myself that this tests should work but this is not true. Tests are run as **Release** (in Travis, Debug in AppVeyor) so NLog library is in Release so mono (and also .Net) **Debug.WriteXXX()** are empty - those methods has **[ConditionalAttribute("Debug")]**. Now in test (also in **Release**) I set **#define DEBUG** directive. But this directive scope is only for this particular *.cs file. This is why this test works in debug and doesn't work in Release. My proposition for mono (write to Trace) is not good. It works for AppVeyor because tests are in **Debug** insted of travis where **Release** is used.

I check this changing configuration to Debug for Travis in this commit: 671e742.

Maybe mono (travis) should also be tested in Debug or my tests should be avoided in Release like this:

```
#if DEBUG
    Assert.Equal(1, mockTraceListener.Messages.Count);
    Assert.Equal("NLog: Error Logger1 Hello" + Environment.NewLine, mockTraceListener.Messages.First());
#else
    Assert.Equal(0, mockTraceListener.Messages.Count);
#endif
```

What do you think?

304NotModified:
Hey @bryjamus ,

As far as I understand, the problem is fixed by writing (always) to `trace`? (no `debug`). I think that is good option (we should add to `<remarks>` docs why we have chosen for `trace` instead of `debug`)

bryjamus:
Done. Please checkout remarks for WriteToDiagnostics.

304NotModified:
nice! Will do!

bryjamus:
Yes, in that case name will reflect intention.

304NotModified:
So `LogToDiagnostics` or `LogToTrace` ?

bryjamus:
I mean `LogToTrace`.

304NotModified:
We have also https://github.com/NLog/NLog/wiki/Trace-target

So If you rename `LogToDiagnostics` or `LogToTrace`, then I will merge it :)

bryjamus:
Rename commited

304NotModified:
It has been merged! Great job!

Issue #1203, opened 2016-01-28
NLog 4.2.2 deadlock


dlkay0:
I have a multi-threaded WCF web service which leverages NLog v4.2.2.  Recently my service deadlocked and I captured the stack traces.  I'm using a standalone NLog.config file which I did not modify at runtime.

```xml
<?xml version="1.0"?>
<configuration>
  <configSections>
    <section name="nlog" type="NLog.Config.ConfigSectionHandler, NLog" />
  </configSections>
  <system.web>
    <compilation targetFramework="4.0" />
  </system.web>
  <nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" autoReload="true">
    <variable name="MessageNoPassword"
         value="${replace:searchFor=(&lt;Password[^&gt;]*)&gt;[^&lt;]+&lt;/Password&gt;:wholeWords=false:replaceWith=$1&gt;XXXX&lt;/Password&gt;:regex=true:ignoreCase=true:inner=${Message}}" />
    <variable name="MessageNoLastName"
                                                                  value="${replace:searchFor=(&lt;LastName[^&gt;]*)&gt;[^&lt;]+&lt;/LastName&gt;:wholeWords=false:replaceWith=$1&gt;XXXX&lt;/LastName&gt;:regex=true:ignoreCase=true:inner=${MessageNoPassword}}" />
                <!-- ... -->
    <variable name="MyLayout" value="${longdate} | ${level} | ${mdc:Operator} | ${mdc:SessionId} | ${MessageNoLastName}${onexception: | EXCEPTION FOLLOWS\:${exception:format=tostring}}" />

    <targets async="true">
      <target xsi:type="File"
                                layout="${MyLayout}"
                                name="logFile"
                                fileName="${basedir}/Logs/AppName_${mdc:ClientID}_${shortdate}.log"
                                archiveFileName="${basedir}/Logs/Archive/AppName_${mdc:ClientID}_${shortdate}.{#}.log"
                                archiveEvery="Hour"
                                archiveNumbering="Rolling"
                                maxArchiveFiles="30" />
                  <!-- ... -->
    </targets>

    <rules>
      <logger name="*" minlevel="Info" writeTo="logFile" />      
      <!-- ... -->
    </rules>
  </nlog>
</configuration>
```

[...]

PR #1223, opened 2016-02-06
Fix deadlock on Factory

304NotModified:
- No locking on factory when rendering layout renderers
- Less locking when using GlobalDiagnosticsContext.Get / MappedDiagnosticsContext.Get / MappedDiagnosticsLogicalContext.Get
- added <remarks> for (indirect) locks 
- move ConvertToString from GlobalContext to FormatHelper (internal)
- introduced StringBuilderExt, remove duplicate code (internal)

fixes https://github.com/NLog/NLog/issues/1203

### DIFF
--- a/src/NLog/LayoutRenderers/MdlcLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/MdlcLayoutRenderer.cs
@@ -34,9 +34,9 @@
 namespace NLog.LayoutRenderers
 {
 #if NET4_0 || NET4_5
-    using System;
     using System.Text;
     using Config;
+    using Internal;
 
     /// <summary>
     /// Mapped Diagnostic Logical Context item (based on CallContext).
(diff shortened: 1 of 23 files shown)


src/NLog/LayoutRenderers/MdlcLayoutRenderer.cs
```diff
@@ -58,8 +59,14 @@ public class MdlcLayoutRenderer : LayoutRenderer
...
-            var message = MappedDiagnosticsLogicalContext.Get(Item);
-            builder.Append(message);
+            //don't use MappedDiagnosticsLogicalContext.Get to ensure we are not locking the Factory (indirect by LogManager.Configuration).
+            var o = MappedDiagnosticsLogicalContext.GetObject(this.Item);
+            var formatProvider = logEvent.FormatProvider;
+            if (formatProvider == null && LoggingConfiguration != null)
+            {
+                formatProvider = LoggingConfiguration.DefaultCultureInfo;
+            }
+            builder.Append(Convert.ToString(o, formatProvider));
```
Page-Not-Found: Wouldn't it benefit if we add this logic to the base class and have it is a property?

304NotModified: That's a good idea :)

304NotModified:
@Page-Not-Found I removed the code duplication.

Page-Not-Found:
I did not find any other issues with this change.

304NotModified:
Thanks @Page-Not-Found !

PR #1225, opened 2016-02-07
XmlLoggingConfiguration: Set config values on correct LogFactory object

bhaeussermann:
Related to #1203


src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -472,12 +476,18 @@ private void ParseNLogElement(NLogXmlElement nlogElement, string filePath, bool
             if (filePath != null)
                 this.fileMustAutoReloadLookup[GetFileLookupKey(filePath)] = autoReload;
 
-            LogManager.ThrowExceptions = nlogElement.GetOptionalBooleanAttribute("throwExceptions", LogManager.ThrowExceptions);
+            if (customLogFactory == null)
```
304NotModified: maybe we can pass LogManager as custom factory?

src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -472,12 +476,18 @@ private void ParseNLogElement(NLogXmlElement nlogElement, string filePath, bool
...
+                LogManager.ThrowExceptions = nlogElement.GetOptionalBooleanAttribute("throwExceptions", LogManager.ThrowExceptions);
+            else
+                customLogFactory.ThrowExceptions = nlogElement.GetOptionalBooleanAttribute("throwExceptions", customLogFactory.ThrowExceptions);
             InternalLogger.LogToConsole = nlogElement.GetOptionalBooleanAttribute("internalLogToConsole", InternalLogger.LogToConsole);
             InternalLogger.LogToConsoleError = nlogElement.GetOptionalBooleanAttribute("internalLogToConsoleError", InternalLogger.LogToConsoleError);
             InternalLogger.LogFile = nlogElement.GetOptionalAttribute("internalLogFile", InternalLogger.LogFile);
             InternalLogger.LogLevel = LogLevel.FromString(nlogElement.GetOptionalAttribute("internalLogLevel", InternalLogger.LogLevel.Name));
-            LogManager.GlobalThreshold = LogLevel.FromString(nlogElement.GetOptionalAttribute("globalThreshold", LogManager.GlobalThreshold.Name));
+            if (customLogFactory == null)
```
304NotModified: Idem

src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -82,21 +84,23 @@ private ConfigurationItemFactory ConfigurationItemFactory
...
-        public XmlLoggingConfiguration(string fileName)
-        {
-            using (XmlReader reader = CreateFileReader(fileName))
-            {
-                this.Initialize(reader, fileName, false);
-            }
-        }
+        /// <param name="customLogFactory">The custom <see cref="LogFactory"/> to which to apply any applicable configuration values,
+        /// or <c>null</c> to use the global <see cref="LogFactory"/> (via <see cref="LogManager"/>).</param>
+        public XmlLoggingConfiguration(string fileName, LogFactory customLogFactory = null)
```
304NotModified: Looks good but optional parameters aren't binary compatible. So we need extra ctors and should not alter the current ones.

src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -82,21 +84,23 @@ private ConfigurationItemFactory ConfigurationItemFactory
...
 
         /// <summary>
         /// Initializes a new instance of the <see cref="XmlLoggingConfiguration" /> class.
         /// </summary>
         /// <param name="fileName">Configuration file to be read.</param>
         /// <param name="ignoreErrors">Ignore any errors during configuration.</param>
-        public XmlLoggingConfiguration(string fileName, bool ignoreErrors)
+        /// <param name="customLogFactory">The custom <see cref="LogFactory"/> to which to apply any applicable configuration values,
+        /// or <c>null</c> to use the global <see cref="LogFactory"/> (via <see cref="LogManager"/>).</param>
+        public XmlLoggingConfiguration(string fileName, bool ignoreErrors, LogFactory customLogFactory = null)
```
304NotModified: idem binary compatiblity

tests/NLog.UnitTests/LogFactoryTests.cs
```diff
@@ -112,6 +109,38 @@ public void InvalidXMLConfiguration_ThrowErrorWhen_ThrowExceptionFlagIsSet()
         }
 
         [Fact]
+        public void SecondaryLogFactoryDoesNotTakePrimaryLogFactoryLock()
```
304NotModified: What are where asserting here? no deadlock?

304NotModified:
Looks good. Added some notes. Binary compat  is the most important one.

tests/NLog.UnitTests/LogFactoryTests.cs
bhaeussermann: Yes. We run the thread while holding the lock of the global `LogFactory`. So if the code in the thread takes the same lock, we will deadlock and the `Join()` method will time out.

bhaeussermann: Hence the test simulates what would in some cases cause a deadlock.

304NotModified: Ow yes indeed. Please add that as comments to the method / lines. :)

src/NLog/Config/XmlLoggingConfiguration.cs
bhaeussermann: Good idea!

bhaeussermann:
AppVeyor:
```
LogManager.cs(137,24): error CS1574: Warning as Error: XML comment on 'NLog.LogManager.Configuration' has cref attribute 'LogFactory.Configuration' that could not be resolved [C:\projects\nlog\src\NLog\NLog.Xamarin.Android.csproj]
```

Wha?! I didn't touch this `cref` attribute and the `LogFactory.Configuration` property was not moved or renamed. Any ideas?

304NotModified:
> Wha?! I didn't touch this cref attribute and the LogFactory.Configuration property was not moved or renamed. Any ideas?

A `using` has been removed?

bhaeussermann:
@304NotModified I cannot see what is wrong. If I right-click on any of these `cref`s in Visual Studio and select "Go To Definition", it takes me to the corresponding member. So they are still correct according to Visual Studio.

304NotModified:
well I would recommend to merge with master locally and then check it.

bhaeussermann:
> well I would recommend to merge with master locally and then check it.

The `cref`s still seem fine according to Visual Studio. I cannot see what changes in this PR might cause the `cref`s to break.

304NotModified:
Will check it locally tomorrow. Strange issue. Have to be a missing `using` or something like that.

304NotModified:
https://github.com/bhaeussermann/NLog/pull/1

some ambiguity about Conguration (class name vs property)

304NotModified:
Thanks!

PR #1227, opened 2016-02-08
LogFactory: Add generic-type versions of GetLogger() and GetCurrentClassLogger()

bhaeussermann:
Related to #1203

### DIFF
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -351,6 +351,24 @@ namespace NLog
         }
 
         /// <summary>
+        /// Gets the logger with the name of the current class. 
+        /// </summary>
+        /// <returns>The logger.</returns>
+        /// <remarks>This is a slow-running method. 
+        /// Make sure you're not doing this in a loop.</remarks>
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        public T GetCurrentClassLogger<T>() where T : Logger
+        {
+#if SILVERLIGHT
+            var frame = new StackFrame(1);
+#else
+            var frame = new StackFrame(1, false);
+#endif
+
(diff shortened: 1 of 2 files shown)


304NotModified:
I restarted appveyor build 3 times and same unit test fails. But dunno why as it looks unrelated.

304NotModified:
Please merge (rebase would be even better)  with master.

src/NLog/LogFactory.cs
```diff
@@ -351,6 +351,24 @@ public Logger GetCurrentClassLogger()
         }
 
         /// <summary>
+        /// Gets the logger with the name of the current class. 
+        /// </summary>
+        /// <returns>The logger.</returns>
+        /// <remarks>This is a slow-running method. 
+        /// Make sure you're not doing this in a loop.</remarks>
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        public T GetCurrentClassLogger<T>() where T : Logger
```
304NotModified: should it not be `T: ILogger`?

bhaeussermann: Good call!

I just tried it, but it yields the compiler error `Constraint type 'ILogger' is not CLS-complaint`

Apparently only CLS-compliant types can be used for generic types, and `ILogger` is marked with `[CLSCompliant(false)]`.

bhaeussermann: Besides, all of the other `Get...Logger()` methods return type `Logger`.

304NotModified: CLS compliant is not needed for generic types, but it is for this class (as it's not marked als CLScompliant false). 

Thanks for checking! 

PS. The return type is here `T` ;)

304NotModified:
Thanks again!

PR #1236, opened 2016-02-12
Bugfix: Internallogger creates folder, even when turned off.

### LINKED ISSUE #1229: Directory created even if internalLogLevel set to off?
I noticed that the directory _c:\temp\_ of value `internalLogFile="c:\temp\nlog-internal.log"` is created even when `internalLogLevel` is set to `"Off"`:
```
internalLogLevel="Off"
internalLogFile="C:\temp\nlog-internal.log"
```

Is this on purpose or should the directory _c:\temp\_ only be created when the `internalLogLevel` is not set to `"Off"`? It's a minor issue but now I have an empty _temp_ directory and I didn't know what caused it at first..

Tested with NLog version: 4.2.3.

eduardorascon:
Fixes #1229 

When LogLevel is set to "Off" NLog should not create an emtpy folder or try to write a log file to a folder that doesn´t exists.

### DIFF
--- a/tests/NLog.UnitTests/Common/InternalLoggerTests.cs
+++ b/tests/NLog.UnitTests/Common/InternalLoggerTests.cs
@@ -416,7 +416,7 @@ namespace NLog.UnitTests.Common
             // Set the log file, which will only create the needed directories
             InternalLogger.LogFile = tempFile;
 
-            Assert.True(Directory.Exists(randomSubDirectory) == shouldCreateDirectory);
+            Assert.Equal(Directory.Exists(randomSubDirectory), shouldCreateDirectory);
 
             try
             {
@@ -424,7 +424,7 @@ namespace NLog.UnitTests.Common
 
                 InternalLogger.Log(LogLevel.FromString(rawLogLevel), "File and Directory created.");
 
-                Assert.True(File.Exists(tempFile) == shouldCreateDirectory);
+                Assert.Equal(File.Exists(tempFile), shouldCreateDirectory);
             }
             finally
             {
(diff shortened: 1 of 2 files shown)


tests/NLog.UnitTests/Common/InternalLoggerTests.cs
```diff
@@ -418,22 +416,15 @@ public void CreateDirectoriesIfNeededTests()
             // Set the log file, which will only create the needed directories
             InternalLogger.LogFile = tempFile;
 
-            Assert.True(Directory.Exists(randomSubDirectory));
+            Assert.True(Directory.Exists(randomSubDirectory) == shouldCreateDirectory);
```
304NotModified: Please use assert.equal here

tests/NLog.UnitTests/Common/InternalLoggerTests.cs
```diff
@@ -38,6 +38,7 @@
 using Xunit;
 using NLog.Common;
 using System.Text;
+using Xunit.Extensions;
```
304NotModified: I think we need `#if !SILVERLIGHT` here and at the test.

eduardorascon:
Already made those changes. Should I create another pull request?

304NotModified:
No that's not needed

tests/NLog.UnitTests/Common/InternalLoggerTests.cs
eduardorascon: CreateDirectoriesIfNeededTests is already inside an `#if !SILVERLIGHT` at [line 174](https://github.com/eduardorascon/NLog/blob/fix-create-directories/tests/NLog.UnitTests/Common/InternalLoggerTests.cs#L174).

eduardorascon:
@304NotModified Can you give me some tips on how to fix this appveyor problem?

304NotModified:
I think it's fine now. One multi threaded test is lately unstable. See #1237

304NotModified:
Thanks!

eduardorascon:
Thanks you, you were very helpful.

PR #1238, opened 2016-02-12
Control throwing of NLogConfigurationExceptions (LogManager.ThrowConfigExceptions)

304NotModified:
Added `ThrowConfigExceptions` to control throwing of `NLogConfigurationException`. Before `NLogConfigurationException` where always thrown, which could lead to `System.TypeInitializationException` which is bad and confusing as sending info with the exception is difficult - see [stackoverflow](http://stackoverflow.com/questions/34994470/better-typeinitializationexception-innerexception-is-also-null).

```c#
LogManager.ThrowConfigExceptions  = true | false | null
```
```xml
<nlog throwConfigExceptions="true"|"false"|"">
```

if `ThrowConfigExceptions` is `null` (or empty `string` from XML), then `ThrowExceptions` is used. So `ThrowConfigExceptions` inherits from `ThrowExceptions`.

This isn't fully backwards-compatible, but it is also a bug as we in code and docs is stated "By default exceptions are not thrown under any circumstances.". also this is a good change to include with #1143

With the addition of https://github.com/NLog/NLog/pull/1222, there is no good reason to throw exceptions even with config errors.

### DIFF
--- a/src/NLog/Config/NLogXmlElement.cs
+++ b/src/NLog/Config/NLogXmlElement.cs
@@ -33,6 +33,7 @@
 
 namespace NLog.Config
 {
+    using Internal;
     using System;
     using System.Collections.Generic;
     using System.Globalization;
@@ -166,7 +167,7 @@ namespace NLog.Config
                 return defaultValue;
             }
 
-            if (string.IsNullOrWhiteSpace(value))
+            if (StringHelpers.IsNullOrWhiteSpace(value))
             {
                 return defaultValue;
             }
--- a/src/NLog/Config/NLogXmlElement.cs
+++ b/src/NLog/Config/NLogXmlElement.cs
@@ -153,7 +153,7 @@ namespace NLog.Config
         }
 
         /// <summary>
-        /// Gets the optional boolean attribute value. If whitespace, then returning <paramref name="defaultValue"/>
+        /// Gets the optional boolean attribute value. If whitespace, then returning <c>null</c>.
         /// </summary>
         /// <param name="attributeName">Name of the attribute.</param>
         /// <param name="defaultValue">Default value to return if the attribute is not found.</param>
@@ -164,12 +164,14 @@ namespace NLog.Config
 
             if (!this.AttributeValues.TryGetValue(attributeName, out value))
             {
+                //not defined, don't override default
                 return defaultValue;
             }
 
             if (StringHelpers.IsNullOrWhiteSpace(value))
             {
-                return defaultValue;
+                //not default otherwise no difference between not defined and empty value.
+                return null;
             }
 
             return Convert.ToBoolean(value, CultureInfo.InvariantCulture);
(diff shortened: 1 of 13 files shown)


304NotModified:
Todo. Make throwConfigExceptions nullable (with default of throwExceptions)

304NotModified:
waiting for https://github.com/NLog/NLog/pull/1240

304NotModified:
`UnitTests.Common.InternalLoggerTests.CreateDirectoriesIfNeededTests`

Assert.False() Failure

304NotModified:
@bhaeussermann I would like to have your opinion on this one :)

bhaeussermann:
@304NotModified This seems to be the best solution. I'm also happy with the code. 

However, what is going to happen if the configuration used is `ThrowExceptions = true` and `ThrowConfigExceptions = false`?  
How about we throw an exception whenever `Log()` is called saying something like `Configuration failed. See internal log for more details.`?

304NotModified:
> @304NotModified This seems to be the best solution. I'm also happy with the code.

Thanks for the check!

> However, what is going to happen if the configuration used is ThrowExceptions = true and ThrowConfigExceptions = false?

It will throw runtime exceptions and not config exceptions.

> How about we throw an exception whenever Log() is called saying something like Configuration failed. See internal log for more details.?

Nice idea, but we like to throw less exceptions (by default!) because:

- We state that we never throw exceptions
- Throwing exceptions can lead to TypeInitializationException, which are hard to understand (sometimes even no innerException), hard to debug and can't be prevented (without big loss of performance), see [stackoverflow](http://stackoverflow.com/questions/34994470/better-typeinitializationexception-innerexception-is-also-null)

bhaeussermann:
> > However, what is going to happen if the configuration used is ThrowExceptions = true and ThrowConfigExceptions = false?

> It will throw runtime exceptions and not config exceptions.

My suggestion applied to this situation. What runtime exception would this throw? Would it be clear that something went wrong with configuration? I cannot check that right now.

If it is not clear from that "runtime" exception, then perhaps we should add a check to throw a more meaningful exception for this situation telling the developer to check the internal log.

bhaeussermann:
(My comment above assumes the situation where there is an error in the configuration.)

304NotModified:
I'm not sure if we are on the same line.

So we have now with the PR two properties:

- ThrowExceptions: throw runtime exceptions like, invalid query, unreachable database, wrong filename (in some cases)
- ThrowConfigExceptions: throws exceptions like unknown targetname, unknown target property , unknown layout renderer.

All the exceptions get logged to the internal logger. 

ThrowExceptions was already in NLog (with default `false`). `ThrowConfigExceptions` is new (and you can see it as before `true` and since this PR `false`)

So it's indeed less clear when there is a config mistake now, but this PR will also lead to less errors at runtime (config mistake can be also at runtime when changing the config) and no unclear `TypeInitializationException` by default :)

If you like the previous behavior, just change `ThrowConfigExceptions=true`

304NotModified:
docs:

- https://github.com/NLog/NLog/wiki/Configuration-file

PR #1245, opened 2016-02-16
Add Logger.Swallow(Task task)

### LINKED ISSUE #1231: Add Swallow(Task task)
Feature suggestion: A `Forget` method that is similar to `Logger.SwallowAsync` except that it would return `void` instead of `Task`. Its use would be in fire-and-forget situations where the calling code intentionally does not await an async method. It would prevent the code from having to use some other means of suppressing this warning:

> Warning CS4014: Because this call is not awaited, execution of the current method continues before the call is completed. Consider applying the 'await' operator to the result of the call.

I'd be happy to submit a PR if there is consensus that this is a good idea.

breyed:
fixes #1231.

### DIFF
--- a/src/NLog/ISuppress.cs
+++ b/src/NLog/ISuppress.cs
@@ -44,7 +44,7 @@ namespace NLog
     public interface ISuppress
     {
         /// <summary>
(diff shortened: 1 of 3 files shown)


304NotModified:
Thanks!

PR #1248, opened 2016-02-17
CoreCLR: regular DNU commands instead of Cake file

304NotModified:


304NotModified:
> @304NotModified I'm not sure removing cake will made the process less complicated. Cake is doing a lot of work and allow us to have the same process on windows and linux. 
For example, if you want to work with the dnx/dnu command directly you should call all dnu /dnx command using dnvm exec to run in the right runtime context.


@phenixdotnet Moved the discussion to the related PR.

I have now the following problems with the Cake file

- It's a long script while the DNU commands need at most 10 lines. It tooks me some time to understand it, while the DNU commands are really simple and clear. 
- Configuration is duplicated: in the project.json and cake file
- The DNX platform version is also duplication, in the cake file and the build.cmd
- It's difficult to add / remove a platform (e.g. UWP)
- It's slow. The build time with DNU commands (restore all platforms at once, build all platforms at once), is 2 times faster. (4 min vs 8 min)
- It's more maintenance. The same now for the msbuild project. A lot of users cannot edit those files.

But I have to agree I have also problems with running the unit tests under different platforms. If that can be fixed in the plain CMD file, then that has the preference.

phenixdotnet:
> Configuration is duplicated: in the project.json and cake file. I had multiple times that only one was working (Visual Studio VS build script)

You have the same problem with dnu / dnx command as you must specify the framework and runtime information.

> The DNX platform version is also duplicated; in the cake file and the build.cmd
Its fixed :)

> It's difficult to add / remove a platform (e.g. remove UWP, add net45)
I kindly desagree. Copy/Paste target, change dependency and change framework in cake file and its done.

> It's slow. The build time with DNU commands (restore all platforms at once, build all platforms at once), is 2 times faster. (4 min vs 8 min)
Wooh 8 minutes ? Currently what I can observe is the time took by the unit tests.
The delta between the dnx/dnu command is, may be, because you always call the dnx 4.5.1 version.

On the 2 others points, I agree with you, but please keep in mind that we remove cake you must manage the cmd build file AND the bash file to have travis happy. 

If you want to test the dnu/dnx command in the same way I implemented it in Cake you can use this command:

To restore:
`powershell.exe -NoProfile -NoLogo -ExecutionPolicy unrestricted -File "PathToDnvm.ps1" "exec 1.0.0-rc1-update1 -a x64 -r coreclr dnu.cmd restore"`
 
To build
`powershell.exe -NoProfile -NoLogo -ExecutionPolicy unrestricted -File "PathToDnvm.ps1" "exec 1.0.0-rc1-update1 -a x64 -r coreclr dnu.cmd build -p .\src\NLog\project.json"`

To run test
`powershell.exe -NoProfile -NoLogo -ExecutionPolicy unrestricted -File "PathToDnvm.ps1" "run 1.0.0-rc1-update1 -a x64 -r coreclr --project .\tests\Nlog.UnitTests\project.json test"`

Note: I'm writing those command without any tests, but its the general idea.

304NotModified:
> Configuration is duplicated: in the project.json and cake file. I had multiple times that only one was working (Visual Studio VS build script)
> You have the same problem with dnu / dnx command as you must specify the framework and runtime information.

Isn't `dnu build src/NLog --quiet` works for all  the platforms? 



> If you want to test the dnu/dnx command in the same way I implemented it in Cake you can use this command:

Well I used this one 

```
#!/bin/bash

DNXVERSION="1.0.0-rc1-update1"

if [ ! -f ~/.dnx/dnvm/dnvm.sh ] 
then
echo "Downloading dnvm"
curl -sSL https://raw.githubusercontent.com/aspnet/Home/dev/dnvminstall.sh | DNX_BRANCH=dev sh && source ~/.dnx/dnvm/dnvm.sh
else
echo "Using installed dnvm.sh script"
source ~/.dnx/dnvm/dnvm.sh
fi

dnvm install $DNXVERSION -a x64 -r coreclr
dnvm install $DNXVERSION -a x64 -r mono -alias default

dnvm use default
```

phenixdotnet:
> Isn't dnu build src/NLog --quiet works for all the platforms? 
Yes if your current dnx/dnu env is clr because clr can build coreclr package.

But any way this will not work to run commands.
The script you're using is right to bootstrap and ensure dnx/dnu are installed in the version required but doesn't let you to choose on which framework the commands will be executed.

And just calling the `dnvm use 1.0.0-rc1-update1 -r coreclr -a x64`after build doesn't help because the env and specially the PATH are already set to clr version.
Before taking a lot of time to implement (and fix ^^) dnu/dnx commands in Cake I do a lot of test with cmd only and cmd + powershell alternative but I didn't get it working as expected.

May be can refactor the cake script, move some part of the script to a full, real .NET Lib and keep only variable into, also we can (re)test to run build only one time and keep the tests executed for each platform. All of this possible with cake actually.

304NotModified:
OK thanks for the info

304NotModified:
DNU = old. Already replaced

PR #1261, opened 2016-02-24
[WIP] Easier api

304NotModified:
- See API improvements section in: http://304notmodified.github.io/2016/02/08/nlog-4-3-0-is-now-available.html

Changes:

- default name of target from attribute 
- addRule methods
- allTargets will now really return all targets
- isLevelEnabled(off) was throwing an exception.

Refactor changes:

- moved from Array to List internally for scanProperties.
- added extra ctor for `NLogConfigurationException`
- removed duplicate code

Todo

- [ ] every target has an new ctor with `name` (beside default ctor)

To be discussed:

- AddRule methods (see also #1266)
- [default name of target from attribute ](https://github.com/NLog/NLog/commit/47eea556c7430b75f03119392c0df6ea7850ce7d)

PR #1266, opened 2016-02-25
[WIP] Easier api - reorder parameters

304NotModified:
#1261 with 1 extra change:

- changed order of parameters of addRule.

Before:

```
public void AddRule(string loggerNamePattern, LogLevel minLevel, LogLevel maxLevel, Target target)
```

after

```
 public void AddRule(LogLevel minLevel, LogLevel maxLevel, Target target, string loggerNamePattern = "*")
```

Benefits: 

- optional parameter, 
- loggerNamePattern is more advanced

Downsides:

- different order compared with ctor loggingRule
- different order compared with nlog.config (examples)
- optional parameters won't work for logLevels (as it's a class)

### DIFF
--- a/src/NLog/Config/LoggingConfiguration.cs
+++ b/src/NLog/Config/LoggingConfiguration.cs
@@ -206,6 +206,95 @@ namespace NLog.Config
             return FindTargetByName(name) as TTarget;
         }
 
+
+        /// <summary>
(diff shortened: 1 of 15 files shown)


304NotModified:
Superseded by https://github.com/NLog/NLog/pull/1286

PR #1273, opened 2016-02-29
Added the ability to allow virtual paths for SMTP pickup directory

michaeljbaird:


304NotModified:
Cool! Thanks!

michaeljbaird:
Great feedback and I did want to make a helper method for that code but did not want to introduce anything like `ConfigureMailClient ` w/o approval. :+1:  I get an update to the PR sometime tomorrow.

src/NLog/Targets/MailTarget.cs
```diff
@@ -449,6 +451,26 @@ internal void ConfigureMailClient(LogEventInfo lastEvent, ISmtpClient client)
             
         }
 
+        /// <summary>
+        /// Handle <paramref name="pickupDirectoryLocation"/> if it is a virtual directory. Set PickupDirectoryLocation of <paramref name="client"/>
+        /// </summary>
+        /// <param name="client">client to set properties on</param>
+        /// <param name="pickupDirectoryLocation"></param>
+        /// <returns></returns>
+        internal static string GetPickupDirectoryLocation(ISmtpClient client, string pickupDirectoryLocation)
```
304NotModified: Proposal: 
- remove 'ISmtpClient client' (I guess this is a bug)
- rename to "ConvertDirectoryLocation"
- update docs

src/NLog/Targets/MailTarget.cs
```diff
@@ -449,6 +451,26 @@ internal void ConfigureMailClient(LogEventInfo lastEvent, ISmtpClient client)
...
+        {
+            if (!pickupDirectoryLocation.StartsWith("~/"))
+            {
+                return pickupDirectoryLocation;
+            }
+
+            // Support for Virtual Paths
+            var root = AppDomain.CurrentDomain.BaseDirectory;
+            var directory = client.PickupDirectoryLocation.Replace("~/", "").Replace('/', Path.DirectorySeparatorChar);
```
304NotModified: This could be improved:

- "~/" should be a `const string` (e.g. virtualPathPrefix)
- We check before `pickupDirectoryLocation` and not `client.PickupDirectoryLocation`
- We check a StartsWith but we're doing a replace all. We should replace only the first ~/ (or just `.Substring (virtualPathPrefix.length)` )

304NotModified:
Thanks for the changes!

There are some improvements needed before I can merge it. See the code comments.

PS: I don't get an alert when commits are pushed to the PR.

michaeljbaird:
PR Updated

304NotModified:
Thanks!

PR #1298, opened 2016-03-09
NullReferenceException fix for VariableLayoutRenderer

neris:
It is possible that LogManager.Configuration may be null so calling loggingConfiguration.Variables may throw a NullReferenceException.


304NotModified:
Thanks!

Issue #1262, opened 2016-02-24
Fix threadsafe issue of ScanProperties


304NotModified:
**NLog version**:  current (4.3 beta)

unit test: NLog.UnitTests.LogManagerTests.ThreadSafe_getCurrentClassLogger_test
```
Microsoft.Practices.Unity.ResolutionFailedException : Resolution of the dependency failed, type = "NLog.UnitTests.LogManagerTests+ClassA", name = "(none)".

Exception occurred while: Calling constructor NLog.UnitTests.LogManagerTests+ClassC(NLog.UnitTests.LogManagerTests+ClassD dd).

Exception is: TypeInitializationException - The type initializer for 'ClassC' threw an exception.

-----------------------------------------------

At the time of the exception, the container was:



  Resolving NLog.UnitTests.LogManagerTests+ClassA,(none)

  Resolving parameter "dd" of constructor NLog.UnitTests.LogManagerTests+ClassA(NLog.UnitTests.LogManagerTests+ClassB dd)

    Resolving NLog.UnitTests.LogManagerTests+ClassB,(none)

    Resolving parameter "dd" of constructor NLog.UnitTests.LogManagerTests+ClassB(NLog.UnitTests.LogManagerTests+ClassC dd)

      Resolving NLog.UnitTests.LogManagerTests+ClassC,(none)

      Calling constructor NLog.UnitTests.LogManagerTests+ClassC(NLog.UnitTests.LogManagerTests+ClassD dd)



---- System.TypeInitializationException : The type initializer for 'ClassC' threw an exception.

-------- System.InvalidOperationException : Collection was modified after the enumerator was instantiated.
```

See https://ci.appveyor.com/project/nlog/nlog/build/4.0.1976/tests

---

   at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)

   at System.Collections.Generic.Queue`1.Enumerator.MoveNext()

   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)

   at NLog.Internal.ObjectGraphScanner.ScanProperties[T](List`1 result, Object o, Int32 level, HashSet`1 visitedObjects) in C:\projects\nlog\src\NLog\Internal\ObjectGraphScanner.cs:line 129

[...]

PR #1310, opened 2016-03-17
Fix threadsafe issue of GetLogger / GetCurrentClassLogger (+improve performance) 

304NotModified:
another try for https://github.com/NLog/NLog/issues/1262

Fix PrecalculateStackTraceUsage for threadsafe  issues in objectscaner.FindReachableObjects

### DIFF
--- a/src/NLog/Internal/TargetWithFilterChain.cs
+++ b/src/NLog/Internal/TargetWithFilterChain.cs
@@ -31,6 +31,8 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
+using System.Linq;
+
 namespace NLog.Internal
 {
     using System.Collections.Generic;
@@ -43,8 +45,8 @@ namespace NLog.Internal
     /// whether logging should happen.
     /// </summary>
     [NLogConfigurationItem]
-	internal class TargetWithFilterChain
-	{
+    internal class TargetWithFilterChain
+    {
         private StackTraceUsage stackTraceUsage = StackTraceUsage.None;
 
         /// <summary>
@@ -92,19 +94,10 @@ namespace NLog.Internal
 
             // find all objects which may need stack trace
             // and determine maximum
-            foreach (var item in ObjectGraphScanner.FindReachableObjects<IUsesStackTrace>(this))
+            // only the target can have IUsesStackTrace
+            if (Target != null)
             {
-                var stu = item.StackTraceUsage;
-
-                if (stu > this.stackTraceUsage)
-                {
-                    this.stackTraceUsage = stu;
-
-                    if (this.stackTraceUsage >= StackTraceUsage.Max)
-                    {
-                        break;
-                    }
-                }
+                this.stackTraceUsage = Target.GetAllLayouts().OfType<IUsesStackTrace>().DefaultIfEmpty().Max(usage => usage == null ? StackTraceUsage.None : usage.StackTraceUsage);
             }
         }
     }
--- a/src/NLog/Layouts/SimpleLayout.cs
+++ b/src/NLog/Layouts/SimpleLayout.cs
@@ -31,6 +31,9 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
+using System.Collections.Generic;
+using System.Linq;
+
 namespace NLog.Layouts
 {
     using System;
@@ -51,7 +54,7 @@ namespace NLog.Layouts
     [Layout("SimpleLayout")]
     [ThreadAgnostic]
     [AppDomainFixedOutput]
-    public class SimpleLayout : Layout
+    public class SimpleLayout : Layout, IUsesStackTrace
     {
         private const int MaxInitialRenderBufferLength = 16384;
         private int maxRenderedLength;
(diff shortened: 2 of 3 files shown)


src/NLog/Internal/TargetWithFilterChain.cs
```diff
@@ -92,17 +94,21 @@ internal void PrecalculateStackTraceUsage()
 
             // find all objects which may need stack trace
             // and determine maximum
-            foreach (var item in ObjectGraphScanner.FindReachableObjects<IUsesStackTrace>(this))
+            // only the target can have IUsesStackTrace
+            if (Target != null)
```
Taloth: Don't you need to recurse NextInChain as well? In case those LayoutRenderers require stacktraces?

src/NLog/Internal/TargetWithFilterChain.cs
```diff
@@ -92,17 +94,21 @@ internal void PrecalculateStackTraceUsage()
...
             {
-                var stu = item.StackTraceUsage;
-
-                if (stu > this.stackTraceUsage)
+                foreach (var item in Target.GetAllLayouts().OfType<IUsesStackTrace>())
```
Taloth: Afaik only LayoutRenderers implement IUsesStackTrace, not Layout, so OfType will yield nothing.

304NotModified: Thanks. There is indeed a mistake here

Taloth:
I noticed this PR after investigating an initialization performance issue when using ~150 loggers with 6 logging rules (on NLog 4.2.3). PrecalculateStackTraceUsage would end up calling the ObjectGraphScanner quite often, and we all know Reflection isn't fast. Disabling it reduced app startup from 8 sec to 1 sec, but that's by no means a solution.
Imagine my delight when discovering that someone was already improving it :smile:

Doing the scanning during Initialize is definitely an improvement.

Allowing GetAllLayouts to do lazyloading is a bit risky, coz that potentially conflicts with another thread calling GetLogger. I assume there aren't any guarantees that the user calls ReconfigExistingLoggers after configuring, right?

Isn't it ridiculous the amount of work done to determine what kind of stacktrace is needed...

304NotModified:
Well getting stacktrace costs a lot at runtime. 

But there is also a threadsafe issue here which I try to solve without many locks.

src/NLog/Internal/TargetWithFilterChain.cs
304NotModified: AFAIK that isn't needed. This call is invoked for every targerwithchain.  But please correct me if I missed something.

src/NLog/Internal/TargetWithFilterChain.cs
304NotModified: I have to go to simplelayout for this. Was thinking to calculate the maxStackTrace when parsing the simplelayout.

src/NLog/Internal/TargetWithFilterChain.cs
Taloth: From what I can see it's definitely needed: The root of the chain is used [here](https://github.com/NLog/NLog/blob/d40035b5f9389e8eef655ccfec56e11a96555b6d/src/NLog/LoggerImpl.cs#L69) to determine which kind of stacktrace is needed, then the chain itself is processed [later](https://github.com/NLog/NLog/blob/d40035b5f9389e8eef655ccfec56e11a96555b6d/src/NLog/LoggerImpl.cs#L97-L103) in a loop.
The PrecalculateStackTraceUsage is also only [called](https://github.com/NLog/NLog/blob/815bf16546ec4d0eec9d758b8d351ab98b036fb2/src/NLog/LogFactory.cs#L783) for the root of the chain, so only the StackTraceUsage of the root is valid.

304NotModified:
As the unit test fails irregular: 

Rebased other open PRs to repeatedly test this issue: https://github.com/NLog/NLog/pull/1310
https://github.com/NLog/NLog/pull/1309, #1299, #1286, #1266, #1261

304NotModified:
All PRs are succeeding so tested at least 12 times (6 PRs)

Taloth:
Performance is good, same 1 s startup time as when I disabled it entirely.

However, you still need to process `NextInChain`. (I assume you missed my comment from earlier.)   

Also, `FindAllLayouts()` currently recurses into `Layout` itself, which isn't needed. There is a performance gain when adding a predicate to the scanner, but it only becomes noticeable when I run ReconfigExistingLoggers like 100x in a loop, so personally i'm not too worried that it takes a whopping  8ms instead of 6ms.

304NotModified:
> However, you still need to process NextInChain. (I assume you missed my comment from earlier.)

Missed indeed your 2nd comment. Will check now. Thanks!

304NotModified:
> Also, FindAllLayouts() currently recurses into Layout itself, which isn't needed.

Can you elaborate on that? Thanks!

Taloth:
FindAllLayouts uses ObjectGraphScanner, which checks all properties on the Target and it's hierarchy. When it finds a Layout instance it adds it to the resultset but it also it dives into the properties of the Layout and it's children (LayoutRenderers for example). So it scans more objects that strictly required.
But as I said it's no big deal, the performance difference is negligible.

Issue #1330, opened 2016-03-23
FileCharacteristics to struct


304NotModified:
Proposed at https://github.com/NLog/NLog/pull/1100#issuecomment-175927220 by @ilya-g

304NotModified:
Tried this, found not performance gain, and as it against the guidelines (struct to large), this isnt merged. See https://github.com/NLog/NLog/pull/1334

PR #1334, opened 2016-03-26
FileCharacteristics to (nullable) stuct

304NotModified:
Fixes https://github.com/NLog/NLog/issues/1330

### DIFF
--- a/src/NLog/Internal/FileAppenders/BaseFileAppender.cs
+++ b/src/NLog/Internal/FileAppenders/BaseFileAppender.cs
@@ -114,7 +114,7 @@ namespace NLog.Internal.FileAppenders
         /// Gets the file info.
         /// </summary>
         /// <returns>The file characteristics, if the file information was retrieved successfully, otherwise null.</returns>
-        public abstract FileCharacteristics GetFileCharacteristics();
+        public abstract FileCharacteristics? GetFileCharacteristics();
 
         /// <summary>
         /// Performs application-defined tasks associated with freeing, releasing, or resetting unmanaged resources.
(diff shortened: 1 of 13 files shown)


304NotModified:
Needs performance tests. 
. 
Size 20 bytes?  Larger than recommended 16. 

See https://msdn.microsoft.com/en-us/library/ms229017(v=vs.110).aspx

bjornbouetsmith:
I have also looked into this in my pooling branch, and have made a pooled object instead, so that we do not recreate that object every single time.
It is not a gret deal it matteers according to my test, but every bit counts.
Even a struct as suggested above would be better, since it gets allocated on the stack and removed immediately.

304NotModified:
So in summary this PR is probably a good change?

bhaeussermann:
@bjornbouetsmith Would you mind to share some statistics from your performance tests (time before and after change)?

bjornbouetsmith:
There arent really any interesting statistics to share.
In the current version, we allocate one FileCharacteristics per LogEventInfo * Number of targets that need to receive the LogEventInfo.

When they are pooled, we just grab them from the pool. and put them back when done.
Thats a trade off, we get less GC and lose a tiny bit of performance, since new is fast, its only when we have to GC it turns into a problem.

So for low traffic logging, its not really a problem. Each FileCharacteristics hold two datetimes (16 bytes) + the size of the filename, which can be big if you are logging in a deep nested folder + object overhead, which is around 8 bytes as far as I remember.

So for each log entry we are allocating at least 24+size of filename, which is not a lot, but it adds up on a high traffic logging.

the problem is not the allocation as such, but that its a class that gets allocated on the heap, which can impact GC. If it were a struct, it would be allocated on the stack and should not impact GC as much, if at all.

So I am all for changing it into a struct as a good start.
Of course it would mean that it has to be copied across methods, but since it only contains two other structs and on pointer it should not matter much.

304NotModified:
About this PR, I'm unsure if this is an improvement.  Copying 20 bytes instead of 4-8 (reference) (probably multiple times) could impact much in terms of performance. Also the reference is kept short, so in terms of memory the improvement is small. 

Still need to test it, but there is a chance I won't merge this one.

304NotModified:
Copying 20 bytes instead of 4-8 (reference) + more then advice = no merge

PR #1343, opened 2016-03-29
LogFactory: Added current/working directory to candidate config file names for running with DNX.

### LINKED ISSUE #1342: Config file not found when running via .NET Core

```
```

**Type**:
- Bug

bhaeussermann:
Fix #1342


304NotModified:
Thanks! 

Have to think what the impact is on this change.

304NotModified:
I added a extra check so the candidate names won't return twice the same filename (can be confusing or lead to bugs). This is PR #1356

PR #1356, opened 2016-04-04
Also check current working directory for nlog.config (DNX compatible)

### LINKED ISSUE #1342: Config file not found when running via .NET Core

304NotModified:
Also check current working directory for nlog.config (DNX compatible)
But don't try twice the same file.

supersedes #1343

fixes #1342

### DIFF
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -847,9 +847,16 @@ namespace NLog
             yield return "NLog.config";
 #elif !SILVERLIGHT
             // NLog.config from application directory
-            if (CurrentAppDomain.BaseDirectory != null)
+            var baseDirectory = CurrentAppDomain.BaseDirectory;
+            if (baseDirectory != null)
             {
-                yield return Path.Combine(CurrentAppDomain.BaseDirectory, "NLog.config");
+                yield return Path.Combine(baseDirectory, "NLog.config");
+            }
+
+            var currentDirectory = Directory.GetCurrentDirectory();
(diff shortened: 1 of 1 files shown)


bhaeussermann:
Thanks! I will pull this to double-check if it works for me tomorrow before I merge it.

304NotModified:
I think @phenixdotnet is right and this change isn't good :(

bhaeussermann:
> I think @phenixdotnet is right and this change isn't good :(

That is a valid point, and unfortunately I don't know of another way to get to the right folder path.

For now, we are loading the config "manually" by constructing the `XmlLoggingConfiguration` object with the correct path and assigning it to `LogManager.Configuration`. I suppose "tricky" cases like these are to be handled this way until we can find a way for NLog to find the correct path automatically.

304NotModified:
> For now, we are loading the config "manually" by constructing the XmlLoggingConfiguration object with the correct path and assigning it to LogManager.Configuration

Well, that is also not a good solution, see https://github.com/NLog/NLog.Extensions.Logging/issues/24

bhaeussermann:
@304NotModified Thanks! I'll look into that.

phenixdotnet:
@304NotModified No ideas here, this thing drive me crazy. Same call (IE ```PlatformServices.Default.Application.ApplicationBasePath```) give you 2 differents behaviors based on app context

I hope it will be fixed in rc2.

304NotModified:
Anyway this isn't an improvement

PR #1359, opened 2016-04-04
Better CleanupInvalidFilenameChars

### LINKED ISSUE #1350: File target: can't use absolute path at root
**Type**: Bug
**NLog version**: 4.2.3
**Platform**: .Net 4.5
**Current NLog config** 
```xml
    <target fileName="D:\log.txt"  />
```
**What is the current result?**
`log.txt` is created in the current directory on D: drive.
**What is the expected result?**
`log.txt` is created at the root of D: drive.
**Workaround?**
Use `cleanupFileName="false"`.

The reason why it doesn't work is this line of code:
https://github.com/NLog/NLog/blob/master/src/NLog/Targets/FileTarget.cs#L1997

When cleaning up the filename, the current value is split in two but the last separator `\` isn't kept (on purpose or as an off-by-one error, I don't know).
For instance: 
- `some\path\log.txt` becomes `some\path` + `log.txt`
- `d:\log.txt` becomes `d:` + `log.txt`.

The pieces are later put back together with `Path.Combine`, but it won't give the expected results:
- `some\path\log.txt` is ok because the missing separator is added.
- `d:log.txt` is not ok! It becomes: `log.txt` in the current directory of the `d:` folder.

You can fix the line 1997 by keeping the last separator in the path.
But completely parsing Win32 paths properly is very hard! I suggest that you use `Path.GetFileName()` to extract the filename to clean and `Path.GetDirectoryName()` to get the path to the file. Doing so would fix the bug reported here.

**EDIT**: Also, since you seem to try allocations as much as possible in the code of `CleanupInvalidFileNameChars`. If the file was not cleaned up (there was no invalid char), there is no need to build a new result by combining the path and filename. The original can just be returned as is.

jods4:
Fixes #1350

1. Supports files at the root of an absolute path: D:\log.txt
2. No allocation if the filename was already clean. (Removed useless Path.Combine call).

### DIFF
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -1994,7 +1994,10 @@ namespace NLog.Targets
             var lastDirSeparator = fileName.LastIndexOfAny(DirectorySeparatorChars);
 
             var fileName1 = fileName.Substring(lastDirSeparator + 1);
-            var dirName = lastDirSeparator > 0 ? fileName.Substring(0, lastDirSeparator) : string.Empty;
+            // Keep the trailing "\" or "/" if it's not the first char in the string.
+            // This is important for files at the root like C:\log.txt, because Path.Combine("C:", "log.txt")
+            // would return "C:log.txt" which is a different thing: "log.txt" in the current directory of drive "C:"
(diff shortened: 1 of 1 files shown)


304NotModified:
Thanks for this PR, but this is already fixed by https://github.com/NLog/NLog/pull/1352

PR #1362, opened 2016-04-05
autoload nlog extensions also from currentdir (DNX compatible)

304NotModified:


304NotModified:
This solution isn't good

PR #1369, opened 2016-04-11
#reduce locks

emanov:
added the lock to support locking as iw was in the (don't know if it really neccessary)
 src/NLog/Internal/FileAppenders/FileAppenderCache.cs

Convert IsInitialized from the autproperty to the property with the volatile variable under the hood.

### DIFF
--- a/src/NLog/Internal/FileAppenders/FileAppenderCache.cs
+++ b/src/NLog/Internal/FileAppenders/FileAppenderCache.cs
@@ -265,6 +265,7 @@ namespace NLog.Internal.FileAppenders
         /// <param name="expireTime">The time which prior the appenders considered expired</param>
         public void CloseAppenders(DateTime expireTime)
         {
+            lock(this)
             for (int i = 0; i < this.appenders.Length; ++i)
             {
                 if (this.appenders[i] == null)
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -664,8 +664,7 @@ namespace NLog.Targets
                                 while (true)
                                 {
                                     Thread.Sleep(200);
-                                    lock (SyncRoot)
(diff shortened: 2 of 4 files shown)


304NotModified:
As you can see "NLog.UnitTests.Targets.TargetTests.LockingTest" is failing now. I didn't checked the test case in detail, but I think the PR should be fixed

emanov:
Yep. I didn't see the test before. Thanks I will check it.

304NotModified:
I think this isn't easy for fix?

emanov:
it's easy, but I've tested the performance and didn't saw good results compare with base. So, I'll try to optimize and if results will be good enough, I will make commit again.

304NotModified:
OK thanks for the feedback!

src/NLog/Internal/FileAppenders/FileAppenderCache.cs
```diff
@@ -265,6 +265,7 @@ public void CloseAppenders()
         /// <param name="expireTime">The time which prior the appenders considered expired</param>
         public void CloseAppenders(DateTime expireTime)
         {
+            lock(this)
```
bjornbouetsmith: lock(this) is a bad practise, since you lock the entire instance.

cosider:

var appenderCache = xxxx.

lock(appenderCache)
{
Thread.Sleep(TimeSpan.FromHours(5));
}

while inside AppenderCache instance
lock(this)
{
 is blocked 5 hours.
}

Its always better to make a class private lock objects.

304NotModified:
@emanov is this still needed?

emanov:
Yes, it is still needed. The problem is here. now we are using workaround: 10 processes with 120 threads per process.

304NotModified:
@emanov please send a new PR when you got time! thanks!

Issue #1297, opened 2016-03-08
Create constructors with name for every target


304NotModified:
For easier API usage.

~ 38 targets

- [ ] Add Constructor with `string name`. Call `this()`, after the default ctor
- [ ] Add default constructor if it isn't there
- [ ] write unit test with reflection to test all constructors (default + named). test also that default+named ctor will yield the same results.

NB: no extra contructor needed in base class (Target) because this will lead to (more) code duplication.

Before:

```c#
/// <summary>
/// Initializes a new instance of the <see cref="FileTarget" /> class.
/// </summary>
/// <remarks>
/// The default value of the layout is: <code>${longdate}|${level:uppercase=true}|${logger}|${message}</code>
/// </remarks>
public FileTarget()
{
	this.ArchiveNumbering = ArchiveNumberingMode.Sequence;
	this.maxArchiveFiles = 0;
	this.ConcurrentWriteAttemptDelay = 1;
	....
}  
```

after
```c#
/// <summary>
/// Initializes a new instance of the <see cref="FileTarget" /> class.
/// </summary>
/// <remarks>
/// The default value of the layout is: <code>${longdate}|${level:uppercase=true}|${logger}|${message}</code>
/// </remarks>
public FileTarget()
{
	this.ArchiveNumbering = ArchiveNumberingMode.Sequence;
	this.maxArchiveFiles = 0;
	this.ConcurrentWriteAttemptDelay = 1;
      ....
}
/// <summary>
/// Initializes a new instance of the <see cref="FileTarget" /> class.
/// </summary>
/// <param name="name">Name of the target.</param>
/// <remarks>
/// The default value of the layout is: <code>${longdate}|${level:uppercase=true}|${logger}|${message}</code>
/// </remarks>
public FileTarget(string name) : this()
{
	this.Name = name;
}    
```

tetrodoxin:
The tricky part is, to compare resulting objects, because if you compare them property-wise by reflection, you have to deep-compare `Layout` objects, since they do not implement a customized `Equals` method.

304NotModified:
The layout is a simple case, as those are mostly simplelayout and a tostring is sufficient.  Anyway, those are addressed in the PR. :)

PR #1370, opened 2016-04-12
Added constructors with name argument to the target types

flyingcroissant:
Resolution for #1297. Seems to build with Appveyor. Let me know if there are any issues.


304NotModified:
\0/

304NotModified:
Thanks also for the unit test, but locally I get a stackoverflow? 

image

304NotModified:
update that is fixed after rebase

304NotModified:
this need some change as the current unit test wont fully work. Trying to fix it :)

`Typeof(Target).GetProperties()` was only returning "Name", which was excluded in the test case ;)

flyingcroissant:
I guess I should have checked if there were any other public properties, haha :)  The Type.GetProperties MSDN documentation says you can use BindingFlags parameters to change what it returns (eg. BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.Static). That set of flags seems to give the protected properties as well.

When I make that change, the test fails because (at least) the AspResponseTarget.IsInitialized property is different with the two constructors. I haven't looked into it too much yet (it looks like the same thing should be happening for both?), but should I be calling Initialize() on each Target before comparing?

flyingcroissant:
Also FYI, I had to filter out the SyncRoot property in the test since that's expected to be a different object in both. Maybe there's a smarter way to do that filtering and handle future property additions.

flyingcroissant:
I think I fixed the other issue I was having too. It was because the assertion was comparing the property value from one instance against the other instance (ie. not the property value from that instance).

tests/NLog.UnitTests/Targets/TargetTests.cs
```diff
@@ -65,13 +65,16 @@ public void ConstructorsTest()
                         namedConstructedTarget = (Target)Activator.CreateInstance(candidateType, candidateType.ToString());
 
                         // Check that the created targets are the same except for name
-                        foreach (System.Reflection.PropertyInfo pi in (typeof(Target)).GetProperties())
+                        foreach (System.Reflection.PropertyInfo pi in (typeof(Target)).GetProperties(
```
304NotModified: `typeof(Target)` => `typeof(candidateType)` ?

304NotModified:
I see you have changed the test, but the test aren't working now, as it should fail (there are some `this()` calls missing in the PR)

304NotModified:
@flyingcroissant see https://github.com/flyingcroissant/NLog/pull/1

flyingcroissant:
Thanks, I've merged in your changes. I'll see what I can think of for doing the property comparisons.

304NotModified:
superseded by https://github.com/NLog/NLog/pull/1434

PR #1377, opened 2016-04-16
#1351 - Getting the 'From' when UseSystemNetMailSettings is true

### LINKED ISSUE #1351: MailTarget: Read 'from' from appsettings
When reading from .config, we need also read the `from` attribute:

e.g.

```xml
<configuration>
  <system.net>
    <mailSettings>
      <smtp deliveryMethod="SpecifiedPickupDirectory" from="bugs@mail.me">
      </smtp>
    </mailSettings>
  </system.net>
</configuration>
```

See http://stackoverflow.com/questions/36351209/any-way-to-use-from-address-from-system-net-mailsettings-smtp-in-nlog-mail-tar

Double check if we can overwrite the from when `UseSystemNetMailSettings` is set to `false`

MoaidHathot:
fixes #1351 - Using the value of the 'From' attribute in system.net/mailSettings/smtp when UseSystemNetMailSettings is true

### DIFF
--- a/src/NLog/Targets/MailTarget.cs
+++ b/src/NLog/Targets/MailTarget.cs
@@ -33,7 +33,7 @@
 
 using JetBrains.Annotations;
 
-#if !SILVERLIGHT && !__ANDROID__ && !__IOS__
+#if !SILVERLIGHT
 
 namespace NLog.Targets
 {
@@ -42,14 +42,18 @@ namespace NLog.Targets
     using System.ComponentModel;
     using System.Net;
     using System.Net.Mail;
-    using System.Configuration;
-    using System.Net.Configuration;
     using System.Text;
     using System.IO;
     using NLog.Common;
     using NLog.Config;
     using NLog.Internal;
     using NLog.Layouts;
+
+    // For issue #1351 - These are not available for Android or IOS
+#if !__ANDROID__ && !__IOS__
+    using System.Configuration;
+    using System.Net.Configuration;
+#endif
 
     /// <summary>
     /// Sends log messages by email using SMTP protocol.
@@ -105,7 +109,7 @@ namespace NLog.Targets
             this.Timeout = 10000;
         }
 
-
+#if !__ANDROID__ && !__IOS__
         private SmtpSection _currentailSettings;
 
         /// <summary>
@@ -126,6 +130,7 @@ namespace NLog.Targets
 
             set { _currentailSettings = value; }
         }
+#endif
 
         /// <summary>
         /// Gets or sets sender's email address (e.g. joe@domain.com).
@@ -307,7 +312,6 @@ namespace NLog.Targets
         /// </summary>
         protected override void InitializeTarget()
         {
-            InitializeFrom();
             CheckRequiredParameters();
 
             base.InitializeTarget();
@@ -318,12 +322,14 @@ namespace NLog.Targets
         /// Thus, when UseSystemNetMailSettings is enabled we have to read the configuration section of system.net/mailSettings/smtp to initialize the 'From' address.
         /// It will do so only if the 'From' attribute in system.net/mailSettings/smtp is not empty.
         ///  </summary>
-        private void InitializeFrom()
+        private void ConfigureFrom()
         {
-            if (UseSystemNetMailSettings)
-            {
-                From = string.IsNullOrEmpty(MailSettings.From) ? From : MailSettings.From;
-            }
+#if !__ANDROID__ && !__IOS__
+                if (!string.IsNullOrEmpty(MailSettings.From))
+                {
+                    From = MailSettings.From;
+                }
+#endif
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


304NotModified:
Thanks for this PR! I did not checked he code in depth yet, but isn't it strange that we need a different construction for reading the "from"  compared to other values from the config file? 

The build on appveyor fails because smtpsection isn't available on (at least) android. You need `#if !___ANDROID___`

MoaidHathot:
Thanks for your comment!
From what I can see, NLog isn't reading the other config file values by itself. NLog is creating a System.Net.Mail.SmtpClient (by creating a NLog.Internal.MySmtpClient class) object and that smtp client is reading the rest of the values from the config file by itself. Since it is not reading the 'From' value I had to add that functionality to MailTarget.

I'll fix the AppVeyor, thanks!

304NotModified:
The other is `___IOS___` ;)

MoaidHathot:
Can you help me please check why the CI has failed? I can't understand from the logs how my changes caused this.

Thanks :)

304NotModified:
Also unclear to me why it fails. Will retry. 

Anyway, before the MailTarget would work in ios/android, although not sure if that makes sense, now the full target is excluded?

MoaidHathot:
Yes, now the full MailTarget doesn't support Android/ios, as some System.Configuration assemblies are missing.

Thanks for your help!

dimrun:
Hi Guys,
Do you have any plans (expected dates) when 4.3.2 is going to be published?

304NotModified:
Not planned yet but I could release this feature this/next week

MoaidHathot:
Are there anything else that I can do, or that should be done from my side?

304NotModified:
@MoaidHathot thanks!

Will do a code review. Some small things I guess

src/NLog/Targets/MailTarget.cs
```diff
@@ -31,10 +31,9 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
-
 using JetBrains.Annotations;
 
-#if !SILVERLIGHT
+#if !SILVERLIGHT && !__ANDROID__ && !__IOS__
```
304NotModified: please move those `#if !__ANDROID__ && !__IOS__` to a minimal scope. For backwards-compatibility

304NotModified:
image

What happend here? Formatting? Newlines? Can you fix this?

src/NLog/Targets/MailTarget.cs
```diff
@@ -284,13 +307,26 @@ protected override void Write(AsyncLogEventInfo[] logEvents)
         /// </summary>
         protected override void InitializeTarget()
         {
-
+            InitializeFrom();
```
304NotModified: I think this call should be in `ConfigureMailClient`?

src/NLog/Targets/MailTarget.cs
```diff
@@ -284,13 +307,26 @@ protected override void Write(AsyncLogEventInfo[] logEvents)
...
         /// <summary>
+        /// In contrary to other settings, System.Net.Mail.SmtpClient doesn't read the 'From' attribute from the system.net/mailSettings/smtp section in the config file.
+        /// Thus, when UseSystemNetMailSettings is enabled we have to read the configuration section of system.net/mailSettings/smtp to initialize the 'From' address.
+        /// It will do so only if the 'From' attribute in system.net/mailSettings/smtp is not empty.
+        ///  </summary>
+        private void InitializeFrom()
+        {
+            if (UseSystemNetMailSettings)
+            {
+                From = string.IsNullOrEmpty(MailSettings.From) ? From : MailSettings.From;
```
304NotModified: This is correct, but reads a bit strange. If `MailSettings.From == null`, then we do `From = From`. Please rewrite to a `if`

304NotModified:
Reviewed! I need to see the unit test before I can decide if we merge this. But I think I can merge this soon :)

src/NLog/Targets/MailTarget.cs
MoaidHathot: I had put that in the constructor because it is easier to Unit test this way. I'll work on rewriting the Unit Tests if you think it will be better to put it in `ConfigureMailClient` :)

src/NLog/Targets/MailTarget.cs
MoaidHathot: Ok, np. This change will be part of my next commit (probably later tonight or tomorrow)

MoaidHathot:
@304NotModified  I have committed the changes according to your comments :)

304NotModified:
testing this now, also trying to fix the newline issue in the testclass

304NotModified:
found the issue in the test class, the spaces were replaced with tabs ;)

304NotModified:
rebased and fixed it in https://github.com/NLog/NLog/pull/1402

Thanks!

MoaidHathot:
Great, thanks! :D

304NotModified:
@dimrun 

This has been released in 4.3.2 and is now online! https://www.nuget.org/packages/NLog/4.3.2

Issue #1381, opened 2016-04-18
StackOverflowException when File target's fileName uses ${var:mydir} instead of ${mydir}


AndreGleichner:
I'm using a variable "mydir" to hold the actual path where a file target should create its log file.
When referencing it via ${var:mydir} it throws a StackOverflowException.
When using the (no longer recomended?) `${mydir}` it works.
Have used NLog 4.2.3 before which works fine with `${var:mydir}`.
My actual NLog.config is much more complicated, so that the use of a variable makes sense...

**Type**: Bug
**NLog version**: 4.3.0
**Platform**: .Net 4.0 on Win10

**Current NLog config** (xml or C#, if relevant)

```xml
<?xml version="1.0" encoding="utf-8" ?>
<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd"
      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
      xsi:schemaLocation="http://www.nlog-project.org/schemas/NLog.xsd NLog.xsd"
      autoReload="true"
      throwExceptions="false"
      internalLogLevel="Trace" internalLogFile="c:\temp\nlog-internal.log" >
    
    <variable name="mydir" value="c:\temp\HelloNlog"/>
    <targets>
        <target name="fileAll" xsi:type="File" fileName="${var:mydir}\Test.log" />
    </targets>
    <rules>
        <logger name="*" writeTo="fileAll" />
    </rules>
</nlog>

```

In case of a BUG:

- What is the current result? StackOverflowException
- What is the expected result? 
- Did you checked the [Internal log](https://github.com/NLog/NLog/wiki/Internal-Logging)?
 Yes, but gives no clear indication of what went wrong.
- Please post full exception details (message, stacktrace, inner exceptions)
 See below.
- Are there any work arrounds? yes, use ${mydir} instead of ${var:mydir}
- Is there a version in which it did worked? 4.2.3
- Can you help us by writing an unit test? 
 Don't know enough about NLog's inner workings.

It recurses endlessly invoking NLog.Targets.FileTarget.FileName.set.
While doing a short debug session, I wasn't able to spot the root cause.

**Stacktrace:**

```
 	...
 	NLog.dll!**NLog.Targets.FileTarget.FileName.set**(NLog.Layouts.Layout value) Line 240
[...]

PR #1383, opened 2016-04-18
Fix configuration usage in `${var}` renderer

304NotModified:
- Fix configuration usage in `${var}` renderer; don't use `LogManager`, that can lead to locks/inifite recursion
- Init layoutrenders correctly inside other layoutrenders
- fix unit tests (if `.render()` is called, call`.Initialize()` first)

fixes https://github.com/NLog/NLog/issues/1381

### DIFF
--- a/tests/NLog.UnitTests/LogFactoryTests.cs
+++ b/tests/NLog.UnitTests/LogFactoryTests.cs
@@ -234,6 +234,31 @@ namespace NLog.UnitTests
                 </rules>
             </nlog>");
         }
+
+        [Fact]
+        public void ValueWithVariableMustNotCauseInfiniteRecursion()
+        {
+            LogManager.Configuration = null;
+            
+            File.WriteAllText("NLog.config", @"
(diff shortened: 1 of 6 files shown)


304NotModified:
Thanks @bhaeussermann ! Looks good.

Now let's wait on AppVeyor

tests/NLog.UnitTests/LogFactoryTests.cs
```diff
@@ -234,6 +234,31 @@ public void NewAttrOnNLogLevelShouldNotThrowError()
...
+                <targets>
+                    <target name='f' type='file' fileName='${var:dir}\test.log' />
+                </targets>
+                <rules>
+                    <logger name='*' writeTo='f' />
+                </rules>
+            </nlog>");
+            try
+            {
+                LogManager.Configuration.ToString();
```
304NotModified: This is a good test! Can confirm it giving me a stackoverflow in the master! Great work!

PR #1392, opened 2016-04-22
Remove test dependency on existing machine users

luigiberrettini:
Since some tests were not working because requiring a user present on the machine I fixed them.

**NLog version**
4.3.0

**Work arounds**
None

**Config when implemented**
No changes


304NotModified:
Thanks, can you remove the creation of this user in appveyor.yml (CreateTestUsers.cmd etc, those can be removed)

luigiberrettini:
Done :)

304NotModified:
Thanks! This is indeed a better solution.  :)

PR #1394, opened 2016-04-22
Make private tests public

luigiberrettini:
Since some tests were not public I fixed them.


304NotModified:
Why is this needed? Xunit can handle private tests?

luigiberrettini:
XUnit handles them, but if you did not do it on purpose it is common to have public tests

304NotModified:
Thanks, will merge this for consistency reasons.

PR #1397, opened 2016-04-23
stream and byte array pooling to prevent excessive garbage collection

bjornbouetsmith:
Initial version of my pooling changes to prevent excessive garbage collection.

Result of my test program that runs for 30 seconds in a tight loop for each message size and logs results.

nlog configuration for targets is as below

`<target name="async" xsi:type="AsyncWrapper" overflowAction="Block" batchSize="10000" queueLimit="20000">
      <target xsi:type="File" name="f" fileName="e:\nlog_perftest.log"
              layout="${longdate} ${uppercase:${level}} ${message}" discardAll="false" keepFileOpen="true" archiveAboveSize="100000000"/>
    </target>
`
**Unmodified code**

| Message Size (characters) | Logged items | Gen0 collections | Gen1 collections | Gen2 collections |
| --: | --: | --: | --: | --: |
| 2 | 3090001 | 572 | 395 | 197 |
| 4 | 3060001 | 612 | 421 | 209 |
| 8 | 3020001 | 801 | 540 | 268 |
| 16 | 3010001 | 620 | 425 | 212 |
| 32 | 2940001 | 775 | 485 | 226 |
| 64 | 2796724 | 972 | 654 | 303 |
| 128 | 2580001 | 1167 | 791 | 369 |
| 256 | 2240001 | 1538 | 1087 | 451 |
| 512 | 1890001 | 1709 | 1021 | 446 |
| 1024 | 1410001 | 2331 | 1632 | 470 |
| 2048 | 660001 | 1910 | 1183 | 242 |
| 4096 | 400001 | 2094 | 1302 | 165 |
| 8192 | 230001 | 2239 | 1475 | 103 |

Unhandled Exception: OutOfMemoryException.

**The forked code, same config file (First iteration)**

| Message Size (characters) | Logged items | Gen0 collections | Gen1 collections | Gen2 collections |
| --: | --: | --: | --: | --: |
| 2 | 3348023 | 241 | 208 | 6 |
| 4 | 3360001 | 240 | 180 | 5 |
| 8 | 3290001 | 248 | 181 | 5 |
| 16 | 3250001 | 261 | 184 | 5 |
| 32 | 3190001 | 290 | 209 | 6 |
| 64 | 3050001 | 344 | 257 | 6 |
| 128 | 2880001 | 453 | 344 | 6 |
| 256 | 2484483 | 602 | 457 | 12 |
| 512 | 1980018 | 823 | 739 | 32 |
| 1024 | 1380001 | 1070 | 936 | 62 |
| 2048 | 910011 | 1370 | 1239 | 91 |
| 4096 | 560001 | 1574 | 1496 | 60 |
| 8192 | 320001 | 1828 | 1767 | 52 |

As you can see from the results theres a huge difference in garbage collections and even though the garbage collections runs fast in my test program, if you have a service that uses a lot of memory, then this many gc2 collections would take many seconds. 

**The forked code, same config file (Second iteration)**

| Message Size (characters) | Logged items | Gen0 collections | Gen1 collections | Gen2 collections |
| --: | --: | --: | --: | --: |
| 2 | 3201398 | 24 | 6 | 3 |
| 4 | 3177030 | 23 | 5 | 3 |
| 8 | 3205944 | 24 | 5 | 2 |
| 16 | 3040933 | 24 | 5 | 2 |
| 32 | 3160521 | 24 | 6 | 3 |
| 64 | 3121334 | 24 | 5 | 2 |
| 128 | 3116056 | 25 | 6 | 3 |
| 256 | 2785667 | 23 | 5 | 2 |
| 512 | 2717137 | 23 | 5 | 2 |
| 1024 | 2222614 | 23 | 5 | 2 |
| 2048 | 753654 | 17 | 4 | 2 |
| 4096 | 366671 | 15 | 4 | 1 |
| 8192 | 205774 | 8 | 2 | 0 |

**The forked code, same config file (Second iteration) - delegates optimized and some pooled**

| Message Size (characters) | Logged items | Gen0 collections | Gen1 collections | Gen2 collections |
| --: | --: | --: | --: | --: |
| 2 | 3580001 | 12 | 1 | 0 |
| 4 | 3580001 | 11 | 2 | 0 |
| 8 | 3540001 | 11 | 2 | 0 |
| 16 | 3570001 | 11 | 2 | 0 |
| 32 | 3600001 | 11 | 2 | 0 |
| 64 | 3550001 | 11 | 2 | 0 |
| 128 | 3480001 | 11 | 2 | 0 |
| 256 | 3450001 | 11 | 2 | 0 |
| 512 | 3020001 | 10 | 2 | 0 |
| 1024 | 2100001 | 7 | 2 | 0 |
| 2048 | 750001 | 3 | 2 | 0 |
| 4096 | 350001 | 1 | 1 | 0 |
| 8192 | 220001 | 1 | 1 | 0 |

**The forked code, same config file (Third iteration) - more pooling and StringBuilderSlim removed**

| Message Size | Logged items | Gen0 collections | Gen1 collections | Gen2 collections |
| --: | --: | --: | --: | --: |
| 2 | 3350001 | 0 | 0 | 0 |
| 4 | 3410079 | 0 | 0 | 0 |
| 8 | 3330075 | 0 | 0 | 0 |
| 16 | 3310075 | 0 | 0 | 0 |
| 32 | 3450005 | 0 | 0 | 0 |
| 64 | 3340057 | 0 | 0 | 0 |
| 128 | 3160059 | 0 | 0 | 0 |
| 256 | 3150076 | 0 | 0 | 0 |
| 512 | 2820003 | 0 | 0 | 0 |
| 1024 | 2030051 | 0 | 0 | 0 |
| 2048 | 740001 | 0 | 0 | 0 |
| 4096 | 350006 | 0 | 0 | 0 |
| 8192 | 190028 | 0 | 0 | 0 |

---

This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1397)


304NotModified:
:+1:

bjornbouetsmith:
Okay, my code is building locally again - I have similar performance, and a little less degredation in performance with increasing message sizes.

And between 5-10 times less gen2 collections, which is because of the less allocations.

I will commit something to look at tomorrow perhaps, my time is tight in the work week, but I will try.

304NotModified:
Nice progress! 

I'm also busy this week, so no problem is it will be next week. 

Is it correct the the tables in your start post only has headers?

304NotModified:
Ps you need to add the headers to all new files. Just copy it from another file.

bjornbouetsmith:
Okay updated comment above with current performance / gc numbers. Good so far, I am still working on tweaking performance - I am sure it can become even better :-)

bjornbouetsmith:
So updated with more pooling and a few performance tweaks, still more to be pooled, but its looking pretty good so far.

I have a few tests failing that I just need to figure out why, so no merging just yet :-)

I have added some usafe pointer code to speed up character copying and also prevent allocating arrays.

Its wrapped in #if NET4_5 which I hope is for windows machines, if not we have do discuss. And whether or not unsafe code is a no go or not.

bjornbouetsmith:
I have made a compiler setting so that when you compile with POOLING, then caching in the simple layout is turned off, since it defeats the purpose of pooling, since we have to do a ToString() on the string builders and that allocate a string.

304NotModified:
I'm curious what the performance impact of disabling caching will be. 

Nice that compiler conditional, too bad we have too choose before creating a package ;) 

About the pointers, I think they are only interesting if they give a huge performance improvement.  (at least 10%)

bjornbouetsmith:
I have only disabled caching when running with pooling - and if we keep caching and pooling, we lose around 4% performance on a 2 char message and get double the amount of gen0 and gen1 garbage collections and with bigger message sizes the performance loss will be even bigger.

So basically caching is not good when running with pooled objects.

In regards to pointers, if we have to stop using them, then when logging a string, we have to do a .ToCharArray(), which allocates a char array - with pointers we just cast the string to a char* and copy the contents into the cached char array.

So the point of using pointers is not performance, but limiting the amount of garbage we generate, since we can use the data directly when using pointers, but without we have to copy them to a new char array. I have run a test without using pointers, but still keeping the rest of the code.

Performance is similar, but GC impact i huge, the garbage collection increases between 4-100 times dependent on the generation.

bjornbouetsmith:
okay, added another sample of current performance. I pooled a delegate type and prevented allocations of a lot of delegates, so right now Gen0 -Gen2 collections are pretty good :-) I will take a look and see if I can improve performance on larger message sizes. since you can see that performance drop a lot with large messages.

I know most people dont write such big messages, but still it would be nice if performance did not drop that much.

304NotModified:
Maybe I missed it, but what is a big message,  In chars/bytes?

bjornbouetsmith:
As you can see of the tables I have updated in the original post above, there is a steep drop off @ around 256 characters with the new code, and even earlier with the original code.

304NotModified:
nice!

To be clear, I will only merge this after some performance impact is clear, so we need to test this a lot.

bjornbouetsmith:
I have run a lot of performance tests, I can submit my performance test as part of the solution. then you can pull it down yourself and try it out.

The goal when I started was not to increase performance, but to decrease the memory that NLog allocated.

If you look at the first table above, which is the original code and the last table above, you can see that the performance is comparable to the old version. Faster in some cases by 15% and garbage collections are so many times lower its almost ridiculus.

Last problems I am facing is that tests will not run on the server. I have successfully run under .NET4.5 & Mono on my own computer, but on the server something goes wrong. I need to figure out what before any merging takes place naturally.

304NotModified:
Yes the goal of the PR is clear, and reached :) 

But of course this should not negative impact the time/CPU, or at least not significant.

304NotModified:
Ps those losing events are critical (AsyncTargetWrapperFlushTest), see appveyor

bjornbouetsmith:
Yes I know :-) I would not recommend merging before all tests pass as they do on my machine.

304NotModified:
I will certainly not merge to soon ;)

bjornbouetsmith:
Hmm, last tests that fail is on a 32-bit x-unit runner - the 64-bit versions are running just fine - thats weird.

304NotModified:
FYI that unit test is proven stable in the past (too bad not the case for all unit tests)

bjornbouetsmith:
Since its a 32bit test that fails - it might be because the pool sizes are too big to handle. I have increased pool sizes, since that equates to less garbage when running. But I think with the current comitted code its about 1Gig of memory, and that might be problematic.

I will try to decrease the pool sizes to something smaller and see if that fixes the 32bit issue.

304NotModified:
How big is the pool in bytes?

bjornbouetsmith:
Its easier to see if you look in PoolFactory, there are several object pools. But when I run with a memory profiler I can see that LOH is around 900 Megs to 1Gig - and I had pooled objects set to 90k for arrays, which is char arrays and that is at least 180k per array and 1k char arrays so thats 180Megs.
StringBuilder pools of at least same size and other pools, so its big :-), but thats only for testing, before merging it in, I want the pool sized configurable via the nlog configuration.

304NotModified:
Thanks, sounds like a good plan

bjornbouetsmith:
The failing test was broken by design.
The continuation that is doing assertions is running on another thread, so there is no guarantee that it will get the latest value from the field.

I have changed it to use Interlocked instead, since that makes a memory barrier and guarantee that you get the latest value from the address.

bjornbouetsmith:
Hmm - the last few builds have failed with an error deep down in X-Unit and StringBuilder under Mono. I doubt that has anything to do with my code changes. Have you seen that error before?

304NotModified:
> Hmm - the last few builds have failed with an error deep down in X-Unit and StringBuilder under Mono.

Travis (mono) just builds?

Which unit test crashes?

bjornbouetsmith:
How do I get things added to the NLog.xsd?

304NotModified:
> How do I get things added to the NLog.xsd?

That is fixed with an msbuild script. No changes needed from your side

bjornbouetsmith:
When it crashed, I am not sure which test caused it, but it seems like it was x-unit when it was trying to report either progress or an error

bjornbouetsmith:
Surely if I want to add an element directly under NLog something needs to be done?

i.e. 

<nlog>
<pooling enabled="true" autoIncrease="false" preFill="false" outputPoolStatisticsInLogFiles="false
estimatedMaxMessageSize="2048" estimatedLogEventsPerSecond="100"/>
</nlog>

bjornbouetsmith:
Question - do I update the Copyright statement on the files/classes that I have created from scratch? Or do I just leave it as is?

304NotModified:
Yes, just copy it from another file.  See output log.

bjornbouetsmith:
Oh I should have been more precise in my question:

I mean, do I add my list to the list of names, since its not code that I am alterning but creating?

304NotModified:
I'm not sure what's common in Oss, but that's not possible now -   the header is checked with LICENSE.TXT.

bjornbouetsmith:
Okay no worries :-) I just wanted to know.
Progress is good btw, I am testing with the new configuration format for pooling and so far good.

bjornbouetsmith:
FYI current version of the code is not performing as well as the above results. I think it has to do with me changing code so pooling can be enabled via config file.

I will look at it and try to see if I can get it as good as it was before.

304NotModified:
I will check and tear the code when you're done :) 

FYI : I expect some changes are needed in the final version as too many code has changed (all? LayoutRenderer etc) . It should be possible to fix things in a more generic way v

bjornbouetsmith:
Yes a lot of code has changed, but mostly required changes so that allocations could be prevented. I will go through the source code and see if something can be removed.

304NotModified:
Maybe we could move some to LayoutRenderer itself?

304NotModified:
I did not checked in depth, but the following questions arise:

- doneWithEvent should be on LayoutRenderer scope? 

- Is the StringbuilderSlim really needed? 
- is allocating and changing a char array better then a stringbuilder? I'm not sure about that, but isnt good to maintain/keep bug free

bjornbouetsmith:
It might be possible to just have DoneWithEvent at the LayoutRenderer - I just wanted to be sure that all code that touches the LogEventInfo is done with it, so we dont accidently put a LogEventInfo back into the pool and some thread is not done with it.

The reason why I introduced StringBuilderSlim is that it is non-allocating, meaning it can write itself to a stream without generating any garbage.

StringBuilder in itself is nice, but when you want data out of it you have to either loop over all the chars in the stringbuilder, i.e. foreach(char c in strinbuilder) - and write each char to the stream, or do a ToString().
First option is very slow, other option is relatively fast, but generates garbage which is what I am trying to prevent.

I know it sucks to have to maintain and keep custom code bug free, but I will make sure that I have plenty of tests for the string builder slim.

304NotModified:
Too bad I don't think we can get rid of the StringBuilder, as that will break all custom layout renderers - that's a no go.

We have to aim to a sub-optimal performance, with the minimal amount of breaking changes.

bjornbouetsmith:
I am not planning to get rid of stringbuilder, but we can change the internals of the code to use stringbuilderslim.
I added
`internal virtual void Append(StringBuilderSlim builder, LogEventInfo logEvent)` 
so we can change the built in layout renderers to use stringbuilder slim, but custom layout renderers will still use stringbuilder. Which is "fine", since default implementation of the above just calls the Append method that takes a stringbuilder.

bjornbouetsmith:
btw - it seems like its always 32-bit unit tests that is failing for me, is there a history of the 32-bit unit tests being unstable?

304NotModified:
The 32bits tests were more unstable in the past regarding race conditions.  That should be solved.

bjornbouetsmith:
I am beginning to test the impact of removing the custom stringbuilder and replace it with the normal string builder. It seems like it will actually give better performance and if done correctly should not have much of a garbage collection impact.

304NotModified:
Nice!

bjornbouetsmith:
Okay - current version does not allocate very much.

|Message Size|  Logged items|   Gen0 collections|       Gen1 collections|       Gen2 collections|
|--:|   --:|    --:|    --:|    --:|
|2|             3350001|                0|                      0|                      0|
|4|             3410079|                0|                      0|                      0|
|8|             3330075|                0|                      0|                      0|
|16|            3310075|                0|                      0|                      0|
|32|            3450005|                0|                      0|                      0|
|64|            3340057|                0|                      0|                      0|
|128|           3160059|                0|                      0|                      0|
|256|           3150076|                0|                      0|                      0|
|512|           2820003|                0|                      0|                      0|
|1024|          2030051|                0|                      0|                      0|
|2048|          740001|         0|                      0|                      0|
|4096|          350006|         0|                      0|                      0|
|8192|          190028|         0|                      0|                      0|

Even though all numbers are 0, its not all true - but nothing of importance gets allocated when the logging is running. The reason for all the zeros in the above list is that after each run with a new message size I manually call GC.Collect. The numbers above is garbage collections that is forced upon the process when the test is running. Which is none - if I let the test run for 30 minutes, then possibly there would be garbage collections, but all that gets allocated during logging if running with my logging configuration above, is structs and AsyncContinuations. which is very little.

I have not optimized all appenders/targets, but its possible to go through all appenders and optimize them in the same way I optimized the UpperCaseLayoutRenderer, where it uses a char array to do the upper casing, which results in 0 allocations compared to just callling ToUpper on the result.

bjornbouetsmith:
Okay I think we should think about a review of what I have done, I am pretty satisfied so far :-)
So whenever you have time, please take a look at the files changed.
Some of the files I cannot understand why they are marked as changed. They are files I have pulled in from the main branch, whenever github said that I had to do it.

304NotModified:
If you're happy, is is possible to squash and rebase?

304NotModified:
I will have to check how to deal with the breaking changes. Maybe some changes can be moved to non breaking. Otherwise, i cannot merge this for NLog 4

bjornbouetsmith:
What breaking changes do you mean? I have tried to make everything backwards compatible.
I have created new virtual methods, but existing methods are still there?

304NotModified:
You can easily squash with tortiosegit (log, contextmenu, combine to one commit)

bjornbouetsmith:
Everything works perfect except the tests when running under mono.
I cannot debug the tests under mono, since when I run with latest mono, i.e. same version as travis - then it crashes. I am certain that the issue is a combination of mono and the way the unit tests are written. I am certain that the code works, otherwise I would also see errors when running under other environments, like SL etc.
Its really annoying and I have no idea as to how to fix it. I am stumbling blindly tweaking that trying this and nothing helps.

304NotModified:
> I am certain that the issue is a combination of mono and the way the unit tests are written. I am certain that the code works, otherwise I would also see errors when running under other environments, like SL etc.

not sure about that. The unit tests did work, so dunno why those are broken indeed.

bjornbouetsmith:
There is a problem with Tests bleeding over to each other. Thats one of the problems with having a static class.

Example from latest build:

Where a pooling test bled over into another test where pooling is not enabled.

Basically _EVERYTHING_ static needs to be reset to their correct state before every tests. Not just LoggingConfiguration.

 NLog.UnitTests.Fluent.LogBuilderTests.DebugWrite [FAIL]
      Assert.Equal() Failure
      Position: First difference is at position 0
      Expected: Error reading file '{0}'.
      Actual:   Pool Name|Pool Size|Objects in Pool|Objects in use|Objects given out|Objects thrown away|
                NLog.Internal.Pooling.CharArrayPool|1|0|0|0|0|
                NLog.Internal.Pooling.AsyncLogEventInfoListPool|1|0|0|0|0|
                NLog.Internal.Pooling.ByteArrayPool|4|0|0|0|0|
                NLog.Internal.Pooling.StringBuilderPool|4|2|0|0|0|
                NLog.Internal.Pooling.SingleCallContinuationPool|1|1|0|0|0|
                NLog.Internal.Pooling.LogEventInfoPool|2|1|0|0|0|
                NLog.Internal.Pooling.FileNameDictionaryPool|20|0|0|0|0|
                NLog.Internal.Pooling.FileCharacteristicsPool|1|0|0|0|0|
                NLog.Internal.Pooling.AsyncContinuationListPool|4|0|0|0|0|
                NLog.Internal.Pooling.CombinedAsyncContinuationPool|1|1|0|0|0|
                NLog.Internal.Pooling.GenericPool`1[[NLog.Common.Counter, NLog, Version=0.0.0.0, Culture=neutral, PublicKeyToken=5120e14c03d0593c]]|1|0|0|0|0|
                NLog.Internal.Pooling.GenericPool`1[[NLog.Common.ContinueWhenAll, NLog, Version=0.0.0.0, Culture=neutral, PublicKeyToken=5120e14c03d0593c]]|1|0|0|0|0|
                NLog.Internal.Pooling.ExceptionHandlerPool|1|1|0|0|0|
                NLog.Internal.Pooling.AsyncLogEventInfoArrayPool|100|0|0|0|0|
                NLog.Internal.Pooling.MemoryStreamPool|1|0|0|0|0|

304NotModified:
Indeed we try to reset those before every test (xunit creates a new instance of the test class for every test method/run)

bjornbouetsmith:
And it seems like I fixed _ALL_ my mono test problems, because I added a Sync call in NLogTest base, so that the file system is flushed. Its seems like linux is caching too agressively for our tests speed. So by forcing the files to be written, the tests pass :-) Which means my code was working all along as I expected :-)

bjornbouetsmith:
I will go through everything tomorrow and squash so there are fewer commits :-)

304NotModified:
https://tortoisegit.org/docs/tortoisegit/tgit-dug-rebase.html

After rebase, no `git pull`! (but `git push - f`) 

Backup is recommended :)

bjornbouetsmith:
Squash wont work, since I have merged from the NLog base branch, and have my commits mixed between others.

So I think Rebase is the only option - which requires that I resolve some conflicts it seems like.

bjornbouetsmith:
What would you think about me opening another PR, and just copy my working base across to that and close this PR. Then I don't have to think about all the merge conflicts I am processing right now, and it will be easier for me to handle any merge conflicts, since it will only be at most one conflict per file?

304NotModified:
That won't work, as there are also merged commits of others in your working copy?

bjornbouetsmith:
Yes but I wold revert everything but my own changes

bjornbouetsmith:
I am fixing my merge, when everything is fixed I will squash once more :-)

bjornbouetsmith:
okay, squashed and ready for review :-)

304NotModified:
cool!

bjornbouetsmith:
But please note, there are still appenders and layout rendererers that could benefit from pooling, I just had to put my efforts some place, but the pattern should be clear for others to continue with.

bjornbouetsmith:
So how is the review going to be done? By you, or do I have to do more now?

304NotModified:
I will do the review, but first I have to find a good way to review large PRs like this.

After that, I need some tests. There are some test programs in some lost branches (from other PRs), if I found them, I will make a new Gihub repository to move it to them. 

Last but not least, next week I will be on holiday for 2 weeks (back 27 jun). I probably won't merge it before my holiday.

304NotModified:
I did a review :)

I did check all the files, expect the test classes, in short

My findings

## Unneeded changes:
There are a lot of files unneedely changed (not related to pooling). This PR is really big (lot of changed files), so it would be nice if we can make it smaller. There are two options:
1) Send the good changes in another PR, I will merge them soon with the master. Some changes are really nice changes.
2) remove them by checking out the file in git from nlog/master. Eg. `git checkout NLog/master -- path/to/file`

These files should not be changed at least in this PR:
- MessageQueueTarget
- LogEventInfoBuffer
- FileAppenderCache
- LevelLayoutRenderer
- ThreadIdLayoutRenderer
- ThreadNameLayoutRenderer
- OnExceptionLayoutRendererWrapper
- PaddingLayoutRendererWrapper
- ReplaceNewLinesLayoutRendererWrapper
- Rot13LayoutRendererWrapper
- TrimWhiteSpaceLayoutRendererWrapper
- UrlEncodeLayoutRendererWrapper
- WhenEmptyLayoutRendererWrapper
- WrapLineLayoutRendererWrapper
- EventLogTarget
- BufferingTargetWrapper
- LastLogEventListTarget

## Usings

Please add new `using`s to the same section as the other `using`s

## Questions 
I have the following questions:

1. Why are we passing configuration to aysncHelpers?
1. Why are the explicit .toString and casts added?
1. I think it would be nice if the Pools implemtations were in separate namespace/ folder.
1. why is CallSiteLayoutRenderer/CallSiteLineNumberLayoutRenderer not `ThreadAgnostic`
1. LayoutRenderer.Render: call to `this.LoggingConfiguration.PoolingEnabled()` looks expensive (as it is a method) - should it be a property?
1. Why we need Layout.RenderBuilder / RenderTo. Why not in IRenderable?
1. `Target.Write(ArraySegment<AsyncLogEventInfo> logEvents)`` -> could this be `Target.Write(IEnumerable<AsyncLogEventInfo> logEvents)`?
1. CombinedException: is this not a code smell?
1. NLogXmlElement.GetOptionalAttribute: move using, removed uncommented code, write exception to internalLogger (first arg)
1. LongDateLayoutRenderer: what are those magic numbers?
1. JsonEncodeLayoutRendererWrapper.render why looping back?

can we move the following code to a new base class? (so we need to change less files)   
                                  
```c#                             
 //from various targets           
  protected override void Write(AsyncLogEventInfo[] logEvents)
  {
      this.Write(new ArraySegment<AsyncLogEventInfo>(logEvents));
  }
 
```                                              
 
## Remarks     
1. AsyncContinuationListPool: `100` should be const or reset method
1. XmlLoggingConfiguration TODO: move it to Initialize()                                                   
1. logFactory.GlobalThreshold: good change, remove the `?` in your comments :)                                                        
       
## Docs
Some changes are really big. Also there are a lot of classes introduced. We need far more docs otherwise we cannot maintain this. Mostly, we need docs WHY we are doing things, but also I'm curious how some things get so complicated (e.g. UppercaseLayoutRendererWrapper)
All new classes needs docs, also internal ones. 

I would also be nice if we have a list of what is optimized (targets, layout renderes etc.)

NB: Change in DebugTarget should be documented on the wiki

bjornbouetsmith:
Okay first comments.

# Unneeded changes
Reverted the unrelated changes.

Regarding usings, I thought it was hard to find a pattern. Some were added inside the namespace, others outside. Is there a policy? Otherwise I think we should just let people decide.

# Answers to your questions

1. The configuration is needed when we need to grab a pooled object, unfortunately most methods end up in PreventMultipleCalls that needs access to the SingleCallContinuationPool.
2. Explicit ToString() are needed in format strings. Otherwise .NET will box the primitives in an object, which gets allocated on the heap and needs to be garbage collected. When explicitily calling ToString() on the primitives, then we save that allocation.
3. No problem - I will move them
4. As far as I understand. The ThreadAgnosticAttribute is for layouts that is not dependent on the calling thread. This is used to either "Pre Render" the log message or wait until after the AsyncTargetWrapper is dequeing, in the case of async tagets. Since CallSite and CallSiteLineNumber depends on the calling location, which is only available in the stack trace, then its by definition not thread agnostic, since the line number will be changed if the layout is rendered on another thread, since that thread will be starting in AsyncTargetWrapper.
5. this.LoggingConfiguration.PoolingEnabled() is a static extension method, its no more expensive than calling a property. A Property is just syntactical sugar for a method. All properties gets compiled into get_Property and set_Property, so its no different. And an extension method cannot be a property unfortunately
6. I will try to move them to Irenderable - I was not aware that there was an interface. Makes sense to put them there if possible.
7. Unfortunately it cannot be IEnumerable<LogEventInfo>, since we need the number of "filled out" log events, when requesting an array of AsyncLogEventInfo from the pool, we get a fixed size array back, so when the target wraps the AsyncLogEventInfo's it gets called with, we need to know how many there was original, and the only was is to wrap it in an ArraySegment, or somehow in another way pass the count and offset
8. CombinedException is just an easy wrapper around AsyncHelpers.GetCombinedException, I thought it was better to just add a new exception that would defer the rendering and string manipulation until someone actually called .ToString() which will probably never happen unless users are running with setting that tells nlog to log errors or throw errors.
9. Will do
10. What magic numbers? The ones in the Append4DigitsZeroPadded or in the actual Append method? Those in Append4Digits are the old ones, which basically just chops up the digits into their respective bucket. The others are just the hour, minute, second, millisecond max value before change of time unit.
11. Json - I will revert the changes, they are not that important anyway, and probably have to be reimplemented properly.
12. I will try to see if we can somehow not need that change, but it might be hard.

# Remarks

1-3 OK will do :-)

# Docs
I will add documentation to all classes and methods I have introduced.
I will also compile a list of targets/layout renderers optimised.
How do I change stuff on the wiki?

bjornbouetsmith:
Can you queue the Travis build again, i'm pretty sure its a x-unit problem and not a test problem.

304NotModified:
Will try to test/include this for NLog 4.4, but it depends of the performance results and the many conflicts on .NET Core branch (coreCLR)

bjornbouetsmith:
The conflicts with the .NET Core branch - is that something that has to be fixed before releasing, or something that can be fixed after it has been merged into the main branch?

I'm pretty sure the performance results will be good, since I have testest performance throroughly.

304NotModified:
> The conflicts with the .NET Core branch - is that something that has to be fixed before releasing, or something that can be fixed after it has been merged into the main branch?

Both is possible, as we could create a new branch an net PR to the coreCLR branch. But I guess the best way is to do the latter (merge first, fix later)

> I'm pretty sure the performance results will be good, since I have testest performance throroughly.

Good too know! So you're compiling your own NLog assembly?

304NotModified:
Superseded by https://github.com/NLog/NLog/pull/1663

PR #1400, opened 2016-04-25
Added WrapLineLayoutRendererWrapper

mathieubrun:
Implementation following discussion on http://stackoverflow.com/questions/36788754/how-can-i-limit-the-length-of-a-line-in-nlog/36789394

### DIFF
--- a/src/NLog/LayoutRenderers/Wrappers/WrapLineLayoutRendererWrapper.cs
+++ b/src/NLog/LayoutRenderers/Wrappers/WrapLineLayoutRendererWrapper.cs
@@ -72,24 +72,33 @@ namespace NLog.LayoutRenderers.Wrappers
         /// <returns>Post-processed text.</returns>
         protected override string Transform(string text)
         {
-            var chunkLength = WrapLine;
-            var textLength = text.Length;
+            if (WrapLine <= 0)
+            {
+                throw new ArgumentException("WrapLine must be a positive integer");
+            }
 
-            var sb = new StringBuilder(textLength);
+            var chunkLength = WrapLine;
+
+            var result = "";
 
             // based on : http://stackoverflow.com/questions/36788754/how-can-i-limit-the-length-of-a-line-in-nlog/36789394
             // and : http://stackoverflow.com/questions/1450774/splitting-a-string-into-chunks-of-a-certain-size/8944374#8944374 
             for (int pos = 0; pos < text.Length; pos += chunkLength)
             {
-                if (chunkLength + pos > textLength)
-                    chunkLength = textLength - pos;
+                if (chunkLength + pos > text.Length)
+                {
+                    chunkLength = text.Length - pos;
+                }
 
-                sb.Append(text.Substring(pos, chunkLength) + Environment.NewLine);
+                result += text.Substring(pos, chunkLength);
+
+                if (chunkLength + pos < text.Length)
+                {
+                    result += Environment.NewLine;
+                }
             }
 
-            sb.Remove(sb.Length - Environment.NewLine.Length, Environment.NewLine.Length);
-
-            return sb.ToString();
+            return result;
         }
     }
 }
--- a/src/NLog/LayoutRenderers/Wrappers/WrapLineLayoutRendererWrapper.cs
+++ b/src/NLog/LayoutRenderers/Wrappers/WrapLineLayoutRendererWrapper.cs
@@ -33,10 +33,11 @@
 
 namespace NLog.LayoutRenderers.Wrappers
 {
-    using NLog.Config;
     using System;
     using System.ComponentModel;
-    using System.Collections.Generic;
+    using System.Text;
+
+    using NLog.Config;
 
     /// <summary>
     /// Replaces newline characters from the result of another layout renderer with spaces.
@@ -71,27 +72,24 @@ namespace NLog.LayoutRenderers.Wrappers
         /// <returns>Post-processed text.</returns>
         protected override string Transform(string text)
         {
-            return string.Join(Environment.NewLine, MakeChunks(text, WrapLine));
-        }
+            var chunkLength = WrapLine;
+            var textLength = text.Length;
 
-        /// <summary>
-        /// Splits a string into chunks
-        /// </summary>
-        /// <param name="toSplit">The string to split</param>
-        /// <param name="chunkLength">The chunk length</param>
-        /// <returns>Collection of strings</returns>
-        private static IEnumerable<string> MakeChunks(string toSplit, int chunkLength)
-        {
-            if (String.IsNullOrEmpty(toSplit)) throw new ArgumentException();
-            if (chunkLength < 1) throw new ArgumentException();
+            var sb = new StringBuilder(textLength);
 
-            for (int i = 0; i < toSplit.Length; i += chunkLength)
+            // based on : http://stackoverflow.com/questions/36788754/how-can-i-limit-the-length-of-a-line-in-nlog/36789394
+            // and : http://stackoverflow.com/questions/1450774/splitting-a-string-into-chunks-of-a-certain-size/8944374#8944374 
+            for (int pos = 0; pos < text.Length; pos += chunkLength)
             {
-                if (chunkLength + i > toSplit.Length)
-                    chunkLength = toSplit.Length - i;
+                if (chunkLength + pos > textLength)
+                    chunkLength = textLength - pos;
 
-                yield return toSplit.Substring(i, chunkLength);
+                sb.Append(text.Substring(pos, chunkLength) + Environment.NewLine);
             }
+
+            sb.Remove(sb.Length - Environment.NewLine.Length, Environment.NewLine.Length);
+
+            return sb.ToString();
         }
     }
 }
--- a/src/NLog/LayoutRenderers/Wrappers/WrapLineLayoutRendererWrapper.cs
+++ b/src/NLog/LayoutRenderers/Wrappers/WrapLineLayoutRendererWrapper.cs
@@ -79,7 +79,10 @@ namespace NLog.LayoutRenderers.Wrappers
 
             var chunkLength = WrapLine;
 
-            var result = "";
+            if (text.Length <= chunkLength) return text;
+
+            // preallocate correct number of chars
+            var result = new StringBuilder(text.Length + (text.Length / chunkLength) * Environment.NewLine.Length);
 
             // based on : http://stackoverflow.com/questions/36788754/how-can-i-limit-the-length-of-a-line-in-nlog/36789394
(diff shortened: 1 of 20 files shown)


304NotModified:
Nice! Thanks!

304NotModified:
needs some fix for .net35 

> LayoutRenderers\Wrappers\WrapLineLayoutRendererWrapper.cs(74,53): error CS1503: Argument 2: cannot convert from 'System.Collections.Generic.IEnumerable<string>' to 'string[]' [C:\projects\nlog\src\NLog\NLog.netfx35.csproj

304NotModified:
Fix PR: https://github.com/mathieubrun/NLog/pull/1

mathieubrun:
@304NotModified I'll look into it. I also want to change a bit the MakeChunks method

304NotModified:
See the PR is send to you

mathieubrun:
Yes, thanks ! But I changed implementation to a simple StringBuilder. No need to split and join, much simplier :)

304NotModified:
That's indeed also a nice solution, if no unneeded stringbuilder are created :)

mathieubrun:
Done, waiting for AppVeyor to finish !

304NotModified:
A stringbuilder approach is indeed nicer, but also a bit more work too do it right.

src/NLog/LayoutRenderers/Wrappers/WrapLineLayoutRendererWrapper.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// Post-processes the rendered message. 
+        /// </summary>
+        /// <param name="text">The text to be post-processed.</param>
+        /// <returns>Post-processed text.</returns>
+        protected override string Transform(string text)
+        {
+            var chunkLength = WrapLine;
+            var textLength = text.Length;
+
+            var sb = new StringBuilder(textLength);
```
304NotModified: building a stringbuilder isn't always needed

src/NLog/LayoutRenderers/Wrappers/WrapLineLayoutRendererWrapper.cs
```diff
@@ -0,0 +1,95 @@
...
+            // and : http://stackoverflow.com/questions/1450774/splitting-a-string-into-chunks-of-a-certain-size/8944374#8944374 
+            for (int pos = 0; pos < text.Length; pos += chunkLength)
+            {
+                if (chunkLength + pos > textLength)
+                    chunkLength = textLength - pos;
+
+                sb.Append(text.Substring(pos, chunkLength) + Environment.NewLine);
+            }
+
+            sb.Remove(sb.Length - Environment.NewLine.Length, Environment.NewLine.Length);
```
304NotModified: this could be done better. We already checking thing is the `for` loop, so don't add ` Environment.NewLine` and then remove it later.

src/NLog/LayoutRenderers/Wrappers/WrapLineLayoutRendererWrapper.cs
```diff
@@ -0,0 +1,95 @@
...
+        public int WrapLine { get; set; }
+
+        /// <summary>
+        /// Post-processes the rendered message. 
+        /// </summary>
+        /// <param name="text">The text to be post-processed.</param>
+        /// <returns>Post-processed text.</returns>
+        protected override string Transform(string text)
+        {
+            var chunkLength = WrapLine;
```
304NotModified: I think we have to check if `WrapLine > 0`

tests/NLog.UnitTests/LayoutRenderers/Wrappers/WrapLineTests.cs
```diff
@@ -0,0 +1,115 @@
...
+{
+    using NLog;
+    using NLog.Layouts;
+    using NLog.Targets;
+    using Xunit;
+
+    public class WrapLineTests : NLogTestBase
+    {
+        [Fact]
+        public void WrapLineAtPositionInsideTextTest()
```
304NotModified: Nice tests, but we need wrapline=0 and wrapline=-1 (negative) tests IMO

src/NLog/LayoutRenderers/Wrappers/WrapLineLayoutRendererWrapper.cs
mathieubrun: indeed. replaced by a string

src/NLog/LayoutRenderers/Wrappers/WrapLineLayoutRendererWrapper.cs
mathieubrun: done.

src/NLog/LayoutRenderers/Wrappers/WrapLineLayoutRendererWrapper.cs
mathieubrun: Done, threw ArgumentException

tests/NLog.UnitTests/LayoutRenderers/Wrappers/WrapLineTests.cs
mathieubrun: Tests added. I did not find a simple way (other than providing InternalLogger with a StringWriter then test on it (cf InternalLogger). For the moment asserted on output == ""

304NotModified:
Why is the string builder totally removed? I think we had a miscommunication!

The `new stringbuilder` wasn't needed if the line didn't need a wrap (e.g. length of 70). But in other cases a stringbuilder is more preferred than string concats

mathieubrun:
Ahah sorry, I understood to remove it fully :) I put it back

src/NLog/LayoutRenderers/Wrappers/WrapLineLayoutRendererWrapper.cs
```diff
@@ -0,0 +1,107 @@
...
+            {
+                throw new ArgumentException("WrapLine must be a positive integer");
+            }
+
+            var chunkLength = WrapLine;
+
+            if (text.Length <= chunkLength) return text;
+
+            // preallocate correct number of chars
+            var result = new StringBuilder(text.Length + (text.Length / chunkLength) * Environment.NewLine.Length);
```
304NotModified: Good one!

tests/NLog.UnitTests/LayoutRenderers/Wrappers/WrapLineTests.cs
```diff
@@ -0,0 +1,164 @@
...
+    using NLog.Layouts;
+    using NLog.Targets;
+    using Xunit;
+
+    public class WrapLineTests : NLogTestBase
+    {
+        [Fact]
+        public void WrapLineAtPositionOnceTest()
+        {
+            MappedDiagnosticsContext.Clear();
```
304NotModified: Why do we need this?

src/NLog/LayoutRenderers/Wrappers/WrapLineLayoutRendererWrapper.cs
```diff
@@ -0,0 +1,107 @@
...
+        /// <summary>
+        /// Post-processes the rendered message. 
+        /// </summary>
+        /// <param name="text">The text to be post-processed.</param>
+        /// <returns>Post-processed text.</returns>
+        protected override string Transform(string text)
+        {
+            if (WrapLine <= 0)
+            {
+                throw new ArgumentException("WrapLine must be a positive integer");
```
304NotModified: I'm doubting if this is the best way. We can also move the check to the setter. Wat do you think?

304NotModified:
Looks really good! Almost done!

tests/NLog.UnitTests/LayoutRenderers/Wrappers/WrapLineTests.cs
mathieubrun: I saw it in a lot of other tests. I was thinking it is required. Works without it as well.

304NotModified: OK please remove them.  Less is more. :)

src/NLog/LayoutRenderers/Wrappers/WrapLineLayoutRendererWrapper.cs
mathieubrun: Putting it in the setter will throw an exception :
`NLog.NLogConfigurationException : Error when setting property 'WrapLine' on Layout Renderer: ${wrapline}`
-> will fail faster, on Layout creation, with an exception

Leaving it here will log into InternalLogger. I had a look at other WrappingLayoutRenderers, did not find similar case. 
-> will fail later, more silently, producing an empty string, and logging a warning in internal logger.

An alternative could be something like this:

```
protected override string Transform(string text)
{
    if (WrapLine <= 0)
    {
        InternalLogger.Warn("WrapLine must be a positive integer");
        return text;
    }
```

tests/NLog.UnitTests/LayoutRenderers/Wrappers/WrapLineTests.cs
mathieubrun: :+1:

src/NLog/LayoutRenderers/Wrappers/WrapLineLayoutRendererWrapper.cs
304NotModified: I think the last proposal is the best :)

mathieubrun: Ok !

304NotModified:
I think I can merge this after we added some docs to the wiki?

mathieubrun:
I can write the new page, then you can link it on the https://github.com/NLog/NLog/wiki/Layout-Renderers page ?

304NotModified:
that would be great!

mathieubrun:
There you go : https://github.com/NLog/NLog/wiki/WrapLine-layout-renderer

I added tests for 2 other syntaxes : 
```
${wrapline:${message}}
${wrapline:Inner=${message}:WrapLine=80}

```

mathieubrun:
I have no idea WHY commit https://github.com/NLog/NLog/pull/1400/commits/80f9e7d5ba0b621d1491351543b2f7f6033da4f9 went without the method names...

304NotModified:
lol thats wierd indeed.

304NotModified:
Thanks for this great submit!

mathieubrun:
You're welcome ! I'll look if there are other things I can help with !

304NotModified:
This has been released with NLog 4.3.4 :)

Issue #1407, opened 2016-04-27
Logging doesn't work if mail target added to NLog configuration for web application


dimrun:
**Type**: Bug

**NLog version**: 4.3.2

**Platform**: .Net 3.5 / .Net 4.0 / .Net 4.5 / Mono 4 / Silverlight 4 / Silverlight 5

**Current NLog config** (xml or C#, if relevant):
nothing special, just mail target added to web.config

In case of a BUG:

- What is the current result?
Exception occurs:
```
---> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.ArgumentException: exePath must be specified when not running inside a stand alone exe.
  at System.Configuration.ConfigurationManager.OpenExeConfigurationImpl(ConfigurationFileMap fileMap, Boolean isMachine, ConfigurationUserLevel userLevel, String exePath, Boolean preLoad)
  at System.Configuration.ConfigurationManager.OpenExeConfiguration(ConfigurationUserLevel userLevel)
  at NLog.Targets.MailTarget.get_MailSettings()
```

- What is the expected result?
No errors

- Stack trace:
```
---> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.ArgumentException: exePath must be specified when not running inside a stand alone exe.
  at System.Configuration.ConfigurationManager.OpenExeConfigurationImpl(ConfigurationFileMap fileMap, Boolean isMachine, ConfigurationUserLevel userLevel, String exePath, Boolean preLoad)
  at System.Configuration.ConfigurationManager.OpenExeConfiguration(ConfigurationUserLevel userLevel)
  at NLog.Targets.MailTarget.get_MailSettings()
  --- End of inner exception stack trace ---
  at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
  at System.Reflection.RuntimeMethodInfo.UnsafeInvokeInternal(Object obj, Object[] parameters, Object[] arguments)
  at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
  at System.Reflection.RuntimePropertyInfo.GetValue(Object obj, Object[] index)
  at NLog.Internal.ObjectGraphScanner.ScanProperties[T](List`1 result, Object o, Int32 level, HashSet`1 visitedObjects)
[...]

PR #1411, opened 2016-04-28
MailTarget: fix "From" errors (bug introduced in NLog 4.3.2)

304NotModified:
- Removed OpenExeConfiguration, which could lead to errors. (bug introduced in NLog 4.3.2)
- try-catch for reading SmtpSection.
- fixed ordening (Targetsettings takes preference over app.config)
- MailSettings should not be public (only in 4.3.2)
- Improve checking of "From" and overriding
- Improve test cases.

fixes https://github.com/NLog/NLog/issues/1407


304NotModified:
:( Codecov still buggy

PR #1417, opened 2016-04-30
FileTarget prevent stackoverflow after setting FileName property on init

304NotModified:
Execute  RefreshFileArchive() + RefreshArchiveFilePatternToWatch() only after first initialize (because of potentional stackoverflow)

To prevent cases like this (after setting FileTarget.FileName)

image

### DIFF
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -224,21 +224,31 @@ namespace NLog.Targets
             get { return fileName; }
             set
             {
-                var simpleLayout = value as SimpleLayout;
-                if (simpleLayout != null && simpleLayout.IsFixedText)
-                {
-                    cachedCleanedFileNamed = CleanupInvalidFileNameChars(simpleLayout.FixedText);
-                }
-                else
-                {
-                    //clear cache
-                    cachedCleanedFileNamed = null;
-                }
-
+              
                 fileName = value;
 
-                RefreshFileArchive();
-                RefreshArchiveFilePatternToWatch();
+                if (IsInitialized)
+                {
+                    SetCachedCleanedFileNamed(value);
+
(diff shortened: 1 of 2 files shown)


bhaeussermann:
I'm happy with the changes, but what happened with Codecov??

304NotModified:
They moved to v4. Which made a lot of stuff unstable.  Im in discussion with them, but if it isn't fixed this week, I will remove the codecov integration.

src/NLog/Targets/FileTarget.cs
```diff
@@ -237,8 +237,11 @@ public Layout FileName
 
                 fileName = value;
 
-                RefreshFileArchive();
-                RefreshArchiveFilePatternToWatch();
+                if (IsInitialized)
+                {
+                    RefreshFileArchive();
+                    RefreshArchiveFilePatternToWatch();
```
bhaeussermann: These two methods are called from a number of other properties as well. Perhaps we should move the `IsInitialized` check into the `RefreshFileArchive()` and `RefreshArchiveFilePatternToWatch()` methods.

304NotModified: I think it's more clear (for the other properties), to keep the check in the setter, but I will double check the other properties.

src/NLog/Targets/FileTarget.cs
```diff
@@ -2015,6 +2039,7 @@ private byte[] GetLayoutBytes(Layout layout)
         /// </summary>
         /// <param name="fileName">The original file name which might contain invalid characters.</param>
         /// <returns>The cleaned up file name without any invalid characters.</returns>
+        /// <remarks>This method should not set a lock on the config.</remarks>
```
304NotModified: remove

PR #1419, opened 2016-05-01
Fix default values of ambient properties

### LINKED ISSUE #1372: ${level:uppercase} won't work
**Type** (choose one):

- Bug

**NLog version**: (4.2.1)

**Platform**: .Net 4.5

```xml
${level:uppercase}
```

In case of a BUG:

- What is the current result? no UPPERCASE
- What is the expected result? UPPERCASE
- Are there any work arrounds? yes: `${level:uppercase=true}`

mathieubrun:
Proposal for implementation. Fixes #1372

### DIFF
--- a/src/NLog/Layouts/LayoutParser.cs
+++ b/src/NLog/Layouts/LayoutParser.cs
@@ -431,6 +431,20 @@ namespace NLog.Layouts
                     else
                     {
                         InternalLogger.Warn("{0} has no default property", lr.GetType().FullName);
+
+                        Type wrapperType;
+
+                        if (configurationItemFactory.AmbientProperties.TryGetDefinition(parameterName, out wrapperType))
+                        {
+                            LayoutRenderer wrapperRenderer;
+
+                            if (!wrappers.TryGetValue(wrapperType, out wrapperRenderer))
+                            {
+                                wrapperRenderer = configurationItemFactory.AmbientProperties.CreateInstance(parameterName);
+                                wrappers[wrapperType] = wrapperRenderer;
+                                orderedWrappers.Add(wrapperRenderer);
+                            }
+                        }
                     }
                 }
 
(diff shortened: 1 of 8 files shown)


304NotModified:
Thanks! Will look at this in depth this week.

304NotModified:
Thanks for the PR!

Im doubting if this isn't a bug in the docs. Because we have now an ambiguity with the layout renderer has a default parameters, e.g. ${exception}.

- `${exception:uppercase=true}` => wraps uppercase arround exception
- `${message:uppercase}` => wraps uppercase arround message
- `${exception:uppercase}` => set Format (default parameter) of exception to "uppercase" :-1:

mathieubrun:
I think a failing test for 

- `${exception:uppercase}` => set Format (default parameter) of exception to "uppercase" :-1:

could be a starting point.

The layout parser could detect if the ${exception:**uppercase**} is an ambient property name, and not affect the default parameter of ${**exception**:uppercase}.

I just don't know the codebase enough to say if this is a good idea / would have side effects.

304NotModified:
> I just don't know the codebase enough to say if this is a good idea / would have side effects.

Well it's a breaking change ;) I am not sure if this is a good idea, adding a property or ambient property to the code base could have large effects.

mathieubrun:
Fixing the docs is less risky then

Le lundi 2 mai 2016, Julian Verdurmen <notifications@github.com> a écrit :

> I just don't know the codebase enough to say if this is a good idea /
> would have side effects.
>
> Well it's a breaking change ;) I am not sure if this is a good idea,
> adding a property or ambient property to the code base could have large
> effects.
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/NLog/NLog/pull/1419#issuecomment-216370729>
>


-- 
Cordialement,
Mathieu BRUN

304NotModified:
Will rethink about this, but probably that's the best.

304NotModified:
The docs are updated. We will not change this parse behavior, see https://github.com/NLog/NLog/issues/1372

Issue #1420, opened 2016-05-02
Zip archives with .NET 3.5


PeterHric:
Hi! Thanks for reporting this feature/bug/question! 

Please keep / fill in the relevant info from this template so that we can help you as best as possible.

**Type** 
- Feature request
  Support for zipping archive files with .NET 3.5 would greatly help :) We need it since we have to be backward compatible due to many target environments which run on .NET 3.5
- Question
  In on the web site: 
  https://github.com/nlog/NLog/wiki/File-target  
  there is statement:
  "Compress won't work due to .Net restrictions. You can use enableArchiveFileCompression."
  And later below:
  "enableArchiveFileCompression - Indicates whether to compress the archive files into the zip files. Boolean Default: False > Supported in: > \* NLog v4.0 for .NET 4.5"
  My Q:  Is it intrinsically given (which I rather doubt), or just that NLog compiled with .NET 3.5 simply does not have that part of code ?

304NotModified:
Hi,

We use 'System.IO.Compression.ZipFile`, which is introduced in .Net 4.5.

https://msdn.microsoft.com/en-us/library/system.io.compression.ziparchive(v=vs.110).aspx

PeterHric:
Dear all,

Thank you for the good answer.
Do you think there would be any chance to use DotNetLib
<http://stackoverflow.com/questions/593026/is-there-a-built-in-zip-library-in-net-3-5>
for earlier .NET releases and incorporate it to NLog.dll somehow, so that
the option  enableArchiveFileCompression="Boolean"   would work too ?

Even if third party would get involved :)

Regards,

Peter

2016-05-02 14:11 GMT+02:00 Julian Verdurmen <notifications@github.com>:

304NotModified:
I will check if I can include this in a separate package.

PeterHric:
Thanx Julian,

That is highly esteemed ! :)

BR,

Peter

2016-05-02 18:57 GMT+02:00 Julian Verdurmen <notifications@github.com>:

PeterHric:
Hi again,

A small suggestion:
 Using System.IO.Compression.GZipStream, and/or  System.IO.Compression.DeflateStream  for archive files compression with .NET 3.5  would be a good option in my opinion.

See here for more data:
https://msdn.microsoft.com/en-us/library/ms404280(v=vs.90).aspx

304NotModified:
@PeterHric NLog 4.3.4 is released with the option to inject your own file compressor.
[...]

PR #1423, opened 2016-05-02
Injection of zip-compressor for fileTarget

AndreGleichner:
See also Issue #1420.
I need to get archive files zipped on .Net4.0 and thus have added a IFileCompressor interface to LogManager, so that NLog users may call NLog.LogManager.SetFileCompressor() to inject there own preferred file compressor like DotNetZip.


304NotModified:
Sounds good! Will check this in depth this week

304NotModified:
Please add the file headers to the new files so appveyor won't complain

src/NLog/Common/DefaultFileCompressor.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+#if NET4_5
+using System.IO;
+using System.IO.Compression;
+#endif
+
+    /// <summary>
+    /// Uses .Net4.5 ZipArchive
+    /// </summary>
+    public class DefaultFileCompressor : IFileCompressor
```
304NotModified: - I think is better to only include this class in .Net 45. The bool return is then not needed.
- Please use the `<see` tag for `ZipArchive`

src/NLog/Common/IFileCompressor.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace NLog.Common
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public interface IFileCompressor
```
304NotModified: looks good, but need soms docs (also parameters below)

src/NLog/LogManager.cs
```diff
@@ -435,5 +435,42 @@ private static void TurnOffLogging(object sender, EventArgs args)
             Configuration = null;
             InternalLogger.Info("Logger has been shut down.");
         }
+
+        static IFileCompressor _fileCompressor;
```
304NotModified: please move those changes to the `FileTarget` (LogManager should not have knowledge of the targets)

src/NLog/LogManager.cs
```diff
@@ -435,5 +435,42 @@ private static void TurnOffLogging(object sender, EventArgs args)
...
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public static bool HasFileCompressor
+        {
+            get
+            {
+#if NET4_5
+                return true;
```
304NotModified: this special case isn't needed

304NotModified:
reviewed already :)

It's in a good direction, and the following changes makes in perfect IMO:

- The `IFileCompressor` is always used (all platforms)
- The Compressor property is of the FileTarget (see code notes) (instead of LogManager)
- The Compressor property is `null` for non-.Net45 platforms. 
- Some small changes, see code notes.

src/NLog/LogManager.cs
```diff
@@ -435,5 +435,42 @@ private static void TurnOffLogging(object sender, EventArgs args)
...
+        public static void SetFileCompressor(IFileCompressor fileCompressor)
+        {
+            _fileCompressor = fileCompressor;
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <returns></returns>
+        public static IFileCompressor GetFileCompressor()
```
304NotModified: change to property instead of 2 methods

AndreGleichner:
Great to see you already had a chance to check my proposed code. Initially my intention was that IFileCompressor could be used FileTarget independently. After moving all stuff to FileTarget according to your comments, it now seems to be a pretty straightforward to use interface. To be honest i til now only hacked this stuff together w/o any actual test and now have my doubts that i actually can set that FileTarget.DefaultCompressor static property early enough, before NLog actually creates the FileTarget instances. So i guess DefaultCompressor setter would need looping over all FileTarget instances.

304NotModified:
> have my doubts that i actually can set that FileTarget.DefaultCompressor static property early enough, before NLog actually creates the FileTarget instances.

It's only important to set it before writing any logevent?

AndreGleichner:
Ok, i guess i just have to set DefaultCompressor before doing any actual logging...i'm a bit unsure how the NLog bootstrap is done.
Could you point me to the right direction of how to fix that AppVeyor thing?

304NotModified:
> Could you point me to the right direction of how to fix that AppVeyor thing?

see line 164, 171 etc: https://ci.appveyor.com/project/nlog/nlog/build/4.0.2397

src/NLog/Targets/FileTarget.cs
```diff
@@ -562,27 +570,35 @@ public int MaxArchiveFiles
         /// <docgen category='Archival Options' order='10' />
         public ArchiveNumberingMode ArchiveNumbering { get; set; }
 
-#if NET4_5
+        /// <summary>
+        /// Per instance file compressor to be used to compress log files during archiving.
+        /// Defaults to <see cref="DefaultCompressor"/>.
+        /// </summary>
+        public IFileCompressor Compressor { get; set; }
```
304NotModified: proposal: rename to FileCompressor

src/NLog/Targets/FileTarget.cs
```diff
@@ -201,8 +200,17 @@ public FileTarget()
             this.previousFileNames = new Queue<string>(this.maxLogFilenames);
             this.fileAppenderCache = FileAppenderCache.Empty;
             this.CleanupFileName = true;
+            
+            this.Compressor = DefaultCompressor;
```
304NotModified: I'm a bit confused by the Compressor  and DefaultCompressor properties. Are both needed? Is this because of the `internal` implementation?

src/NLog/Targets/IFileCompressor.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// or <see cref="FileTarget.DefaultCompressor"/> for all instances.
+    /// </summary>
+    public interface IFileCompressor
+    {
+        /// <summary>
+        /// Create archiveFileName by compressing fileName.
+        /// </summary>
+        /// <param name="fileName">Absolute path to the log file to compress/zip.</param>
+        /// <param name="archiveFileName">Absolute path to the archive zip file to create.</param>
+        void Compress(string fileName, string archiveFileName);
```
304NotModified: proposal: CompressFile

304NotModified:
\0/ it builds. 

Some last things about naming and I was a bit confused with two properties, see notes.

src/NLog/Targets/FileTarget.cs
AndreGleichner: That would have been my personal preference, but you proposed "Compressor" as the property name above ;)

src/NLog/Targets/FileTarget.cs
AndreGleichner: Compressor is per FileTarget Instance and DefaultCompressor it a static property. Background is that you may wanna set a standard compression implementation for all FileTargets but may choose to set onother for a single instance. Maybe a bit too flexible ;)

AndreGleichner:
I'll remove the per FileTarget Instance Compressor property. Just leaving the static property which I'll rename to FileCompressor. And I'll rename IFileCompressor.Compress() to IFileCompressor.CompressFile().

304NotModified:
Cool! Think we needs some unit tests and then we are ready to go

AndreGleichner:
I'll see if I can invent some unit tests soon.
It just came to my mind to also support non-zip archive formats like 7z...how about extending IFileCompressor by an ArchiveExtension readonly property to return e.g. ".zip", ".7z", etc?
There're just a few places where zip is hardcoded.

304NotModified:
Good idea about supporting .7z etc. But I think already fits in the current interface? AFAIK the .zip is only used as fallback, so a user could inject how FileCompressor and configure something like "archive.7z"

AndreGleichner:
Right, in FileTarget its just a fallback. Its actually just the unit tests hardcoding .zip or .txt as archive extension. But that can stay as is...so you're right no need to have such ArchiveExtension property.

PeterHric:
Hi all,
I appreciate this effort. Would Like to ask if and where is the new .dll available with this patch.
If there was a simple example how to inject FileCompressor and what Interface it must comply to, that would be even fantastisk !

304NotModified:
We needs some unit tests, and then it's ready for merge. After merge I expect the release within a few days. 

I you could help us with unit tests, then I can merge (and release) it earlier.

PeterHric:
OK great, How do I help you ? :)

304NotModified:
1. checkout this PR locally
2. create a new branch locally from 1.
3. add unit tests 
4. Create PR to "AndreGleichner:master"

AndreGleichner:
I'd do such UnitTests...just a bit bussy...my main problem just doing a simple UnitTest, ist that I wasn't easily able figuring out whether xUnit is running tests in parallel. If it would I can't just set FileTarget.FileCompressor to another IFileCompressor implementation based on e.g. DotNetZip, because that may crash the other tests also doing file with zip archiving.
That was one of the reasons why my original approach had a per instance compressor property.

304NotModified:
> my main problem just doing a simple UnitTest, ist that I wasn't easily able figuring out whether xUnit is running tests in parallel.

I can answer that :) We are not running tests in parallel, so you can set the static and revert with a `finally` block. We prefer stable tests above test run time, so we will never turn on parallel testing (and we disable it in xUnit2)

> I'd do such UnitTests...just a bit bussy

Thanks!

304NotModified:
Codecov is lost again... 


@AndreGleichner can you fix the conflict?

304NotModified:
@AndreGleichner is dotnetzip a good library? Then maybe we can release it as separate package. We can then remove it from here and replace it with a mock.

AndreGleichner:
IMO DotNetZip is one of the best pure managed zip libs. The only comparable lib is SharpZipLib. But the latter is GPL thus at work we've used DotNetZip since couple years w/o issues and its licence fits our needs.
Concerning codecov...I'll try to figure out how to increase it....

AndreGleichner:
DotNetZip is only refed by the unit tests...didn't know adding new dependencies there could harm.
To perfom the actual unit test of changed code, it would be enough to inject the ZipArchiveFileCompressor as custom FileCompressor, the mechanics would be the same as with using a real custom implementation.

304NotModified:
The external lib is no problem in the unit tests :) 

But I think it's nice to release the implementation as separate package and then we have twice the same code and tests ;) 

Codecov is buggy. The -4% is wrong. More unit tests are nice, but right now it's unclear to me what isn't covered.

AndreGleichner:
Anything else i could do to get this pr merged?

304NotModified:
Forced a rebuild.  I added an unstable test a few days ago. Not sure what's wrong.

304NotModified:
merged :)

AndreGleichner:
Yeah :)

304NotModified:
@AndreGleichner 

PS hat the difference between DotNetZip and DotNetZip.Reduced

AndreGleichner:
Reduced just means it is w/o self extracting archive support.
Am 11.05.2016 7:46 nachm. schrieb "Julian Verdurmen" <
notifications@github.com>:

304NotModified:
This has been released with NLog 4.3.4 :)

PR #1431, opened 2016-05-03
Prevent Thread.Abort tryout3: event

### LINKED ISSUE #1415: NLog calls Thread.Abort() to exit simple thread loop. It should not
Type: Bug
NLog. 4.3.3
Platform: all

Using Thread.Abort() is just bad practice. Especially for a simple processing loop. It would be better to use a ManualResetEvent, or its more efficient version ManualResetEventSlim

I have attached an example of how this would be implemented. I am not sure about the locking in the FileTarget as there does not seem to be any in this area, but I have added it just in case.

In general, I am not sure about the idea of spinning up a thread like this in the first place, but I do not know enough about the code to figure out a better way. In our application we have 4 log files, so there are 4 threads being spun up. These are looping every 200ms to do something very simple. Would a timer not be a better approach? Why a dedicated thread?

[FileTarget.txt](https://github.com/NLog/NLog/files/243333/FileTarget.txt)

jonreis:
@304NotModified I here is what I was thinking for the removal of the Thread.Abort.

I had to call fileTarget.Dispose in 2 of the unit tests to so that the file handle would be freed up. I don't know if this is an issue or not.

fixes #1415

### DIFF
--- a/src/NLog/Internal/FileAppenders/FileAppenderCache.cs
+++ b/src/NLog/Internal/FileAppenders/FileAppenderCache.cs
@@ -40,12 +40,13 @@ namespace NLog.Internal.FileAppenders
     /// Maintains a collection of file appenders usually associated with file targets.
     /// </summary>
     internal sealed class FileAppenderCache
-    {
+    {        
         private BaseFileAppender[] appenders;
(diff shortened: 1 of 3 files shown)


304NotModified:
superseded by https://github.com/NLog/NLog/pull/1444

Issue #736, opened 2015-06-10
Xml Configuration + loading of NLog assembly from memory may cause invalid path exception


nernst:
This was found while running integration tests with my company's Excel add-in that uses ExcelDNA & NLog. Basic synopsis is that we have an xml-based config for NLog that we embed as a resource into a C# based DLL. At runtime, we extract the resource and use this XML bit to configure NLog. This worked fine for us on NLog 2.0.1.2. But, since upgrading to 3.2.1, I've been failing in our release builds (but not debug).

After a bit (rather a lot) of debugging, realized the difference came down to not NLog, per se, but how the debug vs release builds are packaged, or not. Our debug builds for our ExcelDNA library are pretty bare bones. All of the DLLs sit in a folder with the ExcelDNA scaffolding to create the Excel XLL addin. For our release builds, we distribute 2 XLL files, one for 32-bit Excel, and a separate one for 64-bit (but this doesn't matter to the issue). What matters is the ExcelDNA library provides a facility by which it "packs" .Net libraries into a native DLL (as an XLL file) as resources. When the XLL loads, it handles resolving missing assemblies by trying to find them as native resources. Thing is, when it finds an assembly it needs, it never exists on the filesystem - it's purely in memory. This is where the problem comes in:

Our config, although never modifiable by a user, is stored as an embedded resource in xml format. We utilize the XmlReader constructor interface of XmlLoggingConfiguration to handle the initialization. However, XmlLoggingConfiguration has this member:
```C#
private readonly ConfigurationItemFactory configurationItemFactory = ConfigurationItemFactory.Default;
```
The problem comes in that if the preprocessor flag SILVERLIGHT is not defined, it assumes the NLog assembly has a valid location (in our case, the location is the empty string) @ ConfigurationItemFactory.BuildDefaultFactory (ConfigurationItemFactory.cs, line 238 on master). There is a check if the resulting assembly location is null, but the call to Path.GetDirectoryName(string.Empty) results in an exception that is not trapped, and logging fails to initialize.

I believe I've a workaround by manually creating our configuration via the configuration API (which would be fine because we don't allow our users to change the configuration currently), but I've not yet tested this.

akamyshanov:
As of 4.0.0, the problem is worse. If NLog is loaded from memory (https://github.com/Fody/Costura is used), an exception is thrown from the constructor (even empty) of `FileTarget`, `ConsoleTarget`. Everything is ok with 3.2.1 though. Not sure whether this should be a separate issue.

[...]

PR #1432, opened 2016-05-03
Check if directory NLog.dll is detected in actually exists

gregmac:
Fixes crash when an assembly location is found that doesn't actually exist. 

I believe this happens when using in-memory or bundling loading of the library, as described at https://github.com/NLog/NLog/issues/736#issuecomment-216694452.

### DIFF
--- a/src/NLog/Config/ConfigurationItemFactory.cs
+++ b/src/NLog/Config/ConfigurationItemFactory.cs
@@ -244,9 +244,14 @@ namespace NLog.Config
 #if !SILVERLIGHT
 
             var assemblyLocation = Path.GetDirectoryName(new Uri(nlogAssembly.CodeBase).LocalPath);
-            if (assemblyLocation == null || !Directory.Exists(assemblyLocation))
+            if (assemblyLocation == null)
             {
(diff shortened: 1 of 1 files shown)


304NotModified:
Thanks for this PR.  It's a bit strange that this can happen, but no problem to check it first.

The change is in a good direction, bit I think it wise to:

- write another warning to the InternalLogger in this case
- check related code, like auto loading the xmlconfig (inside xmlloggingconfiguration)

gregmac:
@304NotModified I updated with a separate log message. I was originally going to do that but didn't really have a good idea of what log message to use, hopefully 'not found' is appropriate? 

I'm not sure what should be updated in XmlLoggingConfiguration (though I'm also not familiar with this codebase, other than as a user). All the file operations I see are a result of explicit configuration inside a config file -- so if a user specifies an include path that doesn't exist in a config file, I don't think this change should modify the current behaviour.

304NotModified:
Thanks!

304NotModified:
I checked other places, and those are also OK.

This will be released next minor version. I have changed the log message so also the assemblyLocation will be printed.

PR #1439, opened 2016-05-07
Allow comma separated values (List) for Layout Renderers

304NotModified:
Support voor comma separated list,

e.g.

supported:
- Generic List (`List<T>`), with the following types: 
  - C# built in types (string, int, double, object)
  - enum (use shortname)
  - culture, encoding, Type
  - not supported: Layout

Not supported:
-  Arrays
- Non-generic `List` 
-  `IList<T>` / `IList`
- `IEnumerable<T>` / `IEnumerable` 

Example:

```
${layoutrenderer-with-list:enums=Ignore, Neutral, ignore}
```

``` c#
[LayoutRenderer("layoutrenderer-with-list")]
public class LayoutRendererWithListParam : LayoutRenderer
{
    public List<FilterResult> Enums { get; set; }
   {
       ...
   }
}
```

TO DO
- [x] escape comma -> backtick
-  [x] replace escape with optional quoted
- [x] add unit test for multiple enum

---

This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1439)

### DIFF
--- a/src/NLog/Internal/PropertyHelper.cs
+++ b/src/NLog/Internal/PropertyHelper.cs
@@ -31,6 +31,9 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
+using System.Collections;
+using System.Linq;
+
 namespace NLog.Internal
 {
     using System;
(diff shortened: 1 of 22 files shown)


src/NLog/Internal/PropertyHelper.cs
```diff
@@ -221,15 +225,15 @@ private static bool TryNLogSpecificConversion(Type propertyType, string value, o
...
+        private static bool TryGetEnumValue(Type resultType, string value, out object result, bool flagsEnumAllowed)
         {
             if (!resultType.IsEnum)
             {
                 result = null;
                 return false;
             }
 
-            if (resultType.IsDefined(typeof(FlagsAttribute), false))
+            if (flagsEnumAllowed && resultType.IsDefined(typeof(FlagsAttribute), false))
```
bhaeussermann: What if it's a flags-enum, but `flagsEnumAllowed` is false? The else-part would then execute with the flags-enum.

bhaeussermann: How about adding a test for parsing a list of values of a flag-enum?

bhaeussermann:
Any idea why the test coverage dropped so dramatically? I don't seem to be able to view the codecov diff.

src/NLog/Internal/PropertyHelper.cs
304NotModified: Good idea

304NotModified:
> Any idea why the test coverage dropped so dramatically? I don't seem to be able to view the codecov diff.

It did not. 75.30% (+<.01%) compared to 717373a

but this PR has "only"  74.29% coverage

304NotModified:
I will remove the ecape and will move to a optional quoted approach, 

eg. `a,b,c` or `'a,b',c`

304NotModified:
docs: https://github.com/NLog/NLog/wiki/Properties-constraints-for-custom-extensions

PR #1424, opened 2016-05-02
Prevent Thread.Abort tryout1: ManualResetEvent instead of Thread.Abort

304NotModified:
code by @jonreis, see https://github.com/NLog/NLog/issues/1415

PR #1444, opened 2016-05-08
FileTarget: Use ManualResetEvent instead of Thread.Abort to stop appender-invalidator-thread

### LINKED ISSUE #1415: NLog calls Thread.Abort() to exit simple thread loop. It should not

bhaeussermann:
Fix #1415

### DIFF
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -692,23 +692,18 @@ namespace NLog.Targets
                             {
                                 while (true)
                                 {
-                                    try
-                                    {
-                                        Thread.Sleep(200);
-                                    }
-                                    catch (ThreadAbortException ex)
-                                    {
-                                        //ThreadAbortException will be automatically re-thrown at the end of the try/catch/finally if ResetAbort isn't called.
-                                        Thread.ResetAbort();
-                                        InternalLogger.Trace(ex, "ThreadAbortException in Thread.Sleep");
-                                    }
-                                    catch (Exception ex)
-                                    {
-                                        InternalLogger.Warn(ex, "Exception in Thread.Sleep, most of the time not an issue.");
-                                    }
-                                   
+                                    this.fileAppenderCache.LogArchiveWaitHandle.WaitOne();
+
                                     lock (SyncRoot)
+                                    {
+                                        if (!this.fileAppenderCache.LogFileWasArchived)
+                                        {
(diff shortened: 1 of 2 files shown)


304NotModified:
supersedes https://github.com/NLog/NLog/pull/1424?

bhaeussermann:
> supersedes #1424?

Yes.

src/NLog/Targets/FileTarget.cs
```diff
@@ -694,21 +694,23 @@ private void RefreshArchiveFilePatternToWatch()
...
+                                                break;
+                                            }
+
+                                            this.fileAppenderCache.InvalidateAppendersForInvalidFiles();
+                                        }
                                     }
                                     catch (Exception ex)
                                     {
-                                        InternalLogger.Warn(ex, "Exception in Thread.Sleep, most of the time not an issue.");
+                                        InternalLogger.Info(ex, "Exception in FileTarget appender-invalidator thread.");
```
304NotModified: Maybe debug level is better?

src/NLog/Targets/FileTarget.cs
```diff
@@ -719,17 +721,26 @@ private void RefreshArchiveFilePatternToWatch()
...
-                        this.appenderInvalidatorThread.Abort();
-                        this.appenderInvalidatorThread = null;
-                    }
+                    this.StopAppenderInvalidatorThread();
                 }
             }
 #endif
         }
 
+        private void StopAppenderInvalidatorThread()
```
304NotModified: I think some docs are nice, as this is a pretty difficult case (to test at least)

304NotModified:
Some notes on docs and internal Logging. If I find time, I did this after merge.

304NotModified:
merged :)

PR #1469, opened 2016-05-17
Allow overwriting possible nlog configuration file paths

304NotModified:
- [x] `GetDefaultCandidateConfigFileNames` to public

### DIFF
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -171,7 +171,8 @@ namespace NLog
                     // Retest the condition as we might have loaded a config.
                     if (this.config == null)
                     {
-                        foreach (string configFile in GetCandidateConfigFileNames())
+                        var configFileNames = CandidateConfigFileNames ?? GetDefaultCandidateConfigFileNames();
+                        foreach (string configFile in configFileNames)
                         {
 #if SILVERLIGHT
                             Uri configFileUri = new Uri(configFile, UriKind.Relative);
@@ -352,6 +353,14 @@ namespace NLog
                 return configuration != null ? configuration.DefaultCultureInfo : null;
             }
         }
+
+        /// <summary>
+        /// Overwrite possible file paths (including filename) for possible NLog config files. 
+        /// When this property is <c>null</c>, the default file paths (<see cref="GetDefaultCandidateConfigFileNames"/> are used.
+        /// 
+        /// </summary>
+        public IList<string> CandidateConfigFileNames { get; set; }
+
 
         private void LogConfigurationInitialized()
         {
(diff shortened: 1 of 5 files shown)


304NotModified:
> Instead of providing with set and get property. We can make this variable private and Can create a method "OverwriteConfigFileNames" and set the private variable internally. So it is very obvious API Users. 

Nice idea, but what is the benefit over a get and setter?

Page-Not-Found:
Based on coding standard (https://msdn.microsoft.com/en-us/library/ms182327.aspx), collections should not have setter. This is one reason for the suggestion.

304NotModified:
Good point :)

But I could be a getter only indeed

304NotModified:
@Page-Not-Found your opinion please :)

src/NLog/LogFactory.cs
```diff
@@ -865,10 +872,51 @@ protected virtual void Dispose(bool disposing)
...
+        /// <param name="filePaths">The filepaths to the possible config file</param>
+        public void SetCandidateConfigFilePaths(IEnumerable<string> filePaths)
+        {
+            candidateConfigFilePaths = new List<string>();
+
+            if (filePaths != null)
+            {
+                foreach (var filePath in filePaths)
+                {
+                    candidateConfigFilePaths.Add(filePath);
```
Page-Not-Found: You could just use "AddRange" instead of loop and a add.

src/NLog/LogFactory.cs
```diff
@@ -865,10 +872,51 @@ protected virtual void Dispose(bool disposing)
...
+                }
+            }
+        }
+        /// <summary>
+        /// Clear the candidate file paths and return to the defaults.
+        /// </summary>
+        public void ClearCandidateConfigFilePath()
+        {
```
Page-Not-Found: Can be renamed "Reset...." so it is matching with the "Set...." method.

src/NLog/LogManager.cs
```diff
@@ -178,6 +178,32 @@ public static GetCultureInfo DefaultCultureInfo
...
+            return factory.GetCandidateConfigFilePaths();
+        }
+
+        /// <summary>
+        /// Overwrite the paths (including filename) for the possible NLog config files.
+        /// </summary>
+        /// <param name="filePaths">The filepaths to the possible config file</param>
+        public static void SetCandidateConfigFilePaths(IEnumerable<string> filePaths)
+        {
+            factory.SetCandidateConfigFilePaths(filePaths);
```
Page-Not-Found: Same as above.

Page-Not-Found:
@304NotModified  Minor suggestions.

304NotModified:
Thanks @Page-Not-Found !

One thing, the name of the method. `LogManager.GetCandidateConfigFilePaths`, it really visible (next to autocomplete of `GetCurrentClassLogger` : 

image

Maybe it should move it to somewhere else? What do you think?

Page-Not-Found:
I am not sure. if it is public right now. If it is something added for this Issue. I would recommend hiding it by marking as internal.

src/NLog/LogFactory.cs
304NotModified: Thanks, it was first a `IList` but with the move to `List`, AddRange is indeed better

src/NLog/LogFactory.cs
304NotModified: That name is more clear indeed, thx1

Page-Not-Found: I agree. You can ignore the comment and go ahead.

304NotModified:
It should be public, as I need it for NLog.Extensions.Logging

But I think its better to move the LogManager methods to `XmlLoggingConfiguration` or `LoggingConfiguration`. What do you think?

304NotModified:
moved it to XMLLoggingConfiguration, it's easier to find and less pollutes LogManager IMO

PR #1470, opened 2016-05-17
Parse <Variable> before <Include> in XML config

304NotModified:
Parse `<Variable>` before `<Include>` so we can use them in the included files.

TODO not sure, this is maybe a breaking change? (we cannot overwrite a var in a main.config)


304NotModified:
breaking change = no merge

Issue #1361, opened 2016-04-04
Add "else" option for ${when}


304NotModified:
See https://nlog.userecho.com/topics/28-add-else-option-for-when/

PR #1471, opened 2016-05-17
Add else option to ${when}

304NotModified:
Add else option to ${when} to prevent condition duplication.

example: `${when:good:when=logger=='logger':else=better}`

Also handle potential nullref exception

fixes https://github.com/NLog/NLog/issues/1361

### DIFF
--- a/src/NLog/LayoutRenderers/Wrappers/WhenLayoutRendererWrapper.cs
+++ b/src/NLog/LayoutRenderers/Wrappers/WhenLayoutRendererWrapper.cs
@@ -31,6 +31,8 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
+using NLog.Layouts;
+
 namespace NLog.LayoutRenderers.Wrappers
 {
     using NLog.Conditions;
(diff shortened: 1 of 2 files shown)


janrhansen:
I cant figure out how to use this... could you please provide a simple example like this
<when condition="contains('${exception}', 'already exists in')"  action="Log"></when>
extended with the "else" option? I assume the intention is to select "Log" in one case and "Ignore" in another?
I have achieved that by adding a 
<when condition="equals('1', '1')" action="Ignore"></when>
as my last condition, so only exceptions which I find "interesting" are logged. The rest are ignored. The optimal solution would be to let me switch target, based on filters like:
<when condition="contains('${exception}', 'already exists in')"  good:action="Log":writeTo:"target1" else:action="Ignore"></when>
- but is that possible somehow?

304NotModified:
The else is only added for `${when}` and not for filters. 

Example:

    ${when:good:when=logger=='logger':else=better}

I used `<when condition='true' action="Ignore">` as fallback, but I do have to agree it's not optimal

PR #1474, opened 2016-05-20
FileTarget concurrentWrites performance enhancement when used w/o Async-/BufferingWrapper

AndreGleichner:
When having concurrentWrites/keepFileOpen set to true NLog utilizes a mutex shared with all other processes. Using this synchronization mechanism together with using a FileStream which is auto-flushed during every write is quite costly. This PR improves FileTarget performance by removing the need for using that mutex/FileStream.

On my notebook this gives **30-40% more event writes/sec**.

When FileTarget is used with Async- or BufferingWrapper there's no notable effect.

**Background:**
It is rarely known that when creating/opening a file under Windows with CreateFile(...FILE_APPEND_DATA | SYNCHRONIZE...) that writes are atomically appended, even cross process. See https://blogs.msdn.microsoft.com/oldnewthing/20151127-00/?p=92211/

---

This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1474)

### DIFF
--- a/src/NLog/Internal/FileAppenders/UnleashedMultiProcessFileAppender.cs
+++ b/src/NLog/Internal/FileAppenders/UnleashedMultiProcessFileAppender.cs
@@ -1,0 +1,194 @@
+// 
+// Copyright (c) 2004-2016 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without 
+// modification, are permitted provided that the following conditions 
+// are met:
+// 
+// * Redistributions of source code must retain the above copyright notice, 
+//   this list of conditions and the following disclaimer. 
+// 
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution. 
+// 
+// * Neither the name of Jaroslaw Kowalski nor the names of its 
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission. 
+// 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE 
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE 
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE 
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR 
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS 
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN 
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) 
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF 
+// THE POSSIBILITY OF SUCH DAMAGE.
+// 
+
+#if !SILVERLIGHT && !__IOS__ && !__ANDROID__
+
+namespace NLog.Internal.FileAppenders
+{
+    using System;
+    using System.Security;
+    using NLog.Common;
+    using System.Runtime.InteropServices;
+    /// <summary>
+    /// Provides a multiprocess-safe atomic file append while
+    /// keeping the files open.
+    /// </summary>
+    [SecuritySafeCritical]
+    internal class UnleashedMultiProcessFileAppender : BaseFileAppender
+    {
+        public static readonly IFileAppenderFactory TheFactory = new Factory();
+
+        private Microsoft.Win32.SafeHandles.SafeFileHandle _file;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="UnleashedMultiProcessFileAppender" /> class.
+        /// </summary>
+        /// <param name="fileName">Name of the file.</param>
+        /// <param name="parameters">The parameters.</param>
+        public UnleashedMultiProcessFileAppender(string fileName, ICreateFileParameters parameters) : base(fileName, parameters)
+        {
+            try
+            {
+                CreateAppendOnlyFile(fileName);
+            }
+            catch
+            {
+                if (_file != null)
+                {
+                    if (!_file.IsClosed)
+                        _file.Close();
+                    _file = null;
+                }
+
+                throw;
+            }
+        }
+
+        private void CreateAppendOnlyFile(string fileName)
+        {
+            int fileShare = Win32FileNativeMethods.FILE_SHARE_READ | Win32FileNativeMethods.FILE_SHARE_WRITE;
+
+            if (this.CreateFileParameters.EnableFileDelete)
+            {
+                fileShare |= Win32FileNativeMethods.FILE_SHARE_DELETE;
+            }
+
+            try
+            {
+                // https://blogs.msdn.microsoft.com/oldnewthing/20151127-00/?p=92211/
+                // https://msdn.microsoft.com/en-us/library/ff548289.aspx
+                // If only the FILE_APPEND_DATA and SYNCHRONIZE flags are set, the caller can write only to the end of the file, 
+                // and any offset information about writes to the file is ignored.
+                // However, the file will automatically be extended as necessary for this type of write operation.
+
+                _file = Win32FileNativeMethods.CreateFile(
+                    fileName,
+                    Win32FileNativeMethods.FileAccess.FileAppendData | Win32FileNativeMethods.FileAccess.Synchronize,
+                    fileShare,
+                    IntPtr.Zero,
+                    Win32FileNativeMethods.CreationDisposition.OpenAlways,
+                    this.CreateFileParameters.FileAttributes,
+                    IntPtr.Zero);
+
+                if (_file.IsInvalid)
+                {
+                    Marshal.ThrowExceptionForHR(Marshal.GetHRForLastWin32Error());
+                }
+            }
+            catch
+            {
+                if ((_file != null) && (!_file.IsClosed))
+                    _file.Close();
+
+                _file = null;
+
+                throw;
+            }
+        }
+
+        /// <summary>
+        /// Writes the specified bytes.
+        /// </summary>
+        /// <param name="bytes">The bytes to be written.</param>
+        public override void Write(byte[] bytes)
+        {
+            try
+            {
+                uint written;
+                bool success = Win32FileNativeMethods.WriteFile(_file.DangerousGetHandle(), bytes, (uint)bytes.Length, out written, IntPtr.Zero);
+                if (!success || (uint)bytes.Length != written)
+                {
+                    Marshal.ThrowExceptionForHR(Marshal.GetHRForLastWin32Error());
+                }
+                FileTouched();
+            }
+            finally
+            {
+            }
+        }
+
+        /// <summary>
+        /// Closes this instance.
+        /// </summary>
+        public override void Close()
+        {
+            InternalLogger.Trace("Closing '{0}'", FileName);
+            if (_file != null && !_file.IsClosed)
+            {
+                _file.Close();
+            }
+
+            _file = null;
+            FileTouched();
+        }
+
+        /// <summary>
+        /// Flushes this instance.
+        /// </summary>
+        public override void Flush()
+        {
+            // do nothing, the file is written directly
+        }
+
+        /// <summary>
+        /// Gets the file info.
+        /// </summary>
+        /// <returns>The file characteristics, if the file information was retrieved successfully, otherwise null.</returns>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Reliability", "CA2001:AvoidCallingProblematicMethods", MessageId = "System.Runtime.InteropServices.SafeHandle.DangerousGetHandle", Justification = "Optimization")]
+        public override FileCharacteristics GetFileCharacteristics()
+        {
+            return FileCharacteristicsHelper.Helper.GetFileCharacteristics(FileName, _file.DangerousGetHandle());
+        }
+        
+        /// <summary>
+        /// Factory class.
+        /// </summary>
+        private class Factory : IFileAppenderFactory
+        {
+            /// <summary>
+            /// Opens the appender for given file name and parameters.
+            /// </summary>
+            /// <param name="fileName">Name of the file.</param>
+            /// <param name="parameters">Creation parameters.</param>
+            /// <returns>
+            /// Instance of <see cref="BaseFileAppender"/> which can be used to write to the file.
+            /// </returns>
+            BaseFileAppender IFileAppenderFactory.Open(string fileName, ICreateFileParameters parameters)
+            {
+                return new UnleashedMultiProcessFileAppender(fileName, parameters);
+            }
+        }
+    }
+}
+
+#endif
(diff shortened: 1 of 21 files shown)


src/NLog/Internal/FileAppenders/UnleashedMultiProcessFileAppender.cs
```diff
@@ -0,0 +1,215 @@
...
+    using System.Security;
+    using NLog.Common;
+    using System.Runtime.InteropServices;
+    using System.IO;
+    /// <summary>
+    /// Provides a multiprocess-safe atomic file append while
+    /// keeping the files open.
+    /// </summary>
+    [SecuritySafeCritical]
+    internal class UnleashedMultiProcessFileAppender : BaseFileAppender
```
bhaeussermann: I have reservations about this class name ("unleashed" is not very descriptive to me).  
How about `WindowsMultiProcessFileAppender`?

src/NLog/Internal/FileAppenders/UnleashedMultiProcessFileAppender.cs
```diff
@@ -0,0 +1,215 @@
...
+    {
+        public static readonly IFileAppenderFactory TheFactory = new Factory();
+
+        private Microsoft.Win32.SafeHandles.SafeFileHandle _file;
```
bhaeussermann: Please refer to our other source files for our naming conventions for variables.

src/NLog/Internal/FileAppenders/UnleashedMultiProcessFileAppender.cs
```diff
@@ -0,0 +1,215 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Writes the specified bytes.
+        /// </summary>
+        /// <param name="bytes">The bytes to be written.</param>
+        public override void Write(byte[] bytes)
+        {
+            try
```
bhaeussermann: This `try` does not affect flow of execution and can be removed.

src/NLog/Internal/FileAppenders/UnleashedMultiProcessFileAppender.cs
```diff
@@ -0,0 +1,215 @@
...
+            /// Instance of <see cref="BaseFileAppender"/> which can be used to write to the file.
+            /// </returns>
+            BaseFileAppender IFileAppenderFactory.Open(string fileName, ICreateFileParameters parameters)
+            {
+                try
+                {
+                    if (!parameters.ForceManaged && PlatformDetector.IsDesktopWin32)
+                        return new UnleashedMultiProcessFileAppender(fileName, parameters);
+                }
+                catch
```
bhaeussermann: How about logging the exception to the internal logger as a warning? What do you think @304NotModified ?

bhaeussermann:
I've tested the change successfully with three concurrent processes writing 1000 logs each. Other than my feedback above, everything looks good ;)

src/NLog/Internal/FileAppenders/UnleashedMultiProcessFileAppender.cs
304NotModified: Good idea! Empty catches are a no go.  

If it really not important, then debug level is also OK.

304NotModified:
Thanks for the PR and @bhaeussermann thanks for the review!

src/NLog/Internal/FileAppenders/UnleashedMultiProcessFileAppender.cs
AndreGleichner: I'll change that

src/NLog/Internal/FileAppenders/UnleashedMultiProcessFileAppender.cs
AndreGleichner: Just a reflex to have private prefixed with underscore ;)

src/NLog/Internal/FileAppenders/UnleashedMultiProcessFileAppender.cs
AndreGleichner: Copy/Paste relict from MutexMultiProcessFileAppender.

src/NLog/Internal/FileAppenders/UnleashedMultiProcessFileAppender.cs
AndreGleichner: catch is just so that the fallback to MutexMultiProcessFileAppender will get its chance. So I'll place the return "new MutexMultiProcessFileAppender(..)" directly into the catch block preceded with an internal debug log.

bhaeussermann:
I'm happy with the new changes. What's up with codecov?

AndreGleichner:
Don't know codecov calcs coverage. Possibly it is based on a test run for just one or at least not all platforms!? I guess now the previous MutexMultiProcessFileAppender is now no longer executed during testing. How about adding ForceManaged as a variation to test run for some or all FileTarget tests?

AndreGleichner:
Hmmm, seems there's something odd happening...SimpleConcurrentTest case is randomly failing...

304NotModified:
FYI I think this that some time to merge this one. Need to (performance) test is and double-double check as it can have a big impact.

304NotModified:
Is it possible to add some more comments? This will help double checking the code!

AndreGleichner:
I 'm in the middle of fixing up some issues with concurrent tests, unrelated to my change. They were not actually testing concurrency when using async/buffered writing. In my change there was a racing condition with creationtime fixup. Currently im still struggling with a single test case, were my code fails, most likely because my code is just too fast ;) I believe getting this finished this weekend, also with lots of additional comments.

304NotModified:
:+1:

AndreGleichner:
Are there any existing perftests that i could play with?

bhaeussermann:
@AndreGleichner [Here](https://github.com/bhaeussermann/NLogConcurrencyTest) is the solution that I've been using to test it.

AndreGleichner:
Written my own perf tests, since everything I've seen elsewere is not actually doing a lot of concurrent stuff. See here https://github.com/AndreGleichner/TestFileAppend

Problem with concurrency issues is always that stuff might work correctly in general, but may fail randomly with odd racing condition, etc so we need to stress things...
TestFileAppend can massively stress NLog. I've performed lots of combinations of NLog configuration, number of concurrent processes, number of log events every process is writing and log message length.
Most of the time I did run 100 processes in parallel. When doing async/buffered file target it requires lots of messages to keep every single process up for lengthy times. I witten mostly 100K messages per process.
When doing synchronous writes, stuff is naturally much slower and I usually have written 10K events per process.

Bottomline: I wasn't able to break my code in this PR :)

304NotModified:
Most important: what are the test results differences between this pr and the master?

304NotModified:
PS in on holiday so with limited Internet, time and computer.

src/NLog/Targets/FileTarget.cs
```diff
@@ -645,11 +648,33 @@ public bool EnableArchiveFileCompression
...
+        /// Gets or set a value indicating whether a managed file stream is forced, instead of using the native implementation.
         /// </summary>
         [DefaultValue(false)]
         public bool ForceManaged { get; set; }
 
+#if !SILVERLIGHT && !__IOS__ && !__ANDROID__
+        /// <summary>
+        /// Gets or sets a value indicationg whether file creation calls should be synchronized by a system global mutex.
+        /// </summary>
+        /// [DefaultValue(false)]
```
304NotModified: this should be uncommented?

304NotModified:
Hi! Sorry about the delay.

I like to merge this one, but I'm cautious with these changes. If something is wrong, this will influence a lot - as the file target is used a lot.

We could do two things: 

1). test it more, I created a new repos for that (but too bad had no time to complete it): https://github.com/NLog/NLog-Performance-tests
2) make this change optional and maybe in the future default.

What do you think

AndreGleichner:
Unfortunately I've been extremely busy during the last weeks. I've started some systematic testing to check under which conditions my changes have consideral positive effects but currently can't find the time to get it finished.
What was already clear before I even started, is that only unbuffered writes become faster...up to ~40% but at least it never is slower than the previous code. Beside the pure speed the new code creates fewer objects, although I guess that objects in this area are dying young in gen0, almost never being promoted to gen1, etc. so this generally isn't such a big issue.
I already did massive tests using this project: https://github.com/AndreGleichner/TestFileAppend
I'd say the code is stable, but unless more people have used it in other scenarios you can't really know for sure ;)
Doing concurrent stuff like this right is inherently hard. Previous unit tests were just not good enough to uncover concurrency issues and I tried to make them better.
I'm unsure how we could get more confidence. IMO it requires lots of different integrations into real target projects...some beta phase. If it would be optional, nobody really cares and would enable it.
So maybe get it into 4.4.0 beta is an option? IMO release tests should have some massive stress test suites. I use NLog at work, were the bigger customer installations are generating log files > 1TB/day.

304NotModified:
> If it would be optional, nobody really cares and would enable it.

Not sure about that, as some people really care about the performance. But anyway, this won't gives us confidence as we don't know how many people tried it. 

> Doing concurrent stuff like this right is inherently hard.

Can't agree more! That's also why I'm defensive. 

> So maybe get it into 4.4.0 beta is an option?

I think a beta/RC would be the best approach indeed, too bad is NLog 4.4 already large affected by many changes (for .NET Core). 

Also creating a beta/RC for a patch version (4.3.7) is maybe a bit strange, do you agree? 

Maybe I'm pushing NLog 4.4 to 4.5, but now sure how many people would be unhappy as .NET Core won't work anymore on NLog 4.4

AndreGleichner:
I agree that having something like 4.3.8-beta1 sounds odd. Moving Core support to 4.5 is most likely upsetting people unless you'll issue 4.4-beta1 in parallel to 4.5-beta42, which also is quite strange. I also have a strong interest in having a stable .Net Core Version somewhen this year.
How about just pushing this PR into some of the next patch releases w/o any other fix, so that in case there're problems found in the field with my change, they can simply go back to the previous version or we may have an option to disable it at runtime. And you hopefully receive feedback about such issue. In such case we could either fix it before the next patch release or turn it off by default then.

304NotModified:
@AndreGleichner 

Probable .NET Core will be moved to 5.0, so we could use 4.4 with this change. 

Is is possible to push the test program to https://github.com/NLog/NLog-Performance-tests ? Just add a folder to the root and reference NLog in the submodule

304NotModified:
@AndreGleichner 

Update,

we noticed a few days ago that the performance of the FileTarget was really bad in 4.3. It was 3-4 times slower compared to 4.2.3. (see https://github.com/NLog/NLog-Performance-tests/issues/1), we are in the middle of fixing that, (4.3.8 probably) and this will move some code around. 

Is it possible to run your tests again and compare it with 4.2.3? I'm curious if we also gain speed then. If so, I will definitely will pick this PR up in the future.

AndreGleichner:
I'll try to find some spare time.

304NotModified:
Reviewed 13 of 15 files at r2, 8 of 8 files at r3.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[src/NLog/Targets/FileTarget.cs, line 660 \[r3\]](https://reviewable.io:443/reviews/nlog/nlog/1474#-KPhkvF98MlPbPCy4zn5-r3-660:-KPho4Y_C-jKoNr8BPpW:bs4ld2j) ([raw file](https://github.com/nlog/nlog/blob/12101a19f8727d78a2f0b5e0a639e240761dfc28/src/NLog/Targets/FileTarget.cs#L660)):*
> this should be uncommented?
</details>
this is still to do

---

*[src/NLog/Targets/FileTarget.cs, line 894 \[r3\]](https://reviewable.io:443/reviews/nlog/nlog/1474#-KPhoLE3XcmpBjNkjq7G:-KPhoLE3XcmpBjNkjq7H:b36d8kl) ([raw file](https://github.com/nlog/nlog/blob/12101a19f8727d78a2f0b5e0a639e240761dfc28/src/NLog/Targets/FileTarget.cs#L894)):*
> ```Smalltalk
>                 return MutexMultiProcessFileAppender.TheFactory;
> #else
>                 if (!this.PreferMutexLockedFileCreation && !this.ForceManaged && PlatformDetector.IsDesktopWin32)
> ```

would be nice to create a variable for this

---

*[tests/NLog.UnitTests/Targets/ConcurrentFileTargetTests.cs, line 63 \[r3\]](https://reviewable.io:443/reviews/nlog/nlog/1474#-KPho_N3ojlmarGaOGC8:-KPho_N3ojlmarGaOGC9:b-pu7o28) ([raw file](https://github.com/nlog/nlog/blob/12101a19f8727d78a2f0b5e0a639e240761dfc28/tests/NLog.UnitTests/Targets/ConcurrentFileTargetTests.cs#L63)):*
> ```Smalltalk
>             ft.OpenFileCacheSize = 1;
>             ft.LineEnding = LineEndingMode.LF;
>             ft.PreferMutexLockedFileCreation = modes.Length == 2 && modes[1] == "mutex" ? true : false;
> ```

the string is double, it should be a list/array or flag enum

but this check is problematic. why no contains?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1474)*

304NotModified:
just tried reviewable.io on this PR :)

304NotModified:
Hi,

NLog 4.3.8 is online and we have fixed a performance issue in the FileTarget.

I did now a short benchmark (simple program, one loop, one thread):

```xml
  <target name ="mipLoggertarget" xsi:type="File"
                  fileName="c:/temp/Log/test1.log"
                  layout="${message}" keepFileOpen="true" AutoFlush="false"
                  ConcurrentWrites="true"
                  CleanupFileName="false"/>
``` 

master: (4.3.8) b8362f5913af6e4ec51e62e2a47c6deba1b8f5bd

image


this pr after merge master:
c31b4e3b3fdb882a80d9936e420adb251bb6182f

image

So looks promising! 

could you retest the performance, (after merge) with the master, and post the test code to https://github.com/NLog/NLog-Performance-tests ?

304NotModified:
@AndreGleichner any change to compare it with the current release? I'm not sure if there is a performance benefit here.

I really appreciate your time for this PR, but if there isn't a significant performance gain, we will not merge this one.

snakefoot:
Made a very simple test after merging with main-branch, and WindowsMultiProcessFileAppender is twice as fast, compared to MutexMultiProcessFileAppender.

With a single-process then WindowsMultiProcessFileAppender  could do 350.000 msgs/sec, and MutexMultiProcessFileAppender could do 150.000 msgs/sec.

snakefoot:
Instead making special filehandle, then one could consider just opening the file like this:

```
                bool fileExists = File.Exists(fileName);
                fileStream = new FileStream(
                    fileName,
                    FileMode.Append,
                    System.Security.AccessControl.FileSystemRights.AppendData, // <- Atomic append
                    fileShare,
                    1,  // No internal buffer, write directly from user-buffer
                    FileOptions.None);
                if (fileExists || fileStream.Position > 0)
                {
                                // File was not created by this process
                }
```

One will loose the atomic check if file already exists regarding creationTime. But I guess the small workaround with fileExists also works.

304NotModified:
> Instead making special filehandle, then one could consider just opening the file like this:

What code will be replaced by your proposal?

snakefoot:
This is how my CreateAppendOnlyFile looks like, and all changes for handling native-filehandle are gone:

```
        private void CreateAppendOnlyFile(string fileName)
        {
            string dir = Path.GetDirectoryName(fileName);
            if (!Directory.Exists(dir))
            {
                if (!CreateFileParameters.CreateDirs)
                {
                    throw new DirectoryNotFoundException(dir);
                }
                Directory.CreateDirectory(dir);
            }

            var fileShare = FileShare.ReadWrite;
            if (this.CreateFileParameters.EnableFileDelete)
                fileShare |= FileShare.Delete;

            try
            {
                bool fileExists = File.Exists(fileName);

                fileStream = new FileStream(
                    fileName,
                    FileMode.Append,
                    System.Security.AccessControl.FileSystemRights.AppendData, // <- Atomic append
                    fileShare,
                    1,  // No internal buffer, write directly from user-buffer
                    FileOptions.None);

                if (fileExists || fileStream.Position > 0)
                {
                    this.CreationTime = File.GetCreationTimeUtc(this.FileName);
                    if (this.CreationTime < DateTime.UtcNow - TimeSpan.FromSeconds(2))
                    {
                        Thread.Sleep(10);
                        this.CreationTime = File.GetCreationTimeUtc(this.FileName);
                    }
                }
                else
                {
                      this.CreationTime = DateTime.UtcNow;
                    File.SetCreationTimeUtc(this.FileName, this.CreationTime);
                }
            }
            catch
            {
                if (fileStream != null)
                    fileStream.Dispose();
                fileStream = null;
                throw;
            }
        }
```

304NotModified:
Superseded by https://github.com/NLog/NLog/pull/1756

PR #1481, opened 2016-05-25
Feature: get items for diagnostic contexts

### LINKED ISSUE #1315: Get all key-values pairs from Diagnostics Context classes
**Type** :  Feature request

**NLog version**: >=4.1.1

**Platform**: .NET4.5, mono

**Description**:
I am investigating the use of GlobalDiagnosticsContext and MappedDiagnosticsLogicalContext (MDLC) for our system. 

We use a custom layout renderer that transforms the log to a json message that will be sent to a remote log collecting system. I would like to add _all_ values available in MDLC to this json message, but the methods the MDLC class only allow retrieval of a specific item.

I would like a way to get a public readonly dictionary from these classes that contain all values.

Note that MappedDiagnosticsContext makes its dictionary "internal", so that renderers in the NLog package can access the data.

tiljanssen:
Fixes #1315

### DIFF
--- a/src/NLog/GlobalDiagnosticsContext.cs
+++ b/src/NLog/GlobalDiagnosticsContext.cs
@@ -113,6 +113,18 @@ namespace NLog
         }
 
         /// <summary>
+        /// Returns all item names
+        /// </summary>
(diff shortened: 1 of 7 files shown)


304NotModified:
Thanks!

Issue #1486, opened 2016-05-27
Use Ansi Color Codes for ColoredConsoleTarget on Linux/.Net Core


jp7677:
Hi there,

Thanks a lot for already providing a beta for .Net Core RC2. It already works pretty cool!

I would love to use the ColoredConsoleTarget on Linux. Unfortunately the output is on my Linux box identical with the normal ConsoleTarget, thus no colors. I guess because Console.ForegrondColor and friends are quite different on Linux, of even not implemented on those platforms. Looking at Microsoft's own Aspnet Core Logger they make a distinction between Windows and Non-windows. On Non-Windows they use the regular System.Console class too, but use Ansi Color codes as message prefix instead of using Console.ForegrondColor. See here [https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Console/Internal/AnsiLogConsole.cs](https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Console/Internal/AnsiLogConsole.cs) and [https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs](https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs).

Probably Nlog should do something similar. I guess using Ansi Color Code is also slightly faster..

Regards and thanks a lot for your work.
Jens

304NotModified:
Good idea, although there a some issues with the console logger in AspNet Core:

See https://github.com/aspnet/Logging/issues/281 and https://github.com/aspnet/Logging/issues/428

jp7677:
Thank you for your response. I have played a little bit and this is my result so far [https://github.com/NLog/NLog/compare/coreCLR...jp7677:coreCLR](https://github.com/NLog/NLog/compare/coreCLR...jp7677:coreCLR). Works fine on my Linux box ;) 
Would you be interested in a pull request if I add the missing color codes and have finished the Word highlighting? Though first I need to understand whats happening in the word highlighting code..

Regarding both issues you linked to: Well, the docker boys could use the normal ConsoleTarget or wait till the docker extension in Visual Studio is fixed. The other issue about the colors on Mac is not clear to me. We could use the ansi codes on Linux only.

jp7677:
[...]

PR #1495, opened 2016-06-04
Use ansi color codes on unix based platforrms

jp7677:
See https://github.com/NLog/NLog/issues/1486
Note that this change does effect all unix platforms, thus Linux/Mac on Mono and .NetCore and may be others, so not just .Net Core.
Let me know what you think.

---

This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1495)


jp7677:
I'm quite pleased now with my additions. Let me know what you think.

src/NLog/Targets/AnsiConsoleColor.cs
```diff
@@ -0,0 +1,136 @@
...
+
+namespace NLog.Targets
+{
+    using System;
+    
+    internal static class AnsiConsoleColor
+    {
+        internal static string GetForegroundColorEscapeCode(ConsoleColor color)
+        {
+            switch (color)
```
304NotModified: I prefer a static dictionary (because of cyclomatic complexity). What do you think?

src/NLog/Targets/AnsiConsoleColor.cs
```diff
@@ -0,0 +1,136 @@
...
+                case ConsoleColor.White:
+                    return "\x1B[97m";
+                case ConsoleColor.Yellow:
+                    return "\x1B[93m";
+                default:
+                    return "\x1B[39m"; // default foreground color
+            }
+        }
+        
+        internal static string GetTerminalDefaultForegroundColorEscapeCode()
```
304NotModified: Prefer a property instead of method

src/NLog/Targets/AnsiConsoleColorizer.cs
```diff
@@ -0,0 +1,197 @@
...
+
+    /// <summary>
+    /// Colorizes console output using ansi escape codes.
+    /// See https://en.wikipedia.org/wiki/ANSI_escape_code#Colors for background info.
+    /// </summary>
+    internal class AnsiConsoleColorizer
+    {
+	    internal static readonly IList<ConsoleRowHighlightingRule> DefaultConsoleRowHighlightingRules = new List<ConsoleRowHighlightingRule>()
+        {
+            new ConsoleRowHighlightingRule("level == LogLevel.Fatal", ConsoleOutputColor.DarkRed, ConsoleOutputColor.NoChange),
```
304NotModified: this is duplication with ColoredConsoleTarget, not sure if we could fix that easiliy

src/NLog/Targets/AnsiConsoleColorizer.cs
```diff
@@ -0,0 +1,197 @@
...
+            var formattedMessage = ColorizeRow();
+                if (wordHighlightingRules.Count != 0)
+                    formattedMessage = ApplyWordHighlightingRules();
+                    
+            return formattedMessage;
+        }
+        
+        private string ColorizeRow()
+        {
+            StringBuilder builder = new StringBuilder(1);
```
304NotModified: why (1)?

src/NLog/Targets/AnsiConsoleColorizer.cs
```diff
@@ -0,0 +1,197 @@
...
+                            hl.ForegroundColor, hl.BackgroundColor, 
+                            matchBelowEndOfCurrentMatch != null ? matchBelowEndOfCurrentMatch.ForegroundColor : rowHighlightingRule.ForegroundColor, 
+                            matchBelowEndOfCurrentMatch != null ? matchBelowEndOfCurrentMatch.BackgroundColor : rowHighlightingRule.BackgroundColor);
+                });
+            }
+
+            return message;
+        }
+        
+        private List<HighlightedMatch> BuildMatchList()
```
304NotModified: isn't neater to build a dictionary? (for the matching on Id)?

src/NLog/Targets/AnsiConsoleColorizer.cs
```diff
@@ -0,0 +1,197 @@
...
+            return matches.FindAll(hw => hw.Layer < currentMatch.Layer &&
+                                   hw.Start < currentMatch.End && 
+                                   hw.End > currentMatch.End)
+                          .OrderByDescending(o => o.Layer)
+                          .FirstOrDefault();
+        }
+        
+        private class HighlightedMatch
+        {
+            public int Id;
```
304NotModified: What is the Id here? Could you add some docs for it?

src/NLog/Targets/AnsiConsoleColorizer.cs
```diff
@@ -0,0 +1,197 @@
...
+            public ConsoleOutputColor BackgroundColor;
+        }
+        
+        internal static string ColorizeWord(string word, ConsoleOutputColor matchForegroundColor, ConsoleOutputColor matchBackgroundColor, 
+                                   ConsoleOutputColor nextForegroundColor, ConsoleOutputColor nextBackgroundColor)
+        {
+            var builder = new StringBuilder(5);
+
+            if (matchBackgroundColor != ConsoleOutputColor.NoChange)
+                builder.Append(AnsiConsoleColor.GetBackgroundColorEscapeCode((ConsoleColor)matchBackgroundColor));
```
304NotModified: this construction is duplicate, can we add a helper for it?

src/NLog/Targets/ConsoleColorizer.cs
```diff
@@ -0,0 +1,261 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Text;
+
+    /// <summary>
+    /// Colorizes console output using Console.ForegroundColor and Console.BackgroundColor.
+    /// </summary>
+    internal class ConsoleColorizer
```
304NotModified: I like the extraction :)

but should ConsoleColorizer and AnsiConsoleColorizer not sharing a common interface?

src/NLog/Targets/ColoredConsoleTarget.cs
```diff
@@ -204,51 +176,38 @@ protected override void Write(LogEventInfo logEvent)
...
-                    message = message.Replace("\a", "\a\a");
-                    foreach (ConsoleWordHighlightingRule hl in this.WordHighlightingRules)
-                    {
-                        message = hl.ReplaceWithEscapeSequences(message);
-                    }
+            var consoleStream = this.ErrorStream ? Console.Error : Console.Out;
+            if (!PlatformDetector.IsUnix)
+                OutputUsingConsoleColors(consoleStream, logEvent, message);
+            else
+                OutputUsingAnsiEscapeCodes(consoleStream, logEvent, message);
```
304NotModified: I think we need a common interface instead of two separate methods.

304NotModified:
Back from holiday!

Reviewed the code. Looks good, added some notes.

304NotModified:
Please check: 

> C:\projects\nlog\src\NLog\Targets\ColoredConsoleTarget.cs(118,16): error CS0246: The type or namespace name 'Encoding' could not be found (are you missing a using directive or an assembly reference?)
Compilation failed.
    0 Warning(s)

jp7677:
Thanks a lot for your feedback and very cool that you see something in my approach. I'm getting my hands dirty doing some adjustments ;), I'll let you know when I'm pleased with my changes.

jp7677:
The compile error should already been fixed.

jp7677:
also using an interface for both colorizers..

src/NLog/Targets/AnsiConsoleColor.cs
jp7677: Well, a dictionary would indeed reduce cyclomatic complexity, but as a downside we would keep all codes in memory whereas typically just a few are used. I know it's not much, but also teh small bits count :). Furthermore the Microsoft devs are doing it that way (https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Console/Internal/AnsiLogConsole.cs). I would prefer to keep it like this.

src/NLog/Targets/AnsiConsoleColorizer.cs
jp7677: The colors are slightly different, I used the dark colors only as default to have max. compatibility.

src/NLog/Targets/AnsiConsoleColorizer.cs
jp7677: Thanks, I have changed that, I got the meaning wrong for the constr. param.

src/NLog/Targets/ColoredConsoleTarget.cs
jp7677: Done.

src/NLog/Targets/AnsiConsoleColor.cs
jp7677: Done.

src/NLog/Targets/AnsiConsoleColorizer.cs
jp7677: Done.

src/NLog/Targets/ConsoleColorizer.cs
jp7677: Done.

src/NLog/Targets/AnsiConsoleColorizer.cs
jp7677: Done. Hope that my notes are understandable...

src/NLog/Targets/AnsiConsoleColorizer.cs
jp7677: Done. (Though not sure if it reads easier :))

jp7677:
Please see also my last comment in https://github.com/NLog/NLog/issues/1486

jp7677:
I'm done for now :)

jp7677:
PS: Hope you had a nice vacation!

304NotModified:
Thanks, holiday was very nice! But as always also happy I'm  back :)

src/NLog/Targets/AnsiConsoleColor.cs
```diff
@@ -0,0 +1,136 @@
...
+    internal static class AnsiConsoleColor
+    {
+        internal static string GetForegroundColorEscapeCode(ConsoleColor color)
+        {
+            switch (color)
+            {
+                case ConsoleColor.Black:
+                    return "\x1B[30m";
+                case ConsoleColor.Blue:
+                    return "\x1B[94m";
```
BrennanConroy: One thing to note is the 90-97 codes are not standard, you can see here https://github.com/aspnet/Logging/commit/7457578257cfb1eb2940b26818798a017efc1921 that we recently fixed that issue

src/NLog/Targets/AnsiConsoleColor.cs
```diff
@@ -0,0 +1,136 @@
...
+                case ConsoleColor.DarkMagenta:
+                    return "\x1B[45m";
+                case ConsoleColor.DarkRed:
+                    return "\x1B[41m";
+                case ConsoleColor.DarkYellow:
+                    return "\x1B[43m";
+                case ConsoleColor.Gray:
+                    return "\x1B[47m";
+                case ConsoleColor.Green:
+                    return "\x1B[102m";
```
BrennanConroy: Similarly 100-107 isn't standard either

src/NLog/Targets/AnsiConsoleColor.cs
```diff
@@ -0,0 +1,136 @@
...
+                case ConsoleColor.Yellow:
+                    return "\x1B[103m";
+                default:
+                    return TerminalDefaultBackgroundColorEscapeCode;
+            }
+        }
+        
+        internal static string TerminalDefaultBackgroundColorEscapeCode
+        {
+            get { return "\x1B[0m"; }
```
BrennanConroy: This should be 49 probably

src/NLog/Targets/AnsiConsoleColor.cs
304NotModified: Thanks for the info @BrennanConroy!

jp7677: Hi @BrennanConroy, thanks for stepping by indeed, very much appreciated!

src/NLog/Targets/AnsiConsoleColor.cs
jp7677: Fixed.

src/NLog/Targets/AnsiConsoleColor.cs
jp7677: Adjusted, I'm using bold as bright indicator now too.

src/NLog/Targets/AnsiConsoleColor.cs
jp7677: As far as I can see the bright background colors are not part of the original specifications. I guess I should just document that the bright colors might not be supported on older terminals.

304NotModified:
I will relabel this as performance. We will try to look at this after 4.4.

jp7677:
:+1:

304NotModified:
There are currently no problems in the console targets. 

Thanks for the PR, but this won't fix any open standing issue.

jp7677:
I guess this is the correct decision. You shouldn't include more code if it gives you no benefits, the PR is outdated and I had moved to SeriLog some time ago (that fits my target somewhat better, net core on Linux), so there is also no demand from my side anymore.
Thanks for the discussion.

Issue #1508, opened 2016-06-16
throwConfigExceptions doesn't throw when using invalid extension types/assemblies


jjbott:
**Type** (choose one): Bug

**NLog version**: 4.3.5

**Platform**: .Net 4.5

**Current NLog config** (xml or C#, if relevant)

```xml
  <nlog throwConfigExceptions="true">
    <extensions>
      <add assembly="Gelf4NLog.Target" />
    </extensions>
    <targets>
      <target name="graylog" xsi:type="graylog" />
    </targets>
    <rules>
      <logger name="*" minlevel="Info" writeTo="graylog" />
    </rules>
  </nlog>

```

**Note: In the above scenario, due to other problems, the assembly Gelf4NLog.Target can not be loaded (details to follow)**

**Expected Result**: Since Gelf4NLog.Target can not be loaded, NLog configuration fails. Since ```throwConfigExceptions``` is set to true, the app should fail with an exception.

**Current Result**: No exceptions are thrown. The only hint that NLog had problems is in its internal log. In my case the internal log looks like:

```
2016-06-15 14:06:58.3440 Error Failed to add type 'Gelf4NLog.Target.GelfMessage'. Exception: System.IO.FileLoadException: Could not load file or assembly 'Newtonsoft.Json, Version=4.5.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)
File name: 'Newtonsoft.Json, Version=4.5.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed'
   at System.ModuleHandle.ResolveType(RuntimeModule module, Int32 typeToken, IntPtr* typeInstArgs, Int32 typeInstCount, IntPtr* methodInstArgs, Int32 methodInstCount, ObjectHandleOnStack type)
   at System.ModuleHandle.ResolveTypeHandleInternal(RuntimeModule module, Int32 typeToken, RuntimeTypeHandle[] typeInstantiationContext, RuntimeTypeHandle[] methodInstantiationContext)
   at System.Reflection.RuntimeModule.ResolveType(Int32 metadataToken, Type[] genericTypeArguments, Type[] genericMethodArguments)
[...]

PR #1517, opened 2016-06-23
Fix issue #1508. Throw exception when registering invalid extension assembly/type.

Jeinhaus:
See issue #1508.
The problem seemed to be that there were multiple try-catches around the methods loading the assemblies, so that the exception did not bubble out of `ParseExtensionsElement(NLogXmlElement extensionsElement, string baseDirectory)`.
I removed the try-catches in there, so that a NLogConfigurationException is created in the try-catch in the toplevel method `XmlLoggingConfiguration.Initialize(XmlReader reader, string fileName, bool ignoreErrors)`.

Let me know if this fixes the issue and if you see any problems with removing the inner try-catches.

### DIFF
--- a/src/NLog/Config/XmlLoggingConfiguration.cs
+++ b/src/NLog/Config/XmlLoggingConfiguration.cs
@@ -909,24 +909,62 @@ namespace NLog.Config
                 string type = StripOptionalNamespacePrefix(addElement.GetOptionalAttribute("type", null));
                 if (type != null)
                 {
-                    this.ConfigurationItemFactory.RegisterType(Type.GetType(type, true), prefix);
+                    try
+                    {
+                        this.ConfigurationItemFactory.RegisterType(Type.GetType(type, true), prefix);
+                    }
+                    catch (Exception exception)
+                    {
+                        if (exception.MustBeRethrownImmediately())
+                        {
+                            throw;
+                        }
+
+                        InternalLogger.Error(exception, "Error loading extensions.");
+                        NLogConfigurationException configException =
+                            new NLogConfigurationException("Error loading extensions: " + type, exception);
+
+                        if (configException.MustBeRethrown())
+                        {
+                            throw configException;
+                        }
+                    }
                 }
 
                 string assemblyFile = addElement.GetOptionalAttribute("assemblyFile", null);
                 if (assemblyFile != null)
                 {
+                    try
+                    {
 #if SILVERLIGHT && !WINDOWS_PHONE
                     var si = Application.GetResourceStream(new Uri(assemblyFile, UriKind.Relative));
                     var assemblyPart = new AssemblyPart();
                     Assembly asm = assemblyPart.Load(si.Stream);
 #else
+                        string fullFileName = Path.Combine(baseDirectory, assemblyFile);
+                        InternalLogger.Info("Loading assembly file: {0}", fullFileName);
 
-                    string fullFileName = Path.Combine(baseDirectory, assemblyFile);
-                    InternalLogger.Info("Loading assembly file: {0}", fullFileName);
-
-                    Assembly asm = Assembly.LoadFrom(fullFileName);
+                        Assembly asm = Assembly.LoadFrom(fullFileName);
 #endif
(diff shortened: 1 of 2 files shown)


Jeinhaus:
Mhhh. I don't get why that one test is not passing with appveyor.
And it seems like travis-ci has a problem.

src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -406,7 +406,7 @@ private void Initialize(XmlReader reader, string fileName, bool ignoreErrors)
 
                 if (!ignoreErrors)
                 {
-                    if (exception.MustBeRethrown())
+                    if (configurationException.MustBeRethrown())
```
304NotModified: I think this is a good change

src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -915,70 +915,36 @@ private void ParseExtensionsElement(NLogXmlElement extensionsElement, string bas
...
-                        InternalLogger.Info("Loading assembly file: {0}", fullFileName);
+                    string fullFileName = Path.Combine(baseDirectory, assemblyFile);
+                    InternalLogger.Info("Loading assembly file: {0}", fullFileName);
 
-                        Assembly asm = Assembly.LoadFrom(fullFileName);
+                    Assembly asm = Assembly.LoadFrom(fullFileName);
 #endif
-                        this.ConfigurationItemFactory.RegisterItemsFromAssembly(asm, prefix);
-                    }
-                    catch (Exception exception)
```
304NotModified: I think we still need the catch, but with the same logic as the prev change (with my comment)

src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -915,70 +915,36 @@ private void ParseExtensionsElement(NLogXmlElement extensionsElement, string bas
...
+                    var assemblyPart = new AssemblyPart();
+                    Assembly asm = assemblyPart.Load(si.Stream);
 #else
-                        Assembly asm = Assembly.Load(assemblyName);
+                    Assembly asm = Assembly.Load(assemblyName);
 #endif
 
-                        this.ConfigurationItemFactory.RegisterItemsFromAssembly(asm, prefix);
-                    }
-                    catch (Exception exception)
```
304NotModified: Idem, catch should be there

tests/NLog.UnitTests/Config/ExtensionTests.cs
```diff
@@ -281,6 +281,30 @@ public void ExtensionTest_extensions_not_top_and_used()
         }
 
         [Fact]
+        public void ExtensionShouldThrowNLogConfiguratonExceptionWhenRegisteringInvalidType()
+        {
+            var configXml = @"
```
304NotModified: You need to set InternalLogger.throwsconfigexceptions on true

(without typo, on mobile now ;)

304NotModified: Scratch that :)  read to fast on tiny screen

304NotModified:
Thanks for the PR! 

On holiday now but i checked it in short, see comments.  

One unit test is broken, but I'm sure the 2 catches should be restored. 

I restarted travis-ci.org, it was bugged

Jeinhaus:
Alright, I readded the try-catches. You were right. If they were omitted the node wasn't correctly parsed.
Now it should continue parsing the <add .../> elements, even if one doesn't exist (if ThrowConfigExtensions==false).

304NotModified:
could you check why InvalidXMLConfiguration_DoesNotThrowErrorWhen_ThrowExceptionFlagIsNotSet fails?

Jeinhaus:
Mh, I don't get why that test fails on Mono. It passes on my machine. Do you have any recommendations how to debug this?

304NotModified:
I have fixed the unit test and merged the code in https://github.com/NLog/NLog/pull/1553

Thanks for the PR!

304NotModified:
It's online!

Jeinhaus:
Yey, my first PR got merged! 
Thank you for your help in figuring this one out.

PR #1547, opened 2016-07-19
LogReceiverWebServiceTarget is leaking communication channels

MartinTherriault:
After sending log messages over the communication channel, a completion event is raised to close the communication channel. In this event handler the "sender" parameter is cast to the wrong type hence the channel never get closed until the "receive timeout" expire (from the WCF binding, default to 10 minutes)


304NotModified:
Thanks for the PR!

304NotModified:
It's online!

Issue #1542, opened 2016-07-14
Can't update EventLog's Source property


ArchonMegalon:
[KAV.Tools.Logging.1.0.0.21.zip](https://github.com/NLog/NLog/files/364119/KAV.Tools.Logging.1.0.0.21.zip)

Hi! 

My first report, so please be gentle... ;)

The important thing is happening in the LogInternal function of my custom (from NLog.Logger inherited) Loggerclass:

**logEvent.Properties["MySource"] = useSource;** <-- this is set **per call**

The second time the function is called (and useSource is different), of course the Properties are set, but the EventLogTargets SourceAttribute is not updated! It has always the value from the **first call**.

EventID is updated.

**Type**: Bug
**NLog version**: 4.3.5

**Platform**: .Net 4.5 (don't know what else - haven't tried it)

**Current NLog config**:
```xml
<?xml version="1.0" encoding="utf-8"?>
<nlog throwExceptions="true" xmlns="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <!--   See http://nlog-project.org/wiki/Configuration_file for information on customizing logging rules and outputs. -->
  <!--${shortdate}-->
  <targets>
     <target xsi:type="EventLog" name="ELNoFileName" machineName="."
            source="${event-properties:item=MySource}" log="Application"
            eventId="${event-properties:EventID}"
            layout="ELNoFileName (${event-properties:item=MySource}): ${message}${newline}${newline}Pid: ${processid}${newline}Call site: ${callsite:className=true:methodName=true}${newline}Logger: ${logger}${newline}${exception:format=tostring}"/>
  </targets>
  <rules>
    <logger name="*" minlevel="Trace" writeTo="ELNoFileName">
      <filters>
        <when condition="length('${event-properties:item=FileName}') > 0" action="Ignore" />
      </filters>
    </logger>
  </rules>
</nlog>
```
```C#
public partial class LoggerBase : NLog.Logger
{
 private void LogInternal(LogLevel level, object message = null, Exception exception = null, int? eventId = null)
        {
            var logEvent = new LogEventInfo
            {
                Message = message?.ToString() ?? String.Empty,
                Exception = exception,
                Level = level,
                LoggerName = this.Name

            };


            var stack = new StackTrace(true);
            int userFrameIndex = -1;
            MethodBase userMethod = null;

            for (int i = 0; i < stack.FrameCount; ++i)
            {
[...]

PR #1548, opened 2016-07-20
Bugfix: Can't update EventLog's Source property

Page-Not-Found:
fixes https://github.com/NLog/NLog/issues/1542

### DIFF
--- a/src/NLog/Targets/EventLogTarget.cs
+++ b/src/NLog/Targets/EventLogTarget.cs
@@ -349,9 +349,15 @@ namespace NLog.Targets
         /// <returns></returns>
         private EventLog GetEventLog(LogEventInfo logEvent)
         {
-            if (eventLogInstance?.Source == this.Source.Render(logEvent) && eventLogInstance?.Log == this.Log && eventLogInstance?.MachineName == this.MachineName)
-                return eventLogInstance ?? (eventLogInstance = new EventLog(this.Log, this.MachineName, this.Source.Render(logEvent)));
-            return eventLogInstance = new EventLog(this.Log, this.MachineName, this.Source.Render(logEvent));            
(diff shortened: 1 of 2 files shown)


src/NLog/Targets/EventLogTarget.cs
```diff
@@ -349,7 +349,9 @@ internal string GetFixedSource()
         /// <returns></returns>
         private EventLog GetEventLog(LogEventInfo logEvent)
         {
-            return eventLogInstance ?? (eventLogInstance = new EventLog(this.Log, this.MachineName, this.Source.Render(logEvent)));
+            if (eventLogInstance?.Source == this.Source.Render(logEvent))
```
304NotModified: In worst case we're rendering the `source` two times?

Also there is an (implicit) two times check on `null` and we need to check also log and machine?

304NotModified:
Thanks! Added a note. Not sure why appveyor complains, looks unrelated

src/NLog/Targets/EventLogTarget.cs
Page-Not-Found: Do you mean something like this?
```c#
if (eventLogInstance?.Source == this.Source.Render(logEvent) && eventLogInstance?.Log == this.Log && eventLogInstance?.MachineName == this.MachineName)
                return eventLogInstance ?? (eventLogInstance = new EventLog(this.Log, this.MachineName, this.Source.Render(logEvent)));
            return eventLogInstance = new EventLog(this.Log, this.MachineName, this.Source.Render(logEvent));
```

304NotModified: Think it fixes the issue (hard to read on mobile) but Its still double checking null (`?.` and `==`), which is a bit sloppy

Issue #1543, opened 2016-07-14
Nested layouts (Json layout)


luigiberrettini:
**Type**:
Question

**NLog version**
4.3.5

**Platform**: .Net 4.5

I wonder if there is a way to configure NLog to have nested layouts.
My use case is having a key under which I nest a JSON layout, with the key not wrapped in quotes:

```xml
<nlog>
  <targets>
    <target name='file' type='File' fileName='log.txt'>
      <layout type='SimpleLayout' text="myAmazingText: ">
        <layout type='JsonLayout'>
          <attribute name='time' layout='${longdate}' />
          <attribute name='level' layout='${level:upperCase=true}'/>
          <attribute name='nested' encode='false'  >
            <layout type='JsonLayout'>
              <attribute name='message' layout='${message}' />
              <attribute name='exception' layout='${exception}' />
            </layout>
          </attribute>
        </layout>
      </layout>
    </target>
  </targets>
  <rules>
  </rules>
</nlog>
```

```json
myAmazingText: { "time": "2016-10-30 13:30:55.0000", "level": "INFO", "nested": { "message": "this is message", "exception": "test" } }
```

If this is not currently possible where do you think it would be most useful to add the feature?
JsonLayout.cs adding a public string Name { get; set; } property?

Thanks!
Luigi

304NotModified:
yes this is possible already,

you need encode=false

see https://github.com/nlog/nlog/wiki/JsonLayout#advanced-examples

luigiberrettini:
Sorry @304NotModified but with only one attribute and encoded set to false the attribute will be wrapped by curly braces:
```json
{ myAmazingText: { "time": "2016-10-30 13:30:55.0000", "level": "INFO", "nested": { "message": "this is message", "exception": "test" } } }
```

What I am looking for is instead this:
```json
myAmazingText: { "time": "2016-10-30 13:30:55.0000", "level": "INFO", "nested": { "message": "this is message", "exception": "test" } }
```

304NotModified:
> myAmazingText: { "time": "2016-10-30 13:30:55.0000", "level": "INFO", "nested": { "message": "this is message", "exception": "test" } }

but that isn't valid JSON?  

See also http://jsonlint.com/

luigiberrettini:
On the [wiki page](https://github.com/nlog/nlog/wiki/JsonLayout): "Currently the layout will always create an non-nested object with properties"

[...]

PR #1551, opened 2016-07-22
Add CompoundLayout

luigiberrettini:
As per issue #1543 I added a CompoundLayout Layout to make it possible to use multiple nested heterogeneous layouts and create complex log entries.

Take as an example the following configuration:
```xml
<nlog>
  <targets>
    <target name='file' type='File' fileName='log.txt'>
      <layout type='CompoundLayout'>
        <layout type='SimpleLayout' text="myAmazingText: " />
        <layout type='JsonLayout'>
          <attribute name='time' layout='${longdate}' />
          <attribute name='level' layout='${level:upperCase=true}'/>
          <attribute name='nested' encode='false'  >
            <layout type='JsonLayout'>
              <attribute name='exception' layout='${exception}' />
            </layout>
          </attribute>
        </layout>
      </layout>
    </target>
  </targets>
  <rules>
  </rules>
</nlog>
```

Log entries would be like this:
```json
myAmazingText: { "time": "2016-10-30 13:30:55.0000", "level": "INFO", "nested": { "message": "this is message", "exception": "test" } }
```

### DIFF
--- a/src/NLog/Config/XmlLoggingConfiguration.cs
+++ b/src/NLog/Config/XmlLoggingConfiguration.cs
@@ -530,7 +530,7 @@ namespace NLog.Config
             InternalLogger.LogToConsole = nlogElement.GetOptionalBooleanAttribute("internalLogToConsole", InternalLogger.LogToConsole);
             InternalLogger.LogToConsoleError = nlogElement.GetOptionalBooleanAttribute("internalLogToConsoleError", InternalLogger.LogToConsoleError);
             InternalLogger.LogFile = nlogElement.GetOptionalAttribute("internalLogFile", InternalLogger.LogFile);
-            
+
 #if !SILVERLIGHT && !__IOS__ && !__ANDROID__
             InternalLogger.LogToTrace = nlogElement.GetOptionalBooleanAttribute("internalLogToTrace", InternalLogger.LogToTrace);
 #endif
@@ -1092,7 +1092,9 @@ namespace NLog.Config
             if (elementType != null)
             {
                 IList propertyValue = (IList)propInfo.GetValue(o, null);
-                object arrayItem = FactoryHelper.CreateInstance(elementType);
+                object arrayItem = CreateArrayItemInstance(element, elementType);
+                if (arrayItem == null)
+                    return false;
                 this.ConfigureObjectFromAttributes(arrayItem, element, true);
                 this.ConfigureObjectFromElement(arrayItem, element);
                 propertyValue.Add(arrayItem);
@@ -1100,6 +1102,15 @@ namespace NLog.Config
             }
 
             return false;
+        }
+
+        private object CreateArrayItemInstance(NLogXmlElement element, Type elementType)
+        {
+            if (elementType != typeof(Layout))
+                return FactoryHelper.CreateInstance(elementType);
+
+            string layoutTypeName = StripOptionalNamespacePrefix(element.GetOptionalAttribute("type", null));
+            return layoutTypeName == null ? null : ConfigurationItemFactory.Layouts.CreateInstance(ExpandSimpleVariables(layoutTypeName));
         }
 
         private void ConfigureObjectFromAttributes(object targetObject, NLogXmlElement element, bool ignoreType)
--- a/src/NLog/Config/XmlLoggingConfiguration.cs
+++ b/src/NLog/Config/XmlLoggingConfiguration.cs
@@ -1092,9 +1092,12 @@ namespace NLog.Config
             if (elementType != null)
             {
                 IList propertyValue = (IList)propInfo.GetValue(o, null);
-                object arrayItem = CreateArrayItemInstance(element, elementType);
+
+                object arrayItem = TryCreateLayoutInstance(element, elementType);
+                // arrayItem is not a layout
                 if (arrayItem == null)
-                    return false;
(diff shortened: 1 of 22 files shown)


304NotModified:
\0/

src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -1102,6 +1104,15 @@ private bool AddArrayItemFromElement(object o, NLogXmlElement element)
             return false;
         }
 
+        private object CreateArrayItemInstance(NLogXmlElement element, Type elementType)
```
304NotModified: I think this is OK, but maybe nice to add some (XML) comments.

src/NLog/Layouts/CompoundLayout.cs
```diff
@@ -0,0 +1,77 @@
...
+
+
+namespace NLog.Layouts
+{
+    using Config;
+    using System.Collections.Generic;
+    using System.Text;
+
+    /// <summary>
+    /// A layout containig a nested layouts.
```
304NotModified: Proposal: "A layout containing one or more nested layouts."

src/NLog/Layouts/CompoundLayout.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CompoundLayout"/> class.
+        /// </summary>
+        public CompoundLayout()
+        {
+            Layouts = new List<Layout>();
+        }
+
+        /// <summary>
+        /// Gets the array of inner layouts.
```
304NotModified: Proposal: remove the word "array" as it's a implementation detail

src/NLog/Layouts/CompoundLayout.cs
```diff
@@ -0,0 +1,77 @@
...
+{
+    using Config;
+    using System.Collections.Generic;
+    using System.Text;
+
+    /// <summary>
+    /// A layout containig a nested layouts.
+    /// </summary>
+    [Layout("CompoundLayout")]
+    [ThreadAgnostic]
```
304NotModified: I don't think this is true? It's depended of the layouts used?

src/NLog/Layouts/CompoundLayout.cs
```diff
@@ -0,0 +1,77 @@
...
+    [AppDomainFixedOutput]
```
304NotModified: Same for this

src/NLog/Layouts/CompoundLayout.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// <param name="logEvent">The log event to be formatted.</param>
+        /// <returns>A string representation of the log event.</returns>
+        protected override string GetFormattedMessage(LogEventInfo logEvent)
+        {
+            var sb = new StringBuilder();
+            foreach (var layout in Layouts)
+                sb.Append(layout.Render(logEvent));
+            return sb.ToString();
+        }
+    }
```
304NotModified: I think we need also to implement `InitializeLayout` ?

304NotModified:
Looks good, added some notes!

src/NLog/Layouts/CompoundLayout.cs
luigiberrettini: Should be A layout containing nested/inner layouts (nested to it not layouts with nested layouts)

src/NLog/Layouts/CompoundLayout.cs
luigiberrettini: I initially implemented InitializeLayout but later removed it:

1. CsvLayout implements it but only to set properties that ar not Lists

2. JsonLayout contains a set of attributes and does not implement it... similarly CompoundLayout contains a set of Layouts

3. When I call render each Layout is initialized

I can implement both method as a loop over Layouts that calls them for each layout

src/NLog/Layouts/CompoundLayout.cs
luigiberrettini: You are right I only looked at two layouts who share these attributes.
Do you think it is enough to remove the two attributes or should I, in some way, programmatically determine if all layouts have an attribute and set it?

304NotModified: yes then can be removed

304NotModified:
Any idea if we could test this case:

image

(`CreateArrayItemInstance` returning `null`)
See https://codecov.io/gh/NLog/NLog/compare/698ca65c2ea8ac508a13f8d49d0e1c3c7a012c5c...f6d0d4a92c528cfd97927222847ac8b81ad0de87

304NotModified:
I think we almost ready to merge this. 

Can you add those on the wiki? Public is fine. Please add some examples and "introduced in NLog 4.3.6"

luigiberrettini:
We can add a test for layout whose XML configuration does not have a type: in this case I think an exception will be thrown.
This test does not depend on the CompoundLayout

luigiberrettini:
I just added the documentation.

https://github.com/NLog/NLog/wiki/Layouts
(I suppose the NuGet badge will auto-update to 4.3.6 when you release it)

https://github.com/NLog/NLog/wiki/CompoundLayout

304NotModified:
Great! Thanks!

304NotModified:
It's online!

PR #1580, opened 2016-08-04
Added option to not render empty literals on nested json objects

johnkors:
- Nested objects of null value were rendered as empty json literals. Added option to remove these.

### DIFF
--- a/src/NLog/Layouts/JsonLayout.cs
+++ b/src/NLog/Layouts/JsonLayout.cs
@@ -141,14 +141,5 @@ namespace NLog.Layouts
                 stringBuilder.Append(objectToAppend);
             }
         }
-
-        private static StringBuilder PrependIf<T>(bool condition, StringBuilder stringBuilder, T objectToPrepend)
-        {
-            if (condition)
-            {
-                return new StringBuilder(objectToPrepend + stringBuilder.ToString());
-            }
-            return stringBuilder;
-        }
     }
 }
--- a/src/NLog/Layouts/JsonLayout.cs
+++ b/src/NLog/Layouts/JsonLayout.cs
@@ -126,7 +126,7 @@ namespace NLog.Layouts
 
             var result = sb.ToString();
 
-            if (string.IsNullOrWhiteSpace(result) && !RenderEmptyObject)
+            if (string.IsNullOrEmpty(result.Trim()) && !RenderEmptyObject)
             {
                return string.Empty;
             }
--- a/src/NLog/Layouts/JsonLayout.cs
+++ b/src/NLog/Layouts/JsonLayout.cs
@@ -53,6 +53,7 @@ namespace NLog.Layouts
         public JsonLayout()
         {
             this.Attributes = new List<JsonAttribute>();
+            this.RenderEmptyLiteral = true;
         }
 
         /// <summary>
@@ -66,6 +67,11 @@ namespace NLog.Layouts
         /// Gets or sets the option to suppress the extra spaces in the output json
         /// </summary>
         public bool SuppressSpaces { get; set; }
+
+        /// <summary>
+        /// Gets or sets the option to render the empty object value {}
+        /// </summary>
+        public bool RenderEmptyLiteral { get; set; }
 
         /// <summary>
         /// Formats the log event as a JSON document for writing.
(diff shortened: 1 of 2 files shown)


304NotModified:
Thanks! We review it!

src/NLog/Layouts/JsonLayout.cs
```diff
@@ -68,6 +69,11 @@ public JsonLayout()
         public bool SuppressSpaces { get; set; }
 
         /// <summary>
+        /// Gets or sets the option to render the empty object value {}
+        /// </summary>
+        public bool RenderEmptyLiteral { get; set; }
```
304NotModified: I was thinking of 'RenderEmptyObject`? What do you think?

src/NLog/Layouts/JsonLayout.cs
```diff
@@ -53,6 +53,7 @@ public class JsonLayout : Layout
         public JsonLayout()
         {
             this.Attributes = new List<JsonAttribute>();
+            this.RenderEmptyLiteral = true;
```
304NotModified: I think it's wise to keep the behavior backwards-compatible (so `false` as default)

src/NLog/Layouts/JsonLayout.cs
```diff
@@ -119,7 +125,17 @@ protected override string GetFormattedMessage(LogEventInfo logEvent)
             AppendIf(!this.SuppressSpaces, sb, " ");
             sb.Append("}");
 
-            return sb.ToString();
+
+            var result = sb.ToString();
+
+            if (!RenderEmptyLiteral)
+            {
+                if (result.Replace(" ", "") == "{}")
```
304NotModified: This works, but it is a bit hacky ;)

I was thinking, postpone adding the `{` and `}` (eg. https://github.com/johnkors/NLog/blob/c2d5366d574c155b34591c96c31aa8c05c67730b/src/NLog/Layouts/JsonLayout.cs#L85) and add them if the object isn't empty.

src/NLog/Layouts/JsonLayout.cs
```diff
@@ -119,7 +125,17 @@ protected override string GetFormattedMessage(LogEventInfo logEvent)
...
+                    return null;
```
304NotModified: string.empty please :)

src/NLog/Layouts/JsonLayout.cs
johnkors: That's why i set it to true by default. The behaviour was to render empty literals by default, and I wanted an option of removing them, so one has to opt out by setting it to false.

src/NLog/Layouts/JsonLayout.cs
johnkors: Yeah, I agree :)

src/NLog/Layouts/JsonLayout.cs
johnkors: Yeah, better naming!

src/NLog/Layouts/JsonLayout.cs
```diff
@@ -116,10 +120,18 @@ protected override string GetFormattedMessage(LogEventInfo logEvent)
                 }
             }
 
-            AppendIf(!this.SuppressSpaces, sb, " ");
-            sb.Append("}");
+            var result = sb.ToString();
+
+            if (string.IsNullOrEmpty(result.Trim()) && !RenderEmptyObject)
```
304NotModified: It's maybe even better to keep a boolean value tracking the "empty state"? 

Less String operations are better (e.g. trim and string comparisons)

304NotModified:
One not, let me know what you think

src/NLog/Layouts/JsonLayout.cs
johnkors: Ah, okay! Yeah, not used to thinking performance up front :) Thanks. Will push an update.

304NotModified: Cool!

Well pretty important stuff to most of the users ;)

304NotModified:
perfect!

304NotModified:
Added on wiki https://github.com/NLog/NLog/wiki/JsonLayout

304NotModified:
4.3.7 is online: https://www.nuget.org/packages/NLog/4.3.7

PR #1583, opened 2016-08-05
Invalid Layout Should parse partly

304NotModified:
If we made a typo or forget to include some NLog extensions, there is a change that logging won't work at all! e.g. setting the Message of a filetarget on `{logger} {messagge}` (notice the typo), than the filetarget won't log anything!

This is unneeded and now we are skipping the invalid parts. The invalid parts are replace with an empty string.

If throwExceptions or throwConfigExceptions is `true`, there will be still an error thrown

---

This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1583)


304NotModified:
Maybe add in the future the option to to set the option for replacement: empty (current), space, error, exception (enable throwExceptions)

Issue #1625, opened 2016-09-06
(ExceptionLayoutRenderer) How to specify separator for format "data"?


FroggieFrog:
**Type**: Question
**NLog version**:  4.3.6
**Platform**: .Net 4.0 / .Net 4.5
**Current NLog config**
```xml
<nlog>
  <targets>
<target name="TraceLogger" xsi:type="File" layout="${pad:padding=-24:inner=${longdate}}||${pad:padding=-5:inner=${level:uppercase=true}}||${logger}||${message:whenEmpty=(no message)}${onexception:${newline}${exception:format=Message}${newline}${newline}${exception:format=Data}${newline}${newline}${exception:format=StackTrace:maxInnerExceptionLevel=3}${newline}${newline}${newline}${newline}${exception:format=ToString}" fileName="${var:logDirectory}/${var:traceLogFileName}.txt" archiveFileName="${var:archiveDirectory}/${var:traceLogFileName}.{####}.txt" maxArchiveFiles="7" footer="----------|${pad:padding=-24:inner=${longdate}}|Ende Programm" />
  </targets>
  <rules>
  </rules>
</nlog>
```

When rendering an excepetion with format=data, then the data is logged with a ";" as separator.
**How can I change this behaviour to logging each entry in exception.Data on a new line?**
Also the Data of an InnerException is not logged.

304NotModified:
You should set the options separator and maxInnerExceptionLevel.  
 
Eg `${exception:format=Data:maxInnerExceptionLevel=10:separator=|}`

See https://github.com/nlog/nlog/wiki/Exception-Layout-Renderer

FroggieFrog:
Thank you! I did not know "maxInnerExceptionLevel" defaults to 0.

**NLog version:** 4.3.8

But neither
``
${exception:format=Data:maxInnerExceptionLevel=10:seperator=${newline}}
``
nor
``
${exception:format=Data:maxInnerExceptionLevel=10:seperator=|}
``
is working.

The Data is always seperated by ";". I want them seperated by a new line.

304NotModified:
Yes sorry,

I see in the code that the separator for the data is fixed. 
You should create a subclass for that, e.g.:


```c#
    [LayoutRenderer("exception-custom")]
    [ThreadAgnostic]
    public class CustomExceptionLayoutRendrer : ExceptionLayoutRenderer
    {
        protected override void AppendMessage(System.Text.StringBuilder sb, Exception ex)
        {
[...]

PR #1628, opened 2016-09-07
allow different separator for Exception.Data

FroggieFrog:
this should fix issue #1625

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1628)


304NotModified:
Thanks for the PR but we need a fully backwards solution & unit tests. 

Fully backwards could be arranged by:
1. adding a new separator property
2. or adding a new boolean property 

I think 1 is the best, what do you think?

304NotModified:
@FroggieFrog your opinion please :)

FroggieFrog:
I added a new property and tried to add good unit tests.
One test is failing, but I don't know how to fix it. Maybe you can fix this?

304NotModified:
OK I will check it!

304NotModified:
You should use `\n` for newline and not `${newline}`. You can only use `${}` renderers if the type is `Layout`. We could make the property `Layout`, but that isn't consistent with Separator property and changing that one in not backwards-compatible.

Also, I think the name `SeparatorData` should be changed, as it sounds as "data for separator".

FroggieFrog:
I hope now everything is OK.

304NotModified:
\0/

304NotModified:
added to docs: https://github.com/NLog/NLog/wiki/Exception-layout-renderer

304NotModified:
NLog 4.3.9 is live! https://www.nuget.org/packages/NLog/4.3.9

PR #1639, opened 2016-09-13
Extra  tests + UrlEncode handle null

304NotModified:
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1639)


src/NLog/Internal/UrlHelper.cs
```diff
@@ -51,6 +51,8 @@ internal class UrlHelper
         /// <returns>Encoded url.</returns>
         internal static string UrlEncode(string str, bool spaceAsPlus)
         {
+            if (str == null) return string.Empty;
```
304NotModified: test

PR #1640, opened 2016-09-13
MDLC fix + unit test

vlardn:
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1640)

PR #1642, opened 2016-09-14
MDLC: fixing mutable dictionary issue (improvement)

vlardn:
Actually even better MDLC fix - see improvements in unit test. Improvement of https://github.com/NLog/NLog/pull/1640

This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1642)

### DIFF
--- a/src/NLog/MappedDiagnosticsLogicalContext.cs
+++ b/src/NLog/MappedDiagnosticsLogicalContext.cs
@@ -31,16 +31,12 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
-using System.Linq;
-using System.Threading;
 using NLog.Internal;
 
 namespace NLog
 {
 #if NET4_0 || NET4_5
-    using Config;
     using System;
-    using System.Collections.Concurrent;
     using System.Collections.Generic;
     using System.Runtime.Remoting.Messaging;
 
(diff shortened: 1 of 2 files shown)


src/NLog/MappedDiagnosticsLogicalContext.cs
```diff
@@ -111,10 +107,7 @@ public static string Get(string item, IFormatProvider formatProvider)
         public static object GetObject(string item)
         {
             object value;
-
-            if (!LogicalThreadDictionary.TryGetValue(item, out value))
-                value = null;
-
+            GetLogicalThreadDictionary().TryGetValue(item, out value);
```
304NotModified: do we have a guaranteed `null` return here?

304NotModified: no we don't, it could be `0` or `false`?

vlardn: This was not my but Resharper refactoring suggestion. 

Actually the value as null is guaranteed when function TryGet returns false:
> "...   otherwise [if the key is not found], the default value for the type of the value parameter [is returned]." - https://msdn.microsoft.com/en-us/library/bb347013(v=vs.110).aspx

As the default value for "object" is null - this refactoring seems valid.

304NotModified: yeah, was think it was `0` when it was `int`. Thanks for checking!

304NotModified:
NLog 4.3.9 is live! https://www.nuget.org/packages/NLog/4.3.9

Issue #1653, opened 2016-09-20
Set to the default value when given MaxMessageLength value is zero or negative


UgurAldanmaz:
If given value is zero or negative, then Exception will throw. I developed in this way but I have noticed that it is not an elegant solution.

Set to default would be more elegant then throwing an exception.

snakefoot:
Closing since the ArgumentException have not caused any complaints. I guess an alternative approach would be doing the option-validation inside `InitializeTarget()`

PR #1654, opened 2016-09-20
Remove exception for MaxMessageLength <= 0

UgurAldanmaz:
Set to the default value when given MaxMessageLength value is zero or
negative.

See #1653

---

This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1654)

### DIFF
--- a/src/NLog/Targets/EventLogTarget.cs
+++ b/src/NLog/Targets/EventLogTarget.cs
@@ -151,7 +151,7 @@ namespace NLog.Targets
             set
             {
                 if (value <= 0)
-                    throw new ArgumentException("MaxMessageLength cannot be zero or negative.");
+                    value = 16384;
 
                 this.maxMessageLength = value;
             }
@@ -363,7 +363,7 @@ namespace NLog.Targets
         /// <summary>
         /// (re-)create a event source, if it isn't there. Works only with fixed sourcenames.
         /// </summary>
-        /// <param name="fixedSource">sourcenaam. If source is not fixed (see <see cref="SimpleLayout.IsFixedText"/>, then pass <c>null</c> or emptystring.</param>
+        /// <param name="fixedSource">sourcename. If source is not fixed (see <see cref="SimpleLayout.IsFixedText"/>, then pass <c>null</c> or emptystring.</param>
         /// <param name="alwaysThrowError">always throw an Exception when there is an error</param>
         private void CreateEventSourceIfNeeded(string fixedSource, bool alwaysThrowError)
         {
--- a/tests/NLog.UnitTests/Targets/EventLogTargetTests.cs
+++ b/tests/NLog.UnitTests/Targets/EventLogTargetTests.cs
@@ -89,9 +89,11 @@ namespace NLog.UnitTests.Targets
         [Theory]
         [InlineData(0)]
         [InlineData(-1)]
-        public void ConfigurationShouldThrowException_WhenMaxMessageLengthIsNegativeOrZero(int maxMessageLength)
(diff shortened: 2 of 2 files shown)


UgurAldanmaz:
Wiki should be changed if PR merged.

304NotModified:
Is this consistent with other targets?

UgurAldanmaz:
I will check it out.

Thanks!

304NotModified:
I see it is common in Layout Renderers, but not in Targets:


|Location | Error |
| ---- | --- |
| LayoutRenderers\AllEventPropertiesLayoutRenderer.cs(94):   |                throw new ArgumentException("Invalid format: [key] placeholder is missing.");										|
| LayoutRenderers\AllEventPropertiesLayoutRenderer.cs(97):   |                throw new ArgumentException("Invalid format: [value] placeholder is missing.");									|
| LayoutRenderers\ProcessInfoLayoutRenderer.cs(80):          |     throw new ArgumentException("Property '" + this.propertyInfo + "' not found in System.Diagnostics.Process");					|
| LayoutRenderers\RegistryLayoutRenderer.cs(262):            |      throw new ArgumentException(string.Format("Key name is not supported. Root hive '{0}' not recognized.", hiveName));			|
| LayoutRenderers\RegistryLayoutRenderer.cs(275):            |       throw new ArgumentException("Only RegistryHive.LocalMachine and RegistryHive.CurrentUser are supported.", "hive");			|
| Targets\EventLogTarget.cs(154):                            |      throw new ArgumentException("MaxMessageLength cannot be zero or negative.");													|

304NotModified:
My proposal: 

MaxMessageLength of 0 or less will disable the truncate. (we need a log to the internal logger also I guess)

304NotModified:
let me know what you think!

UgurAldanmaz:
Detailed info. Thanks @304NotModified!

Is your proposal to disable truncate? So we need to choose between split message or discard.

I tested with zero and Split option. Result was infinite loop and no stack overflow exception. It writes empty message endless time.

So we can choose to discard the message but it would be misleading. User may suppose that logger does not works. Also if user set it to zero and give an overflow option Split or Truncate then we need to overwrite this option with Discard.

304NotModified:
my proposal is to change ` if (message.Length > this.MaxMessageLength)` https://github.com/UgurAldanmaz/NLog/blob/942fcd2cd7f0e09a3075c4794715d7f67d966755/src/NLog/Targets/EventLogTarget.cs#L266 to 
to  ` if (this.MaxMessageLength.HasValue && message.Length > this.MaxMessageLength)`

UgurAldanmaz:
So I should MaxMessageLength as nullable int?

304NotModified:
I prefer not, as it isn't backwardscompatible

UgurAldanmaz:
> I prefer not, as it isn't backwardscompatible

So your proposal is not valid because of `HasValue` property.

Maybe it should like

    if(this.MaxMessageLength > 0  && message.Length > this.MaxMessageLength)

304NotModified:
yes your're right!

UgurAldanmaz:
So I will change it today. Thanks @304NotModified!

UgurAldanmaz:
@304NotModified So we decided that continue to throw exception when set to negative or zero?

304NotModified:
https://github.com/NLog/NLog/pull/1654#issuecomment-251517522

UgurAldanmaz:
@304NotModified I need help for writing unit tests. Could you help me?

304NotModified:
Hey @UgurAldanmaz, still needed some he
lp on this?

304NotModified:
@UgurAldanmaz please open a new PR when you're ready! Thx!

Issue #1426, opened 2016-05-02
async="true" loses current thread identity


niemyjski:
I'm setting `async="true"` for perf reasons and I know that causes targets to be called on a secondary thread. I was wondering how I can keep the user identity (even the identity.name) of the thread when the log message was created. I see that the layout renders have access to this but my custom targets do not have identity information cached anywhere. I think if this isn't possible, we should at least pass the existing identity along to the target.

niemyjski:
Looks like the layout cache has this info but it's internal. It would be great if we could just expose this on the logger context!

304NotModified:
do you mean `${windows-identity}`?

niemyjski:
In our nlog target I'd like to set the current user if it hasn't already been set. I try to use different contexts but I wanted a fallback to `System.Threading.CurrentThread.Identity` https://github.com/exceptionless/Exceptionless.Net/blob/master/Source/Platforms/NLog/ExceptionlessClientExtensions.cs#L11

304NotModified:
If you write a custom layout renderer, it should work?

niemyjski:
My target looks like this:

```xml
 <targets async="true">
    <target name="file"
      xsi:type="File"
      layout="${longdate} ${threadid:padding=4} ${level:uppercase=true:padding=5} ${logger:shortName=true} ${message} ${exception:format=tostring}"
      fileName="${logDirectory}/log.txt"
      archiveFileName="${logDirectory}/log.{#}.txt"
      archiveEvery="Day"
      archiveNumbering="Rolling"
      maxArchiveFiles="7"
      concurrentWrites="true"
      createDirs="true"
      autoFlush="true" />
    
    <target name="exceptionless"  xsi:type="Exceptionless">
      <field name="host" layout="${machinename}" />
      <field name="identity" layout="${identity}" />
      <field name="process" layout="${processname}" />
      <field name="project" layout="${event-context:item=project}" />
    </target>
  </targets>
```

I first noticed that we weren't setting the user for some exceptions but our layout had rendered a custom object that had it.. Upon investigation this was from the current thread and the authenticated web api user.. So I was hoping to pass this user through our exceptionless code as a fallback but it's not on any nlog context just internal cached layouts

niemyjski:
I don't want to do that, I just want access to the current threads identity when the log was created.

304NotModified:
Maybe i'm missing it, but `${identity}` won't work well with async?

niemyjski:
[...]

PR #1660, opened 2016-09-24
Added failing unit test for #1426

niemyjski:
See https://github.com/NLog/NLog/issues/1426

This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1660)

### DIFF
--- a/tests/NLog.UnitTests/LayoutRenderers/IdentityTests.cs
+++ b/tests/NLog.UnitTests/LayoutRenderers/IdentityTests.cs
@@ -32,6 +32,7 @@
 // 
 
 using System;
+using System.Threading.Tasks;
 using NLog.Common;
 using NLog.Config;
 using NLog.Targets;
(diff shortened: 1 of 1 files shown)


304NotModified:
Thanks! Will look in depth of it

304NotModified:
Too bad AppVeyor is broken :(

304NotModified:
Related

- http://stackoverflow.com/questions/31346574/setting-thread-currentprincipal-with-async-await
- http://stackoverflow.com/questions/16552539/set-thread-currentprincipal-asynchronously

niemyjski:
@304NotModified This unit test might not be the best to show it but what I'm seeing is that the layouts and writes happen on the same thread but the async calls to the target happen on a different thread. Thus the information is lost. If the information that was used to calculate the layouts was in a cache this wouldn't be an issue.

304NotModified:
@niemyjski I would be nice if the unit test works without `async`. The would be easier to debug. Is that possible?

niemyjski:
You're trying to test async targets without any async? Granted this test doesn't really reproduce the environment I'm running in but does show off a symptom of the real issue.

304NotModified:
Yes with task.run on something like that? It's unclear to me if this is only an issue when using `async`

304NotModified:
> Yes with task.run on something like that? It's unclear to me if this is only an issue when using async

Bump @niemyjski

niemyjski:
Well, the reason this fails is because the thread can change because doing any async operation can change the synchronization context (please correct me if I'm wrong). The core problem is that layouts run on the same calling thread but targets can happen on a completely different background thread due to async targets..

304NotModified:
Well this issue happens because the first possible to record the info (`logger.Debug("test write");`) is on the wrong thread. 

I don't know if this would be possible to fix. We need an additional call on the correct thread I guess.

304NotModified:
the unit test is wrong IMO. This is not testing NLog's functionality, but it tests that the `Thread.CurrentPrincipal ` has been changed.

I will update the test and prove `${identity}` **is** working and also in `async` contexts.

304NotModified:
See https://github.com/niemyjski/NLog/pull/1

304NotModified:
superseded by https://github.com/NLog/NLog/pull/1692

PR #1668, opened 2016-10-02
Console Target Automatic Detect if console is available

snakefoot:
It is easier to just have one nlog.config file, than to have different versions depending on running as live-service or console-app.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1668)

### DIFF
--- a/src/NLog/Targets/ConsoleTarget.cs
+++ b/src/NLog/Targets/ConsoleTarget.cs
@@ -137,6 +137,11 @@ namespace NLog.Targets
             if (DetectConsoleAvailable)
             {
                 PauseLogging = !IsConsoleAvailable();
+                if (PauseLogging && LoggingConfiguration != null)
+                {
+                    foreach (var loggingRule in LoggingConfiguration.LoggingRules)
+                        loggingRule.Targets.Remove(this);
+                }
             }
             base.InitializeTarget();
             if (Header != null)
--- a/src/NLog/Targets/ConsoleTarget.cs
+++ b/src/NLog/Targets/ConsoleTarget.cs
@@ -101,7 +101,7 @@ namespace NLog.Targets
         /// <summary>
         /// Gets or sets a value indicating whether to auto-check if the console is available
         /// </summary>
-        [DefaultValue(false)]
+        [DefaultValue(true)]
         public bool DetectConsoleAvailable { get; set; }
 
         /// <summary>
@@ -113,6 +113,7 @@ namespace NLog.Targets
         public ConsoleTarget() : base()
         {
             PauseLogging = false;
+            DetectConsoleAvailable = true;
         }
 
         /// <summary>
@@ -170,18 +171,18 @@ namespace NLog.Targets
             {
                 if (!Environment.UserInteractive)
                 {
-                    InternalLogger.Info("Environment.UserInteractive = False. Console has been turned off. Disable DetectUserInteractive to skip detection.");
+                    InternalLogger.Info("Environment.UserInteractive = False. Console has been turned off. Disable DetectConsoleAvailable to skip detection.");
                     return false;
                 }
                 else if (Console.OpenStandardInput(1) == Stream.Null)
                 {
-                    InternalLogger.Info("Console.OpenStandardInput = Null. Console has been turned off. Disable DetectUserInteractive to skip detection.");
+                    InternalLogger.Info("Console.OpenStandardInput = Null. Console has been turned off. Disable DetectConsoleAvailable to skip detection.");
                     return false;
                 }
             }
             catch (Exception ex)
             {
-                InternalLogger.Warn(ex, "Failed to detect whether console is available. Console has been turned off. Disable DetectUserInteractive to skip detection.");
+                InternalLogger.Warn(ex, "Failed to detect whether console is available. Console has been turned off. Disable DetectConsoleAvailable to skip detection.");
                 return false;
             }
 #endif
--- a/src/NLog/Targets/ConsoleTarget.cs
+++ b/src/NLog/Targets/ConsoleTarget.cs
@@ -100,6 +100,8 @@ namespace NLog.Targets
 
         /// <summary>
         /// Gets or sets a value indicating whether to auto-check if the console is available
+        ///  - Disables console writing if Environment.UserInteractive = False (Windows Service)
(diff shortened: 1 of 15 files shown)


304NotModified:
Please tell us why we need this.

snakefoot:
I don't want the overhead from having a Console-Logger when running Live, but I like the console-logger when running in debugger.

304NotModified:
thanks! 

OK, and why won't we set it `true` by default? Not sure if it works all the cases?

snakefoot:
Well I have noticed that you like to keep things stable. So that is why I have chosen a conservative setting :)

I have read that MONO and others might not respond properly to Environment-namespace, but since they have been excluded from this feature, then I don't see an issue with having it enabled by default. 

Maybe rename it to SkipConsoleDetection with default value of false?

304NotModified:
`DetectConsoleAvailable` is still fine. You could set the default in the Ctor and mark the property with `[DefaultValue]`.

NB: I have to agree that defaulting all options to `false` would be nice, but that isn't the case in NLog right now.

snakefoot:
So I should change default value to true in constructor, and add DefaultValue-attribute?

304NotModified:
Yes please!

src/NLog/Targets/ConsoleTarget.cs
```diff
@@ -99,6 +99,12 @@ public Encoding Encoding
 #endif
 
         /// <summary>
+        /// Gets or sets a value indicating whether to auto-check if the console is available
+        /// </summary>
```
304NotModified: Maybe we should tell in the comments when this is ckecked and what the concenquenes are. What do you think?

src/NLog/Targets/ConsoleTarget.cs
```diff
@@ -127,6 +135,15 @@ public ConsoleTarget(string name) : this()
         protected override void InitializeTarget()
         {
             PauseLogging = false;
+            if (DetectConsoleAvailable)
+            {
+                PauseLogging = !IsConsoleAvailable();
+                if (PauseLogging && LoggingConfiguration != null)
+                {
+                    foreach (var loggingRule in LoggingConfiguration.LoggingRules)
+                        loggingRule.Targets.Remove(this);
```
304NotModified: I'm not sure if this is the right approach. Would not expect it would change my config.

src/NLog/Targets/ConsoleTarget.cs
```diff
@@ -147,6 +164,31 @@ protected override void CloseTarget()
             base.CloseTarget();
         }
 
+        private static bool IsConsoleAvailable()
```
304NotModified: Is this duplicate code for both targets? Could we extract this to one method?

snakefoot: Well I noticed the duplicate code-smell regarding the PauseLogging-property, so I thought you wanted the copy-paste-code for these two classes.

src/NLog/Targets/ConsoleTarget.cs
snakefoot: I can remove this. The most important part is to avoid performing the unnecessary layout-rendring.

src/NLog/Targets/ConsoleTarget.cs
snakefoot: No idea, but I can extend the summary a bit. Seldom I dare to use any properties in NLog without reading the code first :)

src/NLog/Targets/ConsoleTarget.cs
304NotModified: Please remove it - that would be more transparent

src/NLog/Targets/ConsoleTarget.cs
304NotModified: Fair point. I noticed that non trivial code has been duplicated. We always prefer reuse above duplication

src/NLog/Targets/ConsoleTarget.cs
snakefoot: Has been removed

src/NLog/Targets/ConsoleTarget.cs
snakefoot: Has been moved into separate class

304NotModified:
> Missing headers (copy them form other another file).
1 errors: 
 -------- 
-C:\projects\nlog\src/NLog\Targets\ConsoleTargetHelper.cs

304NotModified:
Thanks!!

304NotModified:
Reviewed 14 of 15 files at r3, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/NLog/Targets/ConsoleTarget.cs, line 167 at r2](https://reviewable.io:443/reviews/nlog/nlog/1668#-KT5ezkrmuvtOWwmpQ2t-r2-167:-KTLk5uzmYEMN4kN93hR:bd5l6mr) ([raw file](https://github.com/nlog/nlog/blob/a7bf47167b5181be5047bd436a56431cc1af7929/src/NLog/Targets/ConsoleTarget.cs#L167)):*
<details><summary><i>Previously, snakefoot (Rolf Kristensen) wrote…</i></summary>
> Has been moved into separate class
</details>
thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1668)*

304NotModified:
Will merge this if AppVeyor succeeds :)

Needs only docs on the Wiki I guess? Could you add that? Please add "introduced in 4.3.10"

snakefoot:
Have updated the Wiki-pages for Console Target and Color Console Target

efvincent:
Setting `DetectConsoleAvailable = true` as default caused an issue in our environment - we deploy microservices using Nomad, and we capture STDOUT for export to splunk. When we attempted to deploy a build that resolved 4.3.11 our logging in the Nomad environment stopped working.

I'm pretty sure it's `Environment.UserInteractive`, which is almost certainly false when running a headless Nomad microservice. I have verified that setting  `DetectConsoleAvailable = false`  in our NLog configuration has rectified our problem.

So... +1 to not deploying new features that change existing behavior ;)

304NotModified:
yes, the default has been changed in 4.4!

PR #1674, opened 2016-10-04
LogEventInfo - Hidden StringBuilder-object optimization to avoid LOH

snakefoot:
Implemented hidden optimization on LogEventInfo that allows a StringBuilder to ride through NLog without allocating string on Large-Object-Heap.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1674)

### DIFF
--- a/src/NLog/Internal/StringBuilderExt.cs
+++ b/src/NLog/Internal/StringBuilderExt.cs
@@ -59,5 +59,26 @@ namespace NLog.Internal
             builder.Append(Convert.ToString(o, formatProvider));
         }
 
+        /// <summary>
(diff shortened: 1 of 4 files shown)


304NotModified:
Hey thanks!

Could you provide some info in which benefits this will give and in which cases? Do you have also did a performance (cpu) test? 

This is making the code really more difficult to manage, so there should be big benefits.

snakefoot:
Have now been performance testing and not able to produce an improved result.  Seems this optimization only makes sense when using object-pools. Will merge it into my existing object-pooling PR.

PR #1669, opened 2016-10-02
AsyncTargetWrapper - Allow TimeToSleepBetweenBatches = 0

snakefoot:
Better than using 1ms that uses Timer-Threads even when nothing to process. Also when not having enabled high-resolution-timers, then 1 ms can become 15 ms.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1669)

PR #1684, opened 2016-10-05
AsyncTargetWrapper - Allow TimeToSleepBetweenBatches = 0 (Fix)

snakefoot:
Found a race-condition in #1669:
- Timer-Thread has taken full batchsize, and reschedules (Has emptied the queue)
- Timer-Thread quickly takes the lock again and sees the queue is still empty, but still holding lock
- Logger-Thread now enqueues LogEvent on queue, sees the queue is empty, but fails to schedule new timer as the lock is held.
- Timer-Thread releases lock, and decides not to schedule new timer.

Solution:
- When full-batch instant reschedule
- When small-batch throttled reschedule
- When empty-batch check RequestQueue whether to reschedule (new)

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1684)

### DIFF
--- a/src/NLog/Targets/Wrappers/AsyncTargetWrapper.cs
+++ b/src/NLog/Targets/Wrappers/AsyncTargetWrapper.cs
@@ -379,7 +379,7 @@ namespace NLog.Targets.Wrappers
             }
             finally
             {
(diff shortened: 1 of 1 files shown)


304NotModified:
not sure why Travis fails - restart

304NotModified:
Nice work!

Issue #1689, opened 2016-10-07
AppDomainWrapper memory leak


sungam3r:
**Type**: Bug

**NLog version**: 4.3.9

**Platform**: .Net 4.5

- What is the current result? - memory leak of AppDomainWrapper with EventHandler callbacks
- What is the expected result? - lack of (or less) AppDomainWrapper instances in managed heap

This causes memory leak:

1.  subscription 
    appDomain.ProcessExit += OnProcessExit;
    appDomain.DomainUnload += OnDomainUnload;
    in AppDomainWrapper ctor (there is no unsubscription at all)

2. Instance created each time  public static AppDomainWrapper CurrentDomain { get { return new AppDomainWrapper(AppDomain.CurrentDomain); } } method called

3. Example of transient instance creation call in FilePathLayout:
            if (_filePathKind == FilePathKind.Relative)
            {
                _baseDir = AppDomainWrapper.CurrentDomain.BaseDirectory;
            }
![memory](https://cloud.githubusercontent.com/assets/21261007/19188992/34710862-8c9e-11e6-923c-e82be18116a6.png)

304NotModified:
hey! Any how to fix this? Could you send a PR?

snakefoot:
One solution could be that LogFactory didn't cache the AppDomainWrapper but just returned a new instance every time.

Another solution could be that AppDomainWrapper unsubscribed itself from both handlers when one of the event-handlers fired.

snakefoot:
Yet another solution would be that LogFactory sets currentAppDomain = null when it is disposed. But the dispose-logic in LogFactory looks very tangled. Ex. the reload-timer is disposed every where.

304NotModified:
@sungam3r please your idea on this / https://github.com/NLog/NLog/pull/1690, thanks!

sungam3r:
Thank you for https://github.com/NLog/NLog/pull/1690
May I ask why not make
```
public interface IAppDomain
{
  event EventHandler ProcessExit;
  event EventHandler DomainUnload;
}
```
and simply delegate event handler subscriptions to underlying AppDomain object in AppDomainWrapper
```
  public event EventHandler ProcessExit
        {
            add
            {
                this.currentAppDomain.ProcessExit += value;
            }
            remove
            {
                this.currentAppDomain.ProcessExit -= value;
            }
        }

        public event EventHandler DomainUnload
        {
            add
            {
                this.currentAppDomain.DomainUnload += value;
            }
            remove
            {
[...]

PR #1690, opened 2016-10-08
Fix memory leak in AppDomainWrapper

snakefoot:
Unsubscribe from AppDomain when closing down

Possible solution for fixing  #1689

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1690)

### DIFF
--- a/src/NLog/Internal/Fakeables/AppDomainWrapper.cs
+++ b/src/NLog/Internal/Fakeables/AppDomainWrapper.cs
@@ -100,19 +100,53 @@ namespace NLog.Internal.Fakeables
         /// <summary>
         /// Process exit event.
         /// </summary>
-        public event EventHandler ProcessExit
+        public event EventHandler<EventArgs> ProcessExit
         {
-            add { this.currentAppDomain.ProcessExit += value; }
-            remove { this.currentAppDomain.ProcessExit -= value; }
+            add
+            {
+                if (this.processExitEvent == null)
+                    this.currentAppDomain.ProcessExit += OnProcessExit;
+                this.processExitEvent += value;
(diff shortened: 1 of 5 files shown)


src/NLog/Internal/Fakeables/AppDomainWrapper.cs
```diff
@@ -102,23 +100,19 @@ public AppDomainWrapper(AppDomain appDomain)
         /// <summary>
         /// Process exit event.
         /// </summary>
-        public event EventHandler<EventArgs> ProcessExit;
+        public event EventHandler ProcessExit
```
304NotModified: Do you know of this is a breaking change? I think it is (binary breaking change)

304NotModified:
thanks!

src/NLog/Internal/Fakeables/AppDomainWrapper.cs
snakefoot: Don't know if it is breaking change, but AppDomainWrapper resides within NLog/Internals, so would be surprised if anyone used it.

Alternative solution would be to make AppDomainWrapper IDisposable and then ensure to Dispose()  everywhere.

304NotModified: Its public but that's mistake IMO. Maybe we should fix that also.

304NotModified: https://github.com/NLog/NLog/pull/1691

not sure what do to with this

304NotModified:
close/reopen to trigger AppVeyor

PR #1692, opened 2016-10-08
Test for async="true" loses current thread identity

### LINKED ISSUE #1426: async="true" loses current thread identity

304NotModified:
fixes #1426 and , see also https://github.com/NLog/NLog/pull/1660

---

This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1692)


304NotModified:
Update: this works locally, also when the thread had been changed.

PR #1693, opened 2016-10-09
Performance (allocation) improvement in Aysnc handling 

snakefoot:
AsyncRequestQueue - DequeueBatch should only allocate one array

Instead of creating growing List (without initial capacity) and then doing ToArray

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1693)


304NotModified:
Good catch! Thanks again!

PR #1695, opened 2016-10-09
Performance: Counter/ProcessId/ThreadId-LayoutRenderer allocations less memory

snakefoot:
These three renderers are using the invariant culture. - StringBuilder Append without allocation

This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1695)

### DIFF
--- a/src/NLog/Internal/StringBuilderExt.cs
+++ b/src/NLog/Internal/StringBuilderExt.cs
@@ -59,5 +59,69 @@ namespace NLog.Internal
             builder.Append(Convert.ToString(o, formatProvider));
         }
 
+        /// <summary>
+        /// Appends int without using culture, and most importantly without garbage
+        /// </summary>
+        /// <param name="builder"></param>
+        /// <param name="value">value to append</param>
+        public static void AppendInvariant(this StringBuilder builder, int value)
+        {
+            // Deal with negative numbers
+            if (value < 0)
+            {
+                builder.Append('-');
+                uint uint_value = uint.MaxValue - ((uint)value) + 1; //< This is to deal with Int32.MinValue
+                AppendInvariant(builder, uint_value);
+            }
+            else
+            {
+                AppendInvariant(builder, (uint)value);
+            }
+        }
+
+        /// <summary>
+        /// Appends uint without using culture, and most importantly without garbage
+        /// 
+        /// Credits Gavin Pugh  - http://www.gavpugh.com/2010/04/01/xnac-avoiding-garbage-when-working-with-stringbuilder/
+        /// </summary>
+        /// <param name="builder"></param>
+        /// <param name="value">value to append</param>
+        public static void AppendInvariant(this StringBuilder builder, uint value)
+        {
+            if (value == 0)
+            {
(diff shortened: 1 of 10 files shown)


src/NLog/Internal/StringBuilderExt.cs
```diff
@@ -59,5 +59,99 @@ public static void Append(this StringBuilder builder, object o, LogEventInfo log
...
+        {
+            if (value == 0)
+            {
+                builder.Append('0');
+                return;
+            }
+
+            // Calculate length of integer when written out
+            int length;
+            if (value >= 10000)
```
304NotModified: uhm, what about `value / 10` ?

tests/NLog.UnitTests/Internal/StringBuilderExtTests.cs
```diff
@@ -0,0 +1,83 @@
...
+        [InlineData(1234)]
+        [InlineData(12345)]
+        [InlineData(123456)]
+        [InlineData(1234567)]
+        [InlineData(12345678)]
+        [InlineData(123456789)]
+        [InlineData(1234567890)]
+        [InlineData(int.MaxValue)]
+        [InlineData(int.MinValue)]
+        void TestAppendInvariant(int input)
```
304NotModified: we're missing negative numbers?

snakefoot: They are inside the test-method. I subtract the value from 0.

src/NLog/Internal/StringBuilderExt.cs
snakefoot: Less-than should be faster than division, and requires fewer iterations when larger values.

snakefoot:
Changed the title from without garbage to without allocation

src/NLog/Internal/StringBuilderExt.cs
snakefoot: But I can revert, when testing then the profiler never picked up the ThreadIdLayoutRenderer whether using division-loop or if-less-than

304NotModified: That sounds like a premature optimization? Also we have now more less-than operations needed. My proposal is to replace the whole block with a single division (and some rounding up or down, dunno)

tests/NLog.UnitTests/Internal/StringBuilderExtTests.cs
304NotModified: missed that. Maybe rename the method?

304NotModified:
PS: I'm renaming the PRs so it is more suitable for the changelog.

tests/NLog.UnitTests/Internal/StringBuilderExtTests.cs
snakefoot: Have no special opinion.  Have a good method name?

304NotModified: TestAppendInvariant_positive_and_negative :P

PR #1698, opened 2016-10-09
remove redudant checks

304NotModified:
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1698)

### DIFF
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -1831,11 +1831,7 @@ namespace NLog.Targets
 #endif
                 try
                 {
-                    archiveFile = this.GetArchiveFileName(fileName, ev, upcomingWriteSize);
-                    if (!string.IsNullOrEmpty(archiveFile))
-                    {
-                        this.DoAutoArchive(archiveFile, ev);
-                    }
+                    this.DoAutoArchive(archiveFile, ev);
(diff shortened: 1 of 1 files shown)


src/NLog/Targets/FileTarget.cs
```diff
@@ -1831,11 +1831,7 @@ private void TryArchiveFile(string fileName, LogEventInfo ev, int upcomingWriteS
 #endif
                 try
                 {
-                    archiveFile = this.GetArchiveFileName(fileName, ev, upcomingWriteSize);
-                    if (!string.IsNullOrEmpty(archiveFile))
-                    {
-                        this.DoAutoArchive(archiveFile, ev);
-                    }
+                    this.DoAutoArchive(archiveFile, ev);
```
UgurAldanmaz: It seems ok. Because already created a file name at line 1815 and already checked if it is null or empty at line 1816. If file name is null, then already this line will not executed.

UgurAldanmaz: It would be nice if the variable name is `archiveFileName`.

304NotModified: well, it's a path I think? ;)

304NotModified:
Thanks for checking @UgurAldanmaz !

Issue #1038, opened 2015-11-13
Performance problem when thread number is big


emanov:
[NLog.config.txt](https://github.com/NLog/NLog/files/33788/NLog.config.txt)
[Processor.cs.txt](https://github.com/NLog/NLog/files/33789/Processor.cs.txt)
[Program.cs.txt](https://github.com/NLog/NLog/files/33790/Program.cs.txt)

Hello!

I'm developing the app with big number of the long running threads (900-2000). When thread number is big the program has high cpu usage. After profiling the program I found that NLog eats 30% of all CPU usage in the method WriteAsyncLogEvents. The Blocking GC is high too.

When I have 10 process with 100 threads there is no any problem

So I write simple test for NLog, where all threads process one collection. Processing of each item do 4 call to Logger.Info() with intermediate pause about 80ms. Running this test with 900 threads leads to 30% CPU usage on my Corei5 processor. But when I start 9 processes with 100 threads per process (total is 900 too) the Cpu usage less then 2-3%.

So, I'm trying to find the way to have 900 threads and low cpu usage.

304NotModified:
Thanks for the test program!

We will try to look at this, this weekend, but it's a bit busy, so it could be also next week.

304NotModified:
Will be next week, as my laptop is really slow after a Windows update :(

emanov:
ok. will wait!

emanov:
Hello!
I see this on my core i5
![nlog900](https://cloud.githubusercontent.com/assets/710570/11194707/b14e528a-8cbe-11e5-98f4-db91d10d6999.png)

Nlog 4.1.2
  <package id="NLog" version="4.1.2" targetFramework="net45" />

Will try 4.2

do you use NLog.config with output to file?

emanov:
The same with 4.2

304NotModified:
> do you use NLog.config with output to the file? Looks like you do test without output or with async-wrapper with default drop-mode.

I test with your program, no line changed! Only difference that NLog is directly references instead of nuget package. I don't think nuget packages could lead to more CPU ;)

I have a SSD, i7, but's tested on a laptop. (which has already performance issues)

304NotModified:
have you tried changing the values of keepFileOpen  / concurrentWrites?

emanov:
I'm using SSD too.
For the test I've used

keepFileOpen="true" concurrentWrites="false"
(it is in the file attached to the first post)

[...]

PR #1700, opened 2016-10-09
Improved concurrency when multiple Logger threads are writing to async Target

snakefoot:
When writing to an async-target, then the Logger is forced to acquire Target.SyncObject-lock and hold it while preparing its LogEventInfo-object for the async-queue. Other Logger-threads writing to the same Target has to wait for this.

Now multiple Logger-threads can post directly to the async-queue without need to aquire and hold the Target.SyncRoot-lock. This means that the lock-held-window is much smaller and thus improves concurrency.

---

This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1700)

### DIFF
--- a/src/NLog/LogEventInfo.cs
+++ b/src/NLog/LogEventInfo.cs
@@ -431,7 +431,7 @@ namespace NLog
         {
             lock (this.layoutCacheLock)
             {
-                if (this.layoutCache == null)
+                if (this.layoutCache == null || this.layoutCache.Count == 0)
                 {
                     value = null;
                     return false;
--- a/src/NLog/Targets/Target.cs
+++ b/src/NLog/Targets/Target.cs
@@ -75,7 +75,21 @@ namespace NLog.Targets
         /// <summary>
         /// Gets a value indicating whether the target has been initialized.
         /// </summary>
-        protected bool IsInitialized { get; private set; }
+        protected bool IsInitialized
+        {
+            get
+            {
+                if (this.isInitialized)
+                    return true;    // Initialization has completed
+
+                // Lets wait for initialization to complete, and then check again
+                lock (this.SyncRoot)
+                {
+                    return this.isInitialized;
+                }
+            }
+        }
+        private volatile bool isInitialized;
 
         /// <summary>
         /// Get all used layouts in this target.
@@ -128,29 +142,35 @@ namespace NLog.Targets
                 throw new ArgumentNullException("asyncContinuation");
             }
 
-            lock (this.SyncRoot)
+            if (!this.IsInitialized)
             {
-                if (!this.IsInitialized)
-                {
-                    asyncContinuation(null);
-                    return;
-                }
+                asyncContinuation(null);
+                return;
+            }
 
-                asyncContinuation = AsyncHelpers.PreventMultipleCalls(asyncContinuation);
+            asyncContinuation = AsyncHelpers.PreventMultipleCalls(asyncContinuation);
 
-                try
+            try
+            {
+                lock (this.SyncRoot)
                 {
+                    if (!this.IsInitialized)
+                    {
+                        // In case target was Closed
+                        asyncContinuation(null);
+                        return;
+                    }
                     this.FlushAsync(asyncContinuation);
                 }
-                catch (Exception exception)
+            }
+            catch (Exception exception)
+            {
+                if (exception.MustBeRethrown())
                 {
-                    if (exception.MustBeRethrown())
-                    {
-                        throw;
-                    }
-
-                    asyncContinuation(exception);
+                    throw;
                 }
+
+                asyncContinuation(exception);
             }
         }
 
@@ -163,16 +183,14 @@ namespace NLog.Targets
         /// </param>
         public void PrecalculateVolatileLayouts(LogEventInfo logEvent)
         {
-            lock (this.SyncRoot)
+            if (this.IsInitialized)
             {
-                if (this.IsInitialized)
+                var currentLayouts = this.allLayouts;
+                if (currentLayouts != null)
                 {
-                    if (this.allLayouts != null)
+                    foreach (Layout l in currentLayouts)
                     {
-                        foreach (Layout l in this.allLayouts)
-                        {
-                            l.Precalculate(logEvent);
-                        }
+                        l.Precalculate(logEvent);
                     }
                 }
             }
@@ -201,35 +219,42 @@ namespace NLog.Targets
         /// <param name="logEvent">Log event to write.</param>
         public void WriteAsyncLogEvent(AsyncLogEventInfo logEvent)
         {
-            lock (this.SyncRoot)
+            if (!this.IsInitialized)
             {
-                if (!this.IsInitialized)
-                {
-                    logEvent.Continuation(null);
-                    return;
-                }
+                logEvent.Continuation(null);
+                return;
+            }
 
-                if (this.initializeException != null)
-                {
-                    logEvent.Continuation(this.CreateInitException());
-                    return;
-                }
+            if (this.initializeException != null)
+            {
+                logEvent.Continuation(this.CreateInitException());
+                return;
+            }
 
-                var wrappedContinuation = AsyncHelpers.PreventMultipleCalls(logEvent.Continuation);
+            var wrappedContinuation = AsyncHelpers.PreventMultipleCalls(logEvent.Continuation);
+            var wrappedLogEvent = logEvent.LogEvent.WithContinuation(wrappedContinuation);
 
-                try
+            try
+            {
+                lock (this.SyncRoot)
                 {
-                    this.Write(logEvent.LogEvent.WithContinuation(wrappedContinuation));
-                }
-                catch (Exception exception)
-                {
-                    if (exception.MustBeRethrown())
+                    if (!this.IsInitialized)
                     {
-                        throw;
+                        // In case target was Closed
+                        logEvent.Continuation(null);
+                        return;
                     }
-
-                    wrappedContinuation(exception);
+                    this.Write(wrappedLogEvent);
                 }
+            }
+            catch (Exception exception)
+            {
+                if (exception.MustBeRethrown())
+                {
+                    throw;
+                }
+
+                wrappedContinuation(exception);
             }
         }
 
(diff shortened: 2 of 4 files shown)


snakefoot:
This is pretty much a re-implementation of #1369 coming from #1038

304NotModified:
changes without newline changes: https://github.com/NLog/NLog/pull/1700/files?w=1

304NotModified:
Reviewed 1 of 2 files at r1.
Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/NLog/Targets/Target.cs, line 155 at r1](https://reviewable.io:443/reviews/nlog/nlog/1700#-KTfLeycEA0M328fP5PW:-KTfLeydATKIPFbrBm1_:b-2z5vlj) ([raw file](https://github.com/nlog/nlog/blob/4cabe77710dc90c962076a170f65e89b33faf2c6/src/NLog/Targets/Target.cs#L155)):*
> ```Smalltalk
>             try
>             {
>                 lock (this.SyncRoot)
> ```

why we need this lock here, as it is already in `this.IsInitialized`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1700)*

snakefoot:
*[src/NLog/Targets/Target.cs, line 155 at r1](https://reviewable.io:443/reviews/nlog/nlog/1700#-KTfLeycEA0M328fP5PW:-KTfMIgo9g0MAW58ccyy:bw3owma) ([raw file](https://github.com/nlog/nlog/blob/4cabe77710dc90c962076a170f65e89b33faf2c6/src/NLog/Targets/Target.cs#L155)):*
> why we need this lock here, as it is already in `this.IsInitialized`?
</details>
IsInitialized only tries to take lock, when initialization has not yet completed. And the lock is then released.

I don't want to change the contract completely, and allow multiple threads inside Target.FlushAsync()

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1700)*

304NotModified:
Reviewed 1 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1700)*

304NotModified:
👍 👍

304NotModified:
test this with this tester: https://github.com/NLog/NLog-Performance-tests/tree/master/ManyThreadsTester/ManyThreadsTester (ec170987d83533cc03678cc8e755552e0a0e55f6). Compared master (639b40f803e3a3fbcd2de946a9d7a017181c7915) with pr/1700 (27335851f405a10a7c888b1055523e4edb232d55), but didn't noticed any significant changes. Any idea how to test this?

snakefoot:
I used #1038, where I change the config to be wrapped in async-target.

The goal for that "test" is to lower the CPU usage by minimizing lock congestion.

304NotModified:
> The goal for that "test" is to lower the CPU usage by minimizing lock congestion.

Agree, did test CPU a first (and after that the timing as that should not change to much). 

Could you post a nlog config?

snakefoot:
```
    <target name="file" type="AsyncWrapper" overflowAction="Block" queueLimit="50000" batchSize="25000" timeToSleepBetweenBatches="15">
      <target name="wrappedfile" type="File" fileName="${basedir}//${machinename}_TaskService ${shortdate}.log" layout="${longdate}|${level}|[${threadid}]||${message} ${exception:format=tostring}" keepFileOpen="true" concurrentWrites="false" />
    </target>

When using the above config then the cpu-usage dropes from 20-25% down to 10-15% with this PR.
```

304NotModified:
@snakefoot moved this open PRs to 4.3.11, as it gives me more time to test it and I have preference to release a version with only performance changes (so in case of issues we could go back)

snakefoot:
*[src/NLog/Targets/Wrappers/BufferingTargetWrapper.cs, line 180 at r2](https://reviewable.io:443/reviews/nlog/nlog/1700#-KVWYf7Z85YdYmbleGbn:-KVWYf7Z85YdYmbleGbo:b-f6hhx3) ([raw file](https://github.com/nlog/nlog/blob/078775498ac5c1a98d05016829bd901024862e70/src/NLog/Targets/Wrappers/BufferingTargetWrapper.cs#L180)):*
> ```Smalltalk
>         protected override void Write(AsyncLogEventInfo logEvent)
>         {
>             this.WrappedTarget.PrecalculateVolatileLayouts(logEvent.LogEvent);
> ```

Btw. any reason why this.MergeEventProperties(logEvent.LogEvent) is not applied here like in AsyncTargetWrapper?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1700)*

304NotModified:
Review status: 1 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/NLog/Targets/Wrappers/BufferingTargetWrapper.cs, line 180 at r2](https://reviewable.io:443/reviews/nlog/nlog/1700#-KVWYf7Z85YdYmbleGbn:-KVWk87ir3_Vjk1tdqzJ:bqhcfwj) ([raw file](https://github.com/nlog/nlog/blob/078775498ac5c1a98d05016829bd901024862e70/src/NLog/Targets/Wrappers/BufferingTargetWrapper.cs#L180)):*
<details><summary><i>Previously, snakefoot (Rolf Kristensen) wrote…</i></summary>
> Btw. any reason why this.MergeEventProperties(logEvent.LogEvent) is not applied here like in AsyncTargetWrapper?
</details>
I will check :)

update: it's called in in AsyncTargetWrapper.Write , just before PrecalculateVolatileLayouts()

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1700)*

304NotModified:
Review status: 1 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/NLog/Targets/Wrappers/BufferingTargetWrapper.cs, line 180 at r2](https://reviewable.io:443/reviews/nlog/nlog/1700#-KVWYf7Z85YdYmbleGbn:-KVWkzqAMe4Dc3teCN9x:b9tc1nj) ([raw file](https://github.com/nlog/nlog/blob/078775498ac5c1a98d05016829bd901024862e70/src/NLog/Targets/Wrappers/BufferingTargetWrapper.cs#L180)):*
> I will check :)
> 
> update: it's called in in AsyncTargetWrapper.Write , just before PrecalculateVolatileLayouts()
</details>
Ow misread your question. It's indeed missing in BufferTargetWrapper. I think it's a mistake

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1700)*

snakefoot:
*[src/NLog/Targets/Wrappers/BufferingTargetWrapper.cs, line 180 at r2](https://reviewable.io:443/reviews/nlog/nlog/1700#-KVWYf7Z85YdYmbleGbn:-KVWmpi64Nvoo_UTNsCg:bvv3vjv) ([raw file](https://github.com/nlog/nlog/blob/078775498ac5c1a98d05016829bd901024862e70/src/NLog/Targets/Wrappers/BufferingTargetWrapper.cs#L180)):*
> Ow misread your question. It's indeed missing in BufferTargetWrapper. I think it's a mistake
</details>
Any reason why AsyncTargetWrapper calls PrecalculateVolatileLayouts() directly instead of using WrappedTarget like BufferingTargetWrapper?

I would guess calling the WrappedTarget.PrecalculateVolatileLayouts() is the correct way.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1700)*

snakefoot:
Changed the logic so lock is held while handling initialization-errors or target-write-errors. This should not affect the happy-path, where lock-congestion remains reduced.

snakefoot:
Not holding the lock while doing layout-precalculation will suddenly exercise the Layout-classes in a new way. Can see some of them are not prepared for concurrent access. Will have to change it, so the lock is held while doing layout-stuff.

snakefoot:
Not sure if I should extend this PR with adding thread-safe-layout-class-attribute to the relevant classes. Or if it should be done with a different PR? By having the class-attribute then it will allow better performance for situations where non-thread-agnostic-layouts are used.

304NotModified:
> Not sure if I should extend this PR with adding thread-safe-layout-class-attribute to the relevant classes. Or if it should be done with a different PR?

I depends. I'm a bit lost to be honest ;)

But if it's a problem introduced in this PR, we should fix it in this PR.

snakefoot:
This PR tries to improve async-target-concurrency. But then I discovered the problem with having multiple threads inside the same Layout-Renderer-object. Some of the Layout-Renderer-objects are not thread safe.

I have now changed the logic so it only locks Target.SyncRoot if the Logger-thread has to precalculate (cache) the layout result before posting to the async-target.

If NLog could detect if the layout-renderer-object supported concurrent access then I could skip the locking of the Target.SyncRoot while pre-calculating the layouts.

304NotModified:
Reviewed 3 of 3 files at r3.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/NLog/Targets/Wrappers/AsyncTargetWrapper.cs, line 320 at r3](https://reviewable.io:443/reviews/nlog/nlog/1700#-KVvf3WJSWARLEt6UzO7:-KVvf3WJSWARLEt6UzO8:bmmdgni) ([raw file](https://github.com/nlog/nlog/blob/90562aa24192dec954fc39b487b13cc92e215d75/src/NLog/Targets/Wrappers/AsyncTargetWrapper.cs#L320)):*
> ```Smalltalk
>                 this.Write(logEvent);
>             }
>             catch (Exception exception)
> ```

Why do we need a try-catch in `WriteAsyncThreadSafe` but not in `Write`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1700)*

snakefoot:
*[src/NLog/Targets/Wrappers/AsyncTargetWrapper.cs, line 320 at r3](https://reviewable.io:443/reviews/nlog/nlog/1700#-KVvf3WJSWARLEt6UzO7:-KVvgurfKnQ0mGgWjMBo:b-grgzl4) ([raw file](https://github.com/nlog/nlog/blob/90562aa24192dec954fc39b487b13cc92e215d75/src/NLog/Targets/Wrappers/AsyncTargetWrapper.cs#L320)):*
> Why do we need a try-catch in `WriteAsyncThreadSafe` but not in `Write`?
</details>
Target.WriteAsyncThreadSafe() is responsible for doing exception-handling in a tread-safe manner.

AsyncTargetWrapper.WriteAsyncThreadSafe() just tries to replicate that but without holding Target.SyncRoot

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1700)*

PR #1719, opened 2016-10-22
Avoid (Internal)Logger-boxing and params-array-allocation on Exception

snakefoot:
No need to box value-types when trying to avoid params-array-allocation.

No need to create empty params-array when reporting Exception to Logger.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1719)


304NotModified:
LOL maybe we should consider to implement `ILogger` or merge both code bases

snakefoot:
Yes could have saved me one PR :). But now I learned the trick with using generics to optimize the params-array.

304NotModified:
But isn't it this case better to implement ILogger? Although we need singleton instead of static. Dunno if backwards comp.

snakefoot:
Since the InternalLogger is a static class, then it will be difficult for it to implement the ILogger-interface.

It could have a static property that implements ILogger, but the InternalLogger would still have to forward its method to this property.

304NotModified:
Could you add the related unit tests?

304NotModified:
Reviewed 5 of 5 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1719)*

304NotModified:
Reviewed 1 of 2 files at r2.
Review status: 6 of 7 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1719)*

snakefoot:
Have updated the UnitTests to cover the new interface-methods.

304NotModified:
Thanks! Why are some unit tests changed?

snakefoot:
For some reason when the last parameter was "null" (without type), then the compiler chose the interface-method taking params-array.

I also added the ${exception} to layout in the other tests to show that the Exception was actually included.

304NotModified:
Reviewed 1 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1719)*

304NotModified:
Thanks!

PR #1722, opened 2016-10-26
AsyncLogEventInfo getter-property-methods changed into readonly members

snakefoot:
Replaced syntax-sugar intance-getter-methods with direct readonly intance-member access.

AsyncLogEventInfo  is a public struct so this is a breaking change, even though none are using the setter-method.

Now it is a proper immutable struct :)

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1722)


304NotModified:
Changed your mind?

snakefoot:
Very small performance gain compared to just changing the parameters to have private setters.

Thus no need for breaking change .

PR #1730, opened 2016-10-29
Comparison in condition, prefer non-string values and try first

### LINKED ISSUE #1727: Using relational operators (>) with event-properties isn't working correctly
Is it it possible to setup a condition with a numeric comparison to an event property?

I want to track 'Activation' events in my application that take over some threshold time to complete.

The following will filter out all 'Activation' events.

```
<when condition="'${event-properties:item=Event}' == 'Activation'" action="Ignore"/>
```

Adding `and 100  > 10` breaks this filter as expected, and allows all the messages through.

```
<when condition="'${event-properties:item=Event}' == 'Activation' and  100 > 10" action="Ignore"/>
```

Changing this to `100 > 1000` and once again all 'Activation' events are filtered out.

```
<when condition="'${event-properties:item=Event}' == 'Activation' and  100 > 1000 " action="Ignore"/>
```

Then I tried adding my `Elapsed` event property.

```
<when condition="'${event-properties:item=Event}' == 'Activation' and  100 >  '${event-properties:item=Elapsed}'" action="Ignore"/>
```

But this still continues to log all Activation events, including those that completed in less then 100 ms.

Is this this an issue with data types? Is is my Elapsed value being compared to 100 as a string rather then a number? If so is there some way to data type conversions in a condition?

304NotModified:
Fixes  #1727

---

This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1730)

### DIFF
--- a/src/NLog/Conditions/ConditionRelationalExpression.cs
+++ b/src/NLog/Conditions/ConditionRelationalExpression.cs
@@ -35,6 +35,8 @@ namespace NLog.Conditions
 {
     using System;
     using System.Globalization;
+    using System.Collections.Generic;
+    using Common;
 
     /// <summary>
     /// Condition relational (<b>==</b>, <b>!=</b>, <b>&lt;</b>, <b>&lt;=</b>,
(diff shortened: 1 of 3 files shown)


304NotModified:
@snakefoot would you like to review my PR?

PR #1732, opened 2016-10-30
Handle duplicate attributes (error or using first occurence)

### LINKED ISSUE #1659: Logger configuration attributes issue with duplicate attributes
**Type**: Bug

**NLog version**: 4.3.9

**Platform**: Net 4.6 

**Current NLog config**: xml 

``` xml
<nlog throwExceptions="false" internalLogFile="nloglog.txt" internalLogLevel="Trace">
    <targets>
      <target name="Console" type="ColoredConsole" useDefaultRowHighlightingRules="true" layout="${longdate} | ${message}"/>
    </targets>
    <rules>
      <logger name="*" minlevel="Info" writeTo="Console" minLevel="Info"/>
    </rules>
  </nlog>

```

The logger declaration contains two attributes with names different only in case ("minlevel" and "minLevel")
- What is the current result?
  Logger does not log anything
  Logger writes logs or throws an exception
  Yes. Logger does not write even an internal log
- Are there any work arrounds?
  No
  Don't know
  https://github.com/Falconwatch/NLogAttributes

nazim9214:
Resolves #1659 .
1. Logger now can write logs when config has duplicate attributes (first occurence is taken)
2. If throwExceptions or throwConfigExceptions option is on - exception is thrown.

---

This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1732)

### DIFF
--- a/src/NLog/Config/NLogXmlElement.cs
+++ b/src/NLog/Config/NLogXmlElement.cs
@@ -222,20 +222,20 @@ namespace NLog.Config
         }
 
         /// <summary>
-        /// Try to get all errors occured during xml parsing.
-        /// A return value indicates whether there any error.
+        /// Returns all errors occured during xml parsing.
+        /// If there is no errors - returns null.
         /// </summary>
-        /// <param name="errors">Array of error messages</param>
-        public bool TryGetParsingErrors(out string[] errors)
+        public string GetParsingErrors()
         {
-            errors = GetParsingErrors().ToArray();
-            return errors.Any();
+            var errors = GetParsingErrorsInternal().ToArray();
+            if (errors.Length == 0) return null;
+            return string.Join(Environment.NewLine, errors);
         }
 
         /// <summary>
         /// Returns all parsing errors from current and all child elements.
         /// </summary>
-        private IEnumerable<string> GetParsingErrors()
+        private IEnumerable<string> GetParsingErrorsInternal()
         {
             foreach (var parsingError in _parsingErrors)
             {
@@ -244,7 +244,7 @@ namespace NLog.Config
 
             foreach (var childElement in this.Children)
             {
-                foreach (var parsingError in childElement.GetParsingErrors())
+                foreach (var parsingError in childElement.GetParsingErrorsInternal())
                 {
                     yield return parsingError;
                 }
--- a/src/NLog/Config/NLogXmlElement.cs
+++ b/src/NLog/Config/NLogXmlElement.cs
@@ -31,6 +31,9 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
+using System.Linq;
+using System.Text;
+
 namespace NLog.Config
 {
     using Internal;
@@ -75,6 +78,7 @@ namespace NLog.Config
         {
             this.AttributeValues = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
             this.Children = new List<NLogXmlElement>();
+            this._parsingErrors = new List<string>();
         }
 
         /// <summary>
@@ -100,7 +104,7 @@ namespace NLog.Config
         /// <summary>
         /// Last error occured during configuration read
         /// </summary>
-        private NLogConfigurationException _lastParsingError;
+        private readonly List<string> _parsingErrors;
 
         /// <summary>
         /// Returns children elements with the specified element name.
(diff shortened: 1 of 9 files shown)


304NotModified:
hey thanks!

I just checked the code, but something is not clear to me. If we allow/ignore duplicate attributes, why do we need the "last error"?

nazim9214:
In current implementation, we throw first of these exceptions in elements tree if throwConfigExceptions is enabled. I can't throw it immediately, because configuration must be loaded, to setup all needed options to Logger (for example throwConfigException option, targets etc.).

If something is not written well or not clear, i can rewrite this code. What do you think?

304NotModified:
@nazim9214 

We could also throw the exception here? https://github.com/NLog/NLog/pull/1732/files#diff-0e7810ffc1c0cd264e0012f3310ed9c2R254

(after checking `configExcception.MustBeRethrown()` ?

nazim9214:
If I correctly  understood all the logic:
1. Firstly, we are parsing XML configuration into NLogXmlElement tree.
2. Next, we are calling ParseTopLevel, and filling LogFactory properties with corresponding values.

If exception is thrown here https://github.com/NLog/NLog/pull/1732/files#diff-0e7810ffc1c0cd264e0012f3310ed9c2R254, the second part won't be executed. All LogFactory properties will be unfilled (LogFactory.ThrowExceptions, LogFactory.ThrowConfigExceptions, targets etc.), so ```configExcception.MustBeRethrown()``` - will always return "false" for this case (it uses LogManager.ThrowConfigExceptions, LogManager.ThrowExceptions).

304NotModified:
Ah now I fully understand it  :) Thx!

I think it's a bit uncommon to store an exception and to throw it later. Maybe we could replace _lastParsingError with _parseError(s) and as `string` or `List<string>`. Then in step 2 we do:

a. check if there are any errors
b. if so, then check if `ThrowConfigExceptions` is enabled
c. if so, then throw **one** Exception with all the messages
d. if not, then write all the messages as warning to the `InternalLogger`.

What do you think?

nazim9214:
Good idea! I'll do it soon.

304NotModified:
👍

src/NLog/Config/NLogXmlElement.cs
```diff
@@ -212,19 +221,59 @@ public void AssertName(params string[] allowedNames)
             throw new InvalidOperationException("Assertion failed. Expected element name '" + string.Join("|", allowedNames) + "', actual: '" + this.LocalName + "'.");
         }
 
+        /// <summary>
+        /// Try to get all errors occured during xml parsing.
+        /// A return value indicates whether there any error.
+        /// </summary>
+        /// <param name="errors">Array of error messages</param>
+        public bool TryGetParsingErrors(out string[] errors)
```
304NotModified: I don't prefer `out` parameters. Could be change this to `public string GetParsingErrors()`? 

The "try" pattern is nice, but IMO more suitable on parsing.

nazim9214:
Thanks for the advice!

src/NLog/Config/NLogXmlElement.cs
```diff
@@ -212,19 +221,59 @@ public void AssertName(params string[] allowedNames)
...
+        {
+            var errors = GetParsingErrorsInternal().ToArray();
+            if (errors.Length == 0) return null;
+            return string.Join(Environment.NewLine, errors);
+        }
+
+        /// <summary>
+        /// Returns all parsing errors from current and all child elements.
+        /// </summary>
+        private IEnumerable<string> GetParsingErrorsInternal()
```
304NotModified: Thanks for the changes!

As from our part, we find it important to have a good public API. I think it would be better API if I could read the details, so that the errors aren't joined to a single string. Do you agree?

The easiest way is to make `GetParsingErrorsInternal()` public and rename it to `GetParsingErrors()` (and drop the `GetParsingErrors()`). What do you think?

tests/NLog.UnitTests/NLog.UnitTests.mono.csproj
```diff
@@ -250,4 +252,7 @@
       <Name>SampleExtensions.mono</Name>
     </ProjectReference>
   </ItemGroup>
+  <ItemGroup>
+    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
```
304NotModified: any idea what this is?

nazim9214:
Wow, i'm failing my tests:) Sorry, too hurried

src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -446,6 +445,33 @@ private void Initialize(XmlReader reader, string fileName, bool ignoreErrors)
...
+        /// Checks whether any error during XML configuration parsing has occured.
+        /// If there are any and <c>ThrowConfigExceptions</c> or <c>ThrowExceptions</c>
+        /// setting is enabled - throws <c>NLogConfigurationException</c>, otherwise
+        /// just write an internal log at Warn level.
+        /// </summary>
+        /// <param name="rootContentElement">Root NLog configuration xml element</param>
+        private void CheckParsingErrors(NLogXmlElement rootContentElement)
+        {
+            var parsingErrors = rootContentElement.GetParsingErrors().ToArray();
+            if(!parsingErrors.Any())
```
304NotModified: this should be `if(parsingErrors.Any())`  I guess ;)

304NotModified:
yes a `!` mistake. 

the CI services are helpful today :)

304NotModified:
:lgtm:

---

Reviewed 5 of 9 files at r1, 1 of 3 files at r2, 2 of 2 files at r4, 3 of 3 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1732#-:-KWUrJHogNpC9Jw0r4DL:bnfp4nl)*

304NotModified:
Merged :) Thanks!!!

Issue #1724, opened 2016-10-27
Special characters with webservice target and WebApi


chwebdude:
**Question**
I would like to send the logdata to a Asp.Net WebApi Backend. But all special characters are encoded wrong. I tried to change the formatters in WebApi and tried all possible combinations of the `HttpPost` protocol (encoding and BOM) of the `WebService` target. But still the special characters are received as the unknown unicode character �.
What can I do?

**NLog version**: 4.3.9

**Platform**: .Net 4.6

**Current NLog config**

```
<target xsi:type='WebService' 
   name='ws' 
   url='http://localhost:8084/Logs' 
   protocol='HttpPost' 
   encoding='utf-8' 
   includeBOM='false'>
...
</target>
```

**WebApi Controller:**

```
[HttpPost]
public async Task<HttpResponseMessage> Post([FromBody] LogEntry logEntry)...
```

**Example**
**_Test Application**_
`logger.Debug("a b c é k è ï ?");`
**_Result**_
![image](https://cloud.githubusercontent.com/assets/7746357/19769716/df34c766-9c5c-11e6-9130-668956b202c6.png)

304NotModified:
Checked the code in NLog (in short) and it should work. Any change to check it with fiddler?

304NotModified:
I could reproduce the issue. See https://github.com/NLog/NLog/pull/1725

I looks like the problem is in `UrlHelper.UrlEncode`, see [source](https://github.com/NLog/NLog/blob/0b79e8b1019b6b4a4a3f1f059a8cb642b34d5760/src/NLog/Internal/UrlHelper.cs#L69-L85)

not sure how to fix.

304NotModified:
I think I could use `Uri.EscapeDataString` ?

chwebdude:
Resolved with #1735

304NotModified:
Thanks! We will close this one as it has been merged :)

PR #1725, opened 2016-10-27
[WIP] Improved webservice tests / tryout   Uri.EscapeDataString

304NotModified:
not working yet, for https://github.com/NLog/NLog/issues/1724

---

This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1725)

PR #1735, opened 2016-10-30
UrlHelper - Added standard support for UTF8 encoding, added support for RFC2396  &  RFC3986

snakefoot:
Grapped the unit-test from #1725, but implemented the encoding manually so it stays the same independent of dot-net version.

Fixes https://github.com/NLog/NLog/issues/1724

---

This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1735)

### DIFF
--- a/src/NLog/Internal/UrlHelper.cs
+++ b/src/NLog/Internal/UrlHelper.cs
@@ -41,17 +41,11 @@ namespace NLog.Internal
     /// </summary>
     internal static class UrlHelper
     {
-        private static readonly char[] hexUpperChars =
-            { '0', '1', '2', '3', '4', '5', '6', '7',
-            '8', '9', 'A', 'B', 'C', 'D', 'E', 'F' };
-        private static readonly char[] hexLowerChars =
-            { '0', '1', '2', '3', '4', '5', '6', '7',
-            '8', '9', 'a', 'b', 'c', 'd', 'e', 'f' };
-
         [Flags]
         public enum EscapeEncodingFlag
         {
-            /// <summary>Use UnreservedMarks instead of ReservedMarks, as specified by chosen RFC</summary>
+            None = 0,
+            /// <summary>Allow UnreservedMarks instead of ReservedMarks, as specified by chosen RFC</summary>
             UriString = 1,
             /// <summary>Use RFC2396 standard (instead of RFC3986)</summary>
             LegacyRfc2396 = 2,
@@ -63,39 +57,15 @@ namespace NLog.Internal
             NLogLegacy = 16 | LegacyRfc2396 | LowerCaseHex | UriString,
         };
 
-        private const string RFC2396ReservedMarks = @";/?:@&=+$,";
-        private const string RFC3986ReservedMarks = @":/?#[]@!$&'()*+,;=";
-        private const string RFC2396UnreservedMarks = @"-_.!~*'()";
-        private const string RFC3986UnreservedMarks = @"-._~";
-
-        /// <summary>
-        /// Url encode an URL
-        /// </summary>
-        /// <param name="str">URL to be encoded</param>
-        /// <param name="spaceAsPlus">space as + or %20? <c>false</c> (%20) is the safe option.</param>
-        /// <returns>Encoded url.</returns>
-        public static string UrlEncode(string str, bool spaceAsPlus)
-        {
-            if (string.IsNullOrEmpty(str))
-                return string.Empty;
-
-            StringBuilder result = new StringBuilder(str.Length + 20);
-            result.Append(str);
-            EscapeEncodingFlag escapeFlags = EscapeEncodingFlag.NLogLegacy;
-            if (spaceAsPlus)
-                escapeFlags |= EscapeEncodingFlag.SpaceAsPlus;
-            EscapeDataEncode(result, escapeFlags);
-            return result.ToString();
-        }
-
         /// <summary>
         /// Escape unicode string data for use in http-requests
         /// </summary>
-        /// <param name="target">unicode string-data to be encoded</param>
+        /// <param name="source">unicode string-data to be encoded</param>
+        /// <param name="target">target for the encoded result</param>
         /// <param name="flags"><see cref="EscapeEncodingFlag"/>s for how to perform the encoding</param>
-        public static void EscapeDataEncode(StringBuilder target, EscapeEncodingFlag flags)
+        public static void EscapeDataEncode(string source, StringBuilder target, EscapeEncodingFlag flags)
         {
-            if (target.Length == 0)
+            if (string.IsNullOrEmpty(source))
                 return;
 
             bool isUriString = (flags & EscapeEncodingFlag.UriString) == EscapeEncodingFlag.UriString;
@@ -108,9 +78,10 @@ namespace NLog.Internal
             byte[] byteArray = null;
             char[] hexChars = isLowerCaseHex ? hexLowerChars : hexUpperChars;
 
-            for (int i = 0; i < target.Length; ++i)
+            for (int i = 0; i < source.Length; ++i)
             {
-                char ch = target[i];
+                char ch = source[i];
+                target.Append(ch);
                 if (ch >= 'a' && ch <= 'z')
                     continue;
                 if (ch >= 'A' && ch <= 'Z')
@@ -119,7 +90,7 @@ namespace NLog.Internal
                     continue;
                 if (isSpaceAsPlus && ch == ' ')
                 {
-                    target[i] = '+';
+                    target[target.Length - 1] = '+';
                     continue;
                 }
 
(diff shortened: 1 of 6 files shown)


304NotModified:
Cool! Will check this! Thanks!

src/NLog/Internal/UrlHelper.cs
```diff
@@ -33,82 +33,152 @@
...
 
         /// <summary>
-        /// Url encode and URL
+        /// Url encode an URL
         /// </summary>
         /// <param name="str">URL to be encoded</param>
         /// <param name="spaceAsPlus">space as + or %20? <c>false</c> (%20) is the safe option.</param>
         /// <returns>Encoded url.</returns>
-        internal static string UrlEncode(string str, bool spaceAsPlus)
+        public static string UrlEncode(string str, bool spaceAsPlus)
```
304NotModified: Why public?

src/NLog/Internal/UrlHelper.cs
```diff
@@ -33,82 +33,152 @@
...
+        /// Escape unicode string data for use in http-requests
+        /// </summary>
+        /// <param name="target">unicode string-data to be encoded</param>
+        /// <param name="flags"><see cref="EscapeEncodingFlag"/>s for how to perform the encoding</param>
+        public static void EscapeDataEncode(StringBuilder target, EscapeEncodingFlag flags)
+        {
+            if (target.Length == 0)
+                return;
+
+            bool isUriString = ((int)flags & (int)EscapeEncodingFlag.UriString) == (int)EscapeEncodingFlag.UriString;
```
304NotModified: I think it's performance wise to create a variable for `(int)flags`?  *maybe also for others)

src/NLog/Internal/UrlHelper.cs
snakefoot: Well it is still hidden by the internal class protection, and it is meant to be used outside this class

src/NLog/Internal/UrlHelper.cs
snakefoot: Not sure I understand. Is there a performance hit by converting an enum to an integer? Actually thought I avoided the performance hit of the normal HasFlag() method :)

304NotModified: Casting in general could take some performance. Not sure in this case.

snakefoot:
Have added some properties to UrlEncodeLayoutRendererWrapper and WebServiceTarget, so one can control the encoding/escaping.

By default Rfc2396 is enabled, as this was the standard the existing NLog tried to implement. But I have changed so one have to enable EscapeDataNLogLegacy to get the original NLog encoding.

304NotModified:
Great!

Do you think we should release this in 4.3.11 or 4.4?

snakefoot:
No idea about the timing. The default config chosen by me, makes it a breaking change.

I guess the default config could be changed to a non-breaking change (EscapeDataNLogLegacy  = default true), depending on the urgency for 4.3.11.

304NotModified:
What's your opinion on EscapeDataNLogLegacy ? Is that a plausible option or just buggy legacy stuff? If it's a bug, then it isn't a breaking change :)

(Very good you have added the option, don't wanna change that)

snakefoot:
I would say EscapeDataNLogLegacy is non-standard legacy stuff. Normal url-data without unicode characters and "unwise" special characters will not be affected by this change.

304NotModified:
Reviewed 1 of 4 files at r1, 3 of 4 files at r3.
Review status: 4 of 6 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1735)*

304NotModified:
Reviewed 1 of 4 files at r1, 1 of 4 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1735)*

304NotModified:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[src/NLog/Internal/UrlHelper.cs, line 101 at r1](https://reviewable.io:443/reviews/nlog/nlog/1735#-KVM_wS390xBJ8-mZWw0-r1-101:-KVv_Ral5vXKmrg_utys:b-dqd85w) ([raw file](https://github.com/nlog/nlog/blob/fe5eab6b5fe12db6e8a4a6723cdcac959e2f3f7e/src/NLog/Internal/UrlHelper.cs#L101)):*
> Casting in general could take some performance. Not sure in this case. 
</details>
Checked it, in this case no performance hit.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1735)*

304NotModified:
Will merge this for NLog 4.4! Thanks for this great contribution! 

We need some docs here, right?

snakefoot:
> We need some docs here, right?
Yes breaking change along with new parameters for url-encode-wrapper and WebServiceTarget to revert to legacy mode.

Update documentation now, and say available from ver. 4.4?

304NotModified:
yes 4.4 and not really a breaking change as IMO it only fixes a bug. 

But because of all the new settings and RFCs, 4.4 (instead of 4.3.x) is IMO a better choice.

tests/NLog.UnitTests/Internal/UrlHelperTests.cs
```diff
@@ -53,15 +51,54 @@ public class UrlHelperTests
...
+        [Theory]
+        [InlineData("", true, "")]
+        [InlineData("", false, "")]
+        [InlineData("ab cd", true, "ab+cd")]
+        [InlineData("ab cd", false, "ab%20cd")]
+        [InlineData("one&two =three", true, "one%26two+%3dthree")]
+        [InlineData("one&two =three", false, "one%26two%20%3dthree")]
+        [InlineData(".()*-_!ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz", true, ".()*-_!ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz")]
+        [InlineData(".()*-_!ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz", false, ".()*-_!ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz")]
+        public void EscapeDataEncodeTestRfc2396(string input, bool spaceAsPlus, string result)
```
304NotModified: shouldnt we add some unit test cases for special chars here?

tests/NLog.UnitTests/Internal/UrlHelperTests.cs
```diff
@@ -53,15 +51,54 @@ public class UrlHelperTests
...
+        }
+
+        [Theory]
+        [InlineData("", true, "")]
+        [InlineData("", false, "")]
+        [InlineData("ab cd", true, "ab+cd")]
+        [InlineData("ab cd", false, "ab%20cd")]
+        [InlineData(@"http://foo:bar@host:123/path?query#fragment", true, @"http://foo:bar@host:123/path?query#fragment")]
+        [InlineData(@"http://foo:bar@host:123/path?query#fragment", false, @"http://foo:bar@host:123/path?query#fragment")] 
+        public void EscapeDataEncodeTestRfc3986(string input, bool spaceAsPlus, string result)
```
304NotModified: same about the special chars

304NotModified:
Great stuff this!

304NotModified:
Review status: 5 of 6 files reviewed at latest revision, 2 unresolved discussions.

---

*[tests/NLog.UnitTests/Internal/UrlHelperTests.cs, line 77 at r3](https://reviewable.io:443/reviews/nlog/nlog/1735#-KVN9W8Y_oT1p1fMu2ft-r3-77:-KWcn36Tr-PDD0Xmmv2f:b-7m37tt) ([raw file](https://github.com/nlog/nlog/blob/9f2949c227a6a8c28c366266b2c7bad458ec120a/tests/NLog.UnitTests/Internal/UrlHelperTests.cs#L77)):*
> shouldnt we add some unit test cases for special chars here?
</details>
I was more thinking of É, é, è etc

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1735)*

304NotModified:
Review status: 5 of 6 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/NLog/Internal/UrlHelper.cs, line 66 at r4](https://reviewable.io:443/reviews/nlog/nlog/1735#-KWcnZS7HOzohVudn1zy:-KWcnZS7HOzohVudn1zz:btdog2b) ([raw file](https://github.com/nlog/nlog/blob/d28aae4cc290046abe480702452d6ebafe03ebd9/src/NLog/Internal/UrlHelper.cs#L66)):*
> ```Smalltalk
>         /// <param name="target">target for the encoded result</param>
>         /// <param name="flags"><see cref="EscapeEncodingFlag"/>s for how to perform the encoding</param>
>         public static void EscapeDataEncode(string source, StringBuilder target, EscapeEncodingFlag flags)
> ```

I think this looks solid, but maybe we have also to provide a decode, at least for the unit tests? 

What do you think?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1735)*

snakefoot:
*[tests/NLog.UnitTests/Internal/UrlHelperTests.cs, line 77 at r3](https://reviewable.io:443/reviews/nlog/nlog/1735#-KVN9W8Y_oT1p1fMu2ft-r3-77:-KWdbIFOcNQqMppTdc8P:b-896fix) ([raw file](https://github.com/nlog/nlog/blob/9f2949c227a6a8c28c366266b2c7bad458ec120a/tests/NLog.UnitTests/Internal/UrlHelperTests.cs#L77)):*
> I was more thinking of É, é, è etc
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1735)*

snakefoot:
*[tests/NLog.UnitTests/Internal/UrlHelperTests.cs, line 94 at r3](https://reviewable.io:443/reviews/nlog/nlog/1735#-KVN9W8Y_oT1p1fMu2ft-r3-94:-KWdbPSkOd1VcyBkRgfH:b-896fix) ([raw file](https://github.com/nlog/nlog/blob/9f2949c227a6a8c28c366266b2c7bad458ec120a/tests/NLog.UnitTests/Internal/UrlHelperTests.cs#L94)):*
> same about the special chars
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1735)*

snakefoot:
*[src/NLog/Internal/UrlHelper.cs, line 66 at r4](https://reviewable.io:443/reviews/nlog/nlog/1735#-KWcnZS7HOzohVudn1zy:-KWdbTlqiYhyMw5a7cb8:b-896fix) ([raw file](https://github.com/nlog/nlog/blob/d28aae4cc290046abe480702452d6ebafe03ebd9/src/NLog/Internal/UrlHelper.cs#L66)):*
> I think this looks solid, but maybe we have also to provide a decode, at least for the unit tests? 
> 
> What do you think?
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1735)*

304NotModified:
Reviewed 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1735)*

304NotModified:
Merged! Thanks!!!

304NotModified:
docs:
https://github.com/NLog/NLog/wiki/WebService-target
https://github.com/NLog/NLog/wiki/Url-Encode-Layout-Renderer

thanks @snakefoot for the docs!

PR #1740, opened 2016-11-02
WebServiceTarget support for JSON & Injecting JSON serializer into NLog

### LINKED ISSUE #700: WebserviceTarget - HttpPost, need for other encoding types (JSON and/or XML)
The current implementation of `WebServiceTarget` posts the content as "application/x-www-form-urlencoded". We need therefor the `[FromBody]` attribute on an REST api. 

Posting XML of JSON would be more conventional.

See unit test `WebServiceTargetTests.WebserviceTest_restapi_httppost` and #699

tetrodoxin:
Added support for POST of JSON and XML documents and added respective tests. I tried to keep it simple, comprehensible and not to break with existing functionality.

I'd have to change existing tests in WebServiceTargetTests to extend their usability to the new types but I wasn't sure about altering existing tests of other authors, so I added new parts.

Have a look and tell me what you think about the direction it's heading.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1740)

fixes #700


304NotModified:
Cool!

Could you undo the moved methods? It would be easier to review and otherwise we have a hard time merging this also in the .NET Core branch. Thanks!

tetrodoxin:
Oh I see, that auto sorting and formatting of VS has created a diff-hell.
I think I'll deal with this tomorrow, just adding the new stuff at the bottom and try to prevent VS from exchanging TAB<=>spaces

Good night.

tetrodoxin:
How do you want it to be 'injected' ?

For XmlSerializer too?

304NotModified:
> For XmlSerializer too?

Maybe, but that's already in non system.web libraries so less a problem IMO

tetrodoxin:
So JSONSerializer injecting by defining an interface and adding a property to the target?

304NotModified:
> So JSONSerializer injecting by defining an interface and adding a property to the target?

Yes, and it would be nice if the property is at higher level than this target. I think`ConfigurationItemFactory` or a new class.

304NotModified:
good night1

tetrodoxin:
So, tried a completed reformatting as good as possible.
Is it feasible this way?

I would then go and deal with JsonFormatter injection.

tetrodoxin:
Two possible ways spontaneously cross my mind:

- Just add an `public IJsonSerializer' JsonSerializer {get; set;}` to `ConfigurationItemFactory` which in turn sets an property to every `WebServiceTarget` instance it creates
- introduce a `IServiceResolver` interface and an `InjectedAttribute` that marks properties of objects created by `ConfigurationItemFactory` as injectable and so is done, if that resolver 'knows' the type of that property 

The first option would be more simple and quickly done, the second option indeed slightly more complex but more extendible and may be used for further scenarios. On the other hand, if no one can think of any additional use cases, that would be a perfect example of overkill.

tetrodoxin:
Interface so far:

```
public interface IJsonSerializer
{
	void Serialize(TextWriter writer, object value);
	void Serialize(TextWriter writer, object value, Type objectType);
	object Deserialize(TextReader reader, Type objectType);
	T Deserialize<T>(TextReader reader);
}
```

304NotModified:
I prefer the first one. More in line with the current options in NLog. :)

tetrodoxin:
After reflecting about it...
Which 'object' shall this serializer convert to JSON, actually? Because by now WebServiceTarget writes parameters that are named in config, one by one. There is no complex object to serialize.

304NotModified:
Well then the simple json serializer will work.

Except that is an issue when there is a quote in the current value AFAIK.

src/NLog/Targets/WebServiceTarget.cs
```diff
@@ -296,125 +320,291 @@ private Uri BuildWebServiceUrl(object[] parameterValues)
...
+                return string.Format("\"{0}\":{1}",
+                    parameter.Name,
+                    getJsonValueString(value));
+            }
 
-            return ms;
+            private static string getJsonValueString(object value)
+            {
+                if (value == null) return "null";
+                else if (value is string) return string.Format("\"{0}\"", value);
```
304NotModified: we need to escape the quote inside the string?

tetrodoxin:
Yup, now as you say it, forgot to escape strings.

tetrodoxin:
I'm on it by tomorrow.

tetrodoxin:
So, since the named parameters in config are just objects, I decided to add a slim serilizer interface, namely `ICompactJsonSerializer`, containing only a single method to serialize a value. In addition there is a default implementation, doing very rudimentary serialization and string escaping.
Tests for serializer are included.
Extended `ConfigurationItemFactory` so, that it provides a virtual method `InterceptCreation`, which is invoked right after a new object has been created via the instantiator in `CreateInstance`.
The base implementation of `InterceptCreation` realizes the injection of the JSON serializer into `WebServiceTarget`, if one was set by `ConfigurationItemFactory.SetJsonSerializer()`

Let me know, what you think... uhm, and I just see, I have to deal with Travis complaints :-/

tetrodoxin:
Don't understand this: AppVeyor is complaining about things, that aren't contained in that files ?!?

304NotModified:
It's he SL build, but I see you already seen it :) 👍

tetrodoxin:
yeah, several fixes later... yeehah, there we go :)

304NotModified:
Reviewed 2 of 3 files at r3, 1 of 8 files at r4, 12 of 21 files at r5, 7 of 11 files at r6, 2 of 2 files at r7.
Review status: 24 of 25 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/NLog/Targets/DefaultJsonSerializer.cs, line 144 at r7](https://reviewable.io:443/reviews/nlog/nlog/1740#-KVyulIZggDw3Uk-F-4I:-KVyulIZggDw3Uk-F-4J:b-tgxutr) ([raw file](https://github.com/nlog/nlog/blob/053e98cc553b183670aef33b6a4534e69fdc0f2c/src/NLog/Targets/DefaultJsonSerializer.cs#L144)):*
> ```Smalltalk
>         }
> 
>         private static string escapeString(string str)
> ```

please proper camels here ;)

(CamelCase)

---

*[src/NLog/Targets/ICompactJsonSerializer.cs, line 42 at r7](https://reviewable.io:443/reviews/nlog/nlog/1740#-KVyuvaMye7FsUO-1HZN:-KVyuvaMye7FsUO-1HZO:bajcz67) ([raw file](https://github.com/nlog/nlog/blob/053e98cc553b183670aef33b6a4534e69fdc0f2c/src/NLog/Targets/ICompactJsonSerializer.cs#L42)):*
> ```Smalltalk
>     /// Useful for wrappers for existing serializers.
>     /// </summary>
>     public interface ICompactJsonSerializer
> ```

I would prefer `IJsonSerializer`

Dunno why it should be "compact". Maybe I would write an implementation which won't "compact".

Let me know what you think!

---

*[src/NLog/Targets/WebServiceProtocol.cs, line 64 at r7](https://reviewable.io:443/reviews/nlog/nlog/1740#-KVyvF1qz9RAGogUPAEn:-KVyvF1rooFAjhCaFs8I:b7l3v78) ([raw file](https://github.com/nlog/nlog/blob/053e98cc553b183670aef33b6a4534e69fdc0f2c/src/NLog/Targets/WebServiceProtocol.cs#L64)):*
> ```Smalltalk
>         /// Do an HTTP POST of a JSON document.
>         /// </summary>
>         Json,
> ```

Maybe JsonPost and JsonXml?

---

*[src/NLog/Targets/WebServiceTarget.cs, line 192 at r7](https://reviewable.io:443/reviews/nlog/nlog/1740#-KVyvShA5AFWh-k_RsIZ:-KVyvShA5AFWh-k_RsI_:b-pgdzaz) ([raw file](https://github.com/nlog/nlog/blob/053e98cc553b183670aef33b6a4534e69fdc0f2c/src/NLog/Targets/WebServiceTarget.cs#L192)):*
> ```Smalltalk
>             get
>             {
>                 if (_jsonSerializer == null) _jsonSerializer =  DefaultJsonSerializer.Instance;
> ```

I think we should take there the `ConfigurationItemFactory.JsonSerializer`?

---

*[tests/NLog.UnitTests/Targets/DefaultJsonSerializerTests.cs, line 42 at r7](https://reviewable.io:443/reviews/nlog/nlog/1740#-KVyvunTGejTmaX_gDDM:-KVyvunTGejTmaX_gDDN:b8ox7pv) ([raw file](https://github.com/nlog/nlog/blob/053e98cc553b183670aef33b6a4534e69fdc0f2c/tests/NLog.UnitTests/Targets/DefaultJsonSerializerTests.cs#L42)):*
> ```Smalltalk
> #endif
>     using System;
>     public class DefaultJsonSerializerTests : NLogTestBase
> ```

thanks! Looks great! 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1740)*

304NotModified:
Thanks for the contributions! I have checked the code!

I think it will work, but I won't prefer the "interception". I think we should read the `ConfigurationItemFactory.JsonSerializer` inside the WebserviceTarget and then we don't need the "interception"?

Also check your camel-casing ;)

tetrodoxin:
Ok, to do: CamelCasing (sorry, lowerCamel for private members is a habit of mine), 
rename WebProtocol (JsonPost, XmlPost).

I used that interception way to make it reusable and to decouple WebServiceTarget from ConfigurationItemFactory, because it don't necessarily have to know about it. But I will switch that.

Compact JSON, because the WebServiceTarget produces a compact JSON and I wanted to prevent indented serialized values inside an actual single line document. I choosed compact JSON since it's quite common to use it for web transfers. But we may also switch that, as you want.

tetrodoxin:
So, I took care for everything except the compact JSON format, meaning a single line document without indentation.
What may I do against the conflict of `ConfigurationItemFactory`?

tetrodoxin:
appveyor is waiting for a while now.

304NotModified:
close/open to wake up AppVeyor ;)

304NotModified:
Could you explain what a "compact json" is?

tetrodoxin:
Compact JSON just means no tabs, no spaces, and no line breaks.
But I read the RFCs and the ECMA again, maybe it just doesn't matter, even if you mix tabbed and single line JSON, so we could drop the comment and change the name of the interface.

tetrodoxin:
I hope it now somewhat fits your idea of how it should be :-)
Let me know if I can do something more about it ;-)

304NotModified:
Reviewed 2 of 5 files at r8, 6 of 7 files at r9, 17 of 17 files at r10.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[src/NLog/Targets/IJsonSerializer.cs, line 50 at r10](https://reviewable.io:443/reviews/nlog/nlog/1740#-KWUmUYk_86rViKIb2Fb:-KWUmUYk_86rViKIb2Fc:b-flu3qq) ([raw file](https://github.com/nlog/nlog/blob/e1d045e6669d64d274ec083ee1fb8e6a3cdf86e4/src/NLog/Targets/IJsonSerializer.cs#L50)):*
> ```Smalltalk
>         /// <param name="value">The object to serialize to JSON.</param>
>         /// <returns>Serialized value.</returns>
>         string SerializeValue(object value);
> ```

I was thinking of `SerializeObject`. What do you think?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1740)*

304NotModified:
Looks very good! Thanks! 

One final thing, was doubting between the method name of the interface (SerializeValue). I think SerializeObject is more appropriate. What do you think?

tetrodoxin:
I think originally I choosed `SerializeValue` since it should just serialize the value part of a JSON key/value-pair. But actually it serializes objects, so... yes, I'll just rename it.

snakefoot:
Just curious regarding unicode characters in the string. Can see JsonEncodeLayoutRendererWrapper has special logic for characters not within the ASCII-range (NeedsEscaping).

304NotModified:
> Just curious regarding unicode characters in the string. Can see JsonEncodeLayoutRendererWrapper has special logic for characters not within the ASCII-range (NeedsEscaping). Or is this handled by URL-Escaping?

Which file do you mean?

snakefoot:
> @snakefoot Which file do you mean?

I'm looking at the exiting JSON-encoder implementation (JsonEncodeLayoutRendererWrapper), and I can see it has special handling of unicode characters. I cannot see this handling in the new JSON.-encoder. So I was just curious why.

304NotModified:
Ah good one!

tetrodoxin:
Didn't find any clue that would be necessary or even mandatory. Since the data is transferred with unicode encoding, I didn't see a cause for escaping characters.

snakefoot:
>    Unicode characters may be placed within the
>    quotation marks except for the characters that must be escaped:
>    quotation mark, reverse solidus, and the control characters (U+0000
>    through U+001F).

>   JSON text SHALL be encoded in Unicode.  The default encoding is UTF-8.

https://tools.ietf.org/html/rfc4627

So it seems that normal unicode issues can be ignored, but control characters needs to be handled. While it  is optional to escape non-ascii characters like this: "\u005C". Again just curious as I  know nothing about JSON.

tetrodoxin:
May be more clean now, all control chars (<$1f) are now escaped. The special ones (\t, \r, \n etc) as before and the rest as unicode codes \00xx

snakefoot:
Now that you have added support for serializing IEnumerable and IDictionary, then you have also added the potential cyclic-graph-pitfall.

Ex. calling SerializeObject on the following object, might take a while:

```
List<object> hello = new List<object>();
hello.Add(hello);
SerializeObject(hello);
```

Maybe have a level counter (default starting at 2) of how many levels down the object-graph that it will explore.

tetrodoxin:
So I'm tempted to say, we keep the serializer rudimentary and leave out serialization of collection types, because there are ready, mature serializers out there one can use, which have been the reason for this interface.

src/NLog/Targets/DefaultJsonSerializer.cs
```diff
@@ -0,0 +1,185 @@
...
+using System.Text;
+
+namespace NLog.Targets
+{
+    /// <summary>
+    /// Default class for serialization of values to JSON format.
+    /// </summary>
+    public class DefaultJsonSerializer : IJsonSerializer
+    {
+        private static List<Type> NumericTypes = new List<Type>
```
304NotModified: please create use a hashset (because of the `contains` futher)

304NotModified:
> Maybe have a level counter (default starting at 2) of how many levels down the object-graph that it will explore.

The easier and better way is to store the reference the a hashset? (and compare if it's already used)

304NotModified:
> So I'm tempted to say, we keep the serializer rudimentary and leave out serialization of collection types,

Good point, doubting as I think we are almost there.

tetrodoxin:
So:
switch do hashset for numeric types
and
use a hashset to track visited references to prevent circular paths.

Allright?

304NotModified:
Yes that would be nice!

tetrodoxin:
I'm on it

tetrodoxin:
Although the question may be: how to handle reference loops? JSON.NET for example per default breaks with an exception. You can alter this behavior to 2 alternatives. 
We could end with an exception or skip serialization of objects, when they're child of themselves. In both cases we track every 'path' of object serialization.

snakefoot:
Consider extracting the method JsonLayout.IsNumeric() into a helper-class, then you can skip the HashSet-Lookup.

304NotModified:
I prefer skip instead of exception

tetrodoxin:
Now it uses Hashset (and Contains() ) for checking current serialization path for loop. In the (unlikely) case of an infinite loop because of special implementations of IEnumerable for example, there is also an maximum recursion depth of 10, just protecting application from a stack overflow or out of memory.

304NotModified:
👍 
cool. Maybe unlikely, but stackoverflows are one the worst issue you could have in an application. So thanks for fixing it!

src/NLog/Targets/WebServiceTarget.cs
```diff
@@ -412,9 +429,18 @@ protected override string GetFormattedContent(string parametersContent)
 
             protected override string GetFormattedParameter(MethodCallParameter parameter, object value)
             {
+                string parameterValue = Convert.ToString(value, CultureInfo.InvariantCulture);
+                string valueString = string.Empty;
+                if (!string.IsNullOrEmpty(parameterValue))
+                {
+                    var sb = new StringBuilder(Math.Max(parameterValue.Length + 20, Target.Parameters.Count > 1 ? 256 : 0));
```
snakefoot: The Target.Parameters.Count-check only makes sense when using the same StringBuilder for streaming all parameters.

Unless you optimize the (re-)usage of the StringBuilder, then it should just be:

var sb = new StringBuilder(parameterValue.Length + 20);

src/NLog/Targets/WebServiceTarget.cs
```diff
@@ -412,9 +429,18 @@ protected override string GetFormattedContent(string parametersContent)
...
+                    UrlHelper.EscapeDataEncode(parameterValue, sb, encodingFlags);
+                    valueString = sb.ToString();
+                }
+
                 return string.Format("{0}={1}",
```
snakefoot: Could it not be changed it into (Could probably be optimized even more if re-using the StringBuilder):

if (string.IsNullOrEmpty(parameterValue))
{
    return string.Concat(parameter.Name, "=");
}
else
{
    var sb = new StringBuilder(parameter.Name.Length + parameterValue.Length + 20);
    sb.Append(parameter.Name);
    sb.Append("=");
    UrlHelper.EscapeDataEncode(parameterValue, sb, encodingFlags);
    return sb.ToString();
}

src/NLog/Targets/WebServiceTarget.cs
```diff
@@ -312,135 +335,302 @@ private Uri BuildWebServiceUrl(object[] parameterValues)
...
+            }
+
+            protected override string Separator
+            {
+                get { return ","; }
+            }
+
+            protected override string GetFormattedContent(string parametersContent)
+            {
+                return string.Format("{{{0}}}", parametersContent);
```
snakefoot: Consider changing into this:

return string.Concat("{{", parametersContent, "}}");

src/NLog/Targets/WebServiceTarget.cs
```diff
@@ -312,135 +335,302 @@ private Uri BuildWebServiceUrl(object[] parameterValues)
...
+                base.InitRequest(request);
+
+                string soapAction;
+                if (Target.Namespace.EndsWith("/", StringComparison.Ordinal))
+                {
+                    soapAction = Target.Namespace + Target.MethodName;
+                }
+                else
+                {
+                    soapAction = Target.Namespace + "/" + Target.MethodName;
```
snakefoot: Consider changing into this (Maybe also use string.Concat for the one without slash for consistency)

soapAction = string.Concat(Target.Namespace, "/", Target.MethodName);

304NotModified:
This is ready to merge right?

src/NLog/Targets/WebServiceTarget.cs
```diff
@@ -312,135 +335,304 @@ private Uri BuildWebServiceUrl(object[] parameterValues)
...
+            {
+                string parameterValue = Convert.ToString(value, CultureInfo.InvariantCulture);
+                string valueString = string.Empty;
+                if (string.IsNullOrEmpty(parameterValue))
+                {
+                    return string.Concat(parameter.Name, "=");
+                }
+                else
+                {
+                    var sb = new StringBuilder(parameterValue.Length + 20);
```
snakefoot: Consider changing it into:

var sb = new StringBuilder(parameter.Name.Length + parameterValue.Length + 20);

304NotModified:
Reviewed 1 of 17 files at r10, 1 of 4 files at r11, 21 of 22 files at r13, 1 of 1 files at r14.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1740)*

304NotModified:
@tetrodoxin I have one final comment: please don't create variables with just 1 char! `l`, `c`, `d` etc. Please do use meaningful names. Only in math single chars are allowed ;)

PR #1745, opened 2016-11-06
FilePathLayout - Reduce memory-allocation for cleanup of filename

snakefoot:
Reuse clean-filename when raw-filename is unchanged.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1745)


src/NLog/Internal/FilePathLayout.cs
```diff
@@ -77,6 +77,9 @@ internal class FilePathLayout : IRenderable
 
         private bool _cleanupInvalidChars;
 
+        private string _cachedPrevRawFileName;
+        private string _cachedPrevCleanFileName;
```
304NotModified: We should document (xmldoc) that `_cachedPrevCleanFileName` is a pair with `_cachedPrevRawFileName`, right?

src/NLog/Internal/FilePathLayout.cs
```diff
@@ -138,56 +141,78 @@ public Layout GetLayout()
...
+            var rawFileName = GetRenderedFileName(logEvent);
+            if (string.IsNullOrEmpty(rawFileName))
+            {
+                return rawFileName;
+            }
+
+            if ((!_cleanupInvalidChars || cleanedFixedResult != null) && _filePathKind == FilePathKind.Absolute)
+                return rawFileName; // Skip clean filename string-allocation
+
+            if (string.CompareOrdinal(_cachedPrevRawFileName, rawFileName) == 0 && _cachedPrevCleanFileName != null)
```
304NotModified: In case of performance, the check `_cachedPrevCleanFileName != null` should be on the left?

snakefoot: _cachedPrevCleanFileName will never be null, and when it is then _cachedPrevRawFileName is also null. I just have a secret agenda when it comes to the PR with PoolSetup.

src/NLog/Internal/FilePathLayout.cs
snakefoot: Yes they are working together. Will update the xml-doc

304NotModified:
Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1745)*

304NotModified:
Thanks again!

PR #1746, opened 2016-11-06
DateLayout - Reduce memory allocation when low time resolution

snakefoot:
Cache the result of the layout when not using custom formatProvider and format has low time resolution (daily or hourly)

When having hourly log-file-format, then no need to re-format for every log-event.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1746)

### DIFF
--- a/src/NLog/LayoutRenderers/DateLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/DateLayoutRenderer.cs
@@ -34,6 +34,7 @@
 namespace NLog.LayoutRenderers
 {
     using System;
+    using System.Collections.Generic;
     using System.ComponentModel;
     using System.Globalization;
     using System.Text;
@@ -66,7 +67,22 @@ namespace NLog.LayoutRenderers
         /// </summary>
         /// <docgen category='Rendering Options' order='10' />
         [DefaultParameter]
-        public string Format { get; set; }
+        public string Format
+        {
+            get { return _format; }
+            set
+            {
+                _format = value;
+
+                // Check if caching should be used
+                DateTime cachedDateTime = IsLowTimeResolutionLayout(_format)
+                    ? DateTime.MaxValue     // Cache can be used, will update cache-value
+                    : DateTime.MinValue;    // No cache support
+                _cachedUtcTime = new KeyValuePair<DateTime, string>(cachedDateTime, string.Empty);
+                _cachedLocalTime = new KeyValuePair<DateTime, string>(cachedDateTime, string.Empty);
+            }
+        }
+        private string _format;
 
         /// <summary>
         /// Gets or sets a value indicating whether to output UTC time instead of local time.
@@ -75,9 +91,12 @@ namespace NLog.LayoutRenderers
         [DefaultValue(false)]
         public bool UniversalTime { get; set; }
 
-        private static char[] lowTimeResolutionChars = new char[] { 'Y', 'y', 'M', 'D', 'd', 'H', 'h' };
-        private System.Collections.Generic.KeyValuePair<DateTime, string> cachedUtcTime = new System.Collections.Generic.KeyValuePair<DateTime, string>();
-        private System.Collections.Generic.KeyValuePair<DateTime, string> cachedLocalTime = new System.Collections.Generic.KeyValuePair<DateTime, string>();
+        private const string _lowTimeResolutionChars = "YyMDdHh";
+
+        /// <summary>Cache-key (Last DateTime.UtcNow) + Cache-Value (DateTime.Format result)</summary>
+        private KeyValuePair<DateTime, string> _cachedUtcTime = new System.Collections.Generic.KeyValuePair<DateTime, string>();
+        /// <summary>Cache-key (Last DateTime.Now) + Cache-Value (DateTime.Format result)</summary>
+        private KeyValuePair<DateTime, string> _cachedLocalTime = new System.Collections.Generic.KeyValuePair<DateTime, string>();
 
         /// <summary>
         /// Renders the current date and appends it to the specified <see cref="StringBuilder" />.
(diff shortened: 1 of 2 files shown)


src/NLog/LayoutRenderers/DateLayoutRenderer.cs
```diff
@@ -74,21 +75,69 @@ public DateLayoutRenderer()
...
+            }
+            else
+            {
+                AppendDateLayout(builder, formatProvider, ts, ref cachedLocalTime);
             }
+        }
 
-            var formatProvider = GetFormatProvider(logEvent, Culture);
-            builder.Append(ts.ToString(this.Format, formatProvider));
+        private static bool IsLowTimeResolutionLayout(string dateTimeFormat)
```
304NotModified: should be not call this method on the setter?

src/NLog/LayoutRenderers/DateLayoutRenderer.cs
```diff
@@ -74,21 +75,69 @@ public DateLayoutRenderer()
...
+        {
+            for (int i = 0; i < lowTimeResolutionChars.Length; ++i)
+                dateTimeFormat = dateTimeFormat.Replace(lowTimeResolutionChars[i], ' ');
```
304NotModified: I'm not sure if this is the best approach. I would:

- create a dictionary of `lowTimeResolutionChars `
- would loop only over `dateTimeFormat` (and not `lowTimeResolutionChars`)

What do you think?

src/NLog/LayoutRenderers/DateLayoutRenderer.cs
```diff
@@ -74,21 +75,69 @@ public DateLayoutRenderer()
...
+                        return; // Cache hit
+                    }
+                }
+                else if (cachedTime.Value == null)
+                {
+                    // Check if caching should be used
+                    if (IsLowTimeResolutionLayout(this.Format))
+                        cachedTime = new System.Collections.Generic.KeyValuePair<DateTime, string>(DateTime.MaxValue, string.Empty);
+                    else
+                        cachedTime = new System.Collections.Generic.KeyValuePair<DateTime, string>(DateTime.MinValue, string.Empty);
```
304NotModified: Could you explain me why we would store `DateTime.MaxValue`/`DateTime.MinValue`? What it's saying?

Also would be nice to extract the datetime to a variable to remove the last code duplication.

src/NLog/LayoutRenderers/DateLayoutRenderer.cs
```diff
@@ -74,21 +75,69 @@ public DateLayoutRenderer()
...
+                }
+            }
+
+            string formatTime = timestamp.ToString(this.Format, formatProvider);
+            if (cachedTime.Key != DateTime.MinValue && ReferenceEquals(formatProvider, CultureInfo.InvariantCulture))
```
304NotModified: seen this check before, maybe extract it to a method?

src/NLog/LayoutRenderers/DateLayoutRenderer.cs
```diff
@@ -74,21 +75,69 @@ public DateLayoutRenderer()
         [DefaultValue(false)]
         public bool UniversalTime { get; set; }
 
+        private static char[] lowTimeResolutionChars = new char[] { 'Y', 'y', 'M', 'D', 'd', 'H', 'h' };
+        private System.Collections.Generic.KeyValuePair<DateTime, string> cachedUtcTime = new System.Collections.Generic.KeyValuePair<DateTime, string>();
```
304NotModified: I think the  key is the datetime to be used, and the value the format? I need a small explanation :)

src/NLog/LayoutRenderers/DateLayoutRenderer.cs
```diff
@@ -74,21 +75,69 @@ public DateLayoutRenderer()
...
+
             var ts = logEvent.TimeStamp;
             if (this.UniversalTime)
             {
                 ts = ts.ToUniversalTime();
+                AppendDateLayout(builder, formatProvider, ts, ref cachedUtcTime);
+            }
+            else
+            {
+                AppendDateLayout(builder, formatProvider, ts, ref cachedLocalTime);
```
304NotModified: please use  a variable to reduce code duplication.

src/NLog/LayoutRenderers/DateLayoutRenderer.cs
snakefoot: Yes it could also be checked on the Format property-setter

src/NLog/LayoutRenderers/DateLayoutRenderer.cs
snakefoot: Could easily change it too loop over the dateTimeFormat, and if letter and not in lowTimeResolutionChars, then it is not seen as being daily/hourly. Will change the logic.

304NotModified: A long there is no `contains` on a non-dictionary/non-hashset, then I'm happy :)

snakefoot:
Have added some more comments and changed IsLowTimeResolutionLayout

src/NLog/LayoutRenderers/DateLayoutRenderer.cs
```diff
@@ -75,8 +75,11 @@ public DateLayoutRenderer()
         [DefaultValue(false)]
         public bool UniversalTime { get; set; }
 
-        private static char[] lowTimeResolutionChars = new char[] { 'Y', 'y', 'M', 'D', 'd', 'H', 'h' };
+        private const string lowTimeResolutionChars = "YyMDdHh";
```
304NotModified: Why no hashset? IndexOf is `O(n)`, hashset is `O(1)`

src/NLog/LayoutRenderers/DateLayoutRenderer.cs
```diff
@@ -74,21 +75,72 @@ public DateLayoutRenderer()
...
+                    // Check if caching should be used
+                    DateTime cachedDateTime = IsLowTimeResolutionLayout(this.Format)
+                        ? DateTime.MaxValue     // Cache can be used, will update cache-value
+                        : DateTime.MinValue;    // No cache support
+                    cachedTime = new System.Collections.Generic.KeyValuePair<DateTime, string>(cachedDateTime, string.Empty);
+                }
+            }
+
+            string formatTime = timestamp.ToString(this.Format, formatProvider);
+            if (cachedTime.Key != DateTime.MinValue && invariantCulture)
```
304NotModified: great improvement! thanks!

src/NLog/LayoutRenderers/DateLayoutRenderer.cs
snakefoot: Well this is one time operation, and when below 100 elements then an array-scan is faster than calculating hash-code and doing index-lookup.

304NotModified: Is it really called once? Cannot see that from the code.

Also AFAIK a hashset with 3 items is already more efficient, when used repeatedly

304NotModified: and the hashset could be build statically ;)

snakefoot:
You won. Have modified the Format-property, so one can see it is a one-time operation.

304NotModified:
Reviewed 1 of 2 files at r1, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1746)*

304NotModified:
Thanks for the changes!



> You won. 

Please note that isn't my goal!

snakefoot:
> Please note that isn't my goal! 

I meant it in a good way :). Readable code is very important.

304NotModified:
:) good to hear!

PR #1752, opened 2016-11-08
Layout processinfo with support for custom Format-string

snakefoot:
Makes it possible to make a filename like this:

> fileName="c:/temp/Log/test1-${processinfo:property=StartTime:format=yyyy-MM-dd_HHmmss}.log"

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1752)

### DIFF
--- a/src/NLog/LayoutRenderers/ProcessInfoLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/ProcessInfoLayoutRenderer.cs
@@ -40,8 +40,8 @@ namespace NLog.LayoutRenderers
     using System.Diagnostics;
     using System.Reflection;
     using System.Text;
-    using NLog.Common;
     using NLog.Config;
+    using NLog.Internal;
 
     /// <summary>
     /// The information about the running process.
@@ -52,8 +52,6 @@ namespace NLog.LayoutRenderers
         private Process process;
 
         private PropertyInfo propertyInfo;
-
-        private MethodInfo propertyToString;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="ProcessInfoLayoutRenderer" /> class.
@@ -89,19 +87,6 @@ namespace NLog.LayoutRenderers
             }
 
             this.process = Process.GetCurrentProcess();
-
-            if (!string.IsNullOrEmpty(Format))
-            {
-                object value = this.propertyInfo.GetValue(this.process, null);
-                if (value != null)
-                {
-                    propertyToString = value.GetType().GetMethod("ToString", new[] { typeof(string), typeof(IFormatProvider) });
-                    if (propertyToString == null)
-                    {
-                        InternalLogger.Info("Layout {processinfo} property {0} does not support formatting {1}", this.propertyInfo, Format);
-                    }
-                }
-            }
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/NLog/LayoutRenderers/ProcessInfoLayoutRenderer.cs
```diff
@@ -81,6 +89,19 @@ protected override void InitializeLayoutRenderer()
             }
 
             this.process = Process.GetCurrentProcess();
+
+            if (!string.IsNullOrEmpty(Format))
```
304NotModified: what about `value.ToStringWithOptionalFormat(...)` ? (in class `FormatHelper`)

304NotModified:
Good idea! But I think all those reflection isn't needed :) See code comment

snakefoot:
Good observation. I have change the logic to use FormatHelper.ToStringWithOptionalFormat, and it still flies.

src/NLog/LayoutRenderers/ProcessInfoLayoutRenderer.cs
```diff
@@ -107,7 +113,11 @@ protected override void Append(StringBuilder builder, LogEventInfo logEvent)
             if (this.propertyInfo != null)
             {
                 var formatProvider = GetFormatProvider(logEvent);
-                builder.Append(Convert.ToString(this.propertyInfo.GetValue(this.process, null), formatProvider));
+                var value = this.propertyInfo.GetValue(this.process, null);
+                if (value != null && this.Format != null)
+                    builder.Append(Internal.FormatHelper.ToStringWithOptionalFormat(value, this.Format, formatProvider));
```
304NotModified: This method already handles null for both.  ;)

(and it's an extension method)

So my proposal: 

```
 builder.Append(value.ToStringWithOptionalFormat(this.Format, formatProvider));

```

snakefoot:
Yes it looks pretty now :)

304NotModified:
\0/ 

I the next major version of NLog we should look how generic handle this for all layoutrenderers. In all cases there should be a Format I think.

304NotModified:
:lgtm:

---

Reviewed 1 of 2 files at r1, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1752#-:-KW5D-eaWonGD6iHuj54:bnfp4nl)*

304NotModified:
Merged!

304NotModified:
docs added https://github.com/NLog/NLog/wiki/ProcessInfo-Layout-Renderer

src/NLog/LayoutRenderers/ProcessInfoLayoutRenderer.cs
```diff
@@ -69,6 +69,12 @@ public ProcessInfoLayoutRenderer()
         public ProcessInfoProperty Property { get; set; }
```
snakefoot: @304NotModified Think it should just have DefaultValue, and not DefaultParameter. Make PR to remove DefaultParameter?

304NotModified: yes please :) and add the other as default. Thx in advance!

Issue #1686, opened 2016-10-06
Provide JSON formatting option for layout=all-event-properties


jefflowe7:
**Type**: Feature request
**NLog version**: 4.3.5
**Platform**: .Net 4.5 

My request seems to already exist at https://github.com/NLog/NLog/issues/723, but it doesn't appear to be resolved.

Many logging services (Sumologic, Kibana) provide a robust JSON parsing capability. When logging an event, much important information is contained in the Properties on the event. However it doesn't appear that those Properties can be formatted as JSON via NLog.

Here's the layout config:

``` xml
  <layout xsi:type="JsonLayout">
      <attribute name="time" layout="${shortdate}" />
      <attribute name="level" layout="${level:upperCase=true}" />
      <attribute name="message" layout="${message}" />
      <attribute name="properties" layout="${all-event-properties}" />
  </layout>
```

Here's an example NLog call:

``` c#
    Exception e = new Exception("Something went wrong.");

    string serializedException = JsonSerialize(e);

    Logger
        .Info()
        .Message("Start.")
        .Property("Exception", serializedException)
        .Property("Host", "http://xyzzy.gov")
        .Write();
```

Here's an example of the output using the layout above:

``` javascript
  {
    "time": "2015-05-30 17:11:50.7131",
    "level": "INFO",
    "message": "Start",
    "properties": "Exception={"Message": "Something went wrong."}, Host="http://xyzzy.gov""
  }
```

Note that the value of properties is NOT JSON.

What we need is:

``` javascript
  {
    "time": "2015-05-30 17:11:50.7131",
    "level": "INFO",
    "message": "Start",
    "properties": {
        "Exception": {
            "Message": "Something went wrong"
        }
        "Host": "http://xyzzy.gov"
      }
  }
```

Note that for the Exception property, the property value as already been logged as serialized JSON.

304NotModified:
> Note that the value of properties is NOT JSON.

Agree, this isn't usefull. What is the result with `encode="false"`, e.g.

```xml
  <layout xsi:type="JsonLayout">
      <attribute name="time" layout="${shortdate}" />
      <attribute name="level" layout="${level:upperCase=true}" />
      <attribute name="message" layout="${message}" />
      <attribute name="properties" layout="${all-event-properties}" encode="false" />
  </layout>
```

jefflowe7:
Tried:

'''xml
    <attribute name="properties" layout="${all-event-properties}" encode="false" />
'''
Still getting:

'''
[...]

PR #1754, opened 2016-11-10
JsonLayout.AllEventProperties

aireq:
Here's an initial attempt at adding "AllEventProperties" functionality to `JsonLayout` as mentioned in issue #1686.  This is the first time I've made a pull request on GitHub. So if I'm not following the proper procedure or conventions let me know.

I have few questions:

1. I'm using a `for each` loop to iterate over the the log event properties, but everywhere else it looks like there's been an effort to use a `for` loop to avoid allocating an Enumerator. There doesn't look like a way to access the items or keys in a dictionary by index though. Is there some other way I could optimize this?

2. `LogEventInfo.Properties` is an `IDictionary<object,object>`. How should the keys be converted into a string property names? Should I just use `ToString()` as I have now? When would the key of a property not be a string?

3. Any ideas on how to handle property values that are collections or objects? I'm guessing we wouldn't want to use reflection to read properties from object values due to the performance hit. However, for my use case I'd like it to be able to identify collection properties and have them returned as an array of values in the JSON rather then "System.Collections.Generic.List`1[System.String]".

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1754)

### DIFF
--- a/src/NLog/Layouts/JsonLayout.cs
+++ b/src/NLog/Layouts/JsonLayout.cs
@@ -99,8 +99,6 @@ namespace NLog.Layouts
             bool first = true;
             bool hasContent = false;
 
-
-
             //Creates a copy of the list of specified attributes
             List<JsonAttribute> attributes = new List<JsonAttribute>(this.Attributes);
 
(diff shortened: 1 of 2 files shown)


src/NLog/Layouts/JsonLayout.cs
```diff
@@ -73,6 +75,18 @@ public JsonLayout()
...
+
+        /// <summary>
+        /// Gets or sets the option to include all properties from the log events
+        /// </summary>
+        public bool IncludeAllProperties { get; set; }
+
+        /// <summary>
+        /// List of property names to exclude when IncludeAllProperties is true
+        /// </summary>
+        public IList<string> ExcludedProperties { get; set; }
```
304NotModified: I don't this will work from the XML, yet! 

See https://github.com/NLog/NLog/pull/1439 (but I won't support IList). 

Is it possible to write an unit test for this, so we won't forget it? (and it will succeed if we rebase after https://github.com/NLog/NLog/pull/1439 has been finished)

src/NLog/Layouts/JsonLayout.cs
```diff
@@ -85,11 +99,71 @@ protected override string GetFormattedMessage(LogEventInfo logEvent)
...
+                    {
+                        attributes.Add(new JsonAttribute()
+                        {
+                            Name = propName,
+                            Encode = false, //Don't put quotes around null values
+                            Layout = "null"
+                        });
+                    }
+                    else if (
+                        prop.Value is bool
```
304NotModified: Maybe we should integrate this with https://github.com/NLog/NLog/pull/1740  (injecting Json serializer) . What do you think?

304NotModified:
> I'm using a for each loop to iterate over the the log event properties, but everywhere else it looks like there's been an effort to use a for loop to avoid allocating an Enumerator. There doesn't look like a way to access the items or keys in a dictionary by index though. Is there some other way I could optimize this?

We try to use `for` loops if we found in tests that would benefit. I think we could loop over the keys in this case, but I doubt if that will be an improvement. TL;DR, it's fine with a `foreach`

304NotModified:
Intent to release this nice feature into 4.4 :)

304NotModified:
We need to do the docs in wiki right?

304NotModified:
docs done:

https://github.com/NLog/NLog/wiki/JsonLayout

AuthorProxy:
needs to update xsd schema now, cause `The attribute is not declared`

304NotModified:
thanks @AuthorProxy , created a new issue for it: https://github.com/NLog/NLog/issues/2037

304NotModified:
@AuthorProxy It seems that nlog.schema was up-to-date, but not the one on nlog-project.org. This has been fixed. 

Have to find out how to do this automatically. (powershell script or something like that)

Issue #1652, opened 2016-09-19
Async: NLog is swallowing log messages


naymore:
**Type** (choose one): Bug

**NLog version**: 4.3.8

**Platform**:  .NET 4.6.2 

**Current NLog config** (xml or C#, if relevant)

**ALSO SEE UPDATES IN MY FIRST ANSWER**

``` xml
<?xml version="1.0" encoding="utf-8" ?>
<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd"
      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
      xsi:schemaLocation="http://www.nlog-project.org/schemas/NLog.xsd NLog.xsd"
      autoReload="true"
      throwExceptions="false"
      internalLogLevel="Warn" internalLogFile="logs\nlog-internal.log" >

  <targets async="true">
    <target xsi:type="ColoredConsole" name="c_stats" layout="${time} ${uppercase:${level:padding=-6:fixedLength=true}} ${threadid:padding=-2:fixedLength=true} ${message}">
      <highlight-row condition="true" foregroundColor="DarkGray"/>
    </target>
    <target xsi:type="ColoredConsole" name="c_darkgreen" layout="${time} ${uppercase:${level:padding=-6:fixedLength=true}} ${threadid:padding=-2:fixedLength=true} ${message}">
      <highlight-row backgroundColor="NoChange" condition="true" foregroundColor="Green"/>
    </target>
    <target xsi:type="ColoredConsole" name="c_highlight" layout="${time} ${uppercase:${level:padding=-6:fixedLength=true}} ${threadid:padding=-2:fixedLength=true} ${message}">
      <highlight-row backgroundColor="NoChange" condition="contains('${message}','Consumer0001')" foregroundColor="White"/>
      <highlight-row backgroundColor="NoChange" condition="contains('${message}','Consumer0002')" foregroundColor="Green"/>
      <highlight-row backgroundColor="NoChange" condition="contains('${message}','Consumer0003')" foregroundColor="Blue"/>
      <highlight-row backgroundColor="NoChange" condition="contains('${message}','Consumer0004')" foregroundColor="Cyan"/>
      <highlight-row backgroundColor="NoChange" condition="contains('${message}','Consumer0005')" foregroundColor="Magenta"/>
      <highlight-row backgroundColor="NoChange" condition="contains('${message}','Consumer0006')" foregroundColor="Yellow"/>
    </target>
[...]

PR #1757, opened 2016-11-10
AsyncTargetWrapper - Change default overflow action to blocking

snakefoot:
Change the default behavior of the AsyncWrapper (**async="true"**):

- Default queue-policy changed from discarding to blocking. Adding the async=true should not change the trusted behavior of the wrapped targets.
- Default timer-interval changed from 50 ms to zero (only trigger timer event on activity). Adding the async=true should not add extra overhead in idle periods (avoid waking up application every 50 ms because of needless timer events).

The combination of **timeToSleepBetweenBatches="0"** + **batchSize="200"** + **fullBatchSizeWriteLimit="5"** makes it easy to write more than 200.000 msgs/sec. Thus diminishing the argument for for having a discard-queue-policy.

See also #1789 and #1652 

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1757)

### DIFF
--- a/src/NLog/Targets/Wrappers/AsyncRequestQueue-T.cs
+++ b/src/NLog/Targets/Wrappers/AsyncRequestQueue-T.cs
@@ -42,7 +42,7 @@ namespace NLog.Targets.Wrappers
     /// </summary>
 	internal class AsyncRequestQueue
     {
-        private readonly Queue<AsyncLogEventInfo> logEventInfoQueue = new Queue<AsyncLogEventInfo>();
+        private readonly Queue<AsyncLogEventInfo> logEventInfoQueue = new Queue<AsyncLogEventInfo>(1000);
 
         /// <summary>
         /// Initializes a new instance of the AsyncRequestQueue class.
--- a/src/NLog/Targets/Wrappers/AsyncTargetWrapper.cs
+++ b/src/NLog/Targets/Wrappers/AsyncTargetWrapper.cs
@@ -106,7 +106,7 @@ namespace NLog.Targets.Wrappers
         /// </summary>
         /// <param name="wrappedTarget">The wrapped target.</param>
         public AsyncTargetWrapper(Target wrappedTarget)
-            : this(wrappedTarget, 10000, AsyncTargetWrapperOverflowAction.Discard)
+            : this(wrappedTarget, 10000, AsyncTargetWrapperOverflowAction.Block)
         {
         }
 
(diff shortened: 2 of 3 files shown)


304NotModified:
> Changed default queue-policy to blocking, instead of discarding random log messages. See also #1652

Yes, and the both defaults have their cons and pros. There isn't a default which fits for everyone. So there is not benefit in changing it, but there is con, it's a breaking change we like to avoid. 

> Increased default performance limit from 2000 msgs/sec to 10000 msgs/sec (BatchSize now 500 instead of 100)

I think that is fine to me

snakefoot:
Was thinking that the breaking change part could be included in NLog ver. 5

304NotModified:
Will think about it!

304NotModified:
If we performance test only the change of the batchsize, then we could change it already in NLog 4.4

I would think that 1000 would be better/more logical. But not sure what the performance impact would be.

snakefoot:
You are welcome to make a PR that increases the batch-size alone. This is is mainly for the block-overflow-action :)

This is the result of writing 200.000 messages using a queueSize=10000, and timeToSleepBetweenBatches=50

| BatchSize  | Time (ms) | Msgs/sec  | GC2 | GC1 | GC0 | CPU (ms) | Mem (MB) |
| 100    |    125937 |      1588 |   0 |   2 |   4 |      156 |     90,3 |
| 500        |     26709 |      7487 |   7 |   7 |   7 |      359 |     65,2 |
| 1000       |     13437 |     14883 |   5 |   7 |   8 |      406 |     77,6 |

But remember that AsyncTargetWrapper uses batchSize for both controlling how many items to dequeue at a time, but also for how many items to put into a single buffer.

100 msgs * 100 bytes = 10K buffer (Not LOH)
500 msgs * 100 bytes = 50K buffer (Not LOH)
10000 msgs * 100 bytes = 500K buffer (Large-object-heap-LOH issues)

snakefoot:
Maybe we could add a new setting, called fullBatchMaxCount. That specifies how many full batches to take in one timer-thread-operation. This could just be a simple for-loop, that continues while full batches are dequeued.

Would could then set batchSize = 100 and fullBatchMaxCount = 10. Then we would have the best of both worlds.

304NotModified:
> Maybe we could add a new setting, called fullBatchMaxCount. That specifies how many full batches to take in one timer-thread-operation. This could just be a simple for-loop, that continues while full batches are dequeued.
> 
> Would then set batchSize = 100 and fullBatchMaxCount = 10. Then we would have the best of both worlds (Writing 1000 elements every 50 ms using small buffers).

Sounds like a great idea! Maybe changing the name, but the change is very nice :)

304NotModified:
Gonna close this as I won't merge it ;)

snakefoot:
I have reconsidered regarding increasing BatchSize from 100 to 500. Instead I think adding a FullBatchSizeWriteLimit=10 on AsyncTargetWrapper is better (See also #1799).

Now the PR is just about changing the default behavior as discussed in #1789 and #1652

snakefoot:
After discovering the huge amount of different NLog targets (Where many depends on unstable network traffic). Then I think it is best to keep the default at discard.

304NotModified:
Could you please elaborate on that?

snakefoot:
Network means latency and timeout. Using grow or block in this case will kill the entire application.


Sent from my Samsung device


304NotModified:
Got it! Thanks!

Issue #1741, opened 2016-11-02
Make it possible to inject JSONSerializer


304NotModified:
So we could make a new package, NLog.JsonNet which injects json.net

We need a (simple) custom interface and should expand autoloading packages. (see `ConfigurationItemFactory.BuildDefaultFactory()`)

Needed for JsonLayout and Webservice targets (see #1740 )

304NotModified:
test

304NotModified:
tet2

304NotModified:
done by https://github.com/NLog/NLog/pull/1740

PR #1763, opened 2016-11-14
JsonLayout - Use StringBuilder instead of string append

snakefoot:
Reduce memory allocations when using JsonLayout (Includes changes from #1754)
 - Removed unnecessary string append allocations when doing final wrapping in { and } 
 - Skip allocation of StringBuilder when Json value doesn't need to be escaped.
 - Skip allocation of JsonEncodeLayoutRendererWrapper for every log-event (Now cached on JsonAttribute, will also activate initialSize for LayoutRenderer-StringBuilder).
 - Skip allocation of LogEvent.Properties-Dictionary, when it has no properties.
 - Reuse same JsonAttribute for all dynamic attributes found on LogEvent.Properties
 - Faster type-checking for dynamic attributes found on LogEvent.Properties.

If accepted then please merge #1754 first, and then I will correct any merge-conflicts in this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1763)

### DIFF
--- a/src/NLog/LayoutRenderers/Wrappers/JsonEncodeLayoutRendererWrapper.cs
+++ b/src/NLog/LayoutRenderers/Wrappers/JsonEncodeLayoutRendererWrapper.cs
@@ -35,9 +35,7 @@ namespace NLog.LayoutRenderers.Wrappers
 {
     using System;
     using System.ComponentModel;
-    using System.Globalization;
     using System.Text;
-    using System.Xml;
     using NLog.Config;
 
     /// <summary>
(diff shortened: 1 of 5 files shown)


304NotModified:
Thanks!

If I merge first https://github.com/NLog/NLog/pull/1754, do you think this PR is still mergable?

snakefoot:
They will conflict, but I will merge this PR with master and resolve the conflicts.

304NotModified:
👍

snakefoot:
After having looked at the code for handling LogEvent.Properties. Then I think it is a good idea to use the JSON-encoder from #1740 to also allow handling of Dictionary- and List-objects within the LogEvent.Properties. No need to perform explicit formatting using {event-properties}-layout.

304NotModified:
Agree :)

snakefoot:
After having looked at the JSON-encoder in #1740, then I think it is a good idea just to keep the simple encoder in the JsonLayout and live with the limits.

Then a later iteration can follow the dream of creating a factory that can return a Json-Serializer-Interface, where an extension can register its custom Json-Serializer (Ex. based on Json.net, see also #1741)

304NotModified:
Conflict :(

snakefoot:
Resolved conflicts as promised :)

304NotModified:
Thanks!!

PR #1768, opened 2016-11-19
ExceptionLayoutRenderer - Added support for AggregateException

### LINKED ISSUE #1738: How to expand/flatten an AggregateException using the layout
Feature Request
 - Why do we need it?
When more than one sub task has throws an exception an AggregateException is formed with each exception stored in the property, AggregateException.InnerExecption**s** (note the s)
At present I can only get NLog to recursively log the InnerExpection (singular)

Given the all the TPL programming these days, I expect other people to run into the same problem where exception data is lost when multiple sub tasks throw exceptions.

**NLog version**: 4.3.10
**Platform**: .Net 4.6
**Current NLog config** C# -> See attach unit test

- What is the current result?
Recursively logs only the InnerException (singular)

Handles the AggregateException.InnerException**s** property to log all inner exceptions.

- Are there any workarounds? 
No, other than manually expanding out the InnerExceptions and writing them separately.

No

Done, see attached. [AggregateExceptionTests.zip] https://github.com/NLog/NLog/files/567541/AggregateExceptionTests.zip)

snakefoot:
Renders all AggregateException.InnerExceptions like the current InnerException-chain. Fixes #1738

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1768)

### DIFF
--- a/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
@@ -36,7 +36,6 @@ namespace NLog.LayoutRenderers
     using System;
     using System.Collections.Generic;
     using System.ComponentModel;
-    using System.Globalization;
     using System.Text;
     using NLog.Common;
     using NLog.Config;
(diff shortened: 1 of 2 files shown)


304NotModified:
Cool & Thanks! Another nice feature for NLog  👍

PR #1769, opened 2016-11-19
refs #1498 keeping variables during configuration reload

### LINKED ISSUE #1498: Don't reset Variables when autoReload triggers
This is a Feature request.
NLog version 4.3.4
Platform: net45

```
<nlog autoReload="true" throwExceptions="true" throwConfigExceptions="true" internalLogFile="nlog.txt" xmlns="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <variable name="Batch" value="" />
  <targets>
    <target name="allLog" xsi:type="File" fileName="Log Batch-${var:Batch}.txt" layout="${longdate}|${level:uppercase=true}|${logger}|${callsite}|${gdc:item=global}|${ndc:separator=->}|${message}" />
    <target name="errorLog" xsi:type="File" fileName="Errors Batch-${var:Batch}.txt" />
  </targets>
  <rules>
    <logger name="LogTest.Form1" writeTo="allLog" />
    <logger name="LogTest.Form1" minLevel="Error" writeTo="errorLog" />
  </rules>
</nlog>
```

Current configuration:

When i change the configuration file, the variables get reset, which means my filenames are off.

I request that on an autoreload, the variables (or at least the ones without values in the config) are taken from the current configuration

nazim9214:
fixes #1498 

On configuration reload - variables with empty values/missing variables are taken from current configuration.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1769)

### DIFF
--- a/src/NLog/Config/LoggingConfiguration.cs
+++ b/src/NLog/Config/LoggingConfiguration.cs
@@ -580,5 +580,22 @@ namespace NLog.Config
             }
             return items.ToList();
         }
+
+        /// <summary>
+        /// Copies missing variables and variables with an empty value, from provided dictionary, 
+        /// into current configuration variables dictionary.
+        /// </summary>
+        /// <param name="newVariables">New variables dictionary</param>
+        internal void MergeVariables(IDictionary<string, SimpleLayout> newVariables)
+        {
+            foreach (var variableKey in newVariables.Keys)
+            {
+                if (!this.Variables.ContainsKey(variableKey) 
+                    || string.IsNullOrEmpty(this.Variables[variableKey].OriginalText))
+                {
+                    Variables[variableKey] = newVariables[variableKey];
+                }
+            }
+        }
     }
 }
--- a/src/NLog/Config/LoggingConfiguration.cs
+++ b/src/NLog/Config/LoggingConfiguration.cs
@@ -582,19 +582,14 @@ namespace NLog.Config
         }
 
         /// <summary>
-        /// Copies missing variables and variables with an empty value, from provided dictionary, 
-        /// into current configuration variables dictionary.
+        /// Copies all variables from provided dictionary into current configuration variables. 
         /// </summary>
-        /// <param name="newVariables">New variables dictionary</param>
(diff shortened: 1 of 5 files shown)


304NotModified:
Thanks! 

I'm doubting, shouldnt this be optional?

nazim9214:
Hmm, it's good idea! What do you think, should it be per config level (i think at 'nlog' element level) or per variable (maybe 'keepOnReload' attribute)?

304NotModified:
I guess on NLog level will be sufficient.

304NotModified:
Thanks for the change! 

Could you also add an unit test for `keepVariablesOnReload=false`? Thx in advance!

nazim9214:
I have added unit test for keepVariablesOnReload=false. Also i changed implementation compared to previous. Now, if keepVariablesOnReload is provided, we copy all variables from current configuration into reloaded (new variables are remain).

nazim9214:
In current implementation, when keepVariablesOnReload attribute isn't specified on reloaded config, i get this value from previous configuration. Is this a good decision?

304NotModified:
Sounds good to me:)

304NotModified:
Reviewed 4 of 5 files at r2, 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1769)*

304NotModified:
Thanks! It has been merged and will be released in 4.4

Could you add some docs to the wiki? Please add "introduced in NLog 4.4". Thanks in advance!

nazim9214:
I have added some documentation, can you check it please? (i'm not sure in my english)
https://github.com/NLog/NLog/wiki/Configuration-file/_compare/8453ae91b65e3a650b3efe7de87c6c4b5b598eff...4d385fe3b66b27aaaf05de1b5a1c2af80116f1e7

304NotModified:
Thanks! Changed it just a bit!

Issue #1513, opened 2016-06-21
Writing to the same file from multiple processes does not function when archiving with concurrentWrites="true" and keepFileOpen="true"


thihasoehtet:
**Type**: Bug
**NLog version**: 4.3.5
**Platform**: .Net 4.5
**Current NLog config**

``` xml
  <nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
    <targets>
      <target name="file"
              xsi:type="File"
              layout="${longdate} (${threadid}) ${level} [${logger}] ${message}"
              fileName="C:/NLogTest/logfile.log"
              archiveFileName="C:/NLogTest/{##}-logfile.log"
              archiveAboveSize="100000"
              archiveEvery="Minute"
              archiveNumbering="DateAndSequence"
              archiveDateFormat="yyyyMMdd"
              maxArchiveFiles="30"
              concurrentWrites="true"
              keepFileOpen="true"
              encoding="iso-8859-2" />
      <target name="Console" type="console" />
    </targets>
    <rules>
      <logger name="*" minlevel="Debug" writeTo="console,file" />
    </rules>
  </nlog>
```

**Test Program**

``` cs
using System;
using System.Diagnostics;
using System.Reflection;
using System.Threading;
using NLog;

namespace ConcurrentFileTargetTest
{
  class Program
  {
    private static readonly ILogger Logger = LogManager.GetCurrentClassLogger();
    static void Main(string[] args)
    {
      string processName;
      if (args.Length == 0)
      {
        processName = "0";
        for (int i = 1; i < 5; ++ i)
        {
          Process.Start(Assembly.GetEntryAssembly().Location, i.ToString());
        }
      }
      else
      {
        processName = args[0];
      }

      bool running = true;
      Console.CancelKeyPress += (sender, eventArgs) => running = false; // shutdown on Control-C
      while (running)
      {
        Logger.Debug("{0} - Happily Running Still", processName);
        Thread.Sleep(1000);
      }
    }
  }
}
```

**What is the current result?**
This is similar to an old issue [#1029].
After archiving, one process is writing to new log file (logfile.log) but other processes are writing to a separate archived log files, likes **process1** is writing to the **logfile.log** but **process2** is writing to **20160620.01-logfile.log** and **process3** is writing to **20160620.02-logfile.log**.
It is able to reproduce with the same test application of the old issue[#1029] and seems
[...]

PR #1771, opened 2016-11-21
FileTarget - Added extra archive check is needed, after closing stale file handles

snakefoot:
As shown by #1513, then it is possible for two (or more) processes to detect that a file should be archived.

There is no protection against the case that one process is faster than the others, and performs the archive and creates the new file with its new entry. The next process will then archive the newly created file, and create its own new file.

To protect against this, then a bonus archive-is-needed-check is done after having taken the archive-mutex (And closed stale file handles). This will ensure the next process doesn't succeed in performing the archive operation (will see the archive operation has been done).

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1771)

### DIFF
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -1727,7 +1727,9 @@ namespace NLog.Targets
             var fileInfo = new FileInfo(fileName);
             if (!fileInfo.Exists)
             {
-                this.fileAppenderCache.CloseAppenders(); // Weird, better close our own stale file handles
+                // Close possible stale file handles
+                this.fileAppenderCache.InvalidateAppender(fileName);
+                this.initializedFiles.Remove(fileName);
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


snakefoot:
Little curious about the disconnect between FileTarget.CleanupInitializedFiles() and FileTarget.fileAppenderCache.CloseAppenders().

I guess the File-Footer-Logic is not intended to be used with ConcurrentWrites.

304NotModified:
> Little curious about the disconnect between FileTarget.CleanupInitializedFiles() and FileTarget.fileAppenderCache.CloseAppenders().

Think that's wrong.

> I guess the File-Footer-Logic is not intended to be used with ConcurrentWrites.

Guess it's forgotten ;)

snakefoot:
Have now changed the call from CloseAppenders() to just invalidating a single filename (And also cleanup the InitializedFiles-map)

304NotModified:
👍 looks good!

304NotModified:
Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1771)*

snakefoot:
After looking at the code a little more. Then it seems that the whole mutex-thing (along with the new extra bonus check) should be moved closer down to FileTarget.ArchiveFile.

The archive-mutex only helps while writing to the log-file, but doesn't protect against races at opening/closing the FileTarget. This seems to be an issue independent of using keepFileOpen or not. But there is a lot bonus-archive logic (Ex. rolling log-files), that also might need protection of the archive-mutex. 

But I think the possible multi-process archive-issues at startup/shutdown are for another PR. And not sure I will be making it :)

304NotModified:
> But I think the possible multi-process archive-issues at startup/shutdown are for another PR. And not sure I will be making it :

Agree on that. Maybe we should create a github issue

Thanks again!

304NotModified:
Does this PR fixes https://github.com/NLog/NLog/pull/1771?

304NotModified:
Thanks!

PR #1779, opened 2016-11-22
Improve performance of filters (2-3 x faster)

snakefoot:
ExpressionFilters - Compile MethodInfo into LateBoundMethod-delegate
===

Using the technique from Nate Kohari:

https://web.archive.org/web/20140226131612/http://kohari.org/2009/03/06/fast-late-bound-invocation-with-expression-trees

Then the MethodInfo is compiled into a delegate, that is much faster. Manual handling of default-/optional-parameters are required.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1779)

### DIFF
--- a/src/NLog/Internal/ReflectionHelpers.cs
+++ b/src/NLog/Internal/ReflectionHelpers.cs
@@ -35,8 +35,8 @@ namespace NLog.Internal
 {
     using System;
     using System.Collections.Generic;
+    using System.Linq.Expressions;
     using System.Reflection;
-    using System.Text;
     using NLog.Common;
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


304NotModified:
Impressive!

304NotModified:
Any idea how much faster?

snakefoot:
| Test | Time | Msgs/sec | GC2 | GC1 | GC0 |
------|------|-----------|------|-----|-------|
| No Filter | 2565ms | 779518 | 39 | 50 | 57 |
| Filter | 9789ms | 204291 | 68 | 93 | 117 |
| LateBound | 3693ms | 541445 | 36 | 54 | 61 |

304NotModified:
\0/

src/NLog/Internal/ReflectionHelpers.cs
```diff
@@ -92,6 +92,72 @@ public static bool IsStaticClass(this Type type)
...
+                    parameterType = parameterType.GetElementType();
+
+                var valueCast = Expression.Convert(valueObj, parameterType);
+
+                parameterExpressions.Add(valueCast);
+            }
+
+            // non-instance for static method, or ((TInstance)instance)
+            var instanceCast = methodInfo.IsStatic ? null :
+                Expression.Convert(instanceParameter, methodInfo.ReflectedType);
```
304NotModified: @snakefoot 

another netstandard challenge, `methodInfo.ReflectedType` isn't there. Any idea how to fix it? I dunno what the `ReflectedType` is in this case.

snakefoot: Seems that we should be using DeclaringType instead of ReflectedType

PR #1785, opened 2016-11-23
Increase default QueueLimit and BatchSize for async / asyncwrapper

304NotModified:
- Batchsize from 100 -> 500
- queue limit from 10.000 -> 100.000 

Fixes https://github.com/NLog/NLog/issues/1283

@snakefoot your opinion please. :)

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1785)

### DIFF
--- a/src/NLog/Targets/Wrappers/AsyncTargetWrapper.cs
+++ b/src/NLog/Targets/Wrappers/AsyncTargetWrapper.cs
@@ -107,8 +107,8 @@ namespace NLog.Targets.Wrappers
         /// </summary>
         /// <param name="wrappedTarget">The wrapped target.</param>
         public AsyncTargetWrapper(Target wrappedTarget)
-            : this(wrappedTarget, 10000, AsyncTargetWrapperOverflowAction.Discard)
-        {
+            : this(wrappedTarget, 100000, AsyncTargetWrapperOverflowAction.Discard)
+        {                         
         }
 
         /// <summary>
@@ -119,9 +119,9 @@ namespace NLog.Targets.Wrappers
         /// <param name="overflowAction">The action to be taken when the queue overflows.</param>
         public AsyncTargetWrapper(Target wrappedTarget, int queueLimit, AsyncTargetWrapperOverflowAction overflowAction)
         {
-            this.RequestQueue = new AsyncRequestQueue(10000, AsyncTargetWrapperOverflowAction.Discard);
+            this.RequestQueue = new AsyncRequestQueue(100000, AsyncTargetWrapperOverflowAction.Discard);
             this.TimeToSleepBetweenBatches = 50;
-            this.BatchSize = 100;
+            this.BatchSize = 500;
             this.WrappedTarget = wrappedTarget;
             this.QueueLimit = queueLimit;
             this.OverflowAction = overflowAction;
@@ -132,7 +132,7 @@ namespace NLog.Targets.Wrappers
         /// by the lazy writer thread.
         /// </summary>
         /// <docgen category='Buffering Options' order='100' />
-        [DefaultValue(100)]
+        [DefaultValue(500)]
         public int BatchSize { get; set; }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


snakefoot:
I think a queue-limit of 10000 is enough for 99.999%. Using 100000 will hit the Large-Object-Heap, which is unnecessary for most. When testing async-performance then it actually dropped for me when over 30000.

304NotModified:
Well, it's a Max size and it still discarded by default.

I was a recommendation here: https://github.com/NLog/NLog/issues/1283 (but maybe I misinterpreted it)

snakefoot:
Yes I can see that now. Strange it doesn't give an initial capacity to AsyncRequestQueue.logEventInfoQueue.

Still if you care about memory usage, and not having NLog hogging the whole system, then having 100.000 queued strings will not help.

304NotModified:
> The funny part is that the case is a block-overflow, and the default is discard :)

oops good point. 

But ok, 100.000 string, 

> Getting 100.000 queued messages because a target receives 2min network-timeout for every log-event is not a good thing.

true, but given that a string is 100 bytes (100 UTF8 chars), that is with 100.000 string below 10 MB? , or am I missing something?

snakefoot:
Well you seem convinced. No issue for me, as I'm not using the default values as they are not working for me (Ex. overflow=block, batchsize=1000).

304NotModified:
> Well you seem convinced. No issue for me, as I'm not using the default values as they are not working for me (Ex. overflow=block, batchsize=1000).

A bit. I was more curious if this is an improvement for the async-is-discarding-stuff-by-default in your opinion.

snakefoot:
Well my application is also a rare animal as garbage-collections should be avoided. Having large queues hurts latency, increases CPU-L3-cache-misses and increases chance of being pushed to GC1/GC2.

304NotModified:
So this is no improvement in your opinion?

snakefoot:
BatchSize=500 is good for performance. QueueSize=100000 doesn't make sense to me when it comes to performance. But yes a higher QueueSize will lower the chance of discarding messages, but I prefer to completely sure and just use overflow=block.

snakefoot:
Just made a simple performance test using this config:

```
      <target name="mipLoggertarget" xsi:type="AsyncWrapper" overflowAction="Block" queueLimit="100000" batchSize="500" timeToSleepBetweenBatches="0">
        <target name ="mipLoggertargetWrapped" xsi:type="File"
                fileName="c:/temp/Log/test1-${date:format=yyyy-MM-dd-HH}.log" FileNameKind="Absolute"
                layout="${longdate} ${uppercase:${level}} ${message} ${exception}"
                keepFileOpen="true" AutoFlush="true" ConcurrentWrites="false" CleanupFileName="false" />
      </target>
```

| QueueSize  | Time (ms)  | Msgs/sec   | GC2 | GC1 | GC0 |
|------------|------------|------------|-----|-----|-----|
| 10000      |      22462 |     445177 | 277 | 424 | 434 |
| 100000     |      27573 |     362668 | 259 | 309 | 318 |

Performance is better with 10000, but funny that GC2 numbers are higher for 10000 :)

304NotModified:
> but funny that GC2 numbers are higher for 10000 :)

also GC0 and G1

304NotModified:
Ok, not a great win for now. We need another proposal.

snakefoot:
There is a minor detail regarding all this testing, and that is the Flush-operation. When having a QueueSize of 100.000 then the final flush will give a boost. This might have affected the testing in #1283

My numbers writting 200.000 messages using BatchSize=500 and timeToSleepBetweenBatches=50, where I have injected an exception into the last logevent, and when that is rendered, then the Flush is performed.

| QueueSize  | Time (ms) | Msgs/sec  | GC2 | GC1 | GC0 | CPU (ms) | Mem (MB) |
| 10000      |     25109 |      7964 |   5 |   6 |   6 |      234 |     73,1 |
| 100000     |     25147 |      7953 |   6 |   7 |   8 |      640 |      119 |

Just looks like the 100000 is using more CPU and more memory, but no performance gains.

304NotModified:
Good point! Thanks for testing!

PR #1807, opened 2016-12-01
FlushTriggerWrapper is implemented

### LINKED ISSUE #1712: Add flushCondition and overflow options to the  BufferingWrapper
As discussed at #1710, add flushCondition and overflow options to the  BufferingWrapper

Example: 

``` xml
<targets>
  <target xsi:type="BufferingWrapper"
          name="ErrorFlushingBuffer"
          slidingTimeout="False"
          bufferSize="100"
          flushTimeout="-1"  <!-- Disabled timed flushes so flush only occurs when flushCondition is met -->
          overflow="discard"> <!-- discard if buffer is full. Default: flush-->
          flushCondition="level>LogLevel.Info">  <!-- Flush the buffer when a Warn, Error, or Fatal message is logged-->

          <target xsi:type="wrappedTargetType" ...target properties... />
  </target>
</targets>
```

nazim9214:
I needed a way to automatically flushing to a File target(autoflush=false, keepOpen=true, concurrentWrites=false), only for special LogLevels(>=Warn). 
After some research i have found this [idea](https://github.com/NLog/NLog/issues/1712) of FlushTriggerWrapper and implemented this wrapper.

fixes #1712

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1807)

### DIFF
--- a/src/NLog/NLog.Xamarin.Android.csproj
+++ b/src/NLog/NLog.Xamarin.Android.csproj
@@ -390,6 +390,7 @@
     <Compile Include="Targets\Wrappers\FallbackGroupTarget.cs" />
     <Compile Include="Targets\Wrappers\FilteringRule.cs" />
     <Compile Include="Targets\Wrappers\FilteringTargetWrapper.cs" />
+    <Compile Include="Targets\Wrappers\FlushTriggerWrapper.cs" />
     <Compile Include="Targets\Wrappers\ImpersonatingTargetWrapper.cs" />
     <Compile Include="Targets\Wrappers\LimitingTargetWrapper.cs" />
     <Compile Include="Targets\Wrappers\LogOnProviderType.cs" />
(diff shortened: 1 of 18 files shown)


nazim9214:
I'm not sure, but we could add Condition property (not required) to AutoFlushTargetWrapper. Or it's better to be separate wrapper?

304NotModified:
Cool ! Looks good!


> I'm not sure, but we could add Condition property (not required) to AutoFlushTargetWrapper. Or it's better to be separate wrapper?

I think that's even a better option! What do you think @snakefoot ?

snakefoot:
As long the flush-triger-logic is not located on the BufferingTargetWrapper, then I'm good :)

But yes adding a condition-filter on the AutoFlushTargetWrapper would also work.

304NotModified:
>Would also work

not better?

304NotModified:
Ps. Rebuild for codecov.

snakefoot:
It will probably be better for the NLog users to have just one wrapper, that is dedicated to flushing.

nazim9214:
Ok, i'll move code to AutoFlushTargetWrapper this weekend

304NotModified:
👍 

please open a new PR then, thanks!

Issue #1673, opened 2016-10-04
Add MaximumKilobytes property for eventlog


Sing303:
It is impossible to change the setting MaximumKilobytes in eventlog. Please add this parameter

304NotModified:
Could you explain your question?

There is a `MaxMessageLength` property on the EventLog, but that is not what you mean?

Sing303:
No, I want to change this property
https://msdn.microsoft.com/en-us/library/system.diagnostics.eventlog.maximumkilobytes(v=vs.110).aspx

304NotModified:
You could do that with c#  of course. But you like to configure this from the XML config? How would the config look like?

Sing303:
Something like this
```
<targets>
  <target xsi:type="EventLog"
          name="String"
          layout="Layout"
          machineName="String"
          source="Layout" 
          category="Layout"
          eventId="Layout"
          log="String"
          maxMessageLength="Integer"
          maximumKilobytes="Integer"/>
</targets>
```

304NotModified:
And then we set the `maximumKilobytes` on first usage? (init)

Sing303:
Yes exactly

ajitpeter:
I  would like to implement this feature, I hope somebody could guide me through the process.

304NotModified:
Cool!

Will give some more info tomorrow!

304NotModified:
Firstly, please read https://github.com/NLog/NLog/blob/master/CONTRIBUTING.md

1. start with the NLog.netfx45.sln, in folder src

2. You need a new property on [EventLogTarget](https://github.com/NLog/NLog/blob/master/src/NLog/Targets/EventLogTarget.cs) (probably Layout type) 
3. Use the property OR in `Write` or in `Install` (latter if is only needed once or admin rights needed). Use YourProperty.Render(event) (and maybe `int.TryParse`)
4. Add new tests in [`EventLogTargetTests`](https://github.com/NLog/NLog/blob/master/tests/NLog.UnitTests/Targets/EventLogTargetTests.cs)
5. If you add new files (not this case), sync the .sln files, see https://github.com/NLog/NLog/blob/master/CONTRIBUTING.md#sync-projects

If you have more questions, let us know!

ajitpeter:
Thanks for information. I will start on it and surely will be in contact if I face any issues/questions.

ajitpeter:
[...]

Issue #1234, opened 2016-02-11
Layout everything 


304NotModified:
I have seen multiple requests for layout renders which have non-layoutable properties (eg string,  int) . 

I would be nice if every property would be layoutable.  

This is breaking change and needs some performance tests / tweaks. We don't expect a big performance change, see https://github.com/NLog/NLog/issues/1228

ps: There is a [`SimpleLayout .Escape(string)` method](https://github.com/NLog/NLog/blob/1899e6a4a5ab4ee0b23129d5b950cad519db19b4/src/NLog/Layouts/SimpleLayout.cs#L171). Maybe we can use that for backwards-compatibly. (e.g. when setting reading the XML)

304NotModified:
Proposal: create a generic Layout and default layouts for all primitive types (only if used)

e.g.

- IntLayout
- LongLayout
- ShortLayout
- BoolLayout

also probably needed 

- DateTimeLayout (LimitingTargetWrapper)
- EnumLayout<TEnum>


etc.

Would be nice if we could easier escape the values, e.g. single quotes don't get expanded (`'${myvar}'` vs `"${myvar}"` in the XML config

304NotModified:
@snakefoot any opinion on https://github.com/NLog/NLog/pull/3119? I like to continue on this one.

PS: there is also a really another approach possible, revert all the properties to just `string`, `int` etc and add a proxy object between the layout renderers - before `write`, it will update all properties. But not sure if we could get that nice in terms of performance.

304NotModified:
@snakefoot It this a no opinion? 👼

snakefoot:
> @304NotModified It this a no opinion?

The issue is that I cannot see any perfect solutions. Already now the NLog-Layout logic is causing a lot confusion for new developers. Have now dived through many NLog-targets and pretty much all them was doing it wrong. Some have been fixed, but all new NLog-targets will probably also do it wrong.

My biggest worry was the performance and memory-allocation-overhead. But after having implemented `IStringValueRenderer` then the overhead has been reduced for static strings. And with some extra effort then it can also be applied to `When` and `WhenEmpty`. Just need to make sure one doesn't call `LogEventInfo.CreateNullEvent()` constantly.

I'm fine with trying to convert all string-properties to Layout. But more care has to be applied to int/enum-properties.

Also the `Layout`-type makes it less obvious what values can be assigned. This will affect the chance of creating an intuitive fluent-config-api as one have to pay attention to make property-assignment natural.

[...]

PR #1810, opened 2016-12-03
EventlogTarget: Support for MaximumKilobytes

ajitpeter:
First PR.  Issue #1673 

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1810)


304NotModified:
Hey thanks for the PR! Will review it!

src/NLog/Targets/EventLogTarget.cs
```diff
@@ -158,6 +158,11 @@ public int MaxMessageLength
         }
 
         /// <summary>
+        /// Gets or sets the maximum Event log size in kilobytes.
+        /// </summary>
+        public Layout MaxKilobytes { get; set; }
```
304NotModified: looks good

src/NLog/Targets/EventLogTarget.cs
```diff
@@ -260,8 +265,17 @@ protected override void Write(LogEventInfo logEvent)
                 category = Convert.ToInt16(this.Category.Render(logEvent), CultureInfo.InvariantCulture);
             }
 
+            long maxKilobytes = 0;
+
+            if(this.MaxKilobytes != null)
+            {
+                maxKilobytes = Convert.ToInt64(this.MaxKilobytes.Render(logEvent), CultureInfo.InvariantCulture);
```
304NotModified: maybe it's a bit odd to set this property, when writing every message?

First I was thinking to move it to `InitializeTarget`, but then you have a problem with `GetEventLog(logEvent)`.

Maybe we could cache the value? I was think to move it to `GetEventLog` and maybe creating a method for `isCacheUpToDate `

304NotModified:
I don't know why the unit tests are failing.

src/NLog/Targets/EventLogTarget.cs
```diff
@@ -260,8 +265,17 @@ protected override void Write(LogEventInfo logEvent)
...
+            }
+
             EventLog eventLog = GetEventLog(logEvent);
 
+            eventLog.MaximumKilobytes = maxKilobytes;
```
304NotModified: if `this.MaxKilobytes` is not set, then this will set it to 0. ;)

Therefor the failing unit tests

snakefoot:
Curious why you use  a Layout-object for setting an integer value. Is it some special trick?

304NotModified:
> Curious why you use a Layout-object for parsing an integer value. Is it some special trick?

Do you mean this line:

` Convert.ToInt64(this.MaxKilobytes.Render(logEvent), CultureInfo.InvariantCulture);`

?

snakefoot:
Yes having an integer property declared as "Layout MaxKilobytes" and then using rendering and parsing to convert it back to an integer.

src/NLog/Targets/EventLogTarget.cs
ajitpeter: I was wondering why the unit tests on my local enviroment did not fail?? Is there any way to simulate the unit tests on appveyor without creating a pull request??

304NotModified:
> Yes having an integer property declared as "Layout MaxKilobytes" and then using rendering and parsing to convert it back to an integer.

Well then you could use layout renderers `${var:myvar}`, `${gdc:eventlogmax}` etc.

I have to agree it is not perfect way, but the best way in NLog 4. Related #1234

snakefoot:
> I have to agree it is not perfect way, but the best way in NLog 4

Yes it would make more sense that the correct type was declared for the property, and then the XML-parsing of values was able to recognize Layout-logic and automatically render/parse to the correct property-value-type.

304NotModified:
> Yes it would make more sense that the correct type was declared for the property, and then the XML-parsing of values was able to recognize Layout-logic and automatically render/parse to the correct property-value-type.

That is possible, but then the property value is not evaluated runtime. 

I was thinking of `Layout<Int>` in the future

snakefoot:
> That is possible, but then the property value is not evaluated runtime.

If the configuration-logic was able to set the correct properties at startup, then the configuration-logic should also be able to set the correct properties on reload (Without using Layout-properties). Would also require some logic regarding perform Reinitialize on a running system.

304NotModified:
@snakefoot let move the discussion to https://github.com/NLog/NLog/issues/1811, thx! :)

304NotModified:
@ajitpeter need some help on this?

304NotModified:
It's now doing nothing ;)

ajitpeter:
@304NotModified  Sorry. These were not intended to be included in the pull request. I was trying out appveyor. I was also trying out what you had mentioned and also debugging simultaneously on my local branch. :-)

src/NLog/Targets/EventLogTarget.cs
```diff
@@ -352,6 +357,7 @@ private EventLog GetEventLog(LogEventInfo logEvent)
             var renderedSource = this.Source != null ? this.Source.Render(logEvent) : null;
             var isCacheUpToDate = eventLogInstance != null && renderedSource == eventLogInstance.Source &&
                                    eventLogInstance.Log == this.Log && eventLogInstance.MachineName == this.MachineName;
+            eventLogInstance.MaximumKilobytes = Convert.ToInt64(this.MaxKilobytes.Render(logEvent));
```
304NotModified: This is problematic when the value is empty/0 ;)

304NotModified:
@ajitpeter needed some help on this?

ajitpeter:
Yes please. I am struggling to replicate the test failure in my local debug environment. Any help would be much appreciated.

304NotModified:
If all the event log test failing locally , run as admin.    

If the test succeeds locally then it's probably due to information leftovers in the event log. 

Another way is to check the code review and just commit and push.

304NotModified:
Change looks good. I do prefer auto properties though. 

Please document the 512 (as it's a default of MS - https://msdn.microsoft.com/en-us/library/system.diagnostics.eventlog.maximumkilobytes(v=vs.110).aspx) 

And we need some unit tests.

304NotModified:
close/repopen for trigger CI

304NotModified:
There was a stackoverflow in your code ;) 

Have removed that one. Besides that, looks good to me. If it all builds, I will merge it. 

Thanks!

ajitpeter:
Thanks for your help. Much appreciated. :-)

304NotModified:
Unfortunately the build still fails. Dunno why

UgurAldanmaz:
[From MSDN](https://msdn.microsoft.com/library/system.diagnostics.eventlog.maximumkilobytes(v=vs.110).aspx):

> The MaximumKilobytes property represents the size limit of the event log file. When the event log reaches the size limit, the configured OverflowAction value determines whether new entries are discarded, or whether new entries overwrite older entries.

[MSDN says](https://msdn.microsoft.com/en-us/library/system.diagnostics.eventlog.overflowaction(v=vs.110).aspx) default value of `OverflowAction` is `OverwriteOlder`.

According to test fail messages, logs may are not written. Because they cannot be found when asserting written test log entry.

I suspect if Retention regedit value of the CI machine is `-1`. Check the [OverflowAction source code](https://referencesource.microsoft.com/#System/services/monitoring/system/diagnosticts/EventLogInternal.cs,400).

Tests use the same event log that named as `application`. Application event log is a general instance. So there are maybe so much event log entries that event log file reaches the size limit, then older entries are not overwritten. As a result, unit tests fail.

UgurAldanmaz:
@ajitpeter,

I am not sure at my previous comment. But I suggest you to try. You can write unit tests for your pull request. You just need to supply a unique event log.

* Check the [CreateEventLogTarget](https://github.com/NLog/NLog/blob/ee1d72b6658aa724c6c36f675eb0e1b546aca0cd/tests/NLog.UnitTests/Targets/EventLogTargetTests.cs#L453). You can overload this method to pass log name.
* Call this method with random log name.
* Write your unit tests that use this randomly named event log.

304NotModified:
Thanks for the info @UgurAldanmaz ,

but maybe i'm missing something. The default is 512 (see [here](https://github.com/NLog/NLog/pull/1810/files#diff-e338ce9c5930f3d84803441ea1c37afeR90) and [here](https://msdn.microsoft.com/en-us/library/system.diagnostics.eventlog.maximumkilobytes(v=vs.110).aspx), and no unit test has been changed, so there is not reason that the unit tests suddenly are reporting errors?

304NotModified:
I like to merge this, but anyone knows why this is failing the unit tests?

ajitpeter:
I tried but couldn't figure out the unit test failures.

304NotModified:
it was a nullref, which was swallowed by NLog

fixed in #2114

Issue #531, opened 2015-01-26
Plan: Remove Obsolete


304NotModified:
Proposal for NLog 4, remove the `Obsolete` marked classes & methods:

| What | Alternative | Obsolete since (in code) | Marked as Obsolete in wiki | comments | obsolete text has NLog version |
| --- | --- | --- | --- | --- | --- |
| `GDC` type | `GlobalDiagnosticsContext` type | 2.0 | yes |  |  |
| `MDC type` | `MappedDiagnosticsContext` type | 2.0 | yes |  |  |
| `NDC type` | `NestedDiagnosticsContext` type | 2.0 | yes |  |  |
| `EventContextLayoutRenderer` type | `EventPropertiesLayoutRenderer` type | 2.0 | yes (but recently) |  |  |
| `LogEventInfo.Context` | `LogEventInfo.Properties` | 2.0 | ? |  |  |
| `LoggerShortName` | ?? | 2.0 | ? |  |  |
| `ExceptionLoggingOldStyle` | new loggingstule | 4.1 | not on wiki | added in 4.1 to make move easier from 3 to 4 |  |
| `DisableLogging()` in LogFactory | `SuspendLogging()` | 4.0 |  |  |  |
| `EnableLogging()` in LogFactory | `ResumeLogging()` | 4.0 |  |  |  |
| `FileTarget.maxLogFilenames` | `FileTarget.maxLogFilenames` | not yet | no impact on nlog XML config |  |  |
| `DatabaseTarget.UseTransactions` | don't use it | 4.2.1 | 4.0 | restored with warning in 4.2.1 | yes |

Xharze:
I think it would be a great idea to remove these classes. Hopefully all has either an alternative or shouldn't be used anymore.

304NotModified:
Updated with listing the alternatives

304NotModified:
Note: https://github.com/NLog/NLog/wiki/EventContext-Layout-Renderer not documented as obsolete in wiki

ie-zero:
My proposal is to document the version a function/class or feature became obsolete and agree a certain number of minor version which will remain for example three. 

After the agreed number of version elapse the feature can be removed at any time.

304NotModified:
Thanks for your input!

According to semver, it's only allowed to remove stuff with a major version. 

But we have to sum up what's obsolete (and what is already proper communicated, like wiki and news post)

ie-zero:
[...]

PR #1814, opened 2016-12-04
Improve [Obsolete] warnings - include the Nlog version when it became obsolete

ie-zero:
Document on XML comment as well as the [Obsolete] attribute the NLog
version where a class, method or property was marked as obsolete.

This might help in the future releases to evaluate if an element should
be removed completely from the code and it provides the users with more
information.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1814)


304NotModified:
Good idea.

Maybe it's event better to use the real version (we should find it with git) and update https://github.com/NLog/NLog/issues/531?

304NotModified:
Thanks! Great work!

PR #1824, opened 2016-12-04
Move null logger to separate class to isolate the logic

ie-zero:
Separates the implementation of null logger from the LogFactory into a new NullLogger class to isolate the logic.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1824)


304NotModified:
Good idea! Thx!

PR #1829, opened 2016-12-05
Log4jXmlEvent - Improve performance (CPU & memory)

snakefoot:
Log4jXmlEvent - Replaced Regex-search with XmlConvert.IsXmlChar  and reduced memory allocations

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1829)

### DIFF
--- a/src/NLog/Internal/XmlHelper.cs
+++ b/src/NLog/Internal/XmlHelper.cs
@@ -70,7 +70,7 @@ namespace NLog.Internal
                 char ch = text[i];
                 if (!XmlConvert.IsXmlChar(ch))
                 {
-                    return FilterValidXmlChars(text);   // rare expensive case
+                    return CreateValidXmlString(text);   // rare expensive case
                 }
             }
             return text;
@@ -83,7 +83,7 @@ namespace NLog.Internal
         /// </summary>
         /// <param name="text">unclean string</param>
         /// <returns>string with only valid XML chars</returns>
-        private static string FilterValidXmlChars(string text)
+        private static string CreateValidXmlString(string text)
         {
             var sb = new StringBuilder(text.Length);
             for (int i = 0; i < text.Length; ++i)
--- a/src/NLog/Internal/XmlHelper.cs
+++ b/src/NLog/Internal/XmlHelper.cs
@@ -65,29 +65,38 @@ namespace NLog.Internal
 #if NET3_5
             return InvalidXmlChars.Replace(text, "");
 #else
-            StringBuilder sb = null;
+            for (int i = 0; i < text.Length; ++i)
+            {
+                char ch = text[i];
+                if (!XmlConvert.IsXmlChar(ch))
+                {
+                    return FilterValidXmlChars(text);   // rare expensive case
+                }
+            }
+            return text;
+#endif
+        }
+
+#if !NET3_5
+        /// <summary>
+        /// Cleans string of any invalid XML chars found
+        /// </summary>
+        /// <param name="text">unclean string</param>
+        /// <returns>string with only valid XML chars</returns>
+        private static string FilterValidXmlChars(string text)
+        {
+            var sb = new StringBuilder(text.Length);
             for (int i = 0; i < text.Length; ++i)
             {
                 char ch = text[i];
                 if (XmlConvert.IsXmlChar(ch))
                 {
-                    if (sb != null)
-                        sb.Append(ch);
-                    continue;
-                }
-
-                if (sb == null)
-                {
-                    // Only allocated string-builder, when string is bad
-                    sb = new StringBuilder(text.Length);
-                    for (int j = 0; j < i; ++j)
-                        sb.Append(text[j]);
+                    sb.Append(ch);
                 }
             }
-            return sb != null ? sb.ToString() : text;
-#endif
+            return sb.ToString();
         }
-
+#endif
 
         /// <summary>
         /// Safe version of WriteAttributeString
--- a/src/NLog/Internal/XmlHelper.cs
+++ b/src/NLog/Internal/XmlHelper.cs
@@ -34,6 +34,7 @@
 namespace NLog.Internal
 {
     using System;
+    using System.Text;
     using System.Text.RegularExpressions;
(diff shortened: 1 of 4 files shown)


src/NLog/Internal/XmlHelper.cs
```diff
@@ -58,7 +59,33 @@ public static class XmlHelper
...
+            return InvalidXmlChars.Replace(text, "");
+#else
+            StringBuilder sb = null;
+            for (int i = 0; i < text.Length; ++i)
+            {
+                char ch = text[i];
+                if (XmlConvert.IsXmlChar(ch))
+                {
+                    if (sb != null)
+                        sb.Append(ch);
```
304NotModified: I think it works, but it looks a bit tricky. Could this be improved in readablity?

304NotModified: Also dunno if this performance wise OK. In the best case it is, but not in the worst case, as adding all chars to a stringBuilder in a for loop is less performant then substring.

snakefoot: We could rename sb to badStringRepair ?

snakefoot: Not sure I understand. Every time we skip a string-allocation, then we improve performance.

snakefoot: Then it has to be:

```
                char ch = text[i];
                if (!XmlConvert.IsXmlChar(ch))
                {
                    // Invalid char found - allocated string-builder for cleaning
                    if (sb == null)
                    {
                        //create a fill with already seen chars
                        sb = new StringBuilder(text.Length);
                        for (int j = 0; j < i; ++j) //or substring
                            sb.Append(text[j]);
                    }
                }
                else
                {
                     if (sb != null)
                        sb.Append(ch);
                } 
```

snakefoot:
Writing 1.000.000 messages of 16 bytes:

| Test Name  | Time (ms) | Msgs/sec  | GC2 | GC1 | GC0 | CPU (ms) | Mem (MB) |
| Old Version   |    22.483 |    44.476 |   6 | 446 | 1993 |   24.062 |   42,711 |

| Test Name  | Time (ms) | Msgs/sec  | GC2 | GC1 | GC0 | CPU (ms) | Mem (MB) |
| New Version |   16.748 |    59.707 |  16 | 464 | 1828 |   18.234 |   50,359 |

src/NLog/Internal/XmlHelper.cs
304NotModified: AFAIK substring is in some cases more efficient then for-append, but maybe not in this case.

Anyway, found a better solution, using `StringBuilder.Append(char[] value, int startIndex, int charCount)`

```c#
            for (int i = 0; i < text.Length; ++i)
            {
                char ch = text[i];
                if (!XmlConvert.IsXmlChar(ch))
                {
                    // Invalid char found - allocated string-builder for cleaning
                    if (sb == null)
                    {
                        //create a fill with already seen chars
                        sb = new StringBuilder(text.Length);
                        sb.Append(text, 0, i);
                    }
                    else
                    {
                        sb.Append(ch);
                    }
                }
            }
```

snakefoot: But then I have to allocate a charArray. Anyway no big issue the chance of meeting the forbidden chars are super low. So it will only be a in rare cases. SubString is fine by me.

304NotModified: we have already text, which is a charArray?

304NotModified:
To be honest, I don't like the repetitive check of `sb ==null` from the readability viewpoint. 

I have another proposal, which is clearer IMO

```c#
            for (int i = 0; i < text.Length; ++i)
            {
                char ch = text[i];
                if (!XmlConvert.IsXmlChar(ch))
                {
                    //first invalid, clean whole string
                    text = CleanValue(text, i);
                    break;
                }
            }

        private static string CleanValue(string text, int firstInvalidIndex)
        {
            var sb = new StringBuilder(text.Length);
            //init with valid values
            sb.Append(text, 0, firstInvalidIndex); 
            for (int i = firstInvalidIndex; i < text.Length; ++i) //not sure if this should be firstInvalidIndex+1
            {
                char ch = text[i];
                if (XmlConvert.IsXmlChar(ch))
                {
                    sb.Append(ch);
                }
            }
            return sb.ToString();
        }
```

snakefoot:
Have no opinion about this thing. The invalid-xml-char is a rare case, so whatever works.

Anyways split into two methods.

src/NLog/LayoutRenderers/Log4JXmlEventLayoutRenderer.cs
```diff
@@ -289,9 +303,9 @@ protected override void Append(StringBuilder builder, LogEventInfo logEvent)
                 xtw.WriteAttributeSafeString("name", "log4jmachinename");
 
 #if SILVERLIGHT
-            xtw.WriteAttributeSafeString("value", "silverlight");
+                xtw.WriteAttributeSafeString("value", "silverlight");
```
304NotModified: guess this `#if/else` could be removed ;)

304NotModified:
Thanks!

src/NLog/Internal/XmlHelper.cs
```diff
@@ -58,9 +59,44 @@ public static class XmlHelper
...
+            if (string.IsNullOrEmpty(text))
+                return string.Empty;
+
+#if NET3_5
+            return InvalidXmlChars.Replace(text, "");
+#else
+            for (int i = 0; i < text.Length; ++i)
+            {
+                char ch = text[i];
+                if (!XmlConvert.IsXmlChar(ch))
```
304NotModified: @snakefoot I;m trying to include the changes of NLog 4.4 to NLog 5 (beta)

I seems that this method is not there in NETSTANDARD :(  (not until NETSTANDARD  2 - Q1 2017)

Any idea of an alternative?

304NotModified: Lol missed that one. The!

PR #1839, opened 2016-12-10
Add missing exception to AddTarget method

ie-zero:
In LoggingConfiguration the AddTarget(name, target) method would fail during the call to the InternalLogger if the target was null.

Also the behavior of the AddTarget(target) and AddTarget(name, target) methods are now consistent as they both throw a ArgumentNullException for target parameter.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1839)


304NotModified:
Thanks! Also for the unit tests :+1:

PR #1841, opened 2016-12-10
Move mutex functionality into BaseMutexFileAppender

ie-zero:
Most of the subclasses of BaseFileAppender are using only a subset of its functionality. There are only two classes require the mutex functionality. Therefore, the mutex functionality is moved into a new abstract class (BaseMutexFileAppender) to re-enforce the single responsibility principle.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1841)


304NotModified:
Good improvement! Thanks!

PR #1847, opened 2016-12-12
LogFactory.Dispose() fixed race condition with reloadtimer

### LINKED ISSUE #1846: Possible big memory leak in DomainUnload event
**Type**: Bug
**NLog version**: 4.3.11
**Platform**: .Net 4.5 / .NET Core  and possible others when hosted on IIS

This is not outlined in MSDN documentation but at least in .NET 4.5 and .NET core platforms the domain unload event is always executed on finalizer thread (you can check this in coreclr source code available on GitHub - you will notice, that the domain unload event is being executed by **ProcessUnloadDomainEventOnFinalizeThread** native call used in **finalizerthread.cpp** file). Having this in mind the code placed in DomainUnload event must follow the same restrictions like code in regular class finalizer - for example it cannot ever try to block current thread and wait for something else to wake it up (like ManualResetEvent.WaitOne). Blocking the finalizer thread can lead to a very big and nasty memory leak, because many objects will never be collected by GC in such scenario.

I think that the **ManualResetEvent** instance should be removed from the body of **Dispose(bool)** method from **LogFactory** class, because in case of reloadTimer being not null the call to ManualResetEvent.WaitOne can freeze the thread on which the Dispose(bool) method is being executed. As you may noticed, the Dispose(bool) method of LogFacory class is being called from **LogManager.TurnOffLogging** static method, which is always executed from DomainUnload event. This can lead to a very nasty memory leak because there is a possibility, that the finalizer thread can be frozen when that event fires.

This bug affects only web applications hosted on IIS, because of the way the IIS server handles AppDomains. IIS restarts appdomains (not the whole worker process) when it detects a change in **web.config** file or in the **/bin** directory (for example after new deploy of web application). In such case the domainunload event is fired from appdomain being unloaded (and it executes on the finalizer thread, which is only one per w3wp.exe process) and then the new appdomain is being created. This is the moment when the problem described above can occur.

Current result: there is a possibility that the domain unload event can freeze the finalizer thread thus making a big memory leak in IIS hosted web applications
Expected result: code from domain unload event shall never try to block the thread it is executing on.

snakefoot:
Attempted to fix #1846, but found no easy solution, so just fixed a race-condition.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1847)

### DIFF
--- a/src/NLog/Config/LoggingConfiguration.cs
+++ b/src/NLog/Config/LoggingConfiguration.cs
@@ -490,6 +490,8 @@ namespace NLog.Config
         /// <param name="asyncContinuation">The asynchronous continuation.</param>
         internal void FlushAllTargets(AsyncContinuation asyncContinuation)
         {
+            InternalLogger.Trace("Flushing all targets...");
+
             var uniqueTargets = new List<Target>();
             var loggingRules = this.LoggingRules.ToList();
             foreach (var rule in loggingRules)
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -709,15 +709,19 @@ namespace NLog
 #if !SILVERLIGHT && !__IOS__ && !__ANDROID__
         internal void ReloadConfigOnTimer(object state)
         {
+            if (this.reloadTimer == null && this.IsDisposing)
+                return;
+
             LoggingConfiguration configurationToReload = (LoggingConfiguration)state;
 
             InternalLogger.Info("Reloading configuration...");
             lock (this.syncRoot)
             {
-                if (this.reloadTimer != null)
+                var currentTimer = this.reloadTimer;
+                if (currentTimer != null)
                 {
-                    this.reloadTimer.Dispose();
                     this.reloadTimer = null;
+                    currentTimer.Dispose();
                 }
 
                 if (this.IsDisposing)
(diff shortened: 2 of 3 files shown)


304NotModified:
Are you sure this is correct?

snakefoot:
Trying to follow this guide, but still trying to resolve the triangle between file-watcher, timer-queue and domain-unload-dispose.

https://blogs.msdn.microsoft.com/tmarq/2007/07/21/an-ounce-of-prevention-using-system-threading-timer-in-an-asp-net-application/

snakefoot:
Now changed dispose order:
- Unhooks from file-watcher-event, so not starting new reload timers.
- If able to dispose the active timer, then it knows the timer will perform early exit.
- Else it checks if the timer is currently active, or never started.

304NotModified:
@plukawski I you have time, please your opinion :) thx!

snakefoot:
Another article, that describes how ProcessExit and DomainUnload-events are dangerous, as too much time spent will cause finalizer-thread to be killed and leaving buffers unflushed:

http://joeduffyblog.com/2006/10/17/clr-locks-and-shutdown/

304NotModified:
I haven't time to read it, so this is still WIP?

Did a merge of master to coreCLR. Took me an hour fixing merge conflicts (and not done yet)

snakefoot:
No longer work WIP. Think I have come to the conclusion that it cannot be easily fixed, since NLog depends on doing a lot of action on these events.

But the commit includes a race-condition fix between reload-timer-thread and thread-calling-dispose.

304NotModified:
So it's an improvement, but won't fix #1846  right?

snakefoot:
Yes an improvement, but won't fix #1846 (Replaced a Wait with a lock, same-same).

Not sure how to implement a secure shutdown where everything is flushed. Especially in the IIS world where AppDomains are recycled and Application-Pools are shutdown.

Guess one can only be optimistic like NLog is now, and just hope everything is completed within the hard timeout.

Maybe consider replacing the ProcessExit and DomainUnload events with a LogFactory-Finalizer. Just to remove the use of these special events. Then introduce a special non-async-fast-flush-method that doesn't rely on async-wrappers or buffer-wrappers special timer-logic.

snakefoot:
Race-condition can be seen here:

https://travis-ci.org/NLog/NLog/builds/181278187

304NotModified:
looks good!

FYI, I dunno if the `Monitor.TryEnter` is better than the `ManualResetEvent`. Have to read in first

snakefoot:
The Monitor.TryEnter is better, as it actually waits for the Timer-thread. The ManualResetEvent could be skipped if the reloadTimer had disposed its own Timer, just before the dispose-thread checks it. Dispose-thread would then continue, eventhough the realodTimer was still active.

snakefoot:
Have now added a new method LogFactory.Close(TimeSpan flushTimeout).

Now it is possible to close down the LogFactory without waiting 15 secs. I have chosen 1.5 sec to be the sweetspot. If not able to complete flush within timeout, then it just continues without closing the targets.

304NotModified:
> There is a breaking change for SILVERLIGHT that apparently never flushed its targets on "shutdown"

Sounds also as a bug to me, so no breaker :)

304NotModified:
> The Monitor.TryEnter is better, as it actually waits for the Timer-thread. The ManualResetEvent could be skipped if the reloadTimer had disposed its own Timer, just before the dispose-thread checks it. Dispose-thread would then continue, eventhough the realodTimer was still active.

Thanks for the explanation!

PR #1854, opened 2016-12-14
Fix immutability of LogLevel properties

ie-zero:
It fixes the immutability of the static properties AllLevels and AllLoggingLevels for LogLevel class.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1854)

### DIFF
--- a/src/NLog/LogLevel.cs
+++ b/src/NLog/LogLevel.cs
@@ -85,22 +85,39 @@ namespace NLog
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Security", "CA2104:DoNotDeclareReadOnlyMutableReferenceTypes", Justification = "Type is immutable")]
         public static readonly LogLevel Off = new LogLevel("Off", 6);
 
-
-        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Security", "CA2104:DoNotDeclareReadOnlyMutableReferenceTypes", Justification = "Type is immutable")]
-        private static readonly IList<LogLevel> allLevels = new List<LogLevel> { Trace, Debug, Info, Warn, Error, Fatal, Off };
-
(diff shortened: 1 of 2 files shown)


snakefoot:
Why use the yield-state-machine for a get-property, when you can just use IList<T>.AsReadOnly() ?

ie-zero:
In the previous implementation the following unit tests should be failing but instead they were passing 

``` C#
[Fact]
public void LogLevel_SetAllLevels()
{
	((ICollection<LogLevel>) LogLevel.AllLevels).Add(LogLevel.Fatal);
	Assert.Equal(
		new List<LogLevel>() { 
			LogLevel.Trace, LogLevel.Debug, LogLevel.Info, 
			LogLevel.Warn, LogLevel.Error, LogLevel.Fatal, LogLevel.Off 
		},
		LogLevel.AllLevels);
}
```

``` C#
[Fact]
public void LogLevel_SetAllLoggingLevels()
{
	((ICollection<LogLevel>)LogLevel.AllLoggingLevels).Add(LogLevel.Fatal);
	Assert.Equal(
		new List<LogLevel>() { 
			LogLevel.Trace, LogLevel.Debug, LogLevel.Info, 
			LogLevel.Warn, LogLevel.Error, LogLevel.Fatal 
		},
		LogLevel.AllLoggingLevels);
}
```

snakefoot:
Well old code didn't call AsReadOnly() on the static-allocated List.

ie-zero:
I have no strong preference. If the preference is IList.AsReadOnly()  I can happily change to that. I simply noticed that the immutability was compromised.

snakefoot:
Then it is only done once, instead of getting a new ienumerable allocated every time.

304NotModified:
I prefer AsReadOnly :)

ie-zero:
To be perfectly honest, I cannot see how those two static properties fit on this object as they refer to a 'collection' rather than an individual object. We can  consider separating the ```LogLevel``` from the ```LogLevelCollection``` in the next major version.

ie-zero:
Fair enough. I will change the code to AsReadOnly than :)

ie-zero:
I will resubmit this PR as you proposed.

Issue #1856, opened 2016-12-15
File target - keepFileOpen: file is not recreated if deleted manually.


thba:
**Type**: Bug
**NLog version**: 4.3.4
**Platform**: .Net 4.5

I have configured a file target:
`<target name="myFileTarget" xsi:type="File"
	fileName="${event-properties:item=myFileDir}\${event-properties:item=myFile}"
	layout="${message}"
	cleanupFileName="false"
	autoFlush="false"
	concurrentWrites="false"
	keepFileOpen="true">
</target>`

When I run my application the file gets created and logged into like expected. When my application is idle (no logs are created) and I delete the logfile and then the application starts to produce logs again, the file is not re-created, the logs go to nowhere and no exception or error occures.

I expect that the file is recreated when it does no longer exist. If I set keepFileOpen = "false" this is the case.

304NotModified:
Not sure if this a bug.

I think keepopen and deleting the file is incompatible

snakefoot:
Would be an expensive operation to re-validate FileStream for every wirte-operation.

Windows allows the user to delete files, while another process has open filehandles to the file. Windows just cheats and hides the file from the user, and deletes the file correctly when all handles are closed.

thba:
Well, I know nothing about NLog's "special technique" to keep the files open, but I wonder what happens internally if the file suddenly no longer exists. I can't even see some errors in ProcMon.

It has not to be, that NLog verifies the file stream with every write-operation, but it would be possible to implement a method to do so manually.

I have a long running application which has an idle mode, where no logs are written, and an active mode where logs are written. The user may delete the logs during an idle period. If he does, he will not have any logs for the next active period. If I had some method to call before my app enters the active mode, everything would be fine. Something to reset the logger completly, but I have not found anything which would help me so far.

snakefoot:
Maybe you could try and use this special tool to close file-handles.

So the trick is:

1) Rename the file in the log-directory to have bak-file-extension
2) Close all file handle to the file in the log-directory (See https://technet.microsoft.com/en-us/sysinternals/bb896655)
3) Delete all file with bak-file-extension in the log-directory

This should make NLog react to the filehandle being broken, and then reopen the file automatically.

304NotModified:
@snakefoot would closing/re-initialing of the target help?

snakefoot:
[...]

PR #1858, opened 2016-12-16
FileTarget - Reopen filehandle when file write fails

snakefoot:
Making FileTarget more resilient to disconnecting network drives, or forceful close of file handles. See also #1856

It is certainly not a perfect solution, as the FileStream will continue to work until the buffer is full, causing the unwritten buffer to be discarded on failure.

Also there is no throttle on error, but I guess that is for another PR. To have a wrapper that take a pauses every second, while hammering (Before it would not hit the disk/network again, but just keep failing internally. But I guess one should always use the RetryingMultiProcessFileAppender for network drives).

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1858)

### DIFF
--- a/src/NLog/Internal/FileAppenders/CountingSingleProcessFileAppender.cs
+++ b/src/NLog/Internal/FileAppenders/CountingSingleProcessFileAppender.cs
@@ -36,6 +36,7 @@ namespace NLog.Internal.FileAppenders
     using System;
     using System.IO;
     using System.Security;
+    using NLog.Common;
 
     /// <summary>
     /// Implementation of <see cref="BaseFileAppender"/> which caches 
@@ -83,8 +84,20 @@ namespace NLog.Internal.FileAppenders
         {
             if (this.file != null)
             {
-                this.file.Close();
-                this.file = null;
+                InternalLogger.Trace("Closing '{0}'", FileName);
+
+                try
+                {
+                    this.file.Close();
+                }
+                catch (Exception ex)
+                {
+                    InternalLogger.Warn(ex, "Failed to close file: '{0}'", FileName);
+                }
+                finally
+                {
+                    this.file = null;
+                }
             }
         }
 
@@ -98,7 +111,14 @@ namespace NLog.Internal.FileAppenders
                 return;
             }
 
-            this.file.Flush();
+            try
+            {
+                this.file.Flush();
+            }
+            catch (Exception ex)
+            {
+                throw new System.IO.IOException("FileStream Flush Failed: " + FileName, ex);
+            }
             FileTouched();
         }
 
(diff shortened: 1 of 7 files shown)


304NotModified:
Is it OK to throw a IOException in user code?

snakefoot:
> Is it OK to throw a IOException in user code?

I have no opinion about this. Think the exceptions not to be thrown by usercode is Stackoverflow- and Nullreference-Exception. I can easily change it.

src/NLog/Internal/FileAppenders/CountingSingleProcessFileAppender.cs
```diff
@@ -98,7 +111,14 @@ public override void Flush()
...
             }
 
-            this.file.Flush();
+            try
+            {
+                this.file.Flush();
+            }
+            catch (Exception ex)
+            {
+                throw new System.IO.IOException("FileStream Flush Failed: " + FileName, ex);
```
304NotModified: I'm not sure if this is how I would implement it.

Isn't better to just rethrow the ex, and maybe add some context to the exception? e.g. 

```c#
catch (Exception ex)
{
    ex.data["info"] = "FileStream Flush Failed: " + FileName;
    throw;
} 
```

src/NLog/Internal/FileAppenders/CountingSingleProcessFileAppender.cs
```diff
@@ -83,8 +84,20 @@ public override void Close()
...
-                this.file = null;
+                InternalLogger.Trace("Closing '{0}'", FileName);
+
+                try
+                {
+                    this.file.Close();
+                }
+                catch (Exception ex)
+                {
+                    InternalLogger.Warn(ex, "Failed to close file: '{0}'", FileName);
```
304NotModified: Think we should mention we are ignoring the exception

snakefoot:
I have now removed the wrapping of the structured-exception being thrown, when file-handle has been broken.

src/NLog/Internal/FileAppenders/CountingSingleProcessFileAppender.cs
snakefoot: I prefer to use the InnerException-pattern instead of modifying the original exception.

src/NLog/Internal/FileAppenders/CountingSingleProcessFileAppender.cs
snakefoot: Comment added

304NotModified:
Is this one ready for review/merge? :)

snakefoot:
Yep :)

304NotModified:
OK then! Merged!

PR #1859, opened 2016-12-17
FileTarget - MONO should also check SupportsSharableMutex

snakefoot:
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1859)

### DIFF
--- a/src/NLog/Internal/FileAppenders/BaseMutexFileAppender.cs
+++ b/src/NLog/Internal/FileAppenders/BaseMutexFileAppender.cs
@@ -72,15 +72,13 @@ namespace NLog.Internal.FileAppenders
 #endif
         }
 
-#if !SILVERLIGHT
+#if SupportsMutex
         /// <summary>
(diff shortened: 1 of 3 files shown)


snakefoot:
Interesting. Seems there is a bug in MutexMultiProcessFileAppender when running on MONO. Or a bug in the tester?

Can see that we favor the UnixMultiProcessFileAppender on MONO, with fallback to MutexMultiProcessFileAppender on Win32.

Guess the PR should be changed from checking ForceMutexConcurrentWrites to SupportsSharableMutex.

304NotModified:
looks good, so in Mono4 we now support MutexMultiProcessFileAppender

snakefoot:
> looks good, so in Mono4 we now support MutexMultiProcessFileAppender

More like in MONO4 on Windows then we support MutexMultiProcessFileAppender

304NotModified:
Thanks, looks good!

PR #1860, opened 2016-12-17
FileAppenderCache implements IDisposable

ie-zero:
IDisposable is implemented by FileAppenderCache class according to Microsoft recommendations "CA1001: Types that own disposable fields should be disposable".

Non-breaking change as the class is internal.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1860)


304NotModified:
👍 thx!

PR #1861, opened 2016-12-17
Helper class ParameterUtils removed

ie-zero:
Helper class ParameterUtils is underused in the project so it is removed.

This is a non-breaking change as the class is internal.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1861)


304NotModified:
Thanks! Also better for the stacktrace :)

Issue #1866, opened 2016-12-18
FileTarget Archiving gets stuck


ivmreg:
**Type**:

- Bug

**NLog version**: 4.4.0

**Platform**: .Net 4.5

**Current NLog config** (xml or C#, if relevant)
```

	        var fileTarget = new FileTarget
	        {
	            FileName = Path.Combine(directoryPath, "${logger}_${date:format=yyyyMMdd}.txt"),
	            Layout = DefaultLayoutFormatter,
                NetworkWrites = false,
                KeepFileOpen = true,
                ConcurrentWrites = true,
	            Name = name,
                MaxArchiveFiles = 10,
                ArchiveAboveSize = 1024 * 20, //1024*1024*10 - we reduce the log size before archiving to 20KB to reach the problematic condition quicker, with log size of 10MB the problem reproduces as well
                ArchiveFileName = Path.Combine(directoryPath, "${logger}_{#}.txt"),
                ArchiveDateFormat = "yyyyMMdd.HHmmss",
                ArchiveNumbering = ArchiveNumberingMode.DateAndSequence
            };
            var fileWrapper = new BufferingTargetWrapper
            {
                WrappedTarget = fileTarget,
                Name = name,
                SlidingTimeout = true
            };

            LogManager.Configuration.AddTarget(name, fileWrapper);

            var logFileLoggingRule = new LoggingRule("*", minLevel, fileTarget);
	        LogManager.Configuration.LoggingRules.Add(logFileLoggingRule);

```

Bug description:
We have a "host" application that spins up multiple child "runner" applications.
The "host" application can write to it's own single "host" log.
Multiple "Runner" application can write to a single "runners" log as well as write some diagnostic information in "host" log.
Therefore, for both "runner" and "host" logs there are multiple writers.
The configuration above reflects that.

- What is the current result?

At some point during the runtime of the application when archiving of "host" log happens, it gets stuck on "host" application side (at least) with the following message:

```

2016-12-18 21:15:36.4364 Info Deleting old archive D:\Temp\Logs\HOST_20161218.211223.0.txt.
2016-12-18 21:15:36.4364 Warn Failed to archive file 'D:\Temp\Logs\HOST_20161218.txt'. Exception: System.UnauthorizedAccessException: Access to the path 'D:\Temp\Logs\HOST_20161218.211223.0.txt' is denied.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
[...]

PR #1867, opened 2016-12-18
FileTarget - Failing to delete old archive files, should not stop logging

snakefoot:
See also #1866 

Have also changed the compression logic code from waiting forever to "only" 12 secs.

This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1867)


304NotModified:
Thanks!

PR #1853, opened 2016-12-14
AsyncTargetWrapper - Flush should start immediately without waiting

snakefoot:
AsyncTargetWrapper has the following changes:

- FlushAsync now schedules an "instant" QueueUserWorkItem, instead of waiting for the time-thread-event. Instead of waiting for all LogEvents to be written, then explicit Flush is called on WrappedTarget right after having written all pending logevents.  (minor breaking change)
   - This means that putting a BufferingTargetWrapper inside "targets async=true" will no longer cause the AsyncTargetWrapper to wait for the BufferingTargetWrapper-flush-delay.
    - This also means that Targets with broken Flush-method (not keeping track of pending logevents) will not get help from the AsyncTargetWrapper during flush.
- CloseTarget now flushes all pending messages (instead of just the first BatchSize). CloseTarget-caller was also blocking before, so not much change here.

BufferingTargetWrapper has the following changes:

- FlushAsync now lets the WrappedTarget decide when it has flushed. Instead of waiting for all LogEvents to be written and then call explicit Flush on WrappedTarget. Flush-caller was also blocked before, so not much change here.
    - This also means that Targets with broken Flush-method (not keeping track of pending logevents) will not get help from the BufferingTargetWrapper during flush.
- CloseTarget now flushes any pending messages (minor breaking change, if not bugfix)
- Do not hold Target.SyncRoot while writing messages on Timer-thread (Performance)

Reasons for these changes:
- Closer alignment of AsyncTargetWrapper and BufferingTargetWrapper
- Refactoring to increase the chance of making an instant flush on shutdown, without blocking the finalizer thread too much so it is terminated without anything written (See also #1846)
- Could consider to change AsyncTargetWrapper.FlushAsync to perform synchrone flush instead of ThreadPool.QueueUserWorkItem, but then it would be a major breaking change, as the Flush-caller would be blocked while executing all the writes. (Would avoid spinning up new threads during shutdown, though this is already done by LogFactory.Flush)

Related issues in this area: #134 and #214

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1853)

PR #1868, opened 2016-12-19
WebServiceTarget - FlushAsync - Avoid premature flush

### LINKED ISSUE #259: Lost log statements target on application (AppDomain) close (when async writing)
Since the WebService target is asynchronous messages are lost if the application close before log statements are sent.

Here is a basic example of our program and config:

Program.cs

``` C#
using NLog;

namespace ConsoleTest
{
    internal class Program
    {
        private static void Main()
        {
            var logger = LogManager.GetLogger("MyLog");
            logger.Info("Hello, world!");
            logger.Info("Hello, world again!");
        }
    }
}
```

nlog.config

``` XML
<nlog>
  <targets>
      <target name="ws" xsi:type="WebService" namespace="http://tempuri.org" methodName="SaveLog" url="http://ourserver.com/">
        <parameter name="message" type="String" layout="${message}"/>
      </target>
  </targets>

  <rules>
    <logger name="MyLog" minlevel="Info" writeTo="ws" />
  </rules>
</nlog>
```

We have tried some different approaches so far, such as wrapping the target _ws_ in `AutoFlushWrapper` (writeTo that wrapper) and even calling `LogManager.Flush()` but this did not help.  
If we use `AsyncWrapper` we seem to be able to flush out one of the messages, but not all.

snakefoot:
Keeps track of pendingRequests. Tries to fix #259. Instead of expecting async-target to almost handle it #1853

First idea was to keep track of each individual LogEvent (in a dictionary), but the logic (and the allocations) became a little heavy. Now it just keeps track of a pending request count, so it is not completely correct when races occurs, but I think it is good enough.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1868)

### DIFF
--- a/src/NLog/Targets/WebServiceTarget.cs
+++ b/src/NLog/Targets/WebServiceTarget.cs
@@ -180,7 +180,7 @@ namespace NLog.Targets
         /// <docgen category='Web Service Options' order='10' />
         public string XmlRootNamespace { get; set; }
 
-        private readonly Dictionary<HttpWebRequest, KeyValuePair<DateTime,AsyncContinuation>> _pendingRequests = new Dictionary<HttpWebRequest, KeyValuePair<DateTime,AsyncContinuation>>();
+        private readonly AsyncOperationCounter pendingManualFlushList = new AsyncOperationCounter();
 
         /// <summary>
         /// Calls the target method. Must be implemented in concrete classes.
@@ -227,15 +227,7 @@ namespace NLog.Targets
                 {
                     if (ex != null)
                     {
-                        KeyValuePair<DateTime, AsyncContinuation> flushCallback;
-                        lock (this.SyncRoot)
-                        {
-                            if (_pendingRequests.TryGetValue(request, out flushCallback))
-                                _pendingRequests.Remove(request);
-                        }
-                        continuation(ex);
-                        if (flushCallback.Value != null)
-                            flushCallback.Value(ex);
+                        DoInvokeCompleted(continuation, ex);
                         return;
                     }
 
@@ -250,15 +242,7 @@ namespace NLog.Targets
                                     {
                                     }
 
-                                    KeyValuePair<DateTime, AsyncContinuation> flushCallback;
-                                    lock (this.SyncRoot)
-                                    {
-                                        if (_pendingRequests.TryGetValue(request, out flushCallback))
-                                            _pendingRequests.Remove(request);
-                                    }
-                                    continuation(null);
-                                    if (flushCallback.Value != null)
-                                        flushCallback.Value(null);
+                                    DoInvokeCompleted(continuation, null);
                                 }
                                 catch (Exception ex2)
                                 {
@@ -268,15 +252,7 @@ namespace NLog.Targets
                                         throw;
                                     }
 
-                                    KeyValuePair<DateTime, AsyncContinuation> flushCallback;
-                                    lock (this.SyncRoot)
-                                    {
-                                        if (_pendingRequests.TryGetValue(request, out flushCallback))
-                                            _pendingRequests.Remove(request);
-                                    }
-                                    continuation(ex2);
-                                    if (flushCallback.Value != null)
-                                        flushCallback.Value(ex2);
+                                    DoInvokeCompleted(continuation, ex2);
                                 }
                             },
                             null);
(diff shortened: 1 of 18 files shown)


tests/NLog.UnitTests/Targets/WebServiceTargetTests.cs
```diff
@@ -454,6 +454,15 @@ public void WebserviceTest_restapi_httppost_checkingLost()
...
 
+                // Make triple-flush to fully exercise the async flushing logic
+                try
+                {
+                    LogManager.Flush(0);
+                }
+                catch (NLog.NLogRuntimeException)
+                { }
+                LogManager.Flush(); // Waits for flush (Scheduled on top of the previous flush)
+                LogManager.Flush(); // Nothing to flush
```
304NotModified: should users also do a triple-flush?

src/NLog/Targets/WebServiceTarget.cs
```diff
@@ -180,6 +180,7 @@ public WebServiceTarget(string name) : this()
         /// <docgen category='Web Service Options' order='10' />
         public string XmlRootNamespace { get; set; }
 
+        private readonly Dictionary<HttpWebRequest, KeyValuePair<DateTime,AsyncContinuation>> _pendingRequests = new Dictionary<HttpWebRequest, KeyValuePair<DateTime,AsyncContinuation>>();
```
304NotModified: isn't `HttpWebRequest` an expensive key value?

304NotModified: Another way would be using a linkedlist and send the info with the `AsyncContinuation` (if possible)?

tests/NLog.UnitTests/Targets/WebServiceTargetTests.cs
snakefoot: No triple flush is only for code-coverage :)

src/NLog/Targets/WebServiceTarget.cs
```diff
@@ -226,74 +227,230 @@ protected override void DoInvoke(object[] parameters, AsyncContinuation continua
...
+                    };
+
+                DateTime utcNowTimeout = DateTime.UtcNow.AddSeconds(-120);    // Default HttpWebRequest.Timeout = 100 secs
+
+                bool foundOldWebRequest = false;
+
+                foreach (var key in new List<HttpWebRequest>(_pendingRequests.Keys))
+                {
+                    KeyValuePair<DateTime, AsyncContinuation> flushCallback;
+                    if (_pendingRequests.TryGetValue(key, out flushCallback))
```
snakefoot: FlushAsync is already protected by this.SyncRoot. Same lock-object is used for the other location.

304NotModified:
> I was hoping it was acting like a normal class, and just used default class-reference-hashcode.



yes 're right! 


 >  // The default implementation returns the sync block index for this instance.
    // Calling it on the same object multiple times will return the same value, so
    // it will technically meet the needs of a hash function, but it's less than ideal.
    // Objects (& especially value classes) should override this method.

https://referencesource.microsoft.com/#mscorlib/system/object.cs,4de9cf234d0d8b16

tests/NLog.UnitTests/Targets/WebServiceTargetTests.cs
304NotModified: sounds we should have split those tests.

src/NLog/Targets/WebServiceTarget.cs
304NotModified: Missed that, thx!

snakefoot:
Have now also added AutoFlushTargetWrapper into the mix, since #259 also complains that using the AutoFlushTargetWrapper didn't solve their problem.

Now I also have the crutch (AutoFlushTargetWrapper) that people can use when having a target without a proper flush-method.

tests/NLog.UnitTests/Targets/WebServiceTargetTests.cs
snakefoot: Done

304NotModified:
Reviewed 15 of 17 files at r2, 3 of 3 files at r3.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[tests/NLog.UnitTests/Targets/WebServiceTargetTests.cs, line 626 at r3](https://reviewable.io:443/reviews/nlog/nlog/1868#-KaEl1TuhuaxsuL6eXyS:-KaEl1TuhuaxsuL6eXyT:b-st8lp5) ([raw file](https://github.com/nlog/nlog/blob/89a85193b1a162bb4437493085f040e445a71222/tests/NLog.UnitTests/Targets/WebServiceTargetTests.cs#L626)):*
> ```Smalltalk
>                     CountdownEvent = new CountdownEvent(expectedMessages);
>                 else
>                     CountdownEvent = null;
> ```

nice one :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1868)*

snakefoot:
@304NotModified  Think I have broken something, as the test-builds are now failing quite frequently.

PR #1799, opened 2016-11-29
FileTarget: performance improvement: 10-70% faster, less garbage collecting (3-4 times less) by reusing buffers

snakefoot:
FileTarget and AsyncTargetWrapper with optimized memory allocations for layout rendering, encoding and streaming. Optimization is disabled by default for all targets with the setting OptimizeBufferReuse (Default = false). It has been explicit enabled for pure NLog-FileTarget and pure NLog-AsyncTargetWrapper.

Introduced a new property AsyncTargetWrapper.FullBatchSizeWriteLimit (Default = 5). It gives better performance to have many small writes. But have increased AsyncTargetWrapper.BatchSize to 200. 

Refactoring of #1663

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1799)

PR #1871, opened 2016-12-21
Filetarget allocations optimization (up to 15% speed improvement)

nazim9214:
1. Removed new string allocation in GetBytesToWrite method. Now we are get count of bytes for text and newline characters and write directly into byte array.

2. Allocate MemoryStream directly with required buffer size. 

Performance results for this test https://github.com/NLog/NLog-Performance-tests/blob/master/NLogPerformance
with following parameters:

| Messages 	| Size 	| Threads 	| Loggers 	|
|----------	|------	|---------	|---------	|
| 1000000  	| 100  	| 2       	| 1       	|

| version   	| Time (ms) 	| Msgs/sec 	| GC2 	| GC1 	| GC0 	| CPU (ms) 	| Mem (MB) 	|
|-----------	|-----------	|----------	|-----	|-----	|-----	|----------	|----------	|
| master    	| 14,885    	| 67,178   	| 328 	| 409 	| 848 	| 28,671   	| 51.051   	|
| optimized 	| 12,914    	| 77,432   	| 179 	| 324 	| 704 	| 25,000   	| 47,344   	|

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1871)

### DIFF
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -1014,39 +1014,57 @@ namespace NLog.Targets
         protected override void Write(AsyncLogEventInfo[] logEvents)
         {
             var buckets = logEvents.BucketSort(c => this.GetFullFileName(c.LogEvent));
-            var pendingContinuations = new List<AsyncContinuation>();
-
-            foreach (var bucket in buckets)
+            using (var ms = new MemoryStream())
             {
-                string fileName = bucket.Key;
-                LogEventInfo firstLogEvent = null;
+                var pendingContinuations = new List<AsyncContinuation>();
 
-                byte[][] logEventsBytes = new byte[bucket.Value.Count][];
-                int bucketByteCount = 0;
-                for (int i = 0; i < bucket.Value.Count; i++)
+                foreach (var bucket in buckets)
                 {
-                    AsyncLogEventInfo ev = bucket.Value[i];
-                    if (firstLogEvent == null)
+                    string fileName = bucket.Key;
+
+                    ms.SetLength(0);
+                    ms.Position = 0;
+
+                    LogEventInfo firstLogEvent = null;
+
+                    for (int i = 0; i < bucket.Value.Count; i++)
                     {
-                        firstLogEvent = ev.LogEvent;
+                        AsyncLogEventInfo ev = bucket.Value[i];
+                        if (firstLogEvent == null)
+                        {
+                            firstLogEvent = ev.LogEvent;
+                        }
+
+                        byte[] bytes = this.GetBytesToWrite(ev.LogEvent);
+
+                        if (ms.Capacity == 0)
+                        {
+                            ms.Capacity = GetMemoryStreamInitialSize(bucket.Value.Count, bytes.Length);
+                        }
+
+                        ms.Write(bytes, 0, bytes.Length);
+                        pendingContinuations.Add(ev.Continuation);
                     }
 
-                    var bytes = this.GetBytesToWrite(ev.LogEvent);
-                    logEventsBytes[i] = bytes;
-                    bucketByteCount += bytes.Length;
-                    pendingContinuations.Add(ev.Continuation);
-                }
-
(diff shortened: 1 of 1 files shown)


304NotModified:
@snakefoot as your are active on the FileTarget, your opinion please :)
Is this one also clashing with another PR?

@nazim9214 thanks!

src/NLog/Targets/FileTarget.cs
```diff
@@ -1096,8 +1100,13 @@ protected virtual string GetFormattedMessage(LogEventInfo logEvent)
...
         protected virtual byte[] GetBytesToWrite(LogEventInfo logEvent)
         {
-            string renderedText = this.GetFormattedMessage(logEvent) + this.NewLineChars;
-            return this.TransformBytes(this.Encoding.GetBytes(renderedText));
+            string text = this.GetFormattedMessage(logEvent);
+            int textBytesCount = this.Encoding.GetByteCount(text);
+            int newLineBytesCount = this.Encoding.GetByteCount(this.NewLineChars);
+            byte[] bytes = new byte[textBytesCount + newLineBytesCount];
+            this.Encoding.GetBytes(text, 0, text.Length, bytes, 0);
+            this.Encoding.GetBytes(this.NewLineChars, 0, this.NewLineChars.Length, bytes, textBytesCount);
```
304NotModified: @nazim9214 

As the NewLineChars & Encoding almost never changes while running, I was thinking to "cache" this one (`static` variable). I noticed before while performance testing that setting a fixed bytes for the newline improves the performance ~10-20%.  (but with a limited test)

Your thoughts please :)

snakefoot:
It is a good optimization of the old code path without breaking interfaces.

Will of course conflict with my PR #1799 that introduces a new code-path that allows reuse of MemoryStream + byte[]-array + StringBuilder.

src/NLog/Targets/FileTarget.cs
nazim9214: @304NotModified 

I wrote a version where 'line ending bytes' are computed only on setting Encoding/LineEnding property of FileTarget (as i understand user can change this properties at runtime in some way?). 
My tests on small and medium message sizes show same throughput as in this implementation. 
So i think we shouldn't add more complexity to codebase, because there isn't any significant benefits.

snakefoot:
Instead of making an array of byte-array to make an exact MemoryAllocation could it not just be:
```
byte[] bytes = this.GetBytesToWrite(ev.LogEvent);
if (ms.Capacity == 0)
{
   if (bucketCount < 5)
      ms.Capacity = ((1 + bucketCount) * bytes.Length);
   else
      ms.Capacity = ((((bucketCount + 1) * bytes.Length) / 1024) * 1024 + 1024);
}
ms.Write(bytes, 0, bytes.Length);
```

Of course this very crude calculation will be cheated when having very uneven string-lengths. And the first string is the shortest one.

nazim9214:
Yes, we could use this approach.
I ran some tests:
1. For constant message size(100 chars, same test as in the PR) this approach better than PR implementation: ~77k msg/sec (exact calculation) vs ~80k msg/sec (approx calculation).
2. For random message size(50 - 100 chars) another situation: ~83k (exact calculation) vs ~81k (approx calculation)

I'm not sure which of the approaches select, your thoughts please @304NotModified @snakefoot

304NotModified:
I think random is more important, but also not sure if <5% improvement is worth the change

snakefoot:
> I think random is more important, but also not sure if <5% improvement is worth the change

It will just be easier for me to merge with #1799, if not needing to keep track of an array of byte-arrays.

I guess in NLog 5.0 then the old code-path will be removed completely, and the MemoryStream will just be reused instead of being allocated and thrown away constantly.

nazim9214:
Ok, then i'll rewrite this piece of code, as you proposed.

304NotModified:
👍 thanks in advance!

304NotModified:
Thanks again!

PR #1880, opened 2016-12-29
FileTarget - Rename to tmp file before starting actual compression

### LINKED ISSUE #1314: File target archive compression goes out of control if log file is held open by other process
**Type** (choose one): Bug

**NLog version**: 4.1.0 and 4.3.0-beta3

**Platform**: .Net 4.5, Windows 10

``` xml
  <targets>
    <default-wrapper xsi:type="AsyncWrapper"/>
    ...
    <target name="file" xsi:type="File" layout="${commonLayoutIndented}" 
            fileName="${configFolder}/MediaService.log" 
            createDirs="true" concurrentWrites="false" keepFileOpen="true"
            archiveNumbering="DateAndSequence" archiveDateFormat="yyyy.MM.dd"
            archiveAboveSize="52428800" maxArchiveFiles="10"
            archiveFileName="${configFolder}/MediaService.log.{#}.zip"
            enableArchiveFileCompression="true"
            encoding="UTF-8"/>
  </targets>
```

(${configFolder} is under C:/ProgramData/)

**Description**: We were experimenting with enableArchiveFileCompression. The above file target works wonderfully if the log files are left alone. We were also testing using tail from the Git for Windows installation (portable version also works, both available at https://git-scm.com/download/win) to watch our log file. When the rollover limit was hit, the log file stopped growing, and several empty archive files started rapidly being created and rolling over, very similar to what is described in #1028 (but we have concurrentWrites="false"). When we stopped the tail process, it seems to recover.

We tried updating to 4.3.0-beta3 since #1028 was fixed for that release, and with that version, it instead just stops logging, creates a single archive, and then nothing happens (no further log messages anywhere), until we stop tail, then it will continue as if nothing happened (spamming all the buffered log events so far to disk).

Tail is probably not the only program that triggers this, but I won't speculate on how tail is opening the file (or what NLog actually does when it rotates/archives the file). It's just a concrete program that can reproduce this.

Nothing special from the internal nlog log, for 4.1.0 we get nothing at all, and from 4.3.0-beta3 we get a single line: 2016-03-18 16:49:42.8018 Info Archiving ...\MediaService.log to zip-archive ...\MediaService.log.2016.03.18.0.zip

Possible workarounds: Don't use stupid tail from Git for Windows :P

On a more serious note, if we remove enableArchiveFileCompression, there are no problems for either version, and archiving and rotation works as expected (tail won't see the new log, but that's its problem)

snakefoot:
The rename will allow other processes to continue without waiting for compression to complete when ConcurrentWrites=true. Will also fix #1314

This will of couse affect performance if the achive-folder is placed on a remote network-drive, since it will do a file-copy of the uncompressed file before doing the compression.

Could consider to perform the compression in a dedicated thread, but then we need extra logic to join with the dedicated-compression-thread when closing file-target. And also control the thread-priority, and possible queuing of file-compression-tasks.

Would be nice if the IFileCompressor took 3 parameters. Input-FilePath, Output-FilePath and Archive-FileName. Then the rename logic didn't have to be done within IFileCompressor.CompressFile(). This is probably something for NLog 5.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1880)

### DIFF
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -1299,9 +1299,20 @@ namespace NLog.Targets
 
             if (EnableArchiveFileCompression)
             {
+                // Would be nice if the FileCompressor used a temporary-output-filename until completed,
+                // but that would require that the FileAppenderCache also reacted to renames in archive-folder
                 InternalLogger.Info("Archiving {0} to compressed {1}", fileName, archiveFileName);
+                // Would be nice if the FileCompressor supported a temporary-input-filename (ex. archiveFileName + ".tmp")
+                // but that would require that the FileCompressor supported 3 parameters (input-filepath, output-filepath and archive-filename)
+                DateTime orgCreationTime = File.GetCreationTime(fileName);
                 FileCompressor.CompressFile(fileName, archiveFileName);
-                DeleteAndWaitForFileDelete(fileName);
+                if (File.Exists(fileName) && File.GetCreationTime(fileName) == orgCreationTime)
+                {
+                    // Manual cleanup after the FileCompressor
+                    DeleteOldArchiveFile(fileName);
+                    if (File.Exists(fileName) && File.GetCreationTime(fileName) == orgCreationTime)
+                        System.Threading.Thread.Sleep(100);
+                }
             }
             else
             {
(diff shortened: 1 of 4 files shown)


snakefoot:
After "discovering" the issue copy the uncompressed file to a remote network drive, then I think that I will close this PR.

I guess the proper solution would be:

- Rename current-filename-ext to current-filename-archive-filename-ext (Within the current folder)
- Perform archive operation on the current-filename-archive-filename-ext
- Add support for recognizing old current-filename-archive-filename-ext, and retry archive operation
- Change the archive-mutex so it becomes try-aquire and repeatedly check if the current-filename-ext already has been renamed by another FileTarget.

I will close this issue.

304NotModified:
> After "discovering" the issue with copy the uncompressed file to a remote network drive, then I think that I will close this PR.

good one.



> I guess the proper solution would be:
> 
> Rename current-filename-ext to current-filename-archive-filename-ext (Within the current folder)
> Perform archive operation on the current-filename-archive-filename-ext
> Add support for recognizing old current-filename-archive-filename-ext, and retry archive operation
> Change the archive-mutex so it becomes try-aquire and repeatedly check if the current-filename-ext already has been renamed by another FileTarget.

So renaming for removing the file lock? Or checking if it's locked?

PR #1881, opened 2016-12-30
WindowsPhone is Silverlight with more file access

### LINKED ISSUE #1875: LogManager.GetCurrentClassLogger() failed on WP81 Silverlight apps [fix proposed]
Hi! 

**Type**: Bug

**NLog version**: v4.0.30319

**Platform**: Windows Phone 8.1 Silverlight apps

Today I tried to repeat steps from this tutorial but in terms of Windows Phone 8.1 http://nlog-project.org/2011/01/09/nlog-for-windows-phone-7.html

Latest sources have only Windows Phone 8.0 support (and auto-upgrade causes removal of some methods in "Logger1.cs" file so you can't compile without reverting that file back to depot version). And even after resolving that caveats, you to receive a runtime error while trying to create any logger instance.

Config file is the same as described in the tutorial:

```xml

  <targets>
    <target xsi:type="LogReceiverService" name="webService" endpointAddress="http://localhost:5000/LogReceiver.svc"/>
  </targets>

  <rules>
    <logger name="*" minlevel="Debug" writeTo="webService" />
  </rules>
</nlog>
```

Code usage is from the tutorial too:
private static NLog.Logger logger = NLog.LogManager.GetCurrentClassLogger();

- What is the current result?
Exception: System.MethodAccessException occurred
  HResult=-2146233072
  Message=Attempt by security transparent method 'NLog.LogFactory.get_Configuration()' to access security critical method 'System.Windows.Application.GetResourceStream(System.Uri)' failed.
  Source=NLog
  StackTrace:
  InnerException: 

Normal logger creation

Have no idea how to achieve this as apps crashes at early stage (static init)

Additional information: Attempt by security transparent method 'NLog.LogFactory.get_Configuration()' to access security critical method 'System.Windows.Application.GetResourceStream(System.Uri)' failed.

You can amend the sources in two places:
1) XmlLoggingConfiguration.cs
===============
---  #if SILVERLIGHT
+++ #if SILVERLIGHT && !WINDOWS_PHONE
                newFileName = newFileName.Replace("\\", "/");
                if (Application.GetResourceStream(new Uri(newFileName, UriKind.Relative)) != null)
#else
                if (File.Exists(newFileName))
#endif
===============

 2) LogFactory.cs
===============
---  #if SILVERLIGHT
+++ #if SILVERLIGHT && !WINDOWS_PHONE
                            Uri configFileUri = new Uri(configFile, UriKind.Relative);
                            if (Application.GetResourceStream(configFileUri) != null)
                            {
                                LoadLoggingConfiguration(configFile);
                                break;
                            }
#else
                            if (File.Exists(configFile))
                            {
                                LoadLoggingConfiguration(configFile);
                                break;
                            }
#endif
===============

Not found

Not sure

snakefoot:
Attempt to fix #1875

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1881)

### DIFF
--- a/src/NLog/Common/InternalLogger.cs
+++ b/src/NLog/Common/InternalLogger.cs
@@ -357,14 +357,14 @@ namespace NLog.Common
         {
             try
             {
-#if !SILVERLIGHT && !__IOS__ && !__ANDROID__
+#if SILVERLIGHT || __IOS__ || __ANDROID__
+                Info(assembly.FullName);
+#else
                 var fileVersionInfo = FileVersionInfo.GetVersionInfo(assembly.Location);
                 Info("{0}. File version: {1}. Product version: {2}.",
                     assembly.FullName,
                     fileVersionInfo.FileVersion,
                     fileVersionInfo.ProductVersion);
-#else
-                Info(assembly.FullName);
 #endif
             }
             catch (Exception ex)
--- a/src/NLog/Common/InternalLogger.cs
+++ b/src/NLog/Common/InternalLogger.cs
@@ -357,14 +357,14 @@ namespace NLog.Common
         {
             try
             {
-#if SILVERLIGHT || __IOS__ || __ANDROID__
-                Info(assembly.FullName);
-#else
+#if !SILVERLIGHT && !__IOS__ && !__ANDROID__
                 var fileVersionInfo = FileVersionInfo.GetVersionInfo(assembly.Location);
                 Info("{0}. File version: {1}. Product version: {2}.",
                     assembly.FullName,
                     fileVersionInfo.FileVersion,
                     fileVersionInfo.ProductVersion);
+#else
+                Info(assembly.FullName);
 #endif
             }
             catch (Exception ex)
--- a/src/NLog/Internal/FileAppenders/BaseFileAppender.cs
+++ b/src/NLog/Internal/FileAppenders/BaseFileAppender.cs
@@ -231,7 +231,7 @@ namespace NLog.Internal.FileAppenders
             throw new InvalidOperationException("Should not be reached.");
         }
 
-#if !SILVERLIGHT && !__IOS__ && !__ANDROID__ && !MONO
+#if !SILVERLIGHT && !MONO && !__IOS__ && !__ANDROID__
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Reliability", "CA2000:Dispose objects before losing scope", Justification = "Objects are disposed elsewhere")]
         private FileStream WindowsCreateFile(string fileName, bool allowFileSharedWriting)
         {
@@ -287,7 +287,7 @@ namespace NLog.Internal.FileAppenders
         {
             UpdateCreationTime();
 
-#if !SILVERLIGHT && !__IOS__ && !__ANDROID__ && !MONO
+#if !SILVERLIGHT && !MONO && !__IOS__ && !__ANDROID__
             try
             {
                 if (!this.CreateFileParameters.ForceManaged && PlatformDetector.IsDesktopWin32)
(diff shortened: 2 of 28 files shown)


src/NLog/Common/InternalLogger.cs
```diff
@@ -357,14 +357,14 @@ public static void LogAssemblyVersion(Assembly assembly)
         {
             try
             {
-#if SILVERLIGHT || __IOS__ || __ANDROID__
-                Info(assembly.FullName);
-#else
+#if !SILVERLIGHT && !__IOS__ && !__ANDROID__
```
304NotModified: Why this change? (I see nothing got changed functional). It's this a better style?

It will lead to potential conflicts on the coreCLR branch.

src/NLog/Internal/FileAppenders/BaseFileAppender.cs
```diff
@@ -231,7 +231,7 @@ protected FileStream CreateFileStream(bool allowFileSharedWriting)
             throw new InvalidOperationException("Should not be reached.");
         }
 
-#if !SILVERLIGHT && !MONO && !__IOS__ && !__ANDROID__
+#if !SILVERLIGHT && !__IOS__ && !__ANDROID__ && !MONO
```
304NotModified: idem. potentional conflict

src/NLog/Internal/FileCharacteristicsHelper.cs
```diff
@@ -45,18 +45,16 @@ internal abstract class FileCharacteristicsHelper
         /// </summary>
         public static FileCharacteristicsHelper CreateHelper(bool forcedManaged)
         {
-#if SILVERLIGHT
-            return new PortableFileCharacteristicsHelper();
-#else
+#if !SILVERLIGHT && !__IOS__ && !__ANDROID__
```
304NotModified: good  one, although I prefer not to change the order again ;)

snakefoot:
The change in order was to make everything more consistent, now the non-mobile platform check looks like this everywhere:

> #if !SILVERLIGHT && !__IOS__ && !__ANDROID__

But I can save that for later, if it causes too much trouble for CoreClr-branch.,

src/NLog/Internal/FileCharacteristicsHelper.cs
snakefoot: Well not changing order, but removing code.

304NotModified:
> But I can save that for later, if it causes too much trouble for CoreClr-branch.,

yes please. Thanks!

snakefoot:
> yes please. Thanks!

Done

304NotModified:
Reviewed 17 of 28 files at r1, 3 of 10 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1881)*

304NotModified:
Reviewed 2 of 10 files at r3.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1881)*

304NotModified:
Thanks man! merged!

304NotModified:
looks like this one is giving issues: https://ci.appveyor.com/project/nlog/nlog/build/4.4.4171

snakefoot:
> The master build fails :(

Think if you re-trigger the build, then it succeeds:

https://ci.appveyor.com/project/nlog/nlog/build/4.4.4177

snakefoot:
> The master build fails :(

Was able to reproduce the bug locally, and it seems the following happens:

The test uses ArchiveEvery = FileArchivePeriod.Minute, so if the test is executed at exact 12:59:59.999 then it will trigger an extra unexpected archive operation. As the LogEvent timestamp will then be one-minute older than the creation-time of the current log-file.

The archive operation fails because the log-filename (based on current-time) and the generated archive-filename (based on file-write-time) becomes the same. Because the archive-filename already exists then it attempts to append the contents of the log-filename to the archive-filename and it of course fails.

We could change the ArchiveEvery to Year, and then the test DeleteArchiveFilesByDateWithDateName would only fail at new years eve. 

We could also consider to change FileTarget.ArchiveFile() to check if fileName and archiveFileName is the same, and if so then just do nothing but doing a InternalLogger.Info()

snakefoot:
Created #1886

Issue #1349, opened 2016-04-03
Change MONO pre-processor flags into runtime checks => Mono can use Nuget Packages


damageboy:
**Type** (choose one):

Bug

**NLog version**: (e.g. 4.2.3)
Any nlog version

**Platform**: 
Mono 4 

Hi,
As I'm trying to fix a dead-lock that happens in Mono (only), I've come across different commits in the source code were special behavior was put in place for Mono support, for example:
https://github.com/NLog/NLog/commit/e58867547be4fc1ff9a1c484d723b3ae12d7333e

As can be clearly seen from the actual commit code, fix involved using a `#if MONO` pre-processor condition.

While this was perhaps a correct solution for that time (2011, in the above case), it's obviously doesn't play nice with how people consume NLog in 2016.
Namely, NLog is mostly distributed throughout NuGet, where there is no way do provide different binaries for Standard .NET (MS.NET) and Mono (It's also probably important to make note that this is a by-design decision taken by the NuGet/Mono people).

As such, it looks like it would be more correct to change the Mono checks, specifically, to become runtime checks.

There is a well documented way of detecting Mono @ Runtime:

``` c#
public static bool IsRunningOnMono ()
{
    return Type.GetType ("Mono.Runtime") != null;
}
```

I think that the NLog source, as it has to be compiled once into a single .DLL for MS.NET + Mono should transition into using runtime checks to support conditional execution in Mono...

304NotModified:
This would be great to have. 

We consider this as an feature, as the Mono support was always source distributed and not binary (nuget)

304NotModified:
Fixed by #1893 & #1833

PR #1883, opened 2016-12-31
LogFactory.Dispose - Should always close down created targets

snakefoot:
MONO now also automatically close down created targets (and their timers) on LogFactory-shutdown, just like it currently closes the reload-config-timer (but without spinning up threads for flushing). See also:

http://nlog-project.org/2011/10/30/using-nlog-with-mono.html

It is still a good idea for MONO applications to clear the NLog-configuration just before leaving main(), as it will allow a more speedy (multi-threaded) flush on shutdown (with longer timeout):

> LogManager.Configuration = null

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1883)

### DIFF
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -935,27 +935,32 @@ namespace NLog
                     }
 #endif
 
-                    if (flushTimeout != TimeSpan.Zero)
+                    var oldConfig = this.config;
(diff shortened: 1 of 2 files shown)


304NotModified:
Reviewed it, looks good! Thanks again!

snakefoot:
Getting close to fix the issue #1349

Issue #1707, opened 2016-10-12
Append to archive sometimes fails


304NotModified:
probably and result of  #1680 since NLog 4.3.10

https://ci.appveyor.com/project/nlog/nlog/build/4.0.3358

> Strange that this build error has started to come "randomly": 

(see https://github.com/NLog/NLog/pull/1706#issuecomment-253301042)

> ...NLog.UnitTests.Targets.WrappedFileTargetTests.DeleteArchiveFilesByDateWithDateName [FAIL]
>    System.IO.IOException : The process cannot access the file 'C:\Users\appveyor\AppData\Local\Temp\1\38d53f98-4243-431f-921b-13c219a6989c\20161012182959914.txt' because it is being used by another process.
>    Stack Trace:
>       at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
>       at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
>       at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share)
>       at System.IO.File.Open(String path, FileMode mode)
>       C:\projects\nlog\src\NLog\Targets\FileTarget.cs(1307,0): at NLog.Targets.FileTarget.ArchiveFile(String fileName, String archiveFileName)
>       C:\projects\nlog\src\NLog\Targets\FileTarget.cs(1579,0): at NLog.Targets.FileTarget.ArchiveByDate(String fileName, String pattern, LogEventInfo logEvent)
>       C:\projects\nlog\src\NLog\Targets\FileTarget.cs(1764,0): at NLog.Targets.FileTarget.DoAutoArchive(String fileName, LogEventInfo eventInfo)
>       C:\projects\nlog\src\NLog\Targets\FileTarget.cs(1848,0): at NLog.Targets.FileTarget.TryArchiveFile(String fileName, LogEventInfo ev, Int32 upcomingWriteSize)
>       C:\projects\nlog\src\NLog\Targets\FileTarget.cs(1043,0): at NLog.Targets.FileTarget.ProcessLogEvent(LogEventInfo logEvent, String fileName, Byte[] bytesToWrite)
[...]

PR #1886, opened 2017-01-04
FileTarget - Archive should not fail when ArchiveFileName matches FileName

snakefoot:
Attempt to fix the following build error, and also resolve this issue #1707

https://ci.appveyor.com/project/nlog/nlog/build/4.4.4171

Was able to reproduce the bug locally, and it seems the following happens:

The test uses ArchiveEvery = FileArchivePeriod.Minute and FileName = ${date:format=yyyyMMddHHmmssfff}.txt. It sleeps 50 ms between each LogEvent to ensure that each LogEvent get their own file.

If the test is executed at exact 12:59:59.999 then it will trigger an extra unexpected archive operation. As one of LogEvent timestamps will then be one-minute older than the creation-time of the previous log-file.

The archive operation fails because the log-filename (based on current-time) and the generated archive-filename (based on file-write-time) becomes the same. Because the archive-filename already exists then it attempts to append the contents of the log-filename to the archive-filename and it of course fails.

We could change the ArchiveEvery to Year for the test DeleteArchiveFilesByDateWithDateName, and then it would only fail at new years eve.

We could also consider to change FileTarget.ArchiveFile() to check if fileName and archiveFileName is the same, and if so then just do nothing but doing a InternalLogger.Info()

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1886)

### DIFF
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -1297,7 +1297,11 @@ namespace NLog.Targets
             if (!Directory.Exists(archiveFolderPath))
                 Directory.CreateDirectory(archiveFolderPath);
 
-            if (EnableArchiveFileCompression)
+            if (string.Equals(fileName, archiveFileName, StringComparison.OrdinalIgnoreCase))
+            {
+                InternalLogger.Info("Archiving {0} skipped as ArchiveFileName equals FileName", fileName);
+            }
+            else if (EnableArchiveFileCompression)
             {
                 InternalLogger.Info("Archiving {0} to compressed {1}", fileName, archiveFileName);
                 FileCompressor.CompressFile(fileName, archiveFileName);
--- a/tests/NLog.UnitTests/Targets/FileTargetTests.cs
+++ b/tests/NLog.UnitTests/Targets/FileTargetTests.cs
@@ -877,7 +877,7 @@ namespace NLog.UnitTests.Targets
                 {
                     FileName = logFile,
                     ArchiveFileName = Path.Combine(tempPath, "{#}.txt"),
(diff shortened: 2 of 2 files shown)


304NotModified:
Great work! \o/

snakefoot:
Well the test is not depending on FileArchivePeriod.Minutes, so need to introduce more random behavior than needed.

304NotModified:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/1886)*

304NotModified:
merged :)

owerkop:
So is the warning still valid:
"Warning: when deleting archives files is enabled (e.g. maxArchiveFiles ), the **folder of the archives should different than the log files**. Otherwise, all content, not just logs, in the folder that is not log archives from the target will be deleted."
(https://github.com/nlog/NLog/wiki/File-target )
?

snakefoot:
No this was fixed with NLog 4.5, but for NLog 4.4.13 (or older) then the advice is still valid.


Sent from my Sony Xperia


---- Paweł Owerko wrote ----


So is the warning still valid:
"Warning: when deleting archives files is enabled (e.g. maxArchiveFiles ), the folder of the archives should different than the log files. Otherwise, all content, not just logs, in the folder that is not log archives from the target will be deleted."
(https://github.com/nlog/NLog/wiki/File-target )
?

snakefoot:
@owerkop Have updated the Wiki for FileTarget about warning only being relevant for NLog 4.4 (and older)

304NotModified:
Looks great @snakefoot , thanks! (And valid question @owerkop. Also thanks for that!)

Issue #1122, opened 2015-12-31
LogFactory.ConfirationgChanged is passed old,new parameters in reverse order


Sarafian:
The code is
```
this.OnConfigurationChanged(new LoggingConfigurationChangedEventArgs(value, oldConfig));
```

but the contructor is
```
internal LoggingConfigurationChangedEventArgs(LoggingConfiguration oldConfiguration, LoggingConfiguration newConfiguration)
```

304NotModified:
That's a bit pity. Thanks for reporting, but we cannot change this due to backwards compatibility.

Sarafian:
As a workaround can you please describe the conditions of when the ConfigurationReloaded event fires?

UgurAldanmaz:
`ConfigurationReloaded` event fires when your configuration file changed.

You need to enable auto reload:

```
<nlog autoReload="true">
```

See for more info: https://github.com/NLog/NLog/wiki/Configuration-file

304NotModified:
Has your question being answered @Sarafian ?

Sarafian:
Yes after some further investigation into the code.
Let me explain for future reference. We have some unit tests over some additional to NLOG functionality that configures targets loaded from the config file.

We want to make sure that if the file changes then the configuration is re-applied. Therefore we used to depend on the ConfigurationChanged event but after the upgrade to v4 and the bug mentioned we switched to ConfigurationReloaded. The reason I asked was because after modifying the file the event did not fire. It turns out that it takes some time as expected. The test needs to wait a bit, let nlog do its job and then check if the reloaded targets are re-configured.

Thank you very much.

304NotModified:
Will mark this as know issue

UgurAldanmaz:
Can I close this one due to wontfix?

304NotModified:
Yes indeed

Sarafian:
Forgot to answer. Somehow I had missed the email. Ok from me.

PR #1891, opened 2017-01-07
LogFactory - Fix ConfigurationChanged-event has flipped the values

snakefoot:
Fixes bug introduced with #491 (NLog 4)

Also reported by  #1122. Breaking change for .NET 5 ?

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1891)

### DIFF
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -338,7 +338,7 @@ namespace NLog
                         }
                     }
 
-                    this.OnConfigurationChanged(new LoggingConfigurationChangedEventArgs(value, oldConfig));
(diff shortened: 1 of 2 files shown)


304NotModified:
this is a subtle breaking change. 

The old code compiles but probably doesn't work. 

https://github.com/NLog/NLog/issues/1122 has also a wontfix ;)

maybe obsolete the old event and create a new one?

snakefoot:
Yes don't understand how #1122 was marked as wontfix, since the bug was introduced between NLog ver. 3 and 4. Anyone upgrading from NLog 3 to 4 will be affected by this bug.

PR #1892, opened 2017-01-08
FileTarget - FileLockMultiProcessFileAppender Experimental

snakefoot:
Trying out FileStream.Lock / Unlock for doing multi-process file-write without mutex (MONO non-Linux or NET Core v2 non-Windows). See also https://github.com/dotnet/coreclr/pull/8233

Right now using the log-file for the locking, but could also use an empty file in a NLogLock-subfolder (Ex. also for archive mutex)

Because the FileStream.Lock operation is not blocking, then it becomes a polled interface. So when having many processes actively writing to the same file, then it will be faster with named-mutex if available.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1892)

### DIFF
--- a/src/NLog/Internal/FileAppenders/BaseMutexFileAppender.cs
+++ b/src/NLog/Internal/FileAppenders/BaseMutexFileAppender.cs
@@ -31,24 +31,14 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
-#if !SILVERLIGHT && !__ANDROID__ && !__IOS__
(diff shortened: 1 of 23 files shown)


304NotModified:
Experimental = WIP?

snakefoot:
Seems that NET Core now supports global named mutex (Using FileLocks), so no need to have this logic in NLog. Closing this issue, as Mono will probably apply the NET Core solution. See also https://github.com/dotnet/coreclr/issues/3422

PR #1893, opened 2017-01-09
FileTarget - MONO doesn't like using the native Win32 API

snakefoot:
Together with #1883, then it will also resolve the issue #1349 

This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1893)

### DIFF
--- a/src/NLog/Internal/FileAppenders/BaseFileAppender.cs
+++ b/src/NLog/Internal/FileAppenders/BaseFileAppender.cs
@@ -290,7 +290,7 @@ namespace NLog.Internal.FileAppenders
 #if !SILVERLIGHT && !MONO && !__IOS__ && !__ANDROID__
             try
             {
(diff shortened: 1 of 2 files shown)


304NotModified:
thanks!!

src/NLog/Internal/FileAppenders/BaseFileAppender.cs
```diff
@@ -290,7 +290,7 @@ private FileStream TryCreateFileStream(bool allowFileSharedWriting)
 #if !SILVERLIGHT && !MONO && !__IOS__ && !__ANDROID__
```
304NotModified: should !MONO be removed?

snakefoot: Well it saves some PlatformDetector-checks, when running pure MONO.

304NotModified: true :)

PR #1897, opened 2017-01-10
LogFactory - Fixed EventArgs for ConfigurationChanged

snakefoot:
Fixes the issue #1122 with the breaking change introduced with NLog ver. 4 for LoggingConfigurationChangedEventArgs. See also #491

This marks the broken OldConfiguration and NewConfiguration as obsolete, and instead introduces two new properties:

- ActivatedConfiguration
- DeactivatedConfiguration

No source breaking change, and no binary breaking change. Just obsolete warnings and new properties.

### DIFF
--- a/src/NLog/Config/LoggingConfigurationChangedEventArgs.cs
+++ b/src/NLog/Config/LoggingConfigurationChangedEventArgs.cs
@@ -43,24 +43,38 @@ namespace NLog.Config
         /// <summary>
         /// Initializes a new instance of the <see cref="LoggingConfigurationChangedEventArgs" /> class.
         /// </summary>
-        /// <param name="oldConfiguration">The old configuration.</param>
-        /// <param name="newConfiguration">The new configuration.</param>
-        public LoggingConfigurationChangedEventArgs(LoggingConfiguration oldConfiguration, LoggingConfiguration newConfiguration)
+        /// <param name="activatedConfiguration">The new configuration.</param>
+        /// <param name="deactivatedConfiguration">The old configuration.</param>
+        public LoggingConfigurationChangedEventArgs(LoggingConfiguration activatedConfiguration, LoggingConfiguration deactivatedConfiguration)
         {
-            this.OldConfiguration = oldConfiguration;
-            this.NewConfiguration = newConfiguration;
+            this.ActivatedConfiguration = activatedConfiguration;
+            this.DeactivatedConfiguration = deactivatedConfiguration;
         }
 
         /// <summary>
         /// Gets the old configuration.
         /// </summary>
         /// <value>The old configuration.</value>
-        public LoggingConfiguration OldConfiguration { get; private set; }
+        public LoggingConfiguration DeactivatedConfiguration { get; private set; }
 
         /// <summary>
         /// Gets the new configuration.
         /// </summary>
         /// <value>The new configuration.</value>
-        public LoggingConfiguration NewConfiguration { get; private set; }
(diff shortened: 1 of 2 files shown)


src/NLog/LogFactory.cs
```diff
@@ -338,7 +338,7 @@ public LoggingConfiguration Configuration
                         }
                     }
 
-                    this.OnConfigurationChanged(new LoggingConfigurationChangedEventArgs(value, oldConfig));
+                    this.OnConfigurationChanged(new LoggingConfigurationChangedEventArgs(oldConfig, value));
```
304NotModified: so this is still a breaking change. 

I like to resolve this issue, but if it's breaking, it should be NLog 5.

304NotModified:
Not sure if there is a real benefit to introduce a breaking change which isn't really needed (even if it's NLog 5)

src/NLog/LogFactory.cs
snakefoot: No it is not a breaking change as I have "fixed" the now obsolete properties for LoggingConfigurationChangedEventArgs. Guess it should wait for NLog 4.5 because of the obsolete handling

304NotModified: If I call `new LoggingConfigurationChangedEventArgs(value, oldConfig)` after this PR has been merged, then it has a different behavior? Or am I missing something :)

snakefoot:
Well it fooled me, as I tried to add a configuration-changed event handler for the unit-tests, where I enabled throw-exceptions for all loaded configurations. Sadly enough it never worked, because I actually enabled it for the old-configuration. Others also think this is weird as you can see with #1122

304NotModified:
> Others also think this is weird as you can see with #1122

yes could't agree more.

> Sadly enough it never worked, because I actually enabled it for the old-configuration. 

and after this change. it could "magically" won't work for others am I afraid of.

snakefoot:
You are right. People who allocates the NLog LoggingConfigurationChangedEventArgs for use elsewhere will get a breaking change. I have updated the PR.

snakefoot:
@304NotModified This is actually not a breaking change. It now just introduces new properties and marks the old ones as obsolete. There is no change between the constructor-input-parameters and the behavior of the obsolete properties.

snakefoot:
@304NotModified Can we have this included for NLog 4.5 as it is not a breaking change. Just faulty properties marked as obsolete.

304NotModified:
isn't the breaking change here the reordered parameters in the ctor?

snakefoot:
> isn't the breaking change here the reordered parameters in the ctor?

How are the parameters re-ordered? Since NLog 4.0 broke LoggingConfigurationChangedEventArgs, then the first parameter has become the new active configuration.

304NotModified:
`new LoggingConfigurationChangedEventArgs(old, new)` needs to be `new LoggingConfigurationChangedEventArgs(new,old)` after this change?

snakefoot:
> new LoggingConfigurationChangedEventArgs(old, new) needs to be new LoggingConfigurationChangedEventArgs(new,old) after this change?

No matter what then NLog ver 3 users will feel the punishment of the NLog 4 breaking change.

In NLog 4 the first parameter has always been the new active configuration, it was just named wrong in the documentation.

304NotModified:
> In NLog 4 the first parameter has always been the new active configuration, it was just named wrong in the documentation

Ah, indeed, missed that.

304NotModified:
merged!

PR #1906, opened 2017-01-13
Inject basedir - allow one current IAppDomain, open IAppDomain in all platforms

304NotModified:
- having in two classes (LogManager & LogFactory) the (static) currentAppDomain is confusing
  - Moved logic to LogFactory, 
  - removed internal stuff in LogManager
  - same logic will be executed
- cleanup of event-hook so we don't double hook
- don't use AppDomainWrapper everywhere, but the static `IAppDomain` property
- remove old `#if` stuff and handle empty properties of `IAppDomain` 

Goal: inject basedir in another package, by using the `IAppDomain `

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1906)

As the interface is altered for SL, this could a breaking change for SL. But i'm pretty sure there aren't custom implementations of IAppDomain for SL in custom code.

### DIFF
--- a/src/NLog/Config/ConfigSectionHandler.cs
+++ b/src/NLog/Config/ConfigSectionHandler.cs
@@ -73,7 +73,7 @@ namespace NLog.Config
         /// <returns>The created section handler object.</returns>
         object IConfigurationSectionHandler.Create(object parent, object configContext, XmlNode section)
         {
-            return Create(section, AppDomainWrapper.CurrentDomain);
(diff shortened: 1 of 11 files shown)


304NotModified:
@snakefoot would you like to review my change? Thx in advance!

snakefoot:
@304NotModified Looks good to me. Like the removal of the if-defined. 

Since LogManager is pretty much a static interface to the LogFactory, then it makes sense there is no "logic" in LogManager.

PR #1912, opened 2017-01-13
WebServiceTarget - Enable OptimizeBufferReuse by default & Ability to add HTTP request headers

snakefoot:
Waiting for NLog ver. 4.5

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1912)

### DIFF
--- a/src/NLog/Targets/MethodCallParameter.cs
+++ b/src/NLog/Targets/MethodCallParameter.cs
@@ -34,7 +34,6 @@
 namespace NLog.Targets
 {
     using System;
-    using System.Globalization;
     using NLog.Config;
     using NLog.Layouts;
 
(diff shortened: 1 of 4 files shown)


snakefoot:
Added support for HttpWebRequest Headers. Not sure if it should be pushed to NLog 5.0, as it is a breaking change.

304NotModified:
Where is the breaking change?

snakefoot:
> Where is the breaking change?

This method is now throwing an exception:

```
protected override void DoInvoke(object[] parameters, AsyncContinuation continuation)
{
   // method is not used, instead asynchronous overload will be used
}
```

As it has been replaced by this override method:

`protected override void DoInvoke(object[] parameters, AsyncLogEventInfo logEvent)`

snakefoot:
@304NotModified Now modified the code, so it is no longer a breaking change.

304NotModified:
\0/ thanks!

snakefoot:
Updated documentation:

https://github.com/NLog/NLog/wiki/WebService-target

Issue #1530, opened 2016-07-01
NLog Extension Security


INotDisposable:
**Type** (choose one):  Feature request
**NLog version**: 4.3.5
**Platform**: All

Currently NLog will load any extension files mentioned in the config XML, or any DLL with the name 'NLog*.dll'.  Our product is sometimes installed on systems that use the FAT32 or exFAT filesystem, and thus does not have any integrated filesystem security.  Placing a rogue NLog*.dll file in our application folder would allow someone to inject code into our logger.

All of our files are signed and timestamped according to Microsoft's Authenticode standards.  It would be nice if we could have a callback or something executed when NLog wants to load an extension; our application could then check the validity of the extension before allowing it to load.  Perhaps something like this:

```
void LogManager.RegisterExtensionValidator(Func<string, bool> validator)
```

The string passed to the method would be the path to the extension DLL, and it would return 'true' to continue loading the extension or 'false' to prevent it from loading.  I could then use this to validate the file's signature against our signing certificate and the timestamp on the file.

304NotModified:
I see the problem.  But a path validation still looks unsecure? We can replace files. 

More thinking of public key validation or disabling loading.

304NotModified:
Also, as an attacker I could easily fork nlog, add the evil code and replace the full dll. 

;)

INotDisposable:
LOL that's true.  For our product specifically, however, one of the first things we do at startup is validate the signature on all files in our installation directory; if any DLL or EXE does not appear to be signed with our code-signing certificate or if the timestamp is wrong, we shut down.  I think that prevents the NLog.dll replacement issue.

My thought with path validation was giving the host application the ability to validate files in whatever way makes sense for the application.  For our application, the Authenticode validation we perform on all other installed files could also be run on any extension classes NLog finds and wants to load.  Other applications might simply check the file hash, or hard-code a return value of 'false' if they never want extensions to load.

304NotModified:
Perhaps it's neater to fire an event and reading the return value?

e.g.

```c#
LogManager.OnExtensionLoaded += loaded;


loaded(CancelEventArgs e) //todo more context
{
  .. e.Cancel = ...
}
```

INotDisposable:
I like it.  I can take a stab at implementing it if that would be helpful.  Thanks!

304NotModified:
That's would be nice :)

304NotModified:
event will be added in 4.4.4!

PR #1917, opened 2017-01-14
Call NLogPackageLoader.Preload (static) for NLog packages on load

304NotModified:
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1917)

related: https://github.com/NLog/NLog/issues/1530

### DIFF
--- a/src/NLog/Config/ConfigurationItemFactory.cs
+++ b/src/NLog/Config/ConfigurationItemFactory.cs
@@ -65,7 +65,7 @@ namespace NLog.Config
         private readonly Factory<TimeSource, TimeSourceAttribute> timeSources;
 
         private IJsonSerializer jsonSerializer = DefaultJsonSerializer.Instance;
-        
+
         /// <summary>
         /// Initializes a new instance of the <see cref="ConfigurationItemFactory"/> class.
         /// </summary>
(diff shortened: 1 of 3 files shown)


src/NLog/Config/ConfigurationItemFactory.cs
```diff
@@ -223,13 +223,70 @@ public void RegisterItemsFromAssembly(Assembly assembly, string itemNamePrefix)
...
+                        try
+                        {
+                            preloadMethod.Invoke(null, null);
+                        }
+                        catch (Exception e)
+                        {
+                            InternalLogger.Warn(e,"Preload invoked for '{0}' failed", type.FullName);
+                        }
+                       
+                        InternalLogger.Debug("Preload succesfully invoked for '{0}'", type.FullName);
```
UgurAldanmaz: This line will be executed, even if invoke has been failed. I suggest that move this line to the line 267 (after the invoke).

src/NLog/Config/ConfigurationItemFactory.cs
```diff
@@ -223,13 +223,70 @@ public void RegisterItemsFromAssembly(Assembly assembly, string itemNamePrefix)
...
+                    }
+                    else
+                    {
+                        InternalLogger.Debug("NLogPackageLoader preload method, but isn't static");
```
UgurAldanmaz: Log message would be improved as:

NLogPackageLoader contains a preload method, but it isn't static.

src/NLog/Config/ConfigurationItemFactory.cs
```diff
@@ -223,13 +223,70 @@ public void RegisterItemsFromAssembly(Assembly assembly, string itemNamePrefix)
...
+            var types = typesToScan.Where(t => t.Name.Equals("NLogPackageLoader", StringComparison.OrdinalIgnoreCase));
+
+            foreach (var type in types)
+            {
+                CallPreload(type);
+            }
+        }
+
+        /// <summary>
+        /// Call Preload for <paramref name="type"/>. The method should be static.
```
UgurAldanmaz: It would be improved as:

Calls the preload method for `<paramref name="type"/>`. The preload method should be static.

304NotModified:
@UgurAldanmaz thanks for the review!

src/NLog/Config/ConfigurationItemFactory.cs
304NotModified: 👍

src/NLog/Config/ConfigurationItemFactory.cs
304NotModified: oops refactor bug! Thanks! Have also improved the unit test

304NotModified:
@UgurAldanmaz I'm doubting about the naming. Currently it calls `NLogPackageLoader.Preload`, but not sure if the name is good enough (and unique enough). What do you think?

304NotModified:
@UgurAldanmaz thanks for the review

PR #1926, opened 2017-01-22
RetryingWrapper - Allow closing target, even when busy retrying

snakefoot:
If wrapped target is working on retrying for ever, then it should not block one from closing target.

This is a breaking change, as the RetryingWrapper is not sealed. Has to wait for NLog 5.0

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1926)

### DIFF
--- a/src/NLog/Targets/Wrappers/AsyncTargetWrapper.cs
+++ b/src/NLog/Targets/Wrappers/AsyncTargetWrapper.cs
@@ -210,11 +210,14 @@ namespace NLog.Targets.Wrappers
         /// </summary>
         protected override void CloseTarget()
         {
-            lock (this.lockObject)
+            this.StopLazyWriterThread();
+            if (this.lazyWriterTimer == null)
             {
-                this.StopLazyWriterThread();
-                WriteEventsInQueue(int.MaxValue, "Closing Target");
+                // Timer has stopped
+                lock (this.lockObject)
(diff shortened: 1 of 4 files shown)


304NotModified:
And thanks again!!

304NotModified:
> This is a breaking change, as the RetryingWrapper is not sealed. Has to wait for NLog 5.0

Oops missed that one.

304NotModified:
isn't .RetryingWrapper.Write calling RetryingWrapper.WriteAsyncThreadSafe (inidrect) and so it's a non-breaker?

304NotModified:
https://github.com/NLog/NLog/pull/1933 should fix this IMO. I assume that the retry logic is safe without lock?

snakefoot:
> #1933 should fix this IMO. I assume that the sleep-retry logic is safe without lock.

@304NotModified When overriding WriteAsyncThreadSafe(), then it will never reach the normal Write(). This is intended as the normal Write() is called after having taken target-lock. It is the target-lock I want to avoid as it blocks one from flushing and closing the target. If you re-introduce the RetryingTargetWrapper-Write() again then it will never be reached.

The sleep-retry-logic was before protected against reordering by the Write-Target-lock, with this change then it is protected against reordering by the RetrySyncObject-lock.

PR #1927, opened 2017-01-22
InternalLogger - Better support for multiple threads when using file

### LINKED ISSUE #1924: InternalLogger locks its own logging file
Hi.
Maybe i am doing something wrong (hopefully), but i don't think so.

We are using NLog with multiple targets and on multiple threads and the internal logger to log exceptions which could occure in our custom target implementation to a file. While testing this implementation we encountered the following error:

>  VS2015 Output:
>  TearDown : System.IO.IOException : The process cannot access the file "C:\logs\internalException.log" because it is being used by another process.

Our configuration looks like this:

```c#
public class LoggerFactory : ILoggerFactory<LoggerConfigurationBase>
{
	public Contracts.ILogger Create(LoggerConfigurationBase loggerConfig, InternalLogConfiguration internalConfiguration)
    {        
        TargetWithLayout target = loggerConfig.GetTarget();
            
        ConfigurationItemFactory.Default.LayoutRenderers.RegisterDefinition("LogEntryLayoutRenderer", typeof(LogEntryLayoutRenderer));
        target.Layout = "${LogEntryLayoutRenderer}";
    
        LoggingRule loggingRule = new LoggingRule(loggerConfig.LoggerName, LogLevel.Info, target);
    
        LogManager.Configuration.LoggingRules.Add(loggingRule);
    
        if (internalConfiguration != null)
        {
            LogManager.ThrowExceptions = internalConfiguration.ThrowExceptions;
            InternalLogger.LogFile = internalConfiguration.Filename;
            InternalLogger.LogLevel = this.ConvertFromSeverity(internalConfiguration.Severity);
            InternalLogger.LogToTrace = internalConfiguration.LogToTrace;
        }
    
        return new ViaLogger(LogManager.GetLogger(loggerConfig.LoggerName));
    }
}
```

The Test we are looking like this:
```c#
[TestFixture]
[ExcludeFromCodeCoverage]
[Category(TestCategories.IntegrationTest)]
[Category("ExceptionHandlingTest")]
[Parallelizable(ParallelScope.None)]
public class ExceptionHandlingTests
{
    private readonly String internalExceptionLogFile = "C:/logs/internalException.log";

	[OneTimeSetUp]
    public void ClassSetUp()
    {
        if (File.Exists(this.internalExceptionLogFile))
        {
            File.Delete(this.internalExceptionLogFile);
        }
    }

    [OneTimeTearDown]
    public void ClassTearDown()
    {
        if (File.Exists(this.internalExceptionLogFile))
        {
            //System.IO.IOException : The process cannot access the file "Disk\Folder\internalException.log" because it is being used by another process.
            File.Delete(this.internalExceptionLogFile);
        }
    }

    [SetUp]
    public void TestSetup()
    {
        Thread.CurrentThread.CurrentUICulture = new CultureInfo("en-US");
    }
	
	[Test]
	[TestCase(TestName = "Log one entry into text file with invalid path should log exception into internal log")]
	//Test Failed: InternalLogFile contains not expected error message. Expected: True, But was:  False
	public void LogTextFile_PathNotexists_ShouldWriteExceptionInInternalTextFile()
	{
		LoggerFactory textFileFactory = new LoggerFactory();
		ILogger textFilelogger = textFileFactory.Create(
			new TextfileConfiguration(
				"ExceptionHandlingTextFileLoggerWithInternalLogFile",
				"ExceptionHandlingTestTextFileTarget",
				"ABC:/Path_Not_Exists/ExceptionHandlingTest.txt"),
			new InternalLogConfiguration()
			{
				Filename = this.internalExceptionLogFile,
				LogToTrace = true,
				Severity = Severities.Trace
			}
		);
		textFilelogger.Log(Severities.Info, "TextFile TestMessage");
		FileAssert.Exists(this.internalExceptionLogFile);
	
		String fileContent = new StreamReader(this.internalExceptionLogFile).ReadToEnd();
		String expectedErrorMessage = "Error Error has been raised. Exception: System.NotSupportedException: The given path's format is not supported.";
		Assert.That(fileContent.Contains(expectedErrorMessage), Is.True, "InternalLogFile contains not expected error message");
	}
	
	[Test]
	[TestCase(TestName = "Log one entry into elastic search with invalid url should log exception into internal log")]
	//Test passes
	public void LogElasticSearch_UrlNotexists_ShouldWriteExceptionInInternalTextFile()
	{
		Uri[] nodeUris =
		{
			new Uri("https://domain-not.exists")
		};
		LoggerFactory elasticSearchfactory = new LoggerFactory();
		ILogger elasticSearchlogger = elasticSearchfactory.Create(
			new ElasticsearchConfiguration(
				"ExceptionHandlingElasticSearchLoggerWithInternalLogFile",
				"ExceptionHandlingTestElasticTarget",
				"ExceptionHandlingTest",
				nodeUris),
			new InternalLogConfiguration()
			{
				Filename = this.internalExceptionLogFile,
				LogToTrace = true
			}
		);
		elasticSearchlogger.Log(Severities.Info, "ElasticSearch TestMessage");
		FileAssert.Exists(this.internalExceptionLogFile);
	
		String fileContent = new StreamReader(this.internalExceptionLogFile).ReadToEnd();
		String expectedErrorMessage = "PingFailure: Node: https://domain-not.exists/ Exception: PipelineException";
		Assert.That(fileContent.Contains(expectedErrorMessage), Is.True, "InternalLogFile contains not expected error message");
	}
}
```

As you can see we are using nunit to perform our tests. The issue only happen when we run these to test together.

In my understanding the file used by the internal logger should be closed after an logentry is written and the filehandle should be released, which it doesnt't do.

My question now: Am i doing something wrong? If no, is there a solution/workaround for this issue?

snakefoot:
Extra lock-operation has been added when doing exclusive file-lock, to improve support for 
multi-threaded use of the InternalLogger.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1927)

### DIFF
--- a/src/NLog/Common/InternalLogger.cs
+++ b/src/NLog/Common/InternalLogger.cs
@@ -250,9 +250,12 @@ namespace NLog.Common
                 var logFile = LogFile;
                 if (!string.IsNullOrEmpty(logFile))
                 {
-                    using (var textWriter = File.AppendText(logFile))
+                    lock (LockObject)
                     {
-                        textWriter.WriteLine(msg);
+                        using (var textWriter = File.AppendText(logFile))
+                        {
+                            textWriter.WriteLine(msg);
+                        }
                     }
                 }
 
@@ -269,14 +272,21 @@ namespace NLog.Common
                 // log to console
                 if (LogToConsole)
                 {
-                    Console.WriteLine(msg);
+                    lock (LockObject)
+                    {
+                        Console.WriteLine(msg);
+                    }
                 }
 
                 // log to console error
                 if (LogToConsoleError)
                 {
-                    Console.Error.WriteLine(msg);
+                    lock (LockObject)
+                    {
+                        Console.Error.WriteLine(msg);
+                    }
                 }
+
 #if !SILVERLIGHT && !__IOS__ && !__ANDROID__
                 WriteToTrace(msg);
 #endif
@@ -290,7 +300,6 @@ namespace NLog.Common
                 {
                     throw;
                 }
-
             }
         }
 


src/NLog/Common/InternalLogger.cs
```diff
@@ -294,6 +312,63 @@ private static void Write([CanBeNull]Exception ex, LogLevel level, string messag
...
+                        _writeToFileAtomicSuccesCount = -1;
+                    }
+                }
+                return false;
+            }
+            catch
+            {
+                lock (LockObject)
+                {
+                    _writeToFileAtomicSuccesCount = -1;
```
304NotModified: should be use Interlocked.Decrement?

src/NLog/Common/InternalLogger.cs
```diff
@@ -294,6 +312,63 @@ private static void Write([CanBeNull]Exception ex, LogLevel level, string messag
...
+#endif
+
+        static bool WriteToFileAtomic(string msg, string logFile)
+        {
+#if !SILVERLIGHT && !__IOS__ && !__ANDROID__ && !MONO
+            try
+            {
+                if (_writeToFileAtomicSuccesCount == 0)
+                {
+                    if (PlatformDetector.IsMono || !PlatformDetector.IsDesktopWin32)
```
304NotModified: why we need this case?

snakefoot:
> also not sure about this one. The internallogger should be as simple as possible?

I have removed the support for atomic writes, and instead just implemented a lock around the file-write. Then it will at least support multi-threaded file-logging.

304NotModified:
> @304NotModified I have removed the support for atomic writes, and instead just implemented a lock around the file-write. Then it will at least support multi-threaded file-logging.

Good one!

src/NLog/Common/InternalLogger.cs
```diff
@@ -246,40 +246,55 @@ private static void Write([CanBeNull]Exception ex, LogLevel level, string messag
...
                 }
-
-                // log to LogWriter
-                var writer = LogWriter;
-                if (writer != null)
+                finally
                 {
-                    lock (LockObject)
+                    // log to LogWriter
+                    var writer = LogWriter;
```
304NotModified: why is this code in the `finally`?  Is this really by-design? 

I only use `finally` to clean up stuff, never to do functional logic. Am I missing here something?

snakefoot: Just wanted to ensure that it also tried the logger-targets in case the file-logger failed. But I have now reverted it. Not that important to me.

304NotModified:
:shipit:

304NotModified:
Thanks for the changes. More more KISS now :)

Issue #1633, opened 2016-09-11
dotnet core nlog running on Linux logs only one entry


martinto:
**Type** (choose one): Bug
**NLog version**: v4.4.0-beta15
**Platform**: .NET Core 
**Current NLog config** (xml or C#, if relevant)

``` xml
    <?xml version="1.0" encoding="utf-8" ?>
    <nlog 
        xmlns="http://www.nlog-project.org/schemas/NLog.xsd" 
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
        <extensions>
            <!--enable NLog.Web for ASP.NET Core-->
            <add assembly="NLog.Web.AspNetCore"/>
        </extensions>
        <targets>
            <!-- https://github.com/nlog/nlog/wiki/File-target -->
            <!-- http://stackoverflow.com/questions/34679727/use-nlog-in-asp-net-5-application -->
            <target name="allfile" xsi:type="AsyncWrapper" queueLimit="5000" overflowAction="Discard">
                <target name="allfile" xsi:type="File"
                        fileName="${basedir}/logs/all-logfile.csv"
                        archiveFileName="${basedir}/logs/archives/all-log.{#}.csv"
                        archiveEvery="Day"
                        archiveNumbering="Date"
                        archiveDateFormat="yyyyMMdd"
                        maxArchiveFiles="28"
                        concurrentWrites="true"
                        keepFileOpen="false"
                        encoding="utf-8">
                    <layout xsi:type="CSVLayout">
                        <column name="time" layout="${longdate}" />
                        <column name="logger" layout="${logger}" />
                        <column name="authtype" layout="${aspnet-user-authtype}" />
                        <column name="identity" layout="${aspnet-user-identity}" />
                        <column name="level" layout="${uppercase:${level}}" />
                        <column name="message" layout="${message} ${exception:format=ToString,StackTrace}" />
                    </layout>
                </target>
            </target>
            <target name="ownfile" xsi:type="AsyncWrapper" queueLimit="5000" overflowAction="Discard">
                <target name="ownfile" xsi:type="File"
                        fileName="${basedir}/logs/own-logfile.csv"
                        archiveFileName="${basedir}/logs/archives/own-log.{#}.csv"
                        archiveEvery="Day"
                        archiveNumbering="Date"
                        archiveDateFormat="yyyyMMdd"
                        maxArchiveFiles="28"
                        concurrentWrites="true"
                        keepFileOpen="false"
                        encoding="utf-8">
[...]

PR #1931, opened 2017-01-24
FileTarget - Validate File CreationTimeUtc when non-Windows

snakefoot:
Simple fallback to FileLastWriteTimeUtc, when FileCreationTimeUtc is very old (when non-windows). Simple fix for #1633 

Also fixes bug in UnixMultiProcessFileAppender, where the fileExists-check was wrongly done after having created the file-descriptor.

This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1931)

### DIFF
--- a/src/NLog/Internal/FileAppenders/BaseFileAppender.cs
+++ b/src/NLog/Internal/FileAppenders/BaseFileAppender.cs
@@ -337,24 +337,19 @@ namespace NLog.Internal.FileAppenders
 
         private void UpdateCreationTime()
         {
-            if (File.Exists(this.FileName))
+            FileInfo fileInfo = new FileInfo(this.FileName);
+            if (fileInfo.Exists)
             {
-#if !SILVERLIGHT
-                this.CreationTimeUtc = File.GetCreationTimeUtc(this.FileName);
-#else
-                this.CreationTimeUtc = File.GetCreationTime(this.FileName);
-#endif
+                this.CreationTimeUtc = FileCharacteristicsHelper.ValidateFileCreationTime(fileInfo, (f) => f.GetCreationTimeUtc(), (f) => f.GetLastWriteTimeUtc()).Value;
             }
             else
             {
                 File.Create(this.FileName).Dispose();
+                this.CreationTimeUtc = DateTime.UtcNow;
 
 #if !SILVERLIGHT
-                this.CreationTimeUtc = DateTime.UtcNow;
                 // Set the file's creation time to avoid being thwarted by Windows' Tunneling capabilities (https://support.microsoft.com/en-us/kb/172190).
                 File.SetCreationTimeUtc(this.FileName, this.CreationTimeUtc);
-#else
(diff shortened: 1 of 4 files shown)


src/NLog/Internal/FileAppenders/FileAppenderCache.cs
```diff
@@ -439,7 +452,16 @@ public Mutex GetArchiveMutex(string fileName)
                 var fileInfo = new FileInfo(filePath);
                 if (fileInfo.Exists)
                 {
-                    return Time.TimeSource.Current.FromSystemTime(fileInfo.GetCreationTimeUtc());
+                    result = fileInfo.GetCreationTimeUtc();
+                    if (result.Value.Year < 1980)
```
304NotModified: could we get rid of the code duplication?

Maybe something like `useFallBack(dateTime r) => r.Year < 1980 && !PlatformDetector.IsDesktopWin32` ?

304NotModified: or at least move the 1980 to a (internal) const.

snakefoot:
> some code duplication notes

@304NotModified Have now created a helper-method, that handles the fallback logic.

src/NLog/Internal/FileAppenders/BaseFileAppender.cs
```diff
@@ -337,24 +337,19 @@ private FileStream TryCreateFileStream(bool allowFileSharedWriting)
...
-            if (File.Exists(this.FileName))
+            FileInfo fileInfo = new FileInfo(this.FileName);
+            if (fileInfo.Exists)
             {
-#if !SILVERLIGHT
-                this.CreationTimeUtc = File.GetCreationTimeUtc(this.FileName);
-#else
-                this.CreationTimeUtc = File.GetCreationTime(this.FileName);
-#endif
+                this.CreationTimeUtc = FileCharacteristicsHelper.ValidateFileCreationTime(fileInfo, (f) => f.GetCreationTimeUtc(), (f) => f.GetLastWriteTimeUtc()).Value;
```
304NotModified: not trying to be nitpicking, but isn't this infecting the performance? AFAIK lamba's have there costs, and this code is called a lot?

snakefoot: This code is called when opening the file. So it is only when using the RetryingMultiProcessAppender it is called a lot.

Also the lambdas are completely static without capture of this, so the compiler should turn them into zero allocations.

Ran a performance profiler of memory allocations using the RetryingMultiProcessAppender, and it within the BaseFileAppender.UpdateCreationTime() then the most expensive was the FileInfo-allocation, and the second most expensive was assigning the CreationTimeUtc where it used DateTime.ToLocalTime().

304NotModified: 👍

304NotModified:
partly fixes https://github.com/NLog/NLog/issues/1633, as this is merged to master and not (yet) to CoreCLR branch

PR #1932, opened 2017-01-25
FilePathLayout - Faster rendering of filename, when not ThreadAgnostic

snakefoot:
#1071 improved from 550.000 msgs/sec to 1.500.000 msgs/sec.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1932)

### DIFF
--- a/src/NLog/Internal/FilePathLayout.cs
+++ b/src/NLog/Internal/FilePathLayout.cs
@@ -164,6 +164,13 @@ namespace NLog.Internal
 
             if (reusableBuilder != null)
             {
+                if (!_layout.IsThreadAgnostic)
+                {
+                    string cachedResult;
+                    if (logEvent.TryGetCachedLayoutValue(_layout, out cachedResult))
+                        return cachedResult;
+                }
+
                 _layout.RenderAppendBuilder(logEvent, reusableBuilder);
 
                 if (_cachedPrevRawFileName != null && _cachedPrevRawFileName.Length == reusableBuilder.Length)


src/NLog/Internal/FilePathLayout.cs
```diff
@@ -164,6 +164,13 @@ private string GetRenderedFileName(LogEventInfo logEvent, System.Text.StringBuil
 
             if (reusableBuilder != null)
             {
+                if (!_layout.IsThreadAgnostic)
```
304NotModified: I'm doubting if this should be `if (_layout.IsThreadAgnostic)`, as ThreadAgnostic stuff could be calculated on every thread. (and thus could be lazy). 

non-ThreadAgnostic should be calculated on the main thread.

snakefoot: IsThreadAgnostic means that it doesn't have to be precalculated by async-wrappers or buffering-wrappers. As none of the layouts requires the context-information of the logging-thread.

But it it requires the context-information of the logging-thread, and there is a async-wrapper or buffering-wrapper. Then it is possible to extract the precalculated layout-rendering-result.

snakefoot: It actually checks if there is a chance that the main-thread has already rendered the logevent using the layout. And if so then dives into the logevent-cache of pre-rendered-layout-results.

304NotModified:
Thanks for the explanation!

PR #1937, opened 2017-01-28
LogManager.Shutdown - Verify that active config exists

snakefoot:
Small enhancement to  #1899, so it doesn't attempt to load config if it wasn't loaded already.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1937)


304NotModified:
👍

PR #1938, opened 2017-01-28
AsyncTargetWrapper - Always use timerLockObject when changing Timer

snakefoot:
And use writeLockObject to throttle instant timers.

Fixes bug that has been introduced with #1926, that can seen by the following build:

https://ci.appveyor.com/project/nlog/nlog/build/4.4.4435

Have run performance tests, and there is a 1 pct decrease in performance, because it now has to take two locks to schedule the instant timer. 

One might consider to stop using instant timers (Heavy use of Timers will cause lock congestion because of the global TimeQueue.Instance-lock inside the NET framework), but instead just use QueueUserWorkItem. Did a small performance test and there was 0-1 pct increase in performance, but discovered that QueueUserWorkItem generates garbage allocation.

I have now changed the logic so it only schedules a new timer, if something in the queue (instead of being optimistic about more soon entering the queue). This minimizes the number of Timer.Change-calls and should decrease the load on the global TimeQueue.Instance-lock

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1938)

### DIFF
--- a/src/NLog/Targets/Wrappers/AsyncTargetWrapper.cs
+++ b/src/NLog/Targets/Wrappers/AsyncTargetWrapper.cs
@@ -77,7 +77,7 @@ namespace NLog.Targets.Wrappers
     [Target("AsyncWrapper", IsWrapper = true)]
     public class AsyncTargetWrapper : WrapperTargetBase
     {
-        private readonly object lockObject = new object();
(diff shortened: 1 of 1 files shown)


304NotModified:
LGTM

thanks!

PR #1933, opened 2017-01-26
RetryingWrapper  backwardscomp with 4.4.1

304NotModified:
Enhancement of https://github.com/NLog/NLog/pull/1926

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1933)

PR #1943, opened 2017-02-04
RetryingWrapper - Reintroduced Write-method to make it almost non-breaking

snakefoot:
Trying to fix-up the breaking change introduced with #1926

The breaking change is now that RetrySyncObject is used to protect the Write-method, instead of the standard Target.SyncRoot.

Guess it replaces #1933

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1943)


304NotModified:
Thanks!

Issue #1376, opened 2016-04-15
Add structural logging (proposals/discussion/prototypes)


304NotModified:
Add structural logging support like Serilog. This is a call for proposals/discussion/prototypes

Why?
- It's used in ASPNET5 logging  / DNX logging of Microsoft
- It works well with Common Logging
- It seems popular
- I like it :)

We need a prototype on this.

Things to check:
- It should be optional (config setting in XML and LogConfiguration) - one global setting is good enough for now.
- We should check which syntax Microsoft is using
- We should consider if Serillog's syntax is the best
- We must list the syntax we will implement (here)
- The syntax should be future proof (extenable, interpolated strings)
- We should cache parsing the Layouts
- Proposal to write the arguments to "event properties"

See also http://nlog.userecho.com/topics/21-make-structural-logging-possible/

304NotModified:
Serilog syntax:

- `{propertyname}`
- `{propertyname:format}`
- escape `{` by `{{` (like `string.format`)
- prefixes: `@` and `$`
- > If any of the property names are non-numeric, then all property names will be matched from left-to-right with the log method's parameters

304NotModified:
Microsoft's logging in ASP.NET core (aka ASP.NET 5) / DNX

Code, but undocumented:
- https://github.com/aspnet/Logging/blob/c3f04e4b6f5413d93b73e5519d2674dc5e900a0b/src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs

304NotModified:
Some thoughts:

- No need for format or destructors in NLog, we have layout renderers?
- sending an object to a log method and take its properties sounds nice, but check the performance. We need `dynamic ` for performance?

nblumhardt:
Personally I think this would be awesome :-) A couple of other impls: https://github.com/adamchester/messagetemplates, https://github.com/logary/logary/pull/90. Because it's now supported in a few places some have adopted the term "message templates" rather than Serilog-like to describe the named-holes/capturing techinque.

Regarding your note above, Serilog ended up with the design it uses here because in Serilog, the "serialization" can be offloaded for async/background processing. Snapshotting ("destructuring") not-known-to-be-immutable values at the point of logging prevents threading issues in this case.

[...]

PR #1951, opened 2017-02-10
[WIP] Structured events (NLog 4.5)

304NotModified:
Structured events like Serilog.

Fixes https://github.com/NLog/NLog/issues/1376

(Branched from NLog 4.4.2)

TODO

- [ ] Fix builds Xamarin/Wp8 builds - not included due to Nuget restore issues
- [ ] Fix Travis build  - not included due to NuGet restore issues
- [ ] more tests, tests for eventloginfo properties
- [ ] CLS compliant
- [ ] default json serializer 
    - [ ] culture aware (number formats etc)
    - [ ] nullable values, 
    - [ ] list of objects 
    - [ ] etc
- [x] fix infinite depth objectSerializer
- [ ] better default serializer: 
- [ ] fix/implement [milestone 1 issues of NLog.StructuredEvents](https://github.com/NLog/NLog.StructuredEvents/milestone/1)
- [ ] fix Resharper String Format attribute (complains that `{a}` is not valid)
- [ ] fix test "NLog.UnitTests.Internal.AppDomainPartialTrustTests.MediumTrustWithExternalClass" (currently disabled). Current result for all files (debug.txt, Error.txt etc) - 
[see screenshot](https://cloud.githubusercontent.com/assets/5808377/22858210/a07d6970-f0b6-11e6-8775-ae618baa6d59.png) - @snakefoot any idea why this testcase is behaving so strange with this change? It all prints " 7" on every line...

Discussion:
- [ ] Default JSON or JSON5? (single quotes, no quotes for keys)
- [ ] should we print "null" values. E.g `logger.Log("hello {Name}", null);`
- [ ] should we print properties with `null` value?
- [ ] lists: `[1,2,3] ` or `1,2,3` ? or `1, 2, 3` or maybe `1, 2 & 3`!
- [x] capture all properties, and if positional use "1" as key 
- [ ] backwardscomp capture all properties

This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1951)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,7 +1,7 @@
 version: 4.5.{build}
 clone_folder: c:\projects\nlog
 build_script:                                                                                                                
-  - msbuild  src/NLog.proj /verbosity:minimal /t:rebuild /t:xsd /p:BuildNetFX35=true /p:BuildNetFX40=true /p:BuildNetFX45=true  /t:NuGetPackage /p:BuildVersion=4.5-alpha03  /p:AssemblyFileVersion=4.5.%APPVEYOR_BUILD_NUMBER% /p:BuildLastMajorVersion=4.0.0 /p:configuration=release /p:BuildLabelOverride=NONE 
+  - msbuild  src/NLog.proj /verbosity:minimal /t:rebuild /t:xsd /p:BuildNetFX35=true /p:BuildNetFX40=true /p:BuildNetFX45=true  /t:NuGetPackage /p:BuildVersion=4.5-alpha04  /p:AssemblyFileVersion=4.5.%APPVEYOR_BUILD_NUMBER% /p:BuildLastMajorVersion=4.0.0 /p:configuration=release /p:BuildLabelOverride=NONE 
   - msbuild  src/NLog.proj /verbosity:minimal /t:BuildTests /p:BuildNetFX35=true /p:BuildNetFX40=true /p:BuildNetFX45=true 
  
 test_script:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,11 +1,11 @@
-version: 4.4.{build}
+version: 4.5.{build}
 clone_folder: c:\projects\nlog
 build_script:                                                                                                                
-  - msbuild  src/NLog.proj /verbosity:minimal /t:rebuild /t:xsd /t:NuGetPackage /p:BuildVersion=4.4.2  /p:AssemblyFileVersion=4.4.%APPVEYOR_BUILD_NUMBER% /p:BuildLastMajorVersion=4.0.0 /p:configuration=release /p:BuildLabelOverride=NONE 
+  - msbuild  src/NLog.proj /verbosity:minimal /t:rebuild /t:xsd /p:BuildNetFX35=true /p:BuildNetFX40=true /p:BuildNetFX45=true  /t:NuGetPackage /p:BuildVersion=4.5-alpha03  /p:AssemblyFileVersion=4.5.%APPVEYOR_BUILD_NUMBER% /p:BuildLastMajorVersion=4.0.0 /p:configuration=release /p:BuildLabelOverride=NONE 
   - msbuild  src/NLog.proj /verbosity:minimal /t:BuildTests /p:BuildNetFX35=true /p:BuildNetFX40=true /p:BuildNetFX45=true 
  
 test_script:
-  - tools\CheckSourceCode\NLog.SourceCodeTests.exe no-interactive
+#  - tools\CheckSourceCode\NLog.SourceCodeTests.exe no-interactive
   - xunit.console.clr4 "build\bin\Debug\.NET Framework 3.5\NLog.UnitTests.dll" /appveyor /noshadow
   - xunit.console.clr4 "build\bin\Debug\.NET Framework 4.0\NLog.UnitTests.dll" /appveyor /noshadow
   - nuget.exe install OpenCover -ExcludeVersion
--- a/src/NLog.Extended/NLog.Extended.netfx35.csproj
+++ b/src/NLog.Extended/NLog.Extended.netfx35.csproj
@@ -41,6 +41,10 @@
     <DocumentationFile>$(OutputPath)\$(AssemblyName).xml</DocumentationFile>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="NLog.StructuredEvents, Version=0.1.0.0, Culture=neutral, PublicKeyToken=5120e14c03d0593c, processorArchitecture=MSIL">
+      <HintPath>..\packages\NLog.StructuredEvents.0.3.1\lib\net35\NLog.StructuredEvents.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
     <Reference Include="System" />
     <Reference Include="System.Configuration" />
     <Reference Include="System.Core">
@@ -66,6 +70,9 @@
   </ItemGroup>
   <ItemGroup>
     <CodeAnalysisDictionary Include="CustomDictionary.xml" />
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="packages.config" />
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
--- a/src/NLog.Extended/NLog.Extended.netfx40.csproj
+++ b/src/NLog.Extended/NLog.Extended.netfx40.csproj
@@ -44,6 +44,10 @@
     <DocumentationFile>$(OutputPath)\$(AssemblyName).xml</DocumentationFile>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="NLog.StructuredEvents, Version=0.1.0.0, Culture=neutral, PublicKeyToken=5120e14c03d0593c, processorArchitecture=MSIL">
+      <HintPath>..\packages\NLog.StructuredEvents.0.3.1\lib\net35\NLog.StructuredEvents.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
     <Reference Include="System" />
     <Reference Include="System.Configuration" />
     <Reference Include="System.Core">
@@ -70,7 +74,9 @@
       <Name>NLog.netfx40</Name>
     </ProjectReference>
   </ItemGroup>
-  <ItemGroup />
+  <ItemGroup>
+    <None Include="packages.config" />
+  </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
        Other similar extension points exist, see Microsoft.Common.targets.
--- a/src/NLog.Extended/NLog.Extended.netfx45.csproj
+++ b/src/NLog.Extended/NLog.Extended.netfx45.csproj
@@ -48,6 +48,10 @@
     <DocumentationFile>$(OutputPath)\$(AssemblyName).xml</DocumentationFile>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="NLog.StructuredEvents, Version=0.1.0.0, Culture=neutral, PublicKeyToken=5120e14c03d0593c, processorArchitecture=MSIL">
+      <HintPath>..\packages\NLog.StructuredEvents.0.3.1\lib\net35\NLog.StructuredEvents.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
     <Reference Include="System" />
     <Reference Include="System.Configuration" />
     <Reference Include="System.Core">
@@ -74,7 +78,9 @@
       <Name>NLog.netfx45</Name>
     </ProjectReference>
   </ItemGroup>
-  <ItemGroup />
+  <ItemGroup>
+    <None Include="packages.config" />
+  </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
        Other similar extension points exist, see Microsoft.Common.targets.
--- a/src/NLog/LogEventInfo.cs
+++ b/src/NLog/LogEventInfo.cs
@@ -31,6 +31,8 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
+using NLog.StructuredEvents;
+
 namespace NLog
 {
     using System;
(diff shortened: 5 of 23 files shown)


304NotModified:
To do. Sync projects, include sl, xamarin support

biiiipy:
**Default JSON or JSON5? (single quotes, no quotes for keys)**
Max compatibility, so JSON.

**should we print "null" ?**
Didn't quite get what this refers to, but whatever is most JSONly.

**lists: [1,2,3] or 1,2,3 ? or 1, 2, 3 or maybe 1, 2 & 3!**
JSON style for easy parsing (for pushing further into ELK for example), so [1,2,3]


Question: Will it be possible to use JSON.NET instead of default serializer? Cause serializing .NET objects into JSON can be tricky and opinionated.

304NotModified:
> Default JSON or JSON5? (single quotes, no quotes for keys)
> Max compatibility, so JSON

Well there is no reason to parse it, as you will get the full object (which you can serialize to any format) in the event properties. 

I was thinking in a format which will be comparable to Serilog:
- Simple types: primitive types, lists, dictionaries: user friendly format, no JSON 
- Objects: serialized to JSON5 (single quotes, no quotes on keys) - optional replaced by own/JSON.NET implementation

The reason it is that we, humans, have to parse it. Optional you could always replace it with your own format (scrict JSON, XML etc)

> should we print "null" ?
> Didn't quite get what this refers to, but whatever is most JSONly.

I mean't, printing properties with `null` value. Could make it optional. It was first on, but I have now disabled it. 

> Question: Will it be possible to use JSON.NET instead of default serializer?

Yes definitely! :) But I will not reference JSON.NET by default for compatibility issues. It's already injectable as the "NLog.StructuredEvents" doesn't have JSON awareness  - it's now injected by "NLog"

biiiipy:
>Well there is no reason to parse it, as you will get the full object (which you can serialize to any format) in the event properties.

If it's possible to use serializers to ouput structured logs as json, then that's all fine :) 

>I mean't, printing properties with null value. Could make it optional. It was first on, but I have now disabled it.
>edit: but also printing null in logger.Log("hello {Name}", null);

I would say definitely print null, that's a sane default behaviour. We want to know the value, especially if it's null (because of null-pointer-exceptions).

>Yes definitely! :) But I will not reference JSON.NET by default because of potential compatibility issues.

Perfect!

snakefoot:
@304NotModified Just looked at the parser, and see a lot of overhead in the parsing on the string-template.

Any reason why the handling of structured-events cannot be done when creating the LogEventInfo?  The MessageTemplate could just be an object in the Properties-dictionary (along with all the hole-names). The MessageTemplate can be modified to generate a positional-template, that is set on the LogEventInfo as the official string-format. Custom targets that knows how to handle structured-events can then extract this MessageTemplate if needing the original template and parsed holes.

304NotModified:
Yes that would be possible, but I'm not seeing a large overhead? Could you elaborate on this?

snakefoot:
> Yes that would be possible, but I'm not seeing a large overhead? Could you elaborate on this?

Well string.Fomat is super optimized and uses an internal stringbuilder-pool to optimize allocations.

While the current TemplateParser. Allocates a TemplateParser. Allocates a Template. Allocates a Literal-List. Allocates a Hole-list. Allocates a Literal-Array, Allocates a Hole-List. Allocates a non-reusable String-Builder.

304NotModified:
> While the current TemplateParser. Allocates a TemplateParser. Allocates a Template. Allocates a Literal-List. Allocates a Hole-list. Allocates a Literal-Array, Allocates a Hole-Array. 

true, but aren't we need them anyway? We have to use them in EventLogInfo to render the message and fill the properties

> Allocates a non-reusable String-Builder.

Yes that a good one

snakefoot:
> true, but aren't we need them anyway? We have to use them in EventLogInfo to render the message and fill the properties

The current NLog 4.5 Alpha build performs all these allocation independent of positional-template or structured-template. So if using Log.Info("Hello {0}", "World") then it will also allocate, but then discover that the string-template is positional, and then discard everything.

304NotModified:
> So if using Log.Info("Hello {0}", "World") then it will also allocate, but then discover that the string-template is positional, and then discard everything.

That's indeed an option, but not the current implementation. It still uses or own rendering. We have to benchmark if there is a big performance difference.

But you have a very good point. Maybe it's better to integratie the code more with NLog itself. I preferred to:

- keep the current  commits
- make it easier to test
- keep this separate package also

and therefor this is now installed as package in NLog

snakefoot:
> > So if using Log.Info("Hello {0}", "World") then it will also allocate, but then discover that the string-template is positional, and then discard everything.
>
> That's indeed an option, but not the current implementation. 

Not sure I understand. My description matches the current implementation of NLog 4.5:

```
private void ParseTemplate()
{
   this.messageTemplate = TemplateParser.Parse(Message);
   if (!messageTemplate.IsPositional)
   {
      // Fill into LogEventInfo-properties, else just discard the holes
   }
}
```

My suggestion is not to integrate NLog even further with NLog-structured-event. But instead keep the structured-events as a seperate package of extension-log-methods, that can parse structured-templates (with the extra overhead) and call the core NLog-Log-methods.

304NotModified:
> Not sure I understand. My description matches the current implementation of NLog 4.5:

The template is still used, even if `IsPositional`

```c#
  this.formattedMessage = messageTemplate.Render(this.FormatProvider ?? CultureInfo.CurrentCulture, this.parameters);
```

> But instead keep the structured-events as a seperate package of extension-log-methods, that can parse structured-templates (with the extra overhead) and call the core NLog-Log-methods.

Not sure if I fully understand your proposal, is your proposal to keep the current packages but not create a `Template` when it's positional?

snakefoot:
> Not sure if I fully understand your proposal, is your proposal to keep the current packages but not create a Template when it's positional?

My suggestion is not perform the Template-parsing inside core NLog-LogEventInfo, but in an extension method that fills a LogEventInfo based on the string-template.

304NotModified:
I think refactoring the code is fine. 

Adding pooling would be nice. 

About using string.format, we will add caching and I would like to benchmark first before moving to string.format.

snakefoot:
So your plan is to:

- Check every LogEvent-message whether it is using structured-events using the TemplateParser.
- Always render LogEvent-message using the allocated Template.

If the parsing cannot be done by a method creating the LogEventInfo, and it has to be done inside LogEventInfo. Could the plan be changed into:

- Check every LogEvent-message whether it is using structured-events
- Only allocate the Template-object when an actual structured-event-message (Making it possible to inspect the holes)
- Use the string.Format for rendering when having detected the LogEvent-message as positional (without needing the Template-object for rendering)

I have made some experimental code to the TemplateParser, where I have changed it into an enumerator, so it can iterate the Literal + Holes of the message-template. Then making a recursive function that moves through the enumerator and caches the enumerator-values on the stack:
 - If the message-template is recognized as structured-event-message after reaching the end of the enumerator, then it allocates the correct sized hole-array and literal-array without using lists and copies from the stack to the arrays on unwind.
- If the message-template is recognized as positional, then it just unwinds the stack without allocating anything.

This would allow NLog to check the LogEvent-message-template without allocating anything when positional-template. And minimize allocations when structured-event-template. Will still give a performance-hit.

304NotModified:
Im trying not to make a decision until it has been benchmarked. 

The stuctured logging should be also fast and memory friendly. I prefer to benchmark first, then optimize and at the end maybe fallback to string.format. maybe I'm to optimistic, but I expect performance gain by caching the parse results (maybe not from memory perspective) ;)

String.format has to parse the template continuously. It would be also an option to move from "compile" and then render, to evaluate (like string.format)

304NotModified:
I guess template and template parser (and maybe renderer) should be merged to reduce allocations.

biiiipy:
Completely agree with @304NotModified 
"First, make it work, then make it fast."

snakefoot:
> First, make it work, then make it fast

@biiiipy Can only agree. I'm just curious about the ambition level of the feature implementation. If putting the feature deep into the NLog-core, then I expect the performance to match string.Format.

Gutek:
maybe stupid question, but how can i play with it with current 4.5 alpha release of nLog? what targets to set, what layouts to provide?

Thanks!

304NotModified:
Just install the latest NLog 4.5-*

The same targets, layout renderers etc as NLog 4.4 are supported.

Currently only the platforms .net35/net40/net45 (not .NET Core, thats NLog 5)

Gutek:
Mhh, got that, and not sure how i should then save log entry so it will be structured in json format for instance.

so, how make to make this:
```
_log.Trace("User {Name} is {Age} year's old", "Tom", 34); // just an example
```

to save as json with structured data. Configuring jsonLayout did not did a trick :(

so all i got is:

```
2017-02-28 14:54:49.2398|TRACE|ConsoleApplication3.Program|User Tom is 34 year's old 
```

but would like to get smt like:

```
{
   ...
   Age: 34,
   Name: "Tom",
   ...
}
```

Am I missing something?

304NotModified:
Use json layout renderer in combination with eg ${events-properties:Age} or ${all-event-properties}

Ps on mobile now

304NotModified:
> I'm just curious about the ambition level of the feature implementation. If putting the feature deep into the NLog-core, then I expect the performance to match string.Format.

@snakefoot the ambition is to get a performance matching string.Format. In both structural and non-structural logging methods. It's unclear to me if you think that we should drop the extra NuGet package (I read that with "putting the feature deep into the NLog-core")

304NotModified:
Example for writing to JSON

```xml
                <nlog throwExceptions='true'>
                    <targets>
                        <target name='debug' type='Debug'  >
                                <layout type='JsonLayout' includeAllProperties='true' />
                        </target>
                    </targets>
                    <rules>
                        <logger name='*' levels='Error' writeTo='debug' />
                    </rules>
                </nlog>"

```

304NotModified:
Online! https://www.nuget.org/packages/NLog/4.5.0-alpha04

snakefoot:
Kind of ruins the HasProperties-optimization (of not allocating the Properties-dictionary). Now the HasProperties-check will always allocate the Properties-dictionary :)

304NotModified:
Yeah, well the optimization was leading to a bug ;)

Any idea how to get this working and also performing well?

snakefoot:
> Any idea how to get this working and also performing well?

Think I will wait for the NLog 4.5 branch to become final, before suggesting optimizations :)

304NotModified:
@snakefoot feel free to add some commits to this branch. I have faith that you only make things better :)

304NotModified:
maybe first a squash

304NotModified:
squased

snakefoot:
> Any idea how to get this working and also performing well?

Just like with your suggestion about only calling MergeEventProperties when needed. Then I think the Logger should parse the message-template, and register the properties on the LogEventInfo. Instead of the LogEventInfo constantly checking every string if it could be a structured-message-template.

The LogEventInfo should not expose the MessageTemplate, but should just have a bonus dictionary with the message-template-properties (name + values). The properties-layout-render should then be extended to look in this bonus directionary.

304NotModified:
Thanks @snakefoot!

> Then I think the Logger should parse the message-template, and register the properties on the LogEventInfo. Instead of the LogEventInfo checking every format-string if it could be a structured-message-template.

What's the benefit here? Memory? Not CPU AFAIK.

304NotModified:
Ow

> Then I think the Logger should parse the message-template, and register the properties on the LogEventInfo. Instead of the LogEventInfo checking every format-string if it could be a structured-message-template.

What's parse and check here? I assume parse= string->template  and check=string->bool. Then I don't really understand it as we currently not testing anything but only parsing.

snakefoot:
> What's the benefit here? Memory? Not CPU AFAIK.

You asked for a solution for improving the existing lazy-initialized properties-dictionary. My suggestion is just to perform the structured-event-parsing upfront when we know it has to be done.

304NotModified:
Ah yes thanks!

Parse+render, maybe call that transform or convert?

The benefit of parse the template is to cache it. But maybe we could do that optionally. Not sure if I have to duplicate the parse code then.

snakefoot:
> The benefit of parse the template is to cache it

Maybe the MessageTemplate could be cached inside this magic-object that contains, both the Dictionary of formatted values and list of th parameter-names. The Logger will parse and allocate this magic-object.

304NotModified:
Yes but then its still allocated. Isn't that a problem for you?

Ps. Would be cool to parse it in a Fody plugin for real performance :)

snakefoot:
> Yes but then its still allocated. Isn't that a problem for you?

Well now I have option to skip it, and create my own Logger-extension-methods, that just allocates an LogEventInfo without any parsing or bonus allocations.

304NotModified:
Yeah well it was thinking to include a "useStringFormat" option anyway.

snakefoot:
> Yeah well it was thinking to include a "useStringFormat" option anyway.

Also an option. But I still like my idea of a seperate dictionary for the parameter-names + formatted-values, where the keys-collection are the parameter-names sorted by parameter-position. Instead of exposing the MessageTemplate.

304NotModified:
Yes great idea!

snakefoot:
Maybe the "useStringFormat"-property on the LogEventInfo could be operating the internal MessageTemplate-variable.
- By default the MessageTemplate is null, and the useStringFormat-getter will return true.
- When useStringFormat-setter is assigned to false, then the internal MessageTemplate-object is assigned to a static readonly-version-object (Thus all other properties work, but are empty).
- When useStringFormat-setter is forcily assigned to true, then it is assigned to null.

When requesting the new dictionary-object (that is the MessageTemplate), and the internal-property is null, then it will parse and allocate the MessageTemplate.

304NotModified:
@snakefoot I like to continue on this one, and like to use your recommendations.

I think the first steps would be:

1. extract merging outside logeventinfo
2. make structural logging optional
3. remove `template` inside logeventinfo.

Am I correct? Thanks for the input!

snakefoot:
@304NotModified Yes like that idea, if it is possible :)
- Have the message-template-parsing in the Logger-interface (instead of in LogEventInfo), Then it can just be parameter in the LoggingConfiguration (cached on the Logger) that decides whether it should perform message-template-parsing (default = true).
- Extend the LogEventInfo to have a special structure-event-property-container (replacing or hiding the MessageTemplate property) that can contain the captured parameters (not mixing with the other properties, unless adding a property-type to the internal container-key)
- Maybe just format the mesage upfront in the Logger-interface, and give to the LogEventInfo, so it doesn't have to know how. Else give it a helper-delegate that tells it how to perform the message-format.

304NotModified:
superseded by https://github.com/NLog/NLog/pull/2208

PR #1956, opened 2017-02-12
Improve docs ThreadAgnosticAttribute

304NotModified:
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1956)


304NotModified:
@snakefoot could you double check this? I'm always confused with this attribute and the docs where wrong also ;)

snakefoot:
@304NotModified The updated XML-comment now actually matches the intended logic :)

304NotModified:
Thanks!

I was myself still not happy with the explanation. Updated it again. But should be OK now.

Funny that the remarks first (before this PR) was wrong.

PR #1964, opened 2017-02-14
Fix TimeZoneTest for negative offset

marinsky:
Remove '-' sign from the expected string, as in case of negative
offset '-' sign will already be present.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1964)


marinsky:
Test was not failing in AppVeyor, because it always goes through 'positive offset' path.

304NotModified:
Thanks!

PR #1985, opened 2017-02-23
Pass activity ID into child thread to preserve correlation

kuznero:
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1985)


kuznero:
Working on a proper solution...

Issue #1681, opened 2016-10-05
Archiving removes unexpected files


rymk:
**Bug**
**NLog version**: 4.3.3 & 4.3.8  
**Platform**: .NET 4.5.2

**Scenario:**
We have multiple .NET apps running on the same server. Each app writes to its own log file, but all logs are in the same directory (using naming conventions to distinguish between apps). Similarly, all the logs are archived to a different directory but all apps share the same archive directory. _See the example configs below to see the hierarchy_

**Example configs:**
_**App1:**_

``` xml
<nlog xmlns:nlog="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
    <targets>
      <!-- Log in a separate thread, possibly queueing up to 5000 messages. When the queue overflows, discard any extra messages-->
      <default-wrapper xsi:type="AsyncWrapper" queueLimit="5000" overflowAction="Discard">
      </default-wrapper>
      <target name="logfile" xsi:type="File" fileName="C:\Logs\Project.Web.log"
                archiveFileName="C:\Logs\Archives\${shortdate}-Project.Web.log"
                archiveEvery="Day" maxArchiveFiles="7"
                layout="${longdate} [${level}] [${callsite}] ${message}"
                concurrentWrites="true" keepFileOpen="false" />
    </targets>
    <rules>
      <logger name="*" minLevel="Debug" writeTo="logfile" />
    </rules>
</nlog>
```

_**App2:**_

``` xml
<nlog xmlns:nlog="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
    <targets>
      <!-- Log in a separate thread, possibly queueing up to 5000 messages. When the queue overflows, discard any extra messages-->
      <default-wrapper xsi:type="AsyncWrapper" queueLimit="5000" overflowAction="Discard">
      </default-wrapper>
      <target name="logfile" xsi:type="File" fileName="C:\Logs\Assembly.Web.log"
                archiveFileName="C:\Logs\Archives\${shortdate}-Assembly.Web.log"
                archiveEvery="Day" maxArchiveFiles="7"
                layout="${longdate} [${level}] [${callsite}] ${message}"
                concurrentWrites="true" keepFileOpen="false" />
    </targets>
    <rules>
      <logger name="*" minLevel="Debug" writeTo="logfile" />
    </rules>
</nlog>
```

[...]

Issue #1670, opened 2016-10-03
Archiving removes wrong files (when archive and log files are in the same directory)


StanislavSumyatin:
**Bug**
**NLog version**: 4.3.7
**Platform**: .Net 4.5.2

**Current NLog config** (xml or C#, if relevant)

``` xml
<target name="file" type="File"
                        layout="${longdate} | ${level:uppercase=true} }"
                        fileName="${LogFolder}/${LogSource}.log"
                        archiveFileName="${LogFolder}/${LogSource}-${date:format=yyyy-MM-dd}#{##}.log"
        archiveEvery="Hour"
                concurrentWrites="true"
        archiveNumbering="Rolling"
        maxArchiveFiles="84"/>

```

**Current result**: Not only "nlog files" are deleted from LogFolder, but also web.config and logs from other systems when method `DeleteOldArchiveFiles` is called

**Expected result**: Only nlog files are deleting when method DeleteOldArchiveFiles is called. 
Problem is that in method DynamicFileArchive.InitializeForArchiveFolderPath all files from directory LogFolder are queued and then deleting in method DeleteOldArchiveFiles.

Internal log:
2016-09-30 14:26:40.3940 Debug Fatal => file
2016-09-30 14:32:41.5089 Warn Cannot delete old archive file : 'C:\logs\SiteLogs\Test8\AvayaToolkit.log'. Exception: System.IO.IOException: The process cannot access the file 'C:\logs\SiteLogs\Test8\AvayaToolkit.log' because it is being used by another process.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.File.InternalDelete(String path, Boolean checkHost)
   at NLog.Targets.FileTarget.DynamicFileArchive.DeleteOldArchiveFiles()
2016-09-30 14:32:41.5089 Warn Cannot delete old archive file : 'C:\logs\SiteLogs\Test8\EpiServer_Test8.log'. Exception: System.IO.IOException: The process cannot access the file 'C:\logs\SiteLogs\Test8\EpiServer_Test8.log' because it is being used by another process.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.File.InternalDelete(String path, Boolean checkHost)
   at NLog.Targets.FileTarget.DynamicFileArchive.DeleteOldArchiveFiles()

It deletes files AvayaToolkit.log and web.config from diretory.

304NotModified:
[...]

Issue #1748, opened 2016-11-07
Unexpected Archiving Behavior when ArchiveEvery="Day" and MaxArchiveFiles="1"


meadjm:
NLog Unexpected Rolling\Archiving Behavior when ArchiveEvery="Day" and MaxArchiveFiles="1"

With our current NLog configuration, I would expect to have a log `<dayNumber>_AppName.log` (i.e. 07_AppName.log) for every day in the month. When the day log from the previous month exists, I would expect it to roll to the archive 00_AppName.0.log, truncate the existing day log for the new day (at or about 12:00:01 AM) and start logging again. The problem we are intermittently seeing is that the previous days log rolls to the archive file and its day log is either truncated or deleted. The current days day log is truncated as expected (but not archived). This effectively is giving us 2 logs for the entire month (if the files are deleted). The current day and previous day as the archive file or we get empty day logs if the files are truncated.  We have several different applications using this configuration. The logging seems to work as expected for several days and then starts exhibiting this odd behavior. Another application we have consistently starts with this behavior immediately.

I recently updated the application from NLog 4.3.4 to 4.3.10 as it looks an archiving issue was resolved 

Note: We really do not care about the archive file, but this was a work around for issue #1479. The archiving configuration forces the previous months log to be truncated. We are waiting for the fix that supports MaxArchiveFiles=0 to mean no archive files.

Any ideas what could be causing this? 

**Type** Bug

**NLog version**: 4.3.10

**Platform**: .Net 4.0  

**Current NLog config**:

```xml
<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" autoReload="false">
    <variable name="logDirectory" value="D:\Logs"/>
    <variable name="logLayout" value="${longdate} | ${threadid:padding=-2} | ${level:padding=-5} ${callsite} | ${message} | ${onexception:EXCEPTION OCCURRED\:${Exception:format=type,message,method,stacktrace:maxInnerExceptionLevel=5}}"/>
    <targets>
        <default-target-parameters xsi:type="File" keepFileOpen="true" OpenFileCacheTimeout="5" ArchiveEvery="Day" ArchiveNumbering="Rolling" MaxArchiveFiles="1" />
        <target name="AppLog" xsi:type="File" fileName="${logDirectory}\${date:format=dd}_AppName.log" layout ="${logLayout}" />
    </targets>
    <rules>
[...]

PR #1993, opened 2017-02-27
FileArchiveModeFactory - Extracted from FileTarget

snakefoot:
Attempt to resolve the issues #1670 and #1681, which is mostly caused not understanding how to configure the NLog Archive Logic (And also major refactoring of the archive logic in FileTarget).

I guess more effort could be applied to FileArchiveModeDynamicSequence.GenerateFileNameMask() to better recognize any date-format-output.

Enabled the unit-tests from #1788, but had to modify them, so they would work with Mono on Linux (Changed hardcoded windows-backslash to Path.Combine). The tests also suffered from confusing debug- and trace-level (Logged to debug-level and didn't expect it to reach trace-level).

**!! Breaking change !!**
If using ArchiveNumberingMode.Sequence (default) and not having configured an ArchiveFileName (default) and have enabled archiving (MaxArchiveFiles > 0). Then it will now enable dynamic-sequence mode, instead of strict-sequence-mode.

Dynamic-sequence-mode will use the same archive-filename-format (as strict-sequence-mode), but its filemask-wildcard for doing archive-cleanup is a little more greedy as it replaces the largest range of digits (must be more than one) with a wildcard (and not just the sequence-no).

This improves the out-of-box archive-cleanup, when the FileTarget.FileName-Layout includes date-renderer.

**New behavior for dynamice sequence-mode**
New behavior when ArchiveFileName does not contain sequence-no-pattern `(#}`:

- More adaptive filemask-wildcard used when doing archive-cleanup. Before it was just all files with the same file-extension. Now also using the actual LogEvent (Instead of Null-LogEvent), so the filemask for cleanup is a lot better (Doesn't need an isolated archive directory).
- Now delays archive cleanup until first write-operation occurs, instead of FileTarget.Initialize(). This allows archive-cleanup to work properly as it will use the actual LogEvent for rendering archive filemask for cleanup (Instead of Null-LogEvent).
  - This also removes extra overhead at startup where all FileTarget would scan their archive-folders.
  - This generates extra overhead at first new file write, where archive-cleanup occurs. But the archive-cleanup is skipped if new file write is caused by just having executed a file-archive-operation.

This improves the out-of-box archive-cleanup, when the ArchiveFileName-Layout includes date-renderer.

**Other changes:**
- Now delays the monitoring of the archive-folder until the first-write. And the filemask monitored is generated from the actual logevent instead of Null-LogEvent (More correct).
- ArchiveNumberingMode.Date now skips an unnecessary archive-cleanup when moving to new file after just having performed archive of the previous file (with cleanup).
- Attempt to handle that logging-folder and archive-folder is the same folder. Crude Path.GetDirectoryName-comparison that probably can be cheated.
- Attempt to handle file-creation-date-issue on Linux for archiving. This should improve how NLog performs archive-cleanup of dated-archives.
- Attempt to handle non-standard-path-usage, that is different from those returned by the standard filesystem.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1993)


304NotModified:
It's the limited silverlight api. 

> Targets\FileArchiveStyleFactory.cs(583,81): error CS1061: 'System.IO.FileInfo' does not contain a definition for 'CreationTimeUtc' and no extension method 'CreationTimeUtc' accepting a first argument of type 'System.IO.FileInfo' could be found (are you missing a using directive or an assembly reference?) 

There's maybe a helper for this already.

snakefoot:
@304NotModified Thanks that fixed the appveyor-build. The travis-build required some changes to the unit-tests. But now things are looking green. And I have implemented a fix for my worries about performing too many archive cleanups.

src/NLog/Internal/FileInfoExt.cs
```diff
@@ -43,19 +43,16 @@ public static DateTime GetLastWriteTimeUtc(this FileInfo fileInfo)
 #if !SILVERLIGHT
             return fileInfo.LastWriteTimeUtc;
 #else
-            return fileInfo.LastWriteTime;
+            return fileInfo.LastWriteTime.ToUniversalTime();
```
304NotModified: isn't this an expensive method? (dunno how many it will be called)

src/NLog/Targets/FileArchiveStyleFactory.cs
```diff
@@ -0,0 +1,758 @@
...
+        /// <summary>
+        /// Characters determining the end of the <see cref="P:FileNameTemplate.Pattern"/>.
+        /// </summary>
+        public const string PatternEndCharacters = "#}";
+
+        /// <summary>
+        /// File name which is used as template for matching and replacements. 
+        /// It is expected to contain a pattern to match.
+        /// </summary>
+        public string Template { get { return this.template; } }
```
304NotModified: could we move to auto properties? (less code is more)

src/NLog/Targets/FileArchiveStyleFactory.cs
```diff
@@ -0,0 +1,758 @@
...
+        {
+            if (_lastArchiveFilePath == archiveFilePath)
+            {
+                // Skip archive cleanup, when it has just been executed
+                _lastArchiveFilePath = string.Empty;
+                return false;
+            }
+            else
+            {
+                _lastArchiveFilePath = string.Empty;
```
304NotModified: please move before the `if` (as both branches have ` _lastArchiveFilePath = string.Empty;` - or is this a  potential bug?)

src/NLog/Targets/FileArchiveStyleFactory.cs
```diff
@@ -0,0 +1,758 @@
...
+#endif
+            foreach (var fileInfo in existingFiles)
+            {
+                var archiveFileInfo = GenerateArchiveFileInfo(fileInfo, archiveFileNameTemplate);
+                if (archiveFileInfo != null)
+                    existingArchiveFiles.Add(archiveFileInfo);
+            }
+
+            if (existingArchiveFiles.Count > 1)
+                existingArchiveFiles.Sort((x, y) => (x.Date != y.Date && !x.HasSameFormattedDate(y.Date)) ? x.Date.CompareTo(y.Date) : x.Sequence.CompareTo(y.Sequence) != 0 ? x.Sequence.CompareTo(y.Sequence) : string.CompareOrdinal(x.FileName, y.FileName));
```
304NotModified: any idea if we could get this a bit more readable (a helper maybe?)

src/NLog/Targets/FileArchiveStyleFactory.cs
```diff
@@ -0,0 +1,758 @@
...
+        {
+            if (maxArchiveFiles <= 0)
+                yield break;
+
+            _lastArchiveFilePath = archiveFilePath; // Cache that we have just performed cleanup for this archive-path
+
+            if (existingArchiveFiles.Count == 0 || existingArchiveFiles.Count < maxArchiveFiles)
+                yield break;
+
+            for (int i = 0; i < existingArchiveFiles.Count - maxArchiveFiles; ++i)
```
304NotModified: `i++` instead of `++i` ?

I'm curious, what's the benefit of `++i` in a for loop?

edit: it was already there I think

src/NLog/Targets/FileTarget.cs
```diff
@@ -1990,21 +1539,12 @@ private string GetArchiveFileNamePattern(string fileName, LogEventInfo eventInfo
...
-        }
-
-        /// <summary>
         /// Archives the file if it should be archived.
         /// </summary>
         /// <param name="fileName">The file name to check for.</param>
         /// <param name="ev">Log event that the <see cref="FileTarget"/> instance is currently processing.</param>
         /// <param name="upcomingWriteSize">The size in bytes of the next chunk of data to be written in the file.</param>
-        private void TryArchiveFile(string fileName, LogEventInfo ev, int upcomingWriteSize)
+        private bool TryArchiveFile(string fileName, LogEventInfo ev, int upcomingWriteSize)
```
304NotModified: please document the return value.

tests/NLog.UnitTests/Targets/FileTargetTests.cs
```diff
@@ -2770,8 +2837,8 @@ public void MaxArchiveFilesWithDate_in_sameDir(bool changeCreationAndWriteTime)
                     var time = now.AddDays(i);
                     if (changeCreationAndWriteTime)
                     {
-                        File.SetCreationTime(filePath, time);
-                        File.SetLastWriteTime(filePath, time);
+                        File.SetCreationTimeUtc(filePath, time);
```
304NotModified: is this really needed? NLog should also handle local times

src/NLog/Targets/FileArchiveStyleFactory.cs
snakefoot: I just moved the class directly from FileTarget, but I can clean it up.

src/NLog/Internal/FileInfoExt.cs
snakefoot: Well it only happens for silverlight-applications, and since the method should return Utc, then it would make sense to do it correct. But I will remove it, since it is not important to me.

tests/NLog.UnitTests/Targets/FileTargetTests.cs
snakefoot: Leftovers from when I was debugging the failing unit-tests on Mono/Travis. Will undo the changes.

src/NLog/Targets/FileArchiveStyleFactory.cs
304NotModified: Would be great if you would do that as I cannot see what's new or what's moved (easily). Thx in advance!

src/NLog/Internal/FileInfoExt.cs
304NotModified: ow yes, missed the `!`.

I think this is then OK, iif the tests could work without UTC stuff

tests/NLog.UnitTests/Targets/FileTargetTests.cs
304NotModified: 👍

src/NLog/Targets/FileArchiveStyleFactory.cs
snakefoot: Removed the if-statement.

src/NLog/Targets/FileArchiveStyleFactory.cs
snakefoot: Now calling static-method

src/NLog/Targets/FileArchiveStyleFactory.cs
snakefoot: Changed to i++

src/NLog/Targets/FileTarget.cs
snakefoot: Updated XML-docs

snakefoot:
> Please move all the classes to their own file :)

Have now moved the classes into their own individual file.

snakefoot:
@304NotModified Discovered that the "bugs" reported for the NLog-Archive-Logic is actually caused by the problems with understanding how to use the enum ArchiveNumberingMode together with ArchiveFileName-layout (Failing to use the **{#}** properly).

I have now modified this PR into improving the out-of-box experience, when using the NLog-Archive-Logic. And removed the bad behavior of doing cleanup of all files with same file-extension, when failing to use the **{#}** properly.

304NotModified:
Sounds great! Will include into 4.5 due to the large redactor. Is it OK if I merge #2000  first?

snakefoot:
> Is it OK if I merge #2000 first?

Sure no problem with that.

304NotModified:
@snakefoot 

> If using ArchiveNumberingMode.Sequence (default) and not having configured an ArchiveFileName (default) and have enabled archiving (MaxArchiveFiles > 0). Then it will now enable dynamic-sequence mode, instead of strict-sequence-mode.

> Dynamic-sequence-mode will use the same archive-filename-format (as strict-sequence-mode), but its filemask-wildcard for doing archive-cleanup is a little more greedy as it replaces the largest range of digits (must be more than one) with a wildcard (and not just the sequence-no).

> This improves the out-of-box archive-cleanup, when the FileTarget.FileName-Layout includes date-renderer.

Do you have an example for this? (e.g. what is changed). Need that for the news post :)

snakefoot:
> Do you have an example for this?

I guess the most simple example is that you can now get correct archive cleanup when just using the two FileTarget parameters `FileName` and `MaxArchiveFiles`, eventhough `FileName` depends on layout-renders (Like date, logger-name, log-level, etc.). Before it would try and cleanup according to rendering of Null-LogEvent.

The more advanced example is when using `archiveFileName` without `{0}`, then one will also get correct archive cleanup. Before it would try and cleanup anything matching the log-file-extension.

304NotModified:
fixes #2050

304NotModified:
@snakefoot any idea if this also fixes #1748?

snakefoot:
@304NotModified No don't think so, just scanned the code to see if there was any obvious bugs, but didn't find any.

snakefoot:
> Any idea if this also fixes #1748?

@304NotModified Just tested with the provided nlog.config. It rolls every day (as configured), so the result becomes:

> 01_AppName.0.log (Two days ago)
> 02_AppName.0.log (Previous day)
> 03_AppName.log (Active)

I think the wanted behavior is this:

> 01_AppName.log (Two days ago)
> 02_AppName.log (Previous day)
> 03_AppName.log (Active)

This PR can give this result by removing **ArchiveEvery="Day"** and **ArchiveNumbering="Rolling"**. And then change the existing parameter to **MaxArchiveFiles="27"** (Cater for february).

snakefoot:
@304NotModified Have now implemented a unit test that verifies that the above solution actually works for  #1748

304NotModified:
fixes  #1670, fixes #1681, fixes https://github.com/NLog/NLog/issues/1748

PR #2000, opened 2017-03-02
Add weekly archival option to FileTarget

dougthor42:
This PR adds a `Weekly` archiving option to FileTarget.

The reasoning behind this change is that, for some projects, Daily archiving is too often and Monthly archiving is too infrequent.

### Discussion
Typically weekly archival options, such as the [Python TimedRotatingFileHandler](https://docs.python.org/3/library/logging.handlers.html#logging.handlers.TimedRotatingFileHandler), will require the user to specify the day of the week that archiving should occur. To implement this, we have two options:

1. Add a configuration option to the File target. Something like:
    ```xml
    <target xsi:type="File"
            name="file"
            layout="${defaultLayout}"
            fileName="${logFileDir}/my_program.log"
            archiveFileName="${logFileDir}/my_program_{#}.log"
            archiveEvery="Week"
            archiveOn="Sunday"            <-- a new archival option
            archiveNumbering="Date"
            archiveDateFormat="yyyyMMdd"
            maxArchiveFiles="14"
      />
    ```
2. Simply use the day of the week as the `archiveEvery` value:
    ```xml
    <target xsi:type="File"
            name="file"
            layout="${defaultLayout}"
            fileName="${logFileDir}/my_program.log"
            archiveFileName="${logFileDir}/my_program_{#}.log"
            archiveEvery="Tuesday"
            archiveNumbering="Date"
            archiveDateFormat="yyyyMMdd"
            maxArchiveFiles="14"
      />
    ```

Personally I feel that option 2 is easier for the user and also easier to implement, so that's what this PR will be focusing on. If you'd prefer option 1 or have another suggestion, please let me know.

### DIFF
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -1882,8 +1882,11 @@ namespace NLog.Targets
 
         private bool PreviousLogOverlappedPeriod(LogEventInfo logEvent, DateTime lastWrite)
         {
-            if (!previousLogEventTimestamp.HasValue)
+            DateTime timestamp;
+            if(!previousLogEventTimestamp.HasValue)
                 return false;
+            else
+                timestamp = previousLogEventTimestamp.Value;
 
             string formatString = GetArchiveDateFormatString(string.Empty);
             string lastWriteTimeString = lastWrite.ToString(formatString, CultureInfo.InvariantCulture);
@@ -1895,18 +1898,18 @@ namespace NLog.Targets
             DateTime periodAfterPreviousLogEventTime;
             switch (this.ArchiveEvery)
             {
-                case FileArchivePeriod.Year: periodAfterPreviousLogEventTime = previousLogEventTimestamp.Value.AddYears(1); break;
-                case FileArchivePeriod.Month: periodAfterPreviousLogEventTime = previousLogEventTimestamp.Value.AddMonths(1); break;
-                case FileArchivePeriod.Day: periodAfterPreviousLogEventTime = previousLogEventTimestamp.Value.AddDays(1); break;
-                case FileArchivePeriod.Hour: periodAfterPreviousLogEventTime = previousLogEventTimestamp.Value.AddHours(1); break;
-                case FileArchivePeriod.Minute: periodAfterPreviousLogEventTime = previousLogEventTimestamp.Value.AddMinutes(1); break;
-                case FileArchivePeriod.Sunday: periodAfterPreviousLogEventTime = CalculateNextWeekday(previousLogEventTimestamp, DayOfWeek.Sunday); break;
-                case FileArchivePeriod.Monday: periodAfterPreviousLogEventTime = CalculateNextWeekday(previousLogEventTimestamp, DayOfWeek.Monday); break;
-                case FileArchivePeriod.Tuesday: periodAfterPreviousLogEventTime = CalculateNextWeekday(previousLogEventTimestamp, DayOfWeek.Tuesday); break;
-                case FileArchivePeriod.Wednesday: periodAfterPreviousLogEventTime = CalculateNextWeekday(previousLogEventTimestamp, DayOfWeek.Wednesday); break;
-                case FileArchivePeriod.Thursday: periodAfterPreviousLogEventTime = CalculateNextWeekday(previousLogEventTimestamp, DayOfWeek.Thursday); break;
-                case FileArchivePeriod.Friday: periodAfterPreviousLogEventTime = CalculateNextWeekday(previousLogEventTimestamp, DayOfWeek.Friday); break;
-                case FileArchivePeriod.Saturday: periodAfterPreviousLogEventTime = CalculateNextWeekday(previousLogEventTimestamp, DayOfWeek.Saturday); break;
+                case FileArchivePeriod.Year: periodAfterPreviousLogEventTime = timestamp.AddYears(1); break;
+                case FileArchivePeriod.Month: periodAfterPreviousLogEventTime = timestamp.AddMonths(1); break;
+                case FileArchivePeriod.Day: periodAfterPreviousLogEventTime = timestamp.AddDays(1); break;
+                case FileArchivePeriod.Hour: periodAfterPreviousLogEventTime = timestamp.AddHours(1); break;
+                case FileArchivePeriod.Minute: periodAfterPreviousLogEventTime = timestamp.AddMinutes(1); break;
+                case FileArchivePeriod.Sunday: periodAfterPreviousLogEventTime = CalculateNextWeekday(timestamp, DayOfWeek.Sunday); break;
+                case FileArchivePeriod.Monday: periodAfterPreviousLogEventTime = CalculateNextWeekday(timestamp, DayOfWeek.Monday); break;
+                case FileArchivePeriod.Tuesday: periodAfterPreviousLogEventTime = CalculateNextWeekday(timestamp, DayOfWeek.Tuesday); break;
+                case FileArchivePeriod.Wednesday: periodAfterPreviousLogEventTime = CalculateNextWeekday(timestamp, DayOfWeek.Wednesday); break;
+                case FileArchivePeriod.Thursday: periodAfterPreviousLogEventTime = CalculateNextWeekday(timestamp, DayOfWeek.Thursday); break;
+                case FileArchivePeriod.Friday: periodAfterPreviousLogEventTime = CalculateNextWeekday(timestamp, DayOfWeek.Friday); break;
+                case FileArchivePeriod.Saturday: periodAfterPreviousLogEventTime = CalculateNextWeekday(timestamp, DayOfWeek.Saturday); break;
                 default: return false;
             }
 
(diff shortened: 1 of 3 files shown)


304NotModified:
Cool!

But its gonna conflict with https://github.com/NLog/NLog/pull/1993 I guess

dougthor42:
@304NotModified Oh I didn't see that one. How would you like me to handle that? Merge into that branch? Just continue with this PR into master?

dougthor42:
A couple discussion points:

1) I'm not familiar with the `xunit` framework - I typically just use the MSTest framework because my projects are small.
I can add tests for the `CalculateNextWeekday` function, but I'm pretty lost when it comes to actually testing the rollover and whatnot. Any tips?

2) It feels a bit silly essentially recreating the `DayOfWeek` enum. Do you have any suggestions for this?

3) I'll be squashing into reasonable commits (likely just 1 or 2) before removing WIP.

snakefoot:
@304NotModified Right now the changes looks quite minimal, so it doesn't have to wait for #1993 

@dougthor42 Remember also to add your new enums to TruncateArchiveTime() and PreviousLogOverlappedPeriod()

src/NLog/Targets/FileTarget.cs
```diff
@@ -1908,6 +1915,24 @@ private bool PreviousLogOverlappedPeriod(LogEventInfo logEvent, DateTime lastWri
...
+        /// </summary>
+        /// <param name="previousLogEventTimestamp">The DateTime of the previous log event.</param>
+        /// <param name="dayOfWeek">The next occuring day of the week to return a DateTime for.</param>
+        /// <returns>The DateTime of the next occuring dayOfWeek.</returns>
+        /// <remarks>For example: if previousLogEventTimestamp is Thursday 2017-03-02 and dayOfWeek is Sunday, this will return
+        ///  Sunday 2017-03-05. If dayOfWeek is Thursday, this will return *next* Thursday 2017-03-09.</remarks>
+        public DateTime CalculateNextWeekday(DateTime? previousLogEventTimestamp, DayOfWeek dayOfWeek)
+        {
+            // Shamelessly taken from http://stackoverflow.com/a/7611480/1354930
+            int start = (int)previousLogEventTimestamp.Value.DayOfWeek;
```
304NotModified: please handle the  possible nullrefexception here

dougthor42: This function is only called by `PreviouisLogOverlappedPeriod()`. At the begining of that function, a check for nullref with `if(!previousLogEventTimestamp.HasValue) return false;` already exists.

How should I adding nullref handling to `CalcualteNextWeekday`?
1. assume `previousLogEventTimestamp = DateTime.Now` and return the next `dayOfWeek`?
2. simply return `DateTime.Now`?
3. throw the error up the chain?
4. alternatively, I can edit `PreviousLogOverlappedPeriod()` to convert the nullable `previousLogEventTimestamp` to a non-nullable type and then not allow nulls in `CalculateNextWeekday`

I don't know enough about the codebase to say if 1 or 2 would cause other issues (such as never archiving), so I'm leaning towards option 4. Thoughts?

dougthor42:
I wasn't sure how I should handle the possible NullRefEx, so I decided to make it a non-issue. See the conversation on the outdated diff and let me know if you want a different solution.

304NotModified:
thanks! Will try to check in-depth this weekend.

src/NLog/Targets/FileTarget.cs
```diff
@@ -1882,8 +1882,11 @@ private DateTime GetArchiveDate(string fileName, LogEventInfo logEvent)
 
         private bool PreviousLogOverlappedPeriod(LogEventInfo logEvent, DateTime lastWrite)
         {
-            if (!previousLogEventTimestamp.HasValue)
+            DateTime timestamp;
+            if(!previousLogEventTimestamp.HasValue)
                 return false;
+            else
+                timestamp = (DateTime)previousLogEventTimestamp;
```
304NotModified: this is `previousLogEventTimestamp.Value`? (cast could be runtime unsafe)

tests/NLog.UnitTests/Targets/FileTargetTests.cs
```diff
@@ -3269,6 +3269,36 @@ public void TestFilenameCleanup()
...
+        public void TestCalculateNextWeekday()
+        {
+            DateTime today = new DateTime(2017, 03, 02, 15, 27, 34, 651);
+            DateTime expected = new DateTime(2017, 03, 05, 15, 27, 34, 651);
+            FileTarget fileTarget = new FileTarget();
+            DateTime actual = fileTarget.CalculateNextWeekday(today, DayOfWeek.Sunday);
+            Assert.Equal(expected, actual);
+        }
+
+        [Fact]
```
304NotModified: would be nice to combine the unit tests to one test, with `[InlineData]` , see http://www.codinginstinct.com/2008/10/funky-xunit-theory.html

src/NLog/Targets/FileTarget.cs
```diff
@@ -1908,6 +1918,24 @@ private bool PreviousLogOverlappedPeriod(LogEventInfo logEvent, DateTime lastWri
...
 
         /// <summary>
+        /// Calculate the DateTime of the requested day of the week.
+        /// </summary>
+        /// <param name="previousLogEventTimestamp">The DateTime of the previous log event.</param>
+        /// <param name="dayOfWeek">The next occuring day of the week to return a DateTime for.</param>
+        /// <returns>The DateTime of the next occuring dayOfWeek.</returns>
+        /// <remarks>For example: if previousLogEventTimestamp is Thursday 2017-03-02 and dayOfWeek is Sunday, this will return
+        ///  Sunday 2017-03-05. If dayOfWeek is Thursday, this will return *next* Thursday 2017-03-09.</remarks>
+        public DateTime CalculateNextWeekday(DateTime previousLogEventTimestamp, DayOfWeek dayOfWeek)
```
304NotModified: should be static

src/NLog/Targets/FileTarget.cs
304NotModified: > How should I adding nullref handling to CalcualteNextWeekday?

Change the parameter from `DateTime?` to `DateTime`. Then it's fixed in this case :)

(I see indeed it's already checked in the calls)

tests/NLog.UnitTests/Targets/FileTargetTests.cs
304NotModified: one of the great benefits of xUnit / NUnit over MStest ;)

tests/NLog.UnitTests/Targets/FileTargetTests.cs
```diff
@@ -3269,6 +3269,17 @@ public void TestFilenameCleanup()
...
         }
 
+        [Theory]
+        [InlineData(DayOfWeek.Sunday, "2017-03-02 15:27:34.651", "2017-03-05 15:27:34.651")]    // On a Thursday, finding next Sunday
+        [InlineData(DayOfWeek.Thursday, "2017-03-02 15:27:34.651", "2017-03-09 15:27:34:651")]  // On a Thursday, finding next Thursday
+        [InlineData(DayOfWeek.Monday, "2017-03-02 00:00:00.000", "2017-03-06 00:00:00.000")]    // On Thursday at Midnight, finding next Monday
+        public void TestCalculateNextWeekday(DayOfWeek day, string todayString, string expectedString)
+        {
+            DateTime today = DateTime.Parse(todayString);
+            DateTime expected = DateTime.Parse(expectedString);
```
dougthor42: I wish it was easier to put complex types into `InlineData`. Having to parse these strings feels kinda kludgy.

dougthor42:
I think I got everything. Let me know if there's more or if I should squash and remove WIP.

Also, how about documentation? Is that part of the source or is it just on the Wiki? (I'm primarily a Python dev and we have our reStructuredText docs in the source.)

tests/NLog.UnitTests/Targets/FileTargetTests.cs
304NotModified: Yes there are also other ways, but for now its sufficient and easier to maintain :)

304NotModified:
> Also, how about documentation? Is that part of the source or is it just on the Wiki?

Just the wiki. If you could add some docs and  "introduced in NLog 4.4.4", that would be great!

304NotModified:
great will merge it when the docs are ready :)

dougthor42:
I modified the wiki for [FileTarget](https://github.com/NLog/NLog/wiki/File-target). Let me know if there's anything else you'd like me to do. Otherwise, I'm all done!

304NotModified:
an example on https://github.com/NLog/NLog/wiki/File-target (bottom) would be great!

304NotModified:
close/open to trigger CI

dougthor42:
Example added to the wiki

👍

304NotModified:
Merged!

304NotModified:
Thx!

304NotModified:
NLog 4.4.4 is live! https://www.nuget.org/packages/NLog/4.4.4

PR #2006, opened 2017-03-06
Added AsyncTaskTarget - Base class for using async methods

### LINKED ISSUE #2004: Custom NLog target with async writing
**Type**: Question
**NLog version**: 5.0.0-beta05
**Platform**: .NET Core 

I want to write a [custom target](http://github.com/NLog/NLog/wiki/How%20to%20write%20a%20custom%20target) that logs to my database using Entity Framework Core.

In `NLog.Targets.Target` there is this:

    protected virtual void Write(LogEventInfo logEvent);

However my code is async, so I must do:

    protected override /*async Task*/void Write(LogEventInfo logEvent)   // problem is here
    {
        await WriteToEfContext(logEvent);               // problem is here
        // and so on...
    }

But there is no `Task WriteAsync` version of that method.

How do I write a custom target, with async support?

snakefoot:
AsyncTaskTarget tries to resolve #2004. It has the following properties:

   - Only uses a single Task to protect against out-of-order writing of LogEvents (And avoids using the entire thread pool).
   - WriteAsyncTask is called without holding Target-Lock-object to improve concurrency (loggers can schedule more LogEvents while the initial LogEvent call is being started).
   - Having a requestQueue to make it easier to monitor pending requests (instead of navigating in a ContinueWith-chain).
   - Support flush where it correctly notifies when pending write of LogEvents have completed.
   - Support timeout handling, cancels active task if it fails to complete (and continues with the next)

Following features could probably be added:

   - Support batching of multiple LogEvents in a single write operation.
   - Support parallel writing of LogEvents using a limited number of concurrent Tasks.

Example of how to make custom target:
```
        public class AsyncTaskTestTarget : AsyncTaskTarget
        {

            protected override Task WriteAsyncTask(LogEventInfo logEvent, CancellationToken token)
            {
                return CustomWriteAsync(logEvent, token);
            }

            private async Task CustomWriteAsync(LogEventInfo logEvent, CancellationToken token)
            {
                await Task.Delay(10, token).ConfigureAwait(false);   // TODO Do something with logEvent
            }
        }
```

### DIFF
--- a/src/NLog/Targets/AsyncTaskTarget.cs
+++ b/src/NLog/Targets/AsyncTaskTarget.cs
@@ -191,7 +191,12 @@ namespace NLog.Targets
                     lock (this.SyncRoot)
                     {
                         _previousTask = newTask;
+#if (SILVERLIGHT && !WINDOWS_PHONE) || NET4_0
+                        var continuation = logEvent.Continuation;
+                        _previousTask.ContinueWith(completedTask => TaskCompletion(completedTask, continuation), _cancelTokenSource.Token);
+#else
                         _previousTask.ContinueWith(_taskCompletion, logEvent.Continuation, _cancelTokenSource.Token);
+#endif
                         if (_previousTask.Status == TaskStatus.Created)
                             _previousTask.Start(TaskScheduler.Default);
                     }
@@ -237,4 +242,4 @@ namespace NLog.Targets
         }
     }
 #endif
-}
+                    }
--- a/src/NLog/Targets/AsyncTaskTarget.cs
+++ b/src/NLog/Targets/AsyncTaskTarget.cs
@@ -40,26 +40,45 @@ namespace NLog.Targets
 #if !NET3_5 && !SILVERLIGHT4
     using System.Threading.Tasks;
     using NLog.Common;
+    using System.ComponentModel;
 
     /// <summary>
     /// Abstract Target with async Task support
     /// </summary>
     public abstract class AsyncTaskTarget : Target
     {
-        private readonly CancellationTokenSource _cancelTokenSource;
+        private readonly Timer _taskTimeoutTimer;
+        private CancellationTokenSource _cancelTokenSource;
         private readonly Queue<AsyncLogEventInfo> _requestQueue;
         private readonly Action _taskStartNext;
+        private readonly Action _taskCancelledToken;
         private readonly Action<Task, object> _taskCompletion;
         private Task _previousTask;
+
+        /// <summary>
+        /// How many seconds a Task is allowed to run before it is cancelled.
+        /// </summary>
+        [DefaultValue(150)]
+        public int TaskTimeoutSeconds { get; set; }
+
+        /// <summary>
+        /// Task Scheduler used for processing async Tasks
+        /// </summary>
+        protected virtual TaskScheduler TaskScheduler { get { return TaskScheduler.Default; } }
 
         /// <summary>
         /// Constructor
         /// </summary>
         protected AsyncTaskTarget()
         {
-            _taskStartNext = TaskStartNext;
+            TaskTimeoutSeconds = 150;
+
+            _taskStartNext = () => TaskStartNext(null);
             _taskCompletion = TaskCompletion;
+            _taskCancelledToken = TaskCancelledToken;
+            _taskTimeoutTimer = new Timer(TaskTimeout);
             _cancelTokenSource = new CancellationTokenSource();
+            _cancelTokenSource.Token.Register(_taskCancelledToken);
             _requestQueue = new Queue<AsyncLogEventInfo>(10000);
         }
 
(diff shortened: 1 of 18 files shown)


src/NLog/Targets/AsyncTaskTarget.cs
```diff
@@ -0,0 +1,358 @@
...
+                {
+                    lock (this.SyncRoot)
+                    {
+                        _previousTask = newTask;
+
+                        if (TaskTimeoutSeconds > 0)
+                            _taskTimeoutTimer.Change(TaskTimeoutSeconds * 1000, Timeout.Infinite);
+
+                        // NOTE - Not using _cancelTokenSource for ContinueWith, or else they will also be cancelled on timeout
+#if (SILVERLIGHT && !WINDOWS_PHONE) || NET4_0
```
304NotModified: why special code for net40 but not for net35 or net45?

src/NLog/Targets/AsyncTaskTarget.cs
```diff
@@ -0,0 +1,358 @@
...
+                    if (completedTask.Exception != null)
+                        InternalLogger.Warn(completedTask.Exception, "{0} WriteAsyncTask was cancelled", this.Name);
+                    else
+                        InternalLogger.Info("{0} WriteAsyncTask was cancelled", this.Name);
+                }
+                else if (completedTask.Exception != null)
+                {
+                    InternalLogger.Warn(completedTask.Exception, "{0} WriteAsyncTask failed on completion", this.Name);
+                }
+                ((AsyncContinuation)continuation)(completedTask.Exception);
```
304NotModified: I prefer a bit more verbose syntax here :)

src/NLog/Targets/AsyncTaskTarget.cs
```diff
@@ -0,0 +1,358 @@
...
+        {
+            TaskTimeoutSeconds = 150;
+
+            _taskStartNext = () => TaskStartNext(null);
+            _taskCompletion = TaskCompletion;
+            _taskCancelledToken = TaskCancelledToken;
+            _taskTimeoutTimer = new Timer(TaskTimeout);
+            _cancelTokenSource = new CancellationTokenSource();
+            _cancelTokenSource.Token.Register(_taskCancelledToken);
+            _requestQueue = new Queue<AsyncLogEventInfo>(10000);
```
snakefoot: I guess some day there will be a MaxQueueLength, but not planning one now.

src/NLog/Targets/AsyncTaskTarget.cs
snakefoot: NET4_5 (and newer) has the new interface with object-parameter, while SILVERLIGHT5 and NET4_0 needs to perform lambda-context-capture.

src/NLog/Targets/AsyncTaskTarget.cs
snakefoot: Verbose -> continuation.Invoke(completedTask.Exception) ?

304NotModified:
```c#
var asyncContinuation = (AsyncContinuation)continuation; 
asyncContinuation(completedTask.Exception);
```

Otherwise parenthesis mind-overflow ;)

304NotModified:
I can also include this one in 4.4.6. What do you think?

snakefoot:
> I can also include this one in 4.4.6. What do you think?

There is no breaking changes included here. And it seems useful for at least one person. Not sure if the imposed limit of only one pending task is an issue.

304NotModified:
👍 merged!

Hussein-Dahir:
looks fantastic !
resolved issue [#2004](https://github.com/NLog/NLog/issues/2004) for me too !

but @snakefoot , what happens if the queue overflowed, meaning more than 10000 were logged in short time? will the overflowed logs be discarded?

I don't want this to happen, isn't there a way to grow the queue in this case?

I tried to combine both **AsyncTaskTarget** with **AsyncTargetWrapper** to increase the queue size and change overflowAction to "grow", but I think it was a bad idea, and noticed some exception arises.

So, is there a way to resolve [#2004](https://github.com/NLog/NLog/issues/2004) and keep all logs un-discarded?

snakefoot:
@Hussein-Dahir `AsyncTaskTarget` has an internal queue the grows to infinity, so if the Tasks constantly performs timeout then at one point it will have eaten all memory. LogEvents are only lost if the Task completes in failure.

I have been planning to add some extra features to `AsyncTaskTarget`:

- Bulk writing so more than one LogEvent can be written at a time (chunks)
- Async Retry-Logic that ensures correct ordering on ex network-failure.
- Better throttle logic so the internal queue cannot grow forever.

snakefoot:
@Hussein-Dahir What exceptions did you get when combining `AsyncTaskTarget` with `AsyncTargetWrapper` ?

snakefoot:
> @Hussein-Dahir However, if you are intereseted, I can show you the exact exception

Sounds strange with an exception from NLog on the mail-server, but please show the full stacktrace. Then we might figure out what it is.

304NotModified:
@Hussein-Dahir please create a new issue on github, thanks!

Hussein-Dahir:
@snakefoot @304NotModified No need :P
Sorry for that :P, but I found out that the exceptions is not related to NLog at all! (they keep arising even after changing the branch to one the doesn't use NLog).

Thanks for your quick response and for the support anyway !

PR #2017, opened 2017-03-17
WebServiceTarget - Fix boolean parameter conversion for Xml and Json (lowercase) & merged the JSON serializer code into DefaultJsonSerializer

### LINKED ISSUE #2016: target type WebService with protocol XmlPost with boolean parameter do not send valid xml values
Hi!

Please someone change/fix into  [NLog/src/NLog/Targets/WebServiceTarget.cs](https://github.com/NLog/NLog/blob/master/src/NLog/Targets/WebServiceTarget.cs) file into the method WriteAllParametersToCurrenElement the next code snippet.

Original code (at time was copied)
```c#
            protected void WriteAllParametersToCurrenElement(XmlWriter currentXmlWriter, object[] parameterValues)
            {
                for (int i = 0; i < Target.Parameters.Count; i++)
                {
                    currentXmlWriter.WriteElementString(Target.Parameters[i].Name, Convert.ToString(parameterValues[i], CultureInfo.InvariantCulture));
                }
            }
```
with this one
```c#
            protected void WriteAllParametersToCurrenElement(XmlWriter currentXmlWriter, object[] parameterValues)
            {
                for (int i = 0; i < Target.Parameters.Count; i++)
                {
                    currentXmlWriter.WriteElementString(Target.Parameters[i].Name, System.Xml.XmlConvert(parameterValues[i]));
                }
            }
```
if is possible.

The issue is with values which are not string. I encounter a issue with **System.Boolean** type. Where the call `Convert.ToString(parameterValues[i], CultureInfo.InvariantCulture)` will return values 'True', 'False', when valid values for xml are 'true', 'false', '0', '1'. The issue comes (in my case) when http service receives the xml and try to revert back the boolean value from xml text value and throw an exception that is invalid boolean format ...

Regards,
Victor Scorpan

snakefoot:
And also merged the JSON serializer code into DefaultJsonSerializer (Trying to fix #2016)

This also fixes the handling of boolean for the DefaultJsonSerializer, so it is now lowercase without being json-encoded (Fixed the existing test to be case-sensitive). This also fixes the handling of all non-string-objects in DefaultJsonSerializer (Ex DateTime or Guid), so when converted to string, then they are actually json-encoded as strings (Added new tests to verify this).

This also reduces the number of string-allocations in DefaultJsonSerializer, when serializing standard strings (better performance). The original EscapeString (would allocate StringBuilder and call ToString for every value) has been replaced by JsonStringEscape (only allocates StringBuilder when needed).

Modified the use of TypeCode, so it should also work on NET Core (Now using System.Convert.GetTypeCode instead of System.Type.GetTypeCode)

Will probably gives lots of conflicts in the NLog 5.0 Core-Branch (as you have already made change to fix the previous use of System.Type.GetTypeCode). But since I have moved the old code to XmlHelper.cs, then it should be easy to fix.


snakefoot:
Now changed JsonStringValue(object value) to use XmlConvert.ToString() as much as possible to avoid confusion (Though the result is the same).

Also added handling for DateTime-formatting, but it seems to be a dangerous minefield (Many different formatting options, and we only have one). But since one can always specify the format as string, and provide a custom format string, then it will not break anything. The handling of DateTime-formatting is only applied when having specified the parameter as System.DateTime (Just like the handling of Boolean-formatting is only applied when having the parameter as System.Boolean).

Maybe someone can answer why the unit-test enforced the Microsoft JSON-datacontract-formatter instead of the JSON-NET-formatter?

'Formatters.JsonFormatter.UseDataContractJsonSerializer = true'

I had to change it to the default JSON-NET-formatter for the DateTime to work with non-microsoft-wcf format.

src/NLog/Targets/MethodCallParameter.cs
```diff
@@ -98,7 +98,13 @@ public MethodCallParameter(string name, Layout layout, Type type)
         /// </summary>
         /// <docgen category='Parameter Options' order='10' />
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1721:PropertyNamesShouldNotMatchGetMethods", Justification = "Backwards compatibility")]
-        public Type Type { get; set; }
+        public Type Type { get { return ParameterType; } set { ParameterType = value; } }
```
304NotModified: maybe we should tell this is the same as `ParameterType` (with `<see`)

src/NLog/Internal/XmlHelper.cs
```diff
@@ -99,6 +99,99 @@ private static string CreateValidXmlString(string text)
...
+        internal static string XmlConvertToString(object value, out TypeCode objTypeCode)
+        {
+            if (value == null)
+            {
+                objTypeCode = TypeCode.Empty;
+                return "null";
+            }
+
+            objTypeCode = Convert.GetTypeCode(value);
+            switch (objTypeCode)
```
304NotModified: is working with typecode better than type? Any idea of the .NET core support?

snakefoot: Think switch-statement is faster than dictionary-lookup. System.Convert.GetTypeCode should also work in NET Core (It is System.Type.GetTypeCode that has been removed).

src/NLog/Targets/MethodCallParameter.cs
snakefoot: Fixed

304NotModified:
thanks!

PR #2025, opened 2017-03-23
MdlcLayoutRenderer - AsyncLocal for NET4_6 and NETSTANDARD1_3

### LINKED ISSUE #1729: MDLC support for .NET Core
Feature Inquiry/Request

NLog version: 4.4.0 - *

Platform: .Net Core

According to the Platform Support page, it seems that there is not yet planned support of MDLC for .Net Core yet, is this correct? And if so, do we know when can we expect the support to be available? 

Ask because we currently rely on MDLC to track things like correlation IDs and stuff, to keep them "logic context safe" and the render layout is really convenient to output them. 

Thanks.

snakefoot:
Also allow NET3_5 and MONO to use CallContext.LogicalGetData (But not Silverlight and WindowsPhone8)

Fixes  #1729

### DIFF
--- a/src/NLog/LayoutRenderers/MdlcLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/MdlcLayoutRenderer.cs
@@ -31,9 +31,10 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
+#if !SILVERLIGHT && !NETSTANDARD || NETSTANDARD1_3
+
 namespace NLog.LayoutRenderers
 {
-#if NET4_0 || NET4_5 && !NETSTANDARD_1plus
     using System.Text;
     using Config;
     using Internal;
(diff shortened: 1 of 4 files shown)


304NotModified:
Is this needed?

Netstandard and net46 are (only) in the coreclr branch. Currently (in master)net46 uses the net45 binary

snakefoot:
> Is this needed?

Maybe just use it for inspiration when doing the next merge to the CoreClr-branch.

snakefoot:
Now changed the PR to point against coreclr-branch. Had to disable the MdlcLayoutRendererTests as the use of constructor doesn't seem to work properly. Also strange one cannot see the test output for the Travis-build.

snakefoot:
@304NotModified Now squashed the PR, so it is a merge of a single change. Discovered that the MdlcLayoutRendererTests was also failing before (#2030), so I have left it untouched.

304NotModified:
Thanks! Will try to look at it soon.

304NotModified:
great! Thanks! Now we could also fix https://github.com/NLog/NLog.Extensions.Logging/issues/10 

\0/

PR #2038, opened 2017-03-29
Remove old Wix installer / makefile

304NotModified:


304NotModified:
Nuget.exe is still used

PR #1923, opened 2017-01-18
Enable OptimizeBufferReuse for simple Target Wrappers (but not inherited)

snakefoot:
Waiting for NLog ver. 4.5

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1923)

PR #2041, opened 2017-03-29
 Reduce memory allocations (AsyncContinuation exceptionHandler) & refactor

snakefoot:
Introduces ReusableObjectCreator as base-class for OptimizeBufferReuse

Code cleanup and found an extra place to reduce memory allocations (AsyncContinuation exceptionHandler).

### DIFF
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -957,15 +957,15 @@ namespace NLog.Targets
         /// <summary>
         /// Can be used if <see cref="Target.OptimizeBufferReuse"/> has been enabled.
         /// </summary>
-        private readonly ReusableStreamCreator reusableFileWriteStream = new ReusableStreamCreator();
+        private readonly ReusableStreamCreator reusableFileWriteStream = new ReusableStreamCreator(1024);
         /// <summary>
         /// Can be used if <see cref="Target.OptimizeBufferReuse"/> has been enabled.
         /// </summary>
-        private readonly ReusableStreamCreator reusableAsyncFileWriteStream = new ReusableStreamCreator();
+        private readonly ReusableStreamCreator reusableAsyncFileWriteStream = new ReusableStreamCreator(1024);
         /// <summary>
         /// Can be used if <see cref="Target.OptimizeBufferReuse"/> has been enabled.
         /// </summary>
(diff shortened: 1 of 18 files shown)


src/NLog/Targets/FileTarget.cs
```diff
@@ -957,15 +957,15 @@ protected override void CloseTarget()
...
+        private readonly ReusableStreamCreator reusableFileWriteStream = new ReusableStreamCreator(1024);
         /// <summary>
         /// Can be used if <see cref="Target.OptimizeBufferReuse"/> has been enabled.
         /// </summary>
-        private readonly ReusableStreamCreator reusableAsyncFileWriteStream = new ReusableStreamCreator();
+        private readonly ReusableStreamCreator reusableAsyncFileWriteStream = new ReusableStreamCreator(1024);
         /// <summary>
         /// Can be used if <see cref="Target.OptimizeBufferReuse"/> has been enabled.
         /// </summary>
-        private readonly ReusableBufferCreator reusableEncodingBuffer = new ReusableBufferCreator(4096);
```
304NotModified: is this on intentionally changed to 1024?

src/NLog/Targets/FileTarget.cs
```diff
@@ -957,15 +957,15 @@ protected override void CloseTarget()
         /// <summary>
         /// Can be used if <see cref="Target.OptimizeBufferReuse"/> has been enabled.
         /// </summary>
-        private readonly ReusableStreamCreator reusableFileWriteStream = new ReusableStreamCreator();
+        private readonly ReusableStreamCreator reusableFileWriteStream = new ReusableStreamCreator(1024);
```
304NotModified: what was this value (1024) before?

snakefoot: No it was empty MemoryStream-object, but since it is enabled by default, then I wanted a good default value.

src/NLog/Targets/FileTarget.cs
snakefoot: Yes the encoding-buffer is only used for a single LogEvent, and small buffers are preferred.

snakefoot:
@304NotModified Not possible to get into 4.4.6 ?

304NotModified:
Yes, also possible. Was thinking it was related to https://github.com/NLog/NLog/pull/1923 etc

304NotModified:
merged!

PR #2042, opened 2017-03-30
AutoFlushTargetWrapper - Added AsyncFlush property

snakefoot:
Allow the following setup to work out of the box:

```
  <target name="mailautoflush" xsi:type="AutoFlushWrapper" condition="level >= LogLevel.Error">
    <target name="mailbuffer" xsi:type="BufferingWrapper" bufferSize="1000000">
      <target name="mail" xsi:type="Mail" ... />
    </target>
  </target>
```

Before the BufferingWrapper would capture the LogEvent, so it wouldn't reach the final target, so the attached flush-operation was not executed. Disabling AsyncFlush will make AutoFlushWrapper work together with BufferingWrapper.

### DIFF
--- a/src/NLog/Targets/Wrappers/AutoFlushTargetWrapper.cs
+++ b/src/NLog/Targets/Wrappers/AutoFlushTargetWrapper.cs
@@ -63,7 +63,13 @@ namespace NLog.Targets.Wrappers
         /// </summary>
         public ConditionExpression Condition { get; set; }
 
-        private readonly AsyncOperationCounter pendingManualFlushList = new AsyncOperationCounter();
+        /// <summary>
+        /// Delay the flush until the LogEvent has been confirmed as written
+        /// </summary>
+        public bool AsyncFlush { get { return _asyncFlush ?? true; } set { _asyncFlush = value; } }
+        private bool? _asyncFlush;
+
+        private readonly AsyncOperationCounter _pendingManualFlushList = new AsyncOperationCounter();
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


304NotModified:
Good catch!

304NotModified:
added to https://github.com/NLog/NLog/wiki/AutoFlushWrapper-target

FixRM:
Hello! I'm experiencing same issue with Network target wrapped with AutoFlushWrapper and BufferingWrapper. Nothing is passed to final target. NLog 5.2.2 config:
```xml
	  xsi:schemaLocation="n NLog.xsd"
	  autoReload="true" 
	  internalLogToConsole="true" internalLogLevel="Info">
	<targets>
		<target name="Logstash" xsi:type="AutoFlushWrapper" condition="level >= LogLevel.Error" flushOnConditionOnly="true">
			<target xsi:type="BufferingWrapper" bufferSize="5" overflowAction="Discard">
				<!-- TCP protocol must be used and newLine is critical -->
				<target name="Network" xsi:type="Network" address="tcp://localhost:9045" newLine="true" />
					<layout xsi:type="JsonLayout" includeEventProperties="true" maxRecursionLimit="2" excludeEmptyProperties="true"/>
			</target>
		</target>
	</targets>
	<rules>
		<logger name="*" minlevel="Info" writeTo="Logstash" />
	</rules>
</nlog>
```
What am I doing wrong? Final target works well without wrappers.

snakefoot:
@FixRM Please create new issue, instead of waking up old ones.

FixRM:
Thanks, @snakefoot. Done: https://github.com/NLog/NLog/issues/5305

PR #2048, opened 2017-04-03
Layout - Ensure StackTraceUsage works for all types of Layout

### LINKED ISSUE #2044: Sourcefilename and number
**Type** :Question
**NLog version** :4.4.5
**Platform**: .Net 4.0 4.5

Hello together;

Bug or problem ? i dont know

I try to send out into nlogviewer the message + Sourcecodefile and number
thats my konfig for this:
        <target name="Viewer" xsi:type="NLogViewer"
            address="udp4:127.0.0.1:4000"
            includeCallSite="true"
            includeSourceInfo="true"
            includeMdc="true"
            includeNdc="true"
            includeNLogData="true"            
            />            
   
everything works fine but i dont see the Sourcefilename and Linenumber
Thats send to my viewer:

```
<log4j:event` logger="Program" level="TRACE" timestamp="1490959857884" thread="1">
        <log4j:message>
                entering func 'get_Givestring' , Stackframe='22'
        </log4j:message>
        <log4j:locationInfo class="Program.con"
                            method="System.String get_Givestring(System.String)" 
                            file=""
                            line="0"/>
        <log4j:properties>
                <log4j:data name="log4japp" value="Test.exe(12148)"/>
                <log4j:data name="log4jmachinename" value="PC1"/>
        </log4j:properties>
</log4j:event>`
```
thanks for helping 

Klaus

snakefoot:
Fixes #2044 

Little curious whether I'm making a regression of the bug #1262 solved with PR #1310. 

But the difference from that old bug is that now the caching of Layout.StackTraceUsage happens at initialization (single-threaded) instead of at Logger-write (multi-threaded).

This fix has become a very big hammer to a small nail  (Could just have fixed Log4JXmlEventLayout to inherit from IUsesStackTrace). The changes included in this PR:

- Faster allocation of new Loggers, as it now caches the lookup of StackTraceLevel on Target.
- Layout initialization will always scan for StackTraceLevel and ThreadAgnostic. Before it would only scan when initialized with LoggingConfiguration.
- Fixes problem with stale caching of Target.allLayoutsAreThreadAgnostic. Now it is refreshed when calling LogManager.ReconfigExistingLoggers()
  - Same fix is used to refresh the new caching of StackTraceLevel on Target (FindAllLayouts)
- LogManager.ReconfigExistingLoggers() has now become slower as it will now also refresh the caching on all registered Targets (FindAllLayouts)
- LogManager.ReconfigExistingLoggers() is now protected against multiple threads performing re-initialization (will still refresh loggers outside lock)
- Fixes the actual problem with detecting IUsesStackTrace-renders for all Layouts (Also those not depending on SimpleLayout).

### DIFF
--- a/src/NLog/Internal/TargetWithFilterChain.cs
+++ b/src/NLog/Internal/TargetWithFilterChain.cs
@@ -31,8 +31,6 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
-using System.Linq;
-
 namespace NLog.Internal
 {
     using System.Collections.Generic;
@@ -97,10 +95,9 @@ namespace NLog.Internal
 
             // find all objects which may need stack trace
             // and determine maximum
-            // only the target can have IUsesStackTrace
             if (Target != null)
             {
-                stackTraceUsage = Target.GetAllLayouts().OfType<IUsesStackTrace>().DefaultIfEmpty().Max(usage => usage == null ? StackTraceUsage.None : usage.StackTraceUsage);
+                stackTraceUsage = Target.StackTraceUsage;
             }
 
             //recurse into chain if not max
--- a/src/NLog/Layouts/Layout.cs
+++ b/src/NLog/Layouts/Layout.cs
@@ -31,14 +31,15 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
-using System;
-
 namespace NLog.Layouts
 {
+    using System;
+    using System.Linq;
     using System.ComponentModel;
     using System.Text;
     using NLog.Config;
     using NLog.Internal;
+    using NLog.Common;
 
     /// <summary>
     /// Abstract interface that layouts must implement.
@@ -51,12 +52,7 @@ namespace NLog.Layouts
         /// Is this layout initialized? See <see cref="Initialize(NLog.Config.LoggingConfiguration)"/>
         /// </summary>
         private bool isInitialized;
-
-        /// <summary>
-        /// Does the layout contains threadAgnostic layout renders? If contains non-threadAgnostic-layoutrendender, then this layout is also not threadAgnostic. 
-        /// See <see cref="IsThreadAgnostic"/> and <see cref="Initialize"/>.
-        /// </summary>
-        private bool threadAgnostic;
+        private bool scannedForObjects;
 
         /// <summary>
         /// Gets a value indicating whether this layout is thread-agnostic (can be rendered on any thread).
@@ -64,12 +60,15 @@ namespace NLog.Layouts
         /// <remarks>
         /// Layout is thread-agnostic if it has been marked with [ThreadAgnostic] attribute and all its children are
         /// like that as well.
+        /// 
         /// Thread-agnostic layouts only use contents of <see cref="LogEventInfo"/> for its output.
         /// </remarks>
-		internal bool IsThreadAgnostic
-		{
-            get { return this.threadAgnostic; }
-        }
+        internal bool ThreadAgnostic { get; private set; }
+
+        /// <summary>
+        /// Gets the level of stack trace information required for rendering.
+        /// </summary>
+        internal StackTraceUsage StackTraceUsage { get; private set; }
 
         private const int MaxInitialRenderBufferLength = 16384;
         private int maxRenderedLength;
@@ -124,7 +123,7 @@ namespace NLog.Layouts
         /// </remarks>
         public virtual void Precalculate(LogEventInfo logEvent)
         {
-            if (!this.threadAgnostic)
+            if (!this.ThreadAgnostic)
             {
                 this.Render(logEvent);
             }
@@ -139,8 +138,7 @@ namespace NLog.Layouts
         {
             if (!this.isInitialized)
             {
-                this.isInitialized = true;
-                this.InitializeLayout();
+                this.Initialize(this.LoggingConfiguration);
             }
 
             return this.GetFormattedMessage(logEvent);
@@ -148,7 +146,7 @@ namespace NLog.Layouts
 
         internal void PrecalculateBuilder(LogEventInfo logEvent, StringBuilder target)
         {
-            if (!this.threadAgnostic)
+            if (!this.ThreadAgnostic)
             {
                 RenderAppendBuilder(logEvent, target, true);
             }
@@ -164,11 +162,10 @@ namespace NLog.Layouts
         {
             if (!this.isInitialized)
             {
-                this.isInitialized = true;
-                this.InitializeLayout();
+                this.Initialize(this.LoggingConfiguration);
             }
 
-            if (!this.threadAgnostic)
+            if (!this.ThreadAgnostic)
             {
                 string cachedValue;
                 if (logEvent.TryGetCachedLayoutValue(this, out cachedValue))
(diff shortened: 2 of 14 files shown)


src/NLog/Internal/TargetWithFilterChain.cs
```diff
@@ -97,10 +97,9 @@ internal StackTraceUsage PrecalculateStackTraceUsage()
 
             // find all objects which may need stack trace
             // and determine maximum
-            // only the target can have IUsesStackTrace
             if (Target != null)
             {
-                stackTraceUsage = Target.GetAllLayouts().OfType<IUsesStackTrace>().DefaultIfEmpty().Max(usage => usage == null ? StackTraceUsage.None : usage.StackTraceUsage);
+                stackTraceUsage = Target.GetAllLayouts().DefaultIfEmpty().Max(layout => layout == null ? StackTraceUsage.None : layout.StackTraceUsage);
```
304NotModified: maybe it's a good time to move this to `Target.Initialize` ?

src/NLog/Layouts/Layout.cs
```diff
@@ -282,13 +295,26 @@ internal void Initialize(LoggingConfiguration configuration)
...
                 foreach (object item in ObjectGraphScanner.FindReachableObjects<object>(this))
                 {
-                    if (!item.GetType().IsDefined(typeof(ThreadAgnosticAttribute), true))
+                    if (this.threadAgnostic)
                     {
-                        this.threadAgnostic = false;
-                        break;
+                        if (!item.GetType().IsDefined(typeof(ThreadAgnosticAttribute), true))
+                        {
+                            this.threadAgnostic = false;
```
304NotModified: I see we lost the `break` here. AFAIK `IsDefined(typeof(ThreadAgnosticAttribute),` not very cheap. 

Maybe we should loop two times (same list) over `ObjectGraphScanner.FindReachableObjects<object>(this)`, and has `breaks` in both loops?

snakefoot: Notice the new if-statement above the IsDefined.

304NotModified: good one. So not in favor of two loops? It's a bit hard to read.

snakefoot: Have now changed it to a Linq-expression.

src/NLog/Internal/TargetWithFilterChain.cs
snakefoot: Agreed.

304NotModified:
oops I broke appveyor.yml

304NotModified:
unbroke it

snakefoot:
@304NotModified Discovered that removing interface from IUsesStackTrace from SimpleLayout, would be a breaking change. Now added it back.

Also discovered that modifying SimpleLayout.Text would automatically refresh the StackTraceUsage. This has also been added back.

NOTE! The caching of StackTraceUsage that you have requested for Target will not be refreshed when creating a new Logger (In case the SimpleLayout-Text has been modified).

304NotModified:
bitdiffer! I like to add that to that to Appveyor. Do you have the command for it?

src/NLog/Layouts/CompoundLayout.cs
```diff
@@ -64,9 +64,9 @@ public CompoundLayout()
         /// </summary>
         protected override void InitializeLayout()
         {
-            base.InitializeLayout();
             foreach (var layout in Layouts)
                 layout.Initialize(this.LoggingConfiguration);
+            base.InitializeLayout();
```
304NotModified: ow this looks like a breaking behavior change.

304NotModified:
> NOTE! The caching of StackTraceUsage that you have requested for Target will not be refreshed when creating a new Logger (In case the SimpleLayout-Text has been modified). But the same problem also exists for Target.allLayoutsAreThreadAgnostic

Could you elaborate on this? I'm missing the reference between the Logger and SimpleLayout.

src/NLog/Layouts/CompoundLayout.cs
snakefoot: Since the base.InitializeLayout() performs object-scanning, then I wanted it to be last, so it would find all relevant objects. But I will revert the change.

snakefoot:
> Could you please elaborate on this?

When creating a Logger, then it calls LogFactory.GetTargetsByLevelForLogger(). Before this would re-check the StackTraceUsage-status for all Layouts for each Target. When the Target caches the StackTraceUsage, then it will not discover that Layout maybe has been modified.

304NotModified:
not sure what was wrong on AppVeyor. (build was stuck - 1 hour timeout). Restarted it

304NotModified:
@snakefoot What do you think? Should this in NLog 4.5 or 4.4.6?

snakefoot:
I would be nice to have this resolved "Fixes problem with stale caching of Target.allLayoutsAreThreadAgnostic" (This bug was introduced with 4.4.2).

The original bug has been there since NLog ver. 4.3, so I don't think it is super urgent. The only issue is that the PR covers several files, so the chance of conflicts is high while waiting for 4.5.

src/NLog/Targets/Target.cs
```diff
@@ -453,18 +451,10 @@ protected virtual void InitializeTarget()
...
-            foreach (Layout layout in this.allLayouts)
-            {
-                if (!layout.IsThreadAgnostic)
-                {
-                    foundNotThreadAgnostic = true;
-                    break;
-                }
-            }
-            this.allLayoutsAreThreadAgnostic = !foundNotThreadAgnostic;
+            this.allLayoutsAreThreadAgnostic = !allLayouts.Any(layout => !layout.ThreadAgnostic);
```
304NotModified: ! Any => .All ? ;)

tests/NLog.UnitTests/Layouts/CsvLayoutTests.cs
```diff
@@ -331,6 +331,7 @@ public void CsvLayoutCachingTest()
                         new CsvColumn("date", "${longdate}"),
                         new CsvColumn("level", "${level}"),
                         new CsvColumn("message", "${message}"),
+                        new CsvColumn("threadid", "${threadid}"),
```
304NotModified: was a new unit test not a better option?

src/NLog/Layouts/Layout.cs
```diff
@@ -271,35 +268,45 @@ void ISupportsInitialize.Close()
...
-		{
+        internal void Initialize(LoggingConfiguration configuration)
+        {
             if (!this.isInitialized)
             {
                 this.LoggingConfiguration = configuration;
                 this.isInitialized = true;
+                this.scannedForObjects = false;
+
+                this.InitializeLayout();
```
304NotModified: this is the only thing i'm doubting about. It's now before the scan (yes I see the default scan in InitializeLayout).

I a subclass is dependent of the scan, then we could have problems if it calls base. InitializeLayout() not (or too late)?

tests/NLog.UnitTests/Layouts/CsvLayoutTests.cs
snakefoot: Well the unit-test expects the render-result to be cached. And after NLog ver. 4.4.2 then it only caches threadAgnostic = false (Default when not initialized properly)

src/NLog/Layouts/Layout.cs
snakefoot: For the normalt NLog Layouts, then the scan is purely done for sake of the Target and the later processing of LogEventInfos. Pretty sure no one knows how threadagnostic and stacktraceusage works :)

snakefoot:
> Do you think we have to performance test this one? As it touches all layout and targets?

All the code changes are related to initial configuration, and reload-configuration.

src/NLog/Targets/Target.cs
snakefoot: Fixed

304NotModified:
Thanks for the explanations and code!

304NotModified:
FYI,

NLog 4.4.6 is live! 

https://www.nuget.org/packages/NLog/4.4.6

Issue #2014, opened 2017-03-15
Add Overloads That Support LogMessageGenerator with Exception


TheXenocide:
**Type**: Feature request

**NLog version**: 4.4.2

**Platform**: .Net 4.5

While working on our internal logging facade I noticed NLog doesn't support deferred message generation on method overloads that also support Exceptions. In many common scenarios I can see how Exceptions would be less likely to be filtered, but given support for logging exceptions as Verbose or Information it seems like it would still be relevant to allow message generation to happen after filtering rules are applied the same as any other message.

TheXenocide:
I figured that would be the rationale, but there are known exceptions that can be handled but logged for diagnostic purposes as verbose/debug/etc. which are much less likely to be automatically redirected to the Event Log than warning/error/fatal. Our project is on a very tight timeline at the moment, so for now we're just invoking our delegate and passing the result along, but it would be nice in the long run so maybe I'll see if I'm able to come back to it later. Still, thought I'd bring it up while passing through. Thanks for everything you guys do :)

304NotModified:
Well forgot it, and I think it should be there for consistency :)

c0shea:
Submitted PR #2051 for this feature

TheXenocide:
Awesome, thanks guys! Out of curiosity, any idea when 4.4.6 will be hitting NuGet?

304NotModified:
closed by #2051

304NotModified:
This week :) 

If you need it earlier, I can push a beta

TheXenocide:
We're really close to the end of our project so I think we'll wait for the stable build just to reduce likelihood of issues, but that's great, thanks a bunch!

PR #2051, opened 2017-04-04
Added LogMessageGenerator overloads for exceptions

c0shea:
Added support for deferred message generation on method overloads that also support Exceptions. See #2014

### DIFF
--- a/src/NLog/ILogger.cs
+++ b/src/NLog/ILogger.cs
@@ -50,54 +50,54 @@ namespace NLog
         /// Gets a value indicating whether logging is enabled for the <c>Trace</c> level.
         /// </summary>
         /// <returns>A value of <see langword="true" /> if logging is enabled for the <c>Trace</c> level, otherwise it returns <see langword="false" />.</returns>
-        bool IsTraceEnabled 
-        { 
-            get; 
+        bool IsTraceEnabled
+        {
+            get;
         }
 
         /// <summary>
         /// Gets a value indicating whether logging is enabled for the <c>Debug</c> level.
         /// </summary>
         /// <returns>A value of <see langword="true" /> if logging is enabled for the <c>Debug</c> level, otherwise it returns <see langword="false" />.</returns>
-        bool IsDebugEnabled 
-        { 
-            get; 
+        bool IsDebugEnabled
+        {
+            get;
         }
 
         /// <summary>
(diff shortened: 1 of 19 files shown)


304NotModified:
Thanks!

Unfortunately this isn't backwards-compatible. I think we should write a C# extensions. What do you think?

PS: changes ignoring whitespace changes: https://github.com/NLog/NLog/pull/2051/files?w=1

c0shea:
Could you clarify what you mean by it being a breaking change so I know for next time?

I think extensions sound good, though. Do you mean something like [this](https://gist.github.com/c0shea/43e51c671681d6a1a3d85d9be0e93ea1)?

304NotModified:
Well extending a public interface is a breaking change as other uses who implement that interface have to implement those new methods. 

(Custom) Code written to NLog 4.x, should always work for every NLog 4.y (where `y>=x`) version (we are taking [semver](semver.org) very serious)

304NotModified:
A full list of what is breaking and what is not: https://github.com/dotnet/corefx/blob/c02d33b18398199f6acc17d375dab154e9a1df66/Documentation/coding-guidelines/breaking-change-rules.md

c0shea:
Thanks for the link! That list makes a lot of sense. How should I back out these breaking changes in order to add the extension methods?

304NotModified:
Well you could add methods to the interface by using c# extensions. That won't break anything :)

c0shea:
Thanks for the suggestion. I added a new class for the extension methods. Let me know what you think.

304NotModified:
Looks good, but the changes to `ILogger` needs to be reverted? 

What's the problem with the CLScompliant? We prefer CLScompliant code

c0shea:
Whoops, I forgot to revert the changes to `ILogger` earlier. They're reverted now.

I couldn't have the extensions be CLS-compliant because the `ILogger` parameter isn't. Both `ILogger-V1.cs` and `ILoggerBase-V1.cs` are decorated with `[CLSCompliant(false)]` and caused the compile to fail.

304NotModified:
Thanks! Merged!  Will be in NLog 4.4.6

304NotModified:
FYI,

NLog 4.4.6 is live! 

https://www.nuget.org/packages/NLog/4.4.6

PR #2053, opened 2017-04-06
Add UWP support

gtbuchanan:
Fixes #1879. I have not tested thoroughly since I really only need basic console logging functionality right now, but this got it to build and log successfully in a UWP project.

PR #2055, opened 2017-04-06
Support pure NETSTANDARD (UWP)

snakefoot:
Ideas stolen from #2053 (Fixed unit tests to pass in pure NETSTANDARD mode, though not activated).

This will allow pure Windows Store Apps to use NLog ver. 5 without depending on System.Threading.Thread (not available until NetStandard2_0).

Added new Uap10.0 build that is a pure NETSTANDARD1_3-build, and created a new define NETSTANDARD1_3PLUS that includes all the stuff that UWP do not support.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/2055)


tests/NLog.UnitTests/UnitTestApp.xaml.cs
```diff
@@ -1,4 +1,4 @@
-﻿#if NETSTANDARD && !NETSTANDARD1_3
+﻿#if NETSTANDARD && UWP10
```
304NotModified: UWP10 isn't necessary anymore?

snakefoot: Well the code doesn't compile, unless having some UWP libraries. Didn't want to investigate, just wanted to test without NETSTANDARD1_3.

snakefoot:
> Cool, but i'm not sure if this solves the UWP problem.

NLog NetStandard1_3 makes use of libraries not available for UWP (Trace, Process, Mutex, Thread, FileWatcher).

I don't know if it would make sense to only have a pure NetStandard1_4 (UWP) and NetStandard1_5 (Thread + Process + Environment + etc.)

304NotModified:
Wouldn't it be more logical to fix netstandard 1.3 or add uwp10 instead of the netstandard 1.4? 

Or are those libs used in netstandard 1.3 also not used in 1.5?

snakefoot:
> Wouldn't it be more logical to fix netstandard 1.3 or add uwp10 instead of the netstandard 1.4?

Well it seems NetStandard 1.4 is the first NetStandard-version that supports UWP. Don't know if there is a special TargetFramework that specifies NetStandard-mode supported by UWP ?

snakefoot:
Just tested the local compiled NLog NetStandard1_3, and it works when writing to the following location:

- Windows.Storage.ApplicationData.Current.LocalFolder.Path

And when using the current NLog 5.0 Beta6, then one get the following exception:

> System.IO.FileSystem.Watcher, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified.'

304NotModified:
> That should also work. Then NetStandard1_3 would be the pure build, that can be used by UWP (Only depending on basic libraries).

Any idea of the UWP10 framework gives compile errors instead of runtime? 

Making exceptions just for UWP, sounds that we should create a UWP10 build.

snakefoot:
> Any idea of the UWP10 framework gives compile errors instead of runtime?

UWP App compiles fine, but gives runtime exception when trying to access/load the NLog library/functionality.

> Making exceptions just for UWP, sounds that we should create a UWP10 build.

Not sure I understand. What exceptions are we talking about?

304NotModified:
> Not sure I understand. What exceptions are we talking about? NetStandard1_3 is now platform independent, and NetStandard1_5 depends on platform dependent libraries.

The NetStandard1_3 . 

Any idea if a program in .NET 4.6 uses the .NET 4.5 or the NetStandard  DLL?

snakefoot:
> Any idea if a program in .NET 4.6 uses the .NET 4.5 or the NetStandard DLL?

I certainly hope it depends on NET4_5, as the NETSTANDARD is still very crippled until NetStandard2_0.

304NotModified:
> I certainly hope it depends on NET4_5, as the NETSTANDARD is still very crippled until NetStandard2_0.

Agree on that ;)

snakefoot:
> The NetStandard1_3 .

Still not sure I understand, about the exception thing. Is it bad to have two NetStandard-builds? And better to have a UWP_NetStandard1_3 build and a NetStandard1_5 build?

Anyways very satisfied with the changes in this PR independent of NetStandard1_3 becoming platform independent or not. The only difference between NetStandard1_3 and NetStandard1_5 is the ability to use file-watcher (And the environment-layout).

304NotModified:
> And better to have a UWP_NetStandard1_3 build and a NetStandard1_5 build?

Doubting about this. I think it's better to have indeed a UWP build currently.

snakefoot:
> Doubting about this. I think it's better to have indeed a UWP build currently.

So introducing a new NETSTANDARD_UWP that specifies that System.Threading and System.Process is not available?

snakefoot:
@gtbuchanan & @304NotModified I have now added a new define NETSTANDARD1_3PLUS. It means NETSTANDARD1_3 with extra dependencies (non-UWP).

I have no clue (or maybe patience for) how to modify the build-configuration to make an UWP-NETSTANDARD1_3 build (Without the define NETSTANDARD1_3PLUS).

But maybe adding this new UWP-build-definition can be made in a new PR.

src/NLog/project.json
```diff
@@ -78,27 +78,58 @@
     //    "define": [ "NET4_0", "WCF_SUPPORTED", "NETSTANDARD1_0", "NETSTANDARD" ]
     //  }
     //},
+    //"uap": {
```
gtbuchanan: You almost have it. I had UWP building with my `project.json` configuration [here](https://github.com/NLog/NLog/pull/2053/files/bf377d08a97447ff5f94a6961a535570dffa554d#diff-4c95cd96a32a135a3f845aad76498b78R226). I think `uap` needs to be `uap10.0` and `NETStandard.Library` replaced with `Microsoft.NETCore.UniversalWindowsPlatform`.

snakefoot:
@gtbuchanan Would have preferred to make some kind of NetStandard1.3 build without reference to UWP or Uap10.0.

snakefoot:
@304NotModified Have been able to show failing tests on Travis (Caused by NLog tests stealing Console-out). Not sure what is going on for the appveyor-build.

snakefoot:
@304NotModified Have now made AppVeyor build fail without using ErrorActionPreference, but instead by returning the exitcode from dotnet test.

304NotModified:
I think this is still WIP?

304NotModified:
about the failing `CallSiteLineNumberTests` under netstandard. I was thinking I disabled them (skip in xunit), but maybe that's still in another branch

snakefoot:
> I think this is still WIP?

Nah just me trying to fix all the failing unit tests revealed in the Travis build, but it is probably better for another PR.

304NotModified:
Would be nice of we could got appveyor working (by skipping tests on netstandard).

snakefoot:
> Would be nice of we could got appveyor working (by skipping tests on netstandard).

Yes but I think that is for a different PR. My guess is that they are failing because they run in release-mode, instead of debug-mode.

304NotModified:
Could you also skip AssemblyVersionTests + ExtensionTests netstandard? Those are on the todo list. :) thanks! (i think that are the lasts to get appveyor succeeding)

snakefoot:
> Could you also skip AssemblyVersionTests + ExtensionTests netstandard? 

Like I said before, they are not failing on NETSTANDARD. It is the NET4_5 build :)

304NotModified:
> Like I said before, they are not failing on NETSTANDARD. It is the NET4_5 build :)

oops sorry missed that

304NotModified:
the failing ExtensionTests is because of the extra folder in `bin`. But maybe a bit out of scope for this PR ;)

snakefoot:
Have added the missing dependency to the unit-testing-project, and changed how to lookup the full-path of SampleExtensions.dll.

Now extension dll loading is working for NetStandard1.5 on Travis,

304NotModified:
Is this WIP?

snakefoot:
> Is this WIP?

Nope just stumbling around.

gtbuchanan:
Heads up, I'm running into an issue building my app in Release mode. UWP appears to require a [Runtime Directives file (rd.xml) file](https://msdn.microsoft.com/en-us/library/dn600639%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396) to define which types are accessed via reflection so they aren't stripped from the build.

I encounter the following unhandled exception when my code sets `LogManager.Configuration` in Release mode:

> Exception thrown: 'System.Reflection.MissingMetadataException' in System.Private.Reflection.Core.dll
> 
> Additional information: 'System.Text.UTF8Encoding' is missing metadata. For more information, please visit http://go.microsoft.com/fwlink/?LinkID=392859

The call stack looks like this:

 	

```
System.Private.Reflection.Core.dll!System.Reflection.Runtime.TypeInfos.RuntimeNoMetadataNamedTypeInfo.CustomAttributes.get() Line 52	C#
 	System.Reflection.Extensions.dll!System.Reflection.CustomAttributeInheritanceRules.TypeCustomAttributeSearcher.GetDeclaredCustomAttributes(System.Reflection.TypeInfo element) Line 112	C#
 	System.Reflection.Extensions.dll!System.Reflection.CustomAttributeSearcher<System.Reflection.TypeInfo>.GetMatchingCustomAttributesIterator(System.Func<System.Type,bool> rawPassesFilter, System.Reflection.TypeInfo element, bool inherit) Line 114	C#
 	System.Reflection.Extensions.dll!System.Linq.LowLevelEnumerable.Any<System.Reflection.CustomAttributeData>(System.Collections.Generic.IEnumerable<System.Reflection.CustomAttributeData> values) Line 17	C#
 	System.Reflection.Extensions.dll!System.Reflection.CustomAttributeExtensions.IsDefined(System.Reflection.MemberInfo element, System.Type attributeType, bool inherit) Line 292	C#
 	NLog.dll!NLog.Internal.ReflectionHelpers.IsDefined<NLog.Config.NLogConfigurationItemAttribute>(System.Type type, bool inherit)	Unknown
 	NLog.dll!NLog.Internal.ObjectGraphScanner.ScanProperties<object>(System.Collections.Generic.List<object> result, object o, int level, System.Collections.Generic.HashSet<object> visitedObjects)	Unknown
 	NLog.dll!NLog.Internal.ObjectGraphScanner.ScanProperties<object>(System.Collections.Generic.List<object> result, object o, int level, System.Collections.Generic.HashSet<object> visitedObjects)	Unknown
 	NLog.dll!NLog.Internal.ObjectGraphScanner.ScanProperties<object>(System.Collections.Generic.List<object> result, object o, int level, System.Collections.Generic.HashSet<object> visitedObjects)	Unknown
 	NLog.dll!NLog.Internal.ObjectGraphScanner.ScanPropertiesList<object>(System.Collections.Generic.List<object> result, System.Collections.Generic.IEnumerable<object> elements, int level, System.Collections.Generic.HashSet<object> visitedObjects)	Unknown
 	NLog.dll!NLog.Internal.ObjectGraphScanner.ScanProperties<object>(System.Collections.Generic.List<object> result, object o, int level, System.Collections.Generic.HashSet<object> visitedObjects)	Unknown
 	NLog.dll!NLog.Internal.ObjectGraphScanner.FindReachableObjects<object>(object[] rootObjects)	Unknown
 	NLog.dll!NLog.Config.LoggingConfiguration.ValidateConfig()	Unknown
 	NLog.dll!NLog.Config.LoggingConfiguration.InitializeAll()	Unknown
 	NLog.dll!NLog.LogFactory.Configuration.set(NLog.Config.LoggingConfiguration value)	Unknown
 	NLog.dll!NLog.LogManager.Configuration.set(NLog.Config.LoggingConfiguration value)	Unknown

```

This doesn't sound like a fun undertaking. I'm going to try to build rd.xml as I go to work through the errors.

snakefoot:
@gtbuchanan Met the same challenge, seems NET Native is playing tricks with reflection. I have tried to resolve the exceptions without needing to update rd.xml.

Have you tried to use the NuGet-package generated by the appveyor-build (NLog.dll from Uap10.0 build) ?

304NotModified:
Reviewed 15 of 37 files at r1, 48 of 60 files at r2, 7 of 10 files at r3, 1 of 1 files at r4, 4 of 5 files at r5, 6 of 8 files at r7, 2 of 2 files at r8, 7 of 8 files at r9, 1 of 1 files at r10.
Review status: 91 of 95 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[src/NLog/project.json, line 138 at r10](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiWv1ZBElXlGgI9ZS3U:-KiWv1ZBElXlGgI9ZS3V:b3ksz) ([raw file](https://github.com/nlog/nlog/blob/3144c84c9448f5828e2ad57930fe95ff5845b683/src/NLog/project.json#L138)):*
> ```JSON
>         "System.Security.Principal": "4.0.1",
>         "System.Threading.AccessControl": "4.0.0"
>       },
> ```

x

---

*[src/NLog/project.json, line 140 at r10](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiWv1D4qCn9_xhanZ_w:-KiWv1D4qCn9_xhanZ_x:b-b2a021) ([raw file](https://github.com/nlog/nlog/blob/3144c84c9448f5828e2ad57930fe95ff5845b683/src/NLog/project.json#L140)):*
> S0436

---

*[src/NLog/Internal/ObjectGraphScanner.cs, line 92 at r10](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiWvF9vSC_jEDdJLYdm:-KiWvF9vSC_jEDdJLYdn:b3ksz) ([raw file](https://github.com/nlog/nlog/blob/3144c84c9448f5828e2ad57930fe95ff5845b683/src/NLog/Internal/ObjectGraphScanner.cs#L92)):*
> ```Smalltalk
>                 return;
>             }
> ```

x

---

*[src/NLog/Internal/ObjectGraphScanner.cs, line 127 at r10](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiWrS9X2ObfTUrTBjiM:-KiWrS9X2ObfTUrTBjiN:b-eu1hzh) ([raw file](https://github.com/nlog/nlog/blob/3144c84c9448f5828e2ad57930fe95ff5845b683/src/NLog/Internal/ObjectGraphScanner.cs#L127)):*
> ```Smalltalk
>                 catch (System.Exception ex)
>                 {
>                     InternalLogger.Info(ex, "{0}Type reflection not possible for property {1}. Maybe because of .NET Native.", new string(' ', level + 1), prop.Name);
> ```

maybe we should move this to `IsDefined` as this is a helper already?

we should use also the generic one here I guess (see above)

`type.IsDefined<NLogConfigurationItemAttribute>`

---

*[src/NLog/Internal/TargetWithFilterChain.cs, line 119 at r10](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiWrzUzlKQn8sjgJxiD:-KiWrzV-siS25V9Spj03:b-duo7lx) ([raw file](https://github.com/nlog/nlog/blob/3144c84c9448f5828e2ad57930fe95ff5845b683/src/NLog/Internal/TargetWithFilterChain.cs#L119)):*
> ```Smalltalk
>                 try
>                 {
>                     // Check if hack used by LoggerImpl.Write() is possible
> ```

maybe a good candidate to move to StackTraceUsageUtils?

---

*[src/NLog/LayoutRenderers/NLogDirLayoutRenderer.cs, line 61 at r10](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiWsRJvo-LBtt-gERT6:-KiWsRJvo-LBtt-gERT7:bowtw2g) ([raw file](https://github.com/nlog/nlog/blob/3144c84c9448f5828e2ad57930fe95ff5845b683/src/NLog/LayoutRenderers/NLogDirLayoutRenderer.cs#L61)):*
> ```Smalltalk
>             var location = ReflectionHelpers.GetLocation(assembly);
>             if (string.IsNullOrEmpty(location))
>                 location = new Uri(ReflectionHelpers.GetCodeBase(assembly)).LocalPath;
> ```

aren't those ReflectionHelpers c# extensions?

---

*[tests/NLog.UnitTests/LogManagerTests.cs, line 537 at r10](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiWtH1O7MjVMNfNQmKt:-KiWtH1O7MjVMNfNQmKu:b-9rf6zw) ([raw file](https://github.com/nlog/nlog/blob/9d2adfe93a573d18cd2d82d6bb34e32d3996c774/tests/NLog.UnitTests/LogManagerTests.cs#L537)):*
> ```Smalltalk
>                 Parallel.ForEach(r, a =>
>                 {
>                     var res = c.Resolve<ClassA>();
> ```

I'm not sure if we test the same functionality in the new test? This was threadsafety in combination with DI (As far as I remember)

---

*[tests/NLog.UnitTests/Internal/EnumHelpersTests.cs, line 42 at r10](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiWu69yX_GQUy3IoW32:-KiWu69z3zUMc52sHuy6:b875gji) ([raw file](https://github.com/nlog/nlog/blob/9d2adfe93a573d18cd2d82d6bb34e32d3996c774/tests/NLog.UnitTests/Internal/EnumHelpersTests.cs#L42)):*
> ```Smalltalk
> namespace NLog.UnitTests.Internal
> {
>     public class EnumHelpersTests : NLogTestBase
> ```

I guess this is to allow running in parallel? If so, please add this as remarks thanks!

---

*[tests/NLog.UnitTests/Internal/PlatformDetectorTests.cs, line 39 at r10](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiWuGK7q-rDeQJKmTXB:-KiWuGK8NLi4jXcjxAT5:bxgocpn) ([raw file](https://github.com/nlog/nlog/blob/9d2adfe93a573d18cd2d82d6bb34e32d3996c774/tests/NLog.UnitTests/Internal/PlatformDetectorTests.cs#L39)):*
> ```Smalltalk
> namespace NLog.UnitTests.Internal
> {
>     public class PlatformDetectorTests : NLogTestBase
> ```

idem remarks

---

*[tests/NLog.UnitTests/Internal/StringBuilderExtTests.cs, line 41 at r10](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiWuJYfsnUqO4F_ihyH:-KiWuJYfsnUqO4F_ihyI:bxgocpn) ([raw file](https://github.com/nlog/nlog/blob/9d2adfe93a573d18cd2d82d6bb34e32d3996c774/tests/NLog.UnitTests/Internal/StringBuilderExtTests.cs#L41)):*
> ```Smalltalk
> namespace NLog.UnitTests.Internal
> {
>     public class StringBuilderExtTests : NLogTestBase
> ```

idem remarks

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/2055)*

304NotModified:
Great work! Some small questions!

---

Review status: 91 of 95 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/2055#-:-KiWvZnec8wDHym1-wDZ:bq425fd)*

snakefoot:
Thanks. Have tried to reply the questions.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/2055#-:-KiX7bgJAuQH1p0mQCVD:b4ah2yl)*

snakefoot:
Review status: 91 of 95 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[src/NLog/project.json, line 138 at r10](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiWv1ZBElXlGgI9ZS3U:-KiX3kOd4GJfKHdgmQFW:bj94rwm) ([raw file](https://github.com/nlog/nlog/blob/3144c84c9448f5828e2ad57930fe95ff5845b683/src/NLog/project.json#L138)):*
<details><summary><i>Previously, 304NotModified (Julian Verdurmen) wrote…</i></summary><blockquote>

x
</blockquote></details>

Removed the warning.

---

*[src/NLog/project.json, line 140 at r10](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiWv1D4qCn9_xhanZ_w:-KiX3j47e5fNvehxduTD:bj94rwm) ([raw file](https://github.com/nlog/nlog/blob/3144c84c9448f5828e2ad57930fe95ff5845b683/src/NLog/project.json#L140)):*
<details><summary><i>Previously, 304NotModified (Julian Verdurmen) wrote…</i></summary><blockquote>

> S0436
</blockquote></details>

Removed the warning.

---

*[src/NLog/Internal/ObjectGraphScanner.cs, line 92 at r10](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiWvF9vSC_jEDdJLYdm:-KiX6K8KcUob845fpaif:b3kre) ([raw file](https://github.com/nlog/nlog/blob/3144c84c9448f5828e2ad57930fe95ff5845b683/src/NLog/Internal/ObjectGraphScanner.cs#L92)):*
<details><summary><i>Previously, 304NotModified (Julian Verdurmen) wrote…</i></summary><blockquote>

x
</blockquote></details>

?

---

*[src/NLog/Internal/ObjectGraphScanner.cs, line 127 at r10](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiWrS9X2ObfTUrTBjiM:-KiX5TeE7IB6F542FI9-:b-96txd1) ([raw file](https://github.com/nlog/nlog/blob/3144c84c9448f5828e2ad57930fe95ff5845b683/src/NLog/Internal/ObjectGraphScanner.cs#L127)):*
<details><summary><i>Previously, 304NotModified (Julian Verdurmen) wrote…</i></summary><blockquote>

maybe we should move this to `IsDefined` as this is a helper already?

we should use also the generic one here I guess (see above)

`type.IsDefined<NLogConfigurationItemAttribute>`
</blockquote></details>

Well I like the InternalLogger-part.

---

*[src/NLog/Internal/TargetWithFilterChain.cs, line 119 at r10](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiWrzUzlKQn8sjgJxiD:-KiX5YLLagkhbbwzR5vk:b-aomujg) ([raw file](https://github.com/nlog/nlog/blob/3144c84c9448f5828e2ad57930fe95ff5845b683/src/NLog/Internal/TargetWithFilterChain.cs#L119)):*
<details><summary><i>Previously, 304NotModified (Julian Verdurmen) wrote…</i></summary><blockquote>

maybe a good candidate to move to StackTraceUsageUtils?
</blockquote></details>

I have little experience with StackTrace. And how to cater for calling the StackTrace constructor in a nested method. I would prefer that Activator.CreateInstance-hack was not used at all :)

---

*[src/NLog/LayoutRenderers/NLogDirLayoutRenderer.cs, line 61 at r10](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiWsRJvo-LBtt-gERT6:-KiX2eIKnEyAdVsoBErq:b-dqrat) ([raw file](https://github.com/nlog/nlog/blob/3144c84c9448f5828e2ad57930fe95ff5845b683/src/NLog/LayoutRenderers/NLogDirLayoutRenderer.cs#L61)):*
<details><summary><i>Previously, 304NotModified (Julian Verdurmen) wrote…</i></summary><blockquote>

aren't those ReflectionHelpers c# extensions?
</blockquote></details>

Yes better fix that :)

---

*[tests/NLog.UnitTests/LogManagerTests.cs, line 537 at r10](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiWtH1O7MjVMNfNQmKt:-KiX2jDt7MekyZBgXq7_:b1qlttu) ([raw file](https://github.com/nlog/nlog/blob/9d2adfe93a573d18cd2d82d6bb34e32d3996c774/tests/NLog.UnitTests/LogManagerTests.cs#L537)):*
<details><summary><i>Previously, 304NotModified (Julian Verdurmen) wrote…</i></summary><blockquote>

I'm not sure if we test the same functionality in the new test? This was threadsafety in combination with DI (As far as I remember)
</blockquote></details>

The original bug was about creating new loggers from multiple threads, while another thread is working on updating the NLog configuration. At least when looking a the callstack.

---

*[tests/NLog.UnitTests/Internal/EnumHelpersTests.cs, line 42 at r10](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiWu69yX_GQUy3IoW32:-KiX6BU3lcmCVcb2jxlW:bx2pmjw) ([raw file](https://github.com/nlog/nlog/blob/9d2adfe93a573d18cd2d82d6bb34e32d3996c774/tests/NLog.UnitTests/Internal/EnumHelpersTests.cs#L42)):*
<details><summary><i>Previously, 304NotModified (Julian Verdurmen) wrote…</i></summary><blockquote>

I guess this is to allow running in parallel? If so, please add this as remarks thanks!
</blockquote></details>

Yes unit tests not using NLog LoggingConfiguration, doesn't need to inherit from NLogTestBase.

---

*[tests/NLog.UnitTests/Internal/PlatformDetectorTests.cs, line 39 at r10](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiWuGK7q-rDeQJKmTXB:-KiX3XgXTwvGLKS9dTNt:bx2pmjw) ([raw file](https://github.com/nlog/nlog/blob/9d2adfe93a573d18cd2d82d6bb34e32d3996c774/tests/NLog.UnitTests/Internal/PlatformDetectorTests.cs#L39)):*
<details><summary><i>Previously, 304NotModified (Julian Verdurmen) wrote…</i></summary><blockquote>

idem remarks
</blockquote></details>

Yes unit tests not using NLog LoggingConfiguration, doesn't need to inherit from NLogTestBase.

---

*[tests/NLog.UnitTests/Internal/StringBuilderExtTests.cs, line 41 at r10](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiWuJYfsnUqO4F_ihyH:-KiWwAV8q71onwFS71cl:b-fbd5wb) ([raw file](https://github.com/nlog/nlog/blob/9d2adfe93a573d18cd2d82d6bb34e32d3996c774/tests/NLog.UnitTests/Internal/StringBuilderExtTests.cs#L41)):*
<details><summary><i>Previously, 304NotModified (Julian Verdurmen) wrote…</i></summary><blockquote>

idem remarks
</blockquote></details>

Well the NLogTestBase is mostly for resetting NLog-LoggingConfiguration. And in XUNIT2 protect against running multiple NLog-LoggingConfigurations in parallel. Tests not using the NLog-Core-Engine is allowed on run in parallel.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/2055)*

snakefoot:
*[src/NLog/Internal/TargetWithFilterChain.cs, line 119 at r10](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiWrzUzlKQn8sjgJxiD:-KiagAJpv27m8sk1yo5o:bvmsd06) ([raw file](https://github.com/nlog/nlog/blob/3144c84c9448f5828e2ad57930fe95ff5845b683/src/NLog/Internal/TargetWithFilterChain.cs#L119)):*
<details><summary><i>Previously, snakefoot (Rolf Kristensen) wrote…</i></summary><blockquote>

I have little experience with StackTrace. And how to cater for calling the StackTrace constructor in a nested method. I would prefer that Activator.CreateInstance-hack was not used at all :)
</blockquote></details>

You convinced me not to use the StackTrace constructor in UWP Apps, so skipped the validation. Now callsite only works using the NLog.Fluent callsite-attributes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/2055)*

304NotModified:
Reviewed 1 of 8 files at r9, 16 of 17 files at r11.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/NLog/Internal/ObjectGraphScanner.cs, line 92 at r10](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiWvF9vSC_jEDdJLYdm:-Kiaqkv9cdwL4YGAv-_P:b-nxkq5d) ([raw file](https://github.com/nlog/nlog/blob/3144c84c9448f5828e2ad57930fe95ff5845b683/src/NLog/Internal/ObjectGraphScanner.cs#L92)):*
<details><summary><i>Previously, snakefoot (Rolf Kristensen) wrote…</i></summary><blockquote>

?
</blockquote></details>

sorry X is also the shortcut of next file in reviewable.io ;)

---

*[src/NLog/Internal/ObjectGraphScanner.cs, line 127 at r10](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiWrS9X2ObfTUrTBjiM:-Kiar97uUxG6BM-TXR4E:bffgkvv) ([raw file](https://github.com/nlog/nlog/blob/3144c84c9448f5828e2ad57930fe95ff5845b683/src/NLog/Internal/ObjectGraphScanner.cs#L127)):*
<details><summary><i>Previously, snakefoot (Rolf Kristensen) wrote…</i></summary><blockquote>

Well I like the InternalLogger-part.
</blockquote></details>

sure but we could add that inside `prop.IsDefined`? (maybe also as warn?) We only miss the level indent I guess

---

*[src/NLog/LayoutRenderers/CallSiteLineNumberLayoutRenderer.cs, line 87 at r11](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiasEnLJ1pXWsU68bK8:-KiasEnLJ1pXWsU68bK9:bmpsxuc) ([raw file](https://github.com/nlog/nlog/blob/260478414ba040161746326217abd649ffb04cde/src/NLog/LayoutRenderers/CallSiteLineNumberLayoutRenderer.cs#L87)):*
> ```Smalltalk
>                 if (logEvent.Properties.TryGetValue("CallerLineNumber", out callerLineNumber) && callerLineNumber is int)
>                 {
>                     linenumber = (int)callerLineNumber;
> ```

any idea of `as` is more efficient here, even with ints? (boxing)

---

*[src/NLog/Internal/TargetWithFilterChain.cs, line 119 at r10](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiWrzUzlKQn8sjgJxiD:-KiaroTlci48JuvuEBvf:bgke9oz) ([raw file](https://github.com/nlog/nlog/blob/3144c84c9448f5828e2ad57930fe95ff5845b683/src/NLog/Internal/TargetWithFilterChain.cs#L119)):*
> I have little experience with StackTrace. And how to cater for calling the StackTrace constructor in a nested method. 

true about that. Missed that one. 

I think fluent is good enough for now indeed

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/2055)*

snakefoot:
*[src/NLog/LayoutRenderers/CallSiteLineNumberLayoutRenderer.cs, line 87 at r11](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiasEnLJ1pXWsU68bK8:-Kiau-KBCZNlh-6el_d6:b-u0ct29) ([raw file](https://github.com/nlog/nlog/blob/260478414ba040161746326217abd649ffb04cde/src/NLog/LayoutRenderers/CallSiteLineNumberLayoutRenderer.cs#L87)):*
<details><summary><i>Previously, 304NotModified (Julian Verdurmen) wrote…</i></summary><blockquote>

any idea of `as` is more efficient here, even with ints? (boxing)
</blockquote></details>

Usually "as" is much faster than "is". But since integer is a valuetype, then it has restrictions. I just wanted the code to be simple.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/2055)*

snakefoot:
*[src/NLog/Internal/ObjectGraphScanner.cs, line 127 at r10](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiWrS9X2ObfTUrTBjiM:-KiauqB2F9pJN7HMOK7E:b-iy0any) ([raw file](https://github.com/nlog/nlog/blob/3144c84c9448f5828e2ad57930fe95ff5845b683/src/NLog/Internal/ObjectGraphScanner.cs#L127)):*
<details><summary><i>Previously, 304NotModified (Julian Verdurmen) wrote…</i></summary><blockquote>

sure but we could add that inside `prop.IsDefined`? (maybe also as warn?) We only miss the level indent I guess
</blockquote></details>

Don't like use of the InternalLogger in the pure library methods. Might introduce unexpected noise/exceptions at an unwanted location.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/2055)*

304NotModified:
PS: fixed also SL build with 659a7990b65ad8f51f0239d3176f12d093ccbe7f

304NotModified:
https://github.com/snakefoot/NLog/pull/1

snakefoot:
@304NotModified Have applied the fixes, and now there is a green AppVeyor-build

304NotModified:
@snakefoot could you please change the version to beta07 (appveyor.yml)? Then it's easier to test (because beta06 is already on nuget)

snakefoot:
@304NotModified Have updated project.json with new version number (beta07)

304NotModified:
thanks @snakefoot !

Works well for me on asp.net core.

304NotModified:
Reviewed 5 of 5 files at r12, 3 of 3 files at r13, 1 of 1 files at r14.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/NLog/Internal/ObjectGraphScanner.cs, line 127 at r10](https://reviewable.io:443/reviews/nlog/nlog/2055#-KiWrS9X2ObfTUrTBjiM:-KiqCUQmzPdcOgQyX2A9:bb70v9n) ([raw file](https://github.com/nlog/nlog/blob/3144c84c9448f5828e2ad57930fe95ff5845b683/src/NLog/Internal/ObjectGraphScanner.cs#L127)):*
<details><summary><i>Previously, snakefoot (Rolf Kristensen) wrote…</i></summary><blockquote>

Don't like use of the InternalLogger in the pure library methods. Might introduce unexpected noise/exceptions at an unwanted location.
</blockquote></details>

OK 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/nlog/nlog/2055)*

304NotModified:
\0/

image

304NotModified:
summary of what has been changed in support:

Layout renderers:

- `${AssemblyVersion}` - netstandard 1.5
- `${machinename}` - now also netstandard 1.3 (was previously 1.5+)
- `${NLogDir}` - netstandard 1.5


targets:

- OutputDebugString  - netstandard 1.3

304NotModified:
updated https://github.com/NLog/NLog/wiki/platform-support

snakefoot:
@304NotModified These bonus updates are also included:

- ${windows-identity} : Enabled for NetStandard 1.3+
- Auto flush on shutdown : Enabled for NetStandard 1.5+
- File Archive Mutex : Enabled for NetStandard 1.3+
- Mutex File Appender : Enabled for NetStandard 1.3+

304NotModified:
Could you please update this one? https://github.com/nlog/nlog/wiki/platform-support

snakefoot:
@304NotModified Completed. The feature list looks very complete. Seems only a fully working GetCurrentClassLogger() is missing.

Guess one could consider to use the StackTrace-constructor-hack for NETSTANDARD1_3PLUS to capture the calling class.

304NotModified:
That would be nice. I prefer the hack over nothing. But the hack isn't working on all devices?

snakefoot:
@304NotModified StackTrace-constructor-hack will not work for .NET Native, not sure whether the hidden (internal) constructor works on non-windows platforms. Think it is better leave the possible implementation of the hack to someone who is more motivated for fixing and testing it.

Or just wait for Q3 2017 where NetStandard 2.0 is released :)

Issue #2062, opened 2017-04-19
Unicode text gets escaped in JSON layout


biiiipy:
**Type**: Bug
**NLog version**: 4.5.0-alpha04
**Platform**: .Net 4.5

Unicode characters get encoded in JSON layout:
https://github.com/NLog/NLog/blob/master/src/NLog/Targets/DefaultJsonSerializer.cs#L274

This block is not necessary (at least in a UTF8 encoded file target), cause we already escaped control characters in switch statement:
```
if (EscapeChar(ch))
{
     sb.AppendFormat(CultureInfo.InvariantCulture, "\\u{0:x4}", (int)ch);
}
```

My setup:
```
JsonLayout jsonLayout = new JsonLayout()
{
     SuppressSpaces = true
};
jsonLayout.Attributes.Add(new JsonAttribute("msg", @"${message}"));
```

Logging Copyright © symbol: 
``Log.Fatal("©");``
Log output: 
``{"msg":"\u00a9"}``
Expected output: 
``{"msg":"©"}``

Q1: Is there any workaround?
Q2: What's the status with using JSON.Net for serializing?

304NotModified:
thanks for the bug report!

304NotModified:
This will be fixed in 4.4.7 and soon also in 4.5 alpha 5

304NotModified:
fixed in beta 4.5

PR #2063, opened 2017-04-19
JsonLayout - Added JsonAttribute property EscapeUnicode

snakefoot:
Fix for #2062. Allow one to specify on each JsonAttribute, whether it should escape unicode-characters (non ascii).


304NotModified:
added to docs https://github.com/NLog/NLog/wiki/JsonLayout

PR #2067, opened 2017-04-21
${EventProperties}, ${newline}, ${basedir} & ${tempdir} as ThreadAgnostic

snakefoot:
Allow to defer rendering on the async-thread.


304NotModified:
Thanks!

304NotModified:
`System.Diagnostics.Process.GetCurrentProcess()` is also threadagnostic? I think so because all the threads has the same process? 

PS: updated the title of the PR to make it more complete.

PR #2069, opened 2017-04-21
LayoutRenderer Initialize should swallow exceptions when configured

snakefoot:
Breaking change, should be moved to NLog ver. 5.0

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/2069)


c0shea:
Isn't this a breaking change considering `NLogConfigurationException` is not derived from the existing `ArgumentException` and an exception is now being thrown where one was never thrown before?

snakefoot:
> Isn't this a breaking change 

Certainly is. Just me being lazy. Have now updated the description.

304NotModified:
Wrong button? ;)

snakefoot:
More or less. Just had so many stale branches, that the pending ones was mixed together :)

PR #2075, opened 2017-04-24
StackTraceLayoutRenderer with Raw format should display source FileName

### LINKED ISSUE #2074: Stacktrace doesn't show file lines and file names
**Type**: Bug

**NLog version**:   4.4.6

**Platform**: .Net 4.5

**Current NLog config**

```xml
<nlog>
  <targets>
    <target name="logfile" xsi:type="File" fileName="file.txt" layout="${stacktrace:format=raw}"/>
  </targets>
  <rules>
    <logger name="*" minlevel="Trace" writeTo="logfile" />
  </rules>
</nlog>

```
My current log with that configuration is:
_ExecuteAssembly at offset 127 in file:line:column <filename unknown>:0:0
nExecuteAssembly at offset 0 in file:line:column <filename unknown>:0:0
Main at offset 710 in file:line:column <filename unknown>:0:0_

I expect filelines and filenames to be in logfile.
I tried to do same thing with .NET diagnostic (something like _var frames = new StackTrace(true).GetFrames();_) and all filenames and filelines were showed correctly.

Project has full output debug info, all pdb-files are on their place.

I also tried to do that on NLog version 4.2.3 and received same result - no filelines and filenames.

snakefoot:
Attempt to fix #2074

### DIFF
--- a/tests/NLog.UnitTests/LayoutRenderers/StackTraceRendererTests.cs
+++ b/tests/NLog.UnitTests/LayoutRenderers/StackTraceRendererTests.cs
@@ -99,8 +99,9 @@ namespace NLog.UnitTests.LayoutRenderers
 
             AssertDebugLastMessageContains("debug", "RenderStackTrace_raw at offset ");
             AssertDebugLastMessageContains("debug", "RenderMe at offset ");
-            AssertDebugLastMessageContains("debug", "<filename unknown>");
-
+#if !MONO
+            AssertDebugLastMessageContains("debug", "StackTraceRendererTests.cs");
+#endif
         }
 
         [Fact]
@@ -117,7 +118,6 @@ namespace NLog.UnitTests.LayoutRenderers
             RenderMe("I am:");
             AssertDebugLastMessageContains("debug", " => [Void RenderStackTrace_DetailedFlat()] => [Void RenderMe(System.String)]");
         }
-
 
         [MethodImpl(MethodImplOptions.NoInlining)]
         private void RenderMe(string message)
--- a/src/NLog/LayoutRenderers/StackTraceLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/StackTraceLayoutRenderer.cs
@@ -37,7 +37,6 @@ namespace NLog.LayoutRenderers
     using System.Diagnostics;
     using System.Text;
     using NLog.Config;
-
     using NLog.Internal;
 
     /// <summary>
@@ -91,7 +90,7 @@ namespace NLog.LayoutRenderers
         /// <value></value>
         StackTraceUsage IUsesStackTrace.StackTraceUsage
         {
-            get { return StackTraceUsage.WithoutSource; }
+            get { return (Format == StackTraceFormat.Raw) ? StackTraceUsage.Max : StackTraceUsage.WithoutSource; }
         }
 
         /// <summary>
@@ -117,7 +116,6 @@ namespace NLog.LayoutRenderers
                         StackFrame f = logEvent.StackTrace.GetFrame(i);
                         builder.Append(f.ToString());
                     }
-
                     break;
 
                 case StackTraceFormat.Flat:
@@ -144,7 +142,6 @@ namespace NLog.LayoutRenderers
                         builder.Append(f.GetMethod().Name);
                         first = false;
                     }
-
                     break;
 
                 case StackTraceFormat.DetailedFlat:
@@ -161,7 +158,6 @@ namespace NLog.LayoutRenderers
                         builder.Append("]");
                         first = false;
                     }
-
                     break;
             }
         }


tests/NLog.UnitTests/LayoutRenderers/StackTraceRendererTests.cs
```diff
@@ -118,6 +118,27 @@ public void RenderStackTrace_DetailedFlat()
...
+            <nlog>
+                <targets><target name='debug' type='Debug' layout='${message} ${stacktrace:format=RawWithSource}' /></targets>
+                <rules>
+                    <logger name='*' minlevel='Debug' writeTo='debug' />
+                </rules>
+            </nlog>");
+
+            RenderMe("I am:");
+
+            AssertDebugLastMessageContains("debug", "RenderStackTrace_RawWithSource at offset ");
```
304NotModified: not sure if this a useful test now. It can be still ` file:line:column <filename unknown>:0:0  ` ?

src/NLog/LayoutRenderers/StackTraceFormat.cs
```diff
@@ -52,5 +52,10 @@ public enum StackTraceFormat
         /// Detailed flat format (method signatures displayed in a single line).
         /// </summary>
         DetailedFlat,
+
+        /// <summary>
+        /// Raw format (multiline with source info - as returned by StackFrame.ToString() method).
+        /// </summary>
+        RawWithSource,
```
304NotModified: I'm doubting if we should patch `raw` instead of a new option (I know i proposed it earlier ;))

Printing ` file:line:column <filename unknown>:0:0  ` sounds useless to me and the default is `flat`. 

What do you think?

304NotModified:
Thanks! I was thinking to create this, but you we're faster :)

PS: please do use the `fixes` syntax. Thanks!

tests/NLog.UnitTests/LayoutRenderers/StackTraceRendererTests.cs
snakefoot: Well it is combination of the 3 asserts, that all verifies the last logged message. See the original RenderStackTrace_Raw, which does the same.

304NotModified: yeah but should we not test that line number >0 ? 

Need a regex I guess

src/NLog/LayoutRenderers/StackTraceFormat.cs
snakefoot: I'm not using this LayoutRenderer, so fixing the existing Raw-format is fine with me.

tests/NLog.UnitTests/LayoutRenderers/StackTraceRendererTests.cs
snakefoot: Well it is enough for me that it was able to capture the filename (StackTraceRendererTests.cs). Not wanting to dive into the StackTrace-formating abyss.

304NotModified: something like this?

```
string pattern = @"<(?<filename>.*?)>:(?<line>\d+):(?<col>\d+)";
```
https://regex101.com/r/JJ1tUk/1

snakefoot: I'm not committing any regex funkyness. Just close this PR, and create you own :)

src/NLog/LayoutRenderers/StackTraceFormat.cs
304NotModified: I think fix "raw" is more clear :) 

So remove and we need to document the change in Nlog 4.4.7.

tests/NLog.UnitTests/LayoutRenderers/StackTraceRendererTests.cs
304NotModified: am I allowed to push on this branch?

snakefoot: "Allow edits from maintainers" is checked, so I guess you can knock yourself wild :)

tests/NLog.UnitTests/LayoutRenderers/StackTraceRendererTests.cs
```diff
@@ -99,8 +99,9 @@ public void RenderStackTrace_raw()
 
             AssertDebugLastMessageContains("debug", "RenderStackTrace_raw at offset ");
             AssertDebugLastMessageContains("debug", "RenderMe at offset ");
-            AssertDebugLastMessageContains("debug", "<filename unknown>");
-
+#if !MONO
+            AssertDebugLastMessageContains("debug", "StackTraceRendererTests.cs");
```
304NotModified: 👍

304NotModified:
updated docs for this: https://github.com/NLog/NLog/wiki/Stack-Trace-Layout-Renderer

PR #2093, opened 2017-05-05
Fixed bug to logging by day of week [Исправлена ошибка с логированием…

RussianDragon:
Fixed bug to logging by day of week.

Exemple config
```
<target name="testName" xsi:type="File" fileName="C:/log/${logger}.log"
			layout="${message}"
      archiveFileName="C:/log/${logger}_{##}.log" archiveNumbering="Date" archiveEvery="Monday" maxArchiveFiles="10" archiveDateFormat="yyyy-MM-dd-HH-mm"
			concurrentWrites="true" keepFileOpen="false" encoding="UTF-8"/>
<logger name="testLogger" writeTo="testName" levels="Trace,Debug,Info,Warn,Error,Fatal" final="true"/>
```
With this configuration, logs are archived every day, but I expect a new archive file every week.

I fix this bug.

### DIFF
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -2272,13 +2272,19 @@ namespace NLog.Targets
                 case FileArchivePeriod.Minute:
                     return input.AddTicks(-(input.Ticks % TimeSpan.TicksPerMinute));
                 case FileArchivePeriod.Sunday:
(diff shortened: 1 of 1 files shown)


304NotModified:
fixes https://github.com/NLog/NLog/pull/2093

304NotModified:
@RussianDragon

any idea how we could unit test this? So we won't break it in the future?

304NotModified:
unit test could be later :)

PR #2096, opened 2017-05-05
Fixes #2045 by moving rule elements of List<NLogXmlElement> children …

### LINKED ISSUE #2045: In the config file, order of targets and rules blocks may cause errors
I'm new to NLog and just went through a process of setting up NLog within a project of mine. I couldn't get NLog to properly load for awhile. I eventually got NLog to generate the internal log file and got an error saying that NLog could not locate a target of mine. I looked at my config and noticed that my `targets` block was below my `rules` block. So, I swapped it so that `targets` was now above `rules` and everything started working as intended. This remained the case regardless of whether or not I used a stand alone `NLog.config` file or the `app.config` file.

I'm unsure of how NLog loads the config file, but order of these blocks interfering with things seems odds to me. I think that this is either a bug or needs to be mentioned in the config file documentation, if it's not already.

Moafak:
in XmlLoggingConfiguration.ParseNLogElement() method
in switch statement, case "RULES"
the method ParseRulesElement() 
calls ParseLoggerElement()
which has a call to FindTargetByName()

That's why rule wouldn't have target if rules are loaded first

To fix that, rearrange the List<NLogXmlElement> so that targets are handled before rules in the switch statement

From my understanding, I think rules should be loaded last

### DIFF
--- a/src/NLog/Config/XmlLoggingConfiguration.cs
+++ b/src/NLog/Config/XmlLoggingConfiguration.cs
@@ -619,6 +619,13 @@ namespace NLog.Config
                 this.ParseExtensionsElement(extensionsChild, Path.GetDirectoryName(filePath));
             }
 
+            //target elements should be parsed before rule elements, as target elements are referenced by rule elements
+            var targetChilds = children.Where(child => child.LocalName.Equals("TARGETS", StringComparison.InvariantCultureIgnoreCase)).ToList();
+            foreach (var targetChild in targetChilds)
+            {
+                this.ParseTargetsElement(targetChild);
+            }
+
             //parse all other direct elements
             foreach (var child in children)
             {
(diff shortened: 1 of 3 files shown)


304NotModified:
Thanks!

I think this is maybe a better way (performance wise)

https://github.com/Moafak/NLog/blob/5a405bd2cabac667a29dfb7737021da79f6a626f/src/NLog/Config/XmlLoggingConfiguration.cs#L616

and 

https://github.com/Moafak/NLog/blob/5a405bd2cabac667a29dfb7737021da79f6a626f/src/NLog/Config/XmlLoggingConfiguration.cs#L628

Could you also add an unit test? Thanks!

Moafak:
Thanks
I will update the code.

Would you please tell me what the new case I should write the unit test for?

Moafak:
@304NotModified 
Is the unit test good enough?

304NotModified:
Unit test looks good, but code is not correctly updated I think? Also build fails, could you check that?

304NotModified:
I have updated the code into https://github.com/NLog/NLog/pull/2113, thanks for the submit

PR #2100, opened 2017-05-09
MailTarget - Flush should wait for SmtpClient.SendCompleted

### LINKED ISSUE #2065: How to get stack trace also with exceptions through email

**Type** (choose one):
- Question

**NLog version**: 4.4.6

**Platform**: .Net 4.5

```xml
<targets async="true">
    <target name="mail" xsi:type="Mail" html="true" replaceNewlineWithBrTagInHtml="true" smtpServer="-removed-" smtpAuthentication="None" body="${exception:innerFormat=Message,StackTrace}" to="-removed-" from="-removed-" subject="${event-properties:item=Task} for ${event-properties:item=ClientName} has FAILED on ${machinename}"/>
</targets>
<rules>
    <logger name="*" minlevel="Fatal" writeTo="mail"/>
</rules>
```
I have tried different values for body like `${message}, ${exception:format=toString},${exception:innerFormat=Message,StackTrace}"`  etc but in the mail all i get is exception message and parameter name for argumentnullexception. I want to show the stack trace too.
My C# code is as below:
```c#
catch (Exception ex) {
  LogEventInfo logEventInfo = new LogEventInfo(LogLevel.Fatal, logger.Name, null, "", null, ex);
  logEventInfo.Properties["ClientName"] = "Name";
  logEventInfo.Properties["Task"] ="Test";
  logger.Log(logEventInfo);
}
```
This is what i get now:
image

And this how i would prefer:
image

snakefoot:
Attempt to fix #2065

### DIFF
--- a/src/NLog/Internal/ISmtpClient.cs
+++ b/src/NLog/Internal/ISmtpClient.cs
@@ -84,6 +84,11 @@ namespace NLog.Internal
         /// Gets or sets the folder where applications save mail messages to be processed by the local SMTP server.
         /// </summary>
         string PickupDirectoryLocation { get; set; }
+
+        /// <summary>
+        /// Occurs when an asynchronous e-mail send operation completes.
+        /// </summary>
+        event SendCompletedEventHandler SendCompleted;
     }
 }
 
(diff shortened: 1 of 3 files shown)


snakefoot:
Trying to fix the wrong problem. Since using SmtpClient.Send and not SendAsync.

PR #2101, opened 2017-05-09
LogFactory.Shutdown - Add warning on target flush timeout

snakefoot:
Added InternalLogger-Warning when LogFactory.Shutdown does not flush succesfully. See also #2065


304NotModified:
thanks!

PR #2103, opened 2017-05-10
Improve Install of targets / crash Install on Databasetarget.

### LINKED ISSUE #1098: Improve Install of targets / crash Install on Databasetarget.
Calling `Install` on a databasetarget leads to a nullrefexception (https://github.com/NLog/NLog/issues/1094). The work around is to call `Initialize` on the `LoggingConfiguration`, as seen on  [InstallNLogConfig console application](https://github.com/NLog/NLog/blob/master/src/InstallNLogConfig/Program.cs)

A public method should never throw this kind of exception. We have to find out why `Install` isn't calling `Initialize`.

Steps:
- [ ] check if other targets won't call `Initialize` in `Install`
- [ ] discuss the impact of call `Initialize` in `Install`
- [ ] OR fix the nullref, OR call `Initialize` in `Install`

M4ttsson:
My first PR so I hope I got everything right.
Fixes #1098

### DIFF
--- a/src/NLog/Targets/DatabaseTarget.cs
+++ b/src/NLog/Targets/DatabaseTarget.cs
@@ -362,27 +362,35 @@ namespace NLog.Targets
 
             if (!foundProvider)
             {
-                switch (this.DBProvider.ToUpper(CultureInfo.InvariantCulture))
-                {
-                    case "SQLSERVER":
-                    case "MSSQL":
-                    case "MICROSOFT":
-                    case "MSDE":
-                        this.ConnectionType = systemDataAssembly.GetType("System.Data.SqlClient.SqlConnection", true);
-                        break;
+                this.SetConnectionType();
+            }
+        }
 
-                    case "OLEDB":
-                        this.ConnectionType = systemDataAssembly.GetType("System.Data.OleDb.OleDbConnection", true);
-                        break;
+        /// <summary>
+        /// Set the <see cref="ConnectionType"/> to use it for opening connections to the database.
+        /// </summary>
+        private void SetConnectionType()
+        {
+            switch (this.DBProvider.ToUpper(CultureInfo.InvariantCulture))
+            {
+                case "SQLSERVER":
+                case "MSSQL":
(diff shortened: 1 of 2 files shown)


M4ttsson:
Btw, should my unit test run in Travis? I assumed that the other sql test that I "took inspiration" from is not run in Travis for a reason so I used the same if-case to disable it.

304NotModified:
**thanks!**

AFAIK Travis doesn't have a SQL service?  If it has, it would be great to enable (I'm an other pull request ;)

304NotModified:
More info is here: https://docs.travis-ci.com/user/database-setup/

304NotModified:
Thanks! this a great improvement. 

But I was thinking, isn't easier to do something like this?

```c#
public void Install(InstallationContext installationContext) {
  if(!IsInitialized)
  { 
     InitializeTarget(); 
  }
  ...
}

```

M4ttsson:
Yes, it would. I was concerned about any side effects of calling the base.InitializeTarget() that contains calls to a FindAllLayouts() method. I guess it would cause the layouts to be loaded? 
I could change it if you want? I agree that the code would look better that way.

304NotModified:
Indeed the side effect is less clear. 

So for now this is good enough, thanks!

M4ttsson:
Awesome. I think I'll look for another issue to do now. Or should I maybe create an mysql version of at least the install tests so it can be tested in Travis too?

304NotModified:
> Or should I maybe create an mysql version of the install tests so it can be tested in Travis too?

That would be great also to have!

304NotModified:
Fixed in 4.4.10 and the release is queued. Will be online in a few hours!

Thanks for the submit!

M4ttsson:
Awesome. 
I might have some time tomorrow to look at creating the mysql install test we mentioned. Should I just do it and create a pull request or how should I do it?

304NotModified:
Just send a PR. Would be great to have!

PR #2113, opened 2017-05-16
BugFix: <targets> after <rules> won't work

### LINKED ISSUE #2045: In the config file, order of targets and rules blocks may cause errors

304NotModified:
Fixes #2045

### DIFF
--- a/src/NLog/Config/XmlLoggingConfiguration.cs
+++ b/src/NLog/Config/XmlLoggingConfiguration.cs
@@ -614,6 +614,8 @@ namespace NLog.Config
                 this.ParseExtensionsElement(extensionsChild, Path.GetDirectoryName(filePath));
             }
 
+            var rulesList = new List<NLogXmlElement>();
+
             //parse all other direct elements
             foreach (var child in children)
             {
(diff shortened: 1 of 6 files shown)


304NotModified:
supersedes https://github.com/NLog/NLog/pull/2096

snakefoot:
Maybe change the logic, so it always parses elements in this order (and not just targets first):

```
case "EXTENSIONS":
case "INCLUDE":
case "TIME":
case "VARIABLE":
case "APPENDERS":
case "TARGETS":
case "RULES":
```

304NotModified:
Indeed this was too Much quick fix

snakefoot:
Easy fix would be to always parse "RULES" last :)

304NotModified:
Have to look into this. Where variables and include is defined, does matter.

304NotModified:
> Easy fix would be to always parse "RULES" last :)

That's indeed a much easier approach -  keep it simple :)

changed!

304NotModified:
Fixed in 4.4.10 and the release is queued. Will be online in a few hours!

PR #2114, opened 2017-05-16
EventlogTarget: Support for MaximumKilobytes

304NotModified:
supersedes #1810 

thanks @ajitpeter !

### DIFF
--- a/src/NLog/Targets/EventLogTarget.cs
+++ b/src/NLog/Targets/EventLogTarget.cs
@@ -87,6 +87,7 @@ namespace NLog.Targets
             this.Log = "Application";
             this.MachineName = ".";
             this.MaxMessageLength = 16384;
+            this.maxKilobytes = 512;
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


snakefoot:
Could the maxKilobytes-member become nullable with default value null? And only assigned to the EventLog-object when assigned? 

```
        [DefaultValue(512)]
        public long MaxKilobytes
        {
            get { return this.maxKilobytes ?? 512; }
            set
            {   //Event log API restriction
                if (value < 64 || value > 4194240 || (value % 64 != 0))
                    throw new ArgumentException("MaxKilobytes must be a multitude of 64, and between 64 and 4194240");
                this.maxKilobytes = value;
            }
        }
        private long? maxKilobytes;

        if (this.maxKilobytes.HasValue)
                eventLogInstance.MaximumKilobytes = this.MaxKilobytes.Value;
        return eventLogInstance;
```

Not all services are running with full admin priviledges and can modify everything.

304NotModified:
Was also doubting for the nullable. But what's the benefit? We need admin rights to set the value?

304NotModified:
Ow guess I needed to read "security" in https://msdn.microsoft.com/en-us/library/system.diagnostics.eventlog.maximumkilobytes(v=vs.110).aspx

Issue #2127, opened 2017-05-19
JsonLayout with objects should serialize the object (instead of stringify)


304NotModified:
Objects in the logevent properties are currently stringified.

E.g. a dictionary will be:

```
System.Collections.Generic.Dictionary`2[System.String,System.Object]
```

instead of 

```
{ "key1": "value1", "key2": { "nestedObject": "yes"}}
```

see https://github.com/NLog/NLog/issues/723#issuecomment-302522016

related #2122 

The current Json serializer can't serialize objects

304NotModified:
I think fixed in 4.5. Need to test.

PR #2128, opened 2017-05-19
JSON serializer: support objects, improvements and formatting, add memoryCache

304NotModified:
Extends the default Json Serializer

- number formating with formatting string
- support serializing objects (with reflection)
- support timespan and guid
- serializer options object
- some method renames (only internals)
- performance tweaks
- more unit tests

Needed for #1376 and #2127 

todo:
- [x]  make backwardscomp
- [x] remove inner class
- [x] remove json5 support for now
- [x] quote key
- [x] option to disable quote key  -> for structured logging, so don't enable by default
- [x] is formatting really needed? -> for structured logging, so don't enable by default
- [x] props as real caching instead of static?
- [x] escape quotes
- [x] update commit text
- [x] check enum

### DIFF
--- a/src/NLog/Internal/Cache.cs
+++ b/src/NLog/Internal/Cache.cs
@@ -31,7 +31,7 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
-#if NET3_5 || NET4_0 || SILVERLIGHT || __IOS__ || __ANDROID__ || WINDOWS_PHONE
+#if NET3_5 || NET4_0 || SILVERLIGHT || __IOS__ || __ANDROID__ || WINDOWS_PHONE || MONO
 #define NO_MEMORY_CACHE
 using System.Collections.Generic;
 #else
--- a/src/NLog/Internal/Cache.cs
+++ b/src/NLog/Internal/Cache.cs
@@ -159,7 +159,20 @@ namespace NLog.Internal
         {
             // ReSharper disable InconsistentlySynchronizedField
 #if NO_MEMORY_CACHE
-            return _cache[key] as T;
+            try
+            {
+                T value;
+                if (_cache.TryGetValue(key, out value))
+                {
+                    return value;
+                }
+            }
+            catch (Exception ex)
+            {
+                InternalLogger.Debug(ex, "Exception when getting value with key '{0}'", key);
+            }
+            return null;
+
 #else
             return _cache.Get(key) as T;
             // ReSharper restore InconsistentlySynchronizedField
--- a/src/NLog/Internal/Cache.cs
+++ b/src/NLog/Internal/Cache.cs
@@ -1,0 +1,169 @@
+﻿// 
+// Copyright (c) 2004-2016 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without 
+// modification, are permitted provided that the following conditions 
+// are met:
+// 
+// * Redistributions of source code must retain the above copyright notice, 
+//   this list of conditions and the following disclaimer. 
+// 
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution. 
+// 
+// * Neither the name of Jaroslaw Kowalski nor the names of its 
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission. 
+// 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE 
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE 
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE 
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR 
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS 
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN 
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) 
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF 
+// THE POSSIBILITY OF SUCH DAMAGE.
+// 
+
+#if NET3_5 || SILVERLIGHT || __IOS__ || __ANDROID__ || WINDOWS_PHONE
+#define NO_MEMORY_CACHE
+using System.Collections.Generic;
+#else
+using System.Runtime.Caching;
+#endif
+
+using System;
+using NLog.Common;
+
+namespace NLog.Internal
+{
+    /// <summary>
+    /// Cache values, create the cache and store it static!
+    /// 
+    /// Note: <c>null</c> is not allowed as cache value.
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    internal class Cache<T> where T : class
+    {
+        /// <summary>
+        /// Unique name, for logging and separating cache
+        /// </summary>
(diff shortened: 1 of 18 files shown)


snakefoot:
PropCache should support multithreaded access. If using Dictionary then make a clone and throw the old dictionary away instead of modifying the existing dictionary.

304NotModified:
> PropCache should support multithreaded access. If using Dictionary then make a clone and throw the old dictionary away instead of modifying the existing dictionary.

Well in the worst case the props are lost? (lost update), That isn't a problem as it as "cache" imo. 

> Maybe have a parameter to control whether it should call ToString() on custom-object or if it should enumerate the customer-object-properties.

Yes, will add it. Was thinking to add a serializationProperties objects, so we could easily extend it in the future. 

ALso thinking to make a new interface ISerializer (in the correct namespace) and obsolete the current one. 

> Maybe also recognize IEnumerable<KeyValuePair> so they are translated more optimal.

Checking generic interfaces is AFAIK expensive. So doubting about this.

snakefoot:
> Well in the worst case the props are lost? (lost update), That isn't a problem as it as "cache" imo.

In the worst case you will get an exception during TryGetValue because another thread modifying the Dictionary-object. But not if using the clone-update-method.

304NotModified:
could also add a try catch ;)

cloning and locking is expensive :|

snakefoot:
> could also add a try catch ;)

Yes that is also a solution, if we are certain Dictionary never comes into a bad state after the exception (Two threads inserting at the same time).

> cloning and locking is expensive :|

Locking is expensive, but cloning is only expensive in the beginning where object-types are being discovered.

304NotModified:
AFAIK the concurrent dictionary is even slower.   ;)

304NotModified:
Related info: https://www.codeproject.com/Articles/548406/Dictionary-plus-Locking-versus-ConcurrentDictionar

snakefoot:
> AFAIK the concurrent dictionary is even slower. ;)

Yes the ConcurrentDictionary is slower than Dictionary without locks. But if it is faster than Dictionary with lock, when doing lots of reads. Just like the article says:

> It is optimized to allow lock-free reading

> if you do much more reads than writes you will see that the ConcurrentDictionary is still faster

304NotModified:
FYI, when tracing NLog before I found that the cache in the layout renderer (dictionary) was one of time consuming the parts.

On topic. I'm not sure if I need locks. Just try catch.

304NotModified:
Also I didn't finished reading the article ;)

snakefoot:
Maybe also add `DateTimeOffset` to this line:

`if (value is Guid || value is TimeSpan)`

Maybe consider to use this method instead of just calling ToString():

`Convert.ToString(value, System.Globalization.CultureInfo.InvariantCulture);`

304NotModified:
Thanks!

snakefoot:
Could you not change the _propsCache into a _typeSerializerCache, where the serializer for a type is saved. Then you could also put the serializers for TimeSpan, Guid and DateTimeOffset into this cache at startup. 

The same cache can then also be used for saving the KeyValuePair-types that you might meet. Then there will only be an overhead the first time you meet a Type not seen before.

304NotModified:
Do you mean to cache the whole `Type` instead of the `PropertyInfo[]` ?

Ps. Improving the cache is still on the list. Gonna check the `MenoryCache` (and some replacement for net35)

snakefoot:
> Do you mean to cache the whole Type instead of the PropertyInfo[] ?

Right it is declared like this:

`private static Dictionary<Type, PropertyInfo[]> _propsCache = new Dictionary<Type, PropertyInfo[]>(); `

Instead of mapping to `PropertyInfo[]`, then you could map to a function that can serialize the type. This delegate could then have captured PropertyInfo[] for its usage. Then you can register custom functions for custom Types like TimeSpan, Guid and DateTimeOffset (And maybe also KeyValuePair)

304NotModified:
Ah yes cache at a higher level. :) will check that. Thx!

snakefoot:
> Improving the cache is still on the list. Gonna check the MenoryCache (and some replacement for net35)

"Advanced" caches like the MemoryCache usually introduces the ability to evict items that have not been used for a long time (Using timers and other exciting stuff). When an object-serializer has been born, do you plan to remove it from the cache again when not used for some time?

Think Dictionary and ConcurrentDictionary are excellent for doing caching.

304NotModified:
> "Advanced" caches like the MemoryCache usually introduces the ability to evict items that have not been used for a long time (Using timers and other exciting stuff). When an object-serializer has been born, do you plan to remove it from the cache again when not used for some time?

Yes, but then I should no use a dictionary in a memoryCache I guess (which is maybe even better)

304NotModified:
@snakefoot the MemoryCache is using [string as key](https://msdn.microsoft.com/en-us/library/dd988741(v=vs.110).aspx). I don't think It correct to use `Type.FullName` as key? What do you think?

snakefoot:
I think MemoryCache is overkill. Dictionary is excellent for caching on its own. I would consider having a factory and give each layout their own instance of the serialiser. This will remove the multi threaded issue and also allow one to have a reusable string builder. It should be possible to register custom object serialiser in the factory so new instances born knows them. Nlog configuration reload should release and acquire new serialiser instances



Sent from my Samsung device


304NotModified:
> I think MemoryCache is overkill.

not afraid that memory is never released? 

> I would consider having a factory and give each layout their own instance of the serialiser.

So also their own cache?

> This will remove the multi threaded issue and also allow one to have a reusable string builder.

I don't see that, but I think you're right ;)

snakefoot:
Unless the application is planning to log an infinite amount of different types, then I am not afraid of the cache size

Yes an instance for each, also means a cache for each.


Sent from my Samsung device


304NotModified:
@snakefoot I like to merge this one, would you like to review it?

Some things I will add in the future:

- [ ] readable date format
- [ ] better stringbuilder usage, lazyStringBuilder type?
- [ ] force stringify

I have added a caching class and a simple one for .NET 3.5, otherwise I don't feel comfortable to release it.

snakefoot:
Still not sure why you are using MemoryCache. Is there a need for cache expiry? Have you tested performance between Dictionary and MemoryCache?

Remember a try-catch for the unprotected dictionary-lookup without lock, as it can throw exceptions.

304NotModified:
> Still not sure why you are using MemoryCache. Is there a need for cache expiry?

Static variable as cache could lead to memory leaks, as it's never removed from memory. The current max of 1000 is really arbitrary  - systems with more memory usage could how more and systems with less, less. 

Also MemoryCache is specially built for this, so I think it's the best choice.

snakefoot:
> Also MemoryCache is specially built for this, so I think it's the best choice.

Think it is a strange to have a cache that generates garbage (MemoryCacheKey) for every item-lookup :)

304NotModified:
it's shortliving garbage, isn't it/

snakefoot:
Any reason why you need a static-global-cache? Why not have a serializer-factory (that can be exchanged) that can create serializer-instance where needed. Then have multiple json-serializers. Each serializer can then have a lock-object that allows them to have a reusable-string-builder, and their own local type-cache. The serializer-instance can then be created on initialize of target or layout, and discarded on dispose of target or layout.

snakefoot:
Looks like there are issues with the MemoryCache, when using the "new" concurrent garbage-collection:

https://stackoverflow.com/questions/5655439/net-4-memorycache-leaks-with-concurrent-garbage-collection

304NotModified:
> Any reason why you need a static-global-cache?

Well the cache will be less effective when it's not global.

snakefoot:
> Well the cache will be less effective when it's not global.

Not sure I understand. How big are these objects in the cache? Are you worried about the initial prime of the cache?

304NotModified:
> Looks like there are issues with the MemoryCache, when using the "new" concurrent garbage-collection:
> 
> https://stackoverflow.com/questions/5655439/net-4-memorycache-leaks-with-concurrent-garbage-collection

mmm that's bad indeed

snakefoot:
Instead of doing cache-clear when it reaches a 1000 items, then you could maintain a Most-Recently-Used-Cache:

https://stackoverflow.com/questions/22002814/is-this-algorithm-implementation-lru-or-mru

Would probably make the following modifications:
- Change MruItem into a struct (instead of class) that has the members Value + LinkedListNode<TKey> (Avoid unnecessary cache-line-misses after dictionary lookup).
- Change LinkedList into just holding the TKey (instead of MruItem)
- TryGetValue should check if the found item is already located first in the LinkedList, before doing Remove / AddFirst.
- Now instead of clearing the entire cache, then you can just remove the least recently used items.

Also the JsonSerializer should not perform constant lookups into the cache when serializing a list or dictionary of objects. Instead it should perform cache-lookup once, and reuse that for all items in the collection (ofcourse with fallback to cache-lookup if the collection have many different types of objects).

304NotModified:
if you have time, please go ahead.

Unfortunately less time these weeks, busy on a new house, planning, painting, moving etc. Not sure if I have time tomorrow. 

I could also remove the cache, and then we could create a new PR

304NotModified:
superseded by https://github.com/NLog/NLog/pull/2179

PR #2163, opened 2017-06-12
JsonLayout - IncludeMdc and IncludeMdlc are not ThreadAgnostic

snakefoot:
Kind of a special hack. Maybe it is better to make a virtual-method that gets the object-graph as input parameter. Then JsonLayout can override this an check its IncludeMdc and IncludeMdlc.


snakefoot:
Closed this, since #2164 has become the favoire

PR #2109, opened 2017-05-15
JsonLayout - IncludeMdc and IncludeMdlc

snakefoot:
Partial fix for #2108 (Missing NDC)

PR #2164, opened 2017-06-12
JsonLayout - IncludeMdc and IncludeMdlc are not ThreadAgnostic v2

snakefoot:
Different solution to the same problem.

### DIFF
--- a/src/NLog/Layouts/Layout.cs
+++ b/src/NLog/Layouts/Layout.cs
@@ -63,7 +63,7 @@ namespace NLog.Layouts
         /// 
         /// Thread-agnostic layouts only use contents of <see cref="LogEventInfo"/> for its output.
         /// </remarks>
-        internal bool ThreadAgnostic { get; private set; }
+        internal bool ThreadAgnostic { get; set; }
 
         /// <summary>
         /// Gets the level of stack trace information required for rendering.
--- a/src/NLog/Layouts/JsonLayout.cs
+++ b/src/NLog/Layouts/JsonLayout.cs
@@ -101,6 +101,24 @@ namespace NLog.Layouts
 #endif
 
         /// <summary>
+        /// Initializes the layout.
+        /// </summary>
+        protected override void InitializeLayout()
+        {
+            base.InitializeLayout();
+            if (IncludeMdc)
+            {
+                base.ThreadAgnostic = false;
+            }
+#if NET4_0 || NET4_5
+            if (IncludeMdlc)
+            {
+                base.ThreadAgnostic = false;
+            }
+#endif
+        }
+
+        /// <summary>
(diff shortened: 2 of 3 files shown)


src/NLog/Layouts/Layout.cs
```diff
@@ -63,7 +63,7 @@ public abstract class Layout : ISupportsInitialize, IRenderable
         /// 
         /// Thread-agnostic layouts only use contents of <see cref="LogEventInfo"/> for its output.
         /// </remarks>
-        internal bool ThreadAgnostic { get; private set; }
+        internal bool ThreadAgnostic { get; set; }
```
304NotModified: I prefer to make it `protected set`. What do you think?

304NotModified:
I really prefer this solution :)

src/NLog/Layouts/Layout.cs
snakefoot: Well internal is "almost" private so it is more restrictive than protected. You are not allowed to upgrade the setter to protected, so it can be accessed by external classes.

Else the entire property should become protected, and part of the official interface.

304NotModified: I meant `protected internal`. Not sure if that could be set on the setter only. 

Its indeed and internal optimization (just makes the intention more clear).

304NotModified: Scratch that, its then protected or internal. Not *and*

304NotModified:
If you rebase on the latest master, does sonarqube work then?

snakefoot:
> If you rebase on the latest master, does sonarqube work then?

Not that strong with GIT. Can only see one master-branch, and no changes has been made to this for some time, and this PR is branched of that same master. What do you mean?

304NotModified:
Could you do

    git remote -v

snakefoot:
> git remote -v

```
andre   https://github.com/AndreGleichner/NLog.git (fetch)
andre   https://github.com/AndreGleichner/NLog.git (push)
nlog    https://github.com/NLog/NLog/ (fetch)
nlog    https://github.com/NLog/NLog/ (push)
origin  https://github.com/snakefoot/NLog.git (fetch)
origin  https://github.com/snakefoot/NLog.git (push)
```

304NotModified:
Just on mobile, so hopefully no errors

```
git fetch
git rebase nlog/master
```

After fixing stuff (`git rebase --continue`)

When done

    git push --force-with-lease

304NotModified:
Updated it, you need fetch all

snakefoot:
@304NotModified Updated and triggered rebuild

304NotModified:
(reported here: https://github.com/NLog/NLog/pull/2109#issuecomment-307721262)

PR #2173, opened 2017-06-17
NLog 5 - Up-to-date with NLog 4.4.11

304NotModified:


src/NLog/Layouts/JsonLayout.cs
```diff
@@ -108,12 +108,12 @@ protected override void InitializeLayout()
             base.InitializeLayout();
             if (IncludeMdc)
             {
-                ThreadAgnostic = false;
+                base.ThreadAgnostic = false;
             }
-#if !SILVERLIGHT && !NETSTANDARD || NETSTANDARD1_3
```
304NotModified: @snakefoot the check ` !NETSTANDARD || NETSTANDARD1_3` is a bit strange, as every `NETSTANDARD` platfrom is also `NETSTANDARD1_3` in your project.json

snakefoot: If you remove NETSTANDARD then IncludeMdlc is supported by all platforms except SILVERLIGHT. So I guess it should be:

#if !SILVERLIGHT

304NotModified: good one, will make a new PR for it. Thanks.

right now I will that the same `#if` as the `IncludeMdlc` check

PR #2179, opened 2017-06-19
JsonSerializer - Serialize Object Properties into StringBuilder

snakefoot:
Prototype of how one could extend Json Serializer to handle object-properties.

Stolen the unit tests from #2128

### DIFF
--- a/src/NLog/Internal/MruCache.cs
+++ b/src/NLog/Internal/MruCache.cs
@@ -81,7 +81,7 @@ namespace NLog.Internal
                     if (_dictionary.Count >= _maxCapacity)
                     {
                         // Make 3 sweeps:
-                        //  1) Versions below _currentVerison - _maxCapacity / 2
+                        //  1) Versions below _currentVerison - _maxCapacity / 1.5
                         //  2) Versions below _currentVerison - _maxCapacity / 10
                         //  3) Versions below _currentVerison - 2
                         var pruneKeys = new List<TKey>(_dictionary.Count);
@@ -114,7 +114,7 @@ namespace NLog.Internal
 
                         if (_dictionary.Count >= _maxCapacity)
                         {
-                            _dictionary.Clear();
+                            _dictionary.Clear(); // Failed to perform sweep, fallback to fail safe
                         }
                     }
 
@@ -148,6 +148,7 @@ namespace NLog.Internal
 
             if (item.Version != _currentVersion)
             {
+                // Update item version to mark as recently used
                 lock (_dictionary)
                 {
                     var version = _currentVersion;
--- a/src/NLog/Internal/MruCache.cs
+++ b/src/NLog/Internal/MruCache.cs
@@ -77,7 +77,7 @@ namespace NLog.Internal
                             long oldVersion = version - 2;
                             switch (i)
                             {
-                                case 1: oldVersion = version - _maxCapacity / 2; break;
+                                case 1: oldVersion = version - (int)(_maxCapacity / 1.5); break;
                                 case 2: oldVersion = version - _maxCapacity / 10; break;
                             }
                             foreach (var element in _dictionary)
--- a/src/NLog/Internal/MruCache.cs
+++ b/src/NLog/Internal/MruCache.cs
@@ -35,19 +35,32 @@ using System.Collections.Generic;
 
 namespace NLog.Internal
 {
+    /// <summary>
+    /// Most-Recently-Used-Cache, that discards less frequently used items on overflow
+    /// </summary>
     internal class MruCache<TKey, TValue>
     {
         private readonly Dictionary<TKey, MruItem> _dictionary;
         private readonly int _maxCapacity;
         private long _currentVersion;
 
-        public MruCache(int cap)
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="maxCapacity">Max number of items to hold before discarding</param>
+        public MruCache(int maxCapacity)
         {
-            _maxCapacity = cap;
+            _maxCapacity = maxCapacity;
             _dictionary = new Dictionary<TKey, MruItem>(_maxCapacity);
             _currentVersion = 1;
         }
 
+        /// <summary>
+        /// Insert item into caceh
+        /// </summary>
+        /// <param name="key">Dictionary Key</param>
+        /// <param name="value">Dictionary Value</param>
+        /// <returns>Key did not exist already</returns>
         public bool TryAddValue(TKey key, TValue value)
         {
             lock (_dictionary)
@@ -56,9 +69,9 @@ namespace NLog.Internal
                 if (_dictionary.TryGetValue(key, out item))
                 {
                     var version = _currentVersion;
-                    if (item.Version != version)
+                    if (item.Version != version || !EqualityComparer<TValue>.Default.Equals(value, item.Value))
                     {
-                        _dictionary[key] = new MruItem(item.Value, version);
+                        _dictionary[key] = new MruItem(value, version);
                     }
                     return false;   // Already exists
                 }
@@ -111,6 +124,12 @@ namespace NLog.Internal
             }
         }
 
+        /// <summary>
+        /// Lookup existing item in cache
+        /// </summary>
+        /// <param name="key">Dictionary Key</param>
+        /// <param name="value">Dictionary Value [out]</param>
+        /// <returns>Key found</returns>
         public bool TryGetValue(TKey key, out TValue value)
         {
             MruItem item;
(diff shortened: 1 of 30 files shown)


snakefoot:
I think that I can make this PR a little simpler. After I introduced IJsonSerializerV2, then I don't need a Json-serializer instance for each JsonLayout. The MruCache should be able to handle concurrent access, so all the json-factory-logic can be removed/simplified.

snakefoot:
Discovered that trying to optimize the serialization of dictionary and ienumerable would require an expensive GetType()-call to check if the serializer-method still works with the object. Decided not to implement this.

snakefoot:
@304NotModified Have implemented a simple MruCache and optimized for using StringBuilder by default.

304NotModified:
:+1: :+1:  will try to review it this weekend

snakefoot:
Made a small performance test using a simple DTO with 3 properties (string, integer, enum). And it could serialize 1.000.000 items per second. Not the fastest in the world, but should cover the needs of NLog.

src/NLog/Internal/MruCache.cs
```diff
@@ -0,0 +1,185 @@
...
+                        }
+
+                        foreach (var pruneKey in pruneKeys)
+                        {
+                            _dictionary.Remove(pruneKey);
+                        }
+
+                        if (_dictionary.Count >= _maxCapacity)
+                        {
+                            _dictionary.Clear();
```
304NotModified: We can do this before the delete prune?

304NotModified: With different check of course

src/NLog/Internal/MruCache.cs
```diff
@@ -0,0 +1,185 @@
...
+                    var version = _currentVersion;
+                    if (item.Version != version || !EqualityComparer<TValue>.Default.Equals(value, item.Value))
+                    {
+                        _dictionary[key] = new MruItem(value, version);
+                    }
+                    return false;   // Already exists
+                }
+                else
+                {
+                    var version = ++_currentVersion;
```
304NotModified: Postfix ++?

src/NLog/Internal/MruCache.cs
```diff
@@ -0,0 +1,185 @@
...
+                    value = default(TValue);
+                    return false;
+                }
+            }
+            catch
+            {
+                item = default(MruItem);    // Too many people in the same room
+            }
+
+            if (item.Version != _currentVersion)
```
304NotModified: Please some docs here, thanks! (I guess  retrieving the sane is here on purpose)

src/NLog/Layouts/JsonLayout.cs
```diff
@@ -46,6 +46,44 @@ namespace NLog.Layouts
...
     {
+        private NLog.Targets.IJsonSerializerV2 JsonSerializer
+        {
+            get
+            {
+                if (_jsonSerializer == null)
+                {
+                    var jsonSerializer = ConfigurationItemFactory.Default.JsonSerializer;
+                    _jsonSerializer = jsonSerializer as NLog.Targets.IJsonSerializerV2;
+                    if (_jsonSerializer == null && jsonSerializer != null)
```
304NotModified: Maybe nicer if we fix this in the setter?

src/NLog/Targets/DefaultJsonSerializer.cs
```diff
@@ -72,116 +79,390 @@ private DefaultJsonSerializer()
...
-
-            encodeString = true;
-            return JsonStringEscape(stringValue, escapeUnicode);
+            return objTypeCode != TypeCode.String && (objTypeCode == TypeCode.Empty  // Don't put quotes around null values
+                || objTypeCode == TypeCode.Boolean
+                || IsNumericTypeCode(objTypeCode, true));
         }
 
-        private static bool IsNumericTypeCode(TypeCode objTypeCode)
+        private static bool IsNumericTypeCode(TypeCode objTypeCode, bool includeDecimals)
```
304NotModified: Please add some docs why include decimals could be false

src/NLog/Targets/IJsonSerializer.cs
```diff
@@ -49,4 +49,15 @@ public interface IJsonSerializer
...
+
+    internal interface IJsonSerializerV2
+    {
+        /// <summary>
+        /// Returns a serialization of an object
+        /// into JSON format.
+        /// </summary>
+        /// <param name="value">The object to serialize to JSON.</param>
+        /// <param name="builder">Output destination.</param>
+        bool SerializeObject(object value, System.Text.StringBuilder builder);
```
304NotModified: Please document the output value. Is it true if success?

src/NLog/Targets/IJsonSerializer.cs
```diff
@@ -49,4 +49,15 @@ public interface IJsonSerializer
         /// <returns>Serialized value.</returns>
         string SerializeObject(object value);
     }
+
+    internal interface IJsonSerializerV2
```
304NotModified: I think it should be public and in the NLog namespace, what do you think?

304NotModified:
Maybe we have to include also some more test cases. 126 missed lines could be better

src/NLog/Targets/IJsonSerializer.cs
snakefoot: Makes sense to have it public, as it would allow others to optimize for StringBuilder. Not sure about the naming-convention though, by making it internal, then I could just choose a "random" name :).

src/NLog/Internal/MruCache.cs
snakefoot: I want "version" to have the updated value. I guess I could write:

`var version = _currentVersion++ + 1 ;`

src/NLog/Internal/MruCache.cs
snakefoot: The `_dictionary.Clear();` is a fail safe, in case something has gone terrible wrong, and all items has the same version.

src/NLog/Layouts/JsonLayout.cs
snakefoot: Well it is lazy-initialization.

snakefoot:
> Maybe we have to include also some more test cases. 126 missed lines could be better

Have moved SingleItemOptimizedHashSet into its own file, and added some tests.

src/NLog/Internal/MruCache.cs
snakefoot: Done

src/NLog/Targets/DefaultJsonSerializer.cs
snakefoot: Done

src/NLog/Targets/IJsonSerializer.cs
snakefoot: Done

src/NLog/Layouts/JsonLayout.cs
304NotModified: fair enough :)

src/NLog/Targets/IJsonSerializer.cs
```diff
@@ -46,7 +46,19 @@ public interface IJsonSerializer
         /// into JSON format.
         /// </summary>
         /// <param name="value">The object to serialize to JSON.</param>
-        /// <returns>Serialized value.</returns>
+        /// <returns>Serialized value (null = Serialize failed).</returns>
         string SerializeObject(object value);
     }
+
+    internal interface IJsonSerializerV2
```
304NotModified: Ow a commit has not been added from mobile...

I think it's important to have the interface public, maybe we could move it to the "NLog" namespace? (and mark the other as obsolete)

While the default serialization is good enough for most cases, it should be possible to replace it with e.g. JSON.NET

snakefoot: Yes it will not be super optimal for Json.Net as the implementation would be:

```
bool SerializeObject(object value, System.Text.StringBuilder builder)
{
    try
    {
       StringWriter sw = new StringWriter(sb);
       JsonSerializer serializer = new JsonSerializer();
       using (JsonWriter jsonTextWriter = new JsonTextWriter(sw))
       {
            serializer.Serialize(jsonTextWriter, obj);
       }
       return true;
   }
   catch
   {
      return false;
   }
}
```

snakefoot:
@304NotModified I have now introduce NLog.IJsonConverter and marked NLog.Target.IJsonSerializer as obsolete.

snakefoot:
> IJsonConverter?

Converts string to JSON :). Feel free to provide any class name you want

snakefoot:
Now fixed the obsolete-text to say NLog.IJsonConverter. Good catch.

304NotModified:
I think its better to give the hint of the change property instead of the interface in the obsolete text. Have updated it.

304NotModified:
merged! Thanks!

PR #2126, opened 2017-05-19
[WIP] Structured events (NLog 4.5 - alpha5)

304NotModified:
see https://github.com/NLog/NLog/issues/2077

supersedes https://github.com/NLog/NLog/pull/1951

Issue #1962, opened 2017-02-13
Provide access to the parsed template for structured log events


nblumhardt:
**Type** (choose one): Feature request
**NLog version**: 4.5-alpha01
**Platform**: .NET 4.5, others

When processing a structured log event, the underlying `Template` object is useful for a few reasons.

 * Targets that want to pretty-print output need to iterate over the individual template tokens, e.g. see https://github.com/serilog/serilog-sinks-literate and https://github.com/logary/logary/pull/155
 * Targets that capture the template itself for later offline rendering need to pre-render `Hole`s that have a `Format` associated with them, see: https://github.com/datalust/nlog-targets-seq/issues/2

It'd be nice not to have to parse the template twice in this situation, since it's already parsed in `LogEventInfo.CalcFormattedMessage()`.

A simple way to do this would be to add a lazily-initialized `Template` property to `LogEventInfo`.

To avoid extending the `Template`'s lifetime (potentially hurting GC performance), a `GetTemplate()` method might be a better option. This would internally parse the template, then call the equivalent of `CalcFormattedMessage()` passing the parsed `Template` as an argument, before returning it from `GetTemplate()` (no backing `_template` field would thus be needed).

304NotModified:
Will implement this soon :)

304NotModified:
Also in alpha 2!

GetMessageTemplate.

nblumhardt:
Awesome :-) will update ASAP for this and #1963.

nblumhardt:
@304NotModified just one slight issue- the method is cased `GetmessageTemplate()` - lowercase 'm'

304NotModified:
Oops gonna fix that. Was also thinking I should return an interface eventually.

304NotModified:
This is Fixed in alpha3

PR #2208, opened 2017-07-06
Added Structured events / Message Templates

snakefoot:
Have merged https://github.com/NLog/NLog.StructuredEvents together with https://github.com/NLog/NLog

And then introduced a TemplateEnumerator, that only allocates the hole-name-strings (when non-positional), and not any MessageTemplate-object (with friends).

Stolen the unit tests from #2126

Have added some extra performance optimizations:

- New StringBuilderPool so MessageTemplate-renderering only allocates the formatted-string (just like string.Format)
- Using new StringBuilderPool other places in the NLog-core-engine to minimize allocations (AppendBuilderCreator)
- Better reuse of already allocated StringBuilder when possible (AppendBuilderCreator)
- Faster encoding of StringBuilder to byte-array. Instead of doing the encoding twice, to find the exact byte-count, then pre-allocate max-byte-count, and truncate afterwards (if needed).

### DIFF
--- a/src/NLog/Config/ConfigurationItemFactory.cs
+++ b/src/NLog/Config/ConfigurationItemFactory.cs
@@ -205,6 +205,41 @@ namespace NLog.Config
         }
 
         /// <summary>
+        /// Gets or sets the string serializer to use with <see cref="LogEventInfo.MessageTemplateParameters"/>
+        /// </summary>
+        public IValueSerializer ValueSerializer
+        {
+            get { return MessageTemplates.ValueSerializer.Instance; }
+            set { MessageTemplates.ValueSerializer.Instance = value; }
+        }
+
+        /// <summary>
+        /// Perform mesage template parsing and formatting of LogEvent messages (True = Always, False = Never, Null = Auto Detect)
+        /// </summary>
+        public bool? EnableMessageTemplateParser
+        {
+            get
+            {
+                if (ReferenceEquals(LogEventInfo.DefaultMessageFormatter, LogEventInfo.StringFormatMessageFormatter))
+                {
+                    return false;
+                }
+                else if (ReferenceEquals(LogEventInfo.DefaultMessageFormatter, LogMessageTemplateFormatter.Default.MessageFormatter))
+                {
+                    return true;
+                }
+                else
+                {
+                    return null;
+                }
+            }
+            set
+            {
+                LogEventInfo.SetDefaultMessageFormatter(value);
+            }
+        }
+
+        /// <summary>
         /// Gets the time source factory.
         /// </summary>
         /// <value>The time source factory.</value>
@@ -221,8 +256,6 @@ namespace NLog.Config
         {
             get { return this.conditionMethods; }
         }
-
-
 
         /// <summary>
         /// Registers named items from the assembly.
--- a/src/NLog/Config/XmlLoggingConfiguration.cs
+++ b/src/NLog/Config/XmlLoggingConfiguration.cs
@@ -595,6 +595,9 @@ namespace NLog.Config
             InternalLogger.LogToConsoleError = nlogElement.GetOptionalBooleanAttribute("internalLogToConsoleError", InternalLogger.LogToConsoleError);
             InternalLogger.LogFile = nlogElement.GetOptionalAttribute("internalLogFile", InternalLogger.LogFile);
 
+            bool? messageTemplateParser = nlogElement.GetOptionalBooleanAttribute("messageTemplateParser", null);
+            this.ConfigurationItemFactory.EnableMessageTemplateParser = messageTemplateParser;
+
 #if !SILVERLIGHT && !__IOS__ && !__ANDROID__
             InternalLogger.LogToTrace = nlogElement.GetOptionalBooleanAttribute("internalLogToTrace", InternalLogger.LogToTrace);
 #endif
--- a/src/NLog/Internal/ILogMessageFormatter.cs
+++ b/src/NLog/Internal/ILogMessageFormatter.cs
@@ -1,0 +1,57 @@
+﻿// 
+// Copyright (c) 2004-2017 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without 
+// modification, are permitted provided that the following conditions 
+// are met:
+// 
+// * Redistributions of source code must retain the above copyright notice, 
+//   this list of conditions and the following disclaimer. 
+// 
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution. 
+// 
+// * Neither the name of Jaroslaw Kowalski nor the names of its 
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission. 
+// 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE 
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE 
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE 
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR 
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS 
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN 
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) 
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF 
+// THE POSSIBILITY OF SUCH DAMAGE.
+// 
+
+namespace NLog.Internal
+{
+    using System;
+
+    /// <summary>
+    /// Format a log message
+    /// </summary>
+    internal interface ILogMessageFormatter
+    {
+        /// <summary>
+        /// Format the message and return
+        /// </summary>
+        /// <param name="logEvent">LogEvent with message to be formatted</param>
+        /// <returns>formatted message</returns>
+        string FormatMessage(LogEventInfo logEvent);
+
+        /// <summary>
+        /// Has the logevent properties?
+        /// </summary>
+        /// <param name="logEvent">LogEvent with message to be formatted</param>
+        /// <returns>formatted message</returns>
+        bool HasProperties(LogEventInfo logEvent);
+    }
+}
--- a/src/NLog/Internal/LogMessageTemplateFormatter.cs
+++ b/src/NLog/Internal/LogMessageTemplateFormatter.cs
@@ -1,0 +1,101 @@
+﻿// 
+// Copyright (c) 2004-2017 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without 
+// modification, are permitted provided that the following conditions 
+// are met:
+// 
+// * Redistributions of source code must retain the above copyright notice, 
+//   this list of conditions and the following disclaimer. 
+// 
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution. 
+// 
+// * Neither the name of Jaroslaw Kowalski nor the names of its 
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission. 
+// 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE 
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE 
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE 
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR 
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS 
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN 
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) 
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF 
+// THE POSSIBILITY OF SUCH DAMAGE.
+// 
+
+namespace NLog.Internal
+{
+    using System;
+    using System.Globalization;
+    using NLog.MessageTemplates;
+
+    internal sealed class LogMessageTemplateFormatter : ILogMessageFormatter
+    {
+        public static readonly LogMessageTemplateFormatter DefaultAuto = new LogMessageTemplateFormatter(false);
+        public static readonly LogMessageTemplateFormatter Default = new LogMessageTemplateFormatter(true);
+        private static readonly StringBuilderPool _builderPool = new StringBuilderPool(Environment.ProcessorCount * 2);
+
+        /// <summary>
+        /// When true: Do not fallback to StringBuilder.Format for positional templates
+        /// </summary>
+        private readonly bool _forceTemplateRenderer;
+
+        /// <summary>
+        /// New formatter
+        /// </summary>
+        /// <param name="forceTemplateRenderer">When true: Do not fallback to StringBuilder.Format for positional templates</param>
+        private LogMessageTemplateFormatter(bool forceTemplateRenderer)
+        {
+            _forceTemplateRenderer = forceTemplateRenderer;
+            MessageFormatter = FormatMessage;
+        }
+
+        /// <summary>
+        /// The MessageFormatter delegate
+        /// </summary>
+        public LogMessageFormatter MessageFormatter { get; }
+
+        public bool HasProperties(LogEventInfo logEvent)
+        {
+            //if message is empty, there no parameters
+            //null check cheapest, so in-front
+            return logEvent.Parameters != null && !string.IsNullOrEmpty(logEvent.Message) && logEvent.Parameters.Length > 0;
+        }
+
+        public string FormatMessage(LogEventInfo logEvent)
+        {
+            if (!HasProperties(logEvent))
+            {
+                return logEvent.Message;
+            }
+            // Prevent multiple layouts on different targets to render the same properties
+            lock (logEvent)
+            {
+                using (var builder = _builderPool.Acquire())
+                {
+                    logEvent.Message.Render(logEvent.FormatProvider ?? CultureInfo.CurrentCulture, logEvent.Parameters, _forceTemplateRenderer, builder.Item, out var messageTemplateParameterList);
+                    if (logEvent.PropertiesDictionary == null)
+                    {
+                        if (messageTemplateParameterList != null && messageTemplateParameterList.Count > 0)
+                        {
+                            logEvent.PropertiesDictionary = new PropertiesDictionary(messageTemplateParameterList);
+                        }
+                    }
+                    else
+                    {
+                        logEvent.PropertiesDictionary.MessageProperties = messageTemplateParameterList;
+                    }
+                    return builder.Item.ToString();
+                }
+            }
+        }
+    }
+}
--- a/src/NLog/Internal/MruCache.cs
+++ b/src/NLog/Internal/MruCache.cs
@@ -71,56 +71,78 @@ namespace NLog.Internal
                     var version = _currentVersion;
                     if (item.Version != version || !EqualityComparer<TValue>.Default.Equals(value, item.Value))
                     {
-                        _dictionary[key] = new MruItem(value, version);
+                        _dictionary[key] = new MruItem(value, version, false);
                     }
                     return false;   // Already exists
                 }
                 else
                 {
-                    var version = ++_currentVersion;
                     if (_dictionary.Count >= _maxCapacity)
                     {
-                        // Make 3 sweeps:
-                        //  1) Versions below _currentVerison - _maxCapacity / 1.5
-                        //  2) Versions below _currentVerison - _maxCapacity / 10
-                        //  3) Versions below _currentVerison - 2
-                        var pruneKeys = new List<TKey>(_dictionary.Count);
-                        for (int i = 1; i <= 3; ++i)
-                        {
-                            long oldVersion = version - 2;
-                            switch (i)
-                            {
-                                case 1: oldVersion = version - (int)(_maxCapacity / 1.5); break;
-                                case 2: oldVersion = version - _maxCapacity / 10; break;
-                            }
-                            foreach (var element in _dictionary)
-                            {
-                                if (element.Value.Version <= oldVersion)
-                                {
-                                    pruneKeys.Add(element.Key);
-                                    if (_dictionary.Count - pruneKeys.Count < _maxCapacity / 1.5)
-                                    {
-                                        i = 3;
-                                        break;  // Do not clear the entire cache
-                                    }
-                                }
-                            }
-                        }
-
-                        foreach (var pruneKey in pruneKeys)
-                        {
-                            _dictionary.Remove(pruneKey);
-                        }
-
-                        if (_dictionary.Count >= _maxCapacity)
-                        {
-                            _dictionary.Clear(); // Failed to perform sweep, fallback to fail safe
-                        }
+                        ++_currentVersion;
+                        PruneCache();
                     }
 
-                    _dictionary.Add(key, new MruItem(value, version));
+                    _dictionary.Add(key, new MruItem(value, _currentVersion, true));
                     return true;
                 }
+            }
+        }
+
+        private void PruneCache()
+        {
+            // There 3 priorities:
+            //  - High Priority - Non-Virgins with version very close to _currentVersion
+            //  - Medium Priority - Version close to _currentVersion
+            //  - Low Priority - Old-Virgins
+            // Make 3 sweeps:
+            //  - Kill all the old ones
+            //  - Kill all the virgins
+            //  - Slaughterhouse
+            long latestGeneration = _currentVersion - 2;
+            long oldestGeneration = 1;
+            var pruneKeys = new List<TKey>((int)(_dictionary.Count / 2.5));
+            for (int i = 0; i < 3; ++i)
+            {
+                long oldGeneration = _currentVersion - 5;
+                switch (i)
+                {
+                    case 0: oldGeneration = _currentVersion - (int)(_maxCapacity / 1.5); break;
+                    case 1: oldGeneration = _currentVersion - 10; break;
+                }
+                if (oldGeneration < oldestGeneration)
+                    oldGeneration = oldestGeneration;
+
+                oldestGeneration = long.MaxValue;
+
+                long elementGeneration = 0;
+                foreach (var element in _dictionary)
+                {
+                    elementGeneration = element.Value.Version;
+                    if (elementGeneration <= oldGeneration || (element.Value.Virgin && (i != 0 || elementGeneration < latestGeneration)))
+                    {
+                        pruneKeys.Add(element.Key);
+                        if (_dictionary.Count - pruneKeys.Count < _maxCapacity / 1.5)
+                        {
+                            i = 3;
+                            break;  // Do not clear the entire cache
+                        }
+                    }
+                    else if (elementGeneration < oldestGeneration)
+                    {
+                        oldestGeneration = elementGeneration;
+                    }
+                }
+            }
+
+            foreach (var pruneKey in pruneKeys)
+            {
+                _dictionary.Remove(pruneKey);
+            }
+
+            if (_dictionary.Count >= _maxCapacity)
+            {
+                _dictionary.Clear(); // Failed to perform sweep, fallback to fail safe
             }
         }
 
@@ -146,7 +168,7 @@ namespace NLog.Internal
                 item = default(MruItem);    // Too many people in the same room
             }
 
-            if (item.Version != _currentVersion)
+            if (item.Version != _currentVersion || item.Virgin)
             {
                 // Update item version to mark as recently used
                 lock (_dictionary)
@@ -154,9 +176,13 @@ namespace NLog.Internal
                     var version = _currentVersion;
                     if (_dictionary.TryGetValue(key, out item))
                     {
-                        if (item.Version != version)
+                        if (item.Version != version || item.Virgin)
                         {
-                            _dictionary[key] = new MruItem(item.Value, version);
+                            if (item.Virgin)
+                            {
+                                version = ++_currentVersion;
+                            }
+                            _dictionary[key] = new MruItem(item.Value, version, false);
                         }
                     }
                     else
@@ -175,11 +201,13 @@ namespace NLog.Internal
         {
             public readonly TValue Value;
             public readonly long Version;
+            public readonly bool Virgin;
 
-            public MruItem(TValue value, long version)
+            public MruItem(TValue value, long version, bool virgin)
             {
                 Value = value;
                 Version = version;
+                Virgin = virgin;
             }
         }
     }
--- a/src/NLog/Internal/StringBuilderExt.cs
+++ b/src/NLog/Internal/StringBuilderExt.cs
@@ -52,31 +52,15 @@ namespace NLog.Internal
         /// <param name="formatProvider">provider, for example culture</param>
         public static void AppendFormattedValue(this StringBuilder builder, object value, string format, IFormatProvider formatProvider)
         {
-            if (format == "@")
+            string stringValue = value as string;
+            if (stringValue != null && string.IsNullOrEmpty(format))
             {
-                Config.ConfigurationItemFactory.Default.JsonConverter.SerializeObject(value, builder);
-                return;
+                builder.Append(value);  // Avoid automatic quotes
             }
-
-            if (value == null)
+            else if (value != null || !string.IsNullOrEmpty(format))
             {
-                return;
+                MessageTemplates.ValueSerializer.Instance.SerializeObject(value, format, formatProvider, builder);
             }
-
-            if (format == null)
-            {
-                builder.Append(Convert.ToString(value, formatProvider));
-                return;
-            }
-
-            var formattable = value as IFormattable;
-            if (formattable != null)
-            {
-                builder.Append(formattable.ToString(format, formatProvider));
-                return;
-            }
-
-            builder.Append(Convert.ToString(value, formatProvider));
         }
 
         /// <summary>
--- a/src/NLog/Internal/StringBuilderExt.cs
+++ b/src/NLog/Internal/StringBuilderExt.cs
@@ -169,14 +169,19 @@ namespace NLog.Internal
 #if !SILVERLIGHT
             if (transformBuffer != null)
             {
+                int charCount = 0;
+                int byteCount = 0;
                 for (int i = 0; i < builder.Length; i += transformBuffer.Length)
                 {
-                    int charCount = Math.Min(builder.Length - i, transformBuffer.Length);
+                    charCount = Math.Min(builder.Length - i, transformBuffer.Length);
                     builder.CopyTo(i, transformBuffer, 0, charCount);
-                    int byteCount = encoding.GetByteCount(transformBuffer, 0, charCount);
-                    ms.SetLength(ms.Length + byteCount);
-                    encoding.GetBytes(transformBuffer, 0, charCount, ms.GetBuffer(), (int)ms.Position);
-                    ms.Position = ms.Length;
+                    byteCount = encoding.GetMaxByteCount(charCount);
+                    ms.SetLength(ms.Position + byteCount);
+                    byteCount = encoding.GetBytes(transformBuffer, 0, charCount, ms.GetBuffer(), (int)ms.Position);
+                    if ((ms.Position += byteCount) != ms.Length)
+                    {
+                        ms.SetLength(ms.Position);
+                    }
                 }
             }
             else
--- a/src/NLog/Layouts/JsonLayout.cs
+++ b/src/NLog/Layouts/JsonLayout.cs
@@ -138,11 +138,12 @@ namespace NLog.Layouts
         /// Formats the log event as a JSON document for writing.
         /// </summary>
         /// <param name="logEvent">The logging event.</param>
-        /// <param name="target">Initially empty <see cref="StringBuilder"/> for the result</param>
+        /// <param name="target"><see cref="StringBuilder"/> for the result</param>
         protected override void RenderFormattedMessage(LogEventInfo logEvent, StringBuilder target)
         {
+            int orgLength = target.Length;
             RenderJsonFormattedMessage(logEvent, target);
-            if (target.Length == 0 && RenderEmptyObject)
+            if (target.Length == orgLength && RenderEmptyObject)
             {
                 target.Append(SuppressSpaces ? "{}" : "{  }");
             }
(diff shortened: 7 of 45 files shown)


snakefoot:
@304NotModified I know that i have stolen your code, and this is mostly a proof-of-concept. But maybe you can take some of the ideas in this PR to hammer the last nails in place for NLog 4.5.

Not sure I like exposing the Hole-structs directly. Would prefer they were just "Index" + "Name" + "Format". But I like reusing the Hole-array-allocation, instead of creating a new container.

304NotModified:
> Not sure I like exposing the Hole-structs directly.

My idea was to use the event-properties for that. The holes renders fills the properties and if desired, users could overwrite them. Less complicated for the user and can use all the targets/layout renders which are built on the event-properties.

snakefoot:
> My idea was to use the event-properties for that.

Have also implemented that by providing the new property LogEventInfo.ParameterProperties.

But I recall this PR #1962 where someone wanted access to the holes and their format-options, and not just the parameter-names and the parameter-objects

304NotModified:
> But I recall this PR #1962 where someone wanted access to the holes and their format-options, and not just the parameter-names and the parameter-objects

I was under the impression that your proposal for that was: reparse the message.

snakefoot:
>  prefer to keep the structured event logic in a separate package because:

Since the NLog-engine is going to be fully integrated with the structured-event-logic, then I find it hard to split them without it becoming very artificial.

I understand the need to rip out all the random targets and layout-renderers, but the core engine should stay in one package.

snakefoot:
> I was under the impression that your proposal for that was: reparse the message.

My initial thought was to provide a way to request a single parameter to be formatted as string, like specified in the message template. But after seeing that #1962 wants to use his own target-specific-formatter, then my initial thought would be too restrictive.

304NotModified:
We lost the parse and render tests from https://github.com/NLog/NLog.StructuredEvents?

snakefoot:
> We lost the parse and render tests from https://github.com/NLog/NLog.StructuredEvents?

I will inject them

304NotModified:
> I understand the need to rip out all the random targets and layout-renderers

The idea is now, new code is in new packages and current stuff will be separated after NLog 5 RTM

> But when looking at ASP.NET-CORE loggers that seems prepared for Structured-logging, then I find it less optimal that the NLog-core is not able to handle structured-logging out of the box.

It's still out of the box when those are multiple packages. I will think about it, but currently i'm not a fan to merge them. 






> Else we need to improve the LogEventInfo-Properties even further so one can have internal meta-properties

I don't see why we need that. Just write to the properties, drop everything else. 
I agree that having the Tempate inside the EventLogInfo was not the best solution, but IMO it's now even less optimal. It too complicated. I really don't like to expose a Hole list in the LogEventInfo.

snakefoot:
>  I really don't like to expose a Hole list in the LogEventInfo.

Glad you feel the same way :). Was not sure how far you wanted to go to cater for nlog-targets-seq.

But I don't like pushing the parameter-names directly into LogEventInfo.Properties, because of the name-clashing (that you have also tried to handle). It will give unexpected behavior, no matter what you do.

Instead of having the ParameterHoles. Then we could have something called ParameterNames, that is just a IList<string> ?

304NotModified:
> But I don't like pushing the parameter-names directly into LogEventInfo.Properties, because of the potential name-clashing

IMO the structured syntax is another way to define properties. I don't see the benefit of having properties and parameters. Those are on the same level (logevent level) and so separation is only over complicating things IMO.

snakefoot:
> I don't see the benefit of having properties and parameters.

Then one will just be surprised the day one have a custom-logger than always inject some custom-properties, and that logger is suddenly presented with a message-template that clashes.

Also I like how one can get direct mapping from the parameter-names to the parameter-values without allocating a dictionary.

What if we introduce a new MetaProperties-dictionary (that extends the object-key with a meta-type-enum). This will give one properties-collection, that will just allow "duplicate" object-keys. The existing Properties-dictionary will then just return the values included from meta-type-property (and not meta-type-parametername)

304NotModified:
> Then one will just be surprised the day one have a custom-logger that always inject some custom-properties, and that logger is suddenly presented with a message-template that clashes.

That would be also the case now, as I could set the logevent properties with C#

We have now the following contexts:

- GDC
- MDC & MDLC
- NDC & NDLC
- event properties
- event context (deprecated)

Adding a new on event level is makes it even more complex. That the event context is already deprecated is a burden. Only if it's absolute necessary I would consider it.

304NotModified:
PS:

> Merging #2208 into master will decrease coverage by <1%.
> The diff coverage is 80%.

Another issue with merging here, I cannot see easily if the new code is tracked by the unit tests. It doesn't look like it (it was in the other github repos 95% coverage)

But maybe that's an issue that we have to tackle differently, maybe sending the code coverage to sonar would help (but I did not get that working)

snakefoot:
> That would be also the case now, as I could set the logevent properties with C#

Usually the logger will just receive the logevent-message and parameters, and not any properties. But now it is possible for the logger to unwittingly overwrite the properties introduced by message-template.

> Another issue with merging here, I cannot see easily if the new code is tracked by the unit tests

Well that is how proof-of-concept works for me, first you shuffle things around until it looks good, and then you lock things down with unit-tests (and remove unneeded code).

snakefoot:
@304NotModified Have now removed ParameterHoles, and now injecting message-template properties when asking for the normal Properties-dictionary.

snakefoot:
@304NotModified Made a very simple performance test of this PR:

|------------------|------------|------|----|-----|---------|
| SimpleLogger     |  5.000.000 |  128 |   2 |    1 |       1 |

| Test Name  | Time (ms) | Msgs/sec  | GC2 | GC1 | GC0 | CPU (ms) | Mem (MB) |
| NLog-Standard |    21.361 |   234.061 |   0 |  18 |  37 |   23.015 |  113,258 |

| Test Name  | Time (ms) | Msgs/sec  | GC2 | GC1 | GC0 | CPU (ms) | Mem (MB) |
| NLog-Positional    |    24.747 |   202.036 |   0 |  18 |  36 |   26.156 |  118,078 |

The extra CPU time in"NLog-Positional" is caused by the scanning if the template is positional or not.

| Test Name  | Time (ms) | Msgs/sec  | GC2 | GC1 | GC0 | CPU (ms) | Mem (MB) |
| Serilog-Positional    |    34.779 |   143.762 |   0 |  32 |  64 |   69.156 |  102,852 |

| Test Name  | Time (ms) | Msgs/sec  | GC2 | GC1 | GC0 | CPU (ms) | Mem (MB) |
| NLog-Template |    34.724 |   143.989 |   0 |  87 | 175 |   37.687 |  130,496 |

The extra allocations in"NLog-Template" is caused by allocation of mesage-template and stringbuilder for formatting.

| Test Name  | Time (ms) | Msgs/sec  | GC2 | GC1 | GC0 | CPU (ms) | Mem (MB) |
| Serilog-Template    |    34.039 |   146.888 |   0 |  32 |  65 |   67.750 |  100,773 |

I think serilog has an optimization where it caches the MesageTemplates, which helps Serilog in this test. I also think Serilog doesn't allocate the formatted string, but writes it directly to its TextWriter.

NLog could probably improve its performance by not allocating a StringBuilder for every template-format-operation, and maybe attempt to cache the recent message-templates (together with the hole-strings).

304NotModified:
Cool! Can I see the test cases code some where? 

We like to use cache, it was always in the plan. Are you using the mru cache in the test?

Serilog uses a dictionary with a 1000 limit and full clear when over limit.

snakefoot:
> Cool! Can I see the test cases code some where?

It is just the performance test-project where I'm injecting a log-message-arg (with a hole-name of 25 chars) for every 64 char (Thus one in the middle and one at the end of the 128 char log-message).

The serilog-config is this:

```
var serilogConfig = new Serilog.LoggerConfiguration().WriteTo.Async(a => a.File(@"C:\Temp\Log\Serilog.txt"));
var serilogger = serilogConfig.CreateLogger();
```
> We like to use cache, it was always in the plan. Are you using the mru cache in the test?

No the two bonus perf-tests is just me hardcoding things to see how caching affect allocations. The MruCache is not super-concurrent and doesn't expect a constant high insert-rate. All Loggers will hit this code-path and NLog has made an effort in not grapping a global-lock. I have not made any concurrency-tests to see how things scales as the thread-count rises.

> Serilog uses a dictionary with a 1000 limit and full clear when over limit.

They have changed to a HashTable as it allows them to do the lookup without holding a monitor-lock,

304NotModified:
> They have changed to a HashTable as it allows them to do the lookup without holding a monitor-lock.

Ow yes indeed: https://github.com/serilog/serilog/blob/dev/src/Serilog/Core/Pipeline/MessageTemplateCache.cs

304NotModified:
> An alternative approach would be to just cache the parameter-hole-names (for the property-dictionary). Since the MessageTemplate is just discarded after the rendering. Then modify the TemplateEnumerator to render without allocating any MessageTemplate (Would not be an easy task). The number of unique hole-names are probably much less than the number of unique-messages.

Sounds complicated and not really needed IMO



> No the two bonus perf-tests is just me hardcoding things to see how caching affect allocations. The MruCache is not super-concurrent when there is a constant high insert-rate.

Maybe it's a good idea to start with this one? What do you think?

snakefoot:
> The MruCache is not super-concurrent when there is a constant high insert-rate.
>
> Maybe it's a good idea to start with this one? What do you think?

Sure go ahead, you just have to find a lock-free-container with fast lookup, fast insert, fast cleanup and minimum allocation :)

The above test was merely to see how far from the reference-point (Serilog) the current implementation is. Right now it looks pretty close, so maybe it is just a matter of improve the template-parser, and NLog is ready.

Before you start on implementing this caching business, then we might want to create another test, that pushes the concurrency a little:

- Have 6 threads that each are writing to their own log-level
- Each log-level is redirected to their individual target/sink
- Each of the 6 threads will shuffle between 300 unique log-messages.

For me the low hanging fruit is implementing a StringBuilder-cache, so that will probably be next two tasks (builder-cache + new test).

src/NLog/ILogMessageFormatter.cs
```diff
@@ -0,0 +1,122 @@
...
+// THE POSSIBILITY OF SUCH DAMAGE.
+// 
+
+using System;
+using System.Globalization;
+using NLog.StructuredEvents;
+
+namespace NLog
+{
+    interface ILogMessageFormatter
```
304NotModified: I like this approach :)

src/NLog/ILogMessageFormatter.cs
```diff
@@ -0,0 +1,122 @@
...
+    {
+        string FormatMessage(LogEventInfo logEvent);
+        ILogMessageFormatter SetFormatProvider(IFormatProvider formatProvider);
+        IFormatProvider FormatProvider { get; }
+        bool HasProperties(LogEventInfo logEvent);
+    }
+
+    class LogMessageFormatter : ILogMessageFormatter
```
304NotModified: Please split to separate files, also a bit easier to review :)

src/NLog/ILogMessageFormatter.cs
```diff
@@ -0,0 +1,122 @@
...
+
+    class LogMessageTemplateFormatter : ILogMessageFormatter
+    {
+        public static readonly LogMessageTemplateFormatter Default = new LogMessageTemplateFormatter();
+
+        public IFormatProvider FormatProvider { get; private set; }
+
+        public bool HasProperties(LogEventInfo logEvent)
+        {
+            return logEvent.Message != null && logEvent.Parameters != null && logEvent.Parameters.Length > 0;
```
304NotModified: `logEvent.Message != null ` is strange in `HasProperties`. We have to rename or change the behavoir

src/NLog/ILogMessageFormatter.cs
```diff
@@ -0,0 +1,122 @@
...
+                    var properties = logEvent.Properties;
+                    for (int i = 0; i < messageTemplate.Holes.Length; ++i)
+                    {
+                        properties[messageTemplate.Holes[i].Name] = logEvent.Parameters[i];
+                    }
+                    return messageTemplate.Render(FormatProvider ?? CultureInfo.CurrentCulture, logEvent.Parameters);
+                }
+                else
+                {
+                    return string.Format(FormatProvider ?? CultureInfo.CurrentCulture, logEvent.Message, logEvent.Parameters);
```
304NotModified: should we not use `LogMessageFormatter` here?

src/NLog/Layouts/JsonLayout.cs
```diff
@@ -99,6 +99,11 @@ public JsonLayout()
         public bool IncludeAllProperties { get; set; }
 
         /// <summary>
+        /// Gets or sets the option to include all properties from the log events
+        /// </summary>
+        public bool IncludeAllParameters { get; set; }
```
304NotModified: this one is old, isn't it?

src/NLog/LogEventInfo.cs
```diff
@@ -206,8 +205,13 @@ public string Message
             get { return message; }
             set
             {
+                bool rebuildProperties = ResetFormattedProperties();
                 message = value; 
                 ResetFormattedMessage();
+                if (rebuildProperties && !ReferenceEquals(this.FormattedMessage, this.message))
+                {
+                    // Succesfully reformatted message properties
```
304NotModified: is this for debugging purpose?

src/NLog/LogEventInfo.cs
```diff
@@ -231,12 +240,12 @@ public object[] Parameters
         /// </summary>
         public IFormatProvider FormatProvider
         {
-            get { return formatProvider; }
+            get { return this.logMessageFormatter?.FormatProvider; }
             set
             {
-                if (formatProvider != value)
+                if (logMessageFormatter?.FormatProvider != value)
```
304NotModified: got a resharper warning here, not sure if correct

image

src/NLog/LogEventInfo.cs
```diff
@@ -261,7 +270,7 @@ public string FormattedMessage
         /// <summary>
         /// Checks if any per-event context properties (Without allocation)
         /// </summary>
-        public bool HasProperties { get { return this.properties != null && this.properties.Count > 0; } }
+        public bool HasProperties { get { return (this.properties != null && this.properties.Count > 0) || (this.properties == null && this.logMessageFormatter.HasProperties(this)); } }
```
304NotModified: proposal: 

```
        /// <summary>
        /// Checks if any per-event context properties (Without allocation)
        /// </summary>
        public bool HasProperties
        {
            get
            {
                if (this.properties != null)
                {
                    if (this.properties.Count > 0)
                        return true;
                }
                else
                {
                    return this.logMessageFormatter.HasProperties(this);
                }
                return false;
            }
        }

```

As we now checking `this.properties` on `null` twice

src/NLog/LogEventInfo.cs
```diff
@@ -523,6 +536,31 @@ private void CalcFormattedMessage()
             }
         }
 
+        private bool ResetFormattedProperties()
+        {
+            if (this.properties != null && this.properties.Count > 0 && this.logMessageFormatter.HasProperties(this))
```
304NotModified: this is HasProperties? (so duplicate line)

src/NLog/LogEventInfo.cs
```diff
@@ -523,6 +536,31 @@ private void CalcFormattedMessage()
...
+            {
+                for (int i = 0; i < this.parameters.Length; ++i)
+                {
+                    object parameterName = null;
+                    foreach (var property in this.properties)
```
304NotModified: I have to agree this is really the downside of the same collection!

What do think about: internally split it to 2 collections (messageTemplateParameters, properties), and expose it as one collection to the outside?

src/NLog/StructuredEvents/TemplateParser.cs
```diff
@@ -0,0 +1,169 @@
...
+            }
+
+            return null;
+        }
+
+        private static bool TryParseStructuredTemplate(ref TemplateEnumerator enumerator, ref Literal[] literals, ref Hole[] holes, bool isPositional, int literalCount, int holeCount)
+        {
+            LiteralHole capture = enumerator.Current;
+            bool holeFound = capture.Literal.Skip != 0;
+            isPositional = isPositional && (!holeFound || (holeFound && capture.Hole.Index != -1));
```
304NotModified: `|| holeFound` is not needed here (it's always true) - thanks Resharper

src/NLog/StructuredEvents/TemplateParserException.cs
```diff
@@ -0,0 +1,58 @@
...
+{
+    /// <summary>
+    /// Error when parsing a template.
+    /// </summary>
+    public class TemplateParserException : Exception
+    {
+        /// <summary>
+        /// Current index when the error occurred.
+        /// </summary>
+        public int Index { get; set; }
```
304NotModified: we should add the orginal message template to this exception?

src/NLog/StructuredEvents/TemplateRenderer.cs
```diff
@@ -0,0 +1,149 @@
...
+            }
+
+            switch (captureType)
+            {
+                case CaptureType.Stringify:
+                    // TODO: we don't need to support format and alignment here?
+                    sb.Append('"').Append(value.ToString()).Append('"');
+                    break;
+                case CaptureType.Serialize:
+                    Config.ConfigurationItemFactory.Default.JsonConverter.SerializeObject(value, sb);
```
304NotModified: I think we should split this.

e.g. we like to use semi-json for message templates, and json.net for real json in JsonLayout. What do you think? 

Maybe adding MessageTemplateSerializer property on ConfigurationItemFactory?

304NotModified:
> Yes it cannot be that complicated to find a lock-free-container with fast lookup, fast insert, fast cleanup and minimum allocation :)

image

304NotModified:
> Before you start on implementing this caching business, then we might want to create another test, that pushes the concurrency a little (Closer to what a web-application might do):

It's a bit hard to add commits on another fork ;)

src/NLog/LogEventInfo.cs
snakefoot: `logMessageFormatter.HasProperties` checks if the MessageFormatter can produce-properties.

It is an optimization for the case where structured-logging has been disabled (or the LogEventInfo doesn't match the profile).

src/NLog/ILogMessageFormatter.cs
snakefoot: `HasProperties` could probably be translated into `CanContainProperties`. Message-template is needed for hole-name-properties to be produced.

src/NLog/ILogMessageFormatter.cs
snakefoot: Still proof-of-concept code-state :)

src/NLog/LogEventInfo.cs
snakefoot: The TemplateParser could probably generate a "Hole"-dictionary (Wraps a readonly array). And this "Hole"-dicationary, then has the ability to internally allocate a real dictionary if wanting to add/override properties. Then "just" have to make an artificial enumerator.

304NotModified: Sounds good to me!

src/NLog/Layouts/JsonLayout.cs
snakefoot: Yes it has now been removed.

snakefoot:
Made a concurrency test of Serilog and NLog, where I have 6 threads doing random logging between 2000 log-messages at 5 log-levels.

| Test Name  | Time (ms) | Msgs/sec  | GC2 | GC1 | GC0 | CPU (ms) | Mem (MB) |
| NLog-6T   |    30.309 |    79.181 |   2 | 103 | 205 |  107.250 |  172,246 |

| Test Name  | Time (ms) | Msgs/sec  | GC2 | GC1 | GC0 | CPU (ms) | Mem (MB) |
| Serilog-6T  |    71.380 |    33.622 |   7 |  78 | 479 |  279.859 |  160,824 |

My machine only has 4 real cores. So also tried the test with only 4 cores:

| Test Name  | Time (ms) | Msgs/sec  | GC2 | GC1 | GC0 | CPU (ms) | Mem (MB) |
| NLog-4T    |    18.213 |   131.772 |   3 |  67 | 132 |   63.718 |  172,730 |

| Test Name  | Time (ms) | Msgs/sec  | GC2 | GC1 | GC0 | CPU (ms) | Mem (MB) |
| Serilog-4T    |    47.075 |    50.981 |  19 |  73 | 255 |  182.359 |  148,590 |

This is the Serilog-config used:

```
            var serilogConfig = new Serilog.LoggerConfiguration().MinimumLevel.Verbose()
                .WriteTo.Logger(config => config
                    .MinimumLevel.Verbose()
                    .Filter.ByIncludingOnly(e => e.Level == Serilog.Events.LogEventLevel.Verbose)
                    .WriteTo.Async(a => a.File(@"C:\Temp\Log\Serilog-Verbose.txt")))
                .WriteTo.Logger(config => config
                    .MinimumLevel.Verbose()
                    .Filter.ByIncludingOnly(e => e.Level == Serilog.Events.LogEventLevel.Debug)
                    .WriteTo.Async(a => a.File(@"C:\Temp\Log\Serilog-Debug.txt")))
                .WriteTo.Logger(config => config
                    .MinimumLevel.Verbose()
                    .Filter.ByIncludingOnly(e => e.Level == Serilog.Events.LogEventLevel.Information)
                    .WriteTo.Async(a => a.File(@"C:\Temp\Log\Serilog-Info.txt")))
                .WriteTo.Logger(config => config
                    .MinimumLevel.Verbose()
                    .Filter.ByIncludingOnly(e => e.Level == Serilog.Events.LogEventLevel.Warning)
                    .WriteTo.Async(a => a.File(@"C:\Temp\Log\Serilog-Warn.txt")))
                .WriteTo.Logger(config => config
                    .MinimumLevel.Verbose()
                    .Filter.ByIncludingOnly(e => e.Level == Serilog.Events.LogEventLevel.Error)
                    .WriteTo.Async(a => a.File(@"C:\Temp\Log\Serilog-Error.txt")))
                .WriteTo.Logger(config => config
                    .MinimumLevel.Verbose()
                    .Filter.ByIncludingOnly(e => e.Level == Serilog.Events.LogEventLevel.Fatal)
                    .WriteTo.Async(a => a.File(@"C:\Temp\Log\Serilog-Fatal.txt")))
                    ;
```

Made some profiling of Serilog and NLog just see how the Serilog-Template-Cache worked out. It spent as much CPU-time on cache-lookups, as NLog spent on parsing. So the caching implemented in Serilog only helps when always logging the same 1000 messages over and over (with different args).

This is the result when doing random logging of 100 different messages using 4 threads (Allow the caching to work):

| Test Name  | Time (ms) | Msgs/sec  | GC2 | GC1 | GC0 | CPU (ms) | Mem (MB) |
| Serilog-4T     |    30.061 |    79.837 |   5 |  68 | 141 |  116.921 |  156,582 |

Unless we can find a magical cache-container, then I don't think NLog should go down that road. Prefer the cpu-overhead of re-parsing, instead of the increased allocations with blocking garbage-collections to follow.

snakefoot:
Have now implemented a very custom dictionary, that is just a combination of two linked lists (Holes + Meta). The linked-list works best for small collections below 100 items. This allows one to clear all the Holes while leaving the Meta-data untouched.

| Test Name  | Time (ms) | Msgs/sec  | GC2 | GC1 | GC0 | CPU (ms) | Mem (MB) |
| NLog-4T  |    12.585 |   190.693 |   1 |  29 |  58 |   46.359 |  170,984 |

304NotModified:
Great work! I'm a bit busy this week but will try to check it in depth soon. Would be great if I could also run the tests.

snakefoot:
> Would be great if I could also run the tests.

I have made a lot of hacks to the PerformanceTest-project, and added dependency to Serilog. Not sure I want to add those changes to a PR.

Maybe also a good idea, that a different set of eyes conjure up a performance test. Maybe you will find something that I didn't

304NotModified:
> I have made a lot of hacks to the PerformanceTest-project, and added dependency to Serilog. Not sure I want to add those changes to a PR.

In another branch sounds good to me.

snakefoot:
> In another branch sounds good to me.

Well now you can have some fun with this random code (Modifications to NLogPerformance-project):

[NLogPerformance.zip](https://github.com/NLog/NLog/files/1157393/NLogPerformance.zip)

snakefoot:
@304NotModified The changes made to LogEventInfo (new dictionary, new formatter) will probably be binary breaking, so it is not possible have NLog 4.4 running side-by-side NLog 4.5.

Maybe the NLog 4.5 and NLog 5.0 should be merged together, and just become NLog 5.0 ?

src/NLog/ILogMessageFormatter.cs
snakefoot: Now moved to seperate files.

src/NLog/ILogMessageFormatter.cs
snakefoot: Have now changed to re-use the StringBuilder that we have already acquired from the internal StringBuilder-pool.

snakefoot:
> In another branch sounds good to me.

Well now you can have some fun with this random code (Modifications to NLogPerformance-project):

[NLogPerformance.zip](https://github.com/NLog/NLog/files/1168266/NLogPerformance.zip)

src/NLog/LogEventInfo.cs
snakefoot: Have change the logic, so it now forces a parsing of the message-template.

src/NLog/StructuredEvents/TemplateParser.cs
snakefoot: Remove the code, and instead introduced new TemplateRenderer.Render()

snakefoot:
@304NotModified We might also need to add some restriction in the Json-Serializer (And Value-Renderer), so if they get hold of a ienumerable/idictionary that contains more than 100.000 items, then it will not try and serialize them all.

Can also see that the Value-Renderer does not have a cyclic-loop-level-restriction implemented (Like the Json-Serializer has).

304NotModified:
@snakefoot 

I'm sorry but I haven't found time to test this. Currently very busy with moving to a new house. (and got some delays)

snakefoot:
@304NotModified Don't sweat. Houses are never easy, and always gives delays :) And having a roof is usually more important than having a git-repository ;)

snakefoot:
@304NotModified Considering to change the implementation of the parameter-holes-dictionary into a linked list. The Properties-dictionary remains a union of both. Then create a new property on LogEventInfo allowing one to get/set the ParameterNames/ParameterKeys/ParameterHoles-ienumerable.

This will allow one to get the parameter-names directly related to the log-message, and in the correct order so it matches the LogEventInfo-Parameters-array.

nblumhardt:
> Made a concurrency test of Serilog and NLog, where I have 6 threads doing random logging between 2000 log-messages at 5 log-levels.

Just a quick note, `WriteTo.Logger` has a performance penalty in Serilog, so it's probably not the best basis for the comparison. Not sure it's really important (looks like your perf is really healthy!) but the config below is closer to how Serilog would work for this particular scenario:

```csharp
new LoggerConfiguration()
    .MinimumLevel.Verbose()
    .WriteTo.Async(writeTo =>
        writeTo.Map(le => le.Level, (level, wt) =>
            wt.File(@"C:\Temp\Log\Serilog-" + level + ".txt")))
    .CreateLogger();
```

Cheers!

snakefoot:
> @nblumhardt Just a quick note, WriteTo.Logger has a performance penalty in Serilog, so it's probably not the best basis for the comparison

Thank you for the suggestion. When having merged with main-branch again, then I will run the performance test one more time.

src/NLog/Internal/ILogMessageFormatter.cs
```diff
@@ -0,0 +1,45 @@
...
+// 
+
+using System;
+
+namespace NLog.Internal
+{
+    internal interface ILogMessageFormatter
+    {
+        string FormatMessage(LogEventInfo logEvent);
+        ILogMessageFormatter SetFormatProvider(IFormatProvider formatProvider);
```
304NotModified: probably a good reasons, but I don't see it now. Why is there a setter method instead of a setter? (I see the different return type)

src/NLog/Internal/LogMessageTemplateFormatter.cs
```diff
@@ -0,0 +1,139 @@
...
+        {
+            if (HasProperties(logEvent))
+            {
+                // Prevent multiple layouts on different targets to render the same properties
+                lock (logEvent)
+                {
+                    using (var builder = _builderPool.Acquire())
+                    {
+                        var holeProperties = logEvent.HoleProperties;
+                        System.Collections.Generic.IDictionary<object, object> holeDictionary = holeProperties;
```
304NotModified: Isn't this slower because of the interface calls? Or do you think the compiler optimizes this? 

(I think the interface is unneeded because `(Internal.PropertiesDictionary)holeDictionary` below)

src/NLog/Internal/MruCache.cs
```diff
@@ -71,57 +71,79 @@ public bool TryAddValue(TKey key, TValue value)
...
+        private void PruneCache()
+        {
+            // There 3 priorities:
+            //  - High Priority - Non-Virgins with version very close to _currentVersion
+            //  - Medium Priority - Version close to _currentVersion
+            //  - Low Priority - Old-Virgins
+            // Make 3 sweeps:
+            //  - Kill all the old ones
+            //  - Kill all the virgins
+            //  - Slaughterhouse
```
304NotModified: 😈

src/NLog/Internal/StringBuilderExt.cs
```diff
@@ -150,14 +150,18 @@ public static void CopyToStream(this StringBuilder builder, MemoryStream ms, Enc
 #if !SILVERLIGHT
             if (transformBuffer != null)
             {
+                int charCount = 0, byteCount = 0;
```
304NotModified: please move the bytecount to a new declaration (consistent code style)

src/NLog/StructuredEvents/Hole.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        public Hole(string name, string format, CaptureType captureType, int position, int alignment)
+        {
+            Name = name;
+            Format = format;
+            CaptureType = captureType;
+            Index = (short)position;
+            Alignment = (short)alignment;
```
304NotModified: maybe it's more transparent to change the parameter type to short? (2 times)

src/NLog/StructuredEvents/TemplateEnumerator.cs
```diff
@@ -0,0 +1,381 @@
...
+
+                            break;
+                        }
+                }
+
+                format += ReadUntil(TextDelimiters);
+            }
+        }
+
+        private int ParseAlignment()
```
304NotModified: As we don't use this code for string.format, I think we could drop this. What do you think?

src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -1230,6 +1232,26 @@ private void ParseTimeElement(NLogXmlElement timeElement)
             TimeSource.Current = newTimeSource;
         }
 
+        private void ParseTemplateFormatValue(string templateFormatValue)
+        {
+            templateFormatValue = (templateFormatValue ?? string.Empty).Trim().ToLower();
+            if (templateFormatValue == "yes" || templateFormatValue == "true")
```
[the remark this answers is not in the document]
snakefoot: No opinion

src/NLog/StructuredEvents/Hole.cs
snakefoot: No opinion. Just using your code

src/NLog/StructuredEvents/TemplateEnumerator.cs
snakefoot: No opinion. Just using your code

src/NLog/Config/ConfigurationItemFactory.cs
```diff
@@ -205,6 +205,43 @@ public IJsonConverter JsonConverter
         }
 
         /// <summary>
+        /// Perform mesage template parsing and formatting of LogEvent messages (True = Always, False = Never, Null = Auto Detect)
+        /// </summary>
+        public bool? MessageTemplateFormat
```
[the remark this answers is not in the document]
snakefoot: Fixed

snakefoot:
Repeated the performance tests with the updated code:

| Structured Logging  | Time (ms) | Msgs/sec  | GC2 | GC1 | GC0 | CPU (ms) | Mem (MB) |
| NLog-4T    |     6.403 |   780.815 |   1 |  16 |  32 |   22.437 |  187,012 |
| Serilog-4T    |    30.013 |   166.591 |   0 |  33 |  67 |  115.984 |  170,367 |
| Serilog-4T-100 |    28.320 |   176.549 |   0 |  21 |  43 |  111.828 |  111,457 |

| Positional Logging  | Time (ms) | Msgs/sec  | GC2 | GC1 | GC0 | CPU (ms) | Mem (MB) |
| NLog-4T    |     6.491 |   770.200 |   1 |  12 |  23 |   17.843 |  165,109 |
| Serilog-4T    |    29.541 |   169.251 |   1 |  17 |  42 |  116.656 |  121,934 |
| Serilog-4T-100    |    28.441 |   175.796 |   0 |  22 |  44 |  111.453 |  110,777 |

Using the updated Serilog-Configuration have certainly helped on the allocations, so it now matches NLog. But NLog is still doing good after all the merge.

nblumhardt:
_Serilog.Sinks.Async_ moves all work to a single background worker, so this'd explain the core utilization. Looks like NLog is doing really well! 👍

304NotModified:
great results @snakefoot !

src/NLog/Internal/StringBuilderExt.cs
snakefoot: Fixed

src/NLog/Internal/ILogMessageFormatter.cs
snakefoot: Code removed

src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -1230,6 +1234,28 @@ private void ParseTimeElement(NLogXmlElement timeElement)
             TimeSource.Current = newTimeSource;
         }
 
+        private void ParseFormatElement(NLogXmlElement formatElement)
+        {
+            formatElement.AssertName("templateformat");
+
+            string format = formatElement.GetRequiredAttribute("type").Trim().ToLower();
+            if (format == "yes" || format == "true")
```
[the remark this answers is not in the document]
snakefoot: Now only true/false

src/NLog/Internal/LogMessageTemplateFormatter.cs
snakefoot: Code removed

304NotModified:
@snakefoot do you prefer comments or git pushes from me?

Any way, please stop force push for now, i'm losing some comments after a force push. We can squash at the end.

snakefoot:
@304NotModified Whatever works for you. You can also just merge it, and make the changes you want afterwards.

I have been thinking about renaming the folder StructuredEvents into "MessageTemplate"

304NotModified:
Renaming is fine to me!

src/NLog/Internal/LogMessageTemplateFormatter.cs
```diff
@@ -0,0 +1,92 @@
...
+    using System.Collections.Generic;
+    using System.Globalization;
+    using NLog.StructuredEvents;
+
+    internal sealed class LogMessageTemplateFormatter : ILogMessageFormatter
+    {
+        public static readonly LogMessageTemplateFormatter DefaultAuto = new LogMessageTemplateFormatter(false);
+        public static readonly LogMessageTemplateFormatter Default = new LogMessageTemplateFormatter(true);
+        private static readonly StringBuilderPool _builderPool = new StringBuilderPool(Environment.ProcessorCount * 2);
+        private static readonly Func<IDictionary<object, object>> _dictionaryAllocator = () => new PropertiesDictionary();
```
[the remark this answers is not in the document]
snakefoot: Fixed

src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -595,6 +595,8 @@ private void ParseNLogElement(NLogXmlElement nlogElement, string filePath, bool
             InternalLogger.LogToConsoleError = nlogElement.GetOptionalBooleanAttribute("internalLogToConsoleError", InternalLogger.LogToConsoleError);
             InternalLogger.LogFile = nlogElement.GetOptionalAttribute("internalLogFile", InternalLogger.LogFile);
 
+            ParseTemplateFormatValue(nlogElement.GetOptionalAttribute("templateFormat", "auto"));
```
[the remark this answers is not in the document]
snakefoot: Fixed

src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -1230,6 +1232,26 @@ private void ParseTimeElement(NLogXmlElement timeElement)
             TimeSource.Current = newTimeSource;
         }
 
+        private void ParseTemplateFormatValue(string templateFormatValue)
+        {
+            templateFormatValue = (templateFormatValue ?? string.Empty).Trim().ToLower();
+            if (templateFormatValue == "1" || templateFormatValue == "true")
+            {
+                InternalLogger.Info("Message Template Format always enabled");
+                LogEventInfo.CurrentMessageFormatter = LogMessageTemplateFormatter.Default.MessageFormatter;
```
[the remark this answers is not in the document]
snakefoot: Fixed

src/NLog/LogEventInfo.cs
```diff
@@ -321,7 +355,7 @@ public IMessageTemplateParameters MessageTemplateParameters
         /// </summary>
         /// <remarks>This property was marked as obsolete on NLog 2.0 and it may be removed in a future release.</remarks>
         [Obsolete("Use LogEventInfo.Properties instead.  Marked obsolete on NLog 2.0", true)]
-        public IDictionary Context { get { return (this.properties ?? (this.properties = new PropertiesDictionary())).EventContext; } }
+        public IDictionary Context { get { return (Properties as PropertiesDictionary).EventContext; } }
```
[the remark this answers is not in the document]
snakefoot: Want to have the ProperitesDictionary-allocated, which is triggered by LogEventInfo.Properties-dictionary.

snakefoot:
Perform folder move with force-push to avoid too much noise. I have updated the code according to your comments, except the MessageTemplate.ISerializer-logic.

snakefoot:
@304NotModified I think NLog is about to have a lot of parameter/property-renderers. I'm thinking that we only need three:

- The IJsonConverter that should always render in JSON-format.
- The stringify object.ToString() with IFormatProvider support.
- The new IPropertyRenderer that just converts an object into custom string-format

It should be possible to specify in the MessageTemplate which of the 3 types should be used.

304NotModified:
have renamed the namespaces to MessageTemplates (plural) 

> Use plural namespace names if it is semantically appropriate.

https://msdn.microsoft.com/en-us/library/893ke618(v=vs.71).aspx

;)

304NotModified:
oops broke it. Don't fully understand the code then ;) Fixing it

304NotModified:
don't get it. LogEventTemplateShouldHaveProperties_even_when_changed fails locallly, even on dedd1e2!

304NotModified:
ow it only works when I run all tests from `LoggerTests.cs` at once.

304NotModified:
@snakefoot 

I don't see it where exactly is the side-effect, but LogEvent.HasMessageTemplateParameters has a side effect. Could you try to make this more explicit?

snakefoot:
> @304NotModified I don't see it where exactly is the side-effect, but LogEvent.HasMessageTemplateParameters has a side effect. 

Cannot see it either, but this call has a sideeffect:

`!ReferenceEquals(this.Message, this.FormattedMessage)`

It renders the formatted message, and captures properties (if any)

snakefoot:
@304NotModified Interesting. Your test `LogAfterAwait_CleanNamesOfAsyncContinuationsIsFalse_ShouldNotCleanNames` seems to be unstable also on AppVeyor.

304NotModified:
> @304NotModified Interesting. Your test LogAfterAwait_CleanNamesOfAsyncContinuationsIsFalse_ShouldNotCleanNames seems to be unstable also on AppVeyor.

Don't get it. But that's the risk of stack walking. Feel free to skip the test.

snakefoot:
@304NotModified Do we need more changes. Or can I perform a squash of the commits? (Btw. updated the Serilog performance numbers, after I disabled its lossy default behavior).

304NotModified:
I prefer to check it again, but don't think I will find big issues. Currently working on NLog.web which should be asp.net core 2 ready asap.

https://github.com/NLog/NLog/pull/2208#issuecomment-334583579 -> this will be another PR?

snakefoot:
> #2208 (comment) -> this will be another PR?

Maybe start as an issue, and when having decided on something, then create a PR. I have not seen any confirmation or rejection of any of my statements :)

304NotModified:
> I have not seen any confirmation or rejection of any of my statements :)

Unfortunately hadn't find time for that.

src/NLog/MessageTemplates/CaptureType.cs
```diff
@@ -36,7 +36,7 @@ namespace NLog.MessageTemplates
     /// <summary>
     /// The type of the captured hole
     /// </summary>
-    public enum CaptureType : byte
+    internal enum CaptureType : byte
```
304NotModified: 👍 less public = less maintenance

snakefoot:
@304NotModified Think I have found an okay solution for making it possible to provide an external override of the object-serialization.

snakefoot:
@304NotModified Squashed the commits and performed forced push.

304NotModified:
bam!

Sorry for the delay.

thx!

snakefoot:
@304NotModified No issue. Glad that it could be approved.

I have found some extra modification, that I would like to include. See #2319

ericnewton76:
This is cool.  Just found out about Serilog and got led here, I already use NLog all the time.

Does the target only support going to a Seq server?  Will it be possible to just dump into a file?

snakefoot:
@ericnewton76 NLog extracts the named parameters from the structured-message, and injects them into existing [collection of properties](https://github.com/NLog/NLog/wiki/EventProperties-Layout-Renderer), these can be used by any type of Target/Layout. Please post your comments and questions here instead: #2332

304NotModified:
PS for seq there is https://www.nuget.org/packages/NLog.Targets.Seq/1.0.0-dev-00032

PR #2214, opened 2017-07-13
Option to throw exceptions when installation failed (ThrowExceptions on InstallationContext)

rbarillec:
Currently, exceptions thrown during the installation of a target (e.g. database target) are caught by the [`LoggingConfiguration`](https://github.com/rbarillec/NLog/blob/master/src/NLog/Config/LoggingConfiguration.cs#L372) and logged internally. It can be useful for client code to know that the target installation has failed, so that they can for instance let a user know that logging has failed to start.

This change adds a `bool InstallationContext.ThrowExceptions` property which allows installation exceptions to be rethrown by the `LoggingConfiguration`. Note that for the exceptions to be thrown in the first place, `IgnoreFailures` needs to be set to `false` on both the `CommandInfo` and the `InstallationContext` (at least for the [`DatabaseTarget`](https://github.com/rbarillec/NLog/blob/master/src/NLog/Targets/DatabaseTarget.cs#L694)). If `IgnoreFailures` is set to true on either, then the `ThrowExceptions` property will have no effect.

### DIFF
--- a/tests/NLog.UnitTests/Targets/DatabaseTargetTests.cs
+++ b/tests/NLog.UnitTests/Targets/DatabaseTargetTests.cs
@@ -992,7 +992,7 @@ Dispose()
 
             // Use an in memory SQLite database
             // See https://www.sqlite.org/inmemorydb.html
-            var connectionString = String.Format("Uri=file::memory:;Version=3", databaseName);
+            var connectionString = "Uri=file::memory:;Version=3";
 
             LogManager.Configuration = CreateConfigurationFromString(
                 String.Format(nlogXmlConfig, GetSQLiteDbProvider(), connectionString)
--- a/tests/NLog.UnitTests/Targets/DatabaseTargetTests.cs
+++ b/tests/NLog.UnitTests/Targets/DatabaseTargetTests.cs
@@ -855,13 +855,7 @@ Dispose()
 
                 DatabaseTarget testTarget = new DatabaseTarget("TestSqliteTarget");
                 testTarget.ConnectionString = connectionString;
-                string dbProvider = "";
-#if MONO 
-                dbProvider = "Mono.Data.Sqlite.SqliteConnection, Mono.Data.Sqlite";
-#else
-                dbProvider = "System.Data.SQLite.SQLiteConnection, System.Data.SQLite";
-#endif
-                testTarget.DBProvider = dbProvider;
+                testTarget.DBProvider = GetSQLiteDbProvider();
 
                 testTarget.InstallDdlCommands.Add(new DatabaseCommandInfo()
                 {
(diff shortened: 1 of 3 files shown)


tests/NLog.UnitTests/Targets/DatabaseTargetTests.cs
```diff
@@ -855,13 +855,7 @@ public void SQLite_InstallAndLogMessageProgrammatically()
...
                 DatabaseTarget testTarget = new DatabaseTarget("TestSqliteTarget");
                 testTarget.ConnectionString = connectionString;
-                string dbProvider = "";
-#if MONO 
-                dbProvider = "Mono.Data.Sqlite.SqliteConnection, Mono.Data.Sqlite";
-#else
-                dbProvider = "System.Data.SQLite.SQLiteConnection, System.Data.SQLite";
-#endif
-                testTarget.DBProvider = dbProvider;
+                testTarget.DBProvider = GetSQLiteDbProvider();
```
rbarillec: Moved this code to a separate helper `GetSQLiteDbProvider()` so I could reuse it in my test.

I haven't tested that this still works in Mono.

tests/NLog.UnitTests/Targets/DatabaseTargetTests.cs
```diff
@@ -974,6 +973,61 @@ public void SQLite_InstallAndLogMessage()
...
+        private void SetupSqliteConfigWithInvalidInstallCommand(string databaseName)
+        {
+            var nlogXmlConfig = @"
+            <nlog xmlns='http://www.nlog-project.org/schemas/NLog.xsd'
+                  xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance' throwExceptions='false'>
+                <targets>
+                    <target name='database' xsi:type='Database' dbProvider='{0}' connectionstring='{1}' 
+                        commandText='insert into RethrowingInstallExceptionsTable (Message) values (@message);'>
+                        <parameter name='@message' layout='${{message}}' />
+                        <install-command ignoreFailures='false' text='THIS IS NOT VALID SQL;' />
```
rbarillec: I explicitly set `ignoreFailures` to `false`, but I wonder if leaving it out (i.e. using the default) would be better? That way the test would fail if the default ever changes and give us a hint that we may need to update the `ThrowExceptions` doc?

tests/NLog.UnitTests/Targets/DatabaseTargetTests.cs
```diff
@@ -974,6 +973,61 @@ public void SQLite_InstallAndLogMessage()
...
+                    </target>
+                </targets>
+                <rules>
+                    <logger name='*' writeTo='database' />
+                </rules>
+            </nlog>";
+
+            // Use an in memory SQLite database
+            // See https://www.sqlite.org/inmemorydb.html
+            var connectionString = String.Format("FullUri=file:{0}?mode=memory;Version=3", databaseName);
```
rbarillec: I find that in-memory SQLite databases are easier to use in tests (so long as we're not testing the file system): no clean up required, and much faster to use overall than a real file.

tests/NLog.UnitTests/Targets/DatabaseTargetTests.cs
304NotModified: :+1:

tests/NLog.UnitTests/Targets/DatabaseTargetTests.cs
304NotModified: Default would be better indeed!

tests/NLog.UnitTests/Targets/DatabaseTargetTests.cs
304NotModified: Looks like a good improvement

304NotModified:
Looks good!

Got on Travis: 

> Targets/DatabaseTargetTests.cs(1028,27): error CS0246: The type or namespace name 'SQLiteException' could not be found (are you missing a using directive or an assembly reference?)

tests/NLog.UnitTests/Targets/DatabaseTargetTests.cs
rbarillec: OK, will update.

rbarillec:
Yep, I hadn't noticed `SQLiteException` is `SqliteException` on Mono. I've just added a separate section for Mono. 

I know, that case change is a bit of a bummer, really.

tests/NLog.UnitTests/Targets/DatabaseTargetTests.cs
rbarillec: OK, this doesn't seem to be supported in the Mono version. I've changed it to `Uri=file::memory:` which looks like both Mono and .NET support.

rbarillec:
No problem. Thanks for your support with this too!

PR #2221, opened 2017-07-22
NestedDiagnosticsLogicalContext - Support remote AppDomain

snakefoot:


tests/NLog.UnitTests/Internal/AppDomainPartialTrustTests.cs
```diff
@@ -63,7 +63,12 @@ public void MediumTrustWithExternalClass()
                     // ClassUnderTest must extend MarshalByRefObject
                     AppDomain partialTrusted;
                     var classUnderTest = MediumTrustContext.Create<ClassUnderTest>(fileWritePath, out partialTrusted);
-                    classUnderTest.PartialTrustSuccess(times, fileWritePath);
+#if NET4_0 || NET4_5
+                    using (NLog.NestedDiagnosticsLogicalContext.Push("PartialTrust"))
```
304NotModified: Is this a breaking change in the output of the users?

snakefoot: No it is actually a bug-fix, as exception will be thrown when calling into another AppDomain after having pushed something on the NestedDiagnosticsLogicalContext.

304NotModified: Lol I was sleepy I guess, this is test code.

Issue #1952, opened 2017-02-11
BufferingWrapper support grouping of log messages towards WebService target


devJ0n:
**Type**:
Feature request

**NLog version**:
Latest

**Platform**:
.Net 3.5 / .Net 4.0 / .Net 4.5 / Mono 4 / Silverlight 4 / Silverlight 5 / Xamarin Android / Xamarin iOs / .NET Core

**Current NLog config** (xml or C#, if relevant)
```xml
<nlog>
...
<targets>
  <!-- Write logs to webservice -->
    <target name="logMessageWebServiceBuffered" xsi:type="BufferingWrapper" slidingTimeout="true" bufferSize="100" flushTimeout="60000">
      <target xsi:type="WebService" url="https://acme.com/api/log" protocol="JsonPost" encoding="UTF-8" includeBOM="false">
        <parameter name="Message"    layout="${message}" />
        <parameter name="DateTime"    layout="${event-properties:item=DateTime}" />
        <parameter name="SystemName" layout="${event-properties:item=SystemName}" />
        <parameter name="ApplicationName" layout="${event-properties:item=ApplicationName}" />
        <parameter name="JobName" layout="${event-properties:item=JobName}" />
        <parameter name="Status" layout="${event-properties:item=Status}" />
      </target>
    </target>
</targets>
...
</nlog>
```
 
 - Why do we need it?
Currently logging towards an webservice (REST API) using JSON POST is very chatty and the recieving REST API is easily overcumbered with HTTP requests. The BufferingWrapper make no difference when used togheter with the WebService target other than enablinc async logging and then sending for example 100 HTTP requests at once.

A nice performance improvement would be make the BufferingWrapper achive grouping of log messages very much like the BufferingWrapper+Mail target currently gives us.

Maybe not all protocols on the WebService target are able to use this enhancement of the BufferingWrapper like the HttpGet. Other protocols like the JsonPost and XmlPost could very much implement this grouping logic.

A HTTP request for the JsonPost protocol could look like this:
```
POST /api/log HTTP/1.1
Host: acme.com
Content-Type: application/json

[
  {
    'message': 'Order was created',
    'dateTime': 2017-02-08T18:25:42.437Z,
    'systemName': 'OrderSystem',
    'applicationName': 'AspNet MVC WebOrderClient',
    'jobName': 'OrderCreated',
    'status': 83
  },
  {
    'message': 'Order was confirmed',
    'dateTime': 2017-02-08T18:25:43.511Z,
[...]

PR #2223, opened 2017-07-23
WebServiceTarget - Write multiple LogEvents in single request

snakefoot:
Added Group-Layouts to merge multiple LogEvents into single bulk-write. See also #1952

### DIFF
--- a/src/NLog/Targets/MethodCallParameter.cs
+++ b/src/NLog/Targets/MethodCallParameter.cs
@@ -111,5 +111,25 @@ namespace NLog.Targets
         /// <docgen category='Parameter Options' order='10' />
         [RequiredParameter]
         public Layout Layout { get; set; }
+
+        /// <summary>
+        /// Parameter can combine multiple LogEvents into a single parameter value
+        /// </summary>
+        public bool EnableGroupLayout { get; set; }
+
+        /// <summary>
+        /// Group Header when combining multiple LogEvents into a single parameter value (<see cref="EnableGroupLayout"/>)
+        /// </summary>
+        public Layout GroupHeaderLayout { get; set; }
+
+        /// <summary>
+        /// Group Item Separator when combining multiple LogEvents into a single parameter value (<see cref="EnableGroupLayout"/>)
+        /// </summary>
+        public Layout GroupItemSeparatorLayout { get; set; }
+
+        /// <summary>
+        /// Group Footer when combining multiple LogEvents into a single parameter value (<see cref="EnableGroupLayout"/>)
+        /// </summary>
+        public Layout GroupFooterLayout { get; set; }
     }
 }
--- a/src/NLog/Targets/MethodCallTargetBase.cs
+++ b/src/NLog/Targets/MethodCallTargetBase.cs
@@ -36,6 +36,7 @@ namespace NLog.Targets
     using System;
     using System.Collections.Generic;
     using System.Globalization;
+    using System.Text;
     using NLog.Common;
     using NLog.Config;
     using NLog.Internal;
@@ -46,6 +47,8 @@ namespace NLog.Targets
     /// </summary>
     public abstract class MethodCallTargetBase : Target
     {
+        private const int MaxGroupRenderSingleBufferLength = 128 * 1024;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="MethodCallTargetBase" /> class.
         /// </summary>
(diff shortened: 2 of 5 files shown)


src/NLog/Targets/MethodCallTargetBase.cs
```diff
@@ -67,16 +68,72 @@ protected MethodCallTargetBase()
...
+                if (param.GroupHeaderLayout != null)
+                    param.GroupHeaderLayout.RenderAppendBuilder(logEvents[0].LogEvent, sb);
+                for (int x = 0; x < logEvents.Count; ++x)
+                {
+                    if (x != 0 && param.GroupItemSeparatorLayout != null)
+                        param.GroupItemSeparatorLayout.RenderAppendBuilder(logEvents[x].LogEvent, sb);
+
+                    if (param.Layout != null)
+                    {
+                        if (sb.Length < 128 * 1024)
```
304NotModified: could we create a `const` for this? To prevent "magic numbers"

src/NLog/Targets/MethodCallParameter.cs
```diff
@@ -111,5 +111,25 @@ public MethodCallParameter(string name, Layout layout, Type type)
         /// <docgen category='Parameter Options' order='10' />
         [RequiredParameter]
         public Layout Layout { get; set; }
+
+        /// <summary>
+        /// Parameter can combine multiple LogEvents into a single parameter value
+        /// </summary>
+        public bool GroupLayoutActive { get; set; }
```
304NotModified: AFAIK the most targets in NLog uses the `Enable...` names.

So EnableGrouping?

src/NLog/Targets/WebServiceTarget.cs
```diff
@@ -199,6 +200,162 @@ public WebServiceTarget(string name) : this()
...
+                }
+                else
+                {
+                    DoGroupInvoke(logEvents, ArrayHelper.Empty<string>());
+                }
+            }
+        }
+        private SortHelpers.KeySelector<AsyncLogEventInfo, string[]> convetToHeaderArrayDelegate;
+
+        string[] ConvetToHeaderArray(LogEventInfo logEvent)
```
304NotModified: typo, => `convert`

src/NLog/Targets/WebServiceTarget.cs
```diff
@@ -199,6 +200,162 @@ public WebServiceTarget(string name) : this()
 
         private readonly AsyncOperationCounter pendingManualFlushList = new AsyncOperationCounter();
 
+        private bool enableParameterGroups;
+        private bool onlyParameterGroups;
```
304NotModified: why do we need this?

src/NLog/Targets/WebServiceTarget.cs
```diff
@@ -199,6 +200,162 @@ public WebServiceTarget(string name) : this()
...
+                return true;
+            }
+
+            public int GetHashCode(TValue[] obj)
+            {
+                int hashCode = obj.Length.GetHashCode();
+                for (int i = 0; i < obj.Length; ++i)
+                {
+                    if (obj[i] != null)
+                        hashCode = hashCode ^ obj[i].GetHashCode();
```
304NotModified: is this generated ? (with R# for example?)

snakefoot: No written by my own two hands.

src/NLog/Targets/WebServiceTarget.cs
snakefoot: Fixed

src/NLog/Targets/WebServiceTarget.cs
snakefoot: Have added a comment, explaining why.

src/NLog/Targets/MethodCallTargetBase.cs
snakefoot: Fixed

src/NLog/Targets/WebServiceTarget.cs
304NotModified: AFAIK it good practice to use a prime number, see for example http://www.aaronstannard.com/overriding-equality-in-dotnet/

Ps. Why aren't you using R#? I'm convinced that your good enough to skip implementing simple stuff. :)

snakefoot: Well I find Visual Studio really slow and annoying, and RSharper really doesn't help in that area. But maybe I start using it when I have grown a little wiser :). I'm guessing the that the GetHashCode()-methods already uses a prime-number (not using the raw-values). I know that XOR will give many collisions if the parameter-array-values are all the same, but I'm guessing they will not be. And the GetHashCode() doesn't have to be that optimal as there most likely be one or two groups at a time.

304NotModified:
Thanks!

Issue #107, opened 2012-10-07
persistent SequenceId resp. Counters


ghost:
# issue 1

the sequence id was introduced to help merging sort files,

for the same reason they should be resident/persistent

ASP.Net app running on ISS are stopped after reaching a timeout (default 20mins) when not receiving any requests.

```
2012-05-31 01:53:34.8890|INFO|13|message ....

2012-05-31 02:14:40.8908|INFO|14|session terminated pivxphxch....

2012-05-31 02:14:56.8388|INFO|15|application_end event fired

2012-05-31 03:01:44.0970|INFO|1|application_start event fired

2012-05-31 03:01:44.2600|INFO|2|new session started 84.74.27.121 s2ghwv....

2012-05-31 03:01:49.6772|INFO|3|message ....
```
# issue 2

at the moment the SequenceId is only available if Layout log4jxmlevent and includeNLogData=true. can we have the sequenceId also for other simple formats too?

304NotModified:
issue 1 fixed in #2226 

issue 2 -> https://github.com/NLog/NLog/issues/2373

PR #2226, opened 2017-07-25
GuidLayoutRenderer - Unique guid per LogEvent

### LINKED ISSUE #60: Add a Unique ID per log event visible by a Layout Renderer
<p>When logging to various targets there isn&#39;t a specific way (that I&quot;m aware of) to link the log events together. The ${guid} layout renderer generates a unique id for each layout. So could we get a single unique reference id that all targets could expose?</p>

<p>&nbsp;</p>

<p>I&#39;d be happy to look into an implementation of this and provide a patch if needed.</p>

snakefoot:
LogEvent will output the same Guid for all Targets. Resolves #60 (And maybe #107)

### DIFF
--- a/src/NLog/LayoutRenderers/GuidLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/GuidLayoutRenderer.cs
@@ -59,13 +59,38 @@ namespace NLog.LayoutRenderers
         public string Format { get; set; }
 
         /// <summary>
+        /// Generate the Guid from the NLog LogEvent (Will be the same for all targets)
+        /// </summary>
+        [DefaultValue(false)]
+        public bool GeneratedFromLogEvent { get; set; }
+
+        /// <summary>
         /// Renders a newly generated GUID string and appends it to the specified <see cref="StringBuilder" />.
         /// </summary>
         /// <param name="builder">The <see cref="StringBuilder"/> to append the rendered data to.</param>
         /// <param name="logEvent">Logging event.</param>
         protected override void Append(StringBuilder builder, LogEventInfo logEvent)
         {
-            builder.Append(Guid.NewGuid().ToString(this.Format));
+            if (this.GeneratedFromLogEvent)
+            {
+                int hashCode = logEvent.GetHashCode();
+                short b = (short)((hashCode >> 16) & 0XFFFF);
+                short c = (short)(hashCode & 0XFFFF);
+                long zeroDateTicks = LogEventInfo.ZeroDate.Ticks;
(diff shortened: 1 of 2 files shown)


src/NLog/LayoutRenderers/GuidLayoutRenderer.cs
```diff
@@ -59,13 +59,34 @@ public GuidLayoutRenderer()
...
+            {
+                byte d = (byte)(LogEventInfo.ZeroDate.Ticks >> 56);
+                byte e = (byte)(LogEventInfo.ZeroDate.Ticks >> 48);
+                byte f = (byte)(LogEventInfo.ZeroDate.Ticks >> 40);
+                byte g = (byte)(LogEventInfo.ZeroDate.Ticks >> 32);
+                byte h = (byte)(LogEventInfo.ZeroDate.Ticks >> 24);
+                byte i = (byte)(LogEventInfo.ZeroDate.Ticks >> 16);
+                byte j = (byte)(LogEventInfo.ZeroDate.Ticks >> 8);
+                byte k = (byte)(LogEventInfo.ZeroDate.Ticks);
+                builder.Append(new Guid(logEvent.SequenceID, (short)logEvent.TimeStamp.Month, (short)logEvent.TimeStamp.Year, d, e, f, g, h, i, j, k).ToString(this.Format));
```
304NotModified: Clever approach! 

But is this really a guid? (Global, unique)

snakefoot: It is global unique identifier of a LogEvent within the running application. But probably not global in the universe.

src/NLog/LayoutRenderers/GuidLayoutRenderer.cs
```diff
@@ -59,13 +59,38 @@ public GuidLayoutRenderer()
         public string Format { get; set; }
 
         /// <summary>
+        /// Generate the Guid from the NLog LogEvent (Will be the same for all targets)
+        /// </summary>
+        [DefaultValue(false)]
+        public bool LogEventGuid { get; set; }
```
304NotModified: What about "basedOnLogEvent"? Or "GeneratedFromLogEvent"

304NotModified: @snakefoot ? Your opinion please :)

snakefoot: Changed to GeneratedFromLogEvent

PR #2229, opened 2017-07-29
Fix: ReconfigExistingLoggers sometimes throws an Exception #2228

### LINKED ISSUE #2228: ReconfigExistingLoggers sometimes throws an Exception
**Type** (choose one):

- Bug

**NLog version**: (e.g. 4.2.3) 4.4.11

**Platform**: .Net 3.5 / .Net 4.0 / .Net 4.5 / Mono 4 / Silverlight 4 / Silverlight 5 / Xamarin Android / Xamarin iOs / .NET Core 

The fix I contributed for #2130,

```List<WeakReference> loggerReferences = new List<WeakReference>(loggerCache.Values.ToList());```

Sometimes throws an Exception related to the new List being too small to be initialized with the ToList() of loggerCache.Values.  My CI has caught it a few times but I don't recall the exact Exception and can't reproduce it easily as it is concurrency related.  

I investigated the problem when I first saw it and it stems from the fact that the destination list is constructed of size equal to the capacity of the ToList, and if the initializing list grows (via another thread) before the assignment is made the Exception occurs.

I will submit a PR shortly with the following change to this line:
```List<WeakReference> loggerReferences = loggerCache.Values.ToList();```

This was suggested by @snakefoot in the discussion of my original PR and I regret not implementing the change then.

### PR DESCRIPTION
**Pull request by jpdillingham**
Fixes #2228

### DIFF
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -1349,7 +1349,6 @@ namespace NLog
                 // TODO: Test if loggerCache.Values.ToList<Logger>() can be used for the conversion instead.
                 List<Logger> values = new List<Logger>(loggerCache.Count);
 
-                //new list for prevent InvalidOperationException on Travis/Mono.
                 List<WeakReference> loggerReferences = loggerCache.Values.ToList();
 
                 foreach (WeakReference loggerReference in loggerReferences)
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -1349,9 +1349,7 @@ namespace NLog
                 // TODO: Test if loggerCache.Values.ToList<Logger>() can be used for the conversion instead.
                 List<Logger> values = new List<Logger>(loggerCache.Count);
 
-                List<WeakReference> loggerReferences = loggerCache.Values.ToList();
-
-                foreach (WeakReference loggerReference in loggerReferences)
+                foreach (WeakReference loggerReference in loggerCache.Values)
                 {
                     Logger logger = loggerReference.Target as Logger;
                     if (logger != null)
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -1348,7 +1348,7 @@ namespace NLog
                 List<Logger> values = new List<Logger>(loggerCache.Count);
 
                 //new list for prevent InvalidOperationException on Travis/Mono.
-                List<WeakReference> loggerReferences = new List<WeakReference>(loggerCache.Values.ToList());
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by snakefoot** [comment]
If the problem is caused by the Values-Collection-Enumerator is detecting that the dictionary has been modified while enumerating, then the correct fix is to make the clone under a lock (See alternative solution here: #2131)

**Comment by jpdillingham** [comment]
I've added a lock around the references to loggerCache.  I was hesitant to do this from the start because this is my first time looking at the code and I was concerned about reciprocal effects.

**Comment by snakefoot** [comment]
Not sure I understand why you have added a new lock-object. There is already LogFactory.syncRoot which is used all other places before accessing/modifying the loggerCache.

Why not make the change that I initially suggested here: #2131 ?

**Comment by jpdillingham** [comment]
I see what you mean now.  Apologies, I was trying to fix the issue without getting into the weeds.

**Comment by snakefoot** [comment]
@jpdillingham Guess you can remove the stale comment:

`//new list to avoid "Collection was modified; enumeration operation may not execute"`

**Comment by 304NotModified** [comment]
Could you rebase this one on branch release-4.4.12? (and change this PR, not sure if you could do that)

**Comment by jpdillingham** [comment]
I believe this will do it.

File: src/NLog/LogFactory.cs [hunk 1]
```diff
@@ -1348,7 +1350,7 @@ private IEnumerable<Logger> GetLoggers()
                 List<Logger> values = new List<Logger>(loggerCache.Count);
 
                 //new list for prevent InvalidOperationException on Travis/Mono.
```
304NotModified: this is also an old comment?

jpdillingham: I believe that is irrelevant now due to the upstream lock.

snakefoot: @jpdillingham If removing the comment, then you should also remove the ToList()-operation.

jpdillingham:
Just to follow up, I've made all of the requested changes and this should be ready to merge whenever you're ready.

304NotModified:
Thanks for the reminder. @snakefoot this change is also OK for you isn't?

snakefoot:
Yes it looks good to me


Sent from my Samsung device


304NotModified:
Online!

https://www.nuget.org/packages/NLog/4.4.12

Thanks @jpdillingham !

dtatcox:
I just got this same exception from ReconfigExistingLoggers and I am 100% sure i'm on 4.4.12.

304NotModified:
Please create a new issue, including stacktrace. Thx!

PR #2233, opened 2017-08-01
TraceTarget - Configure to always use Trace.WriteLine

### LINKED ISSUE #1968: NLog prepends unspecified data to Trace target.
**Type**: Bug

**NLog version**: 4.4.0

**Platform**:  .Net 4.6

```
      autoReload="true"
      throwExceptions="false"
      internalLogLevel="Trace" internalLogFile="c:\temp\nlog-internal.log">

  <variable name="VisualStudioTrace" value="TEST [${processtime}] ${message}"/>

  <targets>
    <target xsi:type="Trace"
            name="vstrace"
            layout="${VisualStudioTrace}"/>
  </targets>

  <rules>
    <logger name="*" levels="Trace,Debug,Info,Warn,Error" writeTo="vstrace" />
  </rules>
</nlog>


```

In case of a BUG:

- What is the current result?
```
TEST [00:00:00.069] Trace Log
TEST [00:00:00.089] Debug Log
ByteLogging.vshost.exe Information: 0 : TEST [00:00:00.089] Info Log
ByteLogging.vshost.exe Warning: 0 : TEST [00:00:00.100] Warning Log
ByteLogging.vshost.exe Error: 0 : TEST [00:00:00.130] Exception Log
```
```
TEST [00:00:00.069] Trace Log
TEST [00:00:00.089] Debug Log
TEST [00:00:00.089] Info Log
TEST [00:00:00.100] Warning Log
TEST [00:00:00.130] Exception Log
```
Yes it appeared to be working correctly
This is not an exception but logs from the **nlog-internal.log**
```
2017-02-15 15:57:14.6361 Debug Setting 'TraceTarget.name' to 'vstrace'
2017-02-15 15:57:14.6361 Debug Setting 'TraceTarget.layout' to 'TEST [${processtime}] ${message}'
2017-02-15 15:57:14.7501 Trace   Scanning SimpleLayout ''TEST [${processtime}] ${message}''
2017-02-15 15:57:14.8082 Trace Initializing 'TEST [${processtime}] ${message}'
2017-02-15 15:57:14.8082 Trace Scanning SimpleLayout ''TEST [${processtime}] ${message}''
```
not that I am aware of
I haven't tried yet.
Not sure how to write a unit test, sorry.

in case of a FEATURE REQUEST:
 
 - Why do we need it?
A user should be allowed to configure the default log messages to how they would like them to appear.

snakefoot:
Resolves  #1968

### DIFF
--- a/src/NLog/Targets/TraceTarget.cs
+++ b/src/NLog/Targets/TraceTarget.cs
@@ -94,7 +94,8 @@ namespace NLog.Targets
         /// <summary>
         /// Writes the specified logging event to the <see cref="System.Diagnostics.Trace"/> facility.
         /// 
-        /// Redirects the log message depending on <see cref="LogLevel"/>:
+        /// Redirects the log message depending on <see cref="LogLevel"/> and  <see cref="RawWrite"/>. 
+        /// When <see cref="RawWrite"/> is <c>false</c>:
         ///  - <see cref="LogLevel.Fatal"/> writes to <see cref="Trace.Fail(string)" />
         ///  - <see cref="LogLevel.Error"/> writes to <see cref="Trace.TraceError(string)" />
         ///  - <see cref="LogLevel.Warn"/> writes to <see cref="Trace.TraceWarning(string)" />
--- a/src/NLog/Targets/TraceTarget.cs
+++ b/src/NLog/Targets/TraceTarget.cs
@@ -37,6 +37,7 @@
 
 namespace NLog.Targets
 {
+    using System.ComponentModel;
     using System.Diagnostics;
 
     /// <summary>
@@ -62,6 +63,12 @@ namespace NLog.Targets
     public sealed class TraceTarget : TargetWithLayout
     {
         /// <summary>
+        /// Always use <see cref="Trace.WriteLine(string)"/> independent of <see cref="LogLevel"/>
+        /// </summary>
+        [DefaultValue(false)]
+        public bool RawWrite { get; set; }
+
+        /// <summary>
         /// Initializes a new instance of the <see cref="TraceTarget" /> class.
         /// </summary>
         /// <remarks>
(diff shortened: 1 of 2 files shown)


src/NLog/Targets/TraceTarget.cs
```diff
@@ -62,6 +63,12 @@ namespace NLog.Targets
     public sealed class TraceTarget : TargetWithLayout
     {
         /// <summary>
+        /// Always use <see cref="Trace.WriteLine(string)"/> independent of <see cref="LogLevel"/>
+        /// </summary>
+        [DefaultValue(false)]
+        public bool WriteLine { get; set; }
```
304NotModified: I'm doubting if this is a good name. With WriteLine false I would expect Write without newline (for example in the stringBuilder)

snakefoot: Please come with other ideas. ForceWriteLine  / AlwaysWriteLine ?

304NotModified: WithoutPrefix?

snakefoot: Changed to IgnoreLogLevel

304NotModified: mmmm, sounds like IgnoreLogLevel=error.

Don't want to be picky, do you have another idea?

What about EnablePrefix, DontSerializeLogLevel, SkipLogLevelPrefix, RawLayoutOnly,...

304NotModified: Or maybe it's a "header" or "trace event type", see https://stackoverflow.com/questions/26350620/difference-between-trace-write-and-trace-traceinformation

snakefoot: I like IgnoreLogLevel. The link you just provide says that when using TraceInformation, then it converts to TraceEvent with info-log-level. So it actually ignores LogLevel when just using Trace.WriteLine.

snakefoot: Well what about LogLevelNeutral ?

304NotModified: - https://github.com/NLog/NLog/issues/1968: uses "prepend"
- https://github.com/NLog/NLog/issues/2234 uses "prefix"

I know this is depends of the listener, but still I prefer a name with prepend or prefix in it. What about PrependLogLevel?

snakefoot: Not sure why you like "prepend" and "prefix". It is just the Visual Studio Listener that performs some prefixing, it could just as well add suffix depending on LogLevel provided to the Trace-call.

304NotModified: What about: writeMessageOnly, or rawWrite?

snakefoot: Now it is valled RawWrite

304NotModified:
OK thanks, will merge it after succesfull build

PR #2241, opened 2017-08-04
EventPropertiesLayoutRenderer - Support Json Format

snakefoot:
Guess it makes most sense when combined with #2208 (Or #2244)


304NotModified:
Updated the docs, feel free to change.

Issue #2249, opened 2017-08-16
${exception} - separator=${newline} won't work


304NotModified:
Bug:

```
${exception:format=tostring,data:maxInnerExceptionLevel=10:separator=${newline}} 
```

wont work

but this works:

```
${exception:format=tostring,data:maxInnerExceptionLevel=10:separator=\r\n}
```

See https://github.com/NLog/NLog/issues/2248

jojosardez:
Hi 304,
I am able to replicate it. I believe I have fixed it locally. Can I commit it to my fork and raise a pull request?

304NotModified:
Yes that would be great! Please also include an unit test, thanks!

jojosardez:
OK. I'll refine it first and will let you know once its ready.

jojosardez:
Hi @304NotModified ,
The pull request is now ready: https://github.com/NLog/NLog/pull/2256
Thanks!

304NotModified:
Investigated this, this isn't a bug as the type of the parameter is `string` and not `Layout`

xmedeko:
How to make platform compatible line separator? I.e. for Linux `separator=\n` and for Windows `separator=\r\n`?

304NotModified:
I think the only way is to fix this issue (it's a breaking change), or your program needs some code for it.

snakefoot:
Discovered that there is different between:

- `<target type="file" filename="C:\NLog\Rocks" />`

And

- `${literal:inner=C:\NLog\Rocks}`

So when suddenly changing Exception.Separator from `string` to `Layout` then this will change meaning:

- `${exception:separatorr=\r\n}`

Instead of being Windows-newline, then it will translate into the actual string "\r\n". I kind of like that NLog 5.0 is still non-breaking in parsing of `NLog.config`.

304NotModified:
> where escapes are ignored

Isn't that we need an extra `\`? e.g 

   ${exception:format=message:maxInnerExceptionLevel=10:separator=\\t\\r\\n}

But of course, unfortunate for the breaking change.

304NotModified:
another trick is to decompose the `${newline}`  when parsing the file ;)

snakefoot:
> another trick is to decompose the ${newline} when parsing the file ;)

Thinking the same, since all use-cases has been about using `${newline}`.

snakefoot:
> but I think it should be now in LayoutParser, and that could have bigger side-effect

[...]

PR #2256, opened 2017-08-19
Fixed the parameter value parser (ParseParameterValue) #2249

jojosardez:
The reason why "separator=${newline}" doesn't work is because the parameter value parser isn't able to parse layout macros completely. Even after fixing the parser, it will end up displaying the layout macro tag \ name only without rendering so I've implemented rendering within the parameter value parser as well.

### DIFF
--- a/src/NLog/Layouts/LayoutParser.cs
+++ b/src/NLog/Layouts/LayoutParser.cs
@@ -213,13 +213,40 @@ namespace NLog.Layouts
             return nameBuf.ToString();
         }
 
-        private static string ParseParameterValue(SimpleStringReader sr)
+        private static string ParseParameterValue(ConfigurationItemFactory configurationItemFactory, SimpleStringReader sr)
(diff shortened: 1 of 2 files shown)


304NotModified:
Thanks! I'm very busy these days, but will try to review soon!

304NotModified:
looks good, but i have to double check as this is a part of the core of NLog

304NotModified:
Could you please resolve the conflict? Thanks!

304NotModified:
@jojosardez

First of all, thanks for this PR.

Unfortunately I cannot accept this change, as the current behavior is by-design. The parameters that should accept layoutrenders, should be of type `Layout`. If we change the types of the parameters, your test case works! (have to check if this is backwards-compatible though)

Also a parser that renders layout renderers (aka macros), isn't a solution that will be robust, as some layout renderers are dynamic.

PR #2257, opened 2017-08-19
${exception} -  only include separator when items are available 

### LINKED ISSUE #2250: Exception separator should only be inserted for available items
**Type**:
Bug

**NLog version**:
NLog.Web.AspNetCore: 4.4.1

**Platform**:
.NET Core 

**Current NLog config**:

```xml
      autoReload="true"
      internalLogLevel="Warn"
      internalLogFile="../Logs/internal-nlog.txt">

  <!-- define various log targets -->
  <targets>
    <!-- write logs to file -->
    <target xsi:type="File" name="allfile" fileName="../Logs/nlog-${shortdate}.log"
                layout="${longdate}|${event-properties:item=EventId.Id}|${logger}|${uppercase:${level}}|${message} ${exception:format=tostring,data:maxInnerExceptionLevel=10:separator=\r\n}" />

    <target xsi:type="ColoredConsole" name="console" />
  </targets>

  <rules>
    <logger name="*" minlevel="Debug" writeTo="console" />
    <logger name="*" minlevel="Info" writeTo="allfile" />
  </rules>
</nlog>

```

Created based on #2248 
The separator (newline in this case) is inserted unconditionally. So if no data is available I get an empty line in my log. The separator should only be inserted for fields that exist.

jojosardez:
The reason why it is adding the separator is because in the loop, we always add them first before we render the current function. Moreover, the separator is added even if the current render is empty.

The fix is to add a temporary string builder to be used by the render. Then after the render, check the string builder. If there is a content, that's the only time to append the separator plus the temporary string builder to the main string builder, if none then it will just proceed with the next render - without appending the string builder and adding unnecessary separators.

fixes #2250

### DIFF
--- a/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
@@ -197,10 +197,14 @@ namespace NLog.LayoutRenderers
 
                 foreach (ExceptionRenderingFormat renderingFormat in this.Formats)
                 {
-                    sb2.Append(separator);
-
+                    var sbCurrentRender = new StringBuilder();
                     var currentRenderFunction = _renderingfunctions[renderingFormat];
-                    currentRenderFunction(sb2, primaryException);
+                    currentRenderFunction(sbCurrentRender, primaryException);
+                    if (sbCurrentRender.Length > 0)
+                    {
+                        sb2.Append(separator);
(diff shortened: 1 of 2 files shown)


304NotModified:
Thanks again!

304NotModified:
this is a breaking change isn't? Do you think we need to add an option for it?

304NotModified:
think bug, so bugfix :)

Issue #2227, opened 2017-07-26
Removed target still actively logging


rbarillec:
**Type**: Bug

**Summary**:
   When removing a target from the `LoggingConfiguration` using `configuration.RemoveTarget("target name")`, the target still appears in `Configuration.AllTargets` and is still actively logging. My expectation was that the target would be completely removed from the configuration and would no longer appear in the list of targets nor be used by NLog in any way.

**NLog version**: 4.4.11

**Platform**: .Net 4.5

**Test illustrating the issue**:

```csharp
[Fact]
public void RemovingATarget()
{
    // Setup a configuration with 2 console targets
    var config = new LoggingConfiguration();

    config.AddTarget(new ConsoleTarget("ConsoleA") { Layout = "A | ${message}" });
    config.AddTarget(new ConsoleTarget("ConsoleB") { Layout = "B | ${message}" });

    config.AddRule(LogLevel.Debug, LogLevel.Fatal, "ConsoleA");
    config.AddRule(LogLevel.Debug, LogLevel.Fatal, "ConsoleB");

    LogManager.ThrowConfigExceptions = true;
    LogManager.ThrowExceptions = true;

    InternalLogger.LogLevel = LogLevel.Debug;
    InternalLogger.LogFile = "nlog_internal.log";

    // Apply the configuration
    LogManager.Configuration = config;
            
    // Success
    Assert.Equal(
        new[] { "ConsoleA", "ConsoleB" },
        LogManager.Configuration.ConfiguredNamedTargets.Select(target => target.Name)
    );

    // Remove the first target from the configuration
    LogManager.Configuration.RemoveTarget("ConsoleA");

    // Success
    Assert.Equal(
        new[] { "ConsoleB" },
        LogManager.Configuration.ConfiguredNamedTargets.Select(target => target.Name)
    );

    using (var writer = new StringWriter())
    {
        Console.SetOut(writer);
        var logger = LogManager.GetCurrentClassLogger();
        logger.Debug("Hello");

        var output = writer.ToString();

        Assert.Contains("B | Hello", output);           // Success
        Assert.DoesNotContain("A | Hello", output);     // FAILURE
    }

    // FAILURE: contains { "ConsoleA", "ConsoleB" }
    Assert.Equal(
        new[] { "ConsoleB" },
        LogManager.Configuration.AllTargets.Select(target => target.Name)
    );
}
```

In case of a BUG:

- What is the current result?
   The removed target is still logging to console (see test above).

- What is the expected result?
[...]

PR #2259, opened 2017-08-21
Bugfix for logging on a target even after removed #2227

jojosardez:
* Fixed the logging configuration to also remove the target from the configItems.
* Fixed the log factory to skip setting targets by level if the target doesn't exists (removed). 
* Added unit test (from the issue) to test the behavior.

### DIFF
--- a/src/NLog/Config/LoggingConfiguration.cs
+++ b/src/NLog/Config/LoggingConfiguration.cs
@@ -31,6 +31,7 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
+
 namespace NLog.Config
 {
     using JetBrains.Annotations;
@@ -340,12 +341,26 @@ namespace NLog.Config
         public void RemoveTarget(string name)
         {
             this.targets.Remove(name);
+            CloseAndRemoveTarget(name);
+            ValidateConfig();
+        }
+
+        /// <summary>
+        /// Removes a target from the logging rules and flush it.
+        /// </summary>
+        /// <param name="name">Name of the target.</param>
+        private void CloseAndRemoveTarget(string name)
+        {
             var loggingRules = LoggingRules.ToList();
             foreach (var rule in loggingRules)
             {
-                ((List<Target>)rule.Targets)?.RemoveAll(t => t.Name == name);
+                var targets = rule.Targets.Where(t => t.Name == name).ToList();
+                foreach (var target in targets)
+                {
+                    target.Close();
+                    rule.Targets.Remove(target);
+                }
             }
-            ValidateConfig();
         }
 
         /// <summary>
--- a/src/NLog/Config/LoggingConfiguration.cs
+++ b/src/NLog/Config/LoggingConfiguration.cs
@@ -340,13 +340,12 @@ namespace NLog.Config
         public void RemoveTarget(string name)
         {
             this.targets.Remove(name);
-
-            configItems.RemoveAll(c => c is Target && ((Target) c).Name == name);
             var loggingRules = LoggingRules.ToList();
             foreach (var rule in loggingRules)
             {
                 ((List<Target>)rule.Targets)?.RemoveAll(t => t.Name == name);
             }
+            ValidateConfig();
         }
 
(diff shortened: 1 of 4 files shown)


304NotModified:
Thanks! I'm currently without internet on my PC, will try to review next week

jojosardez:
Thanks @304NotModified! By the way, i'm not sure what's wrong with AppVeyor since the same test runs successfully in Travis.

snakefoot:
@jojosardez Would be better to remove the target from the LoggingRules, instead of checking Configuration.AllTargets. The check is also a little wrong incase a LogginRule has multiple targets, then it will `break;` when meeting an unknown target.

jojosardez:
Hi @snakefoot , thanks for the comments! I'll look into it this weekend.

jojosardez:
Hi @snakefoot , your suggestion works. Added the removal of the target from the logging rule and removed the condition and break from in the loop in log factory.

AppVeyor still fails though even if the build is successful in Travis. The problem is unrelated to the changes made. FYI @304NotModified .

snakefoot:
@jojosardez Looks like the LoggingRules should be used like this:

```
var loggingRules = this.LoggingRules.ToList();  // Make clone to minimize concurrency issues
foreach (var rule in loggingRules)
{
    // Blah blah blah
}
```

snakefoot:
@jojosardez And the configItems should be updated like this:

```
var roots = new List<object>();
roots.AddRange(configItems);
roots.AddRange(this.targets);
this.configItems = ObjectGraphScanner.FindReachableObjects<object>(roots.ToArray());
```

Maybe consider to "fix" the existing method ValidateConfig() so it can be re-used for updating `this.configItems`.

snakefoot:
@jojosardez It seems you cannot use RemoveAll() for removing the target.

snakefoot:
@jojosardez One might also consider to Flush/Close/Dispose the target when removing it.

snakefoot:
@jojosardez When removing a target, then one could also consider to iterate through all wrapper-targets, and check if one is pointing to this target, and then remove it.

But if the wrapper-target is able to reference multiple targets, then it will suddenly be very difficult to remove the target (CompoundTargetBase).

Maybe the best thing is just to close/dispose the target (so it is no longer is initialized), and remove it from loggingrules. Anything more will just generate too much headache. With the small catch that if the removed target is a wrapper-target then it should also close/dispose the inner-target (unless it is being used by other logging-rules).

jojosardez:
Hi @snakefoot , sorry I've been very busy this past week. Anyway, I have changed the implementation keeping in mind your suggestions. I ditched the remove all and just remove the targets individually. I also reused ValidateConfig() instead of removing the configItems explicitly. I also added closing of the target and a unit test to check if it really happened, before their removal.

snakefoot:
@jojosardez Multiple loggingRules can include the same target. Instead of closing the target just after having removed it from the loggingRule, then you should delay the closing of target until having removed it from all loggingRules.

Maybe the method should look like this:

```
public void RemoveTarget(string name)
{
    HashSet<Target> removedTargets = new HashSet<Target>();
    Target removedTarget;
    if (this.targets.TryGetValue(name, out removedTarget))
    {
        removedTargets.Add(removedTarget);
        this.targets.Remove(name);
    }

    var loggingRules = LoggingRules.ToList();
    foreach (var rule in loggingRules)
    {
        var targetList = rule.Targets.ToList();
        foreach (var target in targetList)
        {
            if (target.Name == name)
            {
                removedTargets.Add(removedTarget);
                this.targets.Remove(name);
            }
        }
    }

    foreach (var target in removedTargets)
    {
        target.Close(); 
    }
    
    ValidateConfig();
}
```

304NotModified:
> @jojosardez Multiple loggingRules can include the same target.

good point, rules and targets has more on more relationships. Could we make an unit test for this case?

304NotModified:
This need some changes, right?

304NotModified:
please open a new PR when you have time, thanks!

304NotModified:
merged by #2549

PR #2262, opened 2017-08-23
LogEventInfo.MessageTemplate - Subset of LogEventInfo.Properties

snakefoot:
Ordered by position (instead of hashcode).

It is possible to override these properties using the standard LogEventInfo-Properties-IDictionary-interface.

Allows capture of Microsoft FormattedLogValues without loosing the order given in the MessageTemplate. https://github.com/NLog/NLog.Extensions.Logging/pull/125

### DIFF
--- a/src/NLog/Internal/MessageTemplateParameters.cs
+++ b/src/NLog/Internal/MessageTemplateParameters.cs
@@ -1,0 +1,100 @@
+﻿// 
+// Copyright (c) 2004-2016 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without 
+// modification, are permitted provided that the following conditions 
+// are met:
+// 
+// * Redistributions of source code must retain the above copyright notice, 
+//   this list of conditions and the following disclaimer. 
+// 
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution. 
+// 
+// * Neither the name of Jaroslaw Kowalski nor the names of its 
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission. 
+// 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE 
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE 
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE 
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR 
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS 
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN 
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) 
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF 
+// THE POSSIBILITY OF SUCH DAMAGE.
+// 
+
+namespace NLog.Internal
+{
+    using System.Collections;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// Internal implementation of the interface for returning MessageTemplate parameters
+    /// </summary>
+    internal class MessageTemplateParameters : IMessageTemplateParameters
+    {
+        readonly IList<MessageTemplateParameter> _parameters;
+
+        /// <inheritDoc/>
+        public MessageTemplateParameter this[int index] { get { return _parameters[index]; } }
+
+        /// <inheritDoc/>
+        public int Count { get { return _parameters.Count; } }
+
+        /// <inheritDoc/>
+        public IEnumerator<MessageTemplateParameter> GetEnumerator() { return _parameters.GetEnumerator(); }
+
+        /// <inheritDoc/>
+        IEnumerator IEnumerable.GetEnumerator() { return _parameters.GetEnumerator(); }
+
+        /// <summary>
+        /// Constructore for positional parameters
+        /// </summary>
+        public MessageTemplateParameters(object[] parameters)
+        {
+            if (parameters != null && parameters.Length > 0)
+            {
+                _parameters = new MessageTemplateParameter[parameters.Length];
+                for (int i = 0; i < parameters.Length; ++i)
+                {
+                    string parameterName;
+                    switch (i)
+                    {
+                        case 0: parameterName = "0"; break;
+                        case 1: parameterName = "1"; break;
+                        case 2: parameterName = "2"; break;
+                        case 3: parameterName = "3"; break;
+                        case 4: parameterName = "4"; break;
+                        case 5: parameterName = "5"; break;
+                        case 6: parameterName = "6"; break;
+                        case 7: parameterName = "7"; break;
+                        case 8: parameterName = "8"; break;
+                        case 9: parameterName = "9"; break;
+                        default: parameterName = i.ToString(); break;
+                    }
+                    _parameters[i] = new MessageTemplateParameter(parameterName, parameters[i], null);
+                }
+            }
+            else
+            {
+                _parameters = Internal.ArrayHelper.Empty<MessageTemplateParameter>();
+            }
+        }
+
+        /// <summary>
+        /// Constructore for named parameters
+        /// </summary>
+        public MessageTemplateParameters(IList<MessageTemplateParameter> parameters)
+        {
+            _parameters = parameters ?? Internal.ArrayHelper.Empty<MessageTemplateParameter>();
+        }
+    }
+}
--- a/src/NLog/Internal/PropertiesDictionary.cs
+++ b/src/NLog/Internal/PropertiesDictionary.cs
@@ -1,0 +1,642 @@
+﻿// 
+// Copyright (c) 2004-2016 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without 
+// modification, are permitted provided that the following conditions 
+// are met:
+// 
+// * Redistributions of source code must retain the above copyright notice, 
+//   this list of conditions and the following disclaimer. 
+// 
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution. 
+// 
+// * Neither the name of Jaroslaw Kowalski nor the names of its 
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission. 
+// 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE 
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE 
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE 
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR 
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS 
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN 
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) 
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF 
+// THE POSSIBILITY OF SUCH DAMAGE.
+// 
+
+namespace NLog.Internal
+{
+    using System;
+    using System.Collections;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// Dictionary that combines the standard <see cref="LogEventInfo.Properties" /> with the
+    /// MessageTemplate-properties extracted from the <see cref="LogEventInfo.Message" />.
+    /// 
+    /// The <see cref="MessageProperties" /> are returned as the first items
+    /// in the collection, and in positional order.
+    /// </summary>
+    internal sealed class PropertiesDictionary : IDictionary<object, object>
+    {
+        struct PropertyValue
+        {
+            public readonly object Value;
+            public readonly bool MessageProperty;
+
+            public PropertyValue(object value, bool messageProperty)
+            {
+                Value = value;
+                MessageProperty = messageProperty;
+            }
+        }
+
+        private Dictionary<object, PropertyValue> _eventProperties;
+        private IList<MessageTemplateParameter> _messageProperties;
+        private DictionaryCollection _keyCollection;
+        private DictionaryCollection _valueCollection;
+        private IDictionary _eventContextAdapter;
+
+        /// <summary>
+        /// Injects the list of message-template-parameter into the IDictionary-interface
+        /// </summary>
+        /// <param name="parameterList">Message-template-parameters</param>
+        public PropertiesDictionary(IList<MessageTemplateParameter> parameterList = null)
+        {
+            MessageProperties = parameterList;
+        }
+
+        private bool IsEmpty { get { return (_eventProperties == null || _eventProperties.Count == 0) && (_messageProperties == null || _messageProperties.Count == 0); } }
+
+        public IDictionary EventContext { get { return _eventContextAdapter ?? (_eventContextAdapter = new DictionaryAdapter<object, object>(this)); } }
+
+        private Dictionary<object, PropertyValue> EventProperties
+        {
+            get
+            {
+                if (_eventProperties == null)
+                {
+                    if (_messageProperties != null && _messageProperties.Count > 0)
+                    {
+                        _eventProperties = new Dictionary<object, PropertyValue>(_messageProperties.Count);
+                        if (!InsertMessagePropertiesIntoEmptyDictionary(_messageProperties, _eventProperties))
+                        {
+                            _messageProperties = CreateUniqueMessagePropertiesListSlow(_messageProperties, _eventProperties);
+                        }
+                    }
+                    else
+                    {
+                        _eventProperties = new Dictionary<object, PropertyValue>();
+                    }
+                }
+                return _eventProperties;
+            }
+        }
+
+        public IList<MessageTemplateParameter> MessageProperties
+        {
+            get
+            {
+                return _messageProperties ?? ArrayHelper.Empty<MessageTemplateParameter>();
+            }
+            private set
+            {
+                if (value != null && value.Count > 0)
+                {
+                    _messageProperties = _eventProperties == null ? CreateUniqueMessagePropertiesListFast(value) : null;
+                    if (_messageProperties == null)
+                    {
+                        // Dictionary was already allocated, or the message-template-parameters are troublesome
+                        var eventProperties = _eventProperties ?? (_eventProperties = new Dictionary<object, PropertyValue>(value.Count));
+
+                        _messageProperties = new List<MessageTemplateParameter>(value.Count);
+                        for (int i = 0; i < value.Count; ++i)
+                            _messageProperties.Add(value[i]);
+
+                        if (eventProperties.Count != 0 || !InsertMessagePropertiesIntoEmptyDictionary(_messageProperties, eventProperties))
+                        {
+                            _messageProperties = CreateUniqueMessagePropertiesListSlow(_messageProperties, eventProperties);
+                        }
+                    }
+                }
+            }
+        }
+
+        /// <inheritDoc/>
+        public object this[object key]
+        {
+            get
+            {
+                if (!IsEmpty)
+                {
+                    PropertyValue valueItem;
+                    if (EventProperties.TryGetValue(key, out valueItem))
+                    {
+                        return valueItem.Value;
+                    }
+                }
+
+                throw new KeyNotFoundException();
+            }
+            set
+            {
+                EventProperties[key] = new PropertyValue(value, false);
+            }
+        }
+
+        /// <inheritDoc/>
+        public ICollection<object> Keys { get { return KeyCollection; } }
+        /// <inheritDoc/>
+        public ICollection<object> Values { get { return ValueCollection; } }
+
+        private DictionaryCollection KeyCollection
+        {
+            get
+            {
+                if (_keyCollection != null)
+                    return _keyCollection;
+                else if (IsEmpty)
+                    return EmptyKeyCollection;
+                else
+                    return _keyCollection ?? (_keyCollection = new DictionaryCollection(this, true));
+            }
+        }
+
+        private DictionaryCollection ValueCollection
+        {
+            get
+            {
+                if (_valueCollection != null)
+                    return _valueCollection;
+                else if (IsEmpty)
+                    return EmptyValueCollection;
+                else
+                    return _valueCollection ?? (_valueCollection = new DictionaryCollection(this, false));
+            }
+        }
+
+        private static readonly DictionaryCollection EmptyKeyCollection = new DictionaryCollection(new PropertiesDictionary(), true);
+        private static readonly DictionaryCollection EmptyValueCollection = new DictionaryCollection(new PropertiesDictionary(), false);
+
+        /// <inheritDoc/>
+        public int Count { get { return (_eventProperties?.Count) ?? (_messageProperties?.Count) ?? 0; } }
+
+        /// <inheritDoc/>
+        public bool IsReadOnly { get { return false; } }
+
+        /// <inheritDoc/>
+        public void Add(object key, object value)
+        {
+            EventProperties.Add(key, new PropertyValue(value, false));
+        }
+
+        /// <inheritDoc/>
+        public void Add(KeyValuePair<object, object> item)
+        {
+            Add(item.Key, item.Value);
+        }
+
+        /// <inheritDoc/>
+        public void Clear()
+        {
+            if (_eventProperties != null)
+                _eventProperties.Clear();
+            if (_messageProperties != null)
+                _messageProperties = ArrayHelper.Empty<MessageTemplateParameter>();
+        }
+
+        /// <inheritDoc/>
+        public bool Contains(KeyValuePair<object, object> item)
+        {
+            if (!IsEmpty)
+            {
+                if (((IDictionary<object, PropertyValue>)EventProperties).Contains(new KeyValuePair<object, PropertyValue>(item.Key, new PropertyValue(item.Value, false))))
+                    return true;
+
+                if (((IDictionary<object, PropertyValue>)EventProperties).Contains(new KeyValuePair<object, PropertyValue>(item.Key, new PropertyValue(item.Value, true))))
+                    return true;
+            }
+            return false;
+        }
+
+        /// <inheritDoc/>
+        public bool ContainsKey(object key)
+        {
+            if (!IsEmpty)
+            {
+                return EventProperties.ContainsKey(key);
+            }
+            return false;
+        }
+
+        /// <inheritDoc/>
+        public void CopyTo(KeyValuePair<object, object>[] array, int arrayIndex)
+        {
+            if (array == null)
+                throw new ArgumentNullException("array");
+            if (arrayIndex < 0)
+                throw new ArgumentOutOfRangeException("arrayIndex");
+
+            if (!IsEmpty)
+            {
+                foreach (var propertyItem in this)
+                {
+                    array[arrayIndex++] = propertyItem;
+                }
+            }
+        }
+
+        /// <inheritDoc/>
+        public IEnumerator<KeyValuePair<object, object>> GetEnumerator()
+        {
+            if (IsEmpty)
+                return System.Linq.Enumerable.Empty<KeyValuePair<object, object>>().GetEnumerator();
+            else
+                return new DictionaryEnumerator(this);
+        }
+
+        /// <inheritDoc/>
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            if (IsEmpty)
+                return ArrayHelper.Empty<KeyValuePair<object, object>>().GetEnumerator();
+            else
+                return new DictionaryEnumerator(this);
+        }
+
+        /// <inheritDoc/>
+        public bool Remove(object key)
+        {
+            if (!IsEmpty)
+            {
+                return EventProperties.Remove(key);
+            }
+            return false;
+        }
+
+        /// <inheritDoc/>
+        public bool Remove(KeyValuePair<object, object> item)
+        {
+            if (!IsEmpty)
+            {
+                if (((IDictionary<object, PropertyValue>)EventProperties).Remove(new KeyValuePair<object, PropertyValue>(item.Key, new PropertyValue(item.Value, false))))
+                    return true;
+
+                if (((IDictionary<object, PropertyValue>)EventProperties).Remove(new KeyValuePair<object, PropertyValue>(item.Key, new PropertyValue(item.Value, true))))
+                    return true;
+            }
+
+            return false;
+        }
+
+        /// <inheritDoc/>
+        public bool TryGetValue(object key, out object value)
+        {
+            if (!IsEmpty)
+            {
+                PropertyValue valueItem;
+                if (EventProperties.TryGetValue(key, out valueItem))
+                {
+                    value = valueItem.Value;
+                    return true;
+                }
+            }
+
+            value = null;
+            return false;
+        }
+
+        /// <summary>
+        /// Attempt to use the message-template-parameters without allocating a dictionary
+        /// </summary>
+        /// <param name="parameterList">Message-template-parameters</param>
+        /// <returns>List of message-template-parameters if succesful (else null)</returns>
+        static IList<MessageTemplateParameter> CreateUniqueMessagePropertiesListFast(IList<MessageTemplateParameter> parameterList)
+        {
+            if (parameterList.Count <= 10)
+            {
+                bool uniqueMessageProperties = true;
+                for (int i = 0; i < parameterList.Count - 1; ++i)
+                {
+                    for (int j = i + 1; j < parameterList.Count; ++j)
+                    {
+                        if (parameterList[i].Name == parameterList[j].Name)
+                        {
+                            uniqueMessageProperties = false;
+                            break;
+                        }
+                    }
+                }
+                if (uniqueMessageProperties)
+                {
+                    var messageProperties = new MessageTemplateParameter[parameterList.Count];
+                    for (int i = 0; i < parameterList.Count; ++i)
+                        messageProperties[i] = parameterList[i];
+                    return messageProperties;
+                }
+            }
+
+            return null;
+        }
+
+        /// <summary>
+        /// Attempt to insert the message-template-parameters into an empty dictionary
+        /// </summary>
+        /// <param name="messageProperties">Message-template-parameters</param>
+        /// <param name="eventProperties">The initially empty dictionary</param>
+        /// <returns>Message-template-parameters was inserted into dictionary without trouble (true/false)</returns>
+        private static bool InsertMessagePropertiesIntoEmptyDictionary(IList<MessageTemplateParameter> messageProperties, Dictionary<object, PropertyValue> eventProperties)
+        {
+            try
+            {
+                for (int i = 0; i < messageProperties.Count; ++i)
+                {
+                    eventProperties.Add(messageProperties[i].Name, new PropertyValue(messageProperties[i].Value, true));
+                }
+                return true; // We are done
+            }
+            catch (ArgumentException)
+            {
+                // Duplicate keys found, lets try again
+                for (int i = 0; i < messageProperties.Count; ++i)
+                {
+                    eventProperties.Remove(messageProperties[i].Name);
+                }
+                return false;
+            }
+        }
+
+        /// <summary>
+        /// Attempt to override the existing dictionary values using the message-template-parameters 
+        /// </summary>
+        /// <param name="messageProperties">Message-template-parameters</param>
+        /// <param name="eventProperties">The already filled dictionary</param>
+        /// <returns>List of unique message-template-parameters</returns>
+        private static IList<MessageTemplateParameter> CreateUniqueMessagePropertiesListSlow(IList<MessageTemplateParameter> messageProperties, Dictionary<object, PropertyValue> eventProperties)
+        {
+            List<MessageTemplateParameter> messagePropertiesUnique = null;
+            PropertyValue valueItem;
+            for (int i = 0; i < messageProperties.Count; ++i)
+            {
+                if (eventProperties.TryGetValue(messageProperties[i].Name, out valueItem))
+                {
+                    if (valueItem.MessageProperty)
+                    {
+                        if (messagePropertiesUnique == null)
+                        {
+                            messagePropertiesUnique = new List<MessageTemplateParameter>(messageProperties.Count);
+                            for (int j = 0; j < i; ++j)
+                            {
+                                messagePropertiesUnique.Add(messageProperties[j]);
+                            }
+                        }
+                        continue;   // Skip already exists
+                    }
+                }
+
+                eventProperties[messageProperties[i].Name] = new PropertyValue(messageProperties[i].Value, true);
+                if (messagePropertiesUnique != null)
+                {
+                    messagePropertiesUnique.Add(messageProperties[i]);
+                }
+            }
+
+            return messagePropertiesUnique ?? messageProperties;
+        }
+
+        class DictionaryEnumeratorBase
+        {
+            readonly PropertiesDictionary _dictionary;
+            int? _messagePropertiesEnumerator;
+            bool _eventEnumeratorCreated;
+            Dictionary<object, PropertyValue>.Enumerator _eventEnumerator;
+
+            protected DictionaryEnumeratorBase(PropertiesDictionary dictionary)
+            {
+                _dictionary = dictionary;
+            }
+
+            public KeyValuePair<object, object> CurrentPair
+            {
+                get
+                {
+                    if (_messagePropertiesEnumerator.HasValue)
+                    {
+                        var property = _dictionary._messageProperties[_messagePropertiesEnumerator.Value];
+                        return new KeyValuePair<object, object>(property.Name, property.Value);
+                    }
+                    else if (_eventEnumeratorCreated)
+                        return new KeyValuePair<object, object>(_eventEnumerator.Current.Key, _eventEnumerator.Current.Value.Value);
+                    else
+                        throw new InvalidOperationException();
+                }
+            }
+
+            public bool MoveNext()
+            {
+                if (_messagePropertiesEnumerator.HasValue)
+                {
+                    if ((_messagePropertiesEnumerator.Value + 1) < _dictionary._messageProperties.Count)
+                    {
+                        // Move forward to a key that is not overriden
+                        _messagePropertiesEnumerator = FindNextValidMessagePropertyIndex(_messagePropertiesEnumerator.Value + 1);
+                        if (_messagePropertiesEnumerator.HasValue)
+                            return true;
+
+                        _messagePropertiesEnumerator = _dictionary._eventProperties.Count - 1;
+                    }
+
+                    if (_dictionary._eventProperties != null && _dictionary._eventProperties.Count > 0)
+                    {
+                        _messagePropertiesEnumerator = null;
+                        _eventEnumerator = _dictionary._eventProperties.GetEnumerator();
+                        _eventEnumeratorCreated = true;
+                        return MoveNextValidEventProperty();
+                    }
+
+                    return false;
+                }
+                else
+                {
+                    if (_eventEnumeratorCreated)
+                    {
+                        return MoveNextValidEventProperty();
+                    }
+                    else
+                    {
+                        if (_dictionary._messageProperties != null && _dictionary._messageProperties.Count > 0)
+                        {
+                            // Move forward to a key that is not overriden
+                            _messagePropertiesEnumerator = FindNextValidMessagePropertyIndex(0);
+                            if (_messagePropertiesEnumerator.HasValue)
+                            {
+                                return true;
+                            }
+                        }
+
+                        if (_dictionary._eventProperties != null && _dictionary._eventProperties.Count > 0)
+                        {
+                            _eventEnumerator = _dictionary._eventProperties.GetEnumerator();
+                            _eventEnumeratorCreated = true;
+                            return MoveNextValidEventProperty();
+                        }
+
+                        return false;
+                    }
+                }
+            }
+
+            private bool MoveNextValidEventProperty()
+            {
+                while (_eventEnumerator.MoveNext())
+                {
+                    if (!_eventEnumerator.Current.Value.MessageProperty)
+                        return true;
+                }
+                return false;
+            }
+
+            private int? FindNextValidMessagePropertyIndex(int startIndex)
+            {
+                if (_dictionary._eventProperties == null)
+                    return startIndex;
+
+                PropertyValue valueItem;
+                for (int i = startIndex; i < _dictionary._messageProperties.Count; ++i)
+                {
+                    if (_dictionary._eventProperties.TryGetValue(_dictionary._messageProperties[i].Name, out valueItem) && valueItem.MessageProperty)
+                    {
+                        return i;
+                    }
+                }
+
+                return null;
+            }
+
+            public void Dispose()
+            {
+                // Nothing to do
+            }
+
+            public void Reset()
+            {
+                _messagePropertiesEnumerator = null;
+                _eventEnumeratorCreated = false;
+                _eventEnumerator = default(Dictionary<object, PropertyValue>.Enumerator);
+            }
+        }
+
+        class DictionaryEnumerator : DictionaryEnumeratorBase, IEnumerator<KeyValuePair<object, object>>
+        {
+            /// <inheritDoc/>
+            public KeyValuePair<object, object> Current { get { return CurrentPair; } }
+
+            /// <inheritDoc/>
+            object IEnumerator.Current { get { return CurrentPair; } }
+
+            public DictionaryEnumerator(PropertiesDictionary dictionary)
+                : base(dictionary)
+            {
+            }
+        }
+
+        class DictionaryCollection : ICollection<object>
+        {
+            readonly PropertiesDictionary _dictionary;
+            readonly bool _keyCollection;
+
+            public DictionaryCollection(PropertiesDictionary dictionary, bool keyCollection)
+            {
+                _dictionary = dictionary;
+                _keyCollection = keyCollection;
+            }
+
+            /// <inheritDoc/>
+            public int Count { get { return _dictionary.Count; } }
+
+            /// <inheritDoc/>
+            public bool IsReadOnly { get { return true; } }
+
+
+            /// <summary>Will always throw, as collection is readonly</summary>
+            public void Add(object item) { throw new NotSupportedException(); }
+
+            /// <summary>Will always throw, as collection is readonly</summary>
+            public void Clear() { throw new NotSupportedException(); }
+
+            /// <summary>Will always throw, as collection is readonly</summary>
+            public bool Remove(object item) { throw new NotSupportedException(); }
+
+            /// <inheritDoc/>
+            public bool Contains(object item)
+            {
+                if (_keyCollection)
+                {
+                    return _dictionary.ContainsKey(item);
+                }
+                else
+                {
+                    if (!_dictionary.IsEmpty)
+                    {
+                        if (_dictionary.EventProperties.ContainsValue(new PropertyValue(item, false)))
+                            return true;
+
+                        if (_dictionary.EventProperties.ContainsValue(new PropertyValue(item, true)))
+                            return true;
+                    }
+                    return false;
+                }
+            }
+
+            /// <inheritDoc/>
+            public void CopyTo(object[] array, int arrayIndex)
+            {
+                if (array == null)
+                    throw new ArgumentNullException("array");
+                if (arrayIndex < 0)
+                    throw new ArgumentOutOfRangeException("arrayIndex");
+
+                if (!_dictionary.IsEmpty)
+                {
+                    foreach (var propertyItem in _dictionary)
+                    {
+                        array[arrayIndex++] = _keyCollection ? propertyItem.Key : propertyItem.Value;
+                    }
+                }
+            }
+
+            /// <inheritDoc/>
+            public IEnumerator<object> GetEnumerator()
+            {
+                return new DictionaryCollectionEnumerator(_dictionary, _keyCollection);
+            }
+
+            /// <inheritDoc/>
+            IEnumerator IEnumerable.GetEnumerator()
+            {
+                return GetEnumerator();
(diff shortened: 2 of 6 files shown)


snakefoot:
@304NotModified Have fixed merge conflicts, so we can continue the journey.

src/NLog/Internal/PropertiesDictionary.cs
```diff
@@ -0,0 +1,698 @@
...
+                    _messageProperties = MakePropertiesUnique(value);
+                }
+                else
+                {
+                    _messageProperties = value ?? ArrayHelper.Empty<KeyValuePair<string, object>>();
+                }
+            }
+        }
+
+        private int MessagePropertiesIndexOfKey(object key)
```
304NotModified: proposal name: GetMessagePropertiesIndex

src/NLog/Internal/PropertiesDictionary.cs
```diff
@@ -0,0 +1,698 @@
...
+        /// <inheritDoc/>
+        public ICollection<object> Values { get { return (ICollection<object>)ValueCollection ?? new DictionaryCollection(this, false); } }
+
+        private Dictionary<object, object>.KeyCollection KeyCollection
+        {
+            get
+            {
+                if (_eventProperties != null && (_messageProperties == null || _messageProperties.Count == 0))
+                    return _eventProperties.Keys;
+                if (_eventProperties == null && (_messageProperties == null || _messageProperties.Count == 0))
```
304NotModified: I prefer an `else` and nested `if`s here. 

Maybe also a helper for `_messageProperties == null || _messageProperties.Count == 0` (9 times used?)

src/NLog/Internal/PropertiesDictionary.cs
```diff
@@ -0,0 +1,698 @@
...
+        {
+            if (_eventProperties != null && (_messageProperties == null || _messageProperties.Count == 0))
+                return ((IDictionary)_eventProperties).GetEnumerator();
+            return new DictionaryEntryEnumerator(this);
+        }
+
+        /// <inheritDoc/>
+        public bool Remove(object key)
+        {
+            if (_eventProperties != null && _eventProperties.Remove(key))
```
304NotModified: prefer `return (_eventProperties != null && _eventProperties.Remove(key))`

src/NLog/Internal/PropertiesDictionary.cs
```diff
@@ -0,0 +1,698 @@
...
+                        return _eventEnumerator.MoveNext();
+                    }
+
+                    return false;
+                }
+                else
+                {
+                    if (_eventEnumeratorCreated)
+                    {
+                        if (_eventEnumerator.MoveNext())
```
304NotModified: idem, `return _eventEnumerator.MoveNext()`

src/NLog/Internal/PropertiesDictionary.cs
```diff
@@ -0,0 +1,698 @@
...
+                    if (_dictionary._messageProperties != null && _dictionary._messageProperties.Count > 0)
+                    {
+                        EqualityComparer<object> comparer = EqualityComparer<object>.Default;
+                        for (int i = 0; i < _dictionary._messageProperties.Count; ++i)
+                        {
+                            if (comparer.Equals(_dictionary._messageProperties[i].Value, item))
+                                return true;
+                        }
+                    }
+                    if (_dictionary._eventProperties != null && _dictionary._eventProperties.ContainsValue(item))
```
304NotModified: idem, `return...`

src/NLog/Internal/PropertiesDictionary.cs
```diff
@@ -0,0 +1,698 @@
...
+
+        /// <inheritDoc/>
+        bool ICollection.IsSynchronized { get { return false; } }
+
+        /// <inheritDoc/>
+        public void Add(object key, object value)
+        {
+            int keyIndex = MessagePropertiesIndexOfKey(key);
+            if (keyIndex != -1)
+                throw new ArgumentException("duplicate", "key");
```
304NotModified: I think it would be nice to also include the duplicate key in the exception. We can do that in the exceptiondata or the message

src/NLog/Internal/PropertiesDictionary.cs
```diff
@@ -0,0 +1,698 @@
...
+                        {
+                            return i;
+                        }
+                    }
+                }
+            }
+            return -1;
+        }
+
+        private static IList<KeyValuePair<string, object>> MakePropertiesUnique(IList<KeyValuePair<string, object>> value)
```
304NotModified: please private methods below the publics

src/NLog/Internal/PropertiesDictionary.cs
```diff
@@ -0,0 +1,698 @@
...
+        /// <inheritDoc/>
+        object ICollection.SyncRoot { get { return null; } }
+
+        /// <inheritDoc/>
+        bool ICollection.IsSynchronized { get { return false; } }
+
+        /// <inheritDoc/>
+        public void Add(object key, object value)
+        {
+            int keyIndex = MessagePropertiesIndexOfKey(key);
```
304NotModified: so add is now `O(n)`?

src/NLog/Internal/PropertiesDictionary.cs
```diff
@@ -0,0 +1,698 @@
...
+            if (keyIndex != -1)
+                throw new ArgumentException("duplicate", "key");
+
+            EventProperties.Add(key, value);
+        }
+
+        /// <inheritDoc/>
+        public void Add(KeyValuePair<object, object> item)
+        {
+            int keyIndex = MessagePropertiesIndexOfKey(item.Key);
```
304NotModified: call the other `add` method?

snakefoot:
> Al lot of actions are now O(n)? - which is a bit unfortunate, what the use of the dictionary then ;)

As long the number for MessageTemplateProperties is below 100 parameters, then you will not notice O(n). Normal Properties will still work as O(1).

> Correct me if I didn't understand it correctly.

I want to have a property that works just like Microsofts FormattedLogValues, Serilogs MessageTemplate.Tokens / PropertyBinder.ConstructProperties or NLog StructuredEvents Template.Holes.

And using a standard array is much more optimal memory-wise than a dynamic container like a dictionary. Since we have to merge two containers (user-defined-properties + message-defined-properties). Then I suggest that the message-defined-properties uses the most optimal sub-container.

src/NLog/Internal/PropertiesDictionary.cs
snakefoot: Only when using structured-logging, and there will only be a performance hit if using more than 100 message-parameters.

src/NLog/Internal/PropertiesDictionary.cs
snakefoot: Fixed

src/NLog/Internal/PropertiesDictionary.cs
snakefoot: Fixed

src/NLog/Internal/PropertiesDictionary.cs
snakefoot: Fixed

src/NLog/Internal/PropertiesDictionary.cs
snakefoot: Fixed

src/NLog/Internal/PropertiesDictionary.cs
snakefoot: Fixed

src/NLog/Internal/PropertiesDictionary.cs
snakefoot: Fixed

snakefoot:
@304NotModified Alternative then we should just leave the Properties-dictionary alone, and just create a new merged-readonly-view of Properties and MessageTemplateProperties. Then just change all locations to use the merged-readonly-view by default instead of the Properties-dictionary.

snakefoot:
@304NotModified Changed the PropertiesDictionary so it always allocates and uses the dictionary. This gives extra allocation for structured-logging, but gives O(1) lookup speed for both property-types.

304NotModified:
Thanks! These kind of PRs are a bit hard to review, but will try to look at it soon.

snakefoot:
> Thanks! These kind of PRs are a bit hard to review, but will try to look at it soon.

It is a key-stone for the implementation of structured logging, so it mostly depends on how fast you want the structured logging up and running for NLog ver. 4.5.

304NotModified:
Maybe a stupid question (and we discussed it before long time ago i'm afraid),

Why do we need a separate collection for the messageProperties? Why isn't the current `logeventInfo.Properties` not enough? AFAIK we are missing the following features when using `logeventInfo.Properties`:

- the ordering of the messageProperties is lost
- the info of what is from the template and what not is lost.

I prefer adding all properties to the current properties collection and maybe add some more metadata if we need it for the ordering.

snakefoot:
> I prefer adding all properties to the current properties collection and maybe add some more metadata if we need it for the ordering.

That is what the PropertiesDictionary is doing.

304NotModified:
> That is what the PropertiesDictionary is doing.

true, but i'm not keen on the `logEventInfo.MessageTemplateProperties` option. 

> Why do we need a separate collection for the messageProperties? Why isn't the current logeventInfo.Properties not enough? AFAIK we are missing the following features when using (only) logeventInfo.Properties:
> 
> the ordering of the messageProperties is lost
> the info of what is from the template and what not is lost.

Is this sound and complete?

snakefoot:
> true, but i'm not keen on the logEventInfo.MessageTemplateProperties option.

You are welcome to suggest a different name, or a different approach for providing the messagetemplate-properties in the correct order.

The ordering is important, when logging to Json or a no-sql-database.

snakefoot:
@304NotModified We could also make the parsed MessageTemplate fully public (Like you did for NLog 4.5 Alpha) if that is more acceptable. Could properly make it so it is just a single array allocation of PropertyEntries-struct (Name + Value + Format)

304NotModified:
> @304NotModified Else we can replace MessageTemplateProperties with a public method called LogEventInfo.AddProperties(IList<KeyValuePair<string, object>> properties) ?

Yes, Something like that. I'm now creating a small (non-code) proposal



> I'm guessing 9 out of 10 doesn't use a combination of structured-logging-properties and custom-properties. And

I think also this is true

> @304NotModified We could also make the parsed MessageTemplate fully public (Like you did for NLog 4.5 Alpha) if that is more acceptable.

I like this, but I could understand this is a memory eater. Less keep this as plan B

> We can then consider to give a small life-boat, by extending the different property-layoutrenders to dive in and only output the property-subset found in PositionDictionary.MessageProperties

Could you elaborate on this? Which layout renderers should we change? I think `${event-properties}` and `${all-event-properties}`?

snakefoot:
> Could you elaborate on this? Which layout renderers should we change? I think ${event-properties} and ${all-event-properties}?

Yes, and also JsonLayout.IncludeAllProperties and Log4JXmlEventLayout.IncludeAllProperties

304NotModified:
I was thinking of something like this:

image

In code:

```c#
    public interface IProperties : IDictionary<object, object>
    {
        /// <summary>
        /// Add propery to the end of the collection 
        /// </summary>
        /// <param name="key"></param>
        /// <param name="value"></param>
        void AddOrderedPropery(string key, object value);

        /// <summary>
        /// Add properies to the end of the collection 
        /// </summary>
        /// <param name="properties"></param>
        void AddOrderedProperies(IEnumerable<KeyValuePair<string, object>> properties);

        /// <summary>
        /// Remove ordered property by key
        /// </summary>
        /// <param name="key"></param>
        /// <returns>found?</returns>
        bool RemoveOrderedProperty(string key);

        /// <summary>
        /// Get ordered properties in order of the template
        /// </summary>
        /// <returns></returns>
        IList<object> GetOrderedProperties();

        /// <summary>
        /// Get all properties which aren't in the template
        /// </summary>
        /// <returns></returns>
        IDictionary<object, object> GetNonOrderedProperties();

    }
```

```c#
  public class EventProperties : IProperties
    {
        /// <summary>
        /// Dict from key to index
        /// </summary>
        private Dictionary<string, int> _propertyIndexes;


        private Dictionary<object, object> _allProperties;

        public EventProperties()
        {
            _propertyIndexes = new Dictionary<string, int>(); //todo lazy init?
            _allProperties = new Dictionary<object, object>();//todo lazy init?
        }



        #region delegate _allProperties 

        /// <summary>Returns an enumerator that iterates through the collection.</summary>
        /// <returns>An enumerator that can be used to iterate through the collection.</returns>
        public IEnumerator<KeyValuePair<object, object>> GetEnumerator()
        {
            return _allProperties.GetEnumerator();
        }

        /// <summary>Returns an enumerator that iterates through a collection.</summary>
        /// <returns>An <see cref="T:System.Collections.IEnumerator" /> object that can be used to iterate through the collection.</returns>
        IEnumerator IEnumerable.GetEnumerator()
        {
            return ((IEnumerable)_allProperties).GetEnumerator();
        }

        /// <summary>Copies the elements of the <see cref="T:System.Collections.ICollection" /> to an <see cref="T:System.Array" />, starting at a particular <see cref="T:System.Array" /> index.</summary>
        /// <param name="array">The one-dimensional <see cref="T:System.Array" /> that is the destination of the elements copied from <see cref="T:System.Collections.ICollection" />. The <see cref="T:System.Array" /> must have zero-based indexing. </param>
        /// <param name="index">The zero-based index in <paramref name="array" /> at which copying begins. </param>
        /// <exception cref="T:System.ArgumentNullException">
        /// <paramref name="array" /> is null. </exception>
        /// <exception cref="T:System.ArgumentOutOfRangeException">
        /// <paramref name="index" /> is less than zero. </exception>
        /// <exception cref="T:System.ArgumentException">
        /// <paramref name="array" /> is multidimensional.-or- The number of elements in the source <see cref="T:System.Collections.ICollection" /> is greater than the available space from <paramref name="index" /> to the end of the destination <paramref name="array" />.-or-The type of the source <see cref="T:System.Collections.ICollection" /> cannot be cast automatically to the type of the destination <paramref name="array" />.</exception>
        public void CopyTo(Array array, int index)
        {
            ((ICollection)_allProperties).CopyTo(array, index);
        }

        /// <summary>Gets an object that can be used to synchronize access to the <see cref="T:System.Collections.ICollection" />.</summary>
        /// <returns>An object that can be used to synchronize access to the <see cref="T:System.Collections.ICollection" />.</returns>
        public object SyncRoot => ((ICollection)_allProperties).SyncRoot;

        /// <summary>Gets a value indicating whether access to the <see cref="T:System.Collections.ICollection" /> is synchronized (thread safe).</summary>
        /// <returns>true if access to the <see cref="T:System.Collections.ICollection" /> is synchronized (thread safe); otherwise, false.</returns>
        public bool IsSynchronized => ((ICollection)_allProperties).IsSynchronized;

        /// <summary>Determines whether the <see cref="T:System.Collections.IDictionary" /> object contains an element with the specified key.</summary>
        /// <returns>true if the <see cref="T:System.Collections.IDictionary" /> contains an element with the key; otherwise, false.</returns>
        /// <param name="key">The key to locate in the <see cref="T:System.Collections.IDictionary" /> object.</param>
        /// <exception cref="T:System.ArgumentNullException">
        /// <paramref name="key" /> is null. </exception>
        public bool Contains(object key)
        {
            return ((IDictionary)_allProperties).Contains(key);
        }

        /// <summary>Gets a value indicating whether the <see cref="T:System.Collections.IDictionary" /> object has a fixed size.</summary>
        /// <returns>true if the <see cref="T:System.Collections.IDictionary" /> object has a fixed size; otherwise, false.</returns>
        public bool IsFixedSize => ((IDictionary)_allProperties).IsFixedSize;

        /// <summary>Adds the specified key and value to the dictionary.</summary>
        /// <param name="key">The key of the element to add.</param>
        /// <param name="value">The value of the element to add. The value can be null for reference types.</param>
        /// <exception cref="T:System.ArgumentNullException">
        /// <paramref name="key" /> is null.</exception>
        /// <exception cref="T:System.ArgumentException">An element with the same key already exists in the <see cref="T:System.Collections.Generic.Dictionary`2" />.</exception>
        public void Add(object key, object value)
        {
            _allProperties.Add(key, value);
        }

        /// <summary>Adds an item to the <see cref="T:System.Collections.Generic.ICollection`1" />.</summary>
        /// <param name="item">The object to add to the <see cref="T:System.Collections.Generic.ICollection`1" />.</param>
        /// <exception cref="T:System.NotSupportedException">The <see cref="T:System.Collections.Generic.ICollection`1" /> is read-only.</exception>
        public void Add(KeyValuePair<object, object> item)
        {
            _allProperties.Add(item.Key, item.Value);
        }

        /// <summary>Removes all keys and values from the <see cref="T:System.Collections.Generic.Dictionary`2" />.</summary>
        public void Clear()
        {
            _allProperties.Clear();
        }

        /// <summary>Determines whether the <see cref="T:System.Collections.Generic.ICollection`1" /> contains a specific value.</summary>
        /// <returns>true if <paramref name="item" /> is found in the <see cref="T:System.Collections.Generic.ICollection`1" />; otherwise, false.</returns>
        /// <param name="item">The object to locate in the <see cref="T:System.Collections.Generic.ICollection`1" />.</param>
        public bool Contains(KeyValuePair<object, object> item)
        {
        }

        /// <summary>Copies the elements of the <see cref="T:System.Collections.Generic.ICollection`1" /> to an <see cref="T:System.Array" />, starting at a particular <see cref="T:System.Array" /> index.</summary>
        /// <param name="array">The one-dimensional <see cref="T:System.Array" /> that is the destination of the elements copied from <see cref="T:System.Collections.Generic.ICollection`1" />. The <see cref="T:System.Array" /> must have zero-based indexing.</param>
        /// <param name="arrayIndex">The zero-based index in <paramref name="array" /> at which copying begins.</param>
        /// <exception cref="T:System.ArgumentNullException">
        /// <paramref name="array" /> is null.</exception>
        /// <exception cref="T:System.ArgumentOutOfRangeException">
        /// <paramref name="arrayIndex" /> is less than 0.</exception>
        /// <exception cref="T:System.ArgumentException">The number of elements in the source <see cref="T:System.Collections.Generic.ICollection`1" /> is greater than the available space from <paramref name="arrayIndex" /> to the end of the destination <paramref name="array" />.</exception>
        public void CopyTo(KeyValuePair<object, object>[] array, int arrayIndex)
        {
        }

        /// <summary>Removes the first occurrence of a specific object from the <see cref="T:System.Collections.Generic.ICollection`1" />.</summary>
        /// <returns>true if <paramref name="item" /> was successfully removed from the <see cref="T:System.Collections.Generic.ICollection`1" />; otherwise, false. This method also returns false if <paramref name="item" /> is not found in the original <see cref="T:System.Collections.Generic.ICollection`1" />.</returns>
        /// <param name="item">The object to remove from the <see cref="T:System.Collections.Generic.ICollection`1" />.</param>
        /// <exception cref="T:System.NotSupportedException">The <see cref="T:System.Collections.Generic.ICollection`1" /> is read-only.</exception>
        public bool Remove(KeyValuePair<object, object> item)
        {
        }

        /// <summary>Determines whether the <see cref="T:System.Collections.Generic.Dictionary`2" /> contains the specified key.</summary>
        /// <returns>true if the <see cref="T:System.Collections.Generic.Dictionary`2" /> contains an element with the specified key; otherwise, false.</returns>
        /// <param name="key">The key to locate in the <see cref="T:System.Collections.Generic.Dictionary`2" />.</param>
        /// <exception cref="T:System.ArgumentNullException">
        /// <paramref name="key" /> is null.</exception>
        public bool ContainsKey(object key)
        {
            return _allProperties.ContainsKey(key);
        }

        /// <summary>Determines whether the <see cref="T:System.Collections.Generic.Dictionary`2" /> contains a specific value.</summary>
        /// <returns>true if the <see cref="T:System.Collections.Generic.Dictionary`2" /> contains an element with the specified value; otherwise, false.</returns>
        /// <param name="value">The value to locate in the <see cref="T:System.Collections.Generic.Dictionary`2" />. The value can be null for reference types.</param>
        public bool ContainsValue(object value)
        {
            return _allProperties.ContainsValue(value);
        }


        /// <summary>Implements the <see cref="T:System.Runtime.Serialization.ISerializable" /> interface and returns the data needed to serialize the <see cref="T:System.Collections.Generic.Dictionary`2" /> instance.</summary>
        /// <param name="info">A <see cref="T:System.Runtime.Serialization.SerializationInfo" /> object that contains the information required to serialize the <see cref="T:System.Collections.Generic.Dictionary`2" /> instance.</param>
        /// <param name="context">A <see cref="T:System.Runtime.Serialization.StreamingContext" /> structure that contains the source and destination of the serialized stream associated with the <see cref="T:System.Collections.Generic.Dictionary`2" /> instance.</param>
        /// <exception cref="T:System.ArgumentNullException">
        /// <paramref name="info" /> is null.</exception>
        public void GetObjectData(SerializationInfo info, StreamingContext context)
        {
            _allProperties.GetObjectData(info, context);
        }

        /// <summary>Implements the <see cref="T:System.Runtime.Serialization.ISerializable" /> interface and raises the deserialization event when the deserialization is complete.</summary>
        /// <param name="sender">The source of the deserialization event.</param>
        /// <exception cref="T:System.Runtime.Serialization.SerializationException">The <see cref="T:System.Runtime.Serialization.SerializationInfo" /> object associated with the current <see cref="T:System.Collections.Generic.Dictionary`2" /> instance is invalid.</exception>
        public void OnDeserialization(object sender)
        {
            _allProperties.OnDeserialization(sender);
        }

        /// <summary>Removes the value with the specified key from the <see cref="T:System.Collections.Generic.Dictionary`2" />.</summary>
        /// <returns>true if the element is successfully found and removed; otherwise, false.  This method returns false if <paramref name="key" /> is not found in the <see cref="T:System.Collections.Generic.Dictionary`2" />.</returns>
        /// <param name="key">The key of the element to remove.</param>
        /// <exception cref="T:System.ArgumentNullException">
        /// <paramref name="key" /> is null.</exception>
        public bool Remove(object key)
        {
            return _allProperties.Remove(key);
        }

        /// <summary>Gets the value associated with the specified key.</summary>
        /// <returns>true if the <see cref="T:System.Collections.Generic.Dictionary`2" /> contains an element with the specified key; otherwise, false.</returns>
        /// <param name="key">The key of the value to get.</param>
        /// <param name="value">When this method returns, contains the value associated with the specified key, if the key is found; otherwise, the default value for the type of the <paramref name="value" /> parameter. This parameter is passed uninitialized.</param>
        /// <exception cref="T:System.ArgumentNullException">
        /// <paramref name="key" /> is null.</exception>
        public bool TryGetValue(object key, out object value)
        {
            return _allProperties.TryGetValue(key, out value);
        }

        /// <summary>Gets the <see cref="T:System.Collections.Generic.IEqualityComparer`1" /> that is used to determine equality of keys for the dictionary. </summary>
        /// <returns>The <see cref="T:System.Collections.Generic.IEqualityComparer`1" /> generic interface implementation that is used to determine equality of keys for the current <see cref="T:System.Collections.Generic.Dictionary`2" /> and to provide hash values for the keys.</returns>
        public IEqualityComparer<object> Comparer => _allProperties.Comparer;

        /// <summary>Gets the number of key/value pairs contained in the <see cref="T:System.Collections.Generic.Dictionary`2" />.</summary>
        /// <returns>The number of key/value pairs contained in the <see cref="T:System.Collections.Generic.Dictionary`2" />.</returns>
        public int Count => _allProperties.Count;

        /// <summary>Gets a value indicating whether the <see cref="T:System.Collections.Generic.ICollection`1" /> is read-only.</summary>
        /// <returns>true if the <see cref="T:System.Collections.Generic.ICollection`1" /> is read-only; otherwise, false.</returns>
        public bool IsReadOnly { get; }


        /// <summary>Gets or sets the value associated with the specified key.</summary>
        /// <returns>The value associated with the specified key. If the specified key is not found, a get operation throws a <see cref="T:System.Collections.Generic.KeyNotFoundException" />, and a set operation creates a new element with the specified key.</returns>
        /// <param name="key">The key of the value to get or set.</param>
        /// <exception cref="T:System.ArgumentNullException">
        /// <paramref name="key" /> is null.</exception>
        /// <exception cref="T:System.Collections.Generic.KeyNotFoundException">The property is retrieved and <paramref name="key" /> does not exist in the collection.</exception>
        public object this[object key]
        {
            get => _allProperties[key];
            set => _allProperties[key] = value;
        }

        /// <summary>Gets an <see cref="T:System.Collections.Generic.ICollection`1" /> containing the keys of the <see cref="T:System.Collections.Generic.IDictionary`2" />.</summary>
        /// <returns>An <see cref="T:System.Collections.Generic.ICollection`1" /> containing the keys of the object that implements <see cref="T:System.Collections.Generic.IDictionary`2" />.</returns>
        public ICollection<object> Keys { get; }

        /// <summary>Gets an <see cref="T:System.Collections.Generic.ICollection`1" /> containing the values in the <see cref="T:System.Collections.Generic.IDictionary`2" />.</summary>
        /// <returns>An <see cref="T:System.Collections.Generic.ICollection`1" /> containing the values in the object that implements <see cref="T:System.Collections.Generic.IDictionary`2" />.</returns>
        public ICollection<object> Values { get; }

        #endregion

        #region Implementation of IProperties

        /// <summary>
        /// Add propery to the end of the collection 
        /// </summary>
        /// <param name="key"></param>
        /// <param name="value"></param>
        public void AddOrderedPropery(string key, object value)
        {
            _allProperties[key] = value;
            _propertyIndexes[key] = _propertyIndexes.Count;
        }

        /// <summary>
        /// Add properies to the end of the collection 
        /// </summary>
        /// <param name="properties"></param>
        public void AddOrderedProperies(IEnumerable<KeyValuePair<string, object>> properties)
        {
            foreach (var property in properties)
            {
                AddOrderedPropery(property.Key, property.Value);
            }
        }

        /// <summary>
        /// Remove ordered property by key
        /// </summary>
        /// <param name="key"></param>
        /// <returns>found?</returns>
        public bool RemoveOrderedProperty(string key)
        {
            var removed = _allProperties.Remove(key);

            if (removed)
            {
                _propertyIndexes.Remove(key);
            }
            return removed;
        }

        /// <summary>
        /// Get ordered properties in order of the template
        /// </summary>
        /// <returns></returns>
        public IList<object> GetOrderedProperties()
        {
            var arr = new object[_propertyIndexes.Count];
            foreach (var prop in _propertyIndexes)
            {
                arr[prop.Value] = _allProperties[prop.Value];
            }

            return arr;
        }

        /// <summary>
        /// Get all properties which aren't in the template
        /// </summary>
        /// <returns></returns>
        public IDictionary<object, object> GetNonOrderedProperties()
        {
            var dict = new Dictionary<object,object>();

            foreach (var prop in _allProperties)         {
                if (!_propertyIndexes.ContainsKey(prop.Value)) //todo handle string?
                {
                    dict.Add(prop.Key, prop.Value);
                }
            }
            return dict;
        }

        #endregion
    }
```

notes:

- I renamed it to orderedProperty, but maybe templateProperty is better. 
- I've generated a lot of the delegating code with Resharper (haven't checked it yet)
- I'm not sure sure if the orderedProperties names are always strings, anyway this gives issues in `GetNonOrderedProperties` (see code)
- if I'm missing something, please let me know
- feel free to edit / counter propose. Your expertise on memory usage is really appreciated! (i'm more a code complexity guy)

snakefoot:
Will this new IEventProperties-interface have a new LogEventInfo-property, and the old LogEventInfo-property will just point to this one?

snakefoot:
>The idea is that LogEventInfo.Property type will be changed from IDictionary<object, object> to IProperties. As this property is getter only, it's backwards comp.

Was think it would cause issues, but if not then fine with me.

I have now implemented the new interface and removed LogEventInfo.MessageTemplateProperties

tests/NLog.UnitTests/LogMessageFormatterTests.cs
```diff
@@ -34,21 +34,22 @@
...
         [Fact]
         public void StructuredLoggingTest()
         {
             LogEventInfo logEventInfo = new LogEventInfo(LogLevel.Info, string.Empty, "Login request from {Username} for {Application}");
-            logEventInfo.Properties["Username"] = "John";
-            logEventInfo.Properties["Application"] = "BestApplicationEver";
-            logEventInfo.Parameters = new object[] { "Login request from John for BestApplicationEver" };   // Must have parameter to activate MessageFormatter
+            foreach (var item in new KeyValuePair<string, object>[] { new KeyValuePair<string, object>("Username", "John"), new KeyValuePair<string, object>("Application", "BestApplicationEver") })
+                logEventInfo.Properties.AddOrderedProperty(item.Key, item.Value);
+            logEventInfo.Parameters = new object[] { "John", "BestApplicationEver", "Login request from John for BestApplicationEver" };
```
304NotModified: Why do we need this beside the `logEventInfo.Properties`? I think this unit test is doing to much, as I'm confused. Are we testing multiple things here?

src/NLog/IEventProperties.cs
```diff
@@ -0,0 +1,63 @@
...
+        /// </summary>
+        /// <param name="key"></param>
+        /// <returns>found?</returns>
+        bool RemoveOrderedProperty(string key);
+
+        /// <summary>
+        /// Get ordered properties in order of the template
+        /// </summary>
+        /// <returns></returns>
+        IEnumerable<KeyValuePair<string, object>> OrderedProperties { get; }
```
304NotModified: I'm doubting of ` IList<KeyValuePair<string, object>>` instead of ` IEnumerable<KeyValuePair<string, object>>`.  What do you think is best? (note that it's a getter only)

snakefoot:
> The idea is that LogEventInfo.Property type will be changed from IDictionary<object, object> to IProperties. As this property is getter only, it's backwards comp.

Think it is only compatible when doing a recompile, when just replacing an old NLog-assembly with the new that have changed the return-type of the Properties-property, then one get this exception:

```
Unhandled Exception: System.MissingMethodException: Method not found: 'System.Collections.Generic.IDictionary`2<System.Object,System.Object> NLog.LogEventInfo.get_Properties()'.
   at TestApplication.Program.Main(String[] args)
```

This is my advanced test application:
```
        {
            var hello = new NLog.LogEventInfo();
            hello.Properties["Hello"] = "Goodbye";
            Console.ReadKey();
        }
```

304NotModified:
> Think it is only compatible when doing a recompile, when just replacing an old NLog-assembly with the new that have changed the return-type of the Properties-property, then one get this exception:

Oops, binary breaking change I think. See https://github.com/dotnet/corefx/blob/c02d33b18398199f6acc17d375dab154e9a1df66/Documentation/coding-guidelines/breaking-change-rules.md#signatures

Not sure if we could fix this by *also* adding `IDictionary<object, object> LogEventInfo.Properties` and implicit interface implementation (we need also a new interface I think)

tests/NLog.UnitTests/LogMessageFormatterTests.cs
snakefoot: For normal structured logging you get a raw-message and arguments-params. Then based on that you derive the FormattedMessage and fill Properties-dictionary. In the Microsoft Extension Logging world you get a pre-formatted-message and can dig out some message-properties (where the last property is actually the original raw-message). For the Microsoft Extension Logging, then I want to capture the raw-message, capture the microsoft-formatted-message (store it as arguments-params), and fill the Properties-dictionary.

src/NLog/IEventProperties.cs
snakefoot: Well the IList-interface have an Add-method. Not sure I want to support that.

snakefoot:
> Oops, binary breaking change I think. 

Just like I was expecting. Guess we need to keep the LogEventInfo.Properties unchanged. Now we just have to figure out how this OrderedProperties should work.

- Should it only return the properties marked as ordered ?
- Should it just be an IList or be a custom interface that by default return all properties, but has a property to return the subset of ordered-properties?

304NotModified:
> Just like I was expecting. Guess we need to keep the LogEventInfo.Properties unchanged.

Or considering moving to NLog 5. (well replacing NLog 5's current state with this)

snakefoot:
> Or considering moving to NLog 5.

The only reason for getting this OrderedProperties or MessageTemplateProperties or NamedParameters is to have an interface to correctly capture the structured logging details from other frameworks like Microsoft Extension Logging. And maybe also use the interface to correctly forward the structured logging details to another logging framework (can see that logging frameworks often have adapters to each other).

Still confused why we can have LogEventInfo.Parameters and LogEventInfo.Properties, but we cannot have LogEventInfo.NamedParameters.

304NotModified:
> Still confused why we can have LogEventInfo.Parameters and LogEventInfo.Properties, but we cannot have LogEventInfo.NamedParameters (optimized to point to the same PropertiesDictionary).

Is it a NamedParameters? Or NamedButAlsoOrderedParameters? The latter is really a bit confusing. 

> Maybe time for Plan B - MessageTemplate-object. Then Microsoft Extension Logging will just have to pay the penalty of allocating the MessageTemplate-object to correctly forward the structured-logging-details.

Is the MessageTemplate as class or struct? Is the allocating penalty high?

snakefoot:
> Is the MessageTemplate as class or struct? Is the allocating penalty high?

Will probably be the same as allocating a List-object with an internal array of hole-structs (Name + Value + Format + possible message-string-indexof). Will probably be an interface, that we will implemented by the PropertiesDictionary.

I'm guessing you would have a better idea of this interface, since you created NLog.StructuredEvents that I have based the logic upon (So think the Template-object but without the Literals-array)

snakefoot:
@304NotModified Now changed the PropertiesDictionary to also implement IList (that only returns the positional properties, and added LogEventInfo.NamedParameters.

304NotModified:
So a namedparameter is a kind of property? I have difficulties to explain that to someone. 

Counter proposal: drop namedparameter property and add a method: GetPropertiesDefinedInTemplate (work name) that returns the IList, or maybe event better a new IOrdererDictionary (custom made interface). Maybe we need also an "add" method then.

snakefoot:
> So a namedparameter is a kind of property? I have difficulties to explain that to someone.

As easy to explain as MessageTemplate is also kind of a property-collection :). Actually just noticed that Serilog doesn't ensure the Properties-collection is sorted (though it usually is), but Serilog LogEvent.Properties automatically includes the properties in MessageTemplate.

304NotModified:
> As easy to explain as MessageTemplate is also kind of a property-sub-collection :).

true, but it not visible now from the code. 2 properties that are changing the same stuff doesn't look right.

snakefoot:
> true, but it not visible now from the code. 2 properties that are changing the same stuff doesn't look right.

But you want this. Since you want the Properties to include the MessageTemplate-properties.

304NotModified:
All I want is 2 kind properties on the logeventinfo (named and unnamed), not 3.

snakefoot:
> All I want is 2 kind properties on the logeventinfo (named and unnamed), not 3.

Lets make the MessageTemplate then. Then Seq will also be very happy.

304NotModified:
but that one was slow?

snakefoot:
> but that one was slow?

Can make something that is fast. Will work just like now then instead of IList then it will be IMessageTemplate.

304NotModified:
So

Adding `LogEventInfo.MessageTemplate` with type `IMessageTemplate`, correct?

Will `IMessageTemplate` have the full template or is it the `IList` of KeyValue pairs?

snakefoot:
> Will IMessageTemplate have the full template or is it the IList of KeyValue pairs?

Something like this:

```
    struct NamedParameter
    {
        public readonly string Name;
        public readonly string Format;
        public readonly object Value;
    }

    interface IMessageTemplate
    {
        string Message { get; }
        bool Positional { get; }
        IList<NamedParameter> NamedParameters { get; }
    }
```

304NotModified:
What about (only names changed):

```c#
    struct TemplateParameter
    {
        public readonly string Name;
        public readonly string Format;
        public readonly object Value;
    }

    interface IMessageTemplate
    {
        string Template { get; }
        bool Positional { get; }
        IList<TemplateParameter> Parameters{ get; }
    }
```

snakefoot:
> What about (only names changed):

Works for me. Should we keep IList or make it IEnumerable ? Can make it implicit readonly by throwing not supported on Add, Remove, etc.

304NotModified:
> Works for me. Should we keep IList or make it IEnumerable ?

Good one. The `IEnumerable ` is really small and throwing runtime exceptions is bad. What about creating a `NLog.IReadonlyList` (which inherits from `IEnumerable `)?

snakefoot:
Else we can make it like this:

```
interface IMessageTemplateParameterList : IEnumerable<MessageTemplateParameter>
{
        MessageTemplateParameter this[int index] { get; }
        int Count { get; }
}

```
And just have this property on LogEventInfo:

`IMessageTemplateParameterList  MessageTemplateParameterList { get; set; }`

304NotModified:
I was thinking now:

```c#
class LogEventInfo 
{
   ...
  IMessageTemplate MessageTemplate {get;}
}

interface IMessageTemplateParameterList : IEnumerable<MessageTemplateParameter>
{
        MessageTemplateParameter this[int index] { get; }
        int Count { get; }
}


    struct TemplateParameter
    {
        public readonly string Name;
        public readonly string Format;
        public readonly object Value;
    }

    interface IMessageTemplate
    {
        string Template { get; }
        bool Positional { get; }
        IMessageTemplateParameterList Parameters{ get; }
    }

```

snakefoot:
I have made some small hacks that might need extra scrutiny:

- LogEventInfo implements IMessageTemplate (as it already has the Message-property).
- MessageTemplate is an example of how to implement the IMessageTemplate. It inherits directly from List to minimize allocation. Right now only used for testing, but maybe others will find it useful (Like NLog Extension Logging).

304NotModified:
> LogEventInfo implements IMessageTemplate 

Is this needed (for performance?), its a bit strange. I would remove it if not needed. Also, is the message of the logevent the same as the template's message? Isn't the first one rendered?

snakefoot:
>> LogEventInfo implements IMessageTemplate
>
> Is this needed (for performance?), its a bit strange. I would remove it if not needed. 

After a good nights sleep, then I have removed this oddity, and changed it to an allocation of MessageTemplateReadOnly

>  Also, is the message of the logevent the same as the template's message? Isn't the first one rendered?

LogEventInfo.FormattedMessage is one already rendered. LogEventInfo.Message is the raw message. MessageTemplate.Message is the raw messsage.

src/NLog/IMessageTemplate.cs
```diff
@@ -0,0 +1,144 @@
...
+        /// </summary>
+        public string Message { get; }
+        /// <summary>
+        /// 
+        /// </summary>
+        public bool IsPositional { get; }
+        /// <summary>
+        /// 
+        /// </summary>
+        public IMessageTemplateParameterList Parameters { get { return this; } }
```
304NotModified: Is this for performance reasons? It's a bit strange design to have a property that is the same as `this` IMO

src/NLog/Internal/PropertiesDictionary.cs
```diff
@@ -0,0 +1,744 @@
...
+    {
+        struct PropertyValue
+        {
+            public readonly object Value;
+            public readonly bool MessageProperty;
+
+            public PropertyValue(object value, bool messageProperty)
+            {
+                Value = value;
+                MessageProperty = messageProperty;
```
304NotModified: What's messageproperty? The name or the value, or something else?

src/NLog/Internal/PropertiesDictionary.cs
```diff
@@ -0,0 +1,744 @@
...
+        internal void SetMessageTemplateParameterList(IMessageTemplateParameterList value)
+        {
+            ResetMessageProperties();
+
+            if (value != null && value.Count > 0)
+            {
+                if (_eventProperties == null && value.Count <= 10)
+                {
+                    bool uniqueMessageProperties = true;
+                    for (int i = 0; i < value.Count - 1; ++i)
```
304NotModified: What's uniqueMessageProperties  here? What are we trying to do here?

src/NLog/Internal/PropertiesDictionary.cs
```diff
@@ -0,0 +1,744 @@
...
+        void IDictionary.Remove(object key)
+        {
+            Remove(key);
+        }
+
+        /// <inheritDoc/>
+        void ICollection.CopyTo(Array array, int arrayIndex)
+        {
+            if (array == null)
+                throw new ArgumentNullException("array");
```
304NotModified: please use `nameof` :)

src/NLog/Internal/PropertiesDictionary.cs
```diff
@@ -0,0 +1,744 @@
...
+                {
+                    array.SetValue(new DictionaryEntry(propertyItem.Key, propertyItem.Value), arrayIndex);
+                    ++arrayIndex;
+                }
+            }
+        }
+
+        void ResetMessageProperties()
+        {
+            if (_eventProperties != null && _messageProperties != null && _messageProperties.Count > 0)
```
304NotModified: is this `if(IsEmpty)`?

src/NLog/Internal/PropertiesDictionary.cs
```diff
@@ -0,0 +1,744 @@
...
+                        {
+                            _eventProperties.Remove(key);
+                        }
+                    }
+                }
+                _messageProperties = ArrayHelper.Empty<MessageTemplateParameter>();
+            }
+        }
+
+        private static bool InsertMessagePropertiesDictionary(IList<MessageTemplateParameter> messageProperties, Dictionary<object, PropertyValue> eventProperties)
```
304NotModified: for the docs phase: please document bool return values

src/NLog/LogEventInfo.cs
```diff
@@ -281,20 +280,75 @@ public string FormattedMessage
...
+            get
+            {
+                bool positional = true;
+                IList<MessageTemplateParameter> messageTemplateParameters = ArrayHelper.Empty<MessageTemplateParameter>();
+                if (this.properties != null && this.properties.MessageProperties != null && this.properties.MessageProperties.Count > 0)
+                {
+                    messageTemplateParameters = this.properties.MessageProperties;
+                    if (this.properties.MessageProperties[0].Name != null)
+                    {
+                        if (this.properties.MessageProperties[0].Name.Length != 1 || this.properties.MessageProperties[0].Name[0] < '0' || this.properties.MessageProperties[0].Name[0] > '9')
```
304NotModified: why is this needed in the getter? For performance reasons?

tests/NLog.UnitTests/LogMessageFormatterTests.cs
```diff
@@ -34,21 +34,26 @@
...
-            logEventInfo.Properties["Username"] = "John";
-            logEventInfo.Properties["Application"] = "BestApplicationEver";
-            logEventInfo.Parameters = new object[] { "Login request from John for BestApplicationEver" };   // Must have parameter to activate MessageFormatter
+            LogEventInfo logEventInfo = new LogEventInfo(LogLevel.Info, string.Empty, string.Empty);
+            MessageTemplate messageTemplate = new MessageTemplate("Login request from {Username} for {Application}", false, new[]
+            {
+                new MessageTemplateParameter("Username", "John", null),
+                new MessageTemplateParameter("Application", "BestApplicationEver", null)
+            });
+            logEventInfo.Parameters = new object[] { "John", "BestApplicationEver", "Login request from John for BestApplicationEver" };
```
304NotModified: please split the unit test into 2 tests, or don't use the same values (john, BestApplicationEver).

304NotModified:
Also, do we still need the propertiesdictionary (and interface)? It was needed with the 2 properties on logeventinfo, but now we have 3 parameters.

snakefoot:
> Also, do we still need the propertiesdictionary (and interface)?

PropertiesDictionary ensures LogEventInfo.Properties includes the parameters from LogEventInfo.MessageTemplate.

tests/NLog.UnitTests/LogMessageFormatterTests.cs
snakefoot: The unit test displays how I'm planning capture the structured parameters from Microsoft Extension Logging.

src/NLog/LogEventInfo.cs
snakefoot: Will remove it, since it will never happen

src/NLog/Internal/PropertiesDictionary.cs
snakefoot: Please wait with doc-stuff until we have something real

src/NLog/Internal/PropertiesDictionary.cs
snakefoot: Will change it to IsEmpty (with some extra)

src/NLog/Internal/PropertiesDictionary.cs
snakefoot: Still not used to having the C# 6.0 compiler after changing to VS2017

src/NLog/Internal/PropertiesDictionary.cs
snakefoot: Ensure the MessageTemplate-parameters can be used as Dictionary-values. Ex. "Male user: {userid} loves female user: {userid}".

src/NLog/Internal/PropertiesDictionary.cs
snakefoot: Means the dictionary-entry comes from the _messageProperty-list.

src/NLog/IMessageTemplate.cs
snakefoot: It is to minimize allocations and because we are not using IReadOnlyList

src/NLog/LogEventInfo.cs
snakefoot: Fixed

src/NLog/Internal/PropertiesDictionary.cs
snakefoot: Fixed

304NotModified:
> Have been jumping through many hoops so far, without anything being decided yet.

Sorry, it's hard to discuss things, slow-chatting and while the code evolves all the time.  I prefer also a decision before building. 

> Reason why I have now reverted to introducing LogEventInfo.MessageTemplate was that you selected this solution yourself in NLog 4.5 Alpha. 

I think that's the best design when the perf is good enough. 

> Was hoping we could continue on and stop the interface-fight, and focus on implementation.

In my opinion that's just i'm trying to do currently. 

I think we not on the same line now. Sorry about that. What do we think we still have to decide?

snakefoot:
> I think we not on the same line now. Sorry about that. What do we think we still have to decide?

Well comments like "we still need the propertiesdictionary?", usually make me think that we are still stuck in the decision phase :)

> LogEventInfo.MessageTemplate - I think that's the best design when the perf is good enough.

Well I think the performance of this PR is good. Maybe considering to provide internal access to the actual LogEventInfo.PropertiesDictionary and change its enumerator to struct. Then the enumeration would be allocation free for the internal NLog-logic. But that optimization is for a different day.

I cheated in #2208 where I dropped the MessageTemplate completely, because of its overhead. Now the current interface allows me to have it without a serious performance hit.

304NotModified:
ps updated my reaction ;)

304NotModified:
> Also like simple code, but I don't like open questions about design without any counter-suggestion, then one is left in the dark trying to find the light-switch.

Fair enough. But keep in mind it's hard to give good suggestions if code is submitted which isn't clear from itself. Don't get me wrong, but it takes a lot of time to understand some parts and i'm currently have to limited time for that.

Maybe it's a better idea, that would work better for me, to create a PR with only non-optimized code (just a clear API, small changes), and in another PR a (memory optimized) change.   Design is now mixed with (highy)optimized changes.

snakefoot:
> Maybe it's a better idea, that would work better for me, to create a PR with only non-optimized code 

The logic that ensures LogEventInfo.Properties is a combined dictionary of both, and allows replacing of the MessageTemplate.Parameters is quite complicated. Not sure I can make it less complicated.

If you just want to see the new public API, then you just have to look at LogEventInfo.MessageTemplate. Everything else is the same.

You are welcome to present your own simple implementation of the LogEventInfo.MessageTemplate and its interaction with LogEventInfo.Properties.

304NotModified:
hi @snakefoot 

I thought about this and checked the code more in-depth.

In my opinion it's a bit strange that the LogEventInfo.MessageTemplate has a setter. I think this is a result of earlier dissensions (probably mine ;) )
Also I really like the idea behind the PropertyDictionary but as we can't use it well without breaking changes (no `PropertiesDictionary Properties` in LogEventInfo, I;m doubting if we really need it in a new impl.

I have a different, out-of-the-box, proposal, please let me know what you think:

Only add to the LogEventInfo:

     public IPropertiesMetaInfo PropertiesMetaInfo { get; }

Which holds the extra info which is now in the PropertiesDictionary 

The interface:

```c#
    public interface IPropertiesMetaInfo : IDictionary<string, IPropertyMetaInfo>
    {
        
    }

    public interface IPropertyMetaInfo
    {
        int Position {get;}

        string Format { get; }
    }

```

Downside is that it's a bit more work for the client to enumerate all properties in the template, but we could add a helper (dunno where).

If you think the current design is better, I really like to know!

snakefoot:
I don't understand what you are trying to solve with your proposal besides making Nlog more difficult to use (dictionary interface is not simple).

The reason why I have a MessageTemplate setter is because of Microsoft Extensions Logging. We can remove the setter by creating a new LogEventInfo constructor. Then we don't have the need to support only removing the messagetemplate properties. This will simplify the logic, but not get rid of PropertiesDictionary as I want the ordered guarantee



Sent from my Samsung device


snakefoot:
@304NotModified I have now removed the ability to modify the MessageTemplateParameters after they have been initialized, by using a LogEventInfo-constructor. And now it is only possible to access the MessageTemplateParameters if using NET4_5 or NETSTANDARD (Because of IReadOnlyList)

304NotModified:
> I don't understand what you are trying to solve with your proposal besides making Nlog more difficult to use (dictionary interface is not simple).

As for the API design, it's then not clear that the templateParameters is a subset of properties.

The perfect design as we didn't need any changes to the logeventinfo, and use the logeventinfo properties. This was not the case for the position number and so I have created a new property, the metainfo (in earlier design I added it the the PropertiesDictionary, but that wasn't backwardscomp too bad).

snakefoot:
> As for the API design, it's then not clear that the templateParameters is a subset of properties.

Please review my new proposal. I think it satisfies your requirement of not having several ways to modify LogEventInfo.Properties.

src/NLog/LogEventInfo.cs
```diff
@@ -281,16 +296,47 @@ public string FormattedMessage
...
+                            case 3: parameterName = "3"; break;
+                            case 4: parameterName = "4"; break;
+                            case 5: parameterName = "5"; break;
+                            case 6: parameterName = "6"; break;
+                            case 7: parameterName = "7"; break;
+                            case 8: parameterName = "8"; break;
+                            case 9: parameterName = "9"; break;
+                            default: parameterName = i.ToString(); break;
+                        }
+                        messageParameters[i] = new MessageTemplateParameter(parameterName, parameters[i], null);
```
304NotModified: would not it been better to use `yield return` here? (depends on the return interface)

src/NLog/Internal/PropertiesDictionary.cs
304NotModified: Please rename to IsMessageProperty

tests/NLog.UnitTests/LogMessageFormatterTests.cs
304NotModified: This test is a bit confusing because duplicate values  (john, BestApplicationEver). Is this really needed?  What if we forgot one of the two?

src/NLog/LogEventInfo.cs
```diff
@@ -281,16 +296,47 @@ public string FormattedMessage
         /// <summary>
         /// Gets the dictionary of per-event context properties.
         /// </summary>
-        public IDictionary<object, object> Properties
+        public IDictionary<object,object> Properties { get { return this.properties ?? (this.properties = new PropertiesDictionary()); } }
+
+        /// <summary>
+        /// Gets the named parameters extracted from parsing <see cref="Message"/> as MessageTemplate
+        /// </summary>
+        public IEnumerable<MessageTemplateParameter> MessageTemplateParameters
```
304NotModified: Could we go again to our custom interface? Even if it's only implementing `IEnumerable<MessageTemplateParameter>`, it's more flexible for the future.

snakefoot: Fixed

src/NLog/LogEventInfo.cs
snakefoot: I like for-loops better, when not needing the state-machine.

snakefoot:
@304NotModified Have squashed the commits, and made the suggested changes.

Had to re-introduce the DictionaryAdapter, as I could not ensure that it would return an enumerator of DirectoryEntry-objects.

src/NLog/Internal/MessageTemplateParameters.cs
```diff
@@ -0,0 +1,100 @@
...
+        {
+            if (parameters != null && parameters.Length > 0)
+            {
+                _parameters = new MessageTemplateParameter[parameters.Length];
+                for (int i = 0; i < parameters.Length; ++i)
+                {
+                    string parameterName;
+                    switch (i)
+                    {
+                        case 0: parameterName = "0"; break;
```
304NotModified: this is to prevent boxing, isn't?

src/NLog/Internal/PropertiesDictionary.cs
```diff
@@ -0,0 +1,642 @@
...
+        public PropertiesDictionary(IList<MessageTemplateParameter> parameterList = null)
+        {
+            MessageProperties = parameterList;
+        }
+
+        private bool IsEmpty { get { return (_eventProperties == null || _eventProperties.Count == 0) && (_messageProperties == null || _messageProperties.Count == 0); } }
+
+        public IDictionary EventContext { get { return _eventContextAdapter ?? (_eventContextAdapter = new DictionaryAdapter<object, object>(this)); } }
+
+        private Dictionary<object, PropertyValue> EventProperties
```
304NotModified: These are the regular properties, isn;t?

src/NLog/Internal/MessageTemplateParameters.cs
snakefoot: No to prevent allocation of string-objects.

src/NLog/Internal/PropertiesDictionary.cs
```diff
@@ -0,0 +1,642 @@
...
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// Dictionary that combines the standard <see cref="LogEventInfo.Properties" /> with the
+    /// MessageTemplate-properties extracted from the <see cref="LogEventInfo.Message" />.
+    /// 
+    /// The <see cref="MessageProperties" /> are returned as the first items
+    /// in the collection, and in positional order.
+    /// </summary>
+    internal sealed class PropertiesDictionary : IDictionary<object, object>
```
304NotModified: a sugesstion,

Is see a lot of special handling if `isEmpty`. I assume that is for more efficient memory usage.

Would a proxy design pattern be a good candidate here?

src/NLog/Internal/PropertiesDictionary.cs
```diff
@@ -0,0 +1,642 @@
...
+        private Dictionary<object, PropertyValue> EventProperties
+        {
+            get
+            {
+                if (_eventProperties == null)
+                {
+                    if (_messageProperties != null && _messageProperties.Count > 0)
+                    {
+                        _eventProperties = new Dictionary<object, PropertyValue>(_messageProperties.Count);
+                        if (!InsertMessagePropertiesIntoEmptyDictionary(_messageProperties, _eventProperties))
```
304NotModified: Why do we need some more code if `InsertMessagePropertiesIntoEmptyDictionary` = false?

src/NLog/Internal/PropertiesDictionary.cs
```diff
@@ -0,0 +1,642 @@
...
+                for (int i = 0; i < messageProperties.Count; ++i)
+                {
+                    eventProperties.Add(messageProperties[i].Name, new PropertyValue(messageProperties[i].Value, true));
+                }
+                return true; // We are done
+            }
+            catch (ArgumentException)
+            {
+                // Duplicate keys found, lets try again
+                for (int i = 0; i < messageProperties.Count; ++i)
```
304NotModified: Could you explain me how remove is a try-again for add?

src/NLog/Internal/PropertiesDictionary.cs
```diff
@@ -0,0 +1,642 @@
...
+                            }
+                        }
+                        continue;   // Skip already exists
+                    }
+                }
+
+                eventProperties[messageProperties[i].Name] = new PropertyValue(messageProperties[i].Value, true);
+                if (messagePropertiesUnique != null)
+                {
+                    messagePropertiesUnique.Add(messageProperties[i]);
```
304NotModified: I think we could simplify this code by yielding messageProperties (messageProperties[i] and messageProperties[y], what do you think?

304NotModified:
merged! If you have time to response to my comments, that would be great!

src/NLog/Internal/PropertiesDictionary.cs
snakefoot: Yes the IsEmpty checks is to prevent extra allocations, when operating on an empty-dictionary.

src/NLog/Internal/PropertiesDictionary.cs
snakefoot: Not sure I understand the question.

src/NLog/Internal/PropertiesDictionary.cs
snakefoot: Because then there are duplicate names, and _messageProperties must be fixed.

src/NLog/Internal/PropertiesDictionary.cs
snakefoot: Never been a big fan of yield because of the state-machine it allocates.

src/NLog/Internal/PropertiesDictionary.cs
snakefoot: Have fixed the comment

Issue #2253, opened 2017-08-17
Warning MSB3277 and .Net Standard 2.0


danielmoqvist:
**Type**: Question

**NLog version**: 5.0.0-beta09

**Platform**: .NET Core/.Net Standard, Windows 8.1 

**Problem:**
In Visual Studio 2017 Pro, version 15.3, create a new .net standard class library project and add nuget package NLog 5.0.0-Beta09.
Build the project and you get 
`Warning	MSB3277	Found conflicts between different versions of the same dependent assembly that could not be resolved.  These reference conflicts are listed in the build log when log verbosity is set to detailed.	netStandard2Test	C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\MSBuild\15.0\Bin\Microsoft.Common.CurrentVersion.targets`

And the detailed build log shows:
```
Encountered conflict between 'Reference:C:\Program Files\dotnet\sdk\NuGetFallbackFolder\netstandard.library\2.0.0\build\netstandard2.0\\ref\System.AppContext.dll' and 'Reference:C:\Program Files\dotnet\sdk\NuGetFallbackFolder\system.appcontext\4.1.0\ref\netstandard1.6\System.AppContext.dll'.  Choosing 'Reference:C:\Program Files\dotnet\sdk\NuGetFallbackFolder\netstandard.library\2.0.0\build\netstandard2.0\\ref\System.AppContext.dll' because AssemblyVersion '4.1.2.0' is greater than '4.1.0.0'.
1>    Encountered conflict between 'Reference:C:\Program Files\dotnet\sdk\NuGetFallbackFolder\netstandard.library\2.0.0\build\netstandard2.0\\ref\System.Collections.NonGeneric.dll' and 'Reference:C:\Program Files\dotnet\sdk\NuGetFallbackFolder\system.collections.nongeneric\4.0.1\ref\netstandard1.3\System.Collections.NonGeneric.dll'.  Choosing 'Reference:C:\Program Files\dotnet\sdk\NuGetFallbackFolder\netstandard.library\2.0.0\build\netstandard2.0\\ref\System.Collections.NonGeneric.dll' because AssemblyVersion '4.0.3.0' is greater than '4.0.1.0'.
[...]

PR #2263, opened 2017-08-24
Support .NET Standard 2.0 and move to VS 2017

snakefoot:
Initial attempt for converting to new VS2017 csproj file format.

- [x] NetStandard 2.0
- [x] Net45, Net40-client, Net35 (Moved net40 to net40-client nuget-lib-folder)
- [x] Silverlight 4, Silverlight 5, Windows Phone
- [x] XamarinIOS 1.0
- [x] MonoAndroid (Upgraded from 2.3 to 4.4)
- [x] OpenCover
- [x] Travis NetStandard 2.0
- [x] Travis Mono 5.2 Net45 (Ver. 5.2 needed for msbuild support)
- [x] Cherry picking from coreclr-branch (NLog 5.0 BETA)
- [x] Strong Name Version remains static (ver. 4.0.0.0)
- [x] xUnit 2.3.0-beta5 (BETA Needed for NetStandard 2.0)
- [x] xUnit 2.0 testing of net35 and net40-client
- [x] ~NLog-Docs-csproj (Maybe handled by GenerateDocumentationFile=true)~ -> not used
- [x] ~NLog-BuildDocPages~ -> not used for now
- [x] NLog-CheckSourceCode
- [x] NLog.Extended package
- [x] NLog.Schema package (XSD Generation for config-file-intellisense) -> Julian
- [x] NLog.Config package
- [x] NLog.Snippets package (Maybe it is replaced by https://github.com/NLog/NLogSnippetsVSIX)
- [x] NLog.shfbproj (Sandcastle help file) -> revert

Replaces #2175 and #2154 and resolves #2152, resolves #1954 and resolves  #1953

PR #2277, opened 2017-09-09
NetStandard -  ${WindowsIdentity} and System.Security.Principal.Windows dependency

snakefoot:


304NotModified:
Fixes https://github.com/NLog/NLog/issues/2253

src/NLog/LayoutRenderers/WindowsIdentityLayoutRenderer.cs
```diff
@@ -31,7 +31,7 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
-#if !SILVERLIGHT && !NETSTANDARD || NETSTANDARD1_3PLUS
+#if !SILVERLIGHT && !NETSTANDARD
```
304NotModified: Isn't this a quick fix? Disabling the code?

snakefoot: Sure is. But since it has been granted life by me, then I can also revoke its existence :)

luigiberrettini: You can add a dependency on 4.4 in #2263 since it requires .NET Standard 2.0

snakefoot:
Netstandard 2.0 has a default package that includes almost everything but WCF. Only Netstandard 1.X needs explicit packages specified



Sent from my Samsung device


src/NLog/LayoutRenderers/WindowsIdentityLayoutRenderer.cs
snakefoot: Netstandard 2.0 has a default package that includes almost everything but WCF. Only Netstandard 1.X needs explicit packages specified

PR #2281, opened 2017-09-15
Fix for garbled callsite info after await-calls 

### LINKED ISSUE #1805: ${Callsite} method name for async methods are not correct in .NET Core
Partly moved from https://github.com/NLog/NLog.Extensions.Logging/issues/67

Current:` Auth.Controllers.LoginController+<Post>d__4.MoveNext`
should be: `Auth.Controllers.LoginController.Post`

`callsite:CleanNamesOfAnonymousDelegates` won't change anything.

With layout: `layout="${longdate}|${uppercase:${level}}|${callsite:className=true:methodName=true}|${message}${newline}${stacktrace}"`

Using 
`${callsite:CleanNamesOfAnonymousDelegates=true:ClassName=false}
`I get
```
2016-11-29 11:53:39.2654|INFO|Logged In foo@bar.com
MoveNext
ExecutionContext.Run => ExecutionContext.RunInternal => <Post>d__4.MoveNext

```
of course I'm looking to get something like Auth.LoginController.Post (fully qualified class name + method name)

using 
`${callsite:CleanNamesOfAnonymousDelegates=true:MethodName=false}`
I get 
```
2016-11-29 11:54:58.6905|INFO|Logged In foo@bar.com
Auth.Controllers.LoginController+<Post>d__4
ExecutionContext.Run => ExecutionContext.RunInternal => <Post>d__4.MoveNext

```

Current code: 
https://github.com/NLog/NLog/blob/e77dba48c9140caeb8c0a28a54f92a3ee93b82a0/src/NLog/LayoutRenderers/CallSiteLayoutRenderer.cs#L145-L185

tkhaugen:
The current approach to the problem of garbled callsite info in async methods (skipping a stackframe in LoggerImpl) does not work for logging calls after an await-call in an async method.

This fix exploits the fact that the compiler-generated state machine is nested under the user type. It detects the state machine in the current stack frame in AppendClassName and AppendMethodName in CallSiteLayoutRenderer, then fetches class and method info from the DeclaringType property.

A property, CleanNamesOfAsyncContinuations, has been added to toggle the behaviour, following the example of CleanNamesOfAnonymousDelegates.

This fix might possibly replace the "skip-frame" approach in LoggerImpl, if desired. The callstack will not be fixed, but it isn't really fixed now anyway.

fixes #1805


304NotModified:
Thanks!

tests/NLog.UnitTests/LayoutRenderers/CallSiteTests.cs
```diff
@@ -1040,6 +1040,72 @@ public void CallSiteShouldWorkEvenInlined()
...
+                <nlog>
+                    <targets><target name='debug' type='Debug' layout='${callsite:classname=false:cleannamesofasynccontinuations=true}' /></targets>
+                    <rules>
+                        <logger name='*' levels='Debug' writeTo='debug' />
+                    </rules>
+                </nlog>");
+
+            AsyncMethod5().GetAwaiter().GetResult();
+
+            AssertDebugLastMessage("debug", nameof(AsyncMethod5));
```
304NotModified: I prefer using fixed strings (in tests) and no `nameof`, `typeof` etc. Could you please change that? (Also below)

tests/NLog.UnitTests/LayoutRenderers/CallSiteTests.cs
```diff
@@ -1040,6 +1040,72 @@ public void CallSiteShouldWorkEvenInlined()
...
+                    </rules>
+                </nlog>");
+
+            AsyncMethod5().GetAwaiter().GetResult();
+
+            var loggedCallsite = GetDebugLastMessage("debug");
+            Assert.Contains("<", loggedCallsite);
+            Assert.Contains(">", loggedCallsite);
+            Assert.Contains("+", loggedCallsite);
+            Assert.Contains("MoveNext", loggedCallsite);
```
304NotModified: If it's possible, please test with full string and equals.

304NotModified:
oops, merge was not working well. Feel free to remove the commit and force push

tests/NLog.UnitTests/LayoutRenderers/CallSiteTests.cs
tkhaugen: Sure can do

tests/NLog.UnitTests/LayoutRenderers/CallSiteTests.cs
tkhaugen: OK, I have done that now. I'm afraid it can be somewhat brittle, though, since the exact compiler-generated method name varies somewhat, ie. probably when other members are added to the method. Must be prapared to update this test when that happens.

304NotModified:
mmm, that won't worked well (the push)

304NotModified:
replaced with https://github.com/NLog/NLog/pull/2292

304NotModified:
superseded by https://github.com/NLog/NLog/pull/2292

Issue #2288, opened 2017-09-20
Unused targets logged due when inside BufferingWrapper


thesmallbang:
- Bug

**NLog version**: 4.4.12

**Platform**:  .NET Core 

**Current NLog config** (xml or C#, if relevant)

```xml
<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd"
      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
      autoReload="true"
      internalLogLevel="Warn"
      internalLogFile="logs\internal\internal-nlog.txt">

  <extensions>
    <add assembly="NLog.Web.AspNetCore" />
    <add assembly="NLog.MailKit"/>
  </extensions>
    <targets>
      <target xsi:type="BufferingWrapper"
              name="MailBuffer"
              slidingTimeout="false"
              bufferSize="100"
              flushTimeout="10000">
        <target xsi:type="Mail" name="email"
              COMMENTS ="I removed my settings"
       />
      </target>
    </targets>
 <rules>
    <logger name="*" minlevel="Warn" writeTo="MailBuffer" />
  </rules>
</nlog>

```
**I do have other rules and targets in my actual config

In case of a BUG:

- What is the current result?  The internal logger is writing Warn Unused target detected. Add a rule for this target to the configuration. TargetName: email

- What is the expected result? The email target is being used by the MailBuffer target and is not unused so nothing should be detected

- Is there a version in which it did worked? I was not previously using the BufferingWrapper as I did not have a need before the mailkit support was added.

304NotModified:
Thanks for the report! This is indeed wrong! 

Could you help us in fixing this? (unit test, PR)

thesmallbang:
I'd love to help but being fully unfamiliar with this project's code and style..

How would you approach this issue? 

Here is what I was thinking just at a glance:

1) Iterate targets and store a list of the WrappedTargets if the type is WrapperTargetBase 
2) If the normal unused check fails ignore the warning if found in the list from step 1

any thoughts or better direction from people familiar with the code would be appreciated.

304NotModified:
We prefer adding an unit test first (TDD)

- you could add this to  https://github.com/NLog/NLog/blob/master/tests/NLog.UnitTests/Config/RuleConfigurationTests.cs


- the bug is in https://github.com/NLog/NLog/blob/master/src/NLog/Config/XmlLoggingConfiguration.cs, 
(method CheckUnusedTargets)
[...]

PR #2290, opened 2017-09-21
Do not report wrapped targets as unused targets

thesmallbang:
Account for wrapped targets when determining if they are unused targets.

Related to issue: https://github.com/NLog/NLog/issues/2288


src/NLog/Targets/Wrappers/DebugTargetWrapper.cs
```diff
@@ -0,0 +1,76 @@
...
+    using System.ComponentModel;
+    using System.Threading;
+    using NLog.Common;
+    using NLog.Internal;
+
+    /// <summary>
+    /// Wrapper for unit testing with DebugTarget
+    /// </summary>
+    [Target("DebugWrapper", IsWrapper = true)]
+    public class DebugTargetWrapper : WrapperTargetBase
```
304NotModified: Please add this one to the test project, thx!

thesmallbang: Are you asking for that class to be moved into the testing namespace or for unit tests to be created for the class? I see the DebugTarget exists in the main project. Thanks in advance for the clarification.

304NotModified: Move to the test project please, so it's only for the unit tests visible. 

You're right about the debug target. But I prefer to only add public surface with a feature request.

thesmallbang: The target being in the test project does not seem to work. I get a target not found. The same thing originally was occurring (and still is) when i tried to use the MockTargetWrapper. It appears that only targets in the main project are being found.

thesmallbang: Finished. Adding the assembly reference to unit test project in the nlog config string was enough.

thesmallbang:
Anything pending here? Let me know.

304NotModified:
Only failing builds, but it looks like that are unstable tests

304NotModified:
Thanks for the changes! Looks good!

304NotModified:
merged! Thx! will be released in NLog 4.5

PR #2291, opened 2017-09-22
Improve Unit test performance

304NotModified:
smaller sleeps,
replace some integration tests with (mocked) unit tests


304NotModified:
Unit test time from ~180s -> 163sec (3x)

PR #2297, opened 2017-09-28
WhenRepeatedFilter - Log after timeout

### LINKED ISSUE #2295: Can WhenRepeatedFilter prevent logging events until time period has passed?
**Type**
Question

**NLog version**:
4.4.11

**Platform**: .NET 4.6

Hi, is it possible for the WhenRepeated filter to prevent logging the first log message?

I'd like to be able to log events only after they have been occurring for at least 30 minutes.

snakefoot:
Attempt to resolve #2295 

Added new property IncludeFirst=true/false (default=false), that controls whether it applies the configured action to the initial logevent.

When IncludeFirst=true then the initial logevent will start a timeout-period (but without it being logged), so only when the timeout-period expires then the first logevent is logged.

### DIFF
--- a/src/NLog/Filters/WhenRepeatedFilter.cs
+++ b/src/NLog/Filters/WhenRepeatedFilter.cs
@@ -60,6 +60,13 @@ namespace NLog.Filters
         public int MaxLength { get; set; }
 
         /// <summary>
+        /// Applies the configured action to the initial logevent that starts the timeout period.
+        /// Used to configure that it should ignore all events until timeout.
+        /// </summary>
+        [DefaultValue(false)]
+        public bool IncludeFirst { get; set; }
+
+        /// <summary>
         /// Max number of unique filter values to expect simultaneously
         /// </summary>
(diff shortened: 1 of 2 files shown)


304NotModified:
How does this resolve https://github.com/NLog/NLog/issues/2295?

snakefoot:
`whenRepeated action="ignore"` means that it the first logevent is logged, but then ignores everything else until timeout expires.

`whenRepeated action="log"` means that that the first logevent is skipped, and it ignores everything else until timeout expires.

snakefoot:
Could also consider not to abuse the Action-property. But just introduce a new property called skipFirst=true/false (default=false)

Mykezero:
Hi @snakefoot, this is exactly the functionality I was looking for!

304NotModified:
OK great! great work @snakefoot !

snakefoot:
Updated documentation:

https://github.com/NLog/NLog/wiki/WhenRepeated-Filter

PR #2308, opened 2017-10-05
EntityFrameworkTarget fixes #98

### LINKED ISSUE #98: Support for Entity Framework connection strings
<p>Hello. NLog does not currently seem to support Entity Framework connection strings: DatabaseTarget.InitializeTarget() throws an exception because it cannot create a DbProviderFactory from the EF providerName=&quot;System.Data.EntityClient&quot;.</p>

<p>&nbsp;</p>

<p>Unfortunately, because DatabaseTarget is sealed I can&#39;t just customize the connection string behavior as I would with log4net.</p>

<p>&nbsp;</p>

<p>Please consider adding direct support for the EF provider type (might not be too hard, a regular connection string is indeed embedded in there) or unsealing DatabaseTarget.</p>

<p>&nbsp;</p>

<p>Thank you!</p>

Misiu:
I've created new target, currently without tests.
This is my first PR for NLog so guidance is more than welcome.
Currently it doesn't support .NET Standard.

fixes #98

### DIFF
--- a/src/NLog/Targets/EntityFrameworkTarget.cs
+++ b/src/NLog/Targets/EntityFrameworkTarget.cs
@@ -31,6 +31,8 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
+#if !SILVERLIGHT && !__IOS__ && !__ANDROID__
+
 namespace NLog.Targets
 {
 
@@ -128,3 +130,5 @@ namespace NLog.Targets
 
     }
 }
+
+#endif
--- a/src/NLog/Targets/EntityFrameworkTarget.cs
+++ b/src/NLog/Targets/EntityFrameworkTarget.cs
@@ -61,13 +61,20 @@ namespace NLog.Targets
             this.Name = name;
         }
 
+
+
+        /// <summary>
+        /// Initializes the target. Can be used by inheriting classes
+        /// to initialize logging.
+        /// </summary>
         protected override void InitializeTarget()
         {
+#if !NETSTANDARD
             base.InitializeTarget();
 
-#pragma warning disable 618
+#pragma warning disable CS0618 // Type or member is obsolete
             if (UseTransactions.HasValue)
-#pragma warning restore 618
+#pragma warning restore CS0618 // Type or member is obsolete
             {
                 InternalLogger.Warn("UseTransactions is obsolete and will not be used - will be removed in NLog 6");
             }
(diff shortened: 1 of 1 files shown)


src/NLog/Targets/EntityFrameworkTarget.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// <param name="name">Name of the target.</param>
+        public EntityFrameworkTarget(string name) : this() => Name = name;
+
+        /// <summary>
+        /// Initializes the target. Can be used by inheriting classes
+        /// to initialize logging.
+        /// </summary>
+        protected override void InitializeTarget()
+        {
+#if !NETSTANDARD
```
304NotModified: I think it's better to add this to line 34, so for the whole class

src/NLog/Targets/EntityFrameworkTarget.cs
```diff
@@ -0,0 +1,134 @@
...
+            }
+            // read connection string and provider factory from entity framework connection string
+            var cs = this.ConnectionStringsSettings[ConnectionStringName];
+            if (cs == null)
+            {
+                throw new NLogConfigurationException("Connection string '" + ConnectionStringName + "' is not declared in <connectionStrings /> section.");
+            }
+
+            const string providerConnectionStringParameterName = "provider connection string";
+            const string providerParameterName = "provider";
```
304NotModified: As this one is used once, I would inline it (so no variable)

src/NLog/Targets/EntityFrameworkTarget.cs
```diff
@@ -0,0 +1,134 @@
...
+
+            var dbConnectionStringBuilder = new DbConnectionStringBuilder { ConnectionString = cs.ConnectionString };
+
+            //CurrentValues Dictionary in DbConnectionStringBuilder is case insensitive (StringComparer.OrdinalIgnoreCase)
+
+            //should we use SimpleLayout.Escape here?
+            string provider = (string)dbConnectionStringBuilder[providerParameterName];
```
304NotModified: `SimpleLayout.Escape` isn't needed here as provider is a `string` (and not a Layout)

src/NLog/Targets/EntityFrameworkTarget.cs
```diff
@@ -0,0 +1,134 @@
...
+
+#if NET3_5 || MONO
+            if (provider == null || string.IsNullOrEmpty(provider.Trim()))
```
304NotModified: there is a StringHelpers.IsNullOrWhiteSpace in NLog for this (multiple times)

Misiu:
@304NotModified I've applied all suggestions. Didn't know about StringHelpers. Now the worst part - tests.

304NotModified:
@Misiu need soms help on the tests?

Misiu:
@304NotModified it's hard to admit, but I don't have much experience with building tests. Simple one or two would help a lot, then I could try adding more.
Can DatabaseTargetTests be reused?

304NotModified:
I was checking the code for adding a test,

but it's unclear how this works. Why does this work with entity framework and the current database target not? The code is not that different?

Misiu:
I'll try to show what I'm doing on example.
I web.config I have this connection string, which is added by Entity Framework
```
<connectionStrings>
	<add name="ApplicationDbContext" connectionString="metadata=res://*/Model.csdl|res://*/Model.ssdl|res://*/Model.msl;provider=System.Data.SqlClient;provider connection string=&quot;data source=192.168.0.100;initial catalog=TEST_DB;user id=myUser;password=SecretPassword;multipleactiveresultsets=True;application name=EntityFramework&quot;" providerName="System.Data.EntityClient" />
</connectionStrings>
```

As You can see this connection string has name: `ApplicationDbContext` has value:
`metadata=res://*/Model.csdl|res://*/Model.ssdl|res://*/Model.msl;provider=System.Data.SqlClient;provider connection string=&quot;data source=192.168.0.100;initial catalog=TEST_DB;user id=myUser;password=SecretPassword;multipleactiveresultsets=True;application name=EntityFramework&quot;`

I'm extracting connectionString property from that connection string, then I'm creating `DbConnectionStringBuilder` and then I'm extracting real connection string.

This gives me:
`data source=192.168.0.100;initial catalog=TEST_DB;user id=myUser;password=SecretPassword;multipleactiveresultsets=True;application name=EntityFramework`

I can use that "clean" connection string as it was normal DatabaseTarget.
Hope this is more clear right now.

I'm using this target as so:
```
<target xsi:type="EntityFramework" name="db" connectionStringName="ApplicationDbContext">
  <commandText>EXEC Add_Log @Data = @Date, @Msg = @Message;</commandText>
  <parameter name="@Message" layout="${message} ${exception:format=tostring}"/>
  <parameter name="@Date" layout="${date}"/>
</target>
```
My main idea behind this new target was to extract connection string from entity framework and reuse it in existing target.

304NotModified:
> As You can see this connection string has name: ApplicationDbContext has value:
> metadata=res://*/Model.csdl|res://*/Model.ssdl|res://*/Model.msl;provider=System.Data.SqlClient;provider connection string=&quot;data source=192.168.0.100;initial catalog=TEST_DB;user id=myUser;password=SecretPassword;multipleactiveresultsets=True;application name=EntityFramework&quot;
> 
> I'm extracting connectionString property from that connection string, then I'm creating DbConnectionStringBuilder and then I'm extracting real connection string.

A well then we should write a test/method  which goes from the metadata to the connectionstring.

Maybe we could implement it in the current DataBase target, as I think we only need that method?

304NotModified:
superseded by https://github.com/NLog/NLog/pull/2510, fixed in NLog 4.5

thanks @Misiu for the original code!

Misiu:
@304NotModified glad I could help :)
I'll wait for 4.5 and update my apps.

PR #2314, opened 2017-10-08
Refactor InternalLogger class

ie-zero:
Refactor InternalLogger class to improve code readabiity and create further opportunities to improve the code.

There are no functional changes or changes to the public interface.

The method Write() in InternalLogger class split into smaller methods
* FormatMessage()
* WriteToLogFile()
* WriteToTextWriter()
* WriteToConsole()
* WriteToErrorConsole()

LoggingEnabled() method renamed to IsLoggingEnabled() in alignment to IsSeriousException(). It also convey better to the reader the purpose of the method.

### DIFF
--- a/src/NLog/Common/InternalLogger.cs
+++ b/src/NLog/Common/InternalLogger.cs
@@ -288,7 +288,7 @@ namespace NLog.Common
             var formattedMessage =
                 (args == null) ? message : string.Format(CultureInfo.InvariantCulture, message, args);
 
-            var builder = new StringBuilder(formattedMessage.Length + TimeStampFormat.Length + ex?.ToString()?.Length ?? 0 + 25);
+            var builder = new StringBuilder(formattedMessage.Length + TimeStampFormat.Length + (ex?.ToString()?.Length ?? 0) + 25);
             if (IncludeTimestamp)
             {
                 builder
--- a/src/NLog/Common/InternalLogger.cs
+++ b/src/NLog/Common/InternalLogger.cs
@@ -288,7 +288,7 @@ namespace NLog.Common
             var formattedMessage =
                 (args == null) ? message : string.Format(CultureInfo.InvariantCulture, message, args);
 
-            var builder = new StringBuilder(formattedMessage.Length + TimeStampFormat.Length + 8);
+            var builder = new StringBuilder(formattedMessage.Length + TimeStampFormat.Length + ex?.ToString()?.Length ?? 0 + 25);
             if (IncludeTimestamp)
             {
                 builder
@@ -344,6 +344,14 @@ namespace NLog.Common
                    LogWriter != null;
         }
 
+        /// <summary>
+        /// Write internal messages to the log file defined in <see cref="LogFile"/>.
+        /// </summary>
+        /// <param name="message">Message to write.</param>
+        /// <remarks>
+        /// Message will be logged only when the property <see cref="LogFile"/> is not <c>null</c>, otherwise the
+        /// method has no effect.
+        /// </remarks>
         private static void WriteToLogFile(string message)
         {
             var logFile = LogFile;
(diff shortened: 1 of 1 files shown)


src/NLog/Common/InternalLogger.cs
```diff
@@ -336,6 +280,39 @@ private static void Write([CanBeNull]Exception ex, LogLevel level, string messag
...
 
+        private static string FormatMessage([CanBeNull]Exception ex, LogLevel level, string message, [CanBeNull]object[] args)
+        {
+            const string TimeStampFormat = "yyyy-MM-dd HH:mm:ss.ffff";
+            const string FieldSeparator = " ";
+
+            var formattedMessage =
+                (args == null) ? message : string.Format(CultureInfo.InvariantCulture, message, args);
+
+            var builder = new StringBuilder(formattedMessage.Length + TimeStampFormat.Length + 8);
```
304NotModified: Also small change here I see, maybe also check `IncludeTimestamp`?

src/NLog/Common/InternalLogger.cs
```diff
@@ -367,6 +344,64 @@ private static bool LoggingEnabled(LogLevel logLevel)
                    LogWriter != null;
         }
 
+        private static void WriteToLogFile(string message)
```
304NotModified: Please add some docs, most important is here that logging will be skipped if not enabled. (multiple times)

ie-zero: I will do

src/NLog/Common/InternalLogger.cs
```diff
@@ -288,7 +288,7 @@ private static string FormatMessage([CanBeNull]Exception ex, LogLevel level, str
             var formattedMessage =
                 (args == null) ? message : string.Format(CultureInfo.InvariantCulture, message, args);
 
-            var builder = new StringBuilder(formattedMessage.Length + TimeStampFormat.Length + 8);
+            var builder = new StringBuilder(formattedMessage.Length + TimeStampFormat.Length + ex?.ToString()?.Length ?? 0 + 25);
```
304NotModified: 0 + 25? Should be 8+25?

or 0 + 0+ 0+ 0+ 0+ 0+ 0+ 0 + 25 :P

ie-zero: ```formattedMessage.Length``` and ```TimeStampFormat.Length``` (constant length) should never be 0. ```ex?.ToString()?.Length ?? 0``` might be zero and the rest of white-spaces the literal 'Exception' and the LogLevel as a string sum up, if I recall correctly, to 23 characters.

To be honest, I am not certain if pre-calculating the length of the expected string in such detail for optimisation is positive. 

What do you think?

304NotModified: true, pre-optimized could be done later. 

`0 + 25` looks just odd. Its 25 then.

src/NLog/Common/InternalLogger.cs
```diff
@@ -336,6 +280,39 @@ private static void Write([CanBeNull]Exception ex, LogLevel level, string messag
...
 
+        private static string FormatMessage([CanBeNull]Exception ex, LogLevel level, string message, [CanBeNull]object[] args)
+        {
+            const string TimeStampFormat = "yyyy-MM-dd HH:mm:ss.ffff";
+            const string FieldSeparator = " ";
+
+            var formattedMessage =
+                (args == null) ? message : string.Format(CultureInfo.InvariantCulture, message, args);
+
+            var builder = new StringBuilder(formattedMessage.Length + TimeStampFormat.Length + ex?.ToString()?.Length ?? 0 + 25);
```
304NotModified: Resharper says this is  ` ex?.ToString()?.Length ?? (0 + 25)`

So this is wrong ;)

304NotModified:
Merged. Thanks!

PR #2316, opened 2017-10-09
LogManager.LogFactory public

304NotModified:
needed for (in NLog.Web.AspNetCore)

    NLogBuilder.LoadConfig("NLog.config").GetCurrentClassLogger()

(see https://github.com/NLog/NLog.Web/pull/191)

@snakefoot your opinion please :)


snakefoot:
Think it is a good idea. Always found it weird, that one had to create a Logger-object to get access to the LogFactory. The LogFactory is good because it allows you to create your own extension-methods, which is hard to do with the LogManager.

PR #2319, opened 2017-10-09
NLog - MessageTemplateParameter - IsReservedFormat

snakefoot:
Moved MessageTemplate UnitTests into the MessageTemplates-namespace.

### DIFF
--- a/src/NLog/MessageTemplates/MessageTemplateParameter.cs
+++ b/src/NLog/MessageTemplates/MessageTemplateParameter.cs
@@ -62,6 +62,28 @@ namespace NLog.MessageTemplates
         public readonly string Format;
 
         /// <summary>
+        /// Checks if the <see cref="Format"/> contains reserved letters ('@', '$', 'l')
+        /// </summary>
+        public bool IsReservedFormat
+        {
+            get
+            {
+                switch (Format)
+                {
+                    case "@":
+                    case "$":
+                        return true;
+                    case "l":
+                        if (Value is string || Value == null || Value is char)
+                            return true;
+                        else
+                            return false;
+                }
+                return false;
+            }
+        }
+
+        /// <summary>
         /// Constructs a single message template parameter
         /// </summary>
         /// <param name="name">Parameter Name</param>
--- a/src/NLog/MessageTemplates/TemplateRenderer.cs
+++ b/src/NLog/MessageTemplates/TemplateRenderer.cs
@@ -89,7 +89,16 @@ namespace NLog.MessageTemplates
                         {
                             messageTemplateParameters = new MessageTemplateParameter[parameters.Length];
                         }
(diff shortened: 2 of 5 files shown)


304NotModified:
in what case would you like to use the IsReservedFormat?

src/NLog/MessageTemplates/MessageTemplateParameter.cs
```diff
@@ -62,6 +62,28 @@ public struct MessageTemplateParameter
         public readonly string Format;
 
         /// <summary>
+        /// Checks if the <see cref="Format"/> contains reserved letters ('@', '$', 'l')
+        /// </summary>
+        public bool IsReservedFormat
```
304NotModified: Another option is to create a static method, so we don't have to construct a new MessageTemplateParameter to check the format

snakefoot:
> in what case would you like to use the IsReservedFormat?

Was thinking that I wanted to help the Seq-engine. It wants to pass the parameters through its own custom formatter. So it wants to know if it is "$" or "@" or "l", but I also want to tell users they should not use the format-string together with string.Format.

src/NLog/MessageTemplates/MessageTemplateParameter.cs
snakefoot: It is mostly an informational courtesy for outsiders. No one but Seq, and a few others will be messing with the MessageTemplateParameter

snakefoot: And because I have been evil and merged CaptureType into Format

snakefoot: Since door was already open with the custom format "l" (lowercase L)

304NotModified: OK, but why not a static method? I think that's more clean.

304NotModified: Another option is a C# extension, benefit, it "pollutes" the class MessageTemplateParameter less.

snakefoot: Both the C# extension and the static-method that takes the struct as parameter will require stack-allocation of the MessageTemplateParameter. Not sure I understand the problem with a struct-property.

snakefoot: But we can make it a static-method that takes the format-string as input. Would that be better?

304NotModified: Well it's a bit odd here, but the struct reason is a good reason to have it like this.

snakefoot:
@304NotModified If something should be changed, then please tell. Would prefer to have this included before nblumhardt does the testdrive with https://github.com/datalust/nlog-targets-seq

304NotModified:
merged!

PR #2320, opened 2017-10-09
ValueSerializer: support bool

304NotModified:


snakefoot:
Could you move bool-support-cast below the string-support-cast? Believe that "string" should be first to be checked (performance reason as it is the most likely type to be used)

304NotModified:
is the bool check not faster? or is this because of boxing?

304NotModified:
anyway, moved :)

snakefoot:
> is the bool check not faster? or is this because of boxing slower?

Any object-casting is slow. So if 90 pct is string, then it will be faster for 90 pct of the cases.

304NotModified:
Another option is to use `getType` once, and then a switch/dictionary. Do you think that's a better (read faster) option?

snakefoot:
> Another option is to use getType once, and then a switch/dictionary. Do you think that's a better (read faster) option?

Then I would probably use `Convert.GetTypeCode(value)` like done in DefaultJsonSerializer.

PR #2326, opened 2017-10-09
NLog - ValueSerializer - Faster integer and enum

### LINKED ISSUE #2323: ValueSerializer: use  Convert.GetTypeCode(value)
https://github.com/NLog/NLog/blob/master/src/NLog/MessageTemplates/ValueSerializer.cs

use  Convert.GetTypeCode(value) instead of `as`? 

Needs a benchmark I guess

snakefoot:
Trying to resolve #2323

### DIFF
--- a/src/NLog/MessageTemplates/ValueSerializer.cs
+++ b/src/NLog/MessageTemplates/ValueSerializer.cs
@@ -59,6 +59,8 @@ namespace NLog.MessageTemplates
         private const int MaxValueLength = 512 * 1024;
         private const string LiteralFormatSymbol = "l";
 
+        private readonly MruCache<Enum, string> _enumCache = new MruCache<Enum, string>(1500);
+
         /// <inheritDoc/>
         public bool SerializeObject(object value, string format, IFormatProvider formatProvider, StringBuilder builder)
         {
@@ -96,9 +98,6 @@ namespace NLog.MessageTemplates
         {
             // todo support all scalar types: 
 
-
-    
-
             // todo byte[] - hex?
             // todo datetime, timespan, datetimeoffset
             // todo nullables correct?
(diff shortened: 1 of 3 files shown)


src/NLog/MessageTemplates/ValueSerializer.cs
```diff
@@ -59,6 +59,8 @@ private ValueSerializer()
         private const int MaxValueLength = 512 * 1024;
         private const string LiteralFormatSymbol = "l";
 
+        private readonly MruCache<Enum, string> _enumCache = new MruCache<Enum, string>(10000);
```
304NotModified: isn't that a bit big? (10000)

snakefoot: No idea. Just chose the same as for the DefaultJsonSerializer. Maybe there are more than 10000 enums in the world?

snakefoot:
@304NotModified Have cheated a little as Integer.ToString is pretty much the same in all languages. And have also performed the same cheating for Enum.

src/NLog/MessageTemplates/ValueSerializer.cs
snakefoot: Should we change it to 5000 for the both of them ?

304NotModified:
> @304NotModified Have cheated a little as Integer.ToString is pretty much the same in all languages (when no format-string). And have also performed the same cheating for Enum (when no format string).

No problem if it's document in the source ;)

Otherwise it could get lost in refactoring

snakefoot:
> No problem if it's document in the source ;)

Added a comment about the special optimization.

src/NLog/MessageTemplates/ValueSerializer.cs
304NotModified: was more thinking of 1000. Or even lower. Could not image your using 1000 different enum values in 1 project.

304NotModified:
do we need to pref test this? Or are you sure it's faster?

snakefoot:
>do we need to pref test this?

The enum printing is 100 pct faster, and the integer printing is 50 pct. faster.

src/NLog/MessageTemplates/ValueSerializer.cs
snakefoot: Now changed it to 1500 (also for the Json-Converter)

snakefoot:
@304NotModified Should be ready for review (again) :)

304NotModified:
great! Merged!

Issue #2339, opened 2017-10-11
Structured logging in 4.5-beta4 feedback


nblumhardt:
**Type** (choose one):

- Feature request (essentially)

**NLog version**: 4.5-beta4

**Platform**: all

Hi! I've just updated _Seq.Targets.NLog_ to the new 4.5-beta04 🎉 

A few things seem to be missing or are unclear:

 - `MessageTemplateParameter.Format` is null in a case like `"The number is {0:000}"` - I would expect the format to be `"000"` in this case
 - In a positional template, there doesn't seem to be a way to get the position of a placeholder except by parsing the parameter's `Name`
 - By including `@` and `$` directives in `MessageTemplateParameter.Format`, it's ambiguous whether they constitute capture types or .NET format strings (like `$` for currency)

With my limited understanding of the feature at this point, it appears these could be resolved cleanly by:

 - Exposing `CaptureType` on `MessageTemplateParameter`, and _not_ duplicating the capture type in `Format`
 - Always using `Format` for the text-formatting information
 - Providing `int` or `int? Position` on `MessageTemplateParameter`, with a value provided when `IsPositional` is true

Can you please let me know if I'm on the right track, or misunderstanding how NLog should work, here?

Congratulations on shipping the new beta! :-)

snakefoot:
Yes I have not given the MessageTemplateParameters so much love when IsPositional. Will try and improve this.

Did not think about $ being used as actual format-string. Thought I could use it as reserved word.

nblumhardt:
Thanks for the reply. (I just found #2332 - I can close this and post over there, if you'd prefer - sorry about that!)

304NotModified:
> Thanks for the reply. (I just found #2332 - I can close this and post over there, if you'd prefer - sorry about that!)

NP! Thanks for the fast feedback!

snakefoot:
> @nblumhardt or .NET format strings (like $ for currency)

I have looked at the format strings, and I can see one can use "C" to output as localized currency (Ex $), but I cannot see that one can use $ and @ in the format-string.

https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings

If we add the CaptureType to the MessageTemplateParameter, then must also be added as input parameter to the IValueSerializer. What do you think @304NotModified  ?

> Providing int or int? Position on MessageTemplateParameter, with a value provided when 

[...]

PR #2340, opened 2017-10-11
NLog - MessageTemplateParameters - Always parse when IsPositional

snakefoot:
Fixes the IsPositional parsing issue reported by #2339

### DIFF
--- a/src/NLog/MessageTemplates/MessageTemplateParameters.cs
+++ b/src/NLog/MessageTemplates/MessageTemplateParameters.cs
@@ -63,6 +63,8 @@ namespace NLog.MessageTemplates
         /// <summary>
         /// Constructore for positional parameters
         /// </summary>
+        /// <param name="message">log message including any parameter placeholders</param>
+        /// <param name="parameters">All the parameters</param>
         public MessageTemplateParameters(string message, object[] parameters)
         {
             var hasParameters = parameters != null && parameters.Length > 0;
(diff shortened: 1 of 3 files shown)


src/NLog/MessageTemplates/MessageTemplateParameters.cs
```diff
@@ -83,60 +85,35 @@ public MessageTemplateParameters(IList<MessageTemplateParameter> parameters)
...
-                            break;
-                        case 9:
-                            parameterName = "9";
-                            break;
-                        default:
-                            parameterName = i.ToString(System.Globalization.CultureInfo.InvariantCulture);
-                            break;
+                        var hole = templateEnumerator.Current.Hole;
+                        if (hole.Index == -1)
+                            templateParameters.Add(new MessageTemplateParameter(hole.Name, parameters[holeIndex++], hole.Format));
```
304NotModified: as we're using .MoveNext here, isn't `return yield` a better way?

src/NLog/MessageTemplates/MessageTemplateParameters.cs
```diff
@@ -61,15 +63,15 @@ internal class MessageTemplateParameters : IMessageTemplateParameters
         /// <summary>
         /// Constructore for positional parameters
```
304NotModified: typo ;)

src/NLog/MessageTemplates/MessageTemplateParameters.cs
```diff
@@ -61,15 +63,15 @@ internal class MessageTemplateParameters : IMessageTemplateParameters
...
         /// </summary>
-        public MessageTemplateParameters(object[] parameters)
+        public MessageTemplateParameters(string message, object[] parameters)
```
304NotModified: This is the original non-formatted message, isn't?

PR #2349, opened 2017-10-12
NLog - NETSTANDARD1_5 (Fix uppercase with culture)

snakefoot:


snakefoot:
Yes, but I seldom see configurations that are not invariant. So it would probably have taken a while for it to be discovered.

PR #1691, opened 2016-10-08
AppDomainWrapper / IAppDomain moved to internal

304NotModified:
Not sure if we should include this for NLog 4.x or 5

there are more `public` classes in "internal", e.g. for NLog.Extended

Also Not sure if this is used in external libs, e.g. NLog.Web

- [ ] update: IAppDomain is needed to be public, the wrapper doensn't

This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1691)

PR #2354, opened 2017-10-13
NLog - NETSTANDARD1_5 (revert breaking change)

snakefoot:
AppDomainWrapper is public (See also #1691)


304NotModified:
Do you mean: #1691 should be closed?

snakefoot:
> Do you mean: #1691 should be closed?

Nope. #1691 is a good, but breaking change. Just wanted to repeat that I look forward to it being applied.

PR #2357, opened 2017-10-14
ExceptionLayoutRenderer - Support Serialize Format

snakefoot:
Made some minor optimizations:
- Skip preparation for special handling of AggregateException if MaxInnerExceptionLevel = 0 (Default)
- Skip allocation of extra StringBuilder-objects for each render-format, but just use the provided one.

### DIFF
--- a/src/NLog/Config/ExceptionRenderingFormat.cs
+++ b/src/NLog/Config/ExceptionRenderingFormat.cs
@@ -71,6 +71,10 @@ namespace NLog.Config
         /// <summary>
         /// Appends the contents of an Exception's Data property to the specified target.
         /// </summary>
-        Data = 6
+        Data = 6,
+        /// <summary>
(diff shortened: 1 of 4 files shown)


snakefoot:
Suddenly became a very good test of the Json-serializer. Discovered that NetStandard attempted to access NonPublic properties because of GetRuntimeProperties().

Also curious if the internal Json-Converter should not recognize Fields and not just Properties?

304NotModified:
> Also curious if the internal Json-Converter should not recognize Fields and not just Properties?

Nog sure. AFAIK anonymous objects creates properties and so can't think of a cade where a field is needed. Bit maybe it should be an option.

snakefoot:
> maybe it should be an option.

Will leave this task for future generations :)

304NotModified:
:+1:

PR #2359, opened 2017-10-15
WrapperLayoutRendererBase - Transform with access to LogEventInfo

### LINKED ISSUE #1460: WrapperLayoutRendererBase.Transform should recieve the LogEventInfo
current:

```
    public abstract class WrapperLayoutRendererBase : LayoutRenderer
    {
        /// <summary>
        /// Gets or sets the wrapped layout.
        /// 
        /// [DefaultParameter] so Inner: is not required if it's the first
        /// </summary>
        /// <docgen category='Transformation Options' order='10' />
        [DefaultParameter]
        public Layout Inner { get; set; }

        /// <summary>
        /// Renders the inner message, processes it and appends it to the specified <see cref="StringBuilder" />.
        /// </summary>
        /// <param name="builder">The <see cref="StringBuilder"/> to append the rendered data to.</param>
        /// <param name="logEvent">Logging event.</param>
        {
            string msg = this.RenderInner(logEvent);
            builder.Append(this.Transform(msg));
        }

        /// <summary>
        /// Transforms the output of another layout.
        /// </summary>
        /// <param name="text">Output to be transform.</param>
        /// <remarks>If the <see cref="LogEventInfo"/> is needed, overwrite <see cref="Append"/>.</remarks>
        /// <returns>Transformed text.</returns>
        protected abstract string Transform(string text);
```

But Transform does need sometimes the LogEvent.

snakefoot:
Resolves #1460 (And matches  #2358)

### DIFF
--- a/src/NLog/LayoutRenderers/Wrappers/WrapperLayoutRendererBase.cs
+++ b/src/NLog/LayoutRenderers/Wrappers/WrapperLayoutRendererBase.cs
@@ -65,14 +65,24 @@ namespace NLog.LayoutRenderers.Wrappers
         protected override void Append(StringBuilder builder, LogEventInfo logEvent)
         {
             string msg = RenderInner(logEvent);
-            builder.Append(Transform(msg));
+            builder.Append(Transform(logEvent, msg));
+        }
+
+        /// <summary>
+        /// Transforms the output of another layout.
+        /// </summary>
+        /// <param name="logEvent">Logging event.</param>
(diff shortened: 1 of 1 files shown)


304NotModified:
Not a breaking change, is it? (I was wrong in https://github.com/NLog/NLog/issues/1460 then)

snakefoot:
Nope. Not unless some have already declared a method called Transform with the same signature on their derived class. But then it is suddenly very hard to make any changes in NLog.

PR #2361, opened 2017-10-15
Use expression-bodied members

c0shea:
Replace code braces in properties with expression-bodied members in C# 6.

For example, this:
```
public static bool IsTraceEnabled
{
    get { return LogLevel.Trace >= LogLevel; }
}
```

becomes this:
`public static bool IsTraceEnabled => LogLevel.Trace >= LogLevel;`


304NotModified:
thanks again!

PR #2362, opened 2017-10-15
NLog - NETSTANDARD1_5 (Cleanup package references)

snakefoot:
Removed the System.Runtime.InteropServices.RuntimeInformation by using the implicit NetStandard.Library 1.6.0

Apparently one should not reference System.Runtime.InteropServices.RuntimeInformation directly


304NotModified:
looks good!

PR #2363, opened 2017-10-16
 support database parameter DbType and value convertion

hubo0831:


304NotModified:
Thanks! Will looks at this in-depth this week/weekend

src/NLog/Targets/DatabaseParameterConverter.cs
```diff
@@ -0,0 +1,168 @@
...
+    using System.Globalization;
+    using System.Reflection;
+    using NLog.Config;
+    using NLog.Internal;
+    using NLog.Layouts;
+
+    /// <summary>
+    /// Represents SQL command parameter converter.
+    /// </summary>
+    public class DatabaseParameterConverter
```
304NotModified: this could be internal?

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -263,10 +265,30 @@ public DatabaseTarget(string name) : this()
...
+        /// <docgen category='SQL Statement' order='12' />
+        [DefaultValue("DbType")]
+        public string ParameterDbTypePropertyName { get; set; }
+
+        /// <summary>
+        /// Gets or sets converter type of the SQL command parameter value.
+        /// </summary>
+        /// <docgen category='SQL Statement' order='13' />
+        [DefaultValue(typeof(DatabaseParameterConverter))]
+        public Type ParameterConverterType { get; set; }
```
304NotModified: this class is not sealed, so I think subclassing is a better option here (with a virtual method). Do you agree?

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -593,6 +626,33 @@ private void WriteEventToDatabase(LogEventInfo logEvent)
...
+        /// </summary>
+        protected void EnsureResolveParameterInfo(IDbDataParameter p)
+        {
+            if (this.ParameterConverter == null)
+            {
+                lock (this.SyncRoot)
+                {
+                    if (this.ParameterConverter == null)
+                    {
+                        var converter = (DatabaseParameterConverter)Activator.CreateInstance(this.ParameterConverterType);
```
304NotModified: as proposed: replace with virtual call

304NotModified:
thx, added some comments.

src/NLog/Targets/DatabaseParameterConverter.cs
hubo0831: No, Others can override this class to support more DbType.

src/NLog/Targets/DatabaseTarget.cs
hubo0831: Yes, class 'DatabaseParameterConverter' only implement general methods, others can override this class to support more DbType..

hubo0831:
Sorry, I don't comment instantly, I am on a business trip.

src/NLog/Targets/DatabaseTarget.cs
hubo0831: class 'DatabaseTarget' don't need to change, we only inherit from 'DatabaseParameterConverter' to support more DbType.

304NotModified:
Thanks for the great input!

I have combined and refactor it to #2899 and is expected to be released in NLog 4.6

PR #2377, opened 2017-10-30
NDLC - Perform low resolution scope timing

snakefoot:
Implemented async scope timing

### DIFF
--- a/src/NLog/LayoutRenderers/NdlcTimingLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/NdlcTimingLayoutRenderer.cs
@@ -1,0 +1,96 @@
+﻿// 
+// Copyright (c) 2004-2017 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without 
+// modification, are permitted provided that the following conditions 
+// are met:
+// 
+// * Redistributions of source code must retain the above copyright notice, 
+//   this list of conditions and the following disclaimer. 
+// 
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution. 
+// 
+// * Neither the name of Jaroslaw Kowalski nor the names of its 
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission. 
+// 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE 
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE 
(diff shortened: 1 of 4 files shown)


src/NLog/LayoutRenderers/NdlcTimingLayoutRenderer.cs
```diff
@@ -0,0 +1,97 @@
...
+                {
+                    TimeSpan duration = scopeBegin != DateTime.MinValue ? DateTime.UtcNow - scopeBegin : TimeSpan.Zero;
+                    if (duration < TimeSpan.Zero)
+                        duration = TimeSpan.Zero;
+
+                    var formatProvider = GetFormatProvider(logEvent, null);
+#if !NET3_5
+                    builder.Append(duration.ToString(Format, formatProvider));
+#else
+                    builder.Append(duration.ToString());
```
304NotModified: formatProvider not needed here?

src/NLog/NestedDiagnosticsLogicalContext.cs
```diff
@@ -90,6 +90,35 @@ public static object PopObject()
...
 
         /// <summary>
+        /// Peeks the current scope, and returns its start time
+        /// </summary>
+        /// <returns>Scope Creation Time</returns>
+        public static DateTime PeekCurrentScopeBeginTime()
+        {
+            var currentContext = GetThreadLocal();
+            if (currentContext == null)
+                return DateTime.MinValue;
```
304NotModified: what about nullable datetimes?

src/NLog/NestedDiagnosticsLogicalContext.cs
```diff
@@ -90,6 +90,35 @@ public static object PopObject()
         }
 
         /// <summary>
+        /// Peeks the current scope, and returns its start time
+        /// </summary>
+        /// <returns>Scope Creation Time</returns>
+        public static DateTime PeekCurrentScopeBeginTime()
```
304NotModified: could this be internal? (2x)

snakefoot:
> could you please explain me in short why we need this? thx!

If you create a scope to mark that you have started a web-request, then you can use this to extract the timings for this scoped web-request (when logging messages that belongs to the web-request).

I guess one could also implement this by calling NDLC.Push with a complex object, but I guess we then need to provide a PeekLast and PeekFirst to retrieve this object in a custom layout-renderer.

src/NLog/LayoutRenderers/NdlcTimingLayoutRenderer.cs
snakefoot: There is no TimeSpan.ToString with Format or formatProvider in NET3_5

src/NLog/NestedDiagnosticsLogicalContext.cs
snakefoot: Fixed

src/NLog/NestedDiagnosticsLogicalContext.cs
snakefoot: I prefer DateTime.MinValue as magic value, don't need another one.

src/NLog/LayoutRenderers/NdlcTimingLayoutRenderer.cs
304NotModified: Indeed, but I think the call ` var formatProvider = GetFormatProvider(logEvent, null);` is not needed.

snakefoot: Fixed

PR #2378, opened 2017-10-30
Target - Precalculate Layout should ignore sub-layouts for complex layouts

snakefoot:
Performance optimization for async-target, so it will not render and string-allocate for each individual sub-layout.

This will improve performance for complex layouts like the JsonLayout (Has a sub-layout for each json-attribute, but they don't need to be rendered individually, only the top JsonLayout).

### DIFF
--- a/src/NLog/Layouts/CompoundLayout.cs
+++ b/src/NLog/Layouts/CompoundLayout.cs
@@ -34,9 +34,10 @@
 
 namespace NLog.Layouts
 {
-    using Config;
+    using System.Linq;
     using System.Collections.Generic;
     using System.Text;
+    using NLog.Config;
 
     /// <summary>
     /// A layout containing one or more nested layouts.
@@ -103,6 +104,15 @@ namespace NLog.Layouts
             foreach (var layout in Layouts)
                 layout.Close();
             base.CloseLayout();
+        }
+
+        /// <summary>
+        /// Generate description of Compound Layout
+        /// </summary>
+        /// <returns>Compound Layout String Description</returns>
+        public override string ToString()
+        {
+            return ToStringWithNestedItems(Layouts, l => l.ToString());
         }
     }
 }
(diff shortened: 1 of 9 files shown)


src/NLog/Layouts/CsvLayout.cs
```diff
@@ -301,5 +301,24 @@ protected override void RenderFormattedMessage(LogEventInfo logEvent, StringBuil
...
+        public override string ToString()
+        {
+            if (Columns?.Count > 0)
+            {
+                var columnSelection = System.Linq.Enumerable.Select(Columns, c => c.Name);
+#if NET3_5
+                var columnNames = System.Linq.Enumerable.ToArray(columnSelection);
+#else
+                var columnNames = columnSelection;
+#endif
```
304NotModified: could you please move this to a CollectionHelpers? (or LinqHelpers)

snakefoot: Have sacrificed some memory by always doing ToArray before calling string.Join

src/NLog/Layouts/CompoundLayout.cs
```diff
@@ -104,5 +105,19 @@ protected override void CloseLayout()
...
+        /// <summary>
+        /// Generate description of Compound Layout
+        /// </summary>
+        /// <returns>Compound Layout String Description</returns>
+        public override string ToString()
+        {
+            if (Layouts?.Count > 0)
+            {
+                var layoutDetails = Layouts.Select(l => l.ToString()).ToArray();
+                return string.Concat(GetType().Name, "=", string.Join("|", layoutDetails));
```
304NotModified: Could we write an helper in the base class? 

I'm trying to prevent code duplication to keep things as consistent as possible.

snakefoot: Have created an internal method for sub-classes to use, that sacrifices some more memory.

304NotModified:
merged, thx!

PR #2384, opened 2017-11-07
InternalLogger should work when NetCore loads NetFramework DLL

snakefoot:
Make NLog InternalLogger a little more resilient.


304NotModified:
👍

PR #2386, opened 2017-11-13
LogEventInfo.StackTrace moved into CallSiteInformation

### LINKED ISSUE #2382: Nlog ${callsite-linenumber} not working properly with async methods
I am using NLog v4.4.12.

I have an async method where I must log info with line numbers being included. The problem is that if I log anything before a call to async method, line lumber is always 0. However, if I log anything after a call to async method, line number always prints correctly.

    public class TestController : ApiController {
        private static HttpClient Client = new HttpClient(new HttpClientHandler { Proxy = null, UseProxy = false }){ Timeout = TimeSpan.FromSeconds(double.Parse(WebConfigurationManager.AppSettings["httpTimeout"])) };
        private static readonly Logger logger = LogManager.GetCurrentClassLogger();

        public async Task DoSomething(int clientId)
        {
            logger.Info($"Here, line number is always 0");

            try
            {
                var clientBooks = new List<string>();
                Client.DefaultRequestHeaders.Accept.Clear();
                Client.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));

                using (var response = await Client.GetAsync(new Uri(WebConfigurationManager.AppSettings["cb"] + "books?clientId=" + clientId)))
                {
                    if (response.IsSuccessStatusCode)
                    {
                        clientBooks = await response.Content.ReadAsAsync<List<string>>();
                    }
                    else
                    {
                         throw new Exception($"Error getting books");
                    }
                }

                logger.Info($"Here, line number always prints correctly");

                // Here I have some other logic ...
            }
            catch (Exception ex)
            {
                logger.Error(ex);
            }

            logger.Info($"Here, line number always prints correctly");
        }
    }

My NLog configuration:

```
<nlog autoReload="true" xmlns="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
    <variable name="logDirectory" value="${basedir}/Logs/${shortdate}" />
    <variable name="format" value="${longdate}|${activityid}|${logger}|Line:${callsite-linenumber}|${level:uppercase=true}|${message}" />
    <targets>
      <target name="fileLogErrors" xsi:type="File" fileName="${logDirectory}/errors.txt" layout="${format}" encoding="utf-8" />
      <target name="fileLog" xsi:type="File" fileName="${logDirectory}/all.txt" layout="${format}" encoding="utf-8" />
    </targets>
    <rules>
      <logger name="*" minlevel="Warn" writeTo="fileLogErrors" />
      <logger name="*" minlevel="Info" writeTo="fileLog" />
    </rules>
 </nlog>
```

Here is the log that I have after running actual code (log format is ${stacktrace}|Line:${callsite-linenumber}):

```
AsyncTaskMethodBuilder.Start => <<UpdateAttributes>b__0>d.MoveNext => AttributesBaseController.UpdateEquationAccountsInfo|Line:0
ExecutionContext.RunInternal => MoveNextRunner.InvokeMoveNext => <UpdateEquationAccountsInfo>d__9.MoveNext|Line:457
```

snakefoot:
Attempt to resolve #2382 

- Introduced internal CallsiteInformation-class for handling Callsite-state. Also when using Caller Information Attributes.
- Deprecated the old calculation of UserStackFrameNumber, but still returns the legacy value for external partners. Instead the method and filenumber-detail is extracted from the MoveNext-MethodBase.
- Changed Log4JXmlEventLayoutRenderer to handle async-MoveNext out of the box.
- Added lots of TODOs about removing the use of LogEventInfo.Properties for storing Caller Information Attributes (NLog-fluent)

### DIFF
--- a/src/NLog/Internal/CallSiteInformation.cs
+++ b/src/NLog/Internal/CallSiteInformation.cs
@@ -1,0 +1,146 @@
+﻿// 
+// Copyright (c) 2004-2017 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without 
+// modification, are permitted provided that the following conditions 
+// are met:
+// 
+// * Redistributions of source code must retain the above copyright notice, 
+//   this list of conditions and the following disclaimer. 
+// 
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution. 
+// 
+// * Neither the name of Jaroslaw Kowalski nor the names of its 
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission. 
+// 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE 
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE 
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE 
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR 
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS 
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN 
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) 
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF 
+// THE POSSIBILITY OF SUCH DAMAGE.
+// 
+
+namespace NLog.Internal
+{
+    using System;
+    using System.Diagnostics;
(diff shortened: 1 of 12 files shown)


src/NLog/LayoutRenderers/AllEventPropertiesLayoutRenderer.cs
```diff
@@ -135,33 +160,43 @@ protected override void Append(StringBuilder builder, LogEventInfo logEvent)
         /// <summary>
         /// The names of caller information attributes.
         /// https://msdn.microsoft.com/en-us/library/hh534540.aspx
+        /// TODO NLog ver. 5 - Remove these properties
         /// </summary>
-        private static HashSet<string> CallerInformationAttributeNames = new HashSet<string>
+        private static List<string> CallerInformationAttributeNames = new List<string>
```
304NotModified: why no hashset? Afaik it's more efficient with 3+ elements, and it's static

src/NLog/Internal/CallSiteInformation.cs
```diff
@@ -0,0 +1,146 @@
...
+        /// </summary>
+        public StackTrace StackTrace { get; private set; }
+
+        public MethodBase GetCallerStackFrameMethod(int skipFrames)
+        {
+            StackFrame frame = StackTrace?.GetFrame(UserStackFrameNumber + skipFrames);
+            return frame?.GetMethod();
+        }
+
+        public string GetCallerClassName(MethodBase method, bool includeNameSpace, bool cleanAsyncMoveNext, bool cleanAnonymousDelegates)
```
304NotModified: What do you think about a struct/class for the last 3 parameters?

(the same object for `GetCallerMemberName`

Only for readability of the calling site.

src/NLog/LayoutRenderers/AllEventPropertiesLayoutRenderer.cs
snakefoot: Well I don't need the HashSet-lookup, and the struct-enumerator on List is allocation-free.

src/NLog/Internal/CallSiteInformation.cs
snakefoot: Have no opinion. But not that big fan of struct-parameters.

304NotModified: OK,it's internal.so could be done in the future

PR #2387, opened 2017-11-13
JsonConverter - Do not include static properties by default

snakefoot:
Only include instance properties by default.


304NotModified:
👍 great improvement!

PR #2389, opened 2017-11-14
Fixed Sonar Lint code analysis warnings

snakefoot:
And cleaned up PropertyHelper.TryImplicitConversion (Turned ugly with NetStandard1.5)

### DIFF
--- a/src/NLog/NestedDiagnosticsContext.cs
+++ b/src/NLog/NestedDiagnosticsContext.cs
@@ -37,7 +37,7 @@ namespace NLog
     using System.Collections.Generic;
     using System.Linq;
 
-    using Internal;
+    using NLog.Internal;
 
     /// <summary>
     /// Nested Diagnostics Context - a thread-local structure that keeps a stack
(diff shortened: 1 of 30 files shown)


src/NLog/NestedDiagnosticsContext.cs
```diff
@@ -162,8 +162,8 @@ public static object[] GetAllObjects()
         /// </summary>
         private class StackPopper : IDisposable
         {
-            private Stack<object> _stack;
-            private int _previousCount;
+            private readonly Stack<object> _stack;
+            private readonly int _previousCount;
```
304NotModified: note that `readonly` could influence runtime performance (badly)

snakefoot: Only heard that it affect performance for fields in struct, where it will cause the entire struct to be copied on field access. Do you have some details about this other readonly pitfall?

304NotModified: this one, but that's indeed structs 

https://codeblog.jonskeet.uk/2014/07/16/micro-optimization-the-surprising-inefficiency-of-readonly-fields/

Issue #2147, opened 2017-06-03
Change defaults of NLogViewer target


304NotModified:
See https://github.com/NLog/NLog/issues/2145

Change defaults to
- `keepConnection="false" onConnectionOverflow="Block" maxConnections="100"`

stil:
While it works indeed with 100 connections, I think the limit of connections is imposed by `log4view` implementation. So other logging software could have the limit much lower, for example 50 connections or even higher, at 999 connections. Operating system can have even 2^16 = 64K open TCP connections at the moment.

I'd suggest to set default `maxConnections` limit much lower, for example at 16.

304NotModified:
Good one. But don't you think this will be a too big change in defaults then?

stil:
I don't think unlimited connections were good idea at any time. Also, are we changing defaults for both Network target and NLogViewer target or just NLogViewer?

304NotModified:
only NLogViewer target.

304NotModified:
so this is the proposal:
    
    keepConnection="false" onConnectionOverflow="Block" maxConnections="16"

I think this is suitable for the next minor release (4.5)

PR #2404, opened 2017-11-18
Updated NLog viewer target defaults

304NotModified:
moved to defaults which works better out-of-the-box

fixes https://github.com/NLog/NLog/issues/2147

### DIFF
--- a/src/NLog/Targets/NLogViewerTarget.cs
+++ b/src/NLog/Targets/NLogViewerTarget.cs
@@ -79,6 +79,8 @@ namespace NLog.Targets
         {
             Parameters = new List<NLogViewerParameterInfo>();
             Renderer.Parameters = Parameters;
+            OnConnectionOverflow = NetworkTargetConnectionsOverflowAction.Block;
+            MaxConnections = 16;
             NewLine = false;
         }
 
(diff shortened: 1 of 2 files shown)


snakefoot:
@304NotModified Updated documentation https://github.com/NLog/NLog/wiki/NLogViewer-target

snakefoot:
@304NotModified Are you not missing `keepConnection="false"` ? `OnConnectionOverflow` has no effect when `keepConnection="true"` (current default)

304NotModified:
Think you're right, altough the defaults are now better when disabling keepConnection

snakefoot:
Think a much better default for the NLogViewerTarget is to set the following:

`NewLine = true;`

The following config works without loosing any messages in Log4View, running at full speed:

`<target name="log4view" xsi:type="NLogViewer" address="tcp://127.0.0.1:878" NewLine="true" />`

Can see `keepConnection="false"` will cause lots of allocations, so it is a good idea not to enable it.

304NotModified:
That's also fine to me. Goal is to improve the default behavior of the target :)

snakefoot:
@304NotModified Looks like it is just the Log4View-application that is very sensitive. Other applications (Ex. log2console) does not require NewLine=true to work.

304NotModified:
I think you're right. So what should be done, revert this change? Imo unlimited max connections is a bit tricky.

snakefoot:
This change has no effect, besides adding a little confusion by adding settings that are not used.


Sent from my Samsung device

PR #2407, opened 2017-11-18
NLogTraceListener - set DisableFlush true by default

snakefoot:
NLog will automatically flush on AppDomain-shutdown, so going crazy and eating all threads when autoflush=true is not a good idea. See also #353

This gives a better out-of-box-experience, but also a breaking change, so should be deferred to NLog 4.5 or 5.0

Replaces  #2142


snakefoot:
Inspired by #2404 :)

304NotModified:
👍 good defaults are important

PR #2408, opened 2017-11-20
Feature 2049: Added DatabaseTargetWritten event to LogEventInfo w/ Tests

TimKielyVT:
I'm not crazy about putting the call the the event handler(s) within the transaction in DatabaseTarget.WriteEventToDatabase(). 
What do you think?
An alternative would be to create a temporary dictionary of output parameter values and use it to call the event handler(s) outside the TransactionScope and IDBCommand "using" blocks.

### DIFF
--- a/src/NLog/LogEventInfo.cs
+++ b/src/NLog/LogEventInfo.cs
@@ -151,6 +151,31 @@ namespace NLog
         }
 
         /// <summary>
+        /// The delegate for the DatabaseTargetWritten event.
+        /// </summary>
+        /// <param name="name"></param>
+        /// <param name="value"></param>
+        public delegate void DatabaseTargetWrittenHandler(string name, object value);
+
+        /// <summary>
+        /// The event called when a DatabaseTarget with an output parameter has been written to the database.
+        /// </summary>
+        public event DatabaseTargetWrittenHandler DatabaseTargetWritten;
+
+        /// <summary>
+        /// Used internally to call the client's event handler(s)
+        /// </summary>
+        /// <param name="name"></param>
+        /// <param name="value"></param>
+        internal void OnDatabaseTargetWritten(string name, object value)
+        {
+            if (DatabaseTargetWritten != null)
+            {
+                DatabaseTargetWritten(name, value);
+            }
+        }
+
+        /// <summary>
         /// Gets the unique identifier of log event which is automatically generated
         /// and monotonously increasing.
         /// </summary>
@@ -437,7 +462,13 @@ namespace NLog
         /// <returns>Instance of <see cref="LogEventInfo"/>.</returns>
         public static LogEventInfo Create(LogLevel logLevel, string loggerName, IFormatProvider formatProvider, object message)
         {
-            return new LogEventInfo(logLevel, loggerName, formatProvider, "{0}", new[] { message });
+            LogEventInfo retVal = new LogEventInfo(logLevel, loggerName, formatProvider, "{0}", new[] { message });
+            if (message is LogEventInfo)
+            {
+                LogEventInfo logEventInfo = message as LogEventInfo;
+                retVal.DatabaseTargetWritten = logEventInfo.DatabaseTargetWritten;
+            }
+            return retVal;
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


snakefoot:
@TimKielyVT Start with removing the code from LogEventInfo. The event-handler should not be there.

Then adjust DatabaseTarget so it is easy to override a method method to get the LogEventInfo just about to be written (with support for bulk-write).

Then it will be easy for you to inherit from the DatabaseTarget and override the method, and then do whatever crazy stuff you need when calling the base-method. Maybe perform a try-catch around the call of the base-method and notify when the call fails/succeeds.

TimKielyVT:
Hi Rolf,

Thanks for the feedback.

I'd appreciate any insights into the design of NLog that would help me
understand what features are a better fit for NLog.

Why do you say that the event doesn't belong on LogEventInfo?  I thought
that LogEventInfo properties are the way to get runtime data to target
parameters.  Is that correct?

All I'm trying to achieve is to get an output parameter back to the caller
without having to worry about the NLog targets when my code calls NLog,
e.g. "logger.Error(logEvent);".

Thanks again,
Tim

snakefoot:
@TimKielyVT Instead of adding the hook to LogEventInfo, then you should make the hook into the target that you want to hear from. Maybe creating a custom-target that inherits from databasetarget that exposes a global-event handler that provides the logeventinfo along with the exception.

Then the caller can just add hook into this global-event-handler and wait for the specific logeventinfo to be saved into the database.

304NotModified:
I think adding an event to the databasetarget is also fine to me, what do you think @snakefoot ?

snakefoot:
I prefer not to have event-handlers because they have funny dangling surprises. Instead I just think it should be easy to create a custom target based on databasetarget that the intercept the result of saving a bunch of LogEventInfos.

304NotModified:
> I prefer not to have event-handlers in the NLog-core, because they have funny dangling surprises. Instead I just think it should be easy to create a custom target based on databasetarget that the intercept the result of saving a bunch of LogEventInfos.

en what about writing back to the original log event info? (optional, default off)

snakefoot:
> en what about writing back to the original log event info? (optional, default off)

Usually the LogEventInfo should be seen as readonly, and should not be modified after being sent into the NLog-engine. It will have funny side-effects.

And because the author wants the return value from the database-sql-command returned then one cannot use wrapper-target (similar to the retrying-wrapper), as one wants to extract target-specific details about the result.

304NotModified:
> Usually the LogEventInfo should be seen as readonly, and should not be modified after being sent into the NLog-engine. It will have funny side-effects.

well modifying gives create opportunities, e.g. writing them to other targets afterwards

TimKielyVT:
If it is an event on the target, why would you want it to be a subclass of
DatabaseTarget?
I don't see an obvious place to persist the value of the output
parameter(DatabaseParameterInfo doesn't have a "Value" property).
So, I'd have to copy the implementation in DatabaseTarget just to add a
call to the event.

TimKielyVT:
What do you think of adding a 'Value' parameter to DatabaseParameterInfo?
No events are required and the client can retrieve the value of the output parameter after calls to a logging function.
For example:
```C#
DatabaseTarget dbTarget = (DatabaseTarget)LogManager.Configuration.FindTargetByName("DbTarget");
Logger logger = LogManager.GetCurrentClassLogger();
LogEventInfo logEvent = CreateExceptionDatabaseEvent(new Exception("More fun than a barrel of monkeys, ya'll."));
logger.Error(logEvent);
int errorId = Convert.ToInt32(dbTarget.Parameters.Where(p => p.Name == "@EventId").First().Value);
```

snakefoot:
> What do you think of adding a 'Value' parameter to DatabaseParameterInfo?

The more possible solutions you present, the more I think you are using NLog as the wrong hammer.

I believe in modifying the DatabaseTarget so it is easy to extract the return value of the stored-procedure for a CustomDatabaseTarget that inherits from DatabaseTarget.

Providing a highway for a logging-target to return user-defined values back to the original logger is not something that exist in NLog. Usually the point of logging is fire-and-forget, and not to meddle with target-specific-return-values.

PR #2409, opened 2017-11-20
Revert "Avoid struct copy on readonly field access"

snakefoot:
This partially reverts #2399

Misunderstood the issue. The problem is not having readonly fields, but having property-methods on structs (That really hurts when the struct is used as readonly-member-variable).


304NotModified:
OK thanks!

PR #2414, opened 2017-11-22
Revert breaking change

304NotModified:
While the code is now inconsistent with `Pop(IFormatProvider formatProvider)` and `NestedDiagnosticsContext.Pop()`, we should prevent breaking changes at all costs.

There are +500.000 downloads with the previous signature, so I won't like to take the risk on "method not found" errors.


304NotModified:
@snakefoot guess this is the last before RC1. :)

PR #2417, opened 2017-11-23
IsMono - Cache Type.GetType to avoid constant AppDomain.TypeResolve events

### LINKED ISSUE #2416: Appdomain.TypeResolve event being raised for "Mono.Runtime" while Logging everytime
Type: Bug

**NLog version**: (e.g. 4.4.12)

**Platform**: .Net 3.5 

 <nlog >
   <extensions>
     <add assembly ="MyAssembly"/>
   </extensions>
  
 <targets>
  <target name="TargFile" xsi:type="CFile" fileName="c:/logs/Test.txt" />  
 </targets>

   <rules>
     <logger name="*" minlevel="Info" writeTo="TargFile"/>
   </rules>    
 </nlog>
In case of a BUG:

- What is the current result?
LOGGING IS WORKING AS EXPECTED but:
I am getting a ResolveType Exception(I Subscribed to AppDomain.TypeResolve) whenever I try ".log"(or.info,.error etc). It seems the Exception occurs whenever Type.GetType(Mono.Runtime) is called.  

I expect  ResolveType event to not be called.

No new messages appeared in Internal log.

I believe the class "Nlog/Internal/PlatformDetector.cs" is responsible. The ResolveType Exception only has one property : ex.Name,this property says the Type Failing to resolve is "Mono.Runtime". 

 
- Are there any work arounds? By default no one subscribes to Appdomain.TypeResolve, and thus they will not notice the event, that might not exactly be a workaround as i need to listen to Typeresolve from other parts of my code.

Logging Works, but this event is being raised everytime.

Is there any way to prevent PlatformDetector to check for Mono.Runtime type everytime i try logging?My application needs to check for any typeresolve events as well as a lot of others and log them too, but my console and log file is filled with Typeresolve for "Mono.runtime" after every line.

snakefoot:
Resolves  #2416

### DIFF
--- a/src/NLog/Internal/PlatformDetector.cs
+++ b/src/NLog/Internal/PlatformDetector.cs
@@ -65,7 +65,8 @@ namespace NLog.Internal
         /// <summary>
         /// Gets a value indicating whether current runtime is Mono-based
         /// </summary>
-        public static bool IsMono => Type.GetType("Mono.Runtime") != null;
+        public static bool IsMono => _isMono ?? (_isMono = Type.GetType("Mono.Runtime") != null).Value;
+        private static bool? _isMono;
 
         /// <summary>
         /// Gets a value indicating whether current runtime supports use of mutex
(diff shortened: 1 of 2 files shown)


304NotModified:
smart one 👍

Issue #2345, opened 2017-10-12
Remove [StringFormatMethod] on loggers


304NotModified:
Remove the StringFormatMethodAttribute as this gives warnings for structured logging

snakefoot:
Maybe replace with a different attribute. Like MessageTemplateFormatMethodAttribute.

304NotModified:
see also https://github.com/Suchiman/SerilogAnalyzer/issues/15

304NotModified:
> Maybe replace with a different attribute. Like MessageTemplateFormatMethodAttribute. (That allows one use both positional-parameters and structured-parameters)
> 
> In case one wanted to make something like https://github.com/Suchiman/SerilogAnalyzer

I'm not sure how this works. Do we have to add attributes for this as you suggested? What should be in the attribute, or is it a "marker" attribute? thanks!

snakefoot:
> I'm not sure how this works. What should be in the attribute, or is it a "marker" attribute? 

Yes it works as marker attribute. Just like the StringFormatMethod-attribute does. Signalling to a code-analyzer that extra validation can be performed (If it recognizes the attribute).

PR #2419, opened 2017-11-25
Logger - Added missing MessageTemplateFormatMethodAttribute + removed obsolete internal method

snakefoot:
Follow up to #2345 and #2406

### DIFF
--- a/src/NLog/Logger-generated.tt
+++ b/src/NLog/Logger-generated.tt
@@ -189,7 +189,7 @@ namespace NLog
         {
             if (Is<#=level#>Enabled)
             {
-                WriteToTargets(LogLevel.<#=level#>, message, exception);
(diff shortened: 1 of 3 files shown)


src/NLog/Logger-generated.tt
```diff
@@ -189,7 +189,7 @@ namespace NLog
         {
             if (Is<#=level#>Enabled)
             {
-                WriteToTargets(LogLevel.<#=level#>, message, exception);
+                WriteToTargets(LogLevel.<#=level#>, exception, message, null);
```
304NotModified: why is this change needed?

snakefoot: Because it referenced an obsolete internal method, that I have also removed (Leftovers from the reordering of the Exception-parameter to the front).

304NotModified:
ok thx!

PR #2117, opened 2017-05-17
Added IIncludeContext, implemented missing properties, added includeNdlc, sync NDLC and NDC

304NotModified:

PR #2421, opened 2017-11-26
Changed IIncludeContext to internal interface until someone needs it

snakefoot:
Minor fix to #2117


304NotModified:
Good idea +1

PR #2422, opened 2017-11-26
Avoid unnecessary string-allocation, skip unnecessary lock

snakefoot:
Avoid unnecessary string-allocation when appending two StringBuilders.

Skip use of lock-statement when no message-template properties captured.

### DIFF
--- a/src/NLog/Internal/LogMessageTemplateFormatter.cs
+++ b/src/NLog/Internal/LogMessageTemplateFormatter.cs
@@ -76,25 +76,28 @@ namespace NLog.Internal
             {
                 return logEvent.Message;
             }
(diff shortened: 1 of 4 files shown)


src/NLog/Internal/LogMessageTemplateFormatter.cs
```diff
@@ -76,25 +76,28 @@ public string FormatMessage(LogEventInfo logEvent)
             {
                 return logEvent.Message;
             }
-            // Prevent multiple layouts on different targets to render the same properties
-            lock (logEvent)
+            using (var builder = _builderPool.Acquire())
```
304NotModified: this is an other change, isn't?

PR #2429, opened 2017-11-27
License should explicit state BSD 3 Clause

snakefoot:
Instead of trying to guess from the license text (Help Nuget analyzer to deduce license-type)


304NotModified:
You need to change all files then. Also, I think the original text is from the original BSD license template?

snakefoot:
@304NotModified Will be another day then. Just sad that the Nuget-package doesn't display the correct license up front.

304NotModified:
true,

I would be better if we could configure this in the nuget package (or that the used software on nuget.org was better)

PR #2431, opened 2017-11-28
NetCoreApp - Improve auto loading of NLog extension dlls

snakefoot:
NetCore-Console-Applications will usual be compiled to a dll, that is run by the dotnet-tool.

The NLog-package-reference will be loaded from an isolated path, and using this path to load NLog-extension-dlls will result in empty result.

Added extra fallback locations when the NLog-assembly-path returns empty file-list:
- Assembly.GetEntryAssembly() - Works for NetCore-Console-Application, but will be null for other cases (High probability this will be the private assembly path)
- AppDomain.BaseDirectory - Should also work for AspNetCore.

But it will not help with dlls coming from Nuget-packages. So maybe it is better to implement something using `DependencyContext` api to resolve the DLLs and their locations that are part of the application's dependency graph.

### DIFF
--- a/src/NLog/Config/ConfigurationItemFactory.cs
+++ b/src/NLog/Config/ConfigurationItemFactory.cs
@@ -366,35 +366,37 @@ namespace NLog.Config
             var nlogAssembly = typeof(ILogger).GetAssembly();
             var factory = new ConfigurationItemFactory(nlogAssembly);
             factory.RegisterExtendedItems();
-#if !SILVERLIGHT
 
+#if !SILVERLIGHT
             try
             {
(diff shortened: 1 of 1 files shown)


304NotModified:
> But it will not help with dlls coming from Nuget-packages.

So it work only for projects in your solution isn't?

snakefoot:
Yes it will only help in very few cases, and will change behaviour for existing users. Not entirely convinced myself this is a perfect PR :)



Sent from my Samsung device


304NotModified:
small step forward is step forward :)

I think it's nice, maybe a bit more difficult to explain when it works and when don't

PR #2437, opened 2017-12-01
Added "regex-matches" condition method for filtering

### LINKED ISSUE #2324: Add regex.is-match(...) between basic Condition Functions
**Type**: Feature request
Add `regex.is-match(input, pattern, options)` between **basic** Condition Functions like `contains(s1, s2)`, `starts-with(s1, s2)` and others.
It would easily call this method:
```cs
public static bool Regex.IsMatch(string input, string pattern, RegexOptions options)
```
 - Why do we need it?
Regex searching is used **very often** and it should be in my opinion between **basic** Condition Functions.
- Workaround
You can write new condition function and mark the class with the attribute [ConditionMethod].

MikeFH:
Closes #2324 

I went with the `regex-matches` name since it seemed everybody was OK with it in the issue discussion.

With .NET regexes, it is possible to specify options inline as explained [here](https://docs.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-options#specifying-the-options) so the `options` parameter shouldn't be necessary. I find it more usable if it exists but it creates some kind of redundancy and adds some code to parse the options string. Should we leave it ? If yes, should we make it optional ?
At the moment, [all the options you can use inline](https://docs.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-options) are supported by the `options` parameter.

### DIFF
--- a/src/NLog/Conditions/ConditionMethods.cs
+++ b/src/NLog/Conditions/ConditionMethods.cs
@@ -173,32 +173,33 @@ namespace NLog.Conditions
         /// </summary>
         /// <param name="input">The string to search for a match.</param>
         /// <param name="pattern">The regular expression pattern to match.</param>
-        /// <param name="options">A string consisting of the desired options for the test. Available options are : 'i' for ignore case, 'm' for multi-line, 's' for single line and 'x' for ignore pattern whitespace.</param>
+        /// <param name="options">A string consisting of the desired options for the test. The possible values are those of the <see cref="RegexOptions"/> separated by commas.</param>
         /// <returns>true if the regular expression finds a match; otherwise, false.</returns>
         [ConditionMethod("regex-matches")]
-        public static bool RegexMatches(string input, string pattern, string options)
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1026:DefaultParametersShouldNotBeUsed", Justification = "Not called directly, only ever Invoked.")]
+#if SILVERLIGHT
+        public static bool RegexMatches(string input, string pattern, [Optional] string options)
+#else
+        public static bool RegexMatches(string input, string pattern, [Optional, DefaultParameterValue("")] string options)
+#endif
         {
-            RegexOptions regexOpts = RegexOptions.None;
-
-            if (options.IndexOf('i') != -1)
-            {
-                regexOpts |= RegexOptions.IgnoreCase;
-            }
-            if (options.IndexOf('m') != -1)
-            {
-                regexOpts |= RegexOptions.Multiline;
-            }
-            if (options.IndexOf('s') != -1)
-            {
-                regexOpts |= RegexOptions.Singleline;
-            }
-            if (options.IndexOf('x') != -1)
-            {
-                regexOpts |= RegexOptions.IgnorePatternWhitespace;
-            }
-
+            RegexOptions regexOpts = ParseRegexOptions(options);
             return Regex.IsMatch(input, pattern, regexOpts);
         }
 
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="options"></param>
+        /// <returns></returns>
+        private static RegexOptions ParseRegexOptions(string options)
+        {
+            if (string.IsNullOrEmpty(options))
+            {
+                return RegexOptions.None;
+            }
+
+            return (RegexOptions)Enum.Parse(typeof(RegexOptions), options, true);
+        }
     }
 }
--- a/src/NLog/Conditions/ConditionMethods.cs
+++ b/src/NLog/Conditions/ConditionMethods.cs
@@ -178,12 +178,7 @@ namespace NLog.Conditions
         [ConditionMethod("regex-matches")]
         public static bool RegexMatches(string input, string pattern, string options)
         {
-            RegexOptions regexOpts =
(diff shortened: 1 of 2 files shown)


src/NLog/Conditions/ConditionMethods.cs
```diff
@@ -166,5 +167,43 @@ public static int Length(string text)
...
+        /// <param name="options">A string consisting of the desired options for the test. Available options are : 'i' for ignore case, 'm' for multi-line, 's' for single line and 'x' for ignore pattern whitespace.</param>
+        /// <returns>true if the regular expression finds a match; otherwise, false.</returns>
+        [ConditionMethod("regex-matches")]
+        public static bool RegexMatches(string input, string pattern, string options)
+        {
+            RegexOptions regexOpts =
+#if SILVERLIGHT
+                RegexOptions.None;
+#else
+                RegexOptions.Compiled;
```
304NotModified: please don't compile. never (well almost never). It could lead to memory issues, and the cached version is in 99% of the cases a better idea. 

The static methodes on Regex are already cached

Some articles about it: http://www.allben.net/post/2009/08/06/Performance-Compiled-vs-Interpreted-Regular-Expressions, https://blog.codinghorror.com/to-compile-or-not-to-compile/

src/NLog/Conditions/ConditionMethods.cs
```diff
@@ -166,5 +167,43 @@ public static int Length(string text)
...
+#endif
+
+            if (options.IndexOf('i') != -1)
+            {
+                regexOpts |= RegexOptions.IgnoreCase;
```
304NotModified: It would be nice to create a helper for this. (for at least the indexof, but maybe also the `|=`)

src/NLog/Conditions/ConditionMethods.cs
```diff
@@ -166,5 +167,43 @@ public static int Length(string text)
...
+        public static bool RegexMatches(string input, string pattern, string options)
+        {
+            RegexOptions regexOpts =
+#if SILVERLIGHT
+                RegexOptions.None;
+#else
+                RegexOptions.Compiled;
+#endif
+
+            if (options.IndexOf('i') != -1)
```
304NotModified: I think its nice to create a method, `RegexOptions  CreateOptions(string)`

304NotModified:
> With .NET regexes, it is possible to specify options inline as explained here so the options parameter shouldn't be necessary.

was thinking about this. Maybe the options parameter sense in the following cases:

- we could use longer names (the full names)
- we configure from the API


AFAIK it already works if you create an property `RegexOptions Options` on the class - not sure, it works for targets and layout renderers

src/NLog/Conditions/ConditionMethods.cs
MikeFH: Since the regex specified will probably be used a lot and the doc says that you'll gain run-time performance by compiling, it made sense to me. I'll make it "not compiled" because it's only intuition.

304NotModified: another option is to make it optional (off by default)

MikeFH:
OK so if I get this correctly, we could have an optional `options` parameter using the full names of the `RegexOptions` enumeration values separated by spaces and thus support all the possibilities. For example
`regex-matches('foo', '^foo$', 'ignorecase')` or
`regex-matches('foo', '^foo$', 'ignorecase multiline')`

This also allows to support compiled regexes without imposing it : 
`regex-matches('foo', '^foo$', 'ignorecase compiled')`

I'm sorry but I didn't understand what you meant by
> we configure from the API
Aren't the expressions always specified as strings and then parsed as ConditionExpression ?

304NotModified:
> OK so if I get this correctly, we could have an optional options parameter using the full names of the RegexOptions enumeration values separated by spaces and thus support all the possibilities. For example
> regex-matches('foo', '^foo$', 'ignorecase') or
> regex-matches('foo', '^foo$', 'ignorecase multiline')
> 
> This also allows to support compiled regexes without imposing it :
> regex-matches('foo', '^foo$', 'ignorecase compiled')
> 
> I'm sorry but I didn't understand what you meant by

yes that's correct. (except is should be a comma instead of space) - see https://github.com/NLog/NLog/blob/74e6e26fb50bc1a1396ed8dbeb8e1ec9b31d88e3/tests/NLog.UnitTests/Config/TargetConfigurationTests.cs#L522

But i'm not sure a flag enum property works `ConditionMethod`, it works for `Layout` and `LayoutRenderers`

MikeFH:
I modified the initial version and tried to take into account all the remarks.

304NotModified:
\0/ 

Could you please add some docs? I think here is the best place, please add "introduced in NLog 4.5"

https://github.com/NLog/NLog/wiki/When-filter#condition-functions

Thanks for the great work!

MikeFH:
@304NotModified Thank you for your help !

To documentation is updated.

304NotModified:
Thanks! Nice feature!

MarekPokorny:
NLog became much stronger.
THANK YOU VERY MUCH!!! Good work @MikeFH, thanks to @304NotModified.

304NotModified:
:) 

this feature is online: https://www.nuget.org/packages/NLog/4.5.0-rc02

PR #2447, opened 2017-12-03
CallSiteInformation - Prepare for fast classname lookup from filename and linenumber

snakefoot:
Have an idea for how to optimize the callsite-layout-renderer, so it doesn't have to capture stacktrace for every logevent (only for the first logevent for a log-location).

One could use the caller-member-attributes to capture Member-Function, SourceFile-Name, SourceFile-LineNumber. And cache the ClassName-Lookup performed with StackTrace using the details from caller-member-attributes as cache-lookup-key (Maybe using MruCache).

To implement this correctly, then we need something like StackTraceUsage, but where it is WithCallSiteOnly, which is less than WithoutSource. Sadly enough any change to the StackTraceUsage will be a breaking change.

So instead one probably have to invent a new interface called IStackTraceUsageV2, and if a class implements both IStackTraceUsageV2 and IStackTraceUsage, then it will use IStackTraceUsageV2.

But all this interface-stuff will be a different PR. I just want to prepare for being able to provide the ClassName as CallSite-Information (Right now not a breaking change, as everything is still new in NLog 4.5).

### DIFF
--- a/src/NLog/Fluent/LogBuilder.cs
+++ b/src/NLog/Fluent/LogBuilder.cs
@@ -306,7 +306,7 @@ namespace NLog.Fluent
             if (callerLineNumber != 0)
                 Property("CallerLineNumber", callerLineNumber);
 
-            _logEvent.SetCallerInfo(callerMemberName, callerFilePath, callerLineNumber);
+            _logEvent.SetCallerInfo(null, callerMemberName, callerFilePath, callerLineNumber);
 
             _logger.Log(_logEvent);
         }
--- a/src/NLog/Internal/CallSiteInformation.cs
+++ b/src/NLog/Internal/CallSiteInformation.cs
@@ -55,11 +55,13 @@ namespace NLog.Internal
         /// <summary>
         /// Sets the details retrieved from the Caller Information Attributes
         /// </summary>
+        /// <param name="callerClassName"></param>
         /// <param name="callerMemberName"></param>
         /// <param name="callerFilePath"></param>
         /// <param name="callerLineNumber"></param>
-        public void SetCallerInfo(string callerMemberName, string callerFilePath, int callerLineNumber)
+        public void SetCallerInfo(string callerClassName, string callerMemberName, string callerFilePath, int callerLineNumber)
         {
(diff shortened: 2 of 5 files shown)


snakefoot:
@304NotModified If possible, please include this for NLog 4.5 RC2 :)

304NotModified:
> Have an idea for how to optimize the callsite-layout-renderer, so it doesn't have to capture stacktrace for every logevent (only for the first logevent for a log-location).
> 
> One could use the caller-member-attributes to capture Member-Function, SourceFile-Name, SourceFile-LineNumber. And cache the ClassName-Lookup performed with StackTrace using the details from caller-member-attributes as cache-lookup-key (Maybe using MruCache).

I think this could work :) MruCache sounds like a good idea. Would be very nice to improve performance on this one!

> To implement this correctly, then we need something like StackTraceUsage, but where it is WithCallSiteOnly, which is less than WithoutSource. Sadly enough any change to the StackTraceUsage will be a breaking change.

How is this a breaking changes? We could add a member to the enum StackTraceUsage? AFAIK it's not mandatory to keep the enum (backend) values "ordered"? 

> (Right now not a breaking change, as everything is still new in NLog 4.5).

So this is a must for NLog 4.5? I'm holding RC2 for this now.

snakefoot:
Think enum values are compiled in as integer values. Only allowed to add at the end. Max is locked to 2 and the new value need to be added just after none.


Sent from my Samsung device


304NotModified:
> Think enum values are compiled in as integer values. Only allowed to add at the end. Max is locked to 2 and the new value need to be added just after none.

It's not required to have `Max` as the highest value? Yes, its nice, but no breaking changes are better. Are we using lower/higher comparison on the enum?

304NotModified:
> Are we using lower/higher comparison on the enum?

Ow I see StackTraceUsageUtils.Max - although  it's internal

snakefoot:
Anyways those enum details are for another PR :). Just wanted to markup the playing field



Sent from my Samsung device


304NotModified:
true :)

PR #2452, opened 2017-12-05
ThreadLocalStorageHelper - NetStandard only allocate when needed

snakefoot:
Minor allocation optimization.


304NotModified:
thanks!

PR #2456, opened 2017-12-07
FileTarget: Support byte order mark (optional)

### LINKED ISSUE #26: UTF8 should write BOM (optional) (FileTarget)

KYegres:
I added ability to write byte order mark. fixes #26

### DIFF
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -479,6 +479,12 @@ namespace NLog.Targets
         public bool NetworkWrites { get; set; }
 
         /// <summary>
+        /// Gets or sets a value indicating whether to write BOM (byte order mark) in created files
+        /// </summary>
+        [DefaultValue(false)]
+        public bool WriteBom { get; set; }
+
+        /// <summary>
         /// Gets or sets the number of times the write is appended on the file before NLog
         /// discards the log message.
         /// </summary>
@@ -1941,7 +1947,7 @@ namespace NLog.Targets
             {
                 if (initializedNewFile)
                 {
-                    WriteHeader(appender);
+                    WriteHeaderAndBom(appender);
                 }
 
                 appender.Write(bytes.Array, bytes.Offset, bytes.Count);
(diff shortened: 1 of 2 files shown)


304NotModified:
Thanks! Will review soon!

304NotModified:
nice! Could you please add some docs to https://github.com/NLog/NLog/wiki/File-Target? Please add "introduced in NLog 4.5" - thanks!

PR #2458, opened 2017-12-10
Document and (minor) refactor on MruCache class

ie-zero:
Internal MruCacheItem class was refactored to make the property modifiers explicit.


304NotModified:
Thanks! Nice improvement!

Issue #2465, opened 2017-12-15
AsyncTargetWrapper loses MDLC context when puts event to parallel thread


datsis:
**Type**:
- Bug

**NLog version**: 
- 4.5.0 rc2

**Platform**: 
Checked only on .Net 4.5 

**Current NLog config** (xml or C#, if relevant)
```xml
<nlog>
  <targets>
    <default-wrapper xsi:type="AsyncWrapper" overflowAction="Block" />
    <target name="aiTarget" type="ApplicationInsightsTarget"/>
  </targets>
  <rules>
    <logger name="System.*"    maxlevel="Info" final="true" />
    <logger name="Microsoft.*" maxlevel="Info" final="true" />
    <logger name="*" minlevel="Trace" writeTo="aiTarget" />
  </rules>
</nlog>

```

In case of a BUG:

Context:
There is disposable scope which adds data into MDLC when it's created and removes these data from MDLC when disposing.  Let's name it "scoped data".
LogEvent "X" is generated by Logger.Info(...) which is called inside of scope.

- What is the current result?
Output of aiTarget does not contain "scoped data" for event "X".
- What is the expected result?
Output of aiTarget should contain "scoped data" for event "X".
- Did you checked the [Internal log](https://github.com/NLog/NLog/wiki/Internal-Logging)?
No
- Please post full exception details (message, stacktrace, inner exceptions)
There were no any exceptions
- Are there any work arrounds? yes/no
Yes, I had to create class derived from AsyncWrapperTarget which takes "scoped data" and adds it to logEvent.Properies collection before LogEvent is passed to parallel thread.
- Is there a version in which it did worked? 
Don't know.
- Can you help us by writing an unit test?
Probably.

snakefoot:
Remember you need to access the MDLC-properties using a layout-renderer ex: `${mdlc:item=PropertyName}` -

https://github.com/NLog/NLog/wiki/MDLC-Layout-Renderer

Or by creating a custom layout renderer, and extract the value from the MappedDiagnosticsLogicalContext:

https://github.com/NLog/NLog/wiki/How-to-write-a-custom-layout-renderer

snakefoot:
Also if you are using Microsoft ApplicationInsightsTarget:

https://github.com/Microsoft/ApplicationInsights-dotnet-logging/tree/develop/src/NLogTarget

Then you will notice they are not supporting MDLC-properties. You will have to implement your own.

[...]

PR #2467, opened 2017-12-16
TargetWithContext - Easier to capture snapshot of MDLC and NDLC context

### LINKED ISSUE #254: SplitGroup + AsyncWrapper = trouble for thread affinitive or relying on CallContext custom layout renderers
AsyncWrapper breaks CallContext, simply because it transitions to a dedicated thread using a request queue. Totally legitimate, but the CallContext is lost.

Recognizing this, the `AsyncTargetWrapper` class lets the log event to precalculate all the relevant layouts before enqueuing the log event. The precalculated layouts are stored in the layoutCache of the respective log event object. Perfect.

When the final message is rendered (on the dedicated thread) the layout cache is consulted and if the given layout pattern was already precalculated, then the cached value is returned.

The only problem is that the `SimpleLayout` class does not override the `GetHashCode` function. Nor was the dictionary created with a custom equality comparer. 

Thus the caching is broken and it manifests very clearly when there are custom layout renderers relying on the CallContext feature. The log event cache contains the correct layout text, because it was precalculated in the correct CallContext, but then the cache lookup fails and a wrong layout text is rendered and output.

### LINKED ISSUE #2466: Logger could capture MDC and MDLC properties
I can see that many target writers find it very hard to implement support for MDC and MDLC, especially with async-support.

To support async-mode correctly, then the custom target needs help from LayoutRenderers to capture the async-data and convert it into one or more string-objects.

If the custom target wants to generate json, then it can luckily use the JsonLayout that can help with all this, but if this is not the case, then it is hard work.

One could consider extending the LoggingRule (or the Logger-object) with an option that specifies that it should automatically convert MDC and MDLC-properties into standard LogEventInfo.Properties.

Depending on the level of trusted behavior, then one need to be careful with not capturing any random objects, as they might change state until reaching the actual targe-write. This would give some headache for those involved.

Types that can be safely converted from MDC / MDLC into LogEventInfo-Properties:
- Boxed ValueTypes like int, bool, decimal, datetime, etc.
- String-objects.
- Structs that only contains properties of the above type (doing property reflection might get expensive)
- IEnumerable-containers will require deep-clone and will probably give too much overhead.

I'm not sure if these restrictions will make such a feature into another pitfall for the users of MDLC and MDC.

snakefoot:
Attempt to resolve #2466 and #2465

Guess one could also extend it to enable capture of GDC and Callsite.

Maybe implement a helper method than returns all context properties (IncludeAllProperties, IncludeMdlc, IncludeMdc, IncludeGdc).

Suddenly also become a bugfix for correct precalculation of async data for custom NLog-Layouts. Before it would be impossible for custom NLog-Layouts to correctly capture the async-state. Resolves #254

And a bugfix that allows custom NLog-Layouts to work correctly, when they override Layout.Precalculate and are used with Target.OptimizeBufferReuse = true.

### DIFF
--- a/src/NLog/Internal/ILogMessageFormatter.cs
+++ b/src/NLog/Internal/ILogMessageFormatter.cs
@@ -51,7 +51,7 @@ namespace NLog.Internal
         /// Has the logevent properties?
         /// </summary>
         /// <param name="logEvent">LogEvent with message to be formatted</param>
-        /// <returns>formatted message</returns>
+        /// <returns>False when logevent has no properties to be extracted</returns>
         bool HasProperties(LogEventInfo logEvent);
 
         /// <summary>
--- a/src/NLog/Targets/TargetWithContext.cs
+++ b/src/NLog/Targets/TargetWithContext.cs
@@ -107,8 +107,8 @@ namespace NLog.Targets
         /// Gets the array of custom attributes to be passed into the logevent context
         /// </summary>
         /// <docgen category='Layout Options' order='10' />
-        [ArrayParameter(typeof(TargetWithContextProperty), "contextproperty")]
-        public virtual IList<TargetWithContextProperty> ContextProperties { get; }
+        [ArrayParameter(typeof(TargetPropertyWithContext), "contextproperty")]
+        public virtual IList<TargetPropertyWithContext> ContextProperties { get; }
 
         /// <summary>
         /// Constructor
--- a/src/NLog/Targets/TargetWithContext.cs
+++ b/src/NLog/Targets/TargetWithContext.cs
@@ -1,0 +1,797 @@
+﻿// 
+// Copyright (c) 2004-2017 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without 
+// modification, are permitted provided that the following conditions 
+// are met:
+// 
+// * Redistributions of source code must retain the above copyright notice, 
+//   this list of conditions and the following disclaimer. 
+// 
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution. 
+// 
+// * Neither the name of Jaroslaw Kowalski nor the names of its 
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission. 
+// 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE 
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE 
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE 
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR 
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS 
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN 
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) 
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF 
+// THE POSSIBILITY OF SUCH DAMAGE.
+// 
+
+namespace NLog.Targets
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Globalization;
+    using System.Text;
+    using NLog.Config;
+    using NLog.Internal;
+    using NLog.Layouts;
+
+    /// <summary>
+    /// Represents target that supports context capture using MDLC, MDC, NDLC and NDC
+    /// </summary>
+    public abstract class TargetWithContext : TargetWithLayout, IIncludeContext
+    {
+        /// <inheritdoc/>
+        /// <docgen category='Layout Options' order='1' />
+        public sealed override Layout Layout
+        {
+            get { return _contextLayout; }
+            set
+            {
+                if (_contextLayout != null)
+                    _contextLayout.TargetLayout = value;
+                else
+                    _contextLayout = new TargetWithContextLayout(this, value);
+            }
+        }
+        private TargetWithContextLayout _contextLayout;
+
+        /// <inheritdoc/>
+        bool IIncludeContext.IncludeAllProperties { get => IncludeEventProperties; set => IncludeEventProperties = value; }
+
+        /// <docgen category='Layout Options' order='10' />
+        public bool IncludeEventProperties { get => _contextLayout.IncludeAllProperties; set => _contextLayout.IncludeAllProperties = value; }
+
+        /// <inheritdoc/>
+        /// <docgen category='Layout Options' order='10' />
+        public bool IncludeMdc { get => _contextLayout.IncludeMdc; set => _contextLayout.IncludeMdc = value; }
+
+        /// <inheritdoc/>
+        /// <docgen category='Layout Options' order='10' />
+        public bool IncludeNdc { get => _contextLayout.IncludeNdc; set => _contextLayout.IncludeNdc = value; }
+
+#if !SILVERLIGHT
+        /// <inheritdoc/>
+        /// <docgen category='Layout Options' order='10' />
+        public bool IncludeMdlc { get => _contextLayout.IncludeMdlc; set => _contextLayout.IncludeMdlc = value; }
+
+        /// <inheritdoc/>
+        /// <docgen category='Layout Options' order='10' />
+        public bool IncludeNdlc { get => _contextLayout.IncludeNdlc; set => _contextLayout.IncludeNdlc = value; }
+#endif
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to include contents of the <see cref="GlobalDiagnosticsContext"/> dictionary
+        /// </summary>
+        /// <docgen category='Layout Options' order='10' />
+        public bool IncludeGdc { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to include call site (class and method name) in the <see cref="LogEventInfo" />
+        /// </summary>
+        /// <docgen category='Layout Options' order='10' />
+        public bool IncludeCallSite { get => _contextLayout.IncludeCallSite; set => _contextLayout.IncludeCallSite = value; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to include source info (file name and line number) in the <see cref="LogEventInfo" />
+        /// </summary>
+        /// <docgen category='Layout Options' order='10' />
+        public bool IncludeCallSiteStackTrace { get => _contextLayout.IncludeCallSiteStackTrace; set => _contextLayout.IncludeCallSiteStackTrace = value; }
+
+        /// <summary>
+        /// Gets the array of custom attributes to be passed into the logevent context
+        /// </summary>
+        /// <docgen category='Layout Options' order='10' />
+        [ArrayParameter(typeof(TargetWithContextProperty), "contextproperty")]
+        public virtual IList<TargetWithContextProperty> ContextProperties { get; }
+
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        protected TargetWithContext()
+        {
+            _contextLayout = _contextLayout ?? new TargetWithContextLayout(this, base.Layout);
+            OptimizeBufferReuse = true;
+        }
+
+        /// <summary>
+        /// Check if logevent has properties (or context properties)
+        /// </summary>
+        /// <param name="logEvent"></param>
+        /// <returns>True if properties should be included</returns>
+        protected bool ShouldIncludeProperties(LogEventInfo logEvent)
+        {
+            return IncludeGdc
+            || IncludeMdc
+#if !SILVERLIGHT
+            || IncludeMdlc
+#endif
+            || (IncludeEventProperties && (logEvent?.HasProperties ?? false));
+        }
+
+        /// <summary>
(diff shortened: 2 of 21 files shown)


snakefoot:
An alternative approach would be to implement a different async-target-wrapper using ThreadPool.QueueUserWorkItem (or Tasks). Then the async-state will automatically flow with the LogEventInfo (This will not help MDC + NDC but maybe they are also on their way out).

But such async-wrapper will not be able to perform batch-handling, so not that super interesting. And it would also require a special async-buffering-target-wrapper.

snakefoot:
Another thing that could be improved in NLog is the ability to handle batch-writing. Many are able to override the `Write(IList<AsyncLogEventInfo>)`, but then fail to correctly handle:

- Correctly notify if the write-operation succceded or failed by always calling AsyncLogEventInfo.Continuation
- Ensure Max Batch Size so only sending ex. 10 logevents at a time (instead of 100 or 10000 depending on the batch-wrapper).

snakefoot:
@304NotModified Implementing an easier approach for handling batching of log-events, became to messy. But I like this direction for making it easier for custom targets to make use of MDLC and NDLC without forcing them to use JsonLayout.

304NotModified:
> But I like this direction for making it easier for custom targets to make use of MDLC and NDLC without forcing them to use JsonLayout.

Indeed, that would be a nice improvement! 

Will try to review the PR this weekend.

snakefoot:
Maybe consider to change the default value of the parameter `IncludeMdlc = true`.

src/NLog/Internal/ILogMessageFormatter.cs
```diff
@@ -51,7 +51,7 @@ internal interface ILogMessageFormatter
         /// Has the logevent properties?
         /// </summary>
         /// <param name="logEvent">LogEvent with message to be formatted</param>
-        /// <returns>formatted message</returns>
+        /// <returns>False when logevent has no properties to be extracted</returns>
```
304NotModified: good catch :)

src/NLog/LayoutRenderers/Wrappers/Rot13LayoutRendererWrapper.cs
```diff
@@ -43,7 +43,6 @@ namespace NLog.LayoutRenderers.Wrappers
     /// See <a href="http://en.wikipedia.org/wiki/ROT13">http://en.wikipedia.org/wiki/ROT13</a>.
     /// </remarks>
     [LayoutRenderer("rot13")]
-    [AppDomainFixedOutput]
```
304NotModified: why is this not needed anymore? I don't see it.

src/NLog/Targets/TargetWithContext.cs
```diff
@@ -0,0 +1,747 @@
...
+                    contextLayout.IncludeNdc = true;
+#if !SILVERLIGHT
+                if (IncludeMdlc && !contextLayout.IncludeMdlc)
+                    contextLayout.IncludeMdlc = true;
+                if (IncludeNdlc && !contextLayout.IncludeNdlc)
+                    contextLayout.IncludeNdlc = true;
+#endif
+                _contextLayout = true;
+            }
+            else if (Layout is JsonLayout jsonLayout)
```
304NotModified: I'm not sure if this is good for the maintainability. Could we use/introduce an interface and handle IIncludeContext / JsonLayout  the same? (same for setters). What do you think? 

Maybe we could also check the type of Layout in init?

src/NLog/LayoutRenderers/Wrappers/Rot13LayoutRendererWrapper.cs
snakefoot: Thought the attribute meant that the output of the layoutrenderer would be completely static in the life time of the AppDomain (Ex like for newline-layoutrenderer). Cannot see how the same goes for Rot13. Anyway not important for me, so I can re-introduce it, if you want.

src/NLog/Targets/TargetWithContext.cs
snakefoot: I agree, and have now changed JsonLayout so it implements IIncludeContext.

snakefoot:
@304NotModified Not sure if property-name IncludeSourceInfo is the best (Stolen from Log4JXmlEventLayoutRenderer)

Maybe `IncludeCallSiteWithSource` is easier to understand?

I guess when the CallerMember-attributes becomes part of the official logger interface, then we also need `IncludeCallSiteWithStackTrace`.

304NotModified:
Oops wrong button ;)

IncludeCallSiteWithSource is indeed much better!

src/NLog/LayoutRenderers/Wrappers/Rot13LayoutRendererWrapper.cs
304NotModified: I was thinking it was same output when same input, so independent of other renderers. Need to check that.

snakefoot:
> IncludeCallSiteWithSource is indeed much better!

Done

src/NLog/LayoutRenderers/Wrappers/Rot13LayoutRendererWrapper.cs
304NotModified: Checked this, I think ` [AppDomainFixedOutput]` is correct here.  From the code:

a layout(renderer) could be converted to a literal when: 
- The layout and inner layouts are SimpleLayouts, or `[AppDomainFixedOutput]`

So if rot13 has a `SimpleLayout`, it will always have the same output.

snakefoot:
@304NotModified Resolve the merge conflicts, and reverted the rot13 change.

src/NLog/LayoutRenderers/Wrappers/Rot13LayoutRendererWrapper.cs
snakefoot: Yes you are right. Actually means that all wrapper layouts should have AppDomainFixedOutput

src/NLog/Targets/TargetWithContext.cs
```diff
@@ -0,0 +1,802 @@
...
+        /// <summary>Internal Layout that allows capture of NDC context</summary>
+        public Layout InternalLayoutNdc { get; }
+#if !SILVERLIGHT
+        /// <summary>Internal Layout that allows capture of MDLC context</summary>
+        public Layout InternalLayoutMdlc { get; }
+        /// <summary>Internal Layout that allows capture of NDLC context</summary>
+        public Layout InternalLayoutNdlc { get; }
+#endif
+        /// <summary>Internal Layout that allows capture of callsite context</summary>
+        public Layout InternalLayoutCallSite { get; }
```
304NotModified: I think this is a "trick" for the objectScanner, is that correct?

I'm not sure if exposing those to public (guess it's needed now), is a good idea. I found it confusing and when I subclass this class, it won't understand what to do with those properties.

304NotModified:
I will release 4.5 RC 4 now,

this could be 4.5 RC5, maybe 4.6 or 5.0

src/NLog/Targets/TargetWithContext.cs
304NotModified: if so, then my proposal

- make these layouts private
- make a new interface: 

```c#
internal interface IInternalLayouts{ //or maybe public
Layout[] getLayouts(); //or list/or ienumerable
}
```

- implement this interface explicitly (so members are hidden).
- check first for `IInternalLayouts`, combine them with the other found Layouts

src/NLog/Targets/TargetWithContext.cs
```diff
@@ -0,0 +1,802 @@
...
+
+        /// <inheritdoc/>
+        /// <docgen category='General Options' order='10' />
+        bool IncludeEventProperties
+        {
+            get { return _includeEventProperties && !_contextLayout; }
+            set
+            {
+                _includeEventProperties = value;
+                if (Layout is IIncludeContext contextLayout)
```
304NotModified: I think it would be nice to always set the contextLayout, but it should be a fake, then create a "fake" impl.

So all the properties just do   `contextLayout.IncludeAllProperties = value;`

src/NLog/Targets/TargetWithContext.cs
snakefoot: I have found a better trick, that I'm working on now. I will create nested Layouts, where the TopLevel becomes the Target.Layout (And will be scanned). The TopLevel will be similar to CompoundLayout.

Then override the Target.Layout-getter/setter so it will just be a single sub-Layout that lives under the TopLevel-Layout (along with all the context layouts).

304NotModified: Smart one!

snakefoot:
@304NotModified Now changed into nested Layouts, so there are no ugly public properties

src/NLog/Targets/TargetWithContext.cs
```diff
@@ -0,0 +1,802 @@
...
+                    contextLayout.IncludeAllProperties = value;
+            }
+        }
+        private bool _includeEventProperties;
+
+        /// <inheritdoc/>
+        /// <docgen category='TargetWithContext Options' order='10' />
+        public bool IncludeGdc
```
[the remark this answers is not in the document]
snakefoot: The idea is also to allow the end-user to control what properties to include automatically (using Xml)

snakefoot:
@304NotModified Now removed the automatic forwarding of properties to the inner-layout, think it will be more confusing than helping.

src/NLog/Targets/TargetWithContextProperty.cs
```diff
@@ -0,0 +1,81 @@
...
+        /// Gets or sets the layout that will be rendered as the attribute's value.
+        /// </summary>
+        /// <docgen category='Property Options' order='10' />
+        [RequiredParameter]
+        public Layout Layout { get; set; }
+
+        /// <summary>
+        /// Gets or sets when an empty value should cause the property to be excluded
+        /// </summary>
+        public bool ExcludeEmptyValue { get; set; }
```
304NotModified: proposal: change to IncludeEmptyValue (and default = true)

src/NLog/Targets/TargetWithContext.cs
```diff
@@ -0,0 +1,797 @@
...
+
+            return combinedPropties;
+        }
+
+        /// <summary>
+        /// Creates combined dictionary of all configured properties for logEvent
+        /// </summary>
+        /// <param name="logEvent"></param>
+        /// <returns>Dictionary with all collected properties for logEvent</returns>
+        protected IDictionary<string, object> CreateCombinedProperties(LogEventInfo logEvent)
```
304NotModified: proposal rename to GetCombinedProperties or maybe even GetAllProperties.

src/NLog/Targets/TargetWithContext.cs
```diff
@@ -0,0 +1,797 @@
...
+        /// </summary>
+        /// <docgen category='Layout Options' order='10' />
+        public bool IncludeCallSiteStackTrace { get => _contextLayout.IncludeCallSiteStackTrace; set => _contextLayout.IncludeCallSiteStackTrace = value; }
+
+        /// <summary>
+        /// Gets the array of custom attributes to be passed into the logevent context
+        /// </summary>
+        /// <docgen category='Layout Options' order='10' />
+        [ArrayParameter(typeof(TargetWithContextProperty), "contextproperty")]
+        virtual public IList<TargetWithContextProperty> ContextProperties { get; }
```
304NotModified: isn't strange that we can't set this property ever - and is never set?  Do we mean `abstract` instead of `virtual` here?

src/NLog/Targets/TargetWithContext.cs
```diff
@@ -0,0 +1,797 @@
...
+#endif
+            || (IncludeEventProperties && (logEvent?.HasProperties ?? false));
+        }
+
+        /// <summary>
+        /// Checks if any context properties, and if any returns them as a single dictionary
+        /// </summary>
+        /// <param name="logEvent"></param>
+        /// <returns>Dictionary with any context properties for the logEvent (Null if none found)</returns>
+        protected IDictionary<string, object> LookupContextProperties(LogEventInfo logEvent)
```
304NotModified: proposal: rename to GetContextProperties

src/NLog/Targets/TargetWithContext.cs
```diff
@@ -0,0 +1,797 @@
...
+            _contextLayout = _contextLayout ?? new TargetWithContextLayout(this, base.Layout);
+            OptimizeBufferReuse = true;
+        }
+
+        /// <summary>
+        /// Check if logevent has properties (or context properties)
+        /// </summary>
+        /// <param name="logEvent"></param>
+        /// <returns>True if properties should be included</returns>
+        protected bool HasCombinedProperties(LogEventInfo logEvent)
```
304NotModified: proposal to rename (after renames below) -> ShouldIncludeProperties

I prefer removing the term "combined properties", if possible

304NotModified:
oops sorry for breaking the build

src/NLog/Targets/TargetWithContext.cs
snakefoot: The value is set upfront, so you can just call `ContextProperties.Add`, that is the style from other NLog Targets/Layouts. The `virtual` allows one to provide a different item-class with additional properties.

src/NLog/Targets/TargetWithContextProperty.cs
```diff
@@ -0,0 +1,81 @@
...
+{
+    using NLog.Config;
+    using NLog.Layouts;
+
+    /// <summary>
+    /// Attribute details for <see cref="TargetWithContext"/> 
+    /// </summary>
+    [NLogConfigurationItem]
+    [ThreadAgnostic]
+    public class TargetWithContextProperty
```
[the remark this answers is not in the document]
snakefoot: I prefer `TargetWithContext`, since also captures NDC and NDLC

src/NLog/Targets/TargetWithContextProperty.cs
snakefoot: Done

src/NLog/Targets/TargetWithContext.cs
snakefoot: Done

src/NLog/Targets/TargetWithContext.cs
snakefoot: Done

src/NLog/Targets/TargetWithContext.cs
snakefoot: Done

snakefoot:
> edit: can't see what's wrong. Works locally

There is a conflict with my optimizations of the NetworkTarget and this PR. I have now merged with master, and resolved the conflict.

src/NLog/Targets/TargetWithContextProperty.cs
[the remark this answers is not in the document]
snakefoot: I prefer `TargetWithContextProperty` since it belongs to `TargetWithContext`. But I'm not religious about the name.

304NotModified:
merged :)

PR #2471, opened 2017-12-21
TreatWarningsAsErrors = true

snakefoot:
And fix license url, so the link from Nuget-package becomes human readable.


304NotModified:
thanks! great enhancement!

PR #2472, opened 2017-12-21
LogManager.LoadConfiguration - Prepare for custom config file readers

snakefoot:
Thinking that custom config file readers could be registered in NLog, that include their own candidate paths, but also allow direct file loading depending on file-extension (and priority in registration)

Instead of NLog.Extension.Logging and Nlog.Web creates XmlLoggingConfiguration, then they should just call:

`LogManager.LoadConfiguration(configFile);`

In the future NLog can then be extended to load custom config-files, without needing to change NLog.Extension.Logging and NLog.Web.

### DIFF
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -218,7 +218,7 @@ namespace NLog
                     }
 
 #if __ANDROID__
-                    if (this._config == null)
+                    if (_config == null)
                     {
                         TryLoadFromAndroidAssets();
                     }
@@ -319,8 +319,6 @@ namespace NLog
 
                 //ToArray needed for .Net 3.5
                 InternalLogger.Warn(exception, "Cannot start file watching: {0}", string.Join(",", _config.FileNamesToWatch.ToArray()));
-
-                
             }
         }
 #endif
@@ -337,7 +335,7 @@ namespace NLog
                 {
                     if (stream != null)
                     {
-                        LoadLoggingConfiguration(XmlLoggingConfiguration.AssetsPrefix + nlogConfigFilename);
+                        _config = TryLoadLoggingConfiguration(XmlLoggingConfiguration.AssetsPrefix + nlogConfigFilename);
                     }
                 }
             }
@@ -355,16 +353,16 @@ namespace NLog
             foreach (string configFile in configFileNames)
             {
 #if SILVERLIGHT && !WINDOWS_PHONE
-                            Uri configFileUri = new Uri(configFile, UriKind.Relative);
-                            if (Application.GetResourceStream(configFileUri) != null)
-                            {
-                                LoadLoggingConfiguration(configFile);
-                                break;
-                            }
+                Uri configFileUri = new Uri(configFile, UriKind.Relative);
+                if (Application.GetResourceStream(configFileUri) != null)
+                {
+                    _config = TryLoadLoggingConfiguration(configFile);
+                    break;
+                }
 #else
                 if (File.Exists(configFile))
                 {
-                    LoadLoggingConfiguration(configFile);
+                    _config = TryLoadLoggingConfiguration(configFile);
                     break;
                 }
 #endif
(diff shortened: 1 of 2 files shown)


snakefoot:
Abused the same PR to also introduce Logger.WithProperty and Logger.WithLogicalContext. See also #2466

304NotModified:
Nice! (Both)

As you have noticed, busy days. Will try to review in a few days

src/NLog/LogFactory.cs
```diff
@@ -821,7 +819,7 @@ internal void ReloadConfigOnTimer(object state)
 
                     if (newConfig is XmlLoggingConfiguration xmlConfig)
                     {
-                        if (!xmlConfig.InitializeSucceeded.HasValue || !xmlConfig.InitializeSucceeded.Value)
+                        if ((xmlConfig.InitializeSucceeded ?? false) == false)
```
304NotModified: I prefer

`if(xmlConfig.InitializeSucceeded  != true)`

or 

`if(!(xmlConfig.InitializeSucceeded ?? false))`

(`== true` or `==false` is a common code smell)

src/NLog/LogFactory.cs
```diff
@@ -1248,10 +1255,18 @@ private static Logger CreateDefaultLogger(ref LoggerCacheKey cacheKey)
             return newLogger;
         }
 
-        private void LoadLoggingConfiguration(string configFile)
+
+        internal LoggingConfiguration LoadConfiguration(string configFile)
```
304NotModified: I think this should be `public`, as the LogFactory is "in sync" with the LogManager. What do you think?

src/NLog/LogFactory.cs
snakefoot: Changed as suggested

src/NLog/LogFactory.cs
snakefoot: Made public as suggested.

src/NLog/LogFactory.cs
304NotModified: Guess this needs to be pushed? It's still shows `== false` ;)

src/NLog/LogFactory.cs
```diff
@@ -1153,7 +1151,16 @@ private static IEnumerable<string> GetDefaultCandidateConfigFilePaths()
 #endif
         }
 
-        private Logger GetLogger(LoggerCacheKey cacheKey)
+        internal Logger GetContextLogger(string loggerName, Type concreteType)
```
304NotModified: Could you explain this a bit? I see we need a new logger with unique name, isn't? 

Could we rename this to CreateNewUniqueLogger? (or something like that)

src/NLog/LogFactory.cs
```diff
@@ -1386,6 +1408,26 @@ private IEnumerable<Logger> GetLoggers()
 
                 return values;
             }
+
+            public void PruneLoggerCache()
```
304NotModified: We're removing the "guid" like loggers here, isn;t?

304NotModified:
> bused the same PR to also introduce Logger.WithProperty and Logger.WithLogicalContext.

Could you please split this in to 2 PRs, or at least 2 commits. 

I'm not sure if the Logger.WithProperty approach is optimal (with the temporary "guid" loggers, as far as i understand).

Thanks in advance!

src/NLog/LogFactory.cs
snakefoot: Yes when chaining multiple WithProperty-calls, then it will leave several loggers entries. If loggers has no static references, then it will slowly fill up the logger-cache. Only doing the prune, when activating the context-feature.

src/NLog/LogFactory.cs
snakefoot: Sure

snakefoot:
> Could you please split this in to 2 PRs, or at least 2 commits.

Done

304NotModified:
thx!

PR #2475, opened 2017-12-23
Fix XmlLoggingConfiguration reloading

### LINKED ISSUE #2279: NLog stops working when .config updated with invalid XML. Does not recover.
**Type**:

- Bug

**NLog version**: 4.4.12

**Platform**: .Net 4.6.1 

I have XML based configuration. Upon startup a program logs correctly. There is code that updates NLog.config with new XML and it does something to this extend:

```c#
private void Configure(string xmlConfig){
	string nlogConfigFile = $"{AppDomain.CurrentDomain.BaseDirectory}NLog.config";
	File.WriteAllText(nlogConfigFile, xmlConfig);
	LogManager.Configuration = LogManager.Configuration.Reload();
}
```

As soon as broken XML is written into .config file, "Reload()" returns an empty configuration. Calling the same method with correct XML has no effect - "Reload()" never creates a properly created Configuration object. The only way to get LogManager working is to restart the application (next to impossible in my case).

I see that a similar bug was reported and fixed years ago, but it clearly is broken for me: https://github.com/NLog/NLog/issues/273 . Any help is appreciated.

MikeFH:
Fixes #2279 

In the XmlLoggingConfiguration :
In case the file contains an invalid xml, an exception is thrown in the Initialize method and the _originalFileName was never saved. Since Reload() uses the _originalFileName to read the xml file, it was impossible to reload the config.

### DIFF
--- a/src/NLog/Config/XmlLoggingConfiguration.cs
+++ b/src/NLog/Config/XmlLoggingConfiguration.cs
@@ -406,11 +406,11 @@ namespace NLog.Config
             try
             {
                 InitializeSucceeded = null;
+                _originalFileName = fileName;
                 reader.MoveToContent();
                 var content = new NLogXmlElement(reader);
                 if (fileName != null)
-                {
-                    _originalFileName = fileName;
+                {                    
                     ParseTopLevel(content, fileName, autoReloadDefault: false);
 
                     InternalLogger.Info("Configured from an XML element in {0}...", fileName);
--- a/tests/NLog.UnitTests/Config/ReloadTests.cs
+++ b/tests/NLog.UnitTests/Config/ReloadTests.cs
@@ -564,6 +564,39 @@ namespace NLog.UnitTests.Config
 
         }
 
+        [Fact]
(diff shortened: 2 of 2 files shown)


MikeFH:
As a side note, is it normal that we do not throw a `NullArgumentException` if `fileName` is null ?
https://github.com/NLog/NLog/blob/b209b344c331fb917cab25d42bc034ee73027431/src/NLog/Config/XmlLoggingConfiguration.cs#L126

We also allow to call `Reload()` if `_originalFileName` is null (this is possible when the config was initialized from a `XmlReader`)
https://github.com/NLog/NLog/blob/b209b344c331fb917cab25d42bc034ee73027431/src/NLog/Config/XmlLoggingConfiguration.cs#L304

304NotModified:
> As a side note, is it normal that we do not throw a NullArgumentException if fileName is null ?

I think the fileName is optional?

304NotModified:
Thanks! great work!

MikeFH:
Thanks !

PR #2485, opened 2018-01-02
FileTarget - Dispose Archive-Mutex after completing file-archive

snakefoot:
Explicit cleanup, instead of waiting for the garbage collector.


304NotModified:
👍 thanks!

PR #2487, opened 2018-01-03
${AssemblyVersion}: add type (File, Assembly, Informational) option

### LINKED ISSUE #2480: Add version types to AssemblyVersion layout renderer
**Type**: Feature request

**NLog version**: 4.5.0

**Platform**: .Net 3.5 / .Net 4.0 / .Net 4.5 / Mono 4 / Silverlight 4 / Silverlight 5 / Xamarin Android / Xamarin iOs / .NET Core 

**Current NLog config**: N/A

I would like to extend the AssemblyVersion layout renderer to be able to specify different version types, e.g. from [here](https://github.com/dotnet/sdk/blob/1f1485fc216cd87b0289737eb7fe7de417335022/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.NET.GenerateAssemblyInfo.targets#L56-L85):
- AssemblyVersion
- FileVersion
- InformationalVersion

So the current renderer syntax of `${assembly-version:name=string}` could become `${assembly-version:name=string:type=string}` where the value for `type` is one of:
- assembly
- file
- informational

I'm happy to provide this in a pull request with unit tests, etc... if agreed 😃

alexangas:
Fixes #2480.

Requires documentation updates at https://github.com/nlog/nlog/wiki/AssemblyVersion-Layout-Renderer and https://github.com/NLog/NLog/wiki/Platform-Support (see code for details).

### DIFF
--- a/src/NLog/LayoutRenderers/AssemblyVersionLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/AssemblyVersionLayoutRenderer.cs
@@ -40,9 +40,15 @@ namespace NLog.LayoutRenderers
     using NLog.Internal;
 
     /// <summary>
-    /// Assembly version.
+    /// Renders the assembly version information for the entry assembly or a named assembly.
     /// </summary>
-    /// <remarks>The entry assembly can't be found in some cases e.g. ASP.NET, Unit tests etc.</remarks>
+    /// <remarks>
+    /// As this layout renderer uses reflection and version information is unlikely to change during application execution,
+    /// it is recommended to use it in conjunction with the <see cref="NLog.LayoutRenderers.Wrappers.CachedLayoutRendererWrapper"/>.
+    /// </remarks>
+    /// <remarks>
+    /// The entry assembly can't be found in some cases e.g. ASP.NET, unit tests, etc.
+    /// </remarks>
     [LayoutRenderer("assembly-version")]
     public class AssemblyVersionLayoutRenderer : LayoutRenderer
     {
--- a/src/NLog/LayoutRenderers/AssemblyVersionLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/AssemblyVersionLayoutRenderer.cs
@@ -115,7 +115,7 @@ namespace NLog.LayoutRenderers
             }
         }
 
-#elif WINDOWS_UWP
+#elif WINDOWS_UWP && !NETSTANDARD1_5
 
         private string GetVersion()
         {
--- a/src/NLog/LayoutRenderers/AssemblyVersionLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/AssemblyVersionLayoutRenderer.cs
@@ -115,7 +115,7 @@ namespace NLog.LayoutRenderers
             }
         }
 
-#elif WINDOWS_UWP //&& !NETSTANDARD1_5
+#elif WINDOWS_UWP
 
         private string GetVersion()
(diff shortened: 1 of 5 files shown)


304NotModified:
The build server says:

```
'AssemblyVersionLayoutRenderer.GetVersion()': not all code paths return a value [C:\projects\nlog\src\NLog\NLog.csproj]

```

Could you please check that?

src/NLog/LayoutRenderers/AssemblyVersionLayoutRenderer.cs
```diff
@@ -31,60 +31,144 @@
...
         }
 
-        string GetAssemblyApplicationVersion(string assemblyName)
+#if SILVERLIGHT
+
+        private string GetVersion()
         {
-            if (!string.IsNullOrEmpty(assemblyName))
+            var assemblyName = GetAssemblyName();
+            assemblyName.Version.ToString();
```
304NotModified: missing `return` ;)

tests/NLog.UnitTests/LayoutRenderers/AssemblyVersionTests.cs
```diff
@@ -52,5 +53,93 @@ public void AssemblyNameVersionTest()
...
+            var testLoggerType = compiledAssembly.GetType("LogTester.LoggerTest");
+            var logMethod = testLoggerType.GetMethod("TestLog");
+            var testLoggerInstance = Activator.CreateInstance(testLoggerType);
+            logMethod.Invoke(testLoggerInstance, new object[] { logger, compiledAssembly });
+            AssertDebugLastMessage("debug", expected);
+        }
+
+        private static Assembly GenerateTestAssembly()
+        {
+            const string code = @"
```
304NotModified: nice!

304NotModified:
Unfortunately, it fails on Travis

.>  NLog.UnitTests.LayoutRenderers.AssemblyVersionTests.AssemblyVersionTypeTest(type: Assembly, expected: "1.1.1.1") [FAIL]
      System.InvalidOperationException : Unable to find field m_entryAssembly

Any idea why? See https://travis-ci.org/NLog/NLog/builds/324836802

alexangas:
The issue appears to be the way that xunit handles app domains, described in Microsoft/vstest#649. This means that calls to `Assembly.GetEntryAssembly()` may return null (as can happen in existing `EntryAssemblyVersionTest`) or attempts to force set it may fail (as can happen in new `AssemblyVersionTypeTest`).

There may be a way to get around this by changing the xunit config value `xunit.appDomain`, but it looks like this may affect all tests and cause adverse effects.

So in these most recent pushes I have taken the same approach as the existing `EntryAssemblyVersionTest` when this wasn't working where it effectively skips the test if there are problems. It also attempts to log that.

304NotModified:
> So in these most recent pushes I have taken the same approach as the existing EntryAssemblyVersionTest when this wasn't working where it effectively skips the test if there are problems. It also attempts to log that

seems to work great!

Thanks, looks good! 

Could you please update the wiki and add "Introduced in NLog 4.5"? Thanks in advance!

alexangas:
Really appreciate all your help. I've updated https://github.com/nlog/nlog/wiki/AssemblyVersion-Layout-Renderer and https://github.com/NLog/NLog/wiki/Platform-Support.

304NotModified:
thanks! also for adding examples! 👍 👍

PR #2490, opened 2018-01-05
LogEventInfo.MessageTemplateParameters as class instead of interface

### LINKED ISSUE #2489: Make IMessageTemplateParameters internal
Change LogEventInfo.MessageTemplateParameters to return the actual MessageTemplateParameters (and make it public sealed).

This will lead to less breaking changes, in case someone starts to mock IMessageTemplateParameters, and we want to extend MessageTemplateParameters.

snakefoot:
returning a class instead of interface, otherwise we could not add properties to the class/interface in the future

Resolves #2489


snakefoot:
Would really like to have #2467 included for Nlog 4.5. Planning to use it in Seq integration that is missing Mdlc


Sent from my Samsung device


304NotModified:
OK, let's discuss that https://github.com/NLog/NLog/pull/2467. NLog 4.5 RC5 would also be an option

PR #2493, opened 2018-01-08
RegisterItemsFromAssembly - Include assemblies from nuget packages

### LINKED ISSUE #2364: Auto load Nlog.Web.AspNetCore
auto load from factory.

[In this method](https://github.com/NLog/NLog/blob/a4ec69bc991949d8047c7bee0f3726e7081ae15d/src/NLog/Config/ConfigurationItemFactory.cs#L364)

needs a try-catch.

snakefoot:
Attempt to resolve #2364 

Together with #2472, then you can configure NLog completely with just this line (netstandard2.0 + net45 only).

`var logger = LogManager.LoadConfiguration("nlog.config").GetCurrentClassLogger();`

Where it correctly loads extensions and ignores assemblies (without config reload etc.)

### DIFF
--- a/src/NLog/Config/ConfigurationItemFactory.cs
+++ b/src/NLog/Config/ConfigurationItemFactory.cs
@@ -394,6 +394,7 @@ namespace NLog.Config
                 }
 
                 InternalLogger.Debug("Start auto loading, location: {0}", assemblyLocation);
+                HashSet<string> alreadyRegistered = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
                 foreach (var extensionDll in extensionDlls)
                 {
                     InternalLogger.Info("Auto loading assembly file: {0}", extensionDll);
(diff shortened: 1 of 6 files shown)


304NotModified:
:+1: :+1:

304NotModified:
nice one! thanks!

304NotModified:
how will this work if I added a nuget package but it's not referenced in code? Then it's never in the current appdomain and not autoloaded?

snakefoot:
> how will this work if I added a nuget package but it's not referenced in code? 

Believe that if the nuget-package adds assembly references to the project, then it will automatically load the assembly. But not 100 pct sure if it is bullet proof in all cases. Made a simple test,, and it worked okay.

But I can verify that the callsite-add-hidden-assembly-logic works very nicely for NLog.Extensions, NLog.Web and Microsoft.Logging-callsite-filtering.

304NotModified:
OK, thanks!

PR #2496, opened 2018-01-09
Logger with support for WithProperty and WithLogicalContext

snakefoot:
Maybe consider to extend with xml-config support as well. See also #528 

But then it needs to be a capture of a functor, that either returns the WithProperty-object, or calls Layout-render.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/2496)

### DIFF
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -531,7 +531,7 @@ namespace NLog
         /// are not guaranteed to return the same logger reference.</returns>
         public Logger GetLogger(string name)
         {
-            return GetLogger(new LoggerCacheKey(name, typeof(Logger)));
+            return GetLogger(new LoggerCacheKey(name, typeof(Logger)), name);
         }
 
         /// <summary>
@@ -543,7 +543,7 @@ namespace NLog
         /// are not guaranteed to return the same logger reference.</returns>
         public T GetLogger<T>(string name) where T : Logger
         {
-            return (T)GetLogger(new LoggerCacheKey(name, typeof(T)));
+            return (T)GetLogger(new LoggerCacheKey(name, typeof(T)), name);
         }
 
         /// <summary>
@@ -555,7 +555,7 @@ namespace NLog
         /// same argument aren't guaranteed to return the same logger reference.</returns>
         public Logger GetLogger(string name, Type loggerType)
         {
-            return GetLogger(new LoggerCacheKey(name, loggerType));
+            return GetLogger(new LoggerCacheKey(name, loggerType), name);
         }
 
         /// <summary>
@@ -1151,7 +1151,16 @@ namespace NLog
 #endif
         }
 
-        private Logger GetLogger(LoggerCacheKey cacheKey)
+        internal Logger CreateNewUniqueLogger(string loggerName, Type concreteType)
+        {
+            lock (_syncRoot)
+            {
+                _loggerCache.PruneLoggerCache();
+                return GetLogger(new LoggerCacheKey(string.Concat(loggerName, "_", Guid.NewGuid().ToString()), concreteType), loggerName);
+            }
+        }
+
+        private Logger GetLogger(LoggerCacheKey cacheKey, string loggerName)
         {
             lock (_syncRoot)
             {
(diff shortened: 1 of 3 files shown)


304NotModified:
Could you please explain me why we're creating new (temporary guid) loggers?

Is it for the following cases?

```c#
var logger1 = LogManager.GetLogger("logger1");
var logger2 = logger1.WithProperty("prop1", "val1");
```

logger1 wont add the property prop1, but logger 2 does. Is that correct?

Discussion: if logger1 would also add prop1 (and maybe rename it to AddProperty instead of WithProperty), then no temp loggers are needed? Do you agree? 

Benefit in that case, this would work also:

```c#
var logger1 = LogManager.GetLogger("logger1");
 logger1.AddProperty("prop1", "val1"); //don't use the return value
```

snakefoot:
@304NotModified I was thinking of a usage pattern similar to what I can see in Serilog-ForContext, where the new logger can be handed over to a new owner.

If wanting multiple loggers with the same name, that are initialized with different property-values, then your solution will not work.

304NotModified:
> If wanting multiple loggers with the same name, that are initialized with different property-values, then your solution will not work.

true, but it would be easier to merge with the XML config (#528)

doubting about this. What do self think about the temporary guid loggers?

snakefoot:
> @304NotModified But it would be easier to merge with the XML config #528 

Yes the properties added using logging-rules will be "global" and should be automatically applied to the main-logger.

snakefoot:
Must admit I don't need the overhead from either WithProperty or logging-rules-properties.

But I wanted to test out different solutions for solving #2465, because I can see lots of NLog targets not supporting logging-context verify well (besides the logger-name).

304NotModified:
I was looking at this.

I think changing it to AddLogicalContext and AddProperty would be more in sync with the current GDC/MDC etc. (or maybe even CaptureLogicalContext  and SetProperty)

Also it makes it technical easier and more straight forward.


> Must admit I don't need the overhead from either WithProperty or logging-rules-properties. Though I guess their overhead is smaller compared to using MDLC.

it's a bit magic IMO, and I hope it's not needed. Also the non-destructive update is nice, but (unfortunately?) not in-line with the current methods.

snakefoot:
>Maybe we should postpone this to after 4.5?

Good idea

snakefoot:
> So my proposal is to change this to AddLogicalContext/AddProperty or CaptureLogicalContext/SetProperty

Think it should be two PRs. Think both solutions a very valid. 

- AddProperty - That affects all active logger objects
- WithProperty -That only affects the returned logger object (This PR)

304NotModified:
This PR is still open and I like to give it an direction.

While I like fluent style programming, I'm doubting if that will fit in the current NLog API. In NLog it's not common to create multiple loggers in that way. But there is a fluent style already in the NLog.Fluent namespace.

Also, if we really need a like to add a fluent calls, like `.WithProperty`, then I think it needs to return an intermediate (state) object, and it should not reuse the `Logger` for that. I really think it's a bad idea to create guid-temporary loggers and prune them.

So my proposal is:

- or do a fluent style and use an intermediate (state) object, and make it consistent with NLog.Fluent stuff.
- or add `SetProperty` and change the logger object.

304NotModified:
bump @sna

snakefoot:
@304NotModified Well I like WithProperty, what if I change the logic, so instead of using a Guid, then the LoggerCache becomes:

`Dictionary<LoggerCacheKey, KeyValuePair<WeakReference, IList<WeakReference>> _loggerCache;`

So instead of using Guid, then Loggers with special context is inserted in the `IList<WeakReferences>`?

304NotModified:
> So my proposal is:
> 
> or do a fluent style and use an intermediate (state) object, and make it consistent with NLog.Fluent stuff.
> or add SetProperty and change the logger object.

You don't like one of these proposals?

Guid of weakreference, it's has the same issue (temp loggers and prune, but without guid)

>  really think it's a bad idea to create guid-temporary loggers and prune them.

snakefoot:
> or do a fluent style and use an intermediate (state) object,

The returned state object should be a Logger-object or an ILogger-object. Sadly enough the ILogger-interface is broken by design. So I prefer a Logger-object.

> add SetProperty and change the logger object.

Probably useful in some cases, by I prefer to have the dynamic first, and then introduce the global/fixed property afterwards (easy part).

snakefoot:
Closing as superseded by #3298

Issue #2434, opened 2017-11-30
NLog.Schema + NLog.Config for new csproj file format


luigiberrettini:
**Bug**

**NLog version**: 4.4.12

In case of a BUG:

- What is the current result?
XSD added to %USERPROFILE%\.nuget\packages

- What is the expected result?
XSD added to referencing project folder

- Are there any work arrounds? yes/no

- Is there a version in which it did worked? 
https://github.com/NuGet/Home/issues/4803

304NotModified:
hi!

Any idea how to fix this?

304NotModified:
this is also an issue for nlog.config I guess

luigiberrettini:
A csproj to replace the schema nuspec can help if the xsd is packed into
the contentFiles folder.
I am working on the same issue forthe Syslog target and will let you know
as soon as I validate my solution.

304NotModified:
:+1: thanks!

304NotModified:
Would be great if I could add this for NLog 4.5 rc2 ;)

luigiberrettini:
You should run `dotnet pack --no-build` in a folder with a `NLog.Schema.csproj` file like the following one:

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFrameworks>net45;netstandard2.0</TargetFrameworks>

    <Product>NLog.Schema</Product>
    <Copyright>Copyright info</Copyright>

    <Company>Company info</Company>
    <Authors>Authors info</Authors>
    <PackageId>NLog.Schema</PackageId>
    <Title>NLog schema for Intellisense™</Title>
    <Description>Enables Intellisense™ for NLog settings in NLog.config</Description>
    <PackageTags>NLog Log Logging Intellisense xsd</PackageTags>
    <Version>0.1.0-alpha-01-commitHash</Version>
    <PackageLicenseUrl>https://github.com/NLog/NLog/blob/master/LICENSE</PackageLicenseUrl>
    <PackageProjectUrl>https://github.com/NLog/NLog</PackageProjectUrl>
    <PackageIconUrl>http://nlog-project.org/NConfig.png</PackageIconUrl>
    <RepositoryUrl>https://github.com/NLog/NLog</RepositoryUrl>
    <PackageReleaseNotes>https://github.com/NLog/NLog/releases/tag/v0.1.0-alpha-01-commitHash</PackageReleaseNotes>
    <RepositoryType>Git</RepositoryType>
[...]

PR #2517, opened 2018-01-15
Update nuspec NLog.Config and NLog.Schema

304NotModified:
and update nuget.exe

`<ContentFiles>`, nuget 3.3+

fixes https://github.com/NLog/NLog/issues/2434


luigiberrettini:
I suggest to move to a csproj file since NuGet is part of dotnet CLI and nobody knows how long nuspec files and a standalone NuGet.exe will be supported
In #2434 you can find the file: you only have to replace xml tag contents

304NotModified:
The nuspec is the core of nuget? It's in the nupkg.  Also don't is just calling nuget.exe afaik ;)

Thanks for the suggestion, buy I don't like csproj with one file :p

304NotModified:
superseded by https://github.com/NLog/NLog/pull/2570

PR #2519, opened 2018-01-15
RegisterItemsFromAssembly - Include assemblies from nuget packages (Strict)

snakefoot:
Less aggressive, so NLog.Extensions.Logging.Tests doesn't get marked as hidden assembly.

### DIFF
--- a/src/NLog/Config/ConfigurationItemFactory.cs
+++ b/src/NLog/Config/ConfigurationItemFactory.cs
@@ -395,6 +395,7 @@ namespace NLog.Config
 
                 InternalLogger.Debug("Start auto loading, location: {0}", assemblyLocation);
                 HashSet<string> alreadyRegistered = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+                alreadyRegistered.Add(nlogAssembly.FullName);
                 foreach (var extensionDll in extensionDlls)
                 {
                     InternalLogger.Info("Auto loading assembly file: {0}", extensionDll);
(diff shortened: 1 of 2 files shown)


src/NLog/Config/ConfigurationItemFactory.cs
```diff
@@ -427,18 +428,21 @@ private static ConfigurationItemFactory BuildDefaultFactory()
                 var allAssemblies = LogFactory.CurrentAppDomain.GetAssemblies();
                 foreach (var assembly in allAssemblies)
                 {
-                    if (assembly.FullName.StartsWith("NLog.", StringComparison.OrdinalIgnoreCase))
+                    if (assembly.FullName.StartsWith("NLog", StringComparison.OrdinalIgnoreCase))
```
304NotModified: This is less strict? This would match nlogger.dll?

src/NLog/Config/ConfigurationItemFactory.cs
```diff
@@ -427,18 +428,21 @@ private static ConfigurationItemFactory BuildDefaultFactory()
...
-                      || assembly.FullName.StartsWith("Microsoft.Extensions.Logging", StringComparison.OrdinalIgnoreCase)
-                      || assembly.FullName.StartsWith("Microsoft.Logging", StringComparison.OrdinalIgnoreCase))
+                    if ( assembly.FullName.StartsWith("NLog.Extensions.Logging,", StringComparison.OrdinalIgnoreCase)
+                      || assembly.FullName.StartsWith("NLog.Web,", StringComparison.OrdinalIgnoreCase)
+                      || assembly.FullName.StartsWith("NLog.Web.AspNetCore,", StringComparison.OrdinalIgnoreCase)
+                      || assembly.FullName.StartsWith("Microsoft.Extensions.Logging,", StringComparison.OrdinalIgnoreCase)
+                      || assembly.FullName.StartsWith("Microsoft.Extensions.Logging.Abstractions,", StringComparison.OrdinalIgnoreCase)
+                      || assembly.FullName.StartsWith("Microsoft.Extensions.Logging.Filter,", StringComparison.OrdinalIgnoreCase)
+                      || assembly.FullName.StartsWith("Microsoft.Logging,", StringComparison.OrdinalIgnoreCase))
                     {
```
304NotModified: Why the duplicates, it's a "starts with"

src/NLog/Config/ConfigurationItemFactory.cs
snakefoot: It matches the current: `Directory.GetFiles(assemblyLocation, "NLog*.dll")`

It is the hidden assembly logic, that is more strict.

src/NLog/Config/ConfigurationItemFactory.cs
snakefoot: Notice the `,`

src/NLog/Config/ConfigurationItemFactory.cs
snakefoot: Though I don't mind reverting it back to `NLog.` (But then we should consider also fixing the `GetFiles`-call

src/NLog/Config/ConfigurationItemFactory.cs
304NotModified: Ah thanks, hardly to see on mobile.

src/NLog/Config/ConfigurationItemFactory.cs
304NotModified: Changing the getfiles is backwards compatible, isn't? Would prefer that one.

snakefoot: Nope it will suddenly not load `NLogSplunkTarget.dll` (Made Up)

snakefoot: But you have convinced me, matching on filename is different from matching on namespace. Have added the dot back, so it remains strict.

304NotModified: But the filter start with "nlog." was already there?

snakefoot: Yes and now it is back again :)

304NotModified: :)

PR #2524, opened 2018-01-18
FileTarget - Dynamic archive mode with more strict file-mask for cleanup

snakefoot:
Avoid too aggressive cleanup when having multiple log-archives in the same folder.

Make the NLog 4.5 archive logic more userfriendly.

### DIFF
--- a/src/NLog/Targets/FileArchiveModes/FileArchiveModeDynamicSequence.cs
+++ b/src/NLog/Targets/FileArchiveModes/FileArchiveModeDynamicSequence.cs
@@ -65,11 +65,6 @@ namespace NLog.Targets.FileArchiveModes
             _customArchiveFileName = customArchiveFileName;
         }
 
-        protected override FileNameTemplate GenerateFileNameTemplate(string archiveFilePath)
-        {
-            return null;
-        }
-
         private static bool RemoveNonLetters(string fileName, int startPosition, StringBuilder sb, out int digitsRemoved)
         {
             digitsRemoved = 0;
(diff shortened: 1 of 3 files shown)


304NotModified:
> Make the NLog 4.5 archive logic more userfriendly.

Could you elaborate on this?

> more userfriendly

👍 👍

snakefoot:
> Could you elaborate on this?

Just look at the change of the unit-test, and you should see the problem this PR is trying to fix.

LogFile1 has this filename: `log.txt`
LogFile2 has this filename: `log-other.txt`

The automatic wildcard for LogFile1 will become `log*.txt`, which will also cleanup LogFile2 files.

304NotModified:
is this a case we could unit test? (the first red part)

image

304NotModified:
any idea if this could fix the unstable test " NLog.UnitTests.Targets.PlainFileTargetTests.SimpleFileDeleteTest(concurrentWrites: False, keepFileOpen: True, networkWrites: False, forceManaged: True, forceMutexConcurrentWrites: False, optimizeBufferReuse: True) [FAIL]"

See https://travis-ci.org/NLog/NLog/builds/329033070 (branch master...)

snakefoot:
> is this a case we could unit test? (the first red part)

It is just a safe-guard to ensure the double-string-iteration doesn't go bananas. You are welcome to think up a unit test.

snakefoot:
> any idea if this could fix the unstable test " 

Well my "fix" is to skip the keepfileopen-tests, as they seem very on unstable on Travis right now.

Issue #2532, opened 2018-01-22
Handle Mixed message templates better, e.g. {0} {Order}


304NotModified:
I'm working on the docs and examples for NLog 4.5

Maybe we discussed this before, but I'm trying to use it as a regular NLog user.

**Type** (choose one):

- Bug / feature

**NLog version**:  4.5 rc4

**Platform**:  all platforms, tested on .NET Core

**Current NLog config** (xml or C#, if relevant)

```c#

   var order = new Order
            {
                ClientId = 301,
                OrderId = 2500002,
                Remarks = "My First order"
            };

logger.Info("Process order {0} for {@Order}", "CoolClient", order);
```

In case of a BUG:

- What is the current result?

No formatting as all of the message, no properties

The message is currently ` Process order {0} for {@Order}`

- What is the expected result?

Handle as structured template and create properties.

expected result:

```
 Process order "CoolClient" for {"OrderId":2500002, "ClientId":301, "Remarks":"My First order", "Express":false, "Status":"Processing"}
```

NB: `    Process order {@0} for {@Order}` also won't work

- Are there any work arrounds? no

304NotModified:
@snakefoot please your opinion. I could add an unit test if you agree on this.

snakefoot:
Mixed mode should work when having parseMessageTemplate=true

Please create unit trst and fix code if not


Sent from my Samsung device


-------- Original message --------

304NotModified:
and with parseMessageTemplate=auto?

snakefoot:
Then you want performance and none breaking behaviour



Sent from my Samsung device


-------- Original message --------

304NotModified:
true,

as `Process order {0} for {Order}` wasn't allowed first, I don't think would be a breaking change

304NotModified:
The rules are clear, really like to hold on that:



> - Templates that use numeric property names like {0} and {1} exclusively imply that arguments to the template are captured by numeric index
> - If any of the property names are non-numeric, then all arguments are captured by matching left-to-right with holes in the order in which they appear

from https://messagetemplates.org/

Any idea how to fix bullet 2, if the whole parse is now lazy? Could we do a non-lazy parse if we detected is a non-string.format?

snakefoot:
>  Any idea how to fix bullet 2, if the whole parse is now lazy?

Should not be impossible to do an optimistic parse, and then do a full-parse if detecting a mixed mode situation. All the logic you implemented is still there, have not touched it.

[...]

PR #2540, opened 2018-01-23
Added mixed structured logging tests

304NotModified:
See  #2532

### DIFF
--- a/tests/NLog.UnitTests/LoggerTests.cs
+++ b/tests/NLog.UnitTests/LoggerTests.cs
@@ -1814,6 +1814,34 @@ namespace NLog.UnitTests
         }
 
         [Theory]
+        [InlineData(null, "OrderId", "@Client")]
+        [InlineData(true, "OrderId", "@Client")]
+        [InlineData(null, "0", "@Client")] //fails: gives "Process order {0} for {@Client}"
+        [InlineData(true, "0", "@Client")] //fails: gives "...ess order 13424 for 13424"
+        [InlineData(true, "OrderId", "Client")] //succeeeds, but gives JSON like (no quoted key, missing quotes arround string, =, other spacing)
+        public void MixedStructuredEventsConfigTest(bool? parseMessageTemplates, string param1, string param2)
+        {
+            LogManager.Configuration = CreateConfigurationFromString(@"
+                <nlog parseMessageTemplates='" + (parseMessageTemplates?.ToString() ?? string.Empty) + @"'>
+                    <targets><target name='debug' type='Debug' layout='${message}${exception}' /></targets>
+                    <rules>
+                        <logger name='*' writeTo='debug' />
+                    </rules>
+                </nlog>");
+            ILogger logger = LogManager.GetLogger("A");
+            logger.Debug("Process order {"+ param1 + "} for {" + param2 + "}", 13424, new {ClientId = 3001, ClientName = "John Doe"});
+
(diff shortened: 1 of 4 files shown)


tests/NLog.UnitTests/LoggerTests.cs
```diff
@@ -1813,6 +1813,34 @@ public void SingleTargetMessageFormatOptimizationTest()
             AssertDebugLastMessage("target2", "DualTarget|Hello World");
         }
 
+        [Theory]
+        [InlineData(null, "OrderId", "@Client")]
+        [InlineData(true, "OrderId", "@Client")]
+        [InlineData(null, "0", "@Client")] //fails: gives "Process order {0} for {@Client}"
+        [InlineData(true, "0", "@Client")] //fails: gives "...ess order 13424 for 13424"
+        [InlineData(true, "OrderId", "Client")] //succeeeds, but gives JSON like (no quoted key, missing quotes arround string, =, other spacing)
```
304NotModified: not sure if this is expected, at least not the unquoted string

tests/NLog.UnitTests/LoggerTests.cs
```diff
@@ -1813,6 +1813,34 @@ public void SingleTargetMessageFormatOptimizationTest()
             AssertDebugLastMessage("target2", "DualTarget|Hello World");
         }
 
+        [Theory]
+        [InlineData(null, "OrderId", "@Client")]
+        [InlineData(true, "OrderId", "@Client")]
+        [InlineData(null, "0", "@Client")] //fails: gives "Process order {0} for {@Client}"
+        [InlineData(true, "0", "@Client")] //fails: gives "...ess order 13424 for 13424"
```
304NotModified: bug for sure

tests/NLog.UnitTests/LoggerTests.cs
```diff
@@ -1813,6 +1813,34 @@ public void SingleTargetMessageFormatOptimizationTest()
             AssertDebugLastMessage("target2", "DualTarget|Hello World");
         }
 
+        [Theory]
+        [InlineData(null, "OrderId", "@Client")]
+        [InlineData(true, "OrderId", "@Client")]
+        [InlineData(null, "0", "@Client")] //fails: gives "Process order {0} for {@Client}"
```
304NotModified: Doubting if this is a bug, at least it's unwanted

snakefoot:
I agree that parseMessageTemplate = true should work. Please also try with the parameters reverted.



Sent from my Samsung device


tests/NLog.UnitTests/LoggerTests.cs
snakefoot: Works as expected, not valid string.Format input, so it just returns the raw unformatted message. Unless you want to have fallback on exception?

304NotModified:
@snakefoot I hacked the change, but now I have removed the support for index holes in a structured template. 

That's wrong, I need to fix that, but that's a bit hard because the parsing is lazy.

snakefoot:
> now I have removed the support for index holes in a structured template.

Thats sounds worse than not supporting mixed templates.

304NotModified:
> You have also introduced a nice performance overhead, by always allocating message-parameter-objects for index-holes.

that's because of `var currentHole = holeEnumerator.Current.Hole;`?

304NotModified:
superseded by #2544

PR #2544, opened 2018-01-24
MessageTemplate renderer with support for mixed mode templates

snakefoot:
Alternative solution for #2540 and #2532

Not sure what to do about this unit test, since it has wrong expectation of numeric templates:

`"{1} {OrderId}"` (Unit test thinks that `{1}` should print the first parameter, but it prints the second, causing the OrderId to print the first).

This is still not supported (mixed mode cheats the parser):

`"{1} {0} {OrderId}"`

Not sure I like mixed mode templates. I think it should be one or the other. Are they very common?

### DIFF
--- a/src/NLog/LogEventInfo.cs
+++ b/src/NLog/LogEventInfo.cs
@@ -390,7 +390,7 @@ namespace NLog
             {
                 if (_properties != null && _properties.MessageProperties.Count > 0)
                 {
-                    return new MessageTemplateParameters(_properties.MessageProperties);
+                    return new MessageTemplateParameters(_properties.MessageProperties, _message, _parameters);
                 }
                 else
                 {
--- a/src/NLog/MessageTemplates/MessageTemplateParameters.cs
+++ b/src/NLog/MessageTemplates/MessageTemplateParameters.cs
@@ -46,75 +46,114 @@ namespace NLog.MessageTemplates
         private readonly IList<MessageTemplateParameter> _parameters;
 
         /// <inheritDoc/>
-        public MessageTemplateParameter this[int index] => _parameters[index];
-
-        /// <inheritDoc/>
-        public int Count => _parameters.Count;
-
-        /// <inheritDoc/>
         public IEnumerator<MessageTemplateParameter> GetEnumerator() { return _parameters.GetEnumerator(); }
 
         /// <inheritDoc/>
         IEnumerator IEnumerable.GetEnumerator() { return _parameters.GetEnumerator(); }
 
-        /// <inheritDoc/>
+        /// <summary>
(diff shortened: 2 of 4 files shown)


snakefoot:
One could consider changing the constructor protection level to internal for (Before they were public constructors on an internal class):

MessageTemplateParameters
MessageTemplateParameter

One could consider having one public constructor on MessageTemplateParameters, that actually performs the parsing, so the parser is easy to access (Without making Template-class public).

One could consider adding a new property on MessageTemplateParameters, that tells whether the parse message-template is valid or not. And that the parameter-array-length actually matches the number of parsed parameter-holes.

One could consider having some more tests of whether it should return the half-rendered string on exception (ex. having more parameter-holes than parameters), or whether it should just abort everything and only show the message-template without any rendering at all (current default).

snakefoot:
@304NotModified One could consider changing the constructor protection level to internal for (Before they were public constructors on an internal class):

MessageTemplateParameters
MessageTemplateParameter

One could consider having one public constructor on MessageTemplateParameters, that actually performs the parsing, so the parser is easy to access (Without making Template-class public).

One could consider adding a new property on MessageTemplateParameters, that tells whether the parse message-template is valid or not. And that the parameter-array-length actually matches the number of parsed parameter-holes.

Can see Serilog creates special parameters for unmatched-parameter-array-object (Naming is __1, __2, etc.). These will of course only be accessible through the MessageTemplateParameters-object.

One could consider having some more tests of whether it should return the half-rendered string on exception (ex. having more parameter-holes than parameters), or whether it should just abort everything and only show the message-template without any rendering at all (current default).

304NotModified:
> Not sure I like mixed mode templates. I think it should be one or the other. Are they very common?

I hope so not, but I really like to keep to the rules of messageTemplates.org:

> If any of the property names are non-numeric, then all arguments are captured by matching left-to-right with holes in the order in which they appear

Mixing by-order an by-index isn't clear to me.

snakefoot:
> I hope so not, but I really like to keep to the rules of messageTemplates.org:

Not all standards makes sense, but feel free to take a dive into the code, and do some deep under water improvements for a very limited audience.

304NotModified:
Well, as I don't expect this will be a common case, we could maybe rewind after recognizing mixed mode. Less efficient, but semantic correct.

snakefoot:
No matter what, then this PR will improve the correctness of the default message-template-parser. Any improvements you want to make to unwind/rewind can be a different PR.

snakefoot:
Think I have found an okay solution for doing the unwind and rewind.

304NotModified:
great work again! Thanks! Really happy the mixed mode now works correct! 🍰

PR #2547, opened 2018-01-27
Windows EventLog for NetStandard2.0 - new NLog.WindowsEventLog package

snakefoot:
Depending on preview/pre-release nuget-package:

https://www.nuget.org/packages/System.Diagnostics.EventLog/

Expecting to be part of the NetCore compatibility pack:

https://blogs.msdn.microsoft.com/dotnet/2017/11/16/announcing-the-windows-compatibility-pack-for-net-core/

### DIFF
--- a/src/NLog/Common/ConversionHelpers.cs
+++ b/src/NLog/Common/ConversionHelpers.cs
@@ -1,4 +1,4 @@
-// 
+﻿// 
 // Copyright (c) 2004-2018 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
 // 
 // All rights reserved.
@@ -31,12 +31,40 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
-namespace NLog.Internal
+namespace NLog.Common
 {
     using System;
+    using System.Globalization;
+    using NLog.Internal;
 
-    internal static class EnumHelpers
+    /// <summary>
+    /// String Conversion Helpers
+    /// </summary>
+    public static class ConversionHelpers
     {
+        /// <summary>
+        /// Converts input string value into <see cref="System.Enum"/>
+        /// </summary>
+        /// <param name="inputValue">Input value</param>
+        /// <param name="resultValue">Output value</param>
+        /// <param name="defaultValue">Default value</param>
+        /// <returns>Returns failure if the input value could not be parsed</returns>
+        public static bool TryParseEnum<TEnum>(string inputValue, out TEnum resultValue, TEnum defaultValue = default(TEnum)) where TEnum : struct
+        {
+            if (string.IsNullOrEmpty(inputValue))
+            {
+                resultValue = defaultValue;
+                return true;
+            }
+
+            if (!TryParse<TEnum>(inputValue, true, out resultValue))
+            {
+                resultValue = defaultValue;
+                return false;
+            }
+            return true;
+        }
+
         /// <summary>
         /// Converts the string representation of the name or numeric value of one or more enumerated constants to an equivalent enumerated object. A parameter specifies whether the operation is case-sensitive. The return value indicates whether the conversion succeeded.
         /// </summary>
@@ -45,7 +73,7 @@ namespace NLog.Internal
         /// <param name="result">When this method returns, result contains an object of type TEnum whose value is represented by value if the parse operation succeeds. If the parse operation fails, result contains the default value of the underlying type of TEnum. Note that this value need not be a member of the TEnum enumeration. This parameter is passed uninitialized.</param>
         /// <returns><c>true</c> if the value parameter was converted successfully; otherwise, <c>false</c>.</returns>
         /// <remarks>Wrapper because Enum.TryParse is not present in .net 3.5</remarks>
-        public static bool TryParse<TEnum>(string value, out TEnum result) where TEnum : struct
+        internal static bool TryParse<TEnum>(string value, out TEnum result) where TEnum : struct
         {
             return TryParse(value, false, out result);
         }
@@ -59,7 +87,7 @@ namespace NLog.Internal
         /// <param name="result">When this method returns, result contains an object of type TEnum whose value is represented by value if the parse operation succeeds. If the parse operation fails, result contains the default value of the underlying type of TEnum. Note that this value need not be a member of the TEnum enumeration. This parameter is passed uninitialized.</param>
         /// <returns><c>true</c> if the value parameter was converted successfully; otherwise, <c>false</c>.</returns>
         /// <remarks>Wrapper because Enum.TryParse is not present in .net 3.5</remarks>
-        public static bool TryParse<TEnum>(string value, bool ignoreCase, out TEnum result) where TEnum : struct
+        internal static bool TryParse<TEnum>(string value, bool ignoreCase, out TEnum result) where TEnum : struct
         {
 #if NET3_5
             return TryParseEnum_net3(value, ignoreCase, out result);
--- a/src/NLog/Targets/EventLogTarget.cs
+++ b/src/NLog/Targets/EventLogTarget.cs
@@ -92,7 +92,7 @@ namespace NLog.Targets
         /// Initializes a new instance of the <see cref="EventLogTarget"/> class.
         /// </summary>
         /// <param name="name">Name of the target.</param>
-        public EventLogTarget(string name) 
+        public EventLogTarget(string name)
             : this(LogFactory.CurrentAppDomain)
         {
             Name = name;
@@ -270,11 +270,19 @@ namespace NLog.Targets
 
             EventLogEntryType entryType = GetEntryType(logEvent);
 
-            int eventId = EventId.RenderInt(logEvent, 0, "EventLogTarget.EventId");
+            int eventId = 0;
+            string renderEventId = RenderLogEvent(EventId, logEvent);
+            if (!string.IsNullOrEmpty(renderEventId) && !int.TryParse(renderEventId, out eventId))
+            {
+                InternalLogger.Warn("EventLogTarget(Name={0}): WriteEntry failed to parse EventId={1}", Name, renderEventId);
+            }
 
-            short category = Category.RenderShort(logEvent, 0, "EventLogTarget.Category");
-
-          
+            short category = 0;
+            string renderCategory = RenderLogEvent(Category, logEvent);
+            if (!string.IsNullOrEmpty(renderCategory) && !short.TryParse(renderCategory, out category))
+            {
+                InternalLogger.Warn("EventLogTarget(Name={0}): WriteEntry failed to parse Category={1}", Name, renderCategory);
+            }
 
             // limitation of EventLog API
             if (message.Length > MaxMessageLength)
@@ -317,17 +325,15 @@ namespace NLog.Targets
         /// <returns></returns>
         private EventLogEntryType GetEntryType(LogEventInfo logEvent)
         {
-            if (EntryType != null)
+            string renderEntryType = RenderLogEvent(EntryType, logEvent);
+            if (!string.IsNullOrEmpty(renderEntryType))
             {
                 //try parse, if fail,  determine auto
-
-                var value = RenderLogEvent(EntryType, logEvent);
-
-                EventLogEntryType eventLogEntryType;
-                if (EnumHelpers.TryParse(value, true, out eventLogEntryType))
+                if (ConversionHelpers.TryParseEnum(renderEntryType, out EventLogEntryType eventLogEntryType))
                 {
                     return eventLogEntryType;
                 }
+                InternalLogger.Warn("EventLogTarget(Name={0}): WriteEntry failed to parse EntryType={1}", Name, renderEntryType);
             }
 
             // determine auto
@@ -397,13 +403,11 @@ namespace NLog.Targets
         /// <param name="alwaysThrowError">always throw an Exception when there is an error</param>
         private void CreateEventSourceIfNeeded(string fixedSource, bool alwaysThrowError)
         {
-
             if (string.IsNullOrEmpty(fixedSource))
             {
                 InternalLogger.Debug("EventLogTarget(Name={0}): Skipping creation of event source because it contains layout renderers", Name);
                 //we can only create event sources if the source is fixed (no layout)
                 return;
-
             }
 
             // if we throw anywhere, we remain non-operational
@@ -437,11 +441,10 @@ namespace NLog.Targets
             catch (Exception exception)
             {
                 InternalLogger.Error(exception, "EventLogTarget(Name={0}): Error when connecting to EventLog.", Name);
-                if (alwaysThrowError || exception.MustBeRethrown())
+                if (alwaysThrowError || LogManager.ThrowExceptions)
                 {
                     throw;
                 }
-
             }
         }
     }
--- a/src/NLog/Targets/Target.cs
+++ b/src/NLog/Targets/Target.cs
@@ -709,6 +709,9 @@ namespace NLog.Targets
         /// <returns>String representing log event.</returns>
         protected string RenderLogEvent(Layout layout, LogEventInfo logEvent)
         {
+            if (layout == null || logEvent == null)
+                return null;    // Signal that input was wrong
+
             if (OptimizeBufferReuse)
             {
                 SimpleLayout simpleLayout = layout as SimpleLayout;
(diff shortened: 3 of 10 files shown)


304NotModified:
nice!

I think this should be in NLog 5?

snakefoot:
> I think this should be in NLog 5?

It is a seperate Nuget-package, not dependent on any specific version of NLog.

snakefoot:
Think people doing web-applications on the Windows-Platform would love to have this feature.

304NotModified:
Yes, great feature!

src/NLog/Targets/EventLogTarget.cs
```diff
@@ -270,11 +271,20 @@ protected override void Write(LogEventInfo logEvent)
...
-            short category = Category.RenderShort(logEvent, 0, "EventLogTarget.Category");
+            int eventId = 0;
+            if (!string.IsNullOrEmpty(renderEventId) && !int.TryParse(renderEventId, NumberStyles.Integer, CultureInfo.InvariantCulture, out eventId))
+            {
+                InternalLogger.Warn("{0} WriteEntry failed to parse EventId={1}", this, renderEventId);
+            }
 
-          
+            short category = 0;
+            if (!string.IsNullOrEmpty(renderCategory) && !short.TryParse(renderCategory, NumberStyles.Integer, CultureInfo.InvariantCulture, out category))
```
304NotModified: why did we remove the `RenderShort` and `RenderInt?` Because there aren't public?

src/NLog/Targets/EventLogTarget.cs
```diff
@@ -320,11 +330,14 @@ private EventLogEntryType GetEntryType(LogEventInfo logEvent)
             if (EntryType != null)
             {
                 //try parse, if fail,  determine auto
-
                 var value = RenderLogEvent(EntryType, logEvent);
 
                 EventLogEntryType eventLogEntryType;
+#if NETSTANDARD2_0
+                if (Enum.TryParse(value, true, out eventLogEntryType))
```
304NotModified: could we move this to the EnumHelpers?

src/NLog/Targets/EventLogTarget.cs
```diff
@@ -437,11 +448,10 @@ private void CreateEventSourceIfNeeded(string fixedSource, bool alwaysThrowError
             catch (Exception exception)
             {
                 InternalLogger.Error(exception, "Error when connecting to EventLog.");
-                if (alwaysThrowError || exception.MustBeRethrown())
+                if (alwaysThrowError || LogManager.ThrowExceptions)
```
304NotModified: this is a bugfix?

src/NLog/Targets/Target.cs
```diff
@@ -669,6 +669,9 @@ protected void MergeEventProperties(LogEventInfo logEvent)
         /// <returns>String representing log event.</returns>
         protected string RenderLogEvent(Layout layout, LogEventInfo logEvent)
         {
+            if (layout == null || logEvent == null)
+                return string.Empty;
```
304NotModified: I think it would be nice to write a warning (maybe debug level) to the internal logger in this case, what do you think?

304NotModified:
Meta data Looks good :)

image

src/NLog/Targets/EventLogTarget.cs
snakefoot: Not public

src/NLog/Targets/EventLogTarget.cs
snakefoot: Yes not public

src/NLog/Targets/EventLogTarget.cs
snakefoot: Not public

src/NLog/Targets/EventLogTarget.cs
304NotModified: the line below there is the EnumHelpers? ;)

src/NLog/Targets/Target.cs
snakefoot: Not sure I like messages per logevent.

src/NLog/Targets/EventLogTarget.cs
304NotModified: Proposal: or make public, or duplicate the code into this package :)

It's easier to maintain then

please share your thoughts

src/NLog/Targets/EventLogTarget.cs
snakefoot: The NetStandard2.0 assembly is not part of the Nlog.dll, it is a separate package and cannot access internal stuff (unless I open for it like with unit-tests).

src/NLog/Targets/EventLogTarget.cs
snakefoot: I was thinking everything would be changed, when NLog 6.0 comes and explodes everything into smaller pieces.

src/NLog/Targets/Target.cs
304NotModified: true, but not logging anything without any logs is also not that nice. Internal log Trace level otherwise?

src/NLog/Targets/EventLogTarget.cs
304NotModified: > cannot access internal stuff (unless I open for it like with unit-tests).

Maybe that isn't a bad idea, what do you think? Never done that with "public" assemblies, but guess it works?

snakefoot: I don't like replacing a hack  with another hack. The correct solution would be to make an official API. But I can copy-paste the code for this.

304NotModified: OK, please copy-paste then for these cases, thx!

snakefoot: Now performed copy-paste

src/NLog/Targets/Target.cs
snakefoot: Well some Layouts are expected to be blank/null, so I think the important layout logging should be handled by the individual targets.

src/NLog/Targets/EventLogTarget.cs
304NotModified: yep, and it's easier to structural fix it with NLog 6 if we copy-paste now the methods, agree?

src/NLog/Targets/Target.cs
304NotModified: OK, thanks for the input

src/NLog/Targets/EventLogTarget.cs
snakefoot: Have now replace internal LayoutHelpers with public ConversionHelpers.

src/NLog/Common/ConversionHelpers.cs
```diff
@@ -0,0 +1,138 @@
...
+    public static class ConversionHelpers
+    {
+        /// <summary>
+        /// Converts input string value into <see cref="System.Int16"/>
+        /// </summary>
+        /// <param name="inputValue">Input value</param>
+        /// <param name="resultValue">Output value</param>
+        /// <param name="defaultValue">Default value</param>
+        /// <returns>Returns failure if the input value could not be parsed</returns>
+        public static bool TryParseInt16(string inputValue, out short resultValue, short defaultValue = default(short))
```
304NotModified: is there a reason why we use int16 etc, while short is recommend by MS?

src/NLog/Common/ConversionHelpers.cs
```diff
@@ -0,0 +1,138 @@
...
+// 
+
+namespace NLog.Common
+{
+    using System.Globalization;
+
+    /// <summary>
+    /// String Conversion Helpers
+    /// </summary>
+    public static class ConversionHelpers
```
304NotModified: any idea why this isn't a git rename?

src/NLog/Targets/EventLogTarget.cs
304NotModified: also this one, copy code or it should be public

src/NLog/Common/ConversionHelpers.cs
304NotModified: probable because it was `render...` and now `tryparse...`

Could you explain why this is changed? IMO the `render...` was nice, also all rendering and conversing (and internal logging) would be handled consistently.

snakefoot:
Think the target should decide if critical error and if logging is needed


Sent from my Samsung device


304NotModified:
Imo Consistent internal logging is also important

The helpers should help and make rendering easier, including error handling.  In this pr they are tryparse wrappers.

snakefoot:
I prefer the flexibility. What if the input doesn't come from a Layout and LogEventInfo ? What if needing to log multiple parameters when conversion fails?


Sent from my Samsung device


304NotModified:
> I prefer the flexibility.

That's also possible, by not using the render helpers. Also that's not the case here?

I have introduced the helpers because internal logging was forgotten and that gives too less insights when something when wrong. 

I propose to copy paste the current internal helpers, and keep them internal. In the future we could discuss a public API in which we are both happy with :)

src/NLog/Common/ConversionHelpers.cs
snakefoot: Think the target should decide if critical error and if logging is needed

src/NLog/Targets/EventLogTarget.cs
snakefoot: Don't like to pollute with public extension methods on non-NLog-objects. Have now introduced a new public method `LogManager.MustBeRethrown(Exception)`

snakefoot:
> I propose to copy paste the current internal helpers, and keep them internal. In the future we could discuss a public API in which we are both happy with :)

Don't like copy-paste-code, so prefer to agree on some public API

304NotModified:
> Don't like copy-paste-code, so prefer to agree on some public API

I liked the "RenderInt" etc, and I think "TryParseInt32" is confusing because of:

- int32 while it returns a int
- we have already int.TryParse, so why could be confusing when to use one of them.
- it looks like layout.Render

Is the main problem of the RenderInt method that there is not control over the internal logging and maybe because of the parameter `layoutName`? 

Any proprosal to make a nice  RenderInt (etc) methodes, which looks like Render?

304NotModified:
another - more structural - option, is to create special `Layout` types, e.g. `IntLayout`

snakefoot:
> That's also possible, by not using the render helpers. Also that's not the case here?

I'm also fine with not using the render helpers, and wait until some one can dream up something you like

snakefoot:
Now changed the EventLog target to not use the `LayoutHelpers`

src/NLog/LogManager.cs
```diff
@@ -355,6 +355,16 @@ public static bool IsLoggingEnabled()
             return factory.IsLoggingEnabled();
         }
 
+        /// <summary>
+        /// Determines whether the exception must be rethrown and logs the error to the <see cref="InternalLogger"/>
+        /// </summary>
+        /// <param name="exception">The exception to check.</param>
+        /// <returns><c>true</c>if the <paramref name="exception"/> must be rethrown, <c>false</c> otherwise.</returns>
+        public static bool MustBeRethrown(Exception exception)
```
304NotModified: Todo check

src/NLog/Common/ConversionHelpers.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// <summary>
+        /// Converts input string value into <see cref="System.Enum"/>
+        /// </summary>
+        /// <param name="inputValue">Input value</param>
+        /// <param name="resultValue">Output value</param>
+        /// <param name="defaultValue">Default value</param>
+        /// <returns>Returns failure if the input value could not be parsed</returns>
+        public static bool TryParseEnum<TEnum>(string inputValue, out TEnum resultValue, TEnum defaultValue = default(TEnum)) where TEnum : struct
+        {
+            if (string.IsNullOrEmpty(inputValue))
```
304NotModified: maybe it's a good idea to move also `Internal.EnumHelpers.TryParse` to this file, so (internal or not), so the implementations aren't separated between files.

What do you think?

src/NLog/LogManager.cs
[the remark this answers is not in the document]
snakefoot: Then we should fix all MustBeRethrown, so they actually check against the active LoggingConfiguration that the target belongs to. Think this is outside the scope of this PR.

src/NLog/Common/ConversionHelpers.cs
snakefoot: No opinion. Can move the code if you like. Just like code-history.

src/NLog/LogManager.cs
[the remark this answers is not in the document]
304NotModified: any idea where is a good location, beside LogManager ;)

I try to keep the LogManager small and clean.

src/NLog/Common/ConversionHelpers.cs
304NotModified: please do, thanks!

src/NLog/LogManager.cs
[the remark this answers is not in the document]
snakefoot: Well the alternative solution is just to check `LogManager.ThrowExceptions` like I did initially, where you also complained.

src/NLog/Common/ConversionHelpers.cs
snakefoot: Done

src/NLog/LogManager.cs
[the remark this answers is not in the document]
304NotModified: Sure I like a public helper, but IMO the LogManager isn't the right place.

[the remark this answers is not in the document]
snakefoot: Removed and moved back to square one as before.

vlef:
This feature will be really useful. I'm wondering if this Nuget package will enable the EventLog target even if using NLog with ASP.NET Core (ie calling the ILoggingBuilder.UseNLog() in Program.cs)? Thank you for your work.

304NotModified:
> This feature will be really useful. I'm wondering if this Nuget package will enable the EventLog target even if using NLog with ASP.NET Core

Yes, but it's Windows only.

304NotModified:
I will release this hopefully today!

304NotModified:
merged!

snakefoot:
@vlef Has been released: https://www.nuget.org/packages/NLog.WindowsEventLog/

vlef:
Great, thanks guys. However, I keep getting "Target cannot be found: 'EventLog'" when it's parsing my nuget.config. Is there anything else I should be doing other than adding the new nuget package NLog.WindowsEventLog to the project?

snakefoot:
Must run netcore2. But maybe you need `add assembly` at the top of your NLog.config:

```xml
<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" >
  <extensions>
    <add assembly="NLog.WindowsEventLog" />
  </extensions>
  <targets>
```

idurston:
Adding that extension in the config got me past the Target cannot be found error...but now I keep getting:

2018-05-29 08:53:26.7090 Trace AsyncWrapper(Name=eventlog): Writing 1 events (Timer)
2018-05-29 08:53:26.7090 Trace AsyncWrapper(Name=webservice): Writing 1 events (Timer)
2018-05-29 08:53:26.9850 Error Error has been raised. Exception: System.TypeLoadException: Could not load type 'NLog.Common.ConversionHelpers' from assembly 'NLog, Version=4.0.0.0, Culture=neutral, PublicKeyToken=5120e14c03d0593c'.
   at NLog.Targets.EventLogTarget.GetEntryType(LogEventInfo logEvent)
   at NLog.Targets.EventLogTarget.Write(LogEventInfo logEvent)
   at NLog.Targets.Target.Write(AsyncLogEventInfo logEvent)

I updated NLog to 4.5.5.  Any help is appreciated...

danstur:
The `NLog.Common.ConversionHelpers` is not part of the released NLog 4.5.5 nuget package, the class is still `NLog.Internal.EnumHelpers` in the released sources.

This means the current package can not work. We'd need an NLog 4.5.6 from the current master to fix that problem.

snakefoot:
@304NotModified & @idurston  & @danstur Yes this PR was merged AFTER the NLog 4.5.5 release-cut (for some reason). I guess it needs to be re-releasesd with dependency on NLog 4.5.6

304NotModified:
I think I have to release all (NLog and this package) under 4.5.6, ins;t?

304NotModified:
PS: I've unlisted the current package.

snakefoot:
Could you delay the NLog 4.5.6 some hours ? as I would like to fix a bug in JsonLayout


Sent from my Samsung device


304NotModified:
4.5.6 is now on nuget.org: https://www.nuget.org/packages/NLog.WindowsEventLog

(currently indexing on nuget.org)

also a RTM of the dependency is released this evening: 

image

danstur:
I can confirm that using the 4.5.6 NuGets solves the problem.

304NotModified:
@danstur   Thanks for the confirmation!

PR #2550, opened 2018-01-27
DefaultJsonSerializer - Reflection should skip index-item-properties

snakefoot:


304NotModified:
good idea!

Could you please add an unit test for this case? thx!

snakefoot:
> Could you please add an unit test for this case?

Added unit test, with a `StrangeProperty`

304NotModified:
Thanks!

PR #2551, opened 2018-01-28
Blackhole LoggingRule without targets

snakefoot:
Is faster than constantly allocating LogEvents for Null-target.


snakefoot:
Maybe consider updating Wiki:

https://github.com/NLog/NLog/wiki/Filtering-log-messages

304NotModified:
great improvement!

snakefoot:
@304NotModified Updated all the wiki-pages. Discovered that blackhole-logic also worked by just not specifying `writeTo`

304NotModified:
Great! Thanks!

PR #2554, opened 2018-01-30
LogBuilder - Check level before allocation of Properties-dictionary

snakefoot:
And CallSiteInformation-object


304NotModified:
thx again!

PR #2555, opened 2018-01-30
JsonLayout - Use capturetype and added MaxRecursionLimit and set default to 0

snakefoot:
update

default 0: 
  Will enumerate simple collections but not object properties. 

---

To avoid breaking change in NLog 4.5, then the JsonLayout should not by default perform aggressive serialization of object properties. This will avoid the Entity-framework pitfall: https://stackoverflow.com/questions/48516514

Added MaxRecursionLimit as new setting to JsonLayout. It will allow one to increase the recursion to also include object properties. To avoid breaking changes then default MaxRecursionLimit = 0 (No reflection on object properties). Maybe in NLog 5.0 it can be changed to MaxRecursionLimit = 1 (Limited reflection of object properties).

Implemented a special exception for MessageTemplate parameters where capturetype has been specified. If capture-type `@` tells us it is safe, then it will allow "full" reflection.

### DIFF
--- a/src/NLog/Layouts/JsonLayout.cs
+++ b/src/NLog/Layouts/JsonLayout.cs
@@ -46,18 +46,45 @@ namespace NLog.Layouts
     [AppDomainFixedOutput]
     public class JsonLayout : Layout
     {
-        private IJsonConverter JsonConverter
+        private LimitRecursionJsonConvert JsonConverter
         {
-            get => _jsonConverter ?? (_jsonConverter = ConfigurationItemFactory.Default.JsonConverter);
+            get => _jsonConverter ?? (_jsonConverter = new LimitRecursionJsonConvert(MaxRecursionLimit, ConfigurationItemFactory.Default.JsonConverter));
             set => _jsonConverter = value;
         }
-        private IJsonConverter _jsonConverter = null;
+        private LimitRecursionJsonConvert _jsonConverter = null;
         private IValueFormatter ValueFormatter
         {
             get => _valueFormatter ?? (_valueFormatter = ConfigurationItemFactory.Default.ValueFormatter);
             set => _valueFormatter = value;
         }
         private IValueFormatter _valueFormatter = null;
+
+        class LimitRecursionJsonConvert : IJsonConverter
+        {
+            readonly IJsonConverter _converter;
+            readonly Targets.DefaultJsonSerializer _serializer;
+            readonly Targets.JsonSerializeOptions _serializerOptions;
+
+            public LimitRecursionJsonConvert(int maxRecursionLimit, IJsonConverter converter)
+            {
+                _converter = converter;
+                _serializer = converter as Targets.DefaultJsonSerializer;
+                _serializerOptions = new Targets.JsonSerializeOptions() { MaxRecursionLimit = Math.Max(0, maxRecursionLimit) };
+            }
+
+            public bool SerializeObject(object value, StringBuilder builder)
+            {
+                if (_serializer != null)
+                    return _serializer.SerializeObject(value, builder, _serializerOptions);
+                else
+                    return _converter.SerializeObject(value, builder);
+            }
+
+            public bool SerializeObjectNoLimit(object value, StringBuilder builder)
+            {
+                return _converter.SerializeObject(value, builder);
+            }
+        }
 
         /// <summary>
         /// Initializes a new instance of the <see cref="JsonLayout"/> class.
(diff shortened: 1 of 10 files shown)


snakefoot:
@304NotModified Merged with the rename of ValueSerializer to ValueFormatter

snakefoot:
Have now added a new property to `JsonLayout` called `MaxRecursionLimit`

The default value is 0, which means it will not attempt property-reflection by default, but only when the message-template says it is safe. I have changed so collections at initial level will not receive penalty, so when using default value 0, then it will allow enumeration of collections (but not collections of collections).

When value is 1 then it will allow very limited object property reflection.

When value is 10 then it will use the default recursion level of the NLog JsonConverter.

src/NLog/Layouts/JsonLayout.cs
```diff
@@ -113,6 +147,11 @@ public JsonLayout()
         public ISet<string> ExcludeProperties { get; set; }
 #endif
 
+        /// <summary>
+        /// How far should the JSON serializer follow object references before backing off
+        /// </summary>
+        public int MaxRecursionLimit { get; set; }
```
304NotModified: is no max, -1 or should we make it nullable?

src/NLog/MessageTemplates/CaptureType.cs
```diff
@@ -38,6 +38,10 @@ namespace NLog.MessageTemplates
     /// </summary>
     public enum CaptureType : byte
     {
+        /// <summary>
+        /// Not decided
+        /// </summary>
+        Unknown,
```
304NotModified: what do you think about a  nullable CaptureType? Or does this have to much impact?

src/NLog/Layouts/JsonLayout.cs
snakefoot: I guess I can make it unassigned by having a nullable backing field. But until NLog 5.0, then it must have default value 0, or else it will be a breaking change.

src/NLog/MessageTemplates/CaptureType.cs
snakefoot: I prefer to have Unknown

src/NLog/Layouts/JsonLayout.cs
304NotModified: it;s a new property so not a breaking change?

snakefoot: Yes it is a new property.

snakefoot:
@304NotModified Added some unit tests to show how MaxRecursionLimit=0 and MaxRecursionLimit=1 works

304NotModified:
Close/reopen to trigger ci

snakefoot:
Regarding the naming of MaxRecursionLimit, then it looks similar to Microsoft JavaScriptSerializer:

https://msdn.microsoft.com/en-us/library/system.web.script.serialization.javascriptserializer.recursionlimit.aspx

Json.NET has MaxDepth for for deserializing, there is no functionality when serializing:

https://www.newtonsoft.com/json/help/html/MaxDepth.htm

In serilog the same setting is called MaximumDepth:

src/NLog/Layouts/JsonLayout.cs
304NotModified: Default 0 is fine

But I think we should change the type to nullable int, so we could disable the limit clearly. Agree?

snakefoot: I think the limit should always be in place. Not sure if you have recognized, that the MaxRecursionLimit overrides the hardcoded value of 10 in NLog 4.5 BETA (And overrides the hardcoded value of 0 in NLog 4.4). The limit is there to protect the users from dumping the entire world into a single logevent.

304NotModified: > What behavior do you want when having assigned MaxRecursionLimit to null ?

no limit.

snakefoot:
Think int.MaxValue will do just fine. A value higher than 10 will be a shotgun ready to shoot yourself. Please describe a use case



Sent from my Samsung device


src/NLog/Layouts/JsonLayout.cs
snakefoot: Think int.MaxValue will do just fine. A value higher than 10 will be a shotgun ready to shoot yourself. Actually considered an enum-value called NoReflection (=0), SingleReflection (=1), MaxReflection (=10), but thought that an integer would be easier. Now I see with this discussion, then it might have been a mistake.

304NotModified:
OK thanks, have merged it

snakefoot:
Updated documentation: https://github.com/NLog/NLog/wiki/JsonLayout

304NotModified:
thx!

PR #2560, opened 2018-02-02
Network/NLogViewer/Chainsaw Target - Enabled OptimizeBufferReuse by default, but not for sub classes

snakefoot:


src/NLog/Targets/ChainsawTarget.cs
```diff
@@ -67,6 +67,7 @@ public class ChainsawTarget : NLogViewerTarget
         public ChainsawTarget()
         {
             IncludeNLogData = false;
+            OptimizeBufferReuse = GetType() == typeof(ChainsawTarget);
```
304NotModified: Why not just true? Because of custom sub classes?

snakefoot: Yes minimum amount of surprises possible.

304NotModified: Please add this as code comments, thanks in advance!

PR #2563, opened 2018-02-03
WebService-Target fails internally with PlatformNotSupportedException on NetCore

snakefoot:
WebRequest.GetSystemWebProxy fails when attempting to use Default-proxy on NetCore

See also https://github.com/dotnet/corefx/issues/7037

Changed default value for ProxyType to be NoProxy on NetCore (Avoid internal exception on every request)


304NotModified:
thx!

PR #2568, opened 2018-02-10
WebServiceTarget - Rollback added Group-Layout

### LINKED ISSUE #2567: Undo pollution of MethodCallParameter
Think it would be best to remove these parameters from `MethodCallParameter`

```
        /// <summary>
        /// Parameter can combine multiple LogEvents into a single parameter value
        /// </summary>
        public bool EnableGroupLayout { get; set; }

        /// <summary>
        /// Group Header when combining multiple LogEvents into a single parameter value (<see cref="EnableGroupLayout"/>)
        /// </summary>
        public Layout GroupHeaderLayout { get; set; }

        /// <summary>
        /// Group Item Separator when combining multiple LogEvents into a single parameter value (<see cref="EnableGroupLayout"/>)
        /// </summary>
        public Layout GroupItemSeparatorLayout { get; set; }

        /// <summary>
        /// Group Footer when combining multiple LogEvents into a single parameter value (<see cref="EnableGroupLayout"/>)
        /// </summary>
        public Layout GroupFooterLayout { get; set; }
```

They were introduced with #2223, but I'm not sure I like this change after reading my code again. Prefer to keep `MethodCallParameter` simple without being involved in this batching/group-logic.

snakefoot:
Partially resolves  #2567 by reverting #2223. Better solution must come later.

Sorry about the noise. And reopening  #1952

### DIFF
--- a/src/NLog/Targets/MethodCallParameter.cs
+++ b/src/NLog/Targets/MethodCallParameter.cs
@@ -34,8 +34,8 @@
 namespace NLog.Targets
 {
     using System;
-    using Config;
-    using Layouts;
+    using NLog.Config;
+    using NLog.Layouts;
 
     /// <summary>
     /// A parameter to MethodCall.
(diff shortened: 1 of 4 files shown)


304NotModified:
thx!

Issue #2557, opened 2018-02-01
JsonLayout ignores capturing directives in structured log events


nblumhardt:
**Type** (choose one):

- Bug

**NLog version**: (e.g. 4.5.0-rc04)

**Platform**: .NET Core 

**Current NLog config** (xml or C#, if relevant)

```csharp
using System;
using NLog;
using NLog.Targets;
using NLog.Config;
using NLog.Layouts;

namespace nlog_json
{
    class Program
    {
        static void Main(string[] args)
        {
            var config = new LoggingConfiguration();
            var consoleTarget = new ColoredConsoleTarget();
            config.AddTarget("console", consoleTarget);
            var json = new JsonLayout();

            // Include the event properties in the JSON
            json.IncludeAllProperties = true;

            consoleTarget.Layout = json;
            config.LoggingRules.Add(new LoggingRule("*", LogLevel.Debug, consoleTarget));
            LogManager.Configuration = config;

            var logger = LogManager.GetLogger("Example");
            logger.Info("Hello, {Person}", new { Name = "nblumhardt" });
        }
    }
}
```

**Expected**

The `Person` property is serialized using its `ToString()` representation:

```json
{"Person": "{ Name = nblumhardt }"}
```

**Actual**

The property value is serialized:

```json
{ "Person": {"Name":"nblumhardt"} }
```

This is unexpected, because message templates use `@` to control serialization. It's an opt-in; if the property name is not prefixed with `@`, then a string is captured just like in .NET format strings.

If we're going to do structured logging with `JsonLayout`, we shouldn't risk accidentally serializing-the-world. (The canonical footgun here would be if `Person` captured an entity from Entity Framework - these `ToString()` fine just like everything else, but tend to serialize very poorly.)

snakefoot:
It works as expected. JsonLayout will by default serialize everything as Json.

When IncludeAllProperties = true, then it will serialize all property values into Json.

JsonLayout doesn't have any clue about MessageTemplate-syntax.

nblumhardt:
Thanks for the reply, @snakefoot, appreciate you taking the time to answer.

What's the recommended approach for serializing structured log events? The typical use case for structured logging is generally _templated properties -> JSON -> (file or data store)_. This behavior will make it impractical to use templated properties for structured logging, which seems to defeat the whole purpose of including the feature in 4.5.

* `IncludeProperties = true` - I can't do structured logging because of the risk I'll end up serializing things that shouldn't be serialized
[...]

PR #2576, opened 2018-02-13
Fix possible infinite loop in message template parser + better handling incorrect templates

304NotModified:
See https://github.com/NLog/NLog/issues/2557#issuecomment-365200102

### DIFF
--- a/src/NLog/MessageTemplates/MessageTemplateParameters.cs
+++ b/src/NLog/MessageTemplates/MessageTemplateParameters.cs
@@ -114,7 +114,7 @@ namespace NLog.MessageTemplates
 
             try
             {
-                int holeIndex = 0;
+                short holeIndex = 0;
                 TemplateEnumerator templateEnumerator = new TemplateEnumerator(template);
                 while (templateEnumerator.MoveNext())
                 {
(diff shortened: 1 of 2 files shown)


304NotModified:
@snakefoot thanks for reviewing, I will react in this PR to make the discussions a bit more separated.

> No need to use short (16 bit) since new processors prefer 32 bit and 64 bit variables.

This is a side effect `hole.Index`, which is already a 'short' (see https://github.com/NLog/NLog/blob/19de4feb0a6f1292e9d0c69da5c75b73efda3ea1/src/NLog/MessageTemplates/Hole.cs#L66)

> Also no need to perform "safe" lookup, as invalid templates should trigger exceptions (just not unhandled exceptions)

What is the expected output of  `MessageTemplateParameters.Count` with `" {3} {4} {9} {8} {5} {6} {7}"` ? It was 2, which was really unexpected for me. Also if we could make (partly) a good result, that really better then no result. e.g. in  `" {3} {4} {9} {8} {5} {6} {7}"` we could maybe render 6 out of 7, which is better then 2 (and 0).

snakefoot:
I prefer strict mode, but I can see Serilog has undocumented behavior when using invalid templates. Mixing strict with loose.

P.S. Please don't make these constructors internal. See #2581 (Maybe remove one of the struct constructors, but the one with CaptureType is used by NLog.Extensions)


Sent from my Samsung device


304NotModified:
> I prefer strict mode, but I can see Serilog has undocumented behavior when using invalid templates. Mixing strict with loose.

I prefer strict mode when:

- it's a compile error (not run-time)
- and/or it could be easily fixed (read: without recompile)
- if the non-strict mode has undefined behavior.

Issue #2552, opened 2018-01-28
 constructor protection level to internal for `MessageTemplateParameters` and `MessageTemplateParameter`


304NotModified:
constructor protection level to internal for

- MessageTemplateParameters
- MessageTemplateParameter

see https://github.com/NLog/NLog/pull/2544

B3nCr:
Is there anything more to this or could a new contributor just fork and make these changes?

snakefoot:
@B3nCr This is partially resolved by #2544 (Fixes `MessageTemplateParameters`) . Not sure how important it is to hide the constructors on `MessageTemplateParameter`

PR #2581, opened 2018-02-19
MessageTemplateParameter(s)  ctors to internal

304NotModified:
fixes https://github.com/NLog/NLog/issues/2552

+ small c#7  change

### DIFF
--- a/src/NLog/MessageTemplates/MessageTemplateParameter.cs
+++ b/src/NLog/MessageTemplates/MessageTemplateParameter.cs
@@ -89,8 +89,7 @@ namespace NLog.MessageTemplates
                     default:
                         if (Name?.Length >= 1 && Name[0] >= '0' && Name[0] <= '9')
                         {
-                            int parameterIndex;
-                            if (int.TryParse(Name, out parameterIndex))
+                            if (int.TryParse(Name, out var parameterIndex))
                                 return parameterIndex;
                         }
                         return null;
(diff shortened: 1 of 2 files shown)


304NotModified:
@snakefoot decline this and wontfix for #2552?

snakefoot:
I was thinking to allow use of MessageTemplateParameters constructor to access the internal parser. The MessageTemplateParameter struct is used by NLog.Extensions to capture Microsoft Extension Logging Message Properties. But we don't need two constructors, only the one with CaptureType



Sent from my Samsung device


304NotModified:
So for both classes we need public ctors?

304NotModified:
> I was thinking to allow use of MessageTemplateParameters constructor to access the internal parser.

No really a fan of this magic. 

updated it: MessageTemplateParameter full ctor is still public

MessageTemplateParameters is internal ctor. IMO the ctor of MessageTemplateParameters do maybe to much parsing and so it's a good candidate to keep it internal. 

Let me know if you agree 

We can always open the ctor if we need it in the future.

304NotModified:
PS I will try to create rc6 tonight/tomorrow. I hope that the latest RC before the RTM

@snakefoot if your still on holiday, have a good holiday :)

snakefoot:
Yes. I think so


Sent from my Samsung device


snakefoot:
Will also work for me


Sent from my Samsung device

PR #2584, opened 2018-02-19
ColoredConsoleTarget.cs: Whitespace cleanups

perlun:


304NotModified:
I guess you can't change the whitespace of the header, the exact text is enforced on appveyor.

perlun:
@304NotModified There we go, header part removed.

(We should probably fix that also, but that's a much bigger task since it involves changing all C# files in the whole project... so not in scope for this petty PR.)

304NotModified:
thx

Issue #2589, opened 2018-02-21
Add Layout property to WebServiceTarget


mniak:
**Type**: Feature request
**NLog version**: any
**Platform**: any

I suggest creating a `HttpTarget` or enhancing the `WebServiceTarget` to enable specifing a layout for the entire body of a HTTP request.

Example (desired):
```xml
<target type="WebService" name="http" url="http://localhost:9093">
  <protocol>JsonPost</protocol>
  <layout xsi:type="JsonLayout" includeAllProperties="true"> <!-- This property (layout) cannot be specified today -->
    <attribute name="timestamp"   layout="${longdate}"                 />
    <attribute name="level"       layout="${level}"                    />
    <attribute name="logger"      layout="${logger}"                   />
    <attribute name="message"     layout="${message}"                  />
    <attribute name="exception"   layout="${exception:format=tostring}"/>
    <attribute name="thread_name" layout="${threadid}"                 />
    <attribute name="thread_id"   layout="${threadname}"               />
  </layout>
</target>
```

This shows very useful when using structured logging.

I'll try to create a prototype and send a pull request.

snakefoot:
Duplicate of #1905

snakefoot:
Think a simple "hack" would be for `JsonPost` to recognize that when the a single parameter with empty-name, then it doesn't have to wrap the JsonResult with a Parameter-name.

The problem with WebServiceTarget is that it tries to be everything. Notice how `XmlPost` is missing the ability to provide a root-name (Mainly because the NLog-core is missing a XmlLayout)

snakefoot:
@mniak Made an example of my suggestion here: #2590 (Maybe you can dream up a unit-test?)

snakefoot:
Maybe you can do a manual test with this custom build:

https://ci.appveyor.com/project/nlog/nlog/build/4.5.0-beta7219/artifacts

Notice the use of nameless parameter:

```
<target type="WebService" name="http" url="http://localhost:9093">
  <protocol>JsonPost</protocol>
  <parameter name="">
  <layout xsi:type="JsonLayout" includeAllProperties="true">
    <attribute name="timestamp"   layout="${longdate}"                 />
    <attribute name="level"       layout="${level}"                    />
    <attribute name="logger"      layout="${logger}"                   />
    <attribute name="message"     layout="${message}"                  />
    <attribute name="exception"   layout="${exception:format=tostring}"/>
    <attribute name="thread_name" layout="${threadid}"                 />
[...]

Issue #1905, opened 2017-01-13
Allow Layouts in WebService Target


304NotModified:
**Type** (choose one):

- Feature request

**NLog version**: 4.4.1

Allow defining layout for webservice. In this case user will have full control of message which will be send to remote server. For example JsonLayout can be used.

```xml
<nlog>
    <target name="logs-service" xsi:type="AsyncWrapper">
      <target xsi:type="WebService" url="http://localhost:5002" protocol="JsonPost">
        <layout xsi:type="JsonLayout">
          <attribute name='logger' layout='${logger}' />
          <attribute name='level' layout='${level}'/>
          <attribute name="message" layout="${message}" />
          <attribute name='parameters' layout='${custom-parameters}' encode="false"/>
         </layout>
      </target>
    </target>
  <rules>
  </rules>
</nlog>
```

as suggested at https://github.com/NLog/NLog/issues/1900

yotamLP:
Hi can you please give me an example on how to implement this custom parameters. I have a web API which I want to send a single json variable. 

do I need to override the webservice method?

304NotModified:
Do you mean how to implement this feature?

yotamLP:
yes

snakefoot:
The WebService-target needs to have a parameter to provide the payload, because it supports both HttpGet and HttpPost.

Just like there is XmlRoot-parameter to wrap all parameters inside a single XML-tag, then one could create a new bool parameter called SingleParameterRoot. That promotes the first parameter to be root when doing HttpPost. Whether it is Json or Xml. When SingleParameterRoot=true then it will validate that only a single parameter has been configured.

snakefoot:
Could also consider leaving the WebService-target behind, and instead create a new target based on HttpClient (Net45-only). Then one doesn't have to fight with the backward compatibility dragons and can use recommended high performance APIs.

ParallelTask:
One of my projects requires the exact thing, so I have created NLog.RestTarget that solves the exact problem. Please find the reference https://github.com/ParallelTask/Nlog.RestTarget

snakefoot:
There is now a build that supports this:

https://www.nuget.org/packages/NLog/4.5.0-rc07

Notice the use of a single nameless parameter:

```xml
<target xsi:type="WebService" url="http://localhost:5002" protocol="JsonPost">
  <parameter name="">
        <layout xsi:type="JsonLayout">
          <attribute name='logger' layout='${logger}' />
[...]

PR #2590, opened 2018-02-21
WebServiceTarget - JsonPost with support for single nameless parameter

snakefoot:
Small hack to support #1905 and #2589

### DIFF
--- a/src/NLog/Targets/WebServiceTarget.cs
+++ b/src/NLog/Targets/WebServiceTarget.cs
@@ -323,7 +323,7 @@ namespace NLog.Targets
             DoInvoke(parameters, continuation, request, begin, getStream);
         }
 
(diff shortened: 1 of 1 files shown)


src/NLog/Targets/WebServiceTarget.cs
```diff
@@ -733,35 +728,13 @@ protected HttpPostTextFormatterBase(WebServiceTarget target) : base(target)
                 _encodingPreamble = target.Encoding.GetPreamble();
             }
 
-            protected abstract string Separator { get; }
-
-            protected virtual void BeginFormattedMessage(StringBuilder builder)
-            {
-            }
-
-            protected abstract void AppendFormattedParameter(StringBuilder builder, MethodCallParameter parameter, object value);
```
304NotModified: is this a breaking changes? (removing the abstract methods)

snakefoot: Nope those are private classes inside WebServiceTarget

304NotModified:
Ok thanks!

Issue #2526, opened 2018-01-20
Make MDC and MDLC Set method using friendly


MikeFH:
**Type** :
- Feature request
 
**Why do we need it?**
This would make the MDC and MDLC usage easier when the goal is to capture scoped data.
For example :
```csharp
public void DoSomethingWithProduct(Product p) {
	using (MappedDiagnosticContext.Set("productId", p.Id)) {
		//do something	
	}
}
```
This is already present in NDC and NDLC.

I'am not sure about the performance inpact such a feature would have but if the returned `IDisposable` is a struct, maybe it wouldn't impact current users too much.

304NotModified:
> I'am not sure about the performance inpact such a feature would have but if the returned IDisposable is a struct, maybe it wouldn't impact current users too much.

I don't think this will hurt performance


PR would be nice !

304NotModified:
:) fixed, thanks!

MikeFH:
Thanks !

MikeFH:
@304NotModified Should I add something in the doc regarding the new methods ?

304NotModified:
Thanks for the suggestion. Where would you like to add it?

It's anyway already in the API docs, but more exposure would be nice.

MikeFH:
I thought of adding something along the lines of 
> The SetScoped method returns an IDisposable that removes the added item when disposed. It can be used in conjunction with the using statement to limit the scope during which the item will be present in the context.

with a simple code example in those pages
https://github.com/NLog/NLog/wiki/MDLC-Layout-Renderer
https://github.com/NLog/NLog/wiki/MDC-Layout-Renderer

But I totally forgot about the API docs so, I agree, the need is limited.

304NotModified:
Yeah would be nice to add to the mdc and mdlc pages! please do :+1:

MikeFH:
Alright, done !

304NotModified:
thanks!!

PR #2592, opened 2018-02-22
Make Set methods of MDC and MDLC return IDisposable

MikeFH:
Attempt at implementing #2526

### DIFF
--- a/src/NLog/MappedDiagnosticsContext.cs
+++ b/src/NLog/MappedDiagnosticsContext.cs
@@ -53,6 +53,29 @@ namespace NLog
         private static readonly IDictionary<string, object> EmptyDefaultDictionary = new SortHelpers.ReadOnlySingleBucketDictionary<string, object>();
 
         /// <summary>
+        /// 
+        /// </summary>
+        private class ItemRemover : IDisposable
+        {
+            private readonly string _item;
+            private bool _disposed = false;
+
+            public ItemRemover(string item)
+            {
+                this._item = item;
+            }
+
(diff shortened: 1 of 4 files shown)


304NotModified:
Cool! But unfortunately this is a breaking change. 

So proposal to add a new method, `Scope` or something like that?

MikeFH:
Got it.
I guess you can close this PR, I'll open a new one with another solution. I don't know when though, so if someone wants to look at it, it's cool.

snakefoot:
You could also call the method `Push` or `PushProperty`:

https://github.com/serilog/serilog/blob/3cd376c8c75ed49ab307c50cd80beaa06cbaecb7/src/Serilog/Context/LogContext.cs#L77

304NotModified:
> guess you can close this PR, I'll open a new one with another solution

Just push new commits on your branch, we can squash them later.

MikeFH:
I added a new method. It is called `Push` as suggested, but maybe it is a bit misleading since it might imply that the user works works on a stack (like in serilog) and that's not what we have here.

304NotModified:
What do you think about `SetScoped`?

MikeFH:
`SetScoped` it is. 
I can still rename it if someone comes up with a better idea.

304NotModified:
> Build seems to be ok on my appveyor and travis accounts.

Tests are sometime a bit unstable :(

Will restart the CI

304NotModified:
Looks great! merged!

Thanks!

PR #2595, opened 2018-02-24
IsSafeToDeferFormatting - Convert.GetTypeCode is faster and better

snakefoot:
Than GetType().IsPrimitive()

| NullLogger  | 50.000.000 |   64 |    3 |       1 |       1 |

| This PR (32 bit)   |    16.497 | 3.030.675 |   0 |   0 | 721 |   14.359 |    7.248,2 |
| Master (32 bit)   |    17.514 | 2.854.764 |   0 |   0 | 726 |   16.203 |    7.248,2 |
| This PR (64 bit)    |    15.437 | 3.238.959 |   0 |   0 | 489 |   14.468 |   14.114,7 |
| Master (64 bit)    |    15.923 | 3.139.958 |   0 |   0 | 481 |   15.203 |   14.114,7 |

### DIFF
--- a/src/NLog/LogEventInfo.cs
+++ b/src/NLog/LogEventInfo.cs
@@ -607,12 +607,7 @@ namespace NLog
 
         private static bool IsSafeToDeferFormatting(object value)
         {
-            if (value == null)
-            {
-                return true;
-            }
-
-            return value.GetType().IsPrimitive() || (value is string);
+            return value == null || Convert.GetTypeCode(value) != TypeCode.Object;
         }
 
         private static string GetStringFormatMessageFormatter(LogEventInfo logEvent)


304NotModified:
Are we 100% sure it's the same? Don't expect a performance improvement then.

snakefoot:
> Are we 100% sure it's the same? Don't expect a performance improvement then.

No it is not the same, since it is better and faster. Though it will have a side-effect incase some has created custom type that inherits from IConvertible. But only "primitive struct types" should do that.

This the source  code of Convert.GetTypeCode:

```
        public static TypeCode GetTypeCode(object value) {
            if (value == null) return TypeCode.Empty;
            IConvertible temp = value as IConvertible;
            if (temp != null)
            {
                return temp.GetTypeCode();
            }
            return TypeCode.Object;
        }
```

There is a lot of internal compiler optimization when it comes to GetType() and IsPrimitive(), but at the end it finds an internal typecode, and checks if it is one the simple types specified here:

https://msdn.microsoft.com/en-us/library/system.type.isprimitive(v=vs.110).aspx

Issue #2440, opened 2017-12-02
Merge UWP into NLog 4.5  - drop NLog 5


304NotModified:
I think this is double.

Merge UWP target (UWP branch) to NLog 4.5

Then we could drop NLog 5 for now, as NLog 4.5 supports .NET Standard 1.5 (NLog 5 - .NET standard 1.3)

With loosing .NET Standard 1.3+1.4, we won't loose any platforms?

See https://blogs.msdn.microsoft.com/dotnet/2016/09/26/introducing-net-standard/

![image](https://user-images.githubusercontent.com/5808377/33509568-02ab3a86-d703-11e7-9c77-66f03f4600db.png)

304NotModified:
fixed by #2441

mfeingol:
Hi.

> With loosing .NET Standard 1.3+1.4, we won't loose any platforms?

By losing .NET Standard 1.4, you've lost anyone using NLog in a .NET Standard assembly used by any UWP app targeting the Creator's Update and below. As that's my case, I'll be sticking to your 5.x pre-release versions until you can hopefully fix that.

Thanks!

304NotModified:
@mfeingol time to test a new rc with .net standard 1.3 support?

mfeingol:
Absolutely!

304NotModified:
cool! Will release within a few days!

304NotModified:
@mfeingol rc7 is in the build/release queue. Will be published in a few hours

Would be great if you could confirm soon that this will fix your case :)

304NotModified:
Thanks for the feedback!

PR #2597, opened 2018-02-27
Replace WINDOWS_UWP with NETSTANDARD1_3 to support UWP10 shared libraries

snakefoot:
Right now NLog cannot be used by any shared NetStandard14-library that wants to support UWP10. See also #2440

I guess the easy solution would be isolating the logger-interface into a separate assembly, but that is for NLog 6.0.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -12,7 +12,7 @@ $versionProduct = $versionPrefix;
 if (-Not $versionSuffix.Equals(""))
 	{ $versionProduct = $versionProduct + "-" + $versionSuffix }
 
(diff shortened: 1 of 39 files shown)


304NotModified:
do we need a new RC for this, or just add it to the RTM?

snakefoot:
First need to fix this PR to handle the changes in #2599

Sent from my Samsung device


304NotModified:
OK, let me know when I can merge this one, thanks!

snakefoot:
@304NotModified It has been fixed.

304NotModified:
:+1:

snakefoot:
@304NotModified And the build-server have become stable enough to complete the build.

304NotModified:
Merged, thanks!

Think a rc7 is wise for this change

snakefoot:
> Think a rc7 is wise for this change

Requires that you find some UWP developers willing to test it, since everyone else is using net45 or netstandard20

Issue #2598, opened 2018-02-27
NLog 4.4 cannot handle new ReadOnlySpan implicit operator on string


snakefoot:
Spun up a quick docker container as smoke test:

```
    root@8391e3c12c9c:/# mono --trace=exceptions /opt/NzbDrone/NzbDrone.exe
    [0x7f0e0fc97740:] EXCEPTION handling: System.ArgumentException: Object of type 'System.ReadOnlySpan`1[System.Char]' cannot be converted to type 'System.String'.
    [0x7f0e0fc97740:] EXCEPTION handling: NLog.NLogConfigurationException: Error when setting property 'Format' on Layout Renderer: ${exception}
```

Not sure what to make of it, but wild guess some newer dotnet core code in mono broke backward compat. Can't look into it further today, so yeah, you might want to downgrade to the 5.8 snapshot.

https://www.reddit.com/r/sonarr/comments/80g1e2/broken_after_upgrade_to_mono_510_stable/

Think this is causing the issue:

```c#
        [MethodImpl(MethodImplOptions.AggressiveInlining)]
        public static implicit operator ReadOnlySpan<char>(string value) =>
value != null ? new ReadOnlySpan<char>(ref value.GetRawStringData(), value.Length) : default;
```

https://github.com/dotnet/coreclr/blob/9b0a9fd623c8448af4619acb30b0b469aa5dd6b2/src/mscorlib/src/System/String.cs#L701

Which `TryImplicitConversion` finds and then attempts to use. NLog 4.5 doesn't react to implicit conversions on System-types.

snakefoot:
Adding the following lines to the top of `TryImplicitConversion` will fix the issue:

```c#
        private static bool TryImplicitConversion(Type resultType, string value, out object result)
        {
            try
            {
                if (Convert.GetTypeCode(resultType) != TypeCode.Object)
                {
                    result = null;
                    return false;
                }
```

304NotModified:
great work!

304NotModified:
NLog 4.5 -> #2599
NLog 4.4.14 -> #2600

PR #2599, opened 2018-02-27
TryImplicitConversion should only check object types

snakefoot:
TryImplicitConversion should only check object types

Not string and primitive types. Nicer filtering of non-object-types for fixing #2598 

Same solution should probably be implemented for NLog 4.4.13 (Unless NLog 4.5 is done today)

### DIFF
--- a/src/NLog/Internal/PropertyHelper.cs
+++ b/src/NLog/Internal/PropertyHelper.cs
@@ -88,8 +88,8 @@ namespace NLog.Internal
 
                 if (!(TryNLogSpecificConversion(propertyType, value, out newValue, configurationItemFactory)
                     || TryGetEnumValue(propertyType, value, out newValue, true)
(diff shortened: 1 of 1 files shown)


304NotModified:
> (Unless NLog 4.5 is done today)

No over-rushing things, so #2600 :)

304NotModified:
do we need a rc7 for this?

snakefoot:
The NLog 4.5 has protection. I just improved it a little with this PR. It is not affected by the string-conversion-issue.

304NotModified:
ah cool!

304NotModified:
Then I will revert the title :)

304NotModified:
duhh, the title "NLog 4.4 cannot handle ..."  was maybe already clear that it was not a 4.5 prob. 

Long days :P

Issue #2579, opened 2018-02-16
Allow to log ReflectedType instead of DeclaredType in $callsite when logging from base class


SylwesterZarebski:
**Type**: Feature request

**NLog version**: 4.4.11

**Platform**: .Net 4.5

**Current NLog config**: I'm using `${callsite:className=true:methodName=true}`

**Description**:  
I have base class which log some information, but currently $callsite logs DeclaredType and i need to log derived class (ReflectedType).

**Example**:
```
public abstract class A
{
    protected Logger Logger { get; } = null;
    public A()
    {
        Logger = LogManager.GetCurrentClassLogger();
    }
    public void BaseMethod()
    {
        Logger.Trace("Run BaseMethod()");
        ... some other code ...
    }
}
public sealed class B : A
{
    public void DerivedMethod()
    {
         BaseMethod();
    }
}
```

Log has `A:BaseMethod` and i need `B:BaseMethod`, because i have very many (about 20) derived classes and base class has all common code.
Above code is an example, current one is way more complicated with events etc.

snakefoot:
@SylwesterZarebski Have you actually tested that this is possible?

Looking at the documentation then it doesn't seem possible:

> The method that is currently executing may be inherited from a base class, although it is called in a derived class. In this case, the ReflectedType property of the MethodBase object that is returned by GetMethod identifies the base class, **_not the derived class_**.

https://msdn.microsoft.com/en-us/library/system.diagnostics.stackframe.getmethod.aspx

SylwesterZarebski:
I did test simple case using GetMethod("Method").ReflectedType and it works, i did not test with StackTrace. If it behave as written, i do not know how it should be resolved, maybe there is some other way to get this information?
Without this logging from base classes is way harder, every message has to contain extra GetType().

snakefoot:
@SylwesterZarebski If you promise to use `LogManager.GetCurrentClassLogger()` in your classes, then I recommend you always use `${logger}` instead of `${callsite:className=true:methodName=true}`

Because the callsite-logic is super expensive right now, because the overhead of creating the StackTrace is huge. Trying to improve on this with #2527

SylwesterZarebski:
Thanks, then i should have:
`${logger}.${callsite:className=false:methodName=true}`
instead of:
`${callsite:className=true:methodName=true}`
to get:
`<classtype>.<methodname>`
Correct?

snakefoot:
[...]

PR #2609, opened 2018-03-01
NLog - Fix Callsite when wrapping ILogger in Microsoft Extension Logging

snakefoot:
Fix the issue shown here:

https://tpodolak.com/blog/2018/03/01/asp-net-core-custom-iloggert-nlog-callsite-layout-renderer-issue/

Then in the NLog.Extension.Logging the following line needs to be changed in NLogLogger.cs:

> _logger.Log(eventInfo);

To this:

> _logger.Log(typeof(Microsoft.Extensions.Logging.ILogger), eventInfo);

And then callsite will work for custom Loggers, without needing hidden assembly-logic (and moving custom Logger into separate assembly)

### DIFF
--- a/src/NLog/LoggerImpl.cs
+++ b/src/NLog/LoggerImpl.cs
@@ -216,7 +216,7 @@ namespace NLog
         {
             var method = frame.GetMethod();
             Type declaringType = method.DeclaringType;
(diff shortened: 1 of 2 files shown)


304NotModified:
> So #2579 is not possible to solve using standard StackTrace-logic currently used by NLog.

But I can still merge this one, isn't?

snakefoot:
Yes it is just an observation. Please merge.



Sent from my Samsung device


304NotModified:
ok thx!

PR #2610, opened 2018-03-02
Add notifications when Target drops LogEvent or cache size grows, grow *2 instead of +1 

Pomoinytskyi:
Gives ability to be notified that some LogEvents was not stored, and AsyncRequestQueue growing.


snakefoot:
Bad idea to start allocating EventArgs when the system is already stressed, and no one actually listens to the events.

Pomoinytskyi:
@snakefoot 
I am agree that it is too late to handle situation is we are loosing LogEvents. We have system which can have very high logging acceleration, and we need to know that we are loosing messages.
Presumably good to notify when queue full up to some percentage.

snakefoot:
@Pomoinytskyi Was also thinking you could check if anyone actually listens for the event-listener, before start allocating objects.

snakefoot:
@Pomoinytskyi Could be nice that the "Grow"-event only occurred when it actually grew, and not when reaching the initial requestLimit (Maybe update based on queue-capacity)

Pomoinytskyi:
@snakefoot 
Queue has only Count property, it doesn't provide actual size. I can calculate it based on internal queue algorithms but it will be approximated value. 
Or raise event when queue Count became bigger than maximum Count during working time. 
What do you think?

snakefoot:
I would probably multiply RequestLimit with 2 when reaching the limit


Sent from my Samsung device


src/NLog/Targets/Wrappers/AsyncRequestQueue-T.cs
```diff
@@ -185,5 +197,38 @@ public void Clear()
...
+        private void OnLogEventDropped(AsyncLogEventInfo logEventInfo)
+        {
+            var logEventDropped = LogEventDropped;
+            if (logEventDropped == null)
+            {
+                return;
+            }
+
+            LogEventDroppedEventArgs eventArg = new LogEventDroppedEventArgs(logEventInfo);
+            logEventDropped.Invoke(this, eventArg);
```
304NotModified: I prefer ` LogEventDropped?.Invoke(this, new LogEventDroppedEventArgs(logEventInfo));`

the same for `OnLogEventQueueGrows` 

less is more :)

304NotModified:
> I would probably multiply RequestLimit with 2 when reaching the limit

this isn't a change that is for the notification, isn't? I think this should be a different PR

snakefoot:
> this isn't a change that is for the notification, isn't? I think this should be a different PR?

Seems silly to make growth notification when not growing. Ex. RequestLimit set to 10000, when count 10001 is reached then notification is correctly made. Then one is processed and then one is inserted, thus again count 10001 is reached causing the same notification, but no growth has occurred.

Pomoinytskyi:
> Then one is processed and then one is inserted, thus again count 10001 is reached causing the same notification, but no growth has actually occurred.

I multiply RequestLimit if count equals to RequestLimit and not decrease if items was processed. 
On the next iteration we compare _logEventInfoQueue.Count with new RequestLimit and it should not raise event while _logEventInfoQueue.Count < RequestLimit (already doubled)

```
if (_logEventInfoQueue.Count >= RequestLimit)
      {
              InternalLogger.Debug("Async queue is full");
              switch (OnOverflow)
              {
...
                      case AsyncTargetWrapperOverflowAction.Grow:
                            InternalLogger.Debug("The overflow action is Grow, adding element anyway");
                            RequestLimit *= 2;
                            OnLogEventQueueGrows(RequestCount + 1);
                            break;
...
```
Correct me please, maybe i miss something.

Pomoinytskyi:
> Then one is processed and then one is inserted, thus again count 10001 is reached causing the same notification, but no growth has actually occurred.

I multiply RequestLimit if count equals to RequestLimit and not decrease if items was processed. 
On the next iteration we compare _logEventInfoQueue.Count with new RequestLimit and it should not raise event while _logEventInfoQueue.Count < RequestLimit (already doubled)

```
if (_logEventInfoQueue.Count >= RequestLimit)
      {
              InternalLogger.Debug("Async queue is full");
              switch (OnOverflow)
              {
...
                      case AsyncTargetWrapperOverflowAction.Grow:
                            InternalLogger.Debug("The overflow action is Grow, adding element anyway");
                            RequestLimit *= 2;
                            OnLogEventQueueGrows(RequestCount + 1);
                            break;
...
```
Correct me please, maybe i miss something.

snakefoot:
@Pomoinytskyi Yes that was my idea.

Could you also remove the LogEventDropped-event from Target.cs, and just have it on the AsyncWrapper ?

Pomoinytskyi:
@snakefoot 
I thought that LogEventDropped  event could be reused in other targets,  depends on internal implementation. For example some network target can lost events on network error etc.

snakefoot:
@Pomoinytskyi I was hoping one day, that flow-control for NLog-targets was something done by a wrapper, and not something all targets had to implement themselves. Most flow-control and batching-logic I have seen on NLog-targets are very primitive and doesn't work well with recent inventions like the Internet.

snakefoot:
> Targets/TargetDropEventEventArgs.cs(39,34): error CS1574: XML comment has cref attribute 'LogEventDropped' that could not be resolved

Pomoinytskyi:
@snakefoot 
Sorry, stupid mistake.  Fixed.

304NotModified:
@snakefoot I think this one is ready to merge, do you agree?

snakefoot:
I think that these events on AsyncTargetWrapper:

public event EventHandler<TargetDropEventEventArgs> LogEventDropped;
public event EventHandler<LogEventQueueGrowEventArgs> EventQueueGrow;

Should forward directly to the AsyncRequestQueue-T event. This will prevent unnecessary EventArgs-allocation.

See also AppDomainWrapper.cs for an example of how this is done:

https://github.com/NLog/NLog/blob/19de4feb0a6f1292e9d0c69da5c75b73efda3ea1/src/NLog/Internal/Fakeables/AppDomainWrapper.cs#L180

P.S. @304NotModified  Curious why you keep piling up new stuff on NLog 4.5, thought it was stabilization period.

304NotModified:
> P.S. @304NotModified Curious why you keep piling up new stuff on NLog 4.5, thought it was stabilization period.

I don't like open pull requests. Also if I have the time, I prefer to spent it useful.

snakefoot:
> it will be tech debt for reviewing and demotivating for contributors IMO

People waiting for NLog 4.5 RTM within "in weeks", might also be demotivated by hearing the same words 3 months after :)

Pomoinytskyi:
@snakefoot 

> public event EventHandler LogEventDropped;
>public event EventHandler EventQueueGrow;

>Should forward directly to the AsyncRequestQueue-T event. This will prevent unnecessary EventArgs-allocation.

I'd like to have different event, in order to replace sender. I'd like to have AsyncTargetWrapper in sender instead of AsyncRequestQueue-T
What do you think?

snakefoot:
> I'd like to have different event, in order to replace sender. I'd like to have AsyncTargetWrapper in sender instead of AsyncRequestQueue-T

Then AsyncTargetWrapper should only subscribe to AsyncRequestQueue-T, when someone subscribes to AsyncTargetWrapper.

snakefoot:
@Pomoinytskyi Something like this:

```
        public event EventHandler<TargetDropEventEventArgs> LogEventDropped
        {
            add
            {
                if (_logEventDropped == null)
                    RequestQueue.LogEventDropped += OnRequestQueueDropItem;
                _logEventDropped+= value;
            }
            remove
            {
                _logEventDropped -= value;
                if (_logEventDropped== null)
                    RequestQueue.LogEventDropped -= OnRequestQueueDropItem;
            }
        } 
        private event EventHandler<TargetDropEventEventArgs> _logEventDropped;

        private void OnRequestQueueDropItem(object sender, LogEventDroppedEventArgs logEventDroppedEventArgs) 
        {
            _logEventDropped?.Invoke(this, logEventDroppedEventArgs);  // No need to allocate twice
        }
```

304NotModified:
moved to 4.6 as I'm releasing 4.5 rc7 now.

Pomoinytskyi:
@snakefoot 
Thanks. I am in update process.
I just wondering about construction
` if (_logEventDropped == null)
                    RequestQueue.LogEventDropped += OnRequestQueueDropItem;
                _logEventDropped+= value;`
It is not thread safe, when two threads want to subscribe. But i think it has very low probability.

snakefoot:
> It is not thread safe, when two threads want to subscribe. But i think it has very low probability.

You are very welcome to improve on this. But yes in the case where on is subscribing, and another is unsubscribing then there will be a race-condition. If two threads are subscribing, then no harm is done by subscribing twice, only one subscription will be added (Handled by `EventHandler`-delegate)

Pomoinytskyi:
@snakefoot I think we can neglect this subscribe/unsubscribe race condition. Is it Ok?

snakefoot:
> I think we can neglect this subscribe/unsubscribe race condition. Is it Ok?

Yes fine by me :)

snakefoot:
@Pomoinytskyi Think the automatic unsubscription in `CloseTarget` should be removed.

snakefoot:
@Pomoinytskyi For consistent naming of private variables with underscore prefix, then it should be:

```
private event EventHandler<LogEventDroppedEventArgs> _logEventDroppedEvent;
private event EventHandler<LogEventQueueGrowEventArgs> _eventQueueGrowEvent;
```

Guess we need some stylecop logic :)

304NotModified:
@Pomoinytskyi

would it be possible to add some unit tests?

304NotModified:
> @Pomoinytskyi
> 
> would it be possible to add some unit tests?

@Pomoinytskyi polite bump

Pomoinytskyi:
@304NotModified Yes, i will cover it with UT. Hope to the Monday.

304NotModified:
Thanks in advance!

304NotModified:
@Pomoinytskyi Any idea when you have time to add some unit tests? Thanks!

304NotModified:
Hi @Pomoinytskyi ,

We're trying to complete the PRs for NLog 4.6. If you have time, please add some unit tests. Otherwise this will be NLog 4.7

Pomoinytskyi:
Sorry, i was on vacation, today will spend for unit tests.

304NotModified:
Thanks!

It seems the build is failing now. We need to check that.

304NotModified:
I see that the appveyor build was stuck (and cancelled.) restarted it

src/NLog/Targets/Wrappers/AsyncRequestQueueBase.cs
```diff
@@ -0,0 +1,86 @@
...
+using System;
+
+namespace NLog.Targets.Wrappers
+{
+    using System.Collections.Generic;
+    using NLog.Common;
+
+    internal abstract class AsyncRequestQueueBase
+    {
+        public virtual bool IsEmpty { get; }
```
snakefoot: Any reason for not making it abstract ?

Pomoinytskyi: Only reasin is avoid duplicates.

snakefoot: Not sure I understand "avoid duplicates"

Pomoinytskyi: I misunderstand  your question.
It should be abstract.

304NotModified:
@snakefoot I think this is ready for merge, do you agree?

304NotModified:
ok thanks!

304NotModified:
FYI, NLog 4.6 is online which includes this feature!

PR #2612, opened 2018-03-02
add final param to `AddRule` Methods

893949088:
add final param by AddRule Method

### DIFF
--- a/src/NLog/Config/LoggingConfiguration.cs
+++ b/src/NLog/Config/LoggingConfiguration.cs
@@ -268,6 +268,7 @@ namespace NLog.Config
         /// <param name="maxLevel">Maximum log level needed to trigger this rule.</param>
         /// <param name="targetName">Name of the target to be written when the rule matches.</param>
         /// <param name="loggerNamePattern">Logger name pattern. It may include the '*' wildcard at the beginning, at the end or at both ends.</param>
+        /// <param name="final">Gets or sets a value indicating whether to quit processing any further rule when this one matches.</param>
         public void AddRule(LogLevel minLevel, LogLevel maxLevel, string targetName, string loggerNamePattern = "*", bool final = false)
         {
             var target = FindTargetByName(targetName);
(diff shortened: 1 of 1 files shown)


snakefoot:
This is a breaking change. You have to make new methods, instead of adding parameter to existing ones.

893949088:
@snakefoot I don‘t want to configure nlog by xml file.I want configure nlog by LoggingConfiguration.but the AddRule method don't has 'final' param. why this is a breaking change?

snakefoot:
> why this is a breaking change?

Your code-change requires a complete recompile of all source-code using NLog, or else they will get a MIssingMethodException:

Ex. This method is suddenly gone:

`public void AddRuleForOneLevel(LogLevel level, Target target, string loggerNamePattern)`

Instead there is this method, but it has a different signature:

`public void AddRuleForOneLevel(LogLevel level, Target target, string loggerNamePattern, bool final = false)`

893949088:
@snakefoot  So, Should I open a new issue??

snakefoot:
@893949088 You can just add more commits to the same PR. They will be squashed into a single commit before merge.

893949088:
@snakefoot But I don't have more change to commit.    ╯﹏╰ !!

snakefoot:
@893949088 If you want in included within the next year, then you need to fix the breaking change. Else it will be postponed until like forever.

893949088:
@snakefoot  I'm very sad. why is the 'LoggingConfiguration' imperfect？

snakefoot:
> why is the 'LoggingConfiguration' imperfect？

Because of the breaking change you have introduced, by changing the signature of the existing methods.

893949088:
@snakefoot How can I do it to  fix the breaking change？I haven't submit PR before.
^_^"

snakefoot:
@893949088 The breaking change is that you have modified the signature of the existing methods. The very simple solution is not to modify the signature of the existing methods.

So just add the new methods without default parameters, and make the existing methods call the new method:

```C#
public void AddRuleForOneLevel(LogLevel level, Target target, string loggerNamePattern)
{
   // Existing method forwarding to new method
   AddRuleForOneLevel(level, target, loggerNamePattern, false);
}

public void AddRuleForOneLevel(LogLevel level, Target target, string loggerNamePattern, bool final)
{
   // Do the right thing
}
```

893949088:
@snakefoot  Thank you very much!!!

893949088:
〒_〒！〒_〒！〒_〒！

snakefoot:
@893949088 Remember to remove the parameter default values for the new methods.

304NotModified:
looks good, thanks!

Will merge this when Travis gives green (restarted Travis, it was stuck)

Issue #82, opened 2012-10-07
NLog assumes no threads are created by targets


ghost:
<p>Bug no. 2 today, firming up the RabbitMQ target.</p>

<p>&nbsp;</p>

<p>There&#39;s a problem with NLog in that it has no LogManager.Shutdown()-method -- because it doesn&#39;t have this but relies on the assembly being unloaded, not closing targets until then, there&#39;s a problem if the assemblies are not being unloaded because another thread - in this case the connection thread - is still running.</p>

<p>&nbsp;</p>

<p>I need a way to shutdown NLog properly - log4net does it with LogManager.Shutdown() and as a temporary workaround I&#39;m using this code:</p>

<p>&nbsp;</p>

<p>var allTargets = LogManager.Configuration.AllTargets;</p>

<p>&nbsp;</p>

<p>foreach (var target in allTargets)</p>

<p> target.Dispose();</p>

<p>&nbsp;</p>

<p>While this is tractable, there should be a documented way of shutting down NLog :)!</p>

PR #2616, opened 2018-03-08
LogManager.Shutdown - Should disable file watcher and avoid auto reload

snakefoot:
One might consider to call LogFactory.Close(), but that might be a breaking change as it would prevent the engine to startup again after shutdown (Would have to fix unit tests, so they don't call LogManager.Shutdown)

Adventure that continues from #1899 and #82

LogManager.Shutdown() was created to simulate Shutdown in Log4net.

### DIFF
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -63,7 +63,7 @@ namespace NLog
     {
 #if !SILVERLIGHT && !__IOS__ && !__ANDROID__ && !NETSTANDARD1_3
         private const int ReconfigAfterFileChangedTimeout = 1000;
-        internal Timer reloadTimer;
+        internal Timer _reloadTimer;
         private readonly MultiFileWatcher _watcher;
 #endif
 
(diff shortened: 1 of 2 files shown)


snakefoot:
@304NotModified Improved `NLogManager.Shutdown()` so it performs a more correct shutdown (and also disconnects active loggers from using the closed targets)

snakefoot:
@304NotModified Little annoying that creating a new Logger-object actually causes the LoggingConfiguration to be reloaded after having called LogManager.Shutdown().

Have now changed the logic of `LogManager.Shutdown()`, so it requires an explicit assignment of `LogManager.Configuration` to startup again.

snakefoot:
Victims of shutdown-auto-flush not working on Linux (Should call `NLog.LogManager.Shutdown()`)

 https://github.com/dotnet/corefx/issues/27952
https://github.com/NLog/NLog.Extensions.Logging/issues/201

snakefoot:
@304NotModified Would like to have this included in NLog 4.5 RTM if possible.

There are two important changes:

- LogManager.Shutdown() works correctly when autoload=true
- Avoid unhandled exceptions (segmentation fault) when running on Linux (and forgetting to call LogManager.Shutdown)

304NotModified:
Was doubting about that, but it's merged now :)

snakefoot:
Thank you for the merge. Any update on the timeline for NLog 4.5 RTM ?

304NotModified:
Yes, in the weekend 24+25 I have some more free time and i planned the rtm release for then.

snakefoot:
Very exciting. I will be going on easter holiday from monday the 26th.

PR #2507, opened 2018-01-11
Reduce string allocations for logevents with single target destination

snakefoot:
Will make some performance tests later to see how it performs.

Should also reduce pressure on the internal StringBuilderPool, as it will just write directly to the final destination-StringBuilder.

PR #2620, opened 2018-03-15
Single Target optimization should only be done when parseMessageTemplate = null

snakefoot:
Fixes #2507 so mixed-mode templates also works in single-target scenario. See also #2544


304NotModified:
Is this a bugfix or a performance optimization?

snakefoot:
Bugfix

304NotModified:
Thanks!

PR #2622, opened 2018-03-19
NetworkTarget - Http / Https should not throw on async error response

snakefoot:
And a minor allocation reduction.


304NotModified:
thanks1

Issue #1588, opened 2016-08-06
Configure from JSON instead of XML


304NotModified:
Feature,

Configure NLog from a JSON format. 
### How to handle the multiple XML elements to a dict/list?

 e.g. 

``` xml
    <parameter name="@msg" layout="${message}" />
    <parameter name="@level" layout="${level}" />
```

 to 

``` json
   "parameters": {
        "@msg": "${message}",
        "@level": "${level}",
      },
```
### "XSD"

How to generate the schema def? How does this work with objects?
## Large Example

Something like this:

``` json
{
  "internalLog": {
    "level": "Trace",
    "file": "d:\\work\\log.txt"
  },
  "autoReload": "true",
  "variables": {
    "gmailUsername": "${trim-whitespace:${file-contents:${basedir}/gmailusername.txt}}",
    "gmailPassword": "${trim-whitespace:${file-contents:${basedir}/gmailpassword.txt}}"
  },
  "extensions": {
    "NLog.Extended": {},
    "Custom": {
      "filePath": "..."
    }
  },
  "targets": {
    "file": {
      "type": "File",
      "fileName": "${basedir}/logs/${shortdate}.log",
      "layout": "${longdate} ${aspnet-request:servervariable=URL} ${uppercase:${level}} ${message}"
    },
    "db": {
      "type": "Database",
      "commandText": "INSERT INTO [LogEntries](TimeStamp, Message, Level, Logger) VALUES(getutcdate(), @msg, @level, @logger)",
      //todo how to rename? (was no list/dict here)
      "parameters": {
        "@msg": "${message}",
        "@level": "${level}",
        "@logger": "${logger}"
      },
      "dbProvider": "System.Data.SqlClient",
      "connectionString": "server=.\\SQLEXPRESS;database=MyLogs;integrated security=sspi",
      "install-commands": [
        {
          "text": "CREATE DATABASE MyLogs",
          "connectionString": "server=.\\SQLEXPRESS;database=master;integrated security=sspi",
          "ignoreFailures": "true"
        },
        {
          "text": "CREATE TABLE LogEntries( id int primary key not null identity(1,1), TimeStamp datetime2, Message nvarchar(max), level nvarchar(10), logger nvarchar(128))"
        }
      ],
      "uninstall-commands": [
        {
          "text": "DROP DATABASE MyLogs",
          "connectionString": "server=.\\SQLEXPRESS;database=master;integrated security=sspi",
          "ignoreFailures": "true"
        }
      ]
[...]

PR #2624, opened 2018-03-25
LoggingConfigurationReader - Extracted from XmlLoggingConfiguration

snakefoot:
See also  #1588 

This is merely a proof of concept. I want to remove the quirks from LoggingConfigurationReader, so all the special reader logic for XML and JSON is moved into the specialized LoggingConfiguration-classes.

But if the Json-example presented is completely useless, then no need to proceed further down this road.

### DIFF
--- a/src/NLog/Config/LoggingConfiguration.cs
+++ b/src/NLog/Config/LoggingConfiguration.cs
@@ -45,6 +45,7 @@ namespace NLog.Config
     using NLog.Internal;
     using NLog.Layouts;
     using NLog.Targets;
+    using NLog.Targets.Wrappers;
 
     /// <summary>
     /// Keeps logging configuration and provides simple API
@@ -64,10 +65,24 @@ namespace NLog.Config
         private readonly Dictionary<string, SimpleLayout> _variables = new Dictionary<string, SimpleLayout>(StringComparer.OrdinalIgnoreCase);
 
         /// <summary>
+        /// Gets the factory that will be configured
+        /// </summary>
+        public LogFactory LogFactory { get; }
+
+        /// <summary>
         /// Initializes a new instance of the <see cref="LoggingConfiguration" /> class.
         /// </summary>
         public LoggingConfiguration()
+            :this(LogManager.LogFactory)
         {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="LoggingConfiguration" /> class.
+        /// </summary>
+        public LoggingConfiguration(LogFactory logFactory)
+        {
+            LogFactory = logFactory;
             LoggingRules = new List<LoggingRule>();
         }
 
@@ -109,6 +124,7 @@ namespace NLog.Config
 
         private bool TryGetTargetThreadSafe(string name, out Target target) { lock (_targets) return _targets.TryGetValue(name, out target); }
         private List<Target> GetAllTargetsThreadSafe() { lock (_targets) return _targets.Values.ToList(); }
+
         private Target RemoveTargetThreadSafe(string name)
         {
             Target target;
@@ -126,6 +142,7 @@ namespace NLog.Config
             }
             return target;
         }
+
         private void AddTargetThreadSafe(string name, Target target, bool forceOverwrite)
         {
             if (string.IsNullOrEmpty(name) && !forceOverwrite)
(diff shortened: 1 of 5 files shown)


src/NLog/Config/LoggingConfiguration.cs
```diff
@@ -394,6 +411,64 @@ public void AddRuleForAllLevels(Target target, string loggerNamePattern, bool fi
...
+                }
+            }
+
+            if (removedRules.Count > 0)
+            {
+                lock (LoggingRules)
+                {
+                    for (int i = LoggingRules.Count - 1; i >= 0; i--)
+                    {
+                        if (removedRules.Contains(LoggingRules[i]))
```
304NotModified: isn't better to use the rulename here? Or do you think we will have duplicate names?

src/NLog/Config/LoggingConfigurationReader.cs
```diff
@@ -0,0 +1,1077 @@
...
+using NLog.Targets;
+using NLog.Targets.Wrappers;
+using NLog.Time;
+
+namespace NLog.Config
+{
+    /// <summary>
+    /// Interface for accessing configuration details
+    /// </summary>
+    public interface ILoggingConfigurationSection
```
304NotModified: please move classes/interfaces to separate files, thx!

snakefoot:
> Could you please tell me what's new code, (no moves), makes it easier to review, thanks!

Well before you start the review, then you should read the initial comment. Which I will repeat again. Is the Json-config shown in https://github.com/NLog/NLog/issues/1588#issuecomment-375979728 acceptable?

304NotModified:
> Well before you start the review, then you should read the initial comment. Which I will repeat again. Is the Json-config shown in #1588 (comment) acceptable? (Along with all the restrictions)

yes, I saw it. Have to think about the restrictions. .e.g if a variable is using another variable, that could be an issue (now it's dependent of the order)


Anyway, splitting the config reader from the XML config reader it a good step anyway.

snakefoot:
> .e.g if a variable is using another variable, that could be an issue (now it's dependent of the order)

Well the variables in the variables-section are parsed in their correct order. It is only the individual sections that are parsed in alphabetic order.

304NotModified:
I think the targets config in the proposed JSON looks awesome! Would it be possible to change the "internalLog" and "extensions" section as in the opening post? And how should we configure rules without names?

snakefoot:
> how should we configure rules without names?

The same way you configure targets without names. You don't :)

304NotModified:
> Well the variables in the variables-section are parsed in their correct order.

So this goes well? 

```
 "variables": {
            "rootDir": "c:/somewhere"
            "basedirArchives": "${rootDir}/archives",
            "basedirFiles": "${rootDir}/files",
        },

```

I a bit confused why that one isn't parsed alphabeticly. It it because "extentions" etc is the the intermediate child of the root? IMO Json are just properties and arrays. 



> The same way you configure targets without names. You don't :)

That will be hard for users to come up with unique names while they don't use it.

snakefoot:
> Would it be possible to change the "internalLog" section as in the opening post?

I guess it is possible to fiddle with the LoggingConfigurationReader, so can support reading internalLog-parameters from a section different from the nlog-section.

> Would it be possible to change the "extensions" section as in the opening post?

Nope. This Json is ignored by the microsoft json parser:

`"NLog.Extended": {},`

snakefoot:
> That will be hard for users to come up with unique names while they don't use it.

Same can be said for wrapped targets :)

304NotModified:
> Same can be said for wrapped targets :)

true, but wrapped targets are less common/basic and now we have already unnamed rules.

304NotModified:
> There is also no support for autoReload or including external json-files.

There is `reloadOnChange` (https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?tabs=basicconfiguration)

snakefoot:
> There is reloadOnChange

Sounds like you are ready to take over the PR, and fiddle with this :)

snakefoot:
> true, but wrapped targets are less common/basic and now we have already unnamed rules.

I guess the logging-rules can be implemented like the extensions-section-array:

```
        "rules": [
            {
                "logger": "*",
                "minLevel": "Trace",
                "writeTo": "SplitGroup"
            },
            {
                "logger": "*",
                "minLevel": "Trace",
                "writeTo": "postFilteringWrapper"
            },
        ],
```
I just liked that the rules and targets where using the same syntax.

304NotModified:
> Sounds like you are ready to take over the PR, and fiddle with how to hook into the reloadOnChange-event and perform a controlled reload if something happened to the NLog-section.

I will try, but I don't expect if have time for that before the weekend.

304NotModified:
looks like I have an action to do here :)

304NotModified:
I think this could be merged for 4.6, independent of the discussion of JSON config files, do you agree?

snakefoot:
> There is a pure refactoring, isn't ? I like to accept it.

No it is a bad attempt to reuse the XML-logic for reading JSON-files.

The reordering logic (because JSON-keys are ordered) need happen in the JSON-parser (and not in the reader).

304NotModified:
> No it is a bad attempt to reuse the XML-logic for reading JSON-files.

OK, I think I close it then (for now)

Issue #2627, opened 2018-03-26
NLog locks with big number of thread.


Pomoinytskyi:
Type: Bug
NLog version:  4.4.5097.0
Platform:  x64 CLR 4.0.30319.34209
OS:  6.3.9600 (Windows Server 2012 R2)

Out application creates a huge number of threads and each of them logs some information.
Once we found that application stuck, and created memory dump file.
Investigating SyncBlocks by WinDbg, we found such situation:

Index | SyncBlock | MonitorHeld | Recursion | Owning | Thread Info | SyncBlock | Owner
-----|---------|-----------|---------|------|-----------|---------|-----
36 | 00000009a841a8b8 | 615(307 threads) | 1 | 0000000a8254d890 | 301c 2425 | 00000007eedf7880 | System.Object
1275 | 00000009addc7bf8 | 39 (19 threads) | 1 | 0000000accdfa630 | 349c 2625 | 0000000770977718 | System.Collections.Generic.Dictionary`2[...]
326897 | 000000064ea89618 | 3  (1 thread) | 1 | 0000000acccbd250 | 13d4 2684 | 0000000770978348 | System.Collections.Generic.Dictionary`2[...]
708642 | 000000063f038d28 | 3  (1 thread) | 1 | 0000000a822a6410 | 4b08 2350 | 000000072f8eff60 | System.Object
738325 | 000000064086ff18 | 3  (1 thread) | 1 | 0000000ab82c0aa0 | 2284 2440 | 00000008af2c11c0 | System.Object
965485 | 0000000630b3f268 | 15 (7 threads) | 1 | 0000000a8226dcb0 | 68ac 2581 | 00000006703e8a98 | System.Object
1049611 | 000000063feea228 | 3  (1 thread) | 1 | 0000000a825741c0 | 2bb0 2534 | 00000006efd12de8 | PrivateNamespace.GhostEntitiesCache+HostsGhostData
1200527 | 0000000640d8b3f8 | 4353 (2176 threads) | 1 |0000000a5954d3d0 | 180 1869 | 00000006ef8da510 | System.Object

Call stack of thread 0x2284 (2440) stops on BL code.
Threads: 0x349c (2625), 0x13d4 (2684), 0x4b08 (2350), 0x68ac (2581), 0x2bb0 (2534), 0x180 (1869) stops on the same execution poit and have similart end of the call stack. There is one of then:

```OS Thread Id: 0x2bb0 (2534)
[...]

PR #2628, opened 2018-03-26
RoundRobinGroupTarget - Replaced lock with Interlocked for performance

snakefoot:
Makes it possible to create a highly concurrent file-target.

```xml
<target type="RoundRobinGroup" name="superfast">
    <target type="AsyncWrapper" timeToSleepBetweenBatches="0">
          <target type="file" keepFileOpen="false" fileName="single.log" />
    </target>
    <target type="AsyncWrapper" timeToSleepBetweenBatches="0">
          <target type="file" keepFileOpen="false" fileName="single.log" />
    </target>
    <target type="AsyncWrapper" timeToSleepBetweenBatches="0">
          <target type="file" keepFileOpen="false" fileName="single.log" />
    </target>
    <target type="AsyncWrapper" timeToSleepBetweenBatches="0">
          <target type="file" keepFileOpen="false" fileName="single.log" />
    </target>
    <target type="AsyncWrapper" timeToSleepBetweenBatches="0">
          <target type="file" keepFileOpen="false" fileName="single.log" />
    </target>
    <target type="AsyncWrapper" timeToSleepBetweenBatches="0">
          <target type="file" keepFileOpen="false" fileName="single.log" />
    </target>
</target>
```

Allowing multiple threads to write to the same file without being hit by lock congestion in PrecalculateVolatileLayouts. See also #2627

### DIFF
--- a/src/NLog/Targets/Wrappers/RoundRobinGroupTarget.cs
+++ b/src/NLog/Targets/Wrappers/RoundRobinGroupTarget.cs
@@ -60,8 +60,7 @@ namespace NLog.Targets.Wrappers
     [Target("RoundRobinGroup", IsCompound = true)]
     public class RoundRobinGroupTarget : CompoundTargetBase
     {
-        private int _currentTarget = 0;
-        private readonly object _lockObject = new object();
+        private int _currentTarget = -1;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="RoundRobinGroupTarget" /> class.
@@ -93,6 +92,15 @@ namespace NLog.Targets.Wrappers
         }
 
         /// <summary>
+        /// Ensures forwarding happens without holding lock
+        /// </summary>
+        /// <param name="logEvent"></param>
+        protected override void WriteAsyncThreadSafe(AsyncLogEventInfo logEvent)
+        {
+            Write(logEvent);
+        }
+
+        /// <summary>
         /// Forwards the write to one of the targets from
         /// the <see cref="Targets"/> collection.
         /// </summary>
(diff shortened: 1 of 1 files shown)


304NotModified:
Thanks!

@snakefoot I'm thinking to automatic release Dev packages on nuget (for every master commit?)
What do you think? I think it will boost creating releases and people could faster test it. Like the serilog Dev releases.

snakefoot:
@304NotModified Well if release cycle is going to be once every year, then it is a good idea. If the release cycle is once a month, then I think people can wait. Those too impatient can make private builds. Cannot predict the amount of support it will give, or the eagerness for using pre-release-build by users. NLog 4.5-rc was not that popular, and probably only used by those needing NetCore.

304NotModified:
> @304NotModified Well if release cycle is going to be once every year, then it is a good idea. If the release cycle is once a month, then I think people can wait.

true. Any way, it was hoping for more feedback between releases, it would make it easier to release fast. (fix, test in dev package, confirmation, release)

snakefoot:
> true. Any way, it was hoping for more feedback between releases, it would make it easier to release fast

After all the work with implementing structured logging and all the provided release-candidates, then I find the feedback-loop a little weak for the NLog-project. Uncertain if more dev-releases will improve on this.

304NotModified:
> After all the work with implementing structured logging and all the provided release-candidates, then I find the feedback-loop a little weak for the NLog-project. Uncertain if more dev-releases will improve on this.

Agree on that. I only expect feedback of the ones that reports the error unfortunately. 

Anyway, made make the internal nuget packages (AppVeyor) more visible would be nice

PR #2639, opened 2018-04-02
MessageTemplates - Optimize ParseHole for positional templates

snakefoot:
Optimize the Hole-Peek, so it it is less expensive than the entire string.Format()


snakefoot:
Before:
capture
After:
capture3

304NotModified:
👍

PR #2642, opened 2018-04-04
FileTarget - InitializeFile should skip dictionary lookup when same file

snakefoot:


snakefoot:
Before:
capture2
After:
capture

304NotModified:
👍 nice one again!

PR #794, opened 2015-07-08
Support for auto loading UNC paths

mikeobrien:
Re. #793, the [`GetNLogAssemblyLocation` method](https://github.com/NLog/NLog/blob/master/src/NLog/Config/ConfigurationItemFactory.cs#L267) does not support UNC paths. This can simply be replaced with [`Uri.LocalPath`](https://msdn.microsoft.com/en-us/library/system.uri.localpath(v=vs.110).aspx) which does.

Issue #29, opened 2012-10-06
nlog.dll.nlog is not correctly resolved if hosted on UNC path [regression bug]


Xharze:
> `GetCandidateFileNames` returns AbsolutPath instead of LocalPath in Ln 671.

> That causes nlog.dll.nlog hosted on UNC path not to be resolved.

It retruns:

```
/Share/nlog.dll.nlog
```

instead of:

```
\\Server\Share\nlog.dll.nlog
```
Moved from http://nlog.codeplex.com/workitem/7043

304NotModified:
not sure where to fix this

304NotModified:
This seems to be solved before, but broken again in 4.0.1

See
https://github.com/NLog/NLog/commit/9032367cdf88d66a345d49c8f3fdcf00b820233e#commitcomment-11797254

and https://github.com/Mike120Git/NLog/pull/1

304NotModified:
Raising to must have as the solution is almost clear.

304NotModified:
needs unit tests, if possible

304NotModified:
PR merged

PR #2643, opened 2018-04-04
UWP with NetStandard2 on Net Native does not support Assembly.CodeBase + Handle native methods in StackTrace

### LINKED ISSUE #2632: Xamarin UWP doesn't create a logfile
**Type** (choose one):
- Bug

**NLog version**: 4.5.0-rc07

**Platform**: .Net 4.7 

**Current NLog config**
```xml
      autoReload="true"
      throwExceptions="false"
      internalLogLevel="Trace" internalLogFile="C:\Users\<Path>\nlog-internal.log">
    <targets>
        <target name="file" xsi:type="File"
                layout="${date} | ${level} | ${message}"             
                fileName="${specialfolder:folder=ApplicationData}/qkiss_app-${shortdate}.log"
                keepFileOpen="false"
                encoding="utf-8" />
    </targets>
    <rules>
        <logger name="*" minlevel="Info" writeTo="file" />
    </rules>
</nlog>
```
No log file gets created, even when I use the absolute path for my machine: C:/Users/\<UserName>/AppData/Local/Packages/\<PackageName>/LocalState/\<filePrefix>-${shortdate}.log

What is the current result?  
  - **The logger works fine in debug mode (UWP App)**  
  - Log file won't be created under ReleaseMode. -> .Net Native compilation is on and may couse some 
errors (Reflection etc.)
  - ${environment:USERPROFILE} won't work either

What is the expected result?
  - Creating the log file under ReleaseMode like under DebugMode
 
Did you checked the [Internal log](https://github.com/NLog/NLog/wiki/Internal-Logging)?
  - It wasn't created. I did set internalLogLevel=Trace and path for internalLogFile="C:\Users\...\log.log"

Please post full exception details (message, stacktrace, inner exceptions)
  - I don't get an exception. The log file never gets created

Are there any workarounds? 
  - None that I know of

Is there a version in which it did work? 
  - I don't it had ever worked with UWP ReleaseMode

Environment:
- Visual Studio 2017 Version 15.6.4
- .Net Versione 4.7.0.2556
- NetStandard 2.0.1
- NLOG 4.5.0-rc07

snakefoot:
Throws PlatformNotSupportedException when Net Native. Resolves #2632

https://github.com/dotnet/corert/blob/master/src/System.Private.Reflection.Core/src/System/Reflection/Runtime/General/ThunkedApis.cs#L63

And also fixed an issue with GetCurrentClassLogger that returned wrong classname. And also fixed an issue with callsite that threw NullReferenceExceptions. Happy days with Net Native.

### DIFF
--- a/src/NLog/Config/ConfigurationItemFactory.cs
+++ b/src/NLog/Config/ConfigurationItemFactory.cs
@@ -369,14 +369,11 @@ namespace NLog.Config
                 var extensionDlls = GetNLogExtensionFiles(assemblyLocation);
                 if (extensionDlls.Length==0)
                 {
-                    var entryAssembly = Assembly.GetEntryAssembly();
-                    if (!string.IsNullOrEmpty(entryAssembly?.CodeBase))
+                    var entryLocation = GetAssemblyFileLocation(Assembly.GetEntryAssembly());
+                    if (!string.IsNullOrEmpty(entryLocation) && !string.Equals(entryLocation, assemblyLocation, StringComparison.OrdinalIgnoreCase))
                     {
-                        if (!string.Equals(entryAssembly.CodeBase, nlogAssembly.CodeBase, StringComparison.OrdinalIgnoreCase))
-                        {
-                            assemblyLocation = GetAssemblyFileLocation(entryAssembly);
-                            extensionDlls = GetNLogExtensionFiles(assemblyLocation);
-                        }
+                        assemblyLocation = entryLocation;
+                        extensionDlls = GetNLogExtensionFiles(entryLocation);
                     }
                     else
                     {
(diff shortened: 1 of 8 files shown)


304NotModified:
I will check, there was a good reason.

304NotModified:
Related: https://github.com/NLog/NLog/pull/794

snakefoot:
> I will check, there was a good reason.

You don't have to. Think the GAC and unit-test-frameworks are enough reasons not to keep  Assembly.CodeBase. Though one could consider having fallback to Assembly.Location (when Assembly.CodeBase doesn't work), but that is a different PR.

304NotModified:
There was a fall back, but removed. See #29

304NotModified:
Any way, we need to check the unc path. Afaik there is no unit test for it.

304NotModified:
> Could be nice if this callsite/class-lookup logic could extracted to a separate NLog-package, and remove the StackTrace-dependency completely (Ex. for NLog ver. 6).


Don't we need it for every platform?

snakefoot:
> Don't we need it for every platform?

Many people can live without having GetCurrentClassLogger() and Callsite, especially when it means less dependencies and less overhead. Do not recall seeing this logic in Serilog.

snakefoot:
> (Small poll, but better than no poll).

Think those who voted, regret it now as it took 2 years to complete, and everyone was dead in the water :)

>  optional is always nice, but a bit difficult as it's a static method (on logmanager).

Was thinking NLog 6.0 meant breaking old chains. A quick search and replace should be something most programmers are capable of.

snakefoot:
Another thing that should be moved out of NLog is the automatic assembly loading. It has been a pain to maintain, and promiscuous assembly loading is not recommended by security specialists. Instead it should just be an extension to the internal configuration-loader (separate nuget-package).

304NotModified:
> Any way, we need to check the unc path. Afaik there is no unit test for it.

Could you check if a config from UNC still works, or should I do that?

snakefoot:
> Could you check if a config from UNC still works, or should I do that?

I have not changed anything in the assembly loading, besides handling the PlatformNotSupportedException. But please test away.

src/NLog/Config/ConfigurationItemFactory.cs
```diff
@@ -369,14 +369,11 @@ private static ConfigurationItemFactory BuildDefaultFactory()
                 var extensionDlls = GetNLogExtensionFiles(assemblyLocation);
                 if (extensionDlls.Length==0)
                 {
-                    var entryAssembly = Assembly.GetEntryAssembly();
-                    if (!string.IsNullOrEmpty(entryAssembly?.CodeBase))
+                    var entryLocation = GetAssemblyFileLocation(Assembly.GetEntryAssembly());
+                    if (!string.IsNullOrEmpty(entryLocation) && !string.Equals(entryLocation, assemblyLocation, StringComparison.OrdinalIgnoreCase))
```
304NotModified: good check for the `string.Equals` (saves us a `GetNLogExtensionFiles`) 👍

src/NLog/Config/ConfigurationItemFactory.cs
```diff
@@ -473,32 +470,60 @@ private static ConfigurationItemFactory BuildDefaultFactory()
...
+                    InternalLogger.Warn("Skipping auto loading location because location is empty: ({1})", fullName);
+                    return string.Empty;
+                }
+#endif
+
                 Uri assemblyCodeBase;
                 if (!Uri.TryCreate(assembly.CodeBase, UriKind.RelativeOrAbsolute, out assemblyCodeBase))
                 {
-                    InternalLogger.Warn("No auto loading because assembly code base is unknown");
+                    InternalLogger.Warn("Skipping auto loading location because code base is unknown: `{0}` ({1})", assembly.CodeBase, fullName);
```
304NotModified: not very important, but why the backticks and not quotes?

src/NLog/LoggerImpl.cs
```diff
@@ -167,12 +167,12 @@ internal static int SkipToUserStackFrameLegacy(StackFrame[] stackFrames, int fir
...
                     continue;
 
-                if (stackFrame.GetMethod().Name == "MoveNext")
+                if (stackFrame.GetMethod()?.Name == "MoveNext")
                 {
                     if (stackFrames.Length > i)
                     {
                         var nextStackFrame = stackFrames[i + 1];
-                        var declaringType = nextStackFrame.GetMethod().DeclaringType;
+                        var declaringType = nextStackFrame.GetMethod()?.DeclaringType;
```
304NotModified: .GetMethod() cannot return null now, correct?

snakefoot: Not sure I understand. GetMethod() can return null, so free navigation can be dangerous. See the `stackFrames[i + 1]`.

src/NLog/Config/ConfigurationItemFactory.cs
snakefoot: Fixed

snakefoot:
> Also small performance improvement, isn't? 

Don't think I have made any performance improvements on purpose. Mostly refactoring to improve the code-reuse for callsite-logic and GetCurrentClassLogger .

src/NLog/LoggerImpl.cs
304NotModified: yes, missed that, thanks!

304NotModified:
> I have not changed anything in the assembly loading, besides handling the PlatformNotSupportedException. But please test away.

tested it anyway, :) Thanks!

snakefoot:
> @snakefoot I have not changed anything in the assembly loading,

@304NotModified Guess I can withdraw that statement and wait for #2830 to be merged :(

PR #2644, opened 2018-04-05
FallbackGroupTarget: handle async state on fallback correctly

snakefoot:
FallbackGroupTarget must call PrecalculateVolatileLayouts upfront to support async state.

Changed the concurrency-logic a little to reduce the congestion while waiting for PrecalculateVolatileLayouts to complete.

This will increase the overhead when using targets that depends on async state in their Layouts. Because it will allocate upfront for the fallback target even if it will not be needed.


304NotModified:
thanks again! :)

PR #2645, opened 2018-04-05
Minor performance optimization of some layoutrenderers

snakefoot:
StringBuilderExt.AppendInvariant appends without allocation


304NotModified:
thanks!

PR #2650, opened 2018-04-07
Better Async performance with many concurrent logger threads (up to 500%+)  - Async attribute and AsyncWrapper

snakefoot:
ConcurrentQueue works very well in NetCore2. But not sure when the improvements will get to NetFramework.

When 8 threads are attacking the same async-wrapper, then the ConcurrentRequestQueue performs twice as fast on NetCore2.

PR #2653, opened 2018-04-07
LayoutRenderer ThreadSafe Attribute introduced to allow lock free Precalculate

snakefoot:
PrecalculateVolatileLayouts no longer need to hold monitor lock.

Together with #2650, then the AsyncTargetWrapper will become highly concurrent with no locks. Early testing shows that performance is doubled when having 8 threads logging against the same target.

This pretty much touches all Layout-related files, so will probably become a merge hell, as it will wait for NLog 4.6


304NotModified:
> This pretty much touches all Layout-related files, so will probably become a merge hell, as it will wait for NLog 4.6

Do you mean ”merge before others"?

snakefoot:
> Do you mean ”merge before others"?

Nope. I mean it will become a burden to maintain while waiting for NLog 4.6

304NotModified:
But we don't have to wait?

snakefoot:
> But we don't have to wait?

I thought that we were still in hypercase mode for the recently released structured logging and netcore features. So rocking the boat and adding new features, should be delayed until smooth sailing has occurred for a month (or so).

304NotModified:
Could you please give me an example of a non threadsafe layout renderer?

PR #2666, opened 2018-04-15
Allow skipping tests in build.

304NotModified:
skip tests on AppVeyor when the string `skip tests` is found in the last commit


304NotModified:
Would expected that `exit` would only stop the test phase, but it seems that it stops the whole pipeline

PR #2675, opened 2018-04-27
DatabaseTarget - InitializeTarget with more logging

snakefoot:
Make it easier to diagnose provider lookup errors using the InternalLogger. For NLog 4.5.4

### DIFF
--- a/src/NLog/Targets/DatabaseTarget.cs
+++ b/src/NLog/Targets/DatabaseTarget.cs
@@ -363,11 +363,11 @@ namespace NLog.Targets
                     throw new NLogConfigurationException($"Connection string '{ConnectionStringName}' is not declared in <connectionStrings /> section.");
                 }
 
-                ConnectionString = SimpleLayout.Escape(cs.ConnectionString);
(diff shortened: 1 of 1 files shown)


src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -392,34 +392,63 @@ protected override void InitializeTarget()
                 }
                 catch (Exception ex)
                 {
-                    InternalLogger.Warn(ex, "DatabaseTarget(Name={0}): DbConnectionStringBuilder failed to parse ConnectionString", Name);
+#if !NETSTANDARD
+                    if (!string.IsNullOrEmpty(ConnectionStringName))
+                        InternalLogger.Warn(ex, "DatabaseTarget(Name={0}): DbConnectionStringBuilder failed to parse ConnectionString: {1}", Name, ConnectionStringName);
```
304NotModified: I would expect the connectionstring after `ConnectionString:`.  Maybe change it to `ConnectionString name`?

PR #2681, opened 2018-04-29
SonarQube change to SonarScanner.MSBuild.exe

snakefoot:
Random shot in the dark for making SonarQube work again.

`$env:sonar_token` is not available for PullRequest (unless configured), so the real test happens when merging to master.


304NotModified:
Thanks, let's found out!

PR #2684, opened 2018-04-30
DatabaseTarget - Skip static assembly lookup for .Net Standard

snakefoot:
And ignore case on type lookup. Trying to improve behavior on ServiceFabric (See also https://github.com/NLog/NLog.Web/issues/281)


304NotModified:
Thanks!

PR #2685, opened 2018-04-30
SonarLint - Fixed different warnings from SonarQube

snakefoot:
Also tried to fix this very "nasty" bug (Apparently the "last" bug in NLog):

https://sonarcloud.io/project/issues?id=nlog&resolved=false&types=BUG


snakefoot:
> Is this now also a bug fix?

Nope it is not a bug in NLog, but a bug in Sonar that fails to parse the code.

304NotModified:
OK thanks!

PR #2688, opened 2018-05-02
Faulty invalidate of FormattedMessage when getting PropertiesDictionary

snakefoot:
Found a bug in NLog 4.5, where the LogEventInfo.FormattedMessage is recalculated when PropertiesDictionary is accesed.

This doesn't work well with complex parameters that might change after the log-statement.


304NotModified:
Thanks will merge soon!

PR #2689, opened 2018-05-02
LogEventInfo - Structured logging parameters are not always immutable

snakefoot:
The bugfix followup to #2688

Should also consider to implement the same handling for Exceptions, as they also mutate when rethrown after logging (changes stacktrace or the user can modify the Data-dictionary).

Little sad that I don't have a good fix for TargetWithContext.

### DIFF
--- a/src/NLog/Config/MutableUnsafeAttribute.cs
+++ b/src/NLog/Config/MutableUnsafeAttribute.cs
@@ -1,4 +1,4 @@
-// 
+﻿// 
 // Copyright (c) 2004-2018 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
 // 
 // All rights reserved.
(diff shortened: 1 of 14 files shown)


304NotModified:
when should be apply [ MutableUnsafe]? could you add that to the docs? Thx!

snakefoot:
Not sure. Since Nlog never handled the issue with complex properties.

And no one have really complained, but now with Nlog we make heavy use of properties and not just string format. So maybe issue will get worse.

It is already difficult to explain  ThreadAgnostic for outsiders. Trying make people understand this issue is not going to be easier, besides document in code.

The precalculate patch only works when using layouts. Not when making custom targets that access LogEventInfo directly. See very few making custom layouts, mostly custom targets

Sent from my Samsung device


304NotModified:
> It is already difficult to explain  ThreadAgnostic for outsiders.

yeah, the docs was meant for maintainers, like me ;)

304NotModified:
thx, docs could be another time

PR #2696, opened 2018-05-04
Fixed Sonar warnings

304NotModified:


src/NLog/Targets/FileTarget.cs
```diff
@@ -922,13 +917,13 @@ protected override void InitializeTarget()
         {
             base.InitializeTarget();
 
-            _appenderFactory = GetFileAppenderFactory();
+            var appenderFactory = GetFileAppenderFactory();
```
304NotModified: @snakefoot could you please double check this? Thx

snakefoot: Looks sane

304NotModified: Thanks!

PR #2719, opened 2018-05-13
Databasetarget install comand fix parameters usage

Jejuni:
RunInstallCommands should now take into account all parameters passed to it.
WriteEventToDatabase and RunInstallCommands both use parameter looping logic, so that has been extracted into its own private method.

Various additions to the Unit tests. 
Added a reference to Microsoft.Data.Sqlite and removed Microsoft.Data.Sqlite.Core.
This should make all sqlite tests work for both 452 and netcore.

### DIFF
--- a/tests/NLog.UnitTests/Targets/DatabaseTargetTests.cs
+++ b/tests/NLog.UnitTests/Targets/DatabaseTargetTests.cs
@@ -346,6 +346,45 @@ Dispose()
         }
 
         [Fact]
+        public void InstallParameterTest()
+        {
+            MockDbConnection.ClearLog();
+
+            DatabaseCommandInfo installDbCommand = new DatabaseCommandInfo
+            {
+                Text = $"INSERT INTO dbo.SomeTable(SomeColumn) SELECT @paramOne WHERE NOT EXISTS(SELECT 1 FROM dbo.SomeOtherTable WHERE SomeColumn = @paramOne);"
+            };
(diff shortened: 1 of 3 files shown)


tests/NLog.UnitTests/Targets/DatabaseTargetTests.cs
```diff
@@ -1013,7 +1155,12 @@ private void SetupSqliteConfigWithInvalidInstallCommand(string databaseName)
 
             // Use an in memory SQLite database
             // See https://www.sqlite.org/inmemorydb.html
+#if NETSTANDARD
+            var connectionString = "Data Source=:memory:";
```
304NotModified: cool! Thanks for that!

Jejuni: No prob :)

For Sqlite to fully work under net standard you need the package reference to the full Microsoft sqlite package and use some different syntax for the connection string.
Tripped me up quite a bit on one of my own projects just a week or two ago.

304NotModified:
AFAIK we could also configure the parameters in the XML config for the install, ins't?

Maybe we should add an example for it at  https://github.com/nlog/nlog/wiki/Database-target

snakefoot:
> Setting it to "System.Data.SqlClient" in the full framework will fail, because the switch-case for it is excluded

Believe the following line will work just fine for the full-framework:

`ConnectionType = Type.GetType(DBProvider, true, true); // DBProvider = "System.Data.SqlClient"`

PR #2726, opened 2018-05-15
WhenRepeated - Support logging rules with multiple targets

### LINKED ISSUE #2722: Nlog 4.5 WhenRepeated with multiple targets in logger writeTo only writes to first target

snakefoot:
Resolves #2722 (Abuses the knowledge that targets are added sequentially for each logging-rule).

Also a small performance optimization for logging rules with multiple targets, that shares the same filter.

### DIFF
--- a/src/NLog/LoggerImpl.cs
+++ b/src/NLog/LoggerImpl.cs
@@ -95,12 +95,19 @@ namespace NLog
                 logEvent.MessageFormatter = LogMessageTemplateFormatter.DefaultAutoSingleTarget.MessageFormatter;
             }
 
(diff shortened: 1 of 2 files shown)


melmears:
Thanks for the quick fix!

snakefoot:
The workaround is to have a logging rule for each target. But it gives some overhead



Sent from my Samsung device


304NotModified:
> The workaround is to have a logging rule for each target. But it gives some overhead

that's the workaround in the current release, isn't?

snakefoot:
Yes while waiting for this fix to be released



Sent from my Samsung device

Issue #2721, opened 2018-05-15
StackTrace() method of LogEventInfo class of NLog giving null value when upgrade to NLog version 4.5.4


RudraRavi:
Hi! Thanks for reporting this feature/bug/question! 

Please keep / fill in the relevant info from this template so that we can help you as best as possible.

QUESTIONS are preferred on StackOverflow. You could expect a faster response there (but do include all the info below). Use the tag "nlog" https://stackoverflow.com/questions/ask

For .NET Core users, please check the platform support: https://github.com/NLog/NLog/wiki/platform-support

**Type** (choose one):

- Bug
- Feature request
- Question

**NLog version**: (e.g. 4.2.3)

**Platform**: .Net 3.5 / .Net 4.0 / .Net 4.5 / Mono 4 / Silverlight 4 / Silverlight 5 / Xamarin Android / Xamarin iOs / .NET Core 

**Current NLog config** (xml or C#, if relevant)

```xml
<nlog>
  <targets>
  </targets>
  <rules>
  </rules>
</nlog>

```

In case of a BUG:

- What is the current result?
- What is the expected result?
- Did you checked the [Internal log](https://github.com/NLog/NLog/wiki/Internal-Logging)?
- Please post full exception details (message, stacktrace, inner exceptions)
- Are there any workarounds? yes/no
- Is there a version in which it did work? 
- Can you help us by writing an unit test?

In case of a FEATURE REQUEST:
 
 - Why do we need it?
 - An example of the XML config, if relevant.

RudraRavi:
What is the current result? : LogEventInfo.StackTrace() gives null value
What is the expected result? : this should return stack trace
Did you checked the Internal log?: yes
Please post full exception details (message, stacktrace, inner exceptions) : it is null reference exception as we are calculating GetFrame() afterward and due to null value of stacktrace it is giving object reference set to null.
Are there any workarounds? yes/no : not sure
Is there a version in which it did work? : Yes in 4.4.5
Can you help us by writing an unit test? : Yes

304NotModified:
@RudraRavi please share (stripped version of) your config. Thanks!

RudraRavi:
We have custom Target class inherit from Target class of NLog .The only problem we are facing after using latest version of NLog that is null value of LogEventInfo.StackTrace(). This is not null for older version till NLog version 4.4.5. I am not using nlog config as handled in code for target.

snakefoot:
[...]

PR #2727, opened 2018-05-16
Support for custom targets that implements IUsesStackTrace

snakefoot:
Allow custom targets to request LogEventInfo.StackTrace without needing to use magically public properties. "Just" have to implement `NLog.Internal.IUsesStackTrace`

See also #2721. Fixes one more regression from #1310, but still doesn't fix this use-case:

https://github.com/GibraltarSoftware/Gibraltar.Agent.NLog2/blob/master/src/Agent.NLog2/GibraltarTarget.cs


304NotModified:
Thanks! 👍 


> but still doesn't fix this use-case:

That one should implement the interface?

snakefoot:
> That one should implement the interface?

That or having a public property with a Layout that references the `${callsite}`

PR #2733, opened 2018-05-22
ReconfigExistingLoggers should not block for new Loggers

snakefoot:
Allow other threads to allocate new Loggers while ReconfigExistingLoggers are working on initializing targets.

The targets are protected by their own locks, and doesn't need the extra locking.


snakefoot:
#2048 added the extra locking, because the Layouts (and other random objects implementing ISupportsInitialize) are not guaranteed to have their own lock-protection like the Targets have.

Closing again.

PR #2743, opened 2018-05-25
JsonLayout - Reduce allocations when needing to escape string (44% time improvement)

snakefoot:
Reuse truncated StringBuilder, instead of creating new.

Also skip the temporary StringBuilder, when using `Encode=false`. Performance optimization for nested JsonLayouts. But one can also consider using `Encode=false` for simple fields like `${longdate}` or `${level}` as it skips unnecessary validation.

### DIFF
--- a/src/NLog/LayoutRenderers/Wrappers/JsonEncodeLayoutRendererWrapper.cs
+++ b/src/NLog/LayoutRenderers/Wrappers/JsonEncodeLayoutRendererWrapper.cs
@@ -78,9 +78,9 @@ namespace NLog.LayoutRenderers.Wrappers
         {
             if (JsonEncode && RequiresJsonEncode(target))
             {
-                var result = Targets.DefaultJsonSerializer.EscapeString(target.ToString(), EscapeUnicode);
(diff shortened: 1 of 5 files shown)


snakefoot:
Think one could improve performance even more, by writing directly into the StringBuilder, and just truncate when needing to escape.

snakefoot:
| JsonLogger       |  2.500.000 |  128 |    5 |       4 |       1 |

| This PR   |    15.972 |   156.514 |   1 |  10 |  20 |   24.828 |    2.238,2 |
| NLog 4.5.5    |    23.073 |   108.348 |   1 |  29 |  57 |   26.421 |    3.736,2 |

304NotModified:
nice, 44% improvement!

snakefoot:
Think one could optimize all that inherits from `WrapperLayoutRendererBuilderBase` with the same technique used here.

304NotModified:
> Think one could optimize all that inherits from WrapperLayoutRendererBuilderBase with the same technique used here (in my last commit).

sound good :)

PR #2751, opened 2018-05-30
DistinctlyThrottlingWrapperTarget

kvchernyaev:
Added one new wrapper target. It logs first record of each new string immediately but then accumulates for a time and flush by timer. It is extremly useful for sending errors via email - so your email will not be overflowed.

### DIFF
--- a/src/NLog/Targets/Wrappers/DistinctlyThrottlingWrapperTarget.cs
+++ b/src/NLog/Targets/Wrappers/DistinctlyThrottlingWrapperTarget.cs
@@ -31,8 +31,6 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
-//#define NET3_5
-
 
 #region usings
 using System;
@@ -375,7 +373,7 @@ namespace NLog.Targets.Wrappers
                     if (_entriesCounts.Remove(e))
 #endif
                     {
-                        if (count.Item1 > 1 && !string.IsNullOrWhiteSpace(CountAppendFormat))
+                        if (count.Item1 > 1 && !string.IsNullOrEmpty(CountAppendFormat))
                             if (GroupByTemplate && e.LogEvent.Message.Contains("{"))
                                 // cut off the last?? it is separator - i think do not
                                 e.LogEvent.Message = Escape(e.LogEvent.Message) +
--- a/src/NLog/Targets/Wrappers/DistinctlyThrottlingWrapperTarget.cs
+++ b/src/NLog/Targets/Wrappers/DistinctlyThrottlingWrapperTarget.cs
@@ -31,7 +31,7 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
-#define NET3_5
+//#define NET3_5
 
 
 #region usings
@@ -62,6 +62,16 @@ namespace NLog.Targets.Wrappers
 #if !NET3_5
         ConcurrentDictionary<AsyncLogEventInfo, Tuple<int, StringBuilder>> _entriesCounts;
 #else
+        class Tuple<T1, T2>
+        {
+            public T1 Item1;
+            public T2 Item2;
+            public Tuple(T1 item1, T2 item2)
+            {
+                Item2 = item2;
+                Item1 = item1;
+            }
+        }
         Dictionary<AsyncLogEventInfo, Tuple<int, StringBuilder>> _entriesCounts;
 #endif
 
--- a/src/NLog/Targets/Wrappers/DistinctlyThrottlingWrapperTarget.cs
+++ b/src/NLog/Targets/Wrappers/DistinctlyThrottlingWrapperTarget.cs
@@ -111,10 +111,10 @@ namespace NLog.Targets.Wrappers
             public int GetHashCode(AsyncLogEventInfo x)
             {
                 LogEventInfo a = x.LogEvent;
-                int withoutExc = a.LoggerName.GetHashCode() ^
-                                 (_useFormattedMessage ? a.FormattedMessage : a.Message)
-                                 .GetHashCode() ^
-                                 a.Level.GetHashCode();
+                if (a == null) return 0;
+                int withoutExc = (a.LoggerName?.GetHashCode() ?? 0) ^
+                                 ((_useFormattedMessage ? a.FormattedMessage : a.Message)?.GetHashCode() ?? 0) ^
+                                 (a.Level?.GetHashCode() ?? 0);
                 return a.Exception == null
                     ? withoutExc
                     : withoutExc ^ (a.Exception.Message?.GetHashCode() ?? 0) ^
--- a/src/NLog/Targets/Wrappers/DistinctlyThrottlingWrapperTarget.cs
+++ b/src/NLog/Targets/Wrappers/DistinctlyThrottlingWrapperTarget.cs
@@ -1,4 +1,38 @@
-﻿#region usings
+﻿// 
+// Copyright (c) 2004-2018 Konstantin Chernyaev <konstantin.chernyaev@list.ru>
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without 
+// modification, are permitted provided that the following conditions 
+// are met:
+// 
+// * Redistributions of source code must retain the above copyright notice, 
+//   this list of conditions and the following disclaimer. 
+// 
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution. 
+// 
+// * Neither the name of Konstantin Chernyaev nor the names of its 
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission. 
+// 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE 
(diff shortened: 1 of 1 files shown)


snakefoot:
Would it not be possible to create something similar using:

https://github.com/NLog/NLog/wiki/AutoFlushWrapper-target
https://github.com/nlog/NLog/wiki/BufferingWrapper-target
https://github.com/NLog/NLog/wiki/WhenRepeated-Filter

Where `AutoFlushWrapper.Condition` is using the `WhenRepeated`-filter to decide if it should signal flush to the `BufferingWrapper` ?

The `BufferingWrapper` can be configured to have a timer that triggers flush if nothing has occurred for a while.

Maybe your special wrapper can be preconfigured setup of these 3 existing components?

kvchernyaev:
I am very sorry, but not. Such functionality can not be implemented.
BufferingWrapper flushes all messages - imagine if you will receive email with 10000 identical exceptions in it (if our production system has bug it is not very big number). So i need a distinct-like operator. 
And whenRepeated-filter can not send the first message immediately using message equality.

I need behavior - if two errors (for example, "e1" and "e2") are become very often, i want to receive only 2 email every 5 minutes - one for "e1" saying "e1 was thrown 10000 times" and one for "e2" saying "e2 was thrown 123 times".

Sorry for my poor english

snakefoot:
`whenRepeated`-filter will flush on first message.

`BufferingWrapper` has an overflow-mechanism you can configure to discard, so it will only keep as many messages as the queue-limit allows.

304NotModified:
> I need a behavior - if two errors (for example, "e1" and "e2") are become very often, i want to receive only 2 emails every 5 minutes - one for "e1" saying "e1 was thrown 10000 times" and one for "e2" saying "e2 was thrown 123 times".

I think this is a really useful configuration to support. I was thinking it should be implemented in the limitingWrapper, but not sure.

Please check the build logs:

1) Your code isn't .NET 4 compatible (and maybe also not other platforms):

> NLog -> C:\projects\nlog\src\NLog\bin\Release\net40-client\NLog.dll
Targets\Wrappers\DistinctlyThrottlingWrapperTarget.cs(3,26): error CS0234: The type or namespace name 'Concurrent' does not exist in the namespace 'System.Collections' (are you missing an assembly reference?) [C:\projects\nlog\src\NLog\NLog.csproj]
Targets\Wrappers\DistinctlyThrottlingWrapperTarget.cs(9,24): error CS0234: The type or namespace name 'Tasks' does not exist in the namespace 'System.Threading' (are you missing an assembly reference?) [C:\projects\nlog\src\NLog\NLog.csproj]
Targets\Wrappers\DistinctlyThrottlingWrapperTarget.cs(25,18): error CS0246: The type or namespace name 'ConcurrentDictionary<,>' could not be found (are you missing a using directive or an assembly reference?) [C:\projects\nlog\src\NLog\NLog.csproj]

Your target needs a constructor with "Name" (next to the default ctor) see other targets

src/NLog/Targets/Wrappers/DistinctlyThrottlingWrapperTarget.cs
```diff
@@ -0,0 +1,220 @@
...
+
+
+
+        Timer _flushTimer;
+        readonly object _lockObject = new object();
+
+
+        #region ctors
+        /// <summary>
+        ///
```
304NotModified: please add also docs to the ctors.

src/NLog/Targets/Wrappers/DistinctlyThrottlingWrapperTarget.cs
```diff
@@ -0,0 +1,220 @@
...
+                       a.FormattedMessage == b.FormattedMessage &&
+                       a.Level == b.Level &&
+                       a.Exception?.ToString() == b.Exception?.ToString();
+            }
+
+
+            public int GetHashCode(AsyncLogEventInfo x)
+            {
+                LogEventInfo a = x.LogEvent;
+                int noe = a.LoggerName.GetHashCode() ^ a.FormattedMessage.GetHashCode() ^
```
304NotModified: is this written by hand or generated? What's a noe?

src/NLog/Targets/Wrappers/DistinctlyThrottlingWrapperTarget.cs
```diff
@@ -0,0 +1,220 @@
...
+        [RequiredParameter]
+        public int FlushTimeout { get; set; }
+
+
+        /// <summary>
+        /// Append count of accumulated waiting messages to the <see cref="LogEventInfo.Message"/> when this wrapper is flushed
+        /// </summary>
+        /// <docgen category='Rendering Options' order='10' />
+        [DefaultValue(null)]
+        public string AccumulatedCountMessageAppendFormat { get; set; }
```
304NotModified: Maybe another name for this, (but no abbreviations please)

src/NLog/Targets/Wrappers/DistinctlyThrottlingWrapperTarget.cs
```diff
@@ -0,0 +1,220 @@
...
+
+            if (count == 0)
+            {
+                WrappedTarget.WriteAsyncLogEvents(logEvent);
+                TurnOnTimerIfOffline();
+            }
+        }
+
+
+        volatile bool _isTimerOnNow;
```
304NotModified: please move fields a the start of the class

src/NLog/Targets/Wrappers/DistinctlyThrottlingWrapperTarget.cs
```diff
@@ -0,0 +1,220 @@
...
+        {
+            if (WrappedTarget == null)
+            {
+                InternalLogger.Error("BufferingWrapper(Name={0}): WrappedTarget is NULL", Name);
+                return;
+            }
+
+            //Console.WriteLine("FLUSH " + DateTime.Now);
+            //Console.WriteLine(
+            //    $"---{string.Join(",", _entriesCounts.Select(kvp => $"{kvp.Key.LogEvent.FormattedMessage}({kvp.Value})"))}");
```
304NotModified: please remove old code

kvchernyaev:
304NotModified, very thanks for your attention and corrects!

kvchernyaev:
304NotModified, i've pushed corrects that you requested and one question. Please help.
For behavior that i needed i use ConcurrentDictionary class from System.Collections.Concurrent namespace - but it was introduced to .net framework only in 4.0 version. So it is not exists in .net 3.5 that is supported by nlog. Please advise what should i do.

kvchernyaev:
I've done correct:
I intruduced setting _bool GroupByTemplate_. If it set to true, **DistinctlyThrottlingWrapperTarget** groups log messages by raw **Message** (not by **FormattedMessage**!). And that groups are flushed (by timer) with the list of occured **FormattedMessages**.
For example, if developer logs:
_logger.Error("ops! user: {userid}", userid);_
And, of course, _userid_ has different value every time. So email will be sent at first error, and every period of time (**FlushTimeout**):
_ops! user: {userid} - 10000 times:
ops! user: 123
ops! user: 543
ops! user: 756
..._

If formatting is not used, no list will be rendered in email:
_ops! - 10000 times:_

304NotModified:
could you please check the build log?

https://ci.appveyor.com/project/nlog/nlog/build/4.5.0-7897

There are some unstable tests in NLog, you could ignore them (I will check them), but at least it should build for all platforms.

kvchernyaev:
304NotModified, please help.
For behavior that i needed i use **ConcurrentDictionary** class from **System.Collections.Concurrent** namespace - but it was introduced to .net framework only in **4.0** version. So it is not exists in .net **3.5** that is supported by nlog. Please advise what should i do.

304NotModified:
Use `#if !NET4_0` etc

src/NLog/Targets/Wrappers/DistinctlyThrottlingWrapperTarget.cs
```diff
@@ -31,10 +31,14 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
+#define NET3_5
```
304NotModified: the NET3_5 etc is defined already in the csproj for the .NET 3.5 platform. If you set it here, everything is defined as net3_5..

304NotModified:
Do you need help on this?

304NotModified:
> But the property Exception.TargetSite, according to documentation, was in .net from the begigging - https://msdn.microsoft.com/ru-ru/library/system.exception.targetsite(v=vs.90).aspx
> I can not understand it - please advise, what is the problem?

That's strange indeed!

I will try to check locally

304NotModified:
checking now

304NotModified:
The log is very confusing. There is no problem in .NET 3.5, but in .NET standard 1.5:

image

304NotModified:
VS 2017 is also more clear: 
image

304NotModified:
added `#if`

snakefoot:
Some things that need to be fixed:
- Targets should not be modifying the LogEventInfo (and its Properties). The same LogEventInfo is shared by all relevant targets, causing funny business. Only the Logger-object should be modifying the LogEventInfo
- When using `targets async="true"` with this target, then Layouts will not recognize your magic updating of LogEventInfo.Properties.

I recommend you find a solution similar to whenRepeated, or maybe build on top of whenRepeated.

304NotModified:
Thanks for checking @snakefoot

304NotModified:
@kvchernyaev any time to check the review comments of https://github.com/NLog/NLog/pull/2751#issuecomment-413037411

Thanks?

304NotModified:
Please open a new PR (or I could reopen this one), when you have time. Thanks!

PR #2752, opened 2018-05-30
JSON: Fixes issue where char types are not properly escaped

smbecker:
DefaultJsonSerializer was not properly escaping 'char' types, which could result in invalid JSON.

### DIFF
--- a/src/NLog/Targets/DefaultJsonSerializer.cs
+++ b/src/NLog/Targets/DefaultJsonSerializer.cs
@@ -98,18 +98,6 @@ namespace NLog.Targets
             {
                 return "null";
             }
-            else if (value is char c)
-            {
-                if (RequiresJsonEscape(c, options.EscapeUnicode))
-                {
-                   StringBuilder sb = new StringBuilder(5);
-                   sb.Append('"');
-                   AppendStringEscape(sb, c.ToString(), options.EscapeUnicode);
-                   sb.Append('"');
-                   return sb.ToString();
-                }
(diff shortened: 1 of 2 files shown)


snakefoot:
@smbecker Thank you for an excellent unit-test. Instead of adding an extra cast to the system, then we can move the check into `SerializeTypeCodeValue` (Handles simple types):

https://github.com/NLog/NLog/blob/7e4f29d9352d23c5facca49385993884f3fa0127/src/NLog/Targets/DefaultJsonSerializer.cs#L466-L473

Where the `QuoteValue` should be changed into:

```c#
if (SkipQuotes(objTypeCode)) 
{ 
    destination.Append(str); 
} 
else 
{ 
     if (objTypeCode == TypeCode.Char)
     {
        destination.Append('"');
        AppendStringEscape(destination, str, options.EscapeUnicode);
        destination.Append('"');
     }
     else
     {
        QuoteValue(destination, str);
     }
} 

```

And then also fix this line:

https://github.com/NLog/NLog/blob/7e4f29d9352d23c5facca49385993884f3fa0127/src/NLog/Targets/DefaultJsonSerializer.cs#L119

Which should also check for `TypeCode.Char`:

```c#
if (objTypeCode != TypeCode.Object && objTypeCode != TypeCode.Char && StringHelpers.IsNullOrWhiteSpace(options.Format) && options.FormatProvider == null)
```

smbecker:
Made the requested changes

304NotModified:
Thanks!

PR #2755, opened 2018-06-01
Logging with mmm as intended (vs.mmmm)

stic:
I guess it's faster to move and then divide, but if you look at the output you will currently get 4 digits vs expected millisecond.


304NotModified:
I think we have to update the docs instead of the code for this, as this is a behavior change

304NotModified:
See: 

image

stic:
Will do

304NotModified:
thanks!

PR #2756, opened 2018-06-01
Updating comments

stic:
Doesn't seem like a correct default behaviour (I guess you either need ms or ns) but as pointed out it would be breaking change and docs are defining longdate as ss.ffff already


304NotModified:
thanks!

PR #2761, opened 2018-06-06
${Callsite} fix class naming when includeNamespace=false and cleanNamesOfAnonymousDelegates=true

### LINKED ISSUE #2762: Callsite cleanNamesOfAnonymousDelegates doesn't work when includeNamespace is set to false

QUESTIONS are preferred on StackOverflow. You could expect a faster response there (but do include all the info below). Use the tag "nlog" https://stackoverflow.com/questions/ask

**NLog version**: 4.5.6

**Platform**: .Net 4.6.1

Currently when callsite layout renderer's ```includeNamespace``` parameter is set to **false** and ```cleanNamesOfAnonymousDelegates``` parameter is set to **true**, the outcome of callsite will be something like ```<>c```.
Below is a test that will fail.

```c#
[Fact]
public void When_NotIncludeNameSpace_Then_CleanAnonymousDelegateClassNameShouldReturnParentClassName()
{
		<nlog>
			<targets><target name='debug' type='Debug' layout='${callsite:ClassName=true:MethodName=false:IncludeNamespace=false:CleanNamesOfAnonymousDelegates=true}' /></targets>
			<rules>
				<logger name='*' levels='Fatal' writeTo='debug' />
			</rules>
		</nlog>");

	ILogger logger = LogManager.GetLogger("A");

	bool done = false;
	ThreadPool.QueueUserWorkItem(
		state =>
		{
			logger.Fatal("message");
			done = true;
		},
		null);

	while (done == false)
	{
		Thread.Sleep(10);
	}

	if (done == true)
	{
		AssertDebugLastMessage("debug", "CallSiteTests");
	}
}
```
My suggestion is to append additional logic to check if the class is nested and compiler generated, and if so, use its ```DeclaringType``` for naming.
Created PR for this: https://github.com/NLog/NLog/pull/2761

Azatey:
Fixed Callsite class name of compiler generated type when includeNamespace=false and cleanNamesOfAnonymousDelegates=true.

Fixes #2762

### DIFF
--- a/src/NLog/Internal/StackTraceUsageUtils.cs
+++ b/src/NLog/Internal/StackTraceUsageUtils.cs
@@ -115,6 +115,14 @@ namespace NLog.Internal
                 }
             }
 
+            if (!includeNameSpace
+                && callerClassType?.DeclaringType != null
+                && callerClassType.IsNested
+                && callerClassType.GetCustomAttribute<CompilerGeneratedAttribute>() != null)
+            {
+                return callerClassType.DeclaringType.Name;
+            }
+
             string className = includeNameSpace ? callerClassType?.FullName : callerClassType?.Name;
 
             if (cleanAnonymousDelegates && className != null)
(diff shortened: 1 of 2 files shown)


304NotModified:
Thanks!!

PR #2763, opened 2018-06-07
Throw better error when target name is null

masters3d:
If target.Name is null the rest of the code throws some weird error about 'key' being null which is unrelated.


304NotModified:
Thanks!

Issue #522, opened 2015-01-24
Feature request: adding parameter type to database target parameters


304NotModified:
Feature request,

adding parameter type to parameters of [Database target](https://github.com/NLog/NLog/wiki/Database-target). Example:

``` xml
 <parameter layout="Layout" name="String" precision="Byte"
    type="DType.Int32" <!-- new -->
         scale="Byte" size="Integer"/>
```

Using [DBType](https://msdn.microsoft.com/en-us/library/system.data.dbtype%28v=vs.110%29.aspx)

Everything is now a string/varchar because `Layouts` return a `string`

Current code

``` csharp
  string stringValue = par.Layout.Render(logEvent);

  p.Value = stringValue;
  command.Parameters.Add(p);
```

SergeyChanaev:
Waiting for this feature!
For some reason in case you use MS SQL and stored procedure -> date parameter will be passed correctly to the procedure. Same set up, but using PostgreSQL is complaining, that function (PostgreSQL analog of Stored Procedure) is expecting parameter of type "text" instead of type "timestamp without time zone" (PostgreSQL analog of DType.DateTime)

304NotModified:
This would be a nice feature, but other features have higher prio (e.g. structured logging). 

If someone likes to help with this feature, please let us know!

damienbod:
Support for type Json would be great

304NotModified:
json to database? Is there a special DBType form from .NET?

damienbod:
It would be great if the logs could be sent to a PostgreSQL NoSQL which requires a Json Type.

http://www.npgsql.org/doc/types/basic.html?q=json

The default target for this is string, but should be sent as a json type and not a text type.  It would be nice if this was possible using the database target. Otherwise I could always write a new target. 

?

 Greetings Damien

304NotModified:
I prefer extending the target. PR welcome ;)

ObikeDev:
Added pull request #2767

304NotModified:
To do:

- [ ] Complete #2849
- [ ] get all the good ideas from https://github.com/NLog/NLog/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aopen+label%3Adatabase-target+milestone%3A4.6+

304NotModified:
done, in NLog 4.6 RC

PR #2767, opened 2018-06-09
Issue 522: Added DbType 

ObikeDev:
**Issue #522**
Added new DbType to Parameter for changing the underlying database type when utilizing DatabaseTarget.

TargetConfigurationTests.DatabaseParameterTest was added to verify configuration of parameters.
DatabaseTargetTests.ParameterFacetTest was updated to verify setting of DbType.

### DIFF
--- a/src/NLog/Targets/DatabaseParameterInfo.cs
+++ b/src/NLog/Targets/DatabaseParameterInfo.cs
@@ -104,8 +104,8 @@ namespace NLog.Targets
         /// Gets or sets the database parameter type.
         /// </summary>
         /// <docgen category='Parameter Options' order='10' />
-        [DefaultValue(DbType.AnsiString)]
-        public DbType DbType { get; set; }
+        [DefaultValue(null)]
+        public DbType? DbType { get; set; }
     }
 }
 
--- a/src/NLog/Targets/DatabaseParameterInfo.cs
+++ b/src/NLog/Targets/DatabaseParameterInfo.cs
@@ -31,11 +31,12 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
-#if !SILVERLIGHT
+#if !SILVERLIGHT && !__IOS__ && !__ANDROID__
 
 namespace NLog.Targets
 {
     using System.ComponentModel;
+    using System.Data;
     using Config;
     using Layouts;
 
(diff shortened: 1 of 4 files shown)


Cable321:
What needs to happen to push this feature through at this point? Happy to help.

304NotModified:
Thanks @ObikeDev  and @Cable321 

I'm currently on holiday.

It would help if we could test this, in and unit test or just by installing it.

src/NLog/Targets/DatabaseParameterInfo.cs
```diff
@@ -31,11 +31,12 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
-#if !SILVERLIGHT
+#if !SILVERLIGHT && !__IOS__ && !__ANDROID__
```
304NotModified: note: in sync with database target

src/NLog/Targets/DatabaseParameterInfo.cs
```diff
@@ -98,6 +99,13 @@ public DatabaseParameterInfo(string parameterName, Layout parameterLayout)
         /// <docgen category='Parameter Options' order='10' />
         [DefaultValue(0)]
         public byte Scale { get; set; }
+
+        /// <summary>
+        /// Gets or sets the database parameter type.
+        /// </summary>
+        /// <docgen category='Parameter Options' order='10' />
+        [DefaultValue(DbType.AnsiString)]
```
304NotModified: I think this is inconsistent with

The default is NVarChar. (https://msdn.microsoft.com/en-us/library/system.data.sqlclient.sqlparameter.dbtype(v=vs.110).aspx)

Which is `String` IMO (see https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/sql-server-data-type-mappings)

Proposal: make nullable and only set if non-null in databasetarget

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -815,6 +815,11 @@ private void AddParametersToCommand(IDbCommand command, IList<DatabaseParameterI
                     p.Scale = par.Scale;
                 }
 
+                if (par.DbType != DbType.AnsiString)
```
304NotModified: see proposal commit before

304NotModified:
I think this isn't backwardscompatible with the default, so please make it optional (nullable)

304NotModified:
While the change looks good, I don't think this works.

as the current values are `string`s, and there is no parse, I don't think the SqlParameter converts the string to e.g. a int.

We need a parse or a "rawvalue", see #2849

304NotModified:
This PR is superseded by #2899 and is expected to be released in NLog 4.6

Issue #2764, opened 2018-06-07
 NLog.UnitTests.LayoutRenderers.IdentityTests.WindowsIdentityTest is broken


304NotModified:
Fails all the time, see https://ci.appveyor.com/project/nlog/nlog/build/4.5.0-7900

```
dotnet : [xUnit.net 00:01:16.4494583]     NLog.UnitTests.LayoutRenderers.IdentityTests.WindowsIdentityTest [FAIL]
At C:\projects\nlog\run-tests.ps1:25 char:1
+ dotnet test .\tests\NLog.UnitTests\  --configuration release --framew ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: ([xUnit.net 00:0...tityTest [FAIL]:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
Failed   NLog.UnitTests.LayoutRenderers.IdentityTests.WindowsIdentityTest
Error Message:
 Assert.Equal() Failure
           (pos 0)
Expected: APPVYR-WIN\appveyor
Actual:   
           (pos 0)
Stack Trace:
```

snakefoot:
Seems this test is no longer constantly failing.

PR #2788, opened 2018-06-26
${Assembly-version} make GetAssembly protected and virtual

alexangas:
Pre-requisite for inheritance of AssemblyVersionLayoutRenderer in NLog.Web 

This provides the support required for NLog/NLog.Web#293 to progress by making `GetAssembly()` virtual.

It also sets File and Informational versions used by tests so they can be accurately differentiated from Assembly version.


alexangas:
The failed test in the PR build does not appear to be related to these changes but let me know if I broke something and I will fix it!

304NotModified:
Thanks!

The broken unit test is indeed something else (it's #2764)

PR #2797, opened 2018-07-07
Better performance with Activator.CreateInstance

tangdf:
`Activator.CreateInstance` is faster than `ConstructorInfo.Invoke`

Test code:
```c#
   public  class CreateInstanceBenchamrker
   {

       private static readonly Type s_type = typeof(TestClass);


       [Benchmark]
        public object CreateInstance_By_FactoryHelper()
        {
            return FactoryHelper.CreateInstance(s_type);
        }

       [Benchmark]
       public object CreateInstance_By_Activator()
       {
           return Activator.CreateInstance(s_type);
       }
    }


    public class TestClass
    {

    }
```

``` ini

BenchmarkDotNet=v0.10.14, OS=Windows 10.0.14393.2339 (1607/AnniversaryUpdate/Redstone1)
Intel Core i5-8400 CPU 2.80GHz (Coffee Lake), 1 CPU, 6 logical and 6 physical cores
Frequency=2742183 Hz, Resolution=364.6730 ns, Timer=TSC
.NET Core SDK=2.2.100-preview1-009015
  [Host]     : .NET Core 2.1.0 (CoreCLR 4.6.26515.07, CoreFX 4.6.26515.06), 64bit RyuJIT
  DefaultJob : .NET Core 2.1.0 (CoreCLR 4.6.26515.07, CoreFX 4.6.26515.06), 64bit RyuJIT


```
|                          Method |      Mean |     Error |    StdDev |
|-------------------------------- |----------:|----------:|----------:|
| CreateInstance_By_FactoryHelper | 147.57 ns | 0.4914 ns | 0.4597 ns |
|     CreateInstance_By_Activator |  37.62 ns | 0.3379 ns | 0.3160 ns |

### DIFF
--- a/src/NLog/Internal/FactoryHelper.cs
+++ b/src/NLog/Internal/FactoryHelper.cs
@@ -47,14 +47,13 @@ namespace NLog.Internal
 
         internal static object CreateInstance(Type t)
         {
-            ConstructorInfo constructor = t.GetConstructor(ArrayHelper.Empty<Type>());
-            if (constructor != null)
+            try
             {
-                return constructor.Invoke(ArrayHelper.Empty<object>());
+                return Activator.CreateInstance(t);
             }
-            else
+            catch (MissingMethodException exception)
             {
-                throw new NLogConfigurationException($"Cannot access the constructor of type: {t.FullName}. Is the required permission granted?");
+                throw new NLogConfigurationException($"Cannot access the constructor of type: {t.FullName}. Is the required permission granted?", exception);
             }
         }
     }


304NotModified:
Also backed here: https://stackoverflow.com/a/9789971/201303

any idea why it's faster?

tangdf:
Expressions are the best performance, but typically, we need to add a dictionary to cache delegates, so the performance is very close.

The reason for using `Activatator.Createinstance` is that it is very lean.

I'm so sorry, i have no idea  why it's faster!


```c#
    public class CreateInstanceBenchamrker
    {
        private static readonly Type s_type = typeof(TestClass);

        [Benchmark]
        public object CreateInstance_By_FactoryHelper() {
            return FactoryHelper.CreateInstance(s_type);
        }

        [Benchmark]
        public object CreateInstance_By_Activator() {
            return Activator.CreateInstance(s_type);
        }


        [Benchmark]
        public object CreateInstance_By_Expression() {
            return ExpressionFactoryHelper.CreateInstance(s_type);
        }
    }

    public static class ExpressionFactoryHelper
    {
        private static readonly ConcurrentDictionary<Type, Func<object>> s_cache = new ConcurrentDictionary<Type, Func<object>>();

        internal static object CreateInstance(Type t) {

            return s_cache.GetOrAdd(t, CreateFunc).Invoke();
        }

        private static Func<object> CreateFunc(Type t) {

            ConstructorInfo constructor = t.GetConstructor(Array.Empty<Type>());
            if (constructor != null) {
                var newExpression = Expression.New(constructor);
                var convertExpression = Expression.Convert(newExpression,typeof(object));
                return Expression.Lambda<Func<object>>(convertExpression).Compile();
            }
            else {
                return () => throw new InvalidOperationException(
                    $"Cannot access the constructor of type: {t.FullName}. Is the required permission granted?");
            }
        }
    }
```

```
|                          Method |      Mean |     Error |    StdDev |
|-------------------------------- |----------:|----------:|----------:|
| CreateInstance_By_FactoryHelper | 209.61 ns | 0.9511 ns | 0.8897 ns |
|     CreateInstance_By_Activator |  54.38 ns | 0.1308 ns | 0.1224 ns |
|    CreateInstance_By_Expression |  40.09 ns | 0.2844 ns | 0.2521 ns |

304NotModified:
OK thanks! Great results!

PR #2804, opened 2018-07-13
FileTarget - Do not trust Last File Write TimeStamp when AutoFlush=false

snakefoot:
Improve archive timestamp logic when AutoFlush = false. Where delayed flush can cause Last-File-Write-TimeStamp to be "wrongly" later than previous logevent-timestamp.


304NotModified:
Thanks again!

PR #2809, opened 2018-07-21
MethodCallTarget - Support for Lamba method

### LINKED ISSUE #2780: Add a log listener that is a callback to a lambda method
Is it possible to add a log listener that is a lambda method?

snakefoot:
Resolves #2780 

 Added constructor to allow fluent configuration

Not sure if it should be a declared delegate, or just leave it as an Action-delegate.

### DIFF
--- a/src/NLog/Targets/MethodCallTarget.cs
+++ b/src/NLog/Targets/MethodCallTarget.cs
@@ -35,10 +35,8 @@ namespace NLog.Targets
 {
     using System;
     using System.Reflection;
-    using System.Collections.Generic;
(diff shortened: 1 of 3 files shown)


304NotModified:
> Not sure if it should be a declared delegate, or just leave it as an Action-delegate.

Doubting about this also. https://stackoverflow.com/questions/2282476/actiont-vs-delegate-event

snakefoot:
Updated documentation:

https://github.com/NLog/NLog/wiki/MethodCall-target

PR #2814, opened 2018-07-26
LoggingConfiguration - Improves CheckUnusedTargets to handle target wrappers

snakefoot:
Moved the logic from XmlLoggingConfiguration to LoggingConfiguration (Not specific to Xml-config-file)

Maybe consider to perform the check on assignment of LoggingConfiguration to LogFactory (Maybe for NLog 5.0)


304NotModified:
Hi,

is this only a refactor, or a refactor and change? 

(the title says something else than the summary of this PR)

snakefoot:
It improves the check for unused targets to recognize target wrappers (targets async=true)

And moves the updated method to the base class.


Sent from my Samsung device

PR #2825, opened 2018-07-31
${AssemblyVersion}: add format  option

### LINKED ISSUE #2617: Feature request: Allow configuration of assembly-version string
If I use `${assembly-version}` in NLog.config then it gives me a version of format `Major.Minor.Patch.Build`.
However, management wants only to use the first 2 digits, therefore being able to customize this string would be helpful. Some commonly used options like `Major.Minor.Patch` and `Major.Minor` would be sufficient I think instead of allowing full customizability.

chefarbeiter:
Fixes #2617

### DIFF
--- a/src/NLog/LayoutRenderers/AssemblyVersionLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/AssemblyVersionLayoutRenderer.cs
@@ -60,6 +60,7 @@ namespace NLog.LayoutRenderers
         public AssemblyVersionLayoutRenderer()
         {
             Type = AssemblyVersionType.Assembly;
+            Format = DefaultFormat;
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/NLog/LayoutRenderers/AssemblyVersionLayoutRenderer.cs
```diff
@@ -118,6 +128,24 @@ protected override void Append(StringBuilder builder, LogEventInfo logEvent)
             builder.Append(version);
         }
 
+        private string ApplyFormatToVersion(string version)
+        {
+            var lowerFormat = Format.ToLowerInvariant();
+
+            if (lowerFormat == DefaultFormat || string.IsNullOrEmpty(version))
```
304NotModified: for performance reasons, we prefer:

- do the toLowerInvariant in the setter of Format.
- use OrdinalIgnoreCase  comparison. 

Could you please change that? Thanks!

src/NLog/LayoutRenderers/AssemblyVersionLayoutRenderer.cs
```diff
@@ -81,6 +82,15 @@ public AssemblyVersionLayoutRenderer()
         [DefaultValue(nameof(AssemblyVersionType.Assembly))]
         public AssemblyVersionType Type { get; set; }
 
+        private const string DefaultFormat = "major.minor.build.revision";
+
+        /// <summary>
+        /// Gets or sets the custom format of the assembly version output.
```
304NotModified: please describe the options for the format

bergmeister:
@304NotModified I would really like to start using this feature soon. Do you know roughly when the next release will be?

304NotModified:
Not planned yet, I hope within 4 weeks

bergmeister:
Thanks, that is good enough

PR #2830, opened 2018-08-01
Auto loading of assemblies was broken in some cases

snakefoot:
Use NLog-assembly as primary location (Instead of AppDomain-BaseDirectory). Fixes bug introduced with #2643 (NLog ver. 4.5.2)

### DIFF
--- a/src/NLog/Config/ConfigurationItemFactory.cs
+++ b/src/NLog/Config/ConfigurationItemFactory.cs
@@ -366,23 +366,26 @@ namespace NLog.Config
             try
             {
                 var assemblyLocation = GetAssemblyFileLocation(nlogAssembly);
-                var extensionDlls = GetNLogExtensionFiles(assemblyLocation);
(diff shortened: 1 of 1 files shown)


snakefoot:
@304NotModified I guess it is important to get a new NLog.Web + NLog.AspNetCore released, that forces update to NLog 4.5.8 (when released). As it will ensure a more smooth upgrade path away from this bug.

304NotModified:
> @304NotModified I guess it is important to get a new NLog.Web + NLog.AspNetCore released, that forces update to NLog 4.5.8 (when released). As it will ensure a more smooth upgrade path away from this bug.

Thanks for the suggestion. Agree!

snakefoot:
Please notice that my unit test #2833 is missing a check in GetAutoLoadingFileLocations () so it does return path multiple times

Something like this:

if entrylocation != null && assemblylocation != entry location
  yield return

Same for AppDomain BaseDirectory




Sent from my Samsung device

PR #2835, opened 2018-08-02
Unit test after fixing ConfigurationItemFactory auto loading of assemblies

snakefoot:
See also #2830

### DIFF
--- a/tests/NLog.UnitTests/Config/ExtensionTests.cs
+++ b/tests/NLog.UnitTests/Config/ExtensionTests.cs
@@ -389,6 +389,7 @@ namespace NLog.UnitTests.Config
                 Assert.NotEmpty(fileLocations);
                 Assert.NotNull(fileLocations[0].Key);
                 Assert.NotNull(fileLocations[0].Value); // Primary search location is NLog-assembly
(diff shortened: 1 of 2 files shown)


tests/NLog.UnitTests/Config/ExtensionTests.cs
```diff
@@ -389,6 +389,7 @@ public void Extension_should_be_auto_loaded_when_following_NLog_dll_format()
                 Assert.NotEmpty(fileLocations);
                 Assert.NotNull(fileLocations[0].Key);
                 Assert.NotNull(fileLocations[0].Value); // Primary search location is NLog-assembly
+                Assert.Equal(fileLocations.Length, fileLocations.Select(f => f.Key).Distinct().Count());
```
304NotModified: another option is to group-by and check if there aren't duplicate items, isn't?

304NotModified:
Thanks!

tests/NLog.UnitTests/Config/ExtensionTests.cs
snakefoot: If it is more readable for you, then please change it. Don't have an opinion.

PR #2846, opened 2018-08-12
Include Entry Assembly File Location when loading candidate NLog.config

### LINKED ISSUE #1342: Config file not found when running via .NET Core

snakefoot:
Improve the NetCore experience where the BaseDirectory is often different from the bin-location (But the Bin-location is where nlog.config is copied to by default)

Also improve the logic for recognizing identical paths (with or without ending directory-separator)

Think this will resolve #1342 (Atleast for NetStandard1.5 and NetStandard2.0)

### DIFF
--- a/src/NLog/Internal/AssemblyHelpers.cs
+++ b/src/NLog/Internal/AssemblyHelpers.cs
@@ -107,5 +107,81 @@ using System.Windows;
 #endif
         }
 
+#if !SILVERLIGHT && !NETSTANDARD1_3
+        public static string GetAssemblyFileLocation(Assembly assembly)
+        {
+            string fullName = string.Empty;
+
+            try
+            {
+                if (assembly == null)
+                {
+                    return string.Empty;
+                }
+
+                fullName = assembly.FullName;
+
+#if NETSTANDARD
+                if (string.IsNullOrEmpty(assembly.Location))
+                {
+                    // Assembly with no actual location should be skipped (Avoid PlatformNotSupportedException)
+                    InternalLogger.Warn("Ignoring assembly location because location is empty: {0}", fullName);
+                    return string.Empty;
+                }
+#endif
+
+                Uri assemblyCodeBase;
+                if (!Uri.TryCreate(assembly.CodeBase, UriKind.RelativeOrAbsolute, out assemblyCodeBase))
+                {
+                    InternalLogger.Warn("Ignoring assembly location because code base is unknown: '{0}' ({1})", assembly.CodeBase, fullName);
+                    return string.Empty;
+                }
+
+                var assemblyLocation = Path.GetDirectoryName(assemblyCodeBase.LocalPath);
+                if (string.IsNullOrEmpty(assemblyLocation))
+                {
+                    InternalLogger.Warn("Ignoring assembly location because it is not a valid directory: '{0}' ({1})", assemblyCodeBase.LocalPath, fullName);
+                    return string.Empty;
+                }
+
+                DirectoryInfo directoryInfo = new DirectoryInfo(assemblyLocation);
+                if (!directoryInfo.Exists)
+                {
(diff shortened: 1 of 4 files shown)


src/NLog/Internal/AssemblyHelpers.cs
```diff
@@ -107,5 +107,81 @@ public static Assembly LoadFromName(string assemblyName)
 #endif
         }
 
+#if !SILVERLIGHT && !NETSTANDARD1_3
+        public static string GetAssemblyFileLocation(Assembly assembly)
```
304NotModified: is this just a move or are there some changes here?

src/NLog/LogFactory.cs
```diff
@@ -1092,15 +1093,51 @@ public void ResetCandidateConfigFilePath()
         private static IEnumerable<string> GetDefaultCandidateConfigFilePaths()
         {
             // NLog.config from application directory
-            if (CurrentAppDomain?.BaseDirectory != null)
+            string baseDirectory = PathHelpers.TrimDirectorySeparators(CurrentAppDomain?.BaseDirectory);
+            if (!string.IsNullOrEmpty(baseDirectory))
+                yield return Path.Combine(baseDirectory, "NLog.config");
+
+            bool platformFileSystemCaseInsensitive = PlatformDetector.IsWin32;
+            if (!platformFileSystemCaseInsensitive && !string.IsNullOrEmpty(baseDirectory))
```
304NotModified: PS: windows could be case sensitive, but highly unlikely ;)

src/NLog/LogFactory.cs
```diff
@@ -1092,15 +1093,51 @@ public void ResetCandidateConfigFilePath()
         private static IEnumerable<string> GetDefaultCandidateConfigFilePaths()
         {
             // NLog.config from application directory
-            if (CurrentAppDomain?.BaseDirectory != null)
+            string baseDirectory = PathHelpers.TrimDirectorySeparators(CurrentAppDomain?.BaseDirectory);
+            if (!string.IsNullOrEmpty(baseDirectory))
```
304NotModified: please move the check `string.IsNullOrEmpty(baseDirectory)` on top of both `if`s - it's now duplicated

(so we have 3 ifs, 1 top and 2 nested)

snakefoot: Well I like to skip the `PlatformDetector.IsWin32` when possible.

src/NLog/Internal/AssemblyHelpers.cs
snakefoot: It is a move and some change in the internallogger output

src/NLog/LogFactory.cs
snakefoot: Would cause most applications on Windows to fail badly, which is why I skip the extra file-checks

snakefoot:
Removed the usage of NLog-assembly-location for NLog.config loading, as NetCore will most likely reference NLog.dll in nuget-folder.

src/NLog/LogFactory.cs
304NotModified: I'm a bit confused, currently it tests the `PlatformDetector.IsWin32` before the baseDirectory check, so it isn't skipped when possible?

To be clear, the proposal:


```c#

if (!string.IsNullOrEmpty(baseDirectory))
{
    yield return Path.Combine(baseDirectory, "NLog.config");
    bool platformFileSystemCaseInsensitive = PlatformDetector.IsWin32;
    if (!platformFileSystemCaseInsensitive)
    {
         yield return Path.Combine(baseDirectory, "nlog.config");
    }

}

```

`baseDirectory` was checked before, so no reason to check it again IMO

snakefoot: PlatformDetector.IsWin32 is only called once in the current implementation. Your suggestions requires multiple lookups.

304NotModified: OK, Fair enough

PR #1435, opened 2016-05-04
Allow configuring type of parameters for the database target (including fix for Oracle)

304NotModified:
TODO:
- [ ] use the DB type configured in the config instead of typescanning
- [ ] scan for type at init, not at write
- [ ] implement for all DB types: 
  - [ ]  System.Data.EntityClient.EntityParameter
  - [ ] System.Data.Odbc.OdbcParameter
  - [ ] System.Data.OleDb.OleDbParameter
  - [ ] System.Data.OracleClient.OracleParameter
  - [ ] System.Data.SqlClient.SqlParameter
- [ ] parse DB type when parsing config (not on logging)
- [ ] check if parse type is needed
- [ ] create testcase for at least int, bit and oracletype
- [ ] create testcase with .mdf file.
- [ ] add integration tests .

Discuss:
- Maybe 2nd parameter to distinguish DBType vs **x**DbType (e.g. SqlDbType, OracleDbType)?

supersedes https://github.com/NLog/NLog/pull/1194

thanks to @sorvis

---

This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/1435)

Issue #293, opened 2014-01-06
Possible to insert a null into a database column, using Configuration API?


jacquibo:
I don't seem to be able to find a way to insert a null into a database column, using the Configuration API. I have some columns that are not always required, so for some logging will be nulls.

The only way I have found to do this is to use a stored procedure, and pass empty strings to it for any nulls. The stored procedure then puts in null whenever there is an empty string.

Could this feature be added in some way? e.g.

```
var param = new DatabaseParameterInfo();
param.Name = "foo";
param.Layout = null;
target.Parameters.Add(param);
```

304NotModified:
Hi @jacquibo,

@Xharze as far I can see, most `Targets` assume that the Layout renderer won't return `null`. This makes this feature request a bit tricky.

I was thinking to create a special Layout renderer for the `DatabaseTarget`, e.g. `${DBNull}`. 

But this is a bit strange of course, because it's not available on other layout renderers.  What do you think?

Xharze:
I think the best solution if we could allow the layout renderers to return
null in the database target, but I don't know if it's possible without
affecting other targets.
The reason is that you would probably need to use ifs in the layout to
achieve the desired effect if inserting null in different scenarios.

304NotModified:
I think returning `null` by the layout renderes had too much impact. 

To summarize the possible solutions. 

Possible solutions:
1. Create a special Layout renderer (e.g. `DBNullLayoutRenderer`) for the all targets, e.g. `${DBNull}`. - not possible, because `DBNull.Value` isn't a string.
2. Allow Layout return `null` . Unknown impact, will infect also custom Targets. 
3. Create a special Layout renderer (e.g. `DBNullLayoutRenderer`), only for `DatabaseTarget`,. - not possible, got exception, layout renderer not found.
4. Use a new property on `DatabaseParameterInfo`: `IsNull` (`bool`) - ugly
5. Use a fixed string, which isn't in layout renderer format: e.g. `$DBNull` and convert that one to `null` in `DatabaseTarget` - ugly

Non of them are perfect. Any thoughts on this @Xharze .

realityexists:
FWIW, I ran into this as well, but found a simple workaround for my particular case: ISNULL in the insert command (I'm using PostgreSQL). This doesn't allow distinguishing between empty string and null, but for my scenario it's OK. I want to store the exception text if there is an exception, NULL if there is no exception. So my config looks like this:
```xml
      <commandText>
[...]

PR #2849, opened 2018-08-14
IRawValue + dbNullLayoutRenderer

### LINKED ISSUE #1332: [Feature Request] Support for something like log4net IRawLayout for targets that need objects.
- Why do we need it?
  I'm currently working to create an NLog target for a logging system that takes objects in a specific structure.  Many of the properties on the structure are non string (e.g. int, DateTime, etc.), and right now if I would like to take advantage of the Layout renderers to provide a bit more flexibility to users from the configuration I will need to take the string produced by the renderer and parse it back into the appropriate type.

In many cases such as the MDC or MDLC, the structure has all the properties I need in the proper types, so I can just manually lookup the properties myself, but if the MDC layout renderer supported a way to get the raw object version of a property I could skip the costly parsing and also my code won't need to checking in many different locations for values.

```
public interface IRawLayout
{
    public object RenderRaw(LogEventInfo logEvent);
}
```

Anything that supports IRawLayout should essentially be able to fall back to ILayout by just ToStringing the object, but we could also derive IRawLayout from ILayout to force the implementers to ensure they provide for the lowest common denominator.

I haven't looked TOO deeply into how this would work between the Layout/Renderer layer, so I'm not sure how easy it would be, but wanted to start a conversation to find out if this is something that has been discussed already.

log4net `IRawLayout`: https://svn.apache.org/viewvc/logging/log4net/trunk/src/Layout/IRawLayout.cs?view=markup

304NotModified:
needed for #522 (see also #1435 or #2767)

Not sure if this is enough. How to log ${level} as int to the database? We like to prevent toString and int.parse.

options: 

1. database target aware of LogLevel type
2. or make IRawValue generic, e.g. IRawValue<LogLevel>, IRawValue<int>
3. or request type to GetRawValue, e.g. GetRawValue(typeof(int))  - could be optional.

todo decide

- [x] naming of getRawValue and IRawValue
- [x] also work with TypeCode? (provided or request), really useful is TypeCode.DBNull for the database target, see also https://github.com/NLog/NLog/issues/293
- [x] interface (IRawValue) or abstract class? Later for performance reasons. (also possible, virtual method on layout class, but could be tricky with null)

fixes #1332

todo 
 
- [x] impl for all needed layout renderers 
- [x] unit tests

### DIFF
--- a/src/NLog/LayoutRenderers/CounterLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/CounterLayoutRenderer.cs
@@ -43,7 +43,7 @@ namespace NLog.LayoutRenderers
     /// A counter value (increases on each layout rendering).
     /// </summary>
     [LayoutRenderer("counter")]
-    public class CounterLayoutRenderer : LayoutRenderer
+    public class CounterLayoutRenderer : LayoutRenderer, IRawValue
     {
         private static Dictionary<string, int> sequences = new Dictionary<string, int>();
 
@@ -83,6 +83,18 @@ namespace NLog.LayoutRenderers
         /// <param name="logEvent">Logging event.</param>
         protected override void Append(StringBuilder builder, LogEventInfo logEvent)
         {
+            int v = GetNextValue(logEvent);
+
+            builder.AppendInvariant(v);
+        }
+
+        /// <summary>
+        /// Get raw value, updates the sequence
+        /// </summary>
+        object IRawValue.GetRawValue(LogEventInfo logEvent) => GetNextValue(logEvent);
+
+        private int GetNextValue(LogEventInfo logEvent)
+        {
             int v;
 
             if (Sequence != null)
@@ -95,7 +107,7 @@ namespace NLog.LayoutRenderers
                 Value += Increment;
             }
 
-            builder.AppendInvariant(v);
+            return v;
         }
 
         private static int GetNextSequenceValue(string sequenceName, int defaultValue, int increment)
--- a/src/NLog/LayoutRenderers/EventPropertiesLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/EventPropertiesLayoutRenderer.cs
@@ -46,7 +46,7 @@ namespace NLog.LayoutRenderers
     [ThreadAgnostic]
     [ThreadSafe]
     [MutableUnsafe]
-    public class EventPropertiesLayoutRenderer : LayoutRenderer
+    public class EventPropertiesLayoutRenderer : LayoutRenderer, IRawValue
     {
         /// <summary>
         ///  Log event context data with default options.
@@ -83,12 +83,26 @@ namespace NLog.LayoutRenderers
         /// <param name="logEvent">Logging event.</param>
         protected override void Append(StringBuilder builder, LogEventInfo logEvent)
         {
-            object value;
-            if (logEvent.HasProperties && logEvent.Properties.TryGetValue(Item, out value))
+            if (GetValue(logEvent, out var value))
             {
                 var formatProvider = GetFormatProvider(logEvent, Culture);
                 builder.AppendFormattedValue(value, Format, formatProvider);
             }
+        }
+
+        private bool GetValue(LogEventInfo logEvent, out object value)
+        {
+            value = null;
+            return logEvent.HasProperties && logEvent.Properties.TryGetValue(Item, out value);
+        }
+
+        /// <summary>
+        /// Get raw value, updates the sequence
+        /// </summary>
+        object IRawValue.GetRawValue(LogEventInfo logEvent)
+        {
+            GetValue(logEvent, out var value);
+            return value;
         }
     }
 }
--- a/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
@@ -49,7 +49,7 @@ namespace NLog.LayoutRenderers
     [LayoutRenderer("exception")]
     [ThreadAgnostic]
     [ThreadSafe]
-    public class ExceptionLayoutRenderer : LayoutRenderer
+    public class ExceptionLayoutRenderer : LayoutRenderer, IRawValue
     {
         private string _format;
         private string _innerFormat = string.Empty;
@@ -177,6 +177,11 @@ namespace NLog.LayoutRenderers
             get;
             private set;
         }
+
+        /// <summary>
+        /// Get raw value, updates the sequence
+        /// </summary>
+        object IRawValue.GetRawValue(LogEventInfo logEvent) => logEvent.Exception;
 
         /// <summary>
         /// Renders the specified exception information and appends it to the specified <see cref="StringBuilder" />.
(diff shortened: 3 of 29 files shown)


src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -178,6 +178,11 @@ public List<ExceptionRenderingFormat> InnerFormats
             private set;
         }
 
+        /// <summary>
+        /// Get raw value, updates the sequence
+        /// </summary>
+        object IRawValue.GetRawValue(LogEventInfo logEvent) => logEvent.Exception;
```
NLog-SonarQube: CRITICAL Rename parameter 'logEvent' to 'logEventInfo' to match the interface declaration. [rule](https://sonarcloud.io/coding_rules#rule_key=csharpsquid%3AS927)

src/NLog/LayoutRenderers/GuidLayoutRenderer.cs
```diff
@@ -67,13 +67,26 @@ public GuidLayoutRenderer()
         [DefaultValue(false)]
         public bool GeneratedFromLogEvent { get; set; }
 
+        /// <summary>
+        /// Get raw value
+        /// </summary>
+        object IRawValue.GetRawValue(LogEventInfo logEvent) => GetValue(logEvent);
```
NLog-SonarQube: CRITICAL Rename parameter 'logEvent' to 'logEventInfo' to match the interface declaration. [rule](https://sonarcloud.io/coding_rules#rule_key=csharpsquid%3AS927)

src/NLog/LayoutRenderers/CounterLayoutRenderer.cs
```diff
@@ -82,6 +82,18 @@ public CounterLayoutRenderer()
...
+        {
+            int v = GetNextValue(logEvent);
+
+            builder.AppendInvariant(v);
+        }
+
+        /// <summary>
+        /// Get raw value, updates the sequence
+        /// </summary>
+        object IRawValue.GetRawValue(LogEventInfo logEvent) => GetNextValue(logEvent);
```
NLog-SonarQube: CRITICAL Rename parameter 'logEvent' to 'logEventInfo' to match the interface declaration. [rule](https://sonarcloud.io/coding_rules#rule_key=csharpsquid%3AS927)

src/NLog/LayoutRenderers/GdcLayoutRenderer.cs
```diff
@@ -66,10 +66,20 @@ public class GdcLayoutRenderer : LayoutRenderer
...
-            var value = GlobalDiagnosticsContext.GetObject(Item);
+            object value = GetValue();
             var formatProvider = GetFormatProvider(logEvent, null);
             builder.AppendFormattedValue(value, Format, formatProvider);
         }
+
+        /// <summary>
+        /// Get raw value.
+        /// </summary>
+        object IRawValue.GetRawValue(LogEventInfo logEvent) => GetValue();
```
NLog-SonarQube: CRITICAL Rename parameter 'logEvent' to 'logEventInfo' to match the interface declaration. [rule](https://sonarcloud.io/coding_rules#rule_key=csharpsquid%3AS927)

src/NLog/LayoutRenderers/EventPropertiesLayoutRenderer.cs
```diff
@@ -83,12 +83,26 @@ public EventPropertiesLayoutRenderer()
...
+        private bool GetValue(LogEventInfo logEvent, out object value)
+        {
+            value = null;
+            return logEvent.HasProperties && logEvent.Properties.TryGetValue(Item, out value);
+        }
+
+        /// <summary>
+        /// Get raw value, updates the sequence
+        /// </summary>
+        object IRawValue.GetRawValue(LogEventInfo logEvent)
```
NLog-SonarQube: CRITICAL Rename parameter 'logEvent' to 'logEventInfo' to match the interface declaration. [rule](https://sonarcloud.io/coding_rules#rule_key=csharpsquid%3AS927)

NLog-SonarQube:
SonarQube analysis reported 5 issues
* ![CRITICAL][CRITICAL] 5 critical

Watch the comments in this conversation to review them.

[CRITICAL]: https://sonarsource.github.io/sonar-github/severity-critical.png 'Severity: CRITICAL'

304NotModified:
@snakefoot like your opinion on this :)

snakefoot:
@304NotModified Like the RawValue idea. Makes good sense when it comes to converting into parameter-types.

Regarding conversion of a RawValue to other types (Ex. LogLevel to Ordinal) then I guess one could reuse the type-invariance that has been made for when-filtering.

src/NLog/Layouts/SimpleLayout.cs
```diff
@@ -240,6 +240,18 @@ internal void SetRenderers(LayoutRenderer[] renderers, string text)
...
+        /// <inheritdoc />
+        public override object GetRawValue(LogEventInfo logEvent)
+        {
+
+            if (Renderers.Count == 1 && Renderers[0] is IRawValue rawValueLayoutRenderer)
+            {
+                return rawValueLayoutRenderer.GetRawValue(logEvent);
+            }
+
+            return null;
```
304NotModified: doubting if this is good enough..

If we get `null`, then we could call render again and that's not always the best solution

snakefoot: I think it is a good idea that `Layout.GetRawValue` always return null (Good default for CsvLayout + JsonLayout + etc.). Then SimpleLayout need to do the conversion if ``IRawValue`-optimization is not available.

snakefoot: Prefer that the fallback render logic is inside this method, instead of the caller having to handle `null`.

304NotModified: I changed it already to TryGetRawValue, for performance reasons.

If it returns null (out) and true, there is no need to use the string render anymore

snakefoot: Prefer the old logic, where it tries all kinds of ways, and then fallback to a provided default value.

304NotModified: but then I could call twice .render, while that's not needed

snakefoot: You are right `TryGetRawValue` will just be a single helper method, that is used by an generic-extension method that gets the default-value as input paramater and can perform rendering if needed.

src/NLog/LayoutRenderers/IdentityLayoutRenderer.cs
```diff
@@ -95,36 +95,45 @@ public IdentityLayoutRenderer()
...
-                    }
+                    builder.Append(separator);
+                    builder.Append(identity.Name);
                 }
             }
+
+        }
+
+        /// <inheritdoc />
+        object IRawValue.GetRawValue(LogEventInfo logEvent)
```
304NotModified: @snakefoot do you think this layout renderer should implement IRawValue, and if so, should use the options IsAuthenticated/AuthType/Name is the rawValue? (well then it's a string)

304NotModified: OK thanks! done

304NotModified:
some docs here: https://github.com/NLog/NLog/wiki/Database-target

PR #2850, opened 2018-08-14
NLogTraceListener - Reduce allocations for TraceData with single param

snakefoot:
And handle possible null-reference-exception

### DIFF
--- a/src/NLog/NLogTraceListener.cs
+++ b/src/NLog/NLogTraceListener.cs
@@ -39,9 +39,9 @@ namespace NLog
     using System.Collections;
     using System.ComponentModel;
     using System.Diagnostics;
-    using System.Reflection;
     using System.Text;
     using System.Xml;
+    using NLog.Internal;
 
     /// <summary>
     /// TraceListener which routes all messages through NLog.
@@ -257,7 +257,7 @@ namespace NLog
                 }
                 else
                 {
-                    var sb = new StringBuilder();
+                    var sb = new StringBuilder(data.Length * 5 - 2);
                     for (int i = 0; i < data.Length; ++i)
                     {
                         if (i > 0)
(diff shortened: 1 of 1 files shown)


src/NLog/NLogTraceListener.cs
```diff
@@ -251,20 +248,32 @@ public override void TraceData(TraceEventCache eventCache, string source, TraceE
...
             {
-                if (i > 0)
+                if (data.Length == 1)
                 {
-                    sb.Append(", ");
+                    message = "{0}";
+                }
+                else
+                {
+                    var sb = new StringBuilder();
```
304NotModified: maybe set also the initial capacity? (5*(i+1))

src/NLog/NLogTraceListener.cs
```diff
@@ -251,20 +248,32 @@ public override void TraceData(TraceEventCache eventCache, string source, TraceE
...
+                        if (i > 0)
+                        {
+                            sb.Append(", ");
+                        }
+
+                        sb.Append("{");
+                        sb.Append(i);
+                        sb.Append("}");
+                    }
+                    message = sb.ToString();
```
304NotModified: also, isn't string.join more efficient in this way?

src/NLog/NLogTraceListener.cs
snakefoot: Yes good idea.

src/NLog/NLogTraceListener.cs
snakefoot: Not sure I understand string.Join, but I have now changed to use StringBuilderExt.AppendInvariant

304NotModified: I was wrong, but for the reference: https://stackoverflow.com/questions/585860/string-join-vs-stringbuilder-which-is-faster#585897

304NotModified:
Thanks!

Issue #2852, opened 2018-08-17
NuGet Package : Add `repository` metadata. 


MaximRouiller:
Basically, we can now add a repository metadata on a nuspec file to allow the repository to be surfaced on NuGet. This allows you to have a `projectUrl` as well a repository URL now.

I've talked about it [on Twitter](https://twitter.com/MaximRouiller/status/1030468873945075712) but the most basic things to know is here:

[NuSpec Reference](https://docs.microsoft.com/en-us/nuget/reference/nuspec?WT.mc_id=social-github-marouill#repository)

Reference Implementation in Json .NET
![image](https://user-images.githubusercontent.com/209384/44277447-b1c20f00-a218-11e8-9406-4f1f05879d47.png)

304NotModified:
excellent idea, will be fixed in #2853

Thx!

MaximRouiller:
My pleasure! 😄

304NotModified:
PS it was already in the nlog package, but not in nlog.config and nlog.schema

![image](https://user-images.githubusercontent.com/5808377/44282135-59136600-a25a-11e8-8733-c083a2d71673.png)

PR #2853, opened 2018-08-17
adding <repository> to nuspec NLog.Config and NLog.Schema

304NotModified:
closes https://github.com/NLog/NLog/issues/2852


snakefoot:
Why update the nuspec-files ? They should just be deleted. Using dotnet pack and VS2017 csproj.

304NotModified:
because we need then more csproj AFAIK

snakefoot:
> because we need then more csproj files AFAIK (for nlog.schema and nlog.config).

Well then delete  all but those two

304NotModified:
you're right, done!

PR #2865, opened 2018-08-23
JSON encoding should create valid JSON for special double values (hotfix)

### LINKED ISSUE #2860: JSON serialization of doubles
**NLog version**: 4.5.6

**Platform**:  .Net 4.5

I just ran into an issue/inconsistency using JSON formatted logs. It turns out that **NaN**, **PositiveInfinity**, and **NegativeInfinity** are not valid symbols for a double according to the JSON spec,
however, they serialize as symbols in the NLog JSON Serializer.

Example output: 
`
	"Segment": {
		"Name": "Processing",
		"Requested": "2018-06-14T23:06:43.9029798Z",
		"Finished": "2018-06-14T23:07:20.2739469Z",
		"Valid": true,
		"EstimatedRemaining": NaN,
		"Elapsed": 36.3709671
	},
`

There's a lot of discussion about this in various places online. The common workarounds seem to be either converting it to a string (**"NaN"**) or replacing it with **null** or **0**. 

The reason this is having an impact is that the deserializer used in our log aggregator doesn't recognize the log message as valid JSON and refuses to parse it.

snakefoot:
Resolves #2860 by handling NaN and Infinity (Also for floats)

### DIFF
--- a/src/NLog/Targets/DefaultJsonSerializer.cs
+++ b/src/NLog/Targets/DefaultJsonSerializer.cs
@@ -126,7 +126,7 @@ namespace NLog.Targets
                     else
                     {
                         string xmlStr = XmlHelper.XmlConvertToString(value, objTypeCode);
-                        if (SkipQuotes(objTypeCode))
+                        if (SkipQuotes(value, objTypeCode))
                         {
                             return xmlStr;
                         }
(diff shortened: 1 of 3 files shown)


304NotModified:
Thanks!

PR #2883, opened 2018-08-29
Fix LoadConfiguration for not found config file

### LINKED ISSUE #2878: TryLoadLoggingConfiguration throws exception instead of just logging failure.
**NLog version**: 4.5.9

**Platform**: .Net 4.6.1

None

- What is the current result?
 NLog.LogFactory.GetLogger throws an exception if it can't find an nlog.config file.

No exception, just no configuration loaded.  "Failed loading config from ..." logged to InternalLogger.

Yes

Note: Private details changed to ... .
``` C#
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. 
---> System.TypeInitializationException: The type initializer for '...' threw an exception. 
---> System.TypeInitializationException: The type initializer for '...' threw an exception. 
---> System.TypeInitializationException: The type initializer for '...' threw an exception. 
---> System.IO.FileNotFoundException: Could not find file '...NLog.config'.
    at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options, String msgPath, Boolean bFromProxy)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize)
   at System.Xml.XmlDownloadManager.GetStream(Uri uri, ICredentials credentials, IWebProxy proxy, RequestCachePolicy cachePolicy)
   at System.Xml.XmlUrlResolver.GetEntity(Uri absoluteUri, String role, Type ofObjectToReturn)
   at System.Xml.XmlTextReaderImpl.FinishInitUriString()
   at System.Xml.XmlTextReaderImpl..ctor(String uriStr, XmlReaderSettings settings, XmlParserContext context, XmlResolver uriResolver)
   at System.Xml.XmlReaderSettings.CreateReader(String inputUri, XmlParserContext inputContext)
   at System.Xml.XmlReader.Create(String inputUri, XmlReaderSettings settings, XmlParserContext inputContext)
   at System.Xml.XmlReader.Create(String inputUri)
   at NLog.Config.XmlLoggingConfiguration.CreateFileReader(String fileName)
   at NLog.Config.XmlLoggingConfiguration..ctor(String fileName, Boolean ignoreErrors, LogFactory logFactory)
   at NLog.Config.XmlLoggingConfiguration..ctor(String fileName, LogFactory logFactory)
   at NLog.LogFactory.TryLoadLoggingConfiguration(String configFile)
   at NLog.LogFactory.TryLoadFromFilePaths()
   at NLog.LogFactory.GetLogger(LoggerCacheKey cacheKey)
   at NLog.LogFactory.GetLogger(String name)
   at NLog.LogManager.GetLogger(String name)
   at ...  

```

- Are there any workarounds? yes
Add nlog.config that is empty.

Maybe 4.5.7... removed a bad nlog.config file from a dependency that was being copied into the folder.

``` C#
[Test]
        public void LogManager_TryLoadLoggingConfiguration_WhenFileDoesNotExist_ThenDoesNotThrowException()
        {
            //Assign
            string filename = $"{Guid.NewGuid()}.config";
            File.Delete(filename);

            //Act
            NLog.LogFactory.TryLoadLoggingConfiguration(filename);            
        }
```

snakefoot:
TryLoadFromFilePaths - Improve exception handling

Resolves #2878

### DIFF
--- a/tests/NLog.UnitTests/LogFactoryTests.cs
+++ b/tests/NLog.UnitTests/LogFactoryTests.cs
@@ -95,29 +95,75 @@ namespace NLog.UnitTests
         }
 
         [Fact]
-        public void InaccessibleNLogConfigFileDoesNotThrow()
+        public void Configuration_InaccessibleNLog_doesNotThrowException()
         {
-            var tempDirectory = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString("N"));
-            var configFile = Path.Combine(tempDirectory, "NLog.config");
+            string tempDirectory = null;
 
             try
             {
-                Directory.CreateDirectory(tempDirectory);
-                File.WriteAllText(configFile, "<nlog />");
-                LogFactory logFactory = new LogFactory();
-                logFactory.SetCandidateConfigFilePaths(new[] { configFile });
-                using (var fileLock = new FileStream(configFile, FileMode.Open, FileAccess.ReadWrite, FileShare.None))
+                // Arrange
+                var logFactory = CreateEmptyNLogFile(out tempDirectory, out var configFile);
+                using (OpenStream(configFile))
                 {
+                    // Act
                     var loggingConfig = logFactory.Configuration;
+
+                    // Assert
                     Assert.Null(loggingConfig);
-                    Assert.ThrowsAny<Exception>(() => logFactory.LoadConfiguration(configFile));
                 }
+
+                // Assert
+                Assert.NotNull(logFactory.Configuration);
             }
             finally
             {
-                if (Directory.Exists(tempDirectory))
+                if (tempDirectory != null && Directory.Exists(tempDirectory))
                     Directory.Delete(tempDirectory, true);
             }
+        }
+
+        [Fact]
+        public void LoadConfiguration_InaccessibleNLog_throwException()
+        {
+            string tempDirectory = null;
+
+            try
+            {
+                // Arrange
+                var logFactory = CreateEmptyNLogFile(out tempDirectory, out var configFile);
+                using (OpenStream(configFile))
+                {
+                    // Act
+                    var ex = Record.Exception(() => logFactory.LoadConfiguration(configFile));
+
+                    // Assert
+                    Assert.IsType<IOException>(ex);
+                }
+
+                // Assert
+                Assert.NotNull(logFactory.LoadConfiguration(configFile).Configuration);
+            }
+            finally
+            {
+                if (tempDirectory != null && Directory.Exists(tempDirectory))
+                    Directory.Delete(tempDirectory, true);
+            }
+        }
+
+        private static FileStream OpenStream(string configFile)
+        {
+            return new FileStream(configFile, FileMode.Open, FileAccess.ReadWrite, FileShare.None);
+        }
+
+        private static LogFactory CreateEmptyNLogFile(out string tempDirectory, out string filePath)
+        {
+            tempDirectory = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString("N"));
+            filePath = Path.Combine(tempDirectory, "NLog.config");
+            Directory.CreateDirectory(tempDirectory);
+            File.WriteAllText(filePath, "<nlog />");
+            LogFactory logFactory = new LogFactory();
+            logFactory.SetCandidateConfigFilePaths(new[] { filePath });
+            return logFactory;
         }
 
         [Fact]


304NotModified:
so `factory.LoadConfiguration(filename);` still fails even when throwError = false? (see #2880)

snakefoot:
The difference is that `LoadConfiguration` says load this specific file at this expected path.

While `TryLoadFromFilePaths` just jumps around and loads from all kind of random locations (Without the user can do anything).

304NotModified:
true, unfortunately the `TryLoadLoggingConfiguration` isn't really a try IMO, as all the check logic is outside of it. But we could refactor that later.

snakefoot:
@304NotModified Now changed TryLoadLoggingConfiguration to return true/false

src/NLog/LogFactory.cs
```diff
@@ -330,7 +323,7 @@ private void TryLoadFromAndroidAssets()
                 {
                     if (stream != null)
                     {
-                        _config = TryLoadLoggingConfiguration(XmlLoggingConfiguration.AssetsPrefix + nlogConfigFilename);
+                        _config = LoadXmlLoggingConfiguration(XmlLoggingConfiguration.AssetsPrefix + nlogConfigFilename);
```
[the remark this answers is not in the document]
304NotModified: that's also testcase worthy ;)

Not sure if we could test internal tests in NLog (then we could add then tests to the  TryMethod)

snakefoot:
@304NotModified Created unit-test for you

tests/NLog.UnitTests/LogFactoryTests.cs
```diff
@@ -96,6 +94,32 @@ public void InvalidXMLConfiguration_ThrowErrorWhen_ThrowExceptionFlagIsSet()
...
+            {
+                Directory.CreateDirectory(tempDirectory);
+                File.WriteAllText(configFile, "<nlog />");
+                LogFactory logFactory = new LogFactory();
+                logFactory.SetCandidateConfigFilePaths(new[] { configFile });
+                using (var fileLock = new FileStream(configFile, FileMode.Open, FileAccess.ReadWrite, FileShare.None))
+                {
+                    var loggingConfig = logFactory.Configuration;
+                    Assert.Null(loggingConfig);
+                    Assert.ThrowsAny<Exception>(() => logFactory.LoadConfiguration(configFile));
```
304NotModified: i'm a bit confused. The method title says "DoesNotThrow", but the assert says the contradiction.

snakefoot:
There are two asserts. First it tries to load config by random scan. It will fail and returns null without throwing exception. Next call performs explicit load that should fail by throwing exception


Sent from my Samsung device


tests/NLog.UnitTests/LogFactoryTests.cs
304NotModified: Maybe we should split this into 2 unit tests, or rename it. Currently, I don't know what to do with this. I kinda brain overload.

src/NLog/LogFactory.cs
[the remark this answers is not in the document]
304NotModified: PS I try to use these guidelines for unit tests nowadays:

- naming rule 1: https://dzone.com/articles/7-popular-unit-test-naming
-  Arrange, Act, Assert: http://defragdev.com/blog/?p=783

snakefoot:
Well I like to verify that the test-stage actually work while also verifying the code works.

That is why there are two asserts. One for the test of loading random configs and one for the actual stage. Guess I should have a third assert after removing the file lock that both assets now returns not-null



Sent from my Samsung device


tests/NLog.UnitTests/LogFactoryTests.cs
304NotModified: I think these are 2 tests.  --> https://github.com/snakefoot/NLog/pull/4

304NotModified:
see snakefoot#4

304NotModified:
mmm cherry pick wasn't an option?

snakefoot:
@304NotModified Like I said Github wasn't responding to 50 pct of https-requests

304NotModified:
the horror

:P

nice:




any way, thanks!

304NotModified:
note: backported to "dev"

PR #2890, opened 2018-08-31
AsyncTargetWrapper - TimeToSleepBetweenBatches changed default to 1ms

snakefoot:
Means lazy trigger on request. Will work very well together with #2650

### DIFF
--- a/src/NLog/Targets/Wrappers/AsyncRequestQueue-T.cs
+++ b/src/NLog/Targets/Wrappers/AsyncRequestQueue-T.cs
@@ -42,7 +42,7 @@ namespace NLog.Targets.Wrappers
     /// </summary>
 	internal class AsyncRequestQueue : IAsyncRequestQueue
     {
-        private readonly Queue<AsyncLogEventInfo> _logEventInfoQueue = new Queue<AsyncLogEventInfo>();
+        private readonly Queue<AsyncLogEventInfo> _logEventInfoQueue = new Queue<AsyncLogEventInfo>(1000);
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


304NotModified:
What's the impact? Do you think it has low risk (NLog 4.6), or maybe some higher (then NLog 5)

snakefoot:
It works very well with FileTarget. And it works very well with minimizing Timer-events, when system idle.

I guess if some network-based-targets uses the AsyncWrapper to throttle requests, then they will affected. But for that case they should already have configured TimeToSleepBetweenBatches manually to 150 or 200 ms (Because every 50 ms is aggressive).

304NotModified:
> I guess if some network-based-targets uses the AsyncWrapper to throttle requests, then they will affected. But for that case they should already have configured TimeToSleepBetweenBatches manually to 150 or 200 ms (Because every 50 ms is aggressive).

mmm doubting about this. 

Will think about this

Thanks!

snakefoot:
Change TimeToSleepBetweenBatches = 1 to also be on activity. But a little more lazy than TimeToSleepBetweenBatches = 0. To improve batching for "slow" targets.

| PerfLogger       | 10.000.000 |  128 |    0 |       1 |       1 |

| 0ms   |     9.320 | 1.072.874 |   0 |   7 |  14 |   12 |    1.137,0 |

| 1ms   |    24.551 |   407.304 |   0 |   7 |  14 |   45 |    1.137,0 |

| 50ms    |   250.083 |    39.986 |   0 |   7 |  14 |      609 |    1.139,1 |

PR #2899, opened 2018-09-04
Database target: support for DbType for parameters (including SqlDbType, OracleDbType etc.)

304NotModified:
todo:

- [x] fix state in DatabaseParameterTypeSetter -> no problem, parameters are also shared
- [x] DatabaseParameterTypeSetter internal?
- [x] need conversion for rawValue? -> not for now. Injection point is there
- [x] static converter injection
- [x] fix headers
- [x] fix tests for new dummy "createParameter"
- [x] naming IDatabaseValueConverter
- [x] naming ParameterValueConverter
- [x] docs IDatabaseValueConverter, naming methodes, parameters check
- [x] IDatabaseValueConverter pass DatabaseParameterInfo?
- [ ] unit tests for DatabaseValueConverter -> partly now
- [ ] unit tests for DatabaseTarget.GetParameterValue (fallbacks) -> partly now
- [ ] DatabaseValueConverter - object todo, json case?
- [x] unit test for rawValue usage
- [x] skip rawValue if the param-dbType is a string/not set. 
- [x] fix NLog.UnitTests.Targets.DatabaseTargetTests.LevelParameterTest test + create variant

### DIFF
--- a/src/NLog/Targets/DatabaseTarget.cs
+++ b/src/NLog/Targets/DatabaseTarget.cs
@@ -665,7 +665,7 @@ namespace NLog.Targets
         /// <summary>
         /// Resolve Parameter DbType And Value Converter
         /// </summary>
-        protected void EnsureResolveParameterInfo(IDbCommand command, IDbDataParameter dbParameter)
+        private void EnsureResolveParameterInfo(IDbCommand command, IDbDataParameter dbParameter)
         {
             if (this._parameterTypeSetter == null)
             {
@@ -842,52 +842,62 @@ namespace NLog.Targets
             for (int i = 0; i < databaseParameterInfos.Count; ++i)
             {
                 DatabaseParameterInfo par = databaseParameterInfos[i];
-                IDbDataParameter dbParameter = command.CreateParameter();
-                EnsureResolveParameterInfo(command, dbParameter);
-                dbParameter.Direction = ParameterDirection.Input;
-                if (par.Name != null)
-                {
-                    dbParameter.ParameterName = par.Name;
-                }
-
-                if (par.Size != 0)
-                {
-                    dbParameter.Size = par.Size;
-                }
-
-                if (par.Precision != 0)
-                {
-                    dbParameter.Precision = par.Precision;
-                }
-
-                if (par.Scale != 0)
-                {
-                    dbParameter.Scale = par.Scale;
-                }
-
-                _parameterTypeSetter.SetParameterDbType(dbParameter, par);
-                object value;
-
-                try
-                {
-                    value = GetParameterValue(logEvent, par, dbParameter.ParameterName, dbParameter.DbType);
-                }
-                catch (Exception exception)
-                {
-                    // enhance context. Already checked for must-rethrow 
-                    exception.Data["parameter"] = dbParameter;
-                    throw;
-                }
-
-
-                dbParameter.Value = value;
-                InternalLogger.Trace("  DatabaseTarget: Parameter: '{0}' = '{1}' ({2})", dbParameter.ParameterName, value, dbParameter.DbType);
+                var dbParameter = CreateDatabaseParameter(command, par, logEvent);
+                InternalLogger.Trace("  DatabaseTarget: Parameter: '{0}' = '{1}' ({2})", dbParameter.ParameterName, dbParameter.Value, dbParameter.DbType);
                 command.Parameters.Add(dbParameter);
-
-
             }
         }
 
+        /// <summary>
+        /// Create database parameter with value from <paramref name="logEvent"/>
+        /// </summary>
+        /// <param name="command">Current command.</param>
+        /// <param name="parameterInfo">Parameter configuration info.</param>
+        /// <param name="logEvent">Current logevent.</param>
+        /// <returns></returns>
+        protected IDbDataParameter CreateDatabaseParameter(IDbCommand command, DatabaseParameterInfo parameterInfo, LogEventInfo logEvent)
+        {
+            IDbDataParameter dbParameter = command.CreateParameter();
+            EnsureResolveParameterInfo(command, dbParameter);
+            dbParameter.Direction = ParameterDirection.Input;
+            if (parameterInfo.Name != null)
+            {
+                dbParameter.ParameterName = parameterInfo.Name;
+            }
+
+            if (parameterInfo.Size != 0)
+            {
+                dbParameter.Size = parameterInfo.Size;
+            }
+
+            if (parameterInfo.Precision != 0)
+            {
+                dbParameter.Precision = parameterInfo.Precision;
+            }
+
+            if (parameterInfo.Scale != 0)
+            {
+                dbParameter.Scale = parameterInfo.Scale;
+            }
+
+            _parameterTypeSetter.SetParameterDbType(dbParameter, parameterInfo);
+            object value;
+
+            try
+            {
+                value = GetParameterValue(logEvent, parameterInfo, dbParameter.ParameterName, dbParameter.DbType);
+            }
+            catch (Exception exception)
+            {
+                // enhance context. Already checked for must-rethrow 
+                exception.Data["parameter"] = dbParameter;
+                throw;
+            }
+
+
+            dbParameter.Value = value;
+            return dbParameter;
+        }
 
 
         /// <summary>
--- a/src/NLog/Targets/DatabaseTarget.cs
+++ b/src/NLog/Targets/DatabaseTarget.cs
@@ -853,7 +853,7 @@ namespace NLog.Targets
         /// <param name="parameterInfo">Parameter configuration info.</param>
         /// <param name="logEvent">Current logevent.</param>
         /// <returns></returns>
-        protected IDbDataParameter CreateDatabaseParameter(IDbCommand command, DatabaseParameterInfo parameterInfo, LogEventInfo logEvent)
+        protected virtual IDbDataParameter CreateDatabaseParameter(IDbCommand command, DatabaseParameterInfo parameterInfo, LogEventInfo logEvent)
         {
             IDbDataParameter dbParameter = command.CreateParameter();
             EnsureParameterSetter(dbParameter);
--- a/src/NLog/Targets/DatabaseTarget.cs
+++ b/src/NLog/Targets/DatabaseTarget.cs
@@ -101,6 +101,7 @@ namespace NLog.Targets
 #endif
             CommandType = CommandType.Text;
             OptimizeBufferReuse = GetType() == typeof(DatabaseTarget);  // Class not sealed, reduce breaking changes
+            ParameterDbTypePropertyName = "DbType";
         }
 
         /// <summary>
@@ -261,10 +262,30 @@ namespace NLog.Targets
         public CommandType CommandType { get; set; }
 
         /// <summary>
+        /// Gets or sets property name of the SQL command parameter to set parameter DbType.
+        /// </summary>
+        /// <remarks>
+        /// May set strong DbType, for SQL Server is SqlDbType.
+        /// </remarks>
+        /// <docgen category='SQL Statement' order='12' />
+        [DefaultValue("DbType")]
+        public string ParameterDbTypePropertyName { get; set; }
+
+        ///<summary>SQL Command Parameter Converter</summary>
+        private DatabaseParameterTypeSetter _parameterTypeSetter;
+
+        /// <summary>
+        /// Converter for parameter values
+        /// </summary>
+        [Advanced]
+        [NLogConfigurationIgnoreProperty]
+        public static IDatabaseValueConverter ParameterValueConverter { get; set; } = new DatabaseValueConverter();
+
+        /// <summary>
         /// Gets the collection of parameters. Each parameter contains a mapping
         /// between NLog layout and a database named or positional parameter.
         /// </summary>
-        /// <docgen category='SQL Statement' order='12' />
+        /// <docgen category='SQL Statement' order='14' />
         [ArrayParameter(typeof(DatabaseParameterInfo), "parameter")]
         public IList<DatabaseParameterInfo> Parameters { get; private set; }
 
@@ -608,7 +629,9 @@ namespace NLog.Targets
                 }
             }
         }
-
+        /// <summary>
+        /// Write logEvent to database
+        /// </summary>
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Security", "CA2100:Review SQL queries for security vulnerabilities", Justification = "It's up to the user to ensure proper quoting.")]
         private void WriteEventToDatabase(LogEventInfo logEvent)
         {
@@ -635,6 +658,26 @@ namespace NLog.Targets
 
                 //not really needed as there is no transaction at all.
                 transactionScope.Complete();
+            }
+        }
+
+
+        /// <summary>
+        /// Resolve Parameter DbType And Value Converter
+        /// </summary>
+        protected void EnsureResolveParameterInfo(IDbCommand command, IDbDataParameter dbParameter)
+        {
+            if (this._parameterTypeSetter == null)
+            {
+                lock (this.SyncRoot)
+                {
+                    if (this._parameterTypeSetter == null)
+                    {
+                        var converter = new DatabaseParameterTypeSetter();
+                        converter.Resolve(dbParameter, this.ParameterDbTypePropertyName, this.Parameters);
+                        this._parameterTypeSetter = converter;
+                    }
+                }
             }
         }
         /// <summary>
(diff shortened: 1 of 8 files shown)


src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -260,11 +263,42 @@ public DatabaseTarget(string name) : this()
...
+        /// <docgen category='SQL Statement' order='12' />
+        [DefaultValue("DbType")]
+        public string ParameterDbTypePropertyName { get; set; }
+
+        /// <summary>
+        /// Gets or sets converter type of the SQL command parameter value.
+        /// </summary>
+        /// <docgen category='SQL Statement' order='13' />
+        [DefaultValue(typeof(DatabaseParameterTypeSetter))]
+        public Type ParameterConverterType { get; set; }
```
304NotModified: @snakefoot any opinion on this? It's injecting the (factory) type as prop and I think this (could) work from XML config.

src/NLog/Targets/DatabaseValueConverter.cs
```diff
@@ -0,0 +1,96 @@
...
+                default:
+                    return value;
+            }
+        }
+
+        /// <inheritdoc />
+        public object ConvertFromObject(object rawValue, DbType dbType, string format)
+        {
+            // todo
+            return rawValue;
```
304NotModified: @snakefoot any idea if we should convert by our-self the rawvalue to another object, or leave it to ADO.NET?

snakefoot: I would probably use Convert.GetTypeCode() and if TypeCode.String or TypeCode.Object, then parse it through ConvertFromString

src/NLog/Targets/DatabaseTarget.cs
snakefoot: No idea. I would think that people who need custom ParameterConverterType would inherit from DataBaseTarget and inject them directly.

src/NLog/Targets/DatabaseValueConverter.cs
304NotModified: FYI If we are here, I know it's not a string. Also I try to skip any string parses.

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -260,11 +263,42 @@ public DatabaseTarget(string name) : this()
...
+
+        ///<summary>SQL Command Parameter Converter</summary>
+        public DatabaseParameterTypeSetter ParameterTypeSetter { get; set; }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public IDatabaseValueConverter ParameterValueConverter
```
304NotModified: @snakefoot do you think I have to move this to the LogFactory? I prefer injection without creating a subclass of the target.

Another option would be to add a `static` field here

src/NLog/Targets/DatabaseTarget.cs
304NotModified: OK thx, I removed this construction for now. It's nice, but not sure if it's better

src/NLog/Targets/DatabaseValueConverter.cs
304NotModified: I think I leave it like this for now. The ADO.NET conversion is pretty smart.

src/NLog/Targets/DatabaseTarget.cs
snakefoot: Think the DatabaseConverter is very DatabaseTarget specific, so not a direct static variable on LogFactory. Maybe consider extending the LogFactory to have a simple dependency-injection system? Instead of having these global-instances (Like IJsonConverter), then it could be neat to register a factory-method for a random interface-type-implementation.

304NotModified: a static (with interface) is a simple dependency-injection system? ;)

Or should I inject it to the database target from ...?

304NotModified: or are you proposing our own container and containerbuilder?

snakefoot: Prefer static instance on the DatabaseTarget then

snakefoot: We already have lots of containers. They are just hardcoded to Targets, Layouts, etc. Could easily have a container that could handle any interface-type.

304NotModified: > Could easily have a container that could handle any interface-type-string.

Sure, but then we need a register and resolve (see https://github.com/NLog/NLog/pull/2899#discussion_r215438293).

And then i'm afraid we need (in the future), DI scopes, DI lifetimes, cached lookups, lambda support, constructor injection, etc etc.. (reinventing the wheel)

snakefoot: You could also just support a DependencyInjection-interface, allowing people to come with their own dependency-injection-container.

304NotModified:
```
 Saving C:\projects\nlog\src\nlog\\bin\Release\NLogMerged.api.xml...
  Processing API file 'C:\projects\nlog\src\nlog\bin\Release\net45\API\NLog.api'
EXEC : error : System.NotSupportedException: Unknown API type 'NLog.IDatabaseValueConverter'. [C:\projects\nlog\src\NLog.proj]
     at MakeNLogXSD.XsdFileGenerator.GetXsdType(String apiTypeName, Boolean attribute) in C:\projects\nlog\tools\MakeNLogXSD\XsdFileGenerator.cs:line 367
     at MakeNLogXSD.XsdFileGenerator.GetPropertyElement(XElement propertyElement) in C:\projects\nlog\tools\MakeNLogXSD\XsdFileGenerator.cs:line 268
     at MakeNLogXSD.XsdFileGenerator.GetPropertyElements(XElement type) in C:\projects\nlog\tools\MakeNLogXSD\XsdFileGenerator.cs:line 249
     at MakeNLogXSD.XsdFileGenerator.ProcessApiFile(String apiFile) in C:\projects\nlog\tools\MakeNLogXSD\XsdFileGenerator.cs:line 106
     at MakeNLogXSD.Program.Main(String[] args) in C:\projects\nlog\tools\MakeNLogXSD\Program.cs:line 98
C:\projects\nlog\src\NLog.proj(498,5): error MSB3073: The command ""C:\projects\nlog\src\..\build\bin\tools\MakeNLogXSD.exe" -api API\NLog.api -out NLog.netfx45.xsd -xmlns http://www.nlog-project.org/schemas/NLog.netfx45.xsd" exited with code 1.
Command exited with code 1

```

hubo0831:
Sorry! It's been too long. The details are hard to remember, You're working hard.
Sorry! My english is bad!

src/NLog/Targets/DatabaseValueConverter.cs
```diff
@@ -0,0 +1,125 @@
...
+        /// <summary>
+        /// convert layout value to parameter value
+        /// </summary>
+        public object ConvertFromString(string value, DbType dbType, string parseFormat)
+        {
+            switch (dbType)
+            {
+                case DbType.String:
+                case DbType.AnsiString:
+                case DbType.AnsiStringFixedLength: //todo truncate?
```
NLog-SonarQube: INFO Complete the task associated to this 'TODO' comment. [rule](https://sonarcloud.io/coding_rules#rule_key=csharpsquid%3AS1135)

src/NLog/Targets/DatabaseValueConverter.cs
```diff
@@ -0,0 +1,125 @@
...
+                    return ConvertToBinary(value);
+                default:
+                    return value;
+            }
+        }
+
+        /// <inheritdoc />
+        public object ConvertFromObject(object rawValue, DbType dbType, string format)
+        {
+            // todo
```
NLog-SonarQube: INFO Complete the task associated to this 'TODO' comment. [rule](https://sonarcloud.io/coding_rules#rule_key=csharpsquid%3AS1135)

NLog-SonarQube:
SonarQube analysis reported 2 issues
* ![INFO][INFO] 2 info

Watch the comments in this conversation to review them.

[INFO]: https://sonarsource.github.io/sonar-github/severity-info.png 'Severity: INFO'

src/NLog/Targets/DatabaseValueConverter.cs
```diff
@@ -0,0 +1,127 @@
...
+                    return ConvertToBinary(value);
+                default:
+                    return value;
+            }
+        }
+
+        /// <inheritdoc />
+        public object ConvertFromObject(object rawValue, DbType dbType, DatabaseParameterInfo parameterInfo)
+        {
+            // todo
```
NLog-SonarQube: INFO Complete the task associated to this 'TODO' comment. [rule](https://sonarcloud.io/coding_rules#rule_key=csharpsquid%3AS1135)

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -395,7 +416,7 @@ protected override void InitializeTarget()
                         InternalLogger.Warn(ex, "DatabaseTarget(Name={0}): DbConnectionStringBuilder failed to parse '{1}' ConnectionString", Name, ConnectionStringName);
                     else
```
NLog-SonarQube: CRITICAL Use curly braces or indentation to denote the code conditionally executed by this 'else' [rule](https://sonarcloud.io/coding_rules#rule_key=csharpsquid%3AS3973)

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -790,40 +833,137 @@ private void RunInstallCommands(InstallationContext installationContext, IEnumer
...
 
+        /// <summary>
+        /// Get (converted) parameter value, including fallbacks
+        /// </summary>
+        /// <param name="logEvent">Current logevent.</param>
+        /// <param name="parameterInfo">Parameter config</param>
+        /// <param name="parameterName"></param>
+        /// <param name="dbType"></param>
+        /// <returns></returns>
+        private object GetParameterValue(LogEventInfo logEvent, DatabaseParameterInfo parameterInfo, string parameterName, DbType dbType)
```
NLog-SonarQube: CRITICAL Refactor this method to reduce its Cognitive Complexity from 16 to the 15 allowed. [rule](https://sonarcloud.io/coding_rules#rule_key=csharpsquid%3AS3776)

NLog-SonarQube:
SonarQube analysis reported 4 issues
* ![CRITICAL][CRITICAL] 2 critical
* ![INFO][INFO] 2 info

Watch the comments in this conversation to review them.

[CRITICAL]: https://sonarsource.github.io/sonar-github/severity-critical.png 'Severity: CRITICAL'
[INFO]: https://sonarsource.github.io/sonar-github/severity-info.png 'Severity: INFO'

snakefoot:
Should the raw-output not be skipped if having provided explicit formatting options:

`${exception:format=@}`
`${date:format=YYYY}`

Maybe the IRawValue-interface should have a `HasFormat`-property. So one can skip direct use of raw-value when Format has been specified.

304NotModified:
I think the rawValue should be skipped if the value is dbType is a string/not set. So then your examples works.

I think I have to fix that.

snakefoot:
> I think the rawValue should be skipped if the value is dbType is a string/not set. So then your examples works.

Lets say I want to truncate the time (Using format), and then save with non-string DbType ?

304NotModified:
then you need your own DatabaseValueConverter (for now). 

Another option is to add an property to the databaseParameterInfo, e.g. UseRawValue (default true)

snakefoot:
> then you need your own DatabaseValueConverter (for now).

Strange limitation, when it is obvious that the RawValue should be ignored when Format has been specified. Alternative one must register a custom layoutrenderer that does exactly the same, but without implementing the IRawValue-interface

304NotModified:
not really obvious, as the could be many properties that influence the output. But those influence only the string output, and never the rawvalue

snakefoot:
> not really obvious, as the could be many properties that influence the output. But those influence only the string output, and never the rawvalue

So when a user explicit specify a Format then it doesn't mean anything?

304NotModified:
> explicit been specified (format=raw or maybe database-parameter-option)

I'm OK with a opt-out, but not with an opt-in. e.g. dbtype=Datetime and ${date} should work by default without a stringy and parse (which is slow and non-robust)

304NotModified:
> So when a user explicit specify a Format then it doesn't mean anything?

Not if you're using it as an non-string value (for now)

so

${time:format} - format used if dbtype=string
${time:format} - format not used if dbtype=time

snakefoot:
> I'm OK with a opt-out, but not with an opt-in. e.g. dbtype=Datetime and ${date} should work by default without a stringy and parse (which is slow and non-robust)

I never said that ${date} should work nicely together with `dbtype=datetime` without more allocation than the object-boxing. But when having specified a format then it should be respected.

304NotModified:
so  ${date:format} with dbtype=datetime should honor the format? How? Stringy and parse?

IMO the format is used for stringify, and obvious, you don't want a string here ;)

snakefoot:
> so ${date:format} with dbtype=datetime should honor the format? How? Stringy and parse?

Think you explained to me before that the DbType-conversion logic is strong, so just parse a string with the specified DbType. Maybe just have parameter-option whether to pass string or rawvalue (default) into the database-parameter-value.

304NotModified:
> Maybe just have parameter-option whether to pass string or rawvalue (default) into the database-parameter-value.

Yes, I will make that option then.

304NotModified:
added "useRawValue"

304NotModified:
Thanks for checking :)

NLog-SonarQube:
SonarQube analysis reported 3 issues
* ![CRITICAL][CRITICAL] 1 critical
* ![INFO][INFO] 2 info

Watch the comments in this conversation to review them.

[CRITICAL]: https://sonarsource.github.io/sonar-github/severity-critical.png 'Severity: CRITICAL'
[INFO]: https://sonarsource.github.io/sonar-github/severity-info.png 'Severity: INFO'

304NotModified:
> Please make sure these does not become breaking changes, and suddenly performs Exception-ToString()

there should be no breaking changes, when dbtype = not set or dbtype=stringlike-type, it will use the old path.

snakefoot:
What happens if wrapping the DatabaseTarget in the AsyncTargetWrapper, and Layout.Precalculate has been performed.

What value should Layout.TryGetRawValue return when not ThreadAgnostic or  MutableUnsafe?

See also #2973 where I have implemented fallback to string-value when cached as async. Is this a good fallback?

304NotModified:
> What happens if wrapping the DatabaseTarget in the AsyncTargetWrapper, and Layout.Precalculate has been performed.
> 
> What value should Layout.TryGetRawValue return when not ThreadAgnostic or MutableUnsafe?
> 
> See also #2973 where I have implemented fallback to string-value when cached as async. Is this a good fallback?

Phoe that's difficult one.  I don't know. Any suggestion to do? 

I really like to finish this, but limited with testing time. I'm not sure if this PR is good enough for now.

snakefoot:
> Phoe that's difficult one. I don't know. Any suggestion to do?

Maybe the easy solution is just to remove `IRawValue`-interface from classes that doesn't have ThreadAgnostic-attribute?

snakefoot:
@304NotModified Should I update #2973 where I remove `IRawValue`-interface from render-classes without ThreadAgnostic-attribute. And also fix SimpleLayout to have smarter Precalculate when `IRawValue` ?

snakefoot:
@304NotModified Would it be possible to do something else than having LogLevel implementing `IConvertible` ? Have feeling that it might have funny sideeffects for different serializers.

Was thinking that users might want to add conversion-options for external class-types, where they cannot make the external-type implement `IConvertible`

snakefoot:
@304NotModified Was also hoping to reuse the type conversion system for the TargetWithContext, so one can render into a specific type using the contextproperties.

304NotModified:
> @304NotModified Would it be possible to do something else than having LogLevel implementing `IConvertible` ? Have feeling that it might have funny sideeffects for different serializers.

yes, we could do it with a custom interface, but IMO it's pretty strong to use IConvertible in this case as IMO the name is overused (e.g. in a database it should be a number). Or do you have a counter-example? 

> @304NotModified Was also hoping to reuse the type conversion system for the TargetWithContext, so one can render into a specific type using the contextproperties (Returning objects with correct type when calling `GetAllProperties`)

Sound good, what should be done for that?

304NotModified:
I will try to work on this next week, so we could release 4.6 soon.

I will focus on getting the build working. 

@snakefoot Do I need to change some things in this PR after fixing the build? (sorry didn't read all messages in the past)

snakefoot:
> yes, we could do it with a custom interface, but IMO it's pretty strong to use IConvertible in this case as IMO the name is overused (e.g. in a database it should be a number). Or do you have a counter-example?

I know the most optimal way to store and query loglevel in database is integer, but for most people they will probably use a one-letter-conversion T-race, D-ebug, etc. for readability.

One example is the JsonSerializer in NLog that suddenly will convert LogLevel-objects to integer because IConvertible says so. Using IConvertible for the sake of database-target, will have affect all other places. And again having an option for doing type-conversion for custom object-types would also be useful for other objects.

snakefoot:
> > Was also hoping to reuse the type conversion system for the TargetWithContext,
> Sound good, what should be done for that? In this PR or another one?
It can be done in a separate PR, since the IRawValue-interface is already available.

snakefoot:
Created #3060 for IRawValue-support for `TargetWithContext`

304NotModified:
> I know the most optimal way to store and query loglevel in database is integer, but for most people they will probably use a one-letter-conversion T-race, D-ebug, etc. for readability.

But then it's much harder to find all logs with `level > somelevel`

snakefoot:
Any reason for `DatabaseParameterTypeSetter` to be a `public class` (instead of internal class).

snakefoot:
Any reason for having `ParameterDbTypePropertyName` on the target. Instead of having it on the individual `DatabaseParameterInfo` ?

Maybe just use DbType. Ex. `DbType = "DateTime"` or `DbType = "SqlDbType.NVarChar"`. This is more intuitive for me.

snakefoot:
Curious of the huge local function in `GetParameterValue`. Thought the goal was to reduce complexity of methods for easier code-review and maintenance.

304NotModified:
> Any reason for having `ParameterDbTypePropertyName` on the target. Instead of having it on the individual `DatabaseParameterInfo` ?

That the propertyname, e.g. SqlDbType (https://docs.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlparameter.sqldbtype?view=netframework-4.7.2). That one is indeed on parameter level, but should be the same for all parameters of 1 target. (as you will only target one databasesystem).



> Maybe just use DbType. Ex. `DbType = "DateTime"` or `DbType = "SqlDbType.NVarChar"`. This is more intuitive for me.

Not everything is supported with DbType too bad :( That would be far more easier indeed. 



> Think it is ugly to have `EnsureResolveParameterInfo` as protected member on DatabaseTarget. Instead the logic for creating a `IDbDataParameter`-instance should be extracted into a method, that one can override (Receives `command` + `logEvent` + `DatabaseParameterInfo` as input). Then one can call the base-method and then override DbType if needed.

I think we could change that. Not sure how you would see that. Could you please create a code proposal or a PR?

> Curious about the use of huge local function in `GetParameterValue`. Thought the goal was to reduce complexity of methods for easier code-review and maintenance.

Deduplication of error handling code, but I agree it's not the best code I have seen ;) Any proposal to deduplicate it in another way?

304NotModified:
Thanks for reviewing @snakefoot ! I really like to know when you think it's good enough (when we are there ;) )

snakefoot:
> Not everything is supported with DbType too bad :( That would be far more easier indeed.

You already making it happen. I'm suggesting that you just move the logic implemented with ParameterDbTypePropertyName  into DatabaseParameterInfo (Recognizes non DbType-values and then perform advanced assignment).

> Any proposal to deduplicate it in another way?

Extract the local method into a separate private method

snakefoot:
>  I really like to know when you think it's good enough

I think we are very close. Just think the API-interface to handle SqlDbType-values is a little clunky. Most of the logic is hidden nicely away so can be changed without too much trouble if needed.

304NotModified:
> You already making it happen. I'm suggesting that you just move the logic implemented with ParameterDbTypePropertyName into DatabaseParameterInfo (Recognizes non DbType-values and then perform advanced assignment of custom DbType like SqlDbType).

Not sure if I fully understand. Do we need to drop the DatabaseParameterTypeSetter? (ParameterDbTypePropertyName is just a string prop)



> Just extract the code inside this loop, and put into a separate method:

And make it protected?




> Extract the local method into a separate private method

That's an option, but not sure if it's better then, I method will have two extra parameters and will be (only) used in 1 method.

304NotModified:
> Think it is ugly to have `EnsureResolveParameterInfo` as protected member on DatabaseTarget.

This is fixed now, much better indeed!

304NotModified:
> I suggest that DatabaseParameterInfo.DbType should be extended to support `SqlDbType.NVarChar` instead

How would the configuration will look then? `Dbtype=SqlDbType.NVarChar` ? I'm not sure if the type and propertyname (SqlDbType) are in-sync everywhere. If not, then that could be confusing.

304NotModified:
> > And make it protected?
> 
> Maybe just keep it `private` until it is requested to be `protected virtual`. Actually thought that you wanted to provide the ability to override the parameter-assignment-behavior with this PR. But if not then I prefer to keep it `private`.

Forget the `virtual`. Yes, I think we need sometimes a bit more flexible.

snakefoot:
> How would the configuration will look then? Dbtype=SqlDbType.NVarChar ? 

Well the DatabaseParameterInfo.DbType is a string. So the easy way would be `DatabaseParameterInfo.DbType = "SqlDbType.NVarChar"`

304NotModified:
any idea how to "cache" the parsed propertyinfo and dbValue for the whole DB target?  Or do you think that's not needed? 

It's dependent of the runtime type of  IDbDataParameter, which makes it a bit tricky.

snakefoot:
> It's dependent of the runtime type of IDbDataParameter, which makes it a bit tricky.

I think it is very important to perform caching of the assignment. At least when doing non-DbType Enum-assignment. Consider using NLog `CreateLateBoundMethod`

For most optimal user-experience then I would suggest to use InitializeTarget to build assignment-caching on the individual paramters. Then the target can also throw exceptions if the `DBProvider` does not support the configured `DatabaseParameterInfo.DbType`. This of course requires one to create a "dummy" DbCommand on initialization, but it should be not be a problem.

304NotModified:
> I think it is very important to perform caching of the assignment. At least when doing non-DbType Enum-assignment. Consider using NLog `CreateLateBoundMethod`

Yes, that's indeed nice. 

> . This of course requires one to create a "dummy" DbCommand on initialization, but it should be not be a problem.

I tried to prevent that, but maybe it's indeed needed. 

Do you have time for this refactoring? I don't think I have time for that this month :(

snakefoot:
> I tried to prevent that, but maybe it's indeed needed.

It is only needed if using custom DbType like `SqlDbType.NVarChar`.

> Do you have time for this refactoring? I don't think I have time for that this month :(

I guess if you commit what you have to dev, then I can try and make a new PR.

snakefoot:
@304NotModified You should fix `DatabaseParameterTypeSetter.cs` so usings are moved below:

`#if !SILVERLIGHT && !__IOS__ && !__ANDROID__`

304NotModified:
close/reopen to trigger AppVeyor

304NotModified:
> I guess if you commit what you have to dev-branch, then I can try and make a new PR.

Merged :) Thanks in advance!

PR #2907, opened 2018-09-11
New Substring, Left & Right Wrappers

snakefoot:
Examples:
```c#
${substring:${level}:start=2:length=2}    // from index 2, length 2
${substring:${level}:start=-2:length=2}   // from end -2, length 2
${left:${level}:length=2}    // from start, length 2
${right:${level}:length=2}   // from end, length 2
```


304NotModified:
i've added substring :)

snakefoot:
@304NotModified Looks like NetCore doesn't like when doing SubString where Start-offset is higher than the actual string-length:

https://travis-ci.org/NLog/NLog/builds/450439888

304NotModified:
Thanks for fixing @snakefoot

Issue #2844, opened 2018-08-12
How to end a string in a Layout Renderer with a \ ?


pablinos:
I've been trying to configure a default logging directory when an environment variable is empty. The default value for the project needs to be `c:\logs\` - note the trailing slash.

In the target I've been setting the filename to various things, but they all don't work:

- `filename="${environment:whenEmpty=c\:\log\:LOG_DIR}logfilename.log"` - escaping is wrong
- `filename="${environment:whenEmpty='c\:\log\':LOG_DIR}logfilename.log"` - outputs the single quotes
- define a variable  `<variable name="logDirectory" value="c\:\logs\"/>` and
`filename="${environment:whenEmpty=${logDirectory}:LOG_DIR}logfilename.log"` - throws an exception about `Variable` not being supplied as a parameter to `environment`
- use the same variable and `filename="${environment:whenEmpty='${logDirectory}':LOG_DIR}logfilename.log"` - outputs the single quotes

This is on .NET core 1.1 and version 4.4.1 of NLog, so it may just be that I'm on an old version and the project should be upgraded.

Can what I'm trying to do be done?

pablinos:
Thanks for the speedy reply. I had thought about using the `literal:text` renderer but assumed it would have the same problems.

I tried your suggestion and I get nothing from the `${whenEmpty}` renderer and looking at the internal log I get this line:

`Warn Exception in layout renderer. Exception: System.NullReferenceException: Object reference not set to an instance of an object.`

I'm wondering if the trailing slash escapes the closing brace and so it messes up the parsing. Could that be the case and, if so, is there a way I can escape the trailing slash?

304NotModified:
could you please post the full stacktrace? A `NullReferenceException` is always a bug

304NotModified:
Thanks!

304NotModified:
the `\` should be escaped with another `\\`, but still it's an issue because the `\` will also escape the `:`.

I think the best solution for now is to create a variable and use `${var:myDir}`

304NotModified:
I'm sorry, haven't found time to find the solution for this yet. Maybe this is a good question for stackoverflow

lobster2012-user:
Hello @pablinos! 
The problem is still relevant? I could not reproduce the error in myself.

pablinos:
[...]

PR #2908, opened 2018-09-11
Layout Wrappers should not crash when Inner-Layout is null

snakefoot:
Added null-check to handle #2844


304NotModified:
was `null` an issue in #2844?

304NotModified:
Thanks for the change, looks good!

snakefoot:
Seems there is a test that depends on null working.

snakefoot:
> @304NotModified  was null an issue in #2844?

```
Warn Exception in layout renderer. Exception: System.NullReferenceException: Object reference not set to an instance of an object.
```

But this PR doesn't resolve the actual issue of #2844

PR #2909, opened 2018-09-11
AsyncTaskTarget: Added support for BatchSize and RetryCount

snakefoot:
Made it possible to perform batching using async Tasks. Also improved the concurrency when loggers writes  to the internal target-buffer.

Just configure BatchSize=50 in constructor and override this method:

```
protected override Task WriteAsyncTask(IList<LogEventInfo> logEvents, CancellationToken cancellationToken)
{
    // Blah blah
}
```


304NotModified:
nice feature!

PR #2857, opened 2018-08-21
Added support for complex objects in MDLC and NDLC using ObjectHandle

snakefoot:
Resolves https://github.com/NLog/NLog.Extensions.Logging/issues/237

PR #2910, opened 2018-09-11
ObjectHandle - LifeTime avoid remoting exception: has been disconnected

snakefoot:
or does not exist at the server. Followup to #2857 

The recommendation of using ObjectHandle from [David Fowler](https://github.com/Glimpse/Glimpse/issues/632#issuecomment-41484855) was a trap.

Instead perform ToString serialization on-demand when changing AppDomain (Deferring the overhead of remoting until actually AppDomain transfer). Avoid funny business with lifetime management. Avoid funny business with returning null in InitializeLifetimeService, and abusing the finalizer queue for cleanup.


304NotModified:
👍 thanks!

snakefoot:
Should improve performance (Compare to the ToString-hack in NLog/NLog.Extensions.Logging#232), as it can defer the ToString-operation and Type-checking until actually doing AppDomain-transfer.

PR #2914, opened 2018-09-13
AsyncTaskTarget - Retry WriteAsyncTask should be protected with SyncRoot

snakefoot:
Minor fixes to #2909:

- Retry should also lock SyncRoot to protect Layouts against concurrent usage (If needed)
- RetryFailedAsyncTask is now always called on Exceptions, even if RetryCount = 0 (Can be used for target recovery on error, and perform delay before starting next task).
- Retry should also happen when WriteAsyncTask throws exception without returning a Task


304NotModified:
Thanks! Also for the unit test 👍

PR #2915, opened 2018-09-13
Reduce code complexity in different locations

snakefoot:


304NotModified:
Thanks!

Did you use a refactoring tool for this? (e.g. VS2017, R#). Does it need a review?

snakefoot:
Tried using VS2017 method extractor, but it failed at all locations. So I moved the code manually.

Split into individual commits, so they should be easy to review.

304NotModified:
yes, R# is much better.

304NotModified:
Looks good, thanks! 👍

PR #2917, opened 2018-09-15
Logger.Log should initialize LogEventInfo.LoggerName

### LINKED ISSUE #2916: Logger.Log(LogEventInfo logEvent) does not set LoggerName
Hi,

Unlike Logger.Debug(), Logger.Info(), Logger.Error(), etc the Logger.Log(LogEventInfo logEvent) does not set the LoggerName property.

Since we are calling Log method of the Logger object, we know who the logger is.

Why not do this inside the code for the logger class:

```
if(string.IsNullOrEmpty(logEvent.LoggerName))
  logEvent.LoggerName = logger.Name;
```

This way Logger.Log would exhibit similar behavior to Logger.Info(), Logger.Error(), etc --- all of which set the logger name for you auto-magically.

I raised it as a bug b/c I think Logger.Log stands out like a black sheep from other log methods (LoggerName wise) and there is no need for it.

Thanks!!

snakefoot:
Resolves #2916

### DIFF
--- a/src/NLog/Logger.cs
+++ b/src/NLog/Logger.cs
@@ -102,6 +102,8 @@ namespace NLog
         {
             if (IsEnabled(logEvent.Level))
             {
+                if (logEvent.LoggerName == null)
+                    logEvent.LoggerName = Name;
                 WriteToTargets(logEvent);
             }
         }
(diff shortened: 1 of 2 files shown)


clearwaterstream:
thank you!!

Issue #2911, opened 2018-09-12
PerformanceCounter always showing 0 CPU Usage


Snehal14283:
It display always showing 0 value when using following syntax while in csv file :
`<column name="CPUUsage" layout="${performancecounter:Category=Processor:Counter=% Processor Time:Instance=_Total}"/>`

304NotModified:
I think it could be access rights. Please check the internal log (and if this still an issue, please fill in the template)

Snehal14283:
No i am having full rights. Please check the below code which is written in C# console application 

`            cpuCounter = new PerformanceCounter("Processor",  "% Processor Time", "_Total");
            cpuCounter.NextValue();
            System.Threading.Thread.Sleep(1000);
            MessageBox.Show("CPU Usage: " + cpuCounter.NextValue() + "%");
`
PerformanceCounter class will give the correct CPU usage on 2nd .NextValue() call. In between we need make thread sleep for at lease 1 sec. On first call of .NextValue() it will always be 0.

As of me, NLog calling .NextValue() and giving the result but on first call it will always be 0. is it true?

snakefoot:
@Snehal14283 You are very correct the normal way of doing performance counters is this:

```
CounterSample currentSample = perfCounter.NextSample();
float perfValue = CounterSample.Calculate(_prevSample, currentSample);
_prevSample = currentSample;
return perfValue;
```

So the `PerformanceCounterLayoutRenderer` doesn't work well with performance-counters that depends on calculation with previous-sample-value (Like CPU usage). Having a Timer ticking every second in the layoutrenderer could be one option, though not sure if it is pretty.

304NotModified:
What about the `${QPC}` (https://github.com/NLog/NLog/wiki/QPC-Layout-Renderer), is there the same problem? It has a "difference" and "normalize " option

snakefoot:
> What about the ${QPC}

Think that is something different. It is used for high-precision timing measurements. Not related to Processor Usage Performance Counters.

Snehal14283:
I am new to GitHub. Why new thread started #2918 ?  Do its an error in NLog? will i get in next version or do i need to change something in nlog.config?

snakefoot:
@Snehal14283 The new thread is a "Pull Request" which allows one to include source-code-changes.

Sadly enough you have come at a bad time, because right now NLog 4.6 is being prepared, so no official release of NLog in the near feature.

[...]

PR #2918, opened 2018-09-15
PerformanceCounter - Improve behavior for CPU usage calculation + Format and Culture options

snakefoot:
A delay is required for sample calculation. See also #2911

It is not a perfect solution, as it would require a frequent timer, to update previous sample in the background. But if logging frequently, then there should be a valid previous sample, that gives non-zero result.

This is a throttled version of the normal `NextValue()` (See handling of oldSample):

https://referencesource.microsoft.com/#system/services/monitoring/system/diagnosticts/PerformanceCounter.cs,582

See also `The recommended delay time between calls to the NextValue method is one second`:

https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.performancecounter.nextvalue

### DIFF
--- a/src/NLog/LayoutRenderers/PerformanceCounterLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/PerformanceCounterLayoutRenderer.cs
@@ -35,11 +35,13 @@
 
 namespace NLog.LayoutRenderers
 {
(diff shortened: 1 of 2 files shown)


304NotModified:
Thanks! Merged!

PR #2919, opened 2018-09-15
Introduced NLog.AppConfig for .NET 45 + .NET Standard 2.0

snakefoot:
Allow access to appsettings in app.config (or web.config) without using NLog.Extended (That depends on Msmq).

Also added support for NetStandard2.0 (See https://www.nuget.org/packages/System.Configuration.ConfigurationManager/4.4.0)

Decided not to make a file-link into NLog.Extended, because it enforced caching. Instead one have to manually use `${appsetting:cached=true:name=MySetting}`.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -39,6 +39,10 @@ msbuild /t:Restore,Pack .\src\NLog.WindowsIdentity\ /p:VersionPrefix=$versionPre
 if (-Not $LastExitCode -eq 0)
 	{ exit $LastExitCode }
 
+msbuild /t:Restore,Pack .\src\NLog.AppConfig\ /p:VersionPrefix=$versionPrefix /p:VersionSuffix=$versionSuffix /p:FileVersion=$versionFile /p:ProductVersion=$versionProduct /p:Configuration=Release /p:IncludeSymbols=true /p:PackageOutputPath=..\..\artifacts /verbosity:minimal
+if (-Not $LastExitCode -eq 0)
+	{ exit $LastExitCode }
+
 msbuild /t:xsd /t:NuGetSchemaPackage /t:NuGetConfigPackage .\src\NLog.proj /p:Configuration=Release /p:BuildNetFX45=true /p:BuildVersion=$versionProduct /p:Configuration=Release /p:BuildLabelOverride=NONE /verbosity:minimal
 
 exit $LastExitCode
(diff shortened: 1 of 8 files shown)


snakefoot:
Will create another nuget-package NLog.Extension.Configuration for reading appsettings.json.

snakefoot:
Created https://github.com/NLog/NLog.Extensions.Logging/pull/245 for Net Core AppSettings.json

304NotModified:
Why not call it NLog.AppSettings?

This also works for web.config

snakefoot:
> AppConfig is a bit confusing as this also works for web.config

I hope people who knows about WebConfig, also knows about AppConfig. And that they cover the same logic.

But yes I chose the name AppConfig to make it very different from appsettings.json.

304NotModified:
> I guess this nuget-package should also include the logic to load NLog-LoggingConfiguration from AppConfig. Though right now it collides a little with NLog own attempt to load from AppConfig.

good one. I don't think we should hold this PR? I think this should be NLog 5.0?

snakefoot:
> I don't think we should hold this PR? I think this should be NLog 5.0?

No we can start the nuget-package without having this feature. If NLog 5.0 is the "minor" breaking release, and NLog 6.0 is the "major" breaking release. Then I guess auto-loading NLog-Configuration from App.Config should be in NLog 6.0

snakefoot:
> For appsettings.json etc. in ASP.NET core, use ${configsetting:name-MySetting} in https://www.nuget.org/packages/NLog.Extensions.Logging

Should probably be:

> For appsettings.json etc. in ASP.NET core, use ${configsetting:name-MySetting} from https://www.nuget.org/packages/NLog.Extensions.Configuration

snakefoot:
And there came the push :)

304NotModified:
thanks :)

snakefoot:
Created https://github.com/NLog/NLog.Extensions.Logging/pull/248 to include details about `${configsetting}` in https://www.nuget.org/packages/NLog.Extensions.Configuration

304NotModified:
SimpleConcurrentTest is really unstable lately. :(

snakefoot:
Been trying to implement a working solution with automatic loading of nlog-config from app.config, and it has not been a success.

Since the app-config loading only happens for NetFramework, and the NetFramework is entering maintance-mode. Then I don't think that having this dependency for NetFramework will be a big problem.

snakefoot:
Closing this issue in favor of #3053

PR #2926, opened 2018-09-20
NullAppender - Added missing SecuritySafeCritical

snakefoot:


304NotModified:
when should we apply `[SecuritySafeCritical]`?

snakefoot:
> when should we apply [SecuritySafeCritical]?

No idea. But apparently when inheriting from a class that has the [SecuritySafeCritical]-attribute.

304NotModified:
OK merged it!

Issue #2933, opened 2018-09-24
Memory Leak with async target and usage of ${threadid}, GDC or MLDC in the layout.


staufour:
**NLog version**: 4.5.10

**Platform**: .Net 4.5

**Current NLog config**

```xml
<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd"
      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">

  <targets async="true">
    <target name="logfile" xsi:type="File" fileName="log.txt">
      <layout xsi:type="CsvLayout">
        <withHeader>true</withHeader>
        <column name="Date" layout="${date:universalTime=True:format=o}" />
        <column name="Server" layout="${machinename}" />
        <column name="Logger" layout="${logger}" />
        <column name="Source" layout="${callsite}" />
        <column name="Level" layout="${level}" />
        <column name="Message" layout="${message}" />
        <column name="Exception" layout="${exception:format=ToString}" />
      </layout>
    </target>
    <target name="logfile_crash" xsi:type="File" fileName="log_crash.txt">
      <layout xsi:type="CsvLayout">
        <withHeader>true</withHeader>
        <column name="Date" layout="${date:universalTime=True:format=o}" />
        <column name="Server" layout="${machinename}" />
        <!--<column name="ThreadId" layout="${threadid}" />-->
        <column name="Application" layout="${mdlc:item=Application}" />
        <!--<column name="Version" layout="${gdc:item=Version}" />-->
        <column name="Logger" layout="${logger}" />
        <column name="Source" layout="${callsite}" />
        <column name="Level" layout="${level}" />
        <column name="Message" layout="${message}" />
        <column name="Exception" layout="${exception:format=ToString}" />
      </layout>
    </target>
  </targets>

  <rules>
    <logger name="PerfTest" minlevel="Info" writeTo="logfile" />
    <logger name="PerfTestCrash" minlevel="Info" writeTo="logfile_crash" />
  </rules>
</nlog>

```

Hi,

We encounter a memory leak in our application in production, and the exploitation of memory dump led us to suspect NLog.
After many tests, we realized that using an asynchronous target with a ${threadid}, GDC or MLDC in the layout causes the leak.
The current thread hang, the memory increases and the application finally crash.

We can reproduce quickly by logging large messages quickly.

You can find a test project here with the demonstration : https://github.com/staufour/LoggerPerfTest

Thanks for your help.

304NotModified:
[...]

PR #2934, opened 2018-09-24
CsvLayout - Performance&docs improvements, added Quoting also on Column.

snakefoot:
Add support for ThreadSafe-optimization

by including CsvHeaderLayout. Resolve issue reported in #2933

### DIFF
--- a/src/NLog/Layouts/CsvLayout.cs
+++ b/src/NLog/Layouts/CsvLayout.cs
@@ -185,29 +185,29 @@ namespace NLog.Layouts
             for (int i = 0; i < Columns.Count; i++)
             {
                 Layout columnLayout = Columns[i].Layout;
-                RenderColumnLayout(logEvent, columnLayout, target, i);
+                RenderColumnLayout(logEvent, columnLayout, Columns[i]._quoting ?? Quoting, target, i);
             }
         }
 
-        private void RenderColumnLayout(LogEventInfo logEvent, Layout columnLayout, StringBuilder target, int i)
+        private void RenderColumnLayout(LogEventInfo logEvent, Layout columnLayout, CsvQuotingMode quoting, StringBuilder target, int i)
         {
             if (i != 0)
             {
                 target.Append(_actualColumnDelimiter);
             }
 
-            if (Quoting == CsvQuotingMode.All)
+            if (quoting == CsvQuotingMode.All)
             {
                 target.Append(QuoteChar);
             }
 
             int orgLength = target.Length;
             columnLayout.RenderAppendBuilder(logEvent, target);
-            if (orgLength != target.Length && ColumnValueRequiresQuotes(target, orgLength))
+            if (orgLength != target.Length && ColumnValueRequiresQuotes(quoting, target, orgLength))
             {
                 string columnValue = target.ToString(orgLength, target.Length - orgLength);
                 target.Length = orgLength;
-                if (Quoting != CsvQuotingMode.All)
+                if (quoting != CsvQuotingMode.All)
                 {
                     target.Append(QuoteChar);
                 }
@@ -216,7 +216,7 @@ namespace NLog.Layouts
             }
             else
             {
-                if (Quoting == CsvQuotingMode.All)
+                if (quoting == CsvQuotingMode.All)
                 {
                     target.Append(QuoteChar);
                 }
@@ -238,13 +238,13 @@ namespace NLog.Layouts
                 CsvColumn col = Columns[i];
                 var columnLayout = new SimpleLayout(new LayoutRenderers.LayoutRenderer[] { new LayoutRenderers.LiteralLayoutRenderer(col.Name) }, col.Name, ConfigurationItemFactory.Default);
                 columnLayout.Initialize(LoggingConfiguration);
-                RenderColumnLayout(logEvent, columnLayout, sb, i);
+                RenderColumnLayout(logEvent, columnLayout, col._quoting ?? Quoting, sb, i);
             }
         }
 
-        private bool ColumnValueRequiresQuotes(StringBuilder sb, int startPosition)
+        private bool ColumnValueRequiresQuotes(CsvQuotingMode quoting, StringBuilder sb, int startPosition)
         {
-            switch (Quoting)
+            switch (quoting)
             {
                 case CsvQuotingMode.Nothing:
                     return false;
--- a/src/NLog/Layouts/CsvLayout.cs
+++ b/src/NLog/Layouts/CsvLayout.cs
@@ -39,6 +39,7 @@ namespace NLog.Layouts
     using System.Globalization;
     using System.Text;
     using NLog.Config;
+    using NLog.Internal;
 
     /// <summary>
     /// A specialized layout that renders CSV-formatted events.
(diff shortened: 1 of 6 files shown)


src/NLog/Layouts/CsvLayout.cs
```diff
@@ -115,12 +115,13 @@ public CsvLayout()
         /// </summary>
         protected override void InitializeLayout()
         {
-            base.InitializeLayout();
             if (!WithHeader)
             {
                 Header = null;
             }
 
+            base.InitializeLayout();
```
304NotModified: what's the reason for this moved line?

snakefoot: InitializeLayout performs scanning of all referenced objects. No need to scan the Header-property when WithHeader = false.

304NotModified: 👍

304NotModified:
I think merge after confirmation from #2933?

snakefoot:
Not sure if you have read my comments. There is no fix. This is just a small bandaid

Sent from my Sony Xperia


I think merge after confirmation from #2933<https://github.com/NLog/NLog/issues/2933>?

304NotModified:
I saw the comment. I know it isn't a fix, but maybe we need - dunno therefore the question;) -  if this change is significant.

snakefoot:
>  I know it isn't a fix, but maybe we need

Well the Layout-logic caches their async-output on the LogEventInfo. The LogEventInfo are stuffed into the AsyncWrapper-queue, and stays there until written.

If the logger is fast enough, then it will the queue and consume all memory. Not sure how to fix that.

snakefoot:
Made some performance tests, so not just optimizing in the blind:

| Csv Async |      1.000 | 4194304 |    0 |       1 |       1 |

| Master  |    29.286 |        34 | 257 | 257 | 259 |   51.500 |   23.980,0 |
| This PR  |    39.096 |        25 | 124 | 124 | 124 |   58.578 |    7.993,7 |


| Csv Sync |      1.000 | 4194304 |    0 |       1 |       1 |

| Master      |    37.642 |        26 | 225 | 225 | 227 |   41.531 |   15.986,7 |
| This PR   |    43.351 |        23 |   0 |   0 |   0 |   43.281 |        0,2 |


| Csv Async |     10.000 | 153600 |    0 |       1 |       1 |

| Master     |    12.010 |       832 | 1316 | 1316 | 1341 |   18.609 |    8.816,2 |
| This PR    |    14.337 |       697 | 539 | 539 | 539 |   16.687 |    2.939,9 |


| Csv Sync    |     10.000 | 153600 |    0 |       1 |       1 |

| Master     |    13.931 |       717 | 1493 | 1493 | 1493 |   17.968 |    5.875,2 |
| This PR    |    15.674 |       637 |   0 |   0 |   0 |   15.640 |        1,8 |

| Csv Sync   |    500.000 |  150 |    0 |       1 |       1 |

| Master      |     2.746 |   182.021 |   0 |   0 |  65 |    2.734 |      724,9 |
| This PR    |     2.752 |   181.674 |   0 |   0 |   2 |    2.765 |       85,9 |

One can see that the PR introduces a CPU-Time-Increase because the checking for quotes now scans a StringBuilder instead of the allocated string. There is a small optimization if using Quoting=All, then the quote-scan is faster in this PR:

| Csv Sync - Quoting=All |     10.000 | 153600 |    0 |       1 |       1 |

| Master      |    11.740 |       851 | 1387 | 1387 | 1387 |   15.968 |    5.873,9 |
| This PR    |     8.924 |     1.120 |   0 |   0 |   0 |    8.906 |        1,7 |

304NotModified:
that's difficult too choose -  memory vs cpu.  is 23.980,0 MB = ~23 GB? Nice machine if you don't have memory swap issues.

Another option would be to check the "symbols" while writing (in worst case you need a more expensive insert).

snakefoot:
> that's difficult too choose - memory vs cpu.

Not difficult for me. It is only when logging 150KByte messages (or larger) that you see the issue. And in that case you want the reduced allocation. Try looking at the numbers for Size = 150.

> is 23.980,0 MB = ~23 GB? Nice machine if you don't have memory swap issues.

The garbage collector is just doing its job, and cleaning up. It is total allocation not total memory usage.

> Another option would be to check the "symbols" while writing (in worst case you need a more expensive insert).

That would require injection into all LayoutRenderers. Not sure you want to go that way.

304NotModified:
> That would require injection into all LayoutRenderers. Not sure you want to go that way.

What am I missing that we can't use stringbuilder.insert? We are rendering each cell fully isn't?

We have now:

```c#

            if (useQuoting)
            {
                sb.Append(columnValue.Replace(QuoteChar, _doubleQuoteChar));
            }
            else
            {
                sb.Append(columnValue);
            }

```

and that will be something like this?

```c#

                var startIndex = sb.Length;
                var startQuoteWritten = false;
                for (var i = 0; i < columnValue.Length; i++)
                {
                    var c = columnValue[i];

                    if (!startQuoteWritten && isQuoteChar(c))
                    {
                        sb.Insert(startIndex, QuoteChar);
                        startQuoteWritten = true;
                    }

                    sb.Append(c);
                }

                if (startQuoteWritten)
                {
                    sb.Append(QuoteChar);
                }
````

**Comment by snakefoot** [comment]
> What am I missing that we can't use stringbuilder.insert? We are rendering each cell fully isn't?

StringBuilder can only be reused if using Append or truncate by assigning Length. All other operations causes it to discard all internal buffers and reallocate.

**Comment by 304NotModified** [comment]
What about this then? OK, worst case is worse.

```c#
if (useQuoting)
 {
     var startIndex = sb.Length;
     for (var i = 0; i < columnValue.Length; i++)
     {
         var c = columnValue[i];

         if (isQuoteChar(c))
         {
             //discard, we can't use insert
             sb.Length = startIndex;
             //render agian
             sb.Append(QuoteChar);
             sb.Append(columnValue);
             sb.Append(QuoteChar);
         }
         else
         {
             sb.Append(c);
         }
     }
 }
 else
 {
     sb.Append(columnValue);
 }
```

snakefoot:
Also remember that we don't have a columnValue. You have to render it first. And you render into a StringBuilder. The old code then performs a StringBuilder.ToString(), but the new code skips that allocation.

I guess one could have special logic for the message-layoutrenderer so it returns the LogEventInfo.FormatttedMessage / Message (If Raw). But then you would have to recognize layout-wrappers like toupper/truncate/etc.

snakefoot:
I guess if NLog didn't use StringBuilder directly but instead ITextWriter, then it would be easier to recognize transformations on the fly, instead of scanning the StringBuilder. But that would require a heavy rewrite.

304NotModified:
> Also remember that we don't have a columnValue

huh? What's this then? Is it too late today? ;) https://github.com/NLog/NLog/pull/2934/files#diff-422e60153862f43b4bc5eb4941308f02R208

snakefoot:
> huh? What's this then? Is it too late today? ;)

My code:
```
         if (ColumnValueRequiresQuotes(target, orgLength))
         {
                string columnValue = target.ToString(orgLength, target.Length - orgLength);
                target.Length = orgLength;
                target.Append(QuoteChar);
                target.Append(columnValue.Replace(QuoteChar, _doubleQuoteChar));
                target.Append(QuoteChar);
         }
```

Is pretty much identical to your suggested code:
```
         if (charNeedsEscape(c))
         {
             //discard, we can't use insert
             sb.Length = startIndex;
             //render agian
             sb.Append(QuoteChar);
             sb.Append(columnValue);
             sb.Append(QuoteChar);
         }
```

Not sure I understand what you are saying. But yes my code also performs string-allocation when quoting is needed.

snakefoot:
One could consider adding a `Quoting`-property on the CSV-Column. Then one could specify Quoting=All for the specific column with the `${message}` (While the CsvLayout default is Quoting=Auto).

Then it will give performance without forcing quotes for all columns.

304NotModified:
Very nice!

304NotModified:
Thanks, merged!

snakefoot:
Updated docs: https://github.com/NLog/NLog/wiki/CsvLayout

304NotModified:
thanks!

PR #2935, opened 2018-09-25
InternalLogger.LogLevel should default to LogLevel.Off when not configured

### LINKED ISSUE #2913: Configure InternalLogger.LogLevel to Off when loading NLog.config
When loading from LoggingConfiguration and no InternalLogger-destinations has been configured (or LogLevel). Then change the default InternalLogger.LogLevel to Off (Instead of Info).

This will reduce the overhead of the InternalLogger to the absolute minimum.

It will only be users who enable the InternalLogger-destinations (Ex LogWriter) after having loaded Nlog.config that will be affected by this change.

snakefoot:
Resolves  #2913


304NotModified:
thanks!

PR #2937, opened 2018-09-26
JSON encoding should create valid JSON for non-string dictionary-keys

### LINKED ISSUE #2936: ExceptionLayoutRenderer generates invalid JSON for exceptions with Data collection with objects as keys
NLog version: 4.5.6

Layout with exception formatted as JSON:
```xml
<nlog>
    <targets>
        <target name='debug1' type='Debug' layout='${exception:format=@}' />
    </targets>
    <rules>
        <logger minlevel='Info' writeTo='debug1' />
    </rules>
</nlog>
```

Logging exception with Data property where the keys are objects:
```cs
class SomeType {
   public int SomeProperty { get; set; }
}

var someObject = new SomeType() { SomeProperty = 123 };
var ex = new Exception("msg");
ex.Data.Add(someObject, 1);
logger.Error(ex, "msg");
```

Generates invalid JSON in the logs:
```json
{"Message":"msg", "Data":{{_SomeProperty__123}:1}, 
```

It should be:
```json
{"Message":"msg", "Data":{"_SomeProperty__123":1}, 
```

This case is from production system.
`System.Activities.XamlIntegration.ActivityXamlServices.Compile` throw `System.InvalidOperationException` with Data property where the keys are objects.

snakefoot:
Resolve #2936 (See also https://www.newtonsoft.com/json/help/html/SerializationGuide.htm#Dictionarys)

> When serializing a dictionary, the keys of the dictionary are converted to strings and used as the JSON object property names. The string written for a key can be customized by overriding ToString()

### DIFF
--- a/src/NLog/Targets/DefaultJsonSerializer.cs
+++ b/src/NLog/Targets/DefaultJsonSerializer.cs
@@ -355,34 +355,46 @@ namespace NLog.Targets
                     destination.Append(',');
                 }
 
+                if (options.QuoteKeys)
+                {
+                    var typeCode = Convert.GetTypeCode(de.Key);
+                    if (!SerializeObjectAsString(de.Key, typeCode, destination, options))
+                    {
+                        destination.Length = originalLength;
+                        continue;
+                    }
+                }
+                else
+                {
+                    if (!SerializeObject(de.Key, destination, options, objectsInPath, nextDepth))
+                    {
+                        destination.Length = originalLength;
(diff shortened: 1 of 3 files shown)


304NotModified:
Thanks!

snakefoot:
@304NotModified Maybe release NLog 4.5.11 with this bug-fix, since NLog 4.6 is not ready yet.

PR #2940, opened 2018-09-29
JSON encoding should create valid JSON for string dictionary-keys

snakefoot:
Followup to #2937 (Ensure strings are escaped when necessary)


snakefoot:
@304NotModified Maybe release NLog 4.5.11 with this bug-fix, since NLog 4.6 is not ready yet.

PR #2941, opened 2018-09-30
TargetWithContext - Include all properties even when duplicate names (rename with postfix: _1, _2 etc)

snakefoot:
Less surprises when failing to provide unique property names (Without loosing properties)

### DIFF
--- a/src/NLog/Targets/TargetWithContext.cs
+++ b/src/NLog/Targets/TargetWithContext.cs
@@ -152,30 +152,7 @@ namespace NLog.Targets
         /// <returns>Dictionary with any context properties for the logEvent (Null if none found)</returns>
         protected IDictionary<string, object> GetContextProperties(LogEventInfo logEvent, IDictionary<string, object> combinedProperties)
         {
-            if (IncludeGdc)
-            {
-                combinedProperties = CaptureContextGdc(logEvent, combinedProperties);
-            }
-
-            if (IncludeMdc)
-            {
-                if (!CombineProperties(logEvent, _contextLayout.MdcLayout, ref combinedProperties))
-                {
-                    combinedProperties = CaptureContextMdc(logEvent, combinedProperties);
-                }
-            }
-
-#if !SILVERLIGHT
-            if (IncludeMdlc)
-            {
-                if (!CombineProperties(logEvent, _contextLayout.MdlcLayout, ref combinedProperties))
-                {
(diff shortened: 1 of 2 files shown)


304NotModified:
Thanks for this one.

I'm doubting if we should change it like this. It's a behavior change and the naming schema (_1, _2) is debatable. 

Maybe we could do:

- make this optional. E.g. duplicateNamePolicy = lastest | suffix (maybe we should add also "first"). For a minor release it's OK if we change the default, if we could go back to the original behavior. 
- and/or inject the duplicatename ruling. For example, suffixes of just numbers is then possible. Something like 

    ```c#
     /// create dictionary key, stateful 
    interface IKeyGenerator {

        string createKey(string name); 
        bool overwriteDuplicate {get;}
    }
```

snakefoot:
Think you are overenginering. Very few are using TargetWithContext and those are already implementing their own unique key logic

It is the reason it should be default to reduce surprises. Rather have underscore properties than lost information. Think most people agree.

But I can put the duplicate key handling logic in a virtual method so people can override in their custom target

Sent from my Sony Xperia


Thanks for this one.

I'm doubting if we should change it like this. It's a behavior change and the naming schema (_1, _2) is debatable.

Maybe we could do:

  *   make this optional. E.g. duplicateNamePolicy = lastest | suffix (maybe we should add also "first"). For a minor release it's OK if we change the default, if we could go back to the original behavior.

  *   and/or inject the duplicatename ruling. For example, suffixes of just numbers is then possible. Something like

 /// create dictionary key, stateful
interface IKeyGenerator {

    string createKey(string name);
    bool overwriteDuplicate {get;}
}

304NotModified:
> Think you are overenginering

Maybe, but I was triggered by the behavior change. Most important is to me that we could go back to the old behavior 

> But I can put the duplicate key handling logic in a virtual method so people can override in their custom target

That's indeed an easier solution. 👍

snakefoot:
> @304NotModified That's indeed a solution.

Have introduced a new virtual method `GenerateUniqueItemName`

snakefoot:
> Where do you think we should document this behavior (_1, _2 etc)?

Besides as a changelog-note, then one should probably add a section to "how to write custom target" with details about TargetWithContext and AsyncTaskTarget

304NotModified:
Thanks for the changes!

PR #2964, opened 2018-10-22
MethodCallTarget - Added fluent constructor with Layout params

snakefoot:
To allow fluent configuration with additional Layout-logic


snakefoot:
Decided to close this, as I don't like having params in the constructor. Will instead create a fluent-api-method.

PR #2973, opened 2018-10-27
LayoutRenderer - Introduced IStringValueRenderer-interface to skip extra string allocation, removed IRawValue from non-ThreadAgnostic layout renderers

snakefoot:
SimpleLayout with single LayoutRenderer can sometimes convert directly to string without first writing to StringBuilder and then allocating new string:

The most important once are ${logger} and ${level} but there are also settings lookups ${gdc}, ${mdc}, ${mdlc}:

- LevelLayoutRenderer
- LoggerNameLayoutRenderer
- EventPropertiesLayoutRenderer
- GdcLayoutRenderer
- MdlcLayoutRenderer
- MdcLayoutRenderer

Others also included in the optimization:

- TraceActivityIdLayoutRenderer
- ThreadNameLayoutRenderer
- MachineNameLayoutRenderer
- GuidLayoutRenderer
- EnvironmentLayoutRenderer
- DateLayoutRenderer
- CurrentDirLayoutRenderer
- CallSiteFileNameLayoutRenderer

This optimizes Target.RenderLogEvent and Layout.Render for simple layouts that just contains a single dynamic value (based on the above layoutrenderers)

### DIFF
--- a/src/NLog/Internal/FormatHelper.cs
+++ b/src/NLog/Internal/FormatHelper.cs
@@ -49,11 +49,10 @@ namespace NLog.Internal
         /// </remarks>
         internal static string ConvertToString(object o, IFormatProvider formatProvider)
         {
-
             // if no IFormatProvider is specified, use the Configuration.DefaultCultureInfo value.
             if (formatProvider == null)
             {
-                if (!ObjectSupportsFormatProvider(o))
+                if (SkipFormattableToString(o))
                     return o?.ToString() ?? string.Empty;
 
                 if (o is IFormattable)
@@ -69,19 +68,19 @@ namespace NLog.Internal
             return Convert.ToString(o, formatProvider);
         }
 
-        internal static bool ObjectSupportsFormatProvider(object o)
+        private static bool SkipFormattableToString(object o)
         {
             switch (Convert.GetTypeCode(o))
             {
-                case TypeCode.String:   return false;
-                case TypeCode.Empty:    return false;
-                default:                return true;
+                case TypeCode.String:   return true;
+                case TypeCode.Empty:    return true;
+                default:                return false;
             }
         }
 
         internal static string TryFormatToString(object value, string format, IFormatProvider formatProvider)
         {
-            if (!ObjectSupportsFormatProvider(value))
+            if (SkipFormattableToString(value))
                 return value?.ToString() ?? string.Empty;
 
             if (value is IFormattable formattable)
--- a/src/NLog/Internal/FormatHelper.cs
+++ b/src/NLog/Internal/FormatHelper.cs
@@ -49,17 +49,21 @@ namespace NLog.Internal
         /// </remarks>
         internal static string ConvertToString(object o, IFormatProvider formatProvider)
         {
+
             // if no IFormatProvider is specified, use the Configuration.DefaultCultureInfo value.
             if (formatProvider == null)
             {
                 if (!ObjectSupportsFormatProvider(o))
                     return o?.ToString() ?? string.Empty;
 
-                //variable so only 1 lock is needed
-                //TODO this locks the configuration, which can lead to deadlocks.
-                var loggingConfiguration = LogManager.Configuration;
-                if (loggingConfiguration != null)
-                    formatProvider = loggingConfiguration.DefaultCultureInfo;
+                if (o is IFormattable)
+                {
+                    //variable so only 1 lock is needed
+                    //TODO this locks the configuration, which can lead to deadlocks.
+                    var loggingConfiguration = LogManager.Configuration;
+                    if (loggingConfiguration != null)
+                        formatProvider = loggingConfiguration.DefaultCultureInfo;
+                }
             }
 
             return Convert.ToString(o, formatProvider);
@@ -71,8 +75,26 @@ namespace NLog.Internal
             {
                 case TypeCode.String:   return false;
                 case TypeCode.Empty:    return false;
-                case TypeCode.Object:   return o is IFormattable;
                 default:                return true;
+            }
+        }
+
+        internal static string TryFormatToString(object value, string format, IFormatProvider formatProvider)
+        {
+            if (!ObjectSupportsFormatProvider(value))
+                return value?.ToString() ?? string.Empty;
+
+            if (value is IFormattable formattable)
+            {
+                return formattable.ToString(format, formatProvider);
+            }
+            else if (value is System.Collections.IEnumerable)
+            {
+                return null;
+            }
+            else
+            {
+                return value.ToString() ?? string.Empty;
             }
         }
     }
--- a/src/NLog/Internal/IRenderString.cs
+++ b/src/NLog/Internal/IRenderString.cs
@@ -1,4 +1,4 @@
-// 
+﻿// 
 // Copyright (c) 2004-2018 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
 // 
 // All rights reserved.
@@ -31,56 +31,18 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
-namespace NLog.LayoutRenderers
+namespace NLog.Internal
 {
-    using System;
-    using System.ComponentModel;
-    using System.Text;
-    using NLog.Config;
-    using NLog.Internal;
-
     /// <summary>
-    /// The log level.
+    /// Supports rendering of string value without allocation (or seldom)
     /// </summary>
-    [LayoutRenderer("level")]
-    [ThreadAgnostic]
-    [ThreadSafe]
-    public class LevelLayoutRenderer : LayoutRenderer, IRawValue, IRenderString
+    interface IRenderString
     {
         /// <summary>
-        /// Gets or sets a value indicating the output format of the level.
+        /// Renders the value of layout renderer in the context of the specified log event
         /// </summary>
-        /// <docgen category='Rendering Options' order='10' />
-        [DefaultValue(LevelFormat.Name)]
-        public LevelFormat Format { get; set; }
-
-        /// <inheritdoc/>
-        protected override void Append(StringBuilder builder, LogEventInfo logEvent)
-        {
-            LogLevel level = GetValue(logEvent);
-            switch (Format)
-            {
-                case LevelFormat.Name:
-                    builder.Append(level.ToString());
-                    break;
-                case LevelFormat.FirstCharacter:
-                    builder.Append(level.ToString()[0]);
-                    break;
-                case LevelFormat.Ordinal:
-                    builder.AppendInvariant(level.Ordinal);
-                    break;
-            }
-        }
-
-        /// <inheritdoc/>
-        object IRawValue.GetRawValue(LogEventInfo logEvent) => GetValue(logEvent);
-
-        /// <inheritdoc/>
-        string IRenderString.GetFormattedString(LogEventInfo logEvent) => Format == LevelFormat.Name ? GetValue(logEvent).ToString() : null;
-
-        private static LogLevel GetValue(LogEventInfo logEvent)
-        {
-            return logEvent.Level;
-        }
+        /// <param name="logEvent"></param>
+        /// <returns>null if not possible or unknown</returns>
+        string GetFormattedString(LogEventInfo logEvent);
     }
 }
--- a/src/NLog/LayoutRenderers/EventPropertiesLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/EventPropertiesLayoutRenderer.cs
@@ -46,7 +46,7 @@ namespace NLog.LayoutRenderers
     [ThreadAgnostic]
     [ThreadSafe]
     [MutableUnsafe]
-    public class EventPropertiesLayoutRenderer : LayoutRenderer, IRawValue, IRenderString
+    public class EventPropertiesLayoutRenderer : LayoutRenderer, IRawValue, IStringValueRenderer
     {
         /// <summary>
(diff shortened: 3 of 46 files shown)


snakefoot:
Changed behavior for SimpleLayout.TryGetRawValue() so it checks if Layout.Precalculate has been performed, and uses the cached value instead of the raw value (if available).

304NotModified:
> Changed behavior for SimpleLayout.TryGetRawValue(), so it checks if Layout.Precalculate has been performed, and uses the cached value instead of the raw value (if available). Is this a good fallback

I think this is OK, but not sure. The cache is per logevent, isn't? So then I think it's OK

snakefoot:
> I think this is OK, but not sure. The cache is per logevent, isn't? So then I think it's OK

The logic calling TryGetRawValue() just need to be aware that it will suddenly return a string-object instead of ex. a DateTime-object, when the DatabaseTarget is wrapped in Async.

304NotModified:
> The logic calling TryGetRawValue() just need to be aware that it will suddenly return a formatted string-object instead of ex. a DateTime-object, when the DatabaseTarget is wrapped in Async.

Ow there is the TryGetRawValue not built for. Maybe we should wrap it in a "TryGetRawValueOrFallback", so the implementor could use the parse fallback

snakefoot:
> Ow there is the TryGetRawValue not built for. Maybe we should wrap it in a "TryGetRawValueOrFallback", so the implementor could use the parse fallback

Now changed so SimpleLayout will return false in TryGetValue when raw-value comes from MutableUnsafe-Layout and the value is not immutable. Also removed IRawValue-interface from all LayoutRenderers that are not ThreadAgnostic.

snakefoot:
@304NotModified Ready for review. Now IRawValue is only available when thread-agnostic to make it more consistent when using using `targets async="true"`

src/NLog/Layouts/SimpleLayout.cs
```diff
@@ -138,6 +141,11 @@ public string Text
         /// </summary>
         public string FixedText => _fixedText;
 
+        /// <summary>
+        /// Is the message a simple formatted string? (Can skip StringBuilder)
+        /// </summary>
+        internal bool IsSimpleString => _stringValueRenderer != null;
```
304NotModified: Was thinking of `IsSingleStringText`. "simple" is a bit ambiguous here.

src/NLog/Internal/FormatHelper.cs
```diff
@@ -71,9 +75,27 @@ internal static bool ObjectSupportsFormatProvider(object o)
             {
                 case TypeCode.String:   return false;
                 case TypeCode.Empty:    return false;
-                case TypeCode.Object:   return o is IFormattable;
```
304NotModified: maybe we should rename `ObjectSupportsFormatProvider` now to `SkipFormattableToString`, or something like that.

src/NLog/Layouts/SimpleLayout.cs
304NotModified: I think the name of `_stringValueRenderer ` should be in-sync with "simpleString" (or the new name)

src/NLog/Layouts/SimpleLayout.cs
```diff
@@ -222,14 +230,28 @@ internal void SetRenderers(LayoutRenderer[] renderers, string text)
...
+                }
+                else
+                {
+                    if (Renderers[0] is IRawValue rawValueRendrer)
+                    {
+                        _rawValueRenderer = rawValueRendrer;
+                    }
+                    if (Renderers[0] is IRenderString stringValueRendrer)
+                    {
+                        _stringValueRenderer = stringValueRendrer;
```
304NotModified: I like this optimization :)

src/NLog/Internal/IRenderString.cs
```diff
@@ -0,0 +1,48 @@
...
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF 
+// THE POSSIBILITY OF SUCH DAMAGE.
+// 
+
+namespace NLog.Internal
+{
+    /// <summary>
+    /// Supports rendering of string value without allocation (or seldom)
+    /// </summary>
+    interface IRenderString
```
304NotModified: proposal: IStringRenderable, as class/interfaces should not contain verbs

src/NLog/LayoutRenderers/GuidLayoutRenderer.cs
```diff
@@ -44,22 +44,14 @@ namespace NLog.LayoutRenderers
     /// </summary>
     [LayoutRenderer("guid")]
     [ThreadSafe]
-    public class GuidLayoutRenderer : LayoutRenderer, IRawValue
```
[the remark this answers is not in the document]
304NotModified: (e.g. uniqueidentifier  type in SQL Server)

src/NLog/LayoutRenderers/EventPropertiesLayoutRenderer.cs
```diff
@@ -90,19 +78,34 @@ protected override void Append(StringBuilder builder, LogEventInfo logEvent)
...
 
-        /// <summary>
-        /// Get raw value, updates the sequence
-        /// </summary>
-        object IRawValue.GetRawValue(LogEventInfo logEvent)
+        private string GetStringValue(LogEventInfo logEvent)
         {
-            GetValue(logEvent, out var value);
-            return value;
+            if (Format != "@")
```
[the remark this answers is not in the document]
snakefoot: Been there since NLog 4.5. Means serialize as JSON.

src/NLog/LayoutRenderers/GuidLayoutRenderer.cs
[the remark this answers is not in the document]
snakefoot: I guess it can be made ThreadAgnostic. Not like the Guid will be much different being created on Timer-thread.

src/NLog/LayoutRenderers/LoggerNameLayoutRenderer.cs
```diff
@@ -52,29 +53,33 @@ public class LoggerNameLayoutRenderer : LayoutRenderer
...
+            builder.Append(logEvent.LoggerName);
+        }
+
+        /// <inheritdoc/>
+        string IRenderString.GetFormattedString(LogEventInfo logEvent)
+        {
+            if (ShortName)
             {
-                builder.Append(logEvent.LoggerName);
+                int lastDot = logEvent.LoggerName?.LastIndexOf('.') ?? -1;
```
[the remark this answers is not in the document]
snakefoot: It is duplicated for performance reasons

src/NLog/Layouts/SimpleLayout.cs
snakefoot: IsSingleStringText is fine with me. Will fix this

src/NLog/Internal/FormatHelper.cs
snakefoot: Ok. Will fix this.

src/NLog/LayoutRenderers/LoggerNameLayoutRenderer.cs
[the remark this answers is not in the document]
304NotModified: Could you please add a code comment for it then? thx!

src/NLog/Internal/IRenderString.cs
snakefoot: Ok. Will fix this

src/NLog/LayoutRenderers/LoggerNameLayoutRenderer.cs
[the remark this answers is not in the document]
snakefoot: Ok. Will fix this

src/NLog/LayoutRenderers/GuidLayoutRenderer.cs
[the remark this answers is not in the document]
304NotModified: I think this is also ThreadAgnostic, because it creates always a new guid  / or it uses only values of the logevent.

src/NLog/Targets/Target.cs
```diff
@@ -744,6 +746,11 @@ protected string RenderLogEvent(Layout layout, LogEventInfo logEvent)
                     }
                 }
 
+                if (simpleLayout != null && simpleLayout.IsSimpleString)
```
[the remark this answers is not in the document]
snakefoot: Because IsFixedText means it is threadagnostic. But IsSimpleString is not guaranteed to be threadagnostic. So have to check precalculate-cache first.

tests/NLog.UnitTests/Layouts/SimpleLayoutOutputTests.cs
```diff
@@ -188,7 +206,23 @@ public void TryGetRawValue_MultipleLayoutRender_ShouldGiveNullRawValue()
             Assert.Null(value);
         }
 
+        [Fact]
+        public void TryGetRawValue_ImmutableLayoutRender_ShouldGiveRawValue()
```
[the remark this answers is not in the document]
snakefoot: Will fix it so it becomes Guid.

src/NLog/LayoutRenderers/GuidLayoutRenderer.cs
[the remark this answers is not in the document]
snakefoot: Will fix this.

snakefoot:
@304NotModified Implemented all suggested changes

src/NLog/Internal/IRenderString.cs
snakefoot: Renamed to ISimpleStringText

src/NLog/LayoutRenderers/MdcLayoutRenderer.cs
```diff
@@ -58,30 +58,32 @@ public class MdcLayoutRenderer : LayoutRenderer, IRawValue
...
 
-        /// <inheritdoc />
-        object IRawValue.GetRawValue(LogEventInfo logEvent)
+        /// <inheritdoc/>
+        string IRenderString.GetFormattedString(LogEventInfo logEvent) => GetStringValue(logEvent);
+
+        private string GetStringValue(LogEventInfo logEvent)
         {
-            return GetValue();
+            if (Format != "@")
```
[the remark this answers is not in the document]
snakefoot: Placed the magic string-value on ValueFormatter.

src/NLog/Internal/ISimpleStringText.cs
```diff
@@ -0,0 +1,48 @@
...
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF 
+// THE POSSIBILITY OF SUCH DAMAGE.
+// 
+
+namespace NLog.Internal
+{
+    /// <summary>
+    /// Supports rendering of string value without allocation (or seldom)
+    /// </summary>
+    interface ISimpleStringText
```
304NotModified: I was also thinking about this, but not sure if this is a good name. 

- IsSimpleStringText is only when there is one ISimpleStringText, so then that property should be renamed to `IsSingleSimpleText`... 

Another option is `IFormattedTextRendererable` (better name is desirable)

src/NLog/Internal/IRenderString.cs
snakefoot: Maybe ISimpleStringValue is better than ISimpleStringText ? (Similar to IRawValue)

src/NLog/Internal/ISimpleStringText.cs
snakefoot: ISimpleStringValue ? (Similar to IRawValue)

304NotModified: I think that a good idea. :)

snakefoot: Maybe a combination IStringValueRenderer

snakefoot: Maybe even consider changing IRawValue to IRawValueRenderer

src/NLog/Internal/ISimpleStringText.cs
```diff
@@ -0,0 +1,48 @@
...
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN 
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) 
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF 
+// THE POSSIBILITY OF SUCH DAMAGE.
+// 
+
+namespace NLog.Internal
+{
+    /// <summary>
+    /// Supports rendering of string value without allocation (or seldom)
```
304NotModified: Maybe we should explain "seldom" (`<remarks>`?). When should we implement this interface? When is it seldom? e.g. in the happy path no allocations? Or not many allocations?

snakefoot: Seldom means Guid.ToString() or DateTime.ToString. I can change the comment to "limited or no allocations."

snakefoot: Updated the xml-comment

src/NLog/Internal/ISimpleStringText.cs
304NotModified: I don't prefer "renderer" in the interface, it doesn't render IMO (especially not the rawValue).

But luckily they are both internal, so we could change it later.

304NotModified:
Thanks for the changes!

PR #2974, opened 2018-10-28
Feature/update resharper annotations

imanushin:
Part of #2952

PR #2975, opened 2018-10-28
Feature/rename message template format method attribute to resharper like

imanushin:
**Important:** ignore this PR until #2974 will be merged.

Question: are you ok with attribute replacing? Please note, that old attribute was removed, however I put new in the each line.

After change:
image

Before change:
image


304NotModified:
> Question: are you ok with attribute replacing? Please note, that old attribute was removed, however I put new in the each line.

Do you mean `MessageTemplateFormatMethod("format")]` ->	        `[StringFormatMethod("format")]
`

unfortunately, that won't work well with the message templates.

imanushin:
> unfortunately, that won't work well with the message templates.

Why? I can add attribute ```[StringFormatMethod("format")]``` for each method with ```[MessageTemplateFormatMethod("format")]```. The result will be the same with on my screenshots.

So questions:
1. Do we want to have pretty suggestions by ReSharper (as on screenshots)?
2. If yes: are you ok with attributes duplicating?
3. If "3"'s answer is "no" - do you have any suggestions, how to achieve this goal?

304NotModified:
> Why? I can add attribute `[StringFormatMethod("format")]` for each method with `[MessageTemplateFormatMethod("format")]`. The result will be the same with on my screenshots.

You get then "false positives" from Resharper?

304NotModified:
e.g. this:

image


when using ` [StringFormatMethod("message")]`

imanushin:
@304NotModified , I see, I agree. Will reject

PR #2983, opened 2018-10-31
JsonSerializer - Use ReferenceEquals instead of object.Equals

### LINKED ISSUE #2980: Add configurable json serialization option for checking reference via ReferenceEquals instead of Equals method
Hi!
I have generated wsdl, with (little bit strange) structure that contains value wrappers for simple types. StringValue for string, IntValue for int, that contains a property with such type.
For better use in my code, I extended all such values and overridden Equals, GetHashCode, ToString methods to work with inner value. So StringValue with Value = "some string" will return true for Equal on the same string.

The problem is: while serializing argument for a log, there are HashSet used for checking items, already added to the collection. In my case, when StringValue added, it is persisted in a hashset, and when it take Value of it, it's not added, because it Equals returns true, because Equals is overridden.

The proposed solution is to add configuration, that can be called like "useReferenceEqualsComparasion" to json logging section:
`        <attribute name="properties" encode="false">
          <layout  xsi:type='JsonLayout' includeAllProperties="true" includeMdc="true" renderEmptyObject="true" maxRecursionLimit="4" />
        </attribute>`
That will instructs to use custom EqualityComparer for this hashset (in class SingleItemOptimizedHashSet) that will call ReferenceEquals not a Equals for items comparsion.

snakefoot:
When checking for cyclic object loops. Resolves #2980

### DIFF
--- a/src/NLog/Internal/SingleItemOptimizedHashSet.cs
+++ b/src/NLog/Internal/SingleItemOptimizedHashSet.cs
@@ -45,6 +45,9 @@ namespace NLog.Internal
     {
         private readonly T _singleItem;
         private HashSet<T> _hashset;
+        private readonly IEqualityComparer<T> _comparer;
+
+        private IEqualityComparer<T> Comparer => _comparer ?? EqualityComparer<T>.Default;
 
         public struct SingleItemScopedInsert : IDisposable
         {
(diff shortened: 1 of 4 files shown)


snakefoot:
@304NotModified Should be considered for NLog ver. 4.5.11. It is actually a bug that it doesn't use object-reference-compare instead of object-equals.

304NotModified:
@snakefoot @2987 is merged!

304NotModified:
PS: there is no need for a new PR, after force push we can change the target branch. That is preferred

snakefoot:
@304NotModified Will try. But your refactoring of the JsonLayout tests is causing merge conflicts between master and dev.

snakefoot:
@304NotModified Ready for commit to master. But I think you will find it difficult to get into dev.

304NotModified:
Thanks!

snakefoot:
@304NotModified Should I make new PR for dev-branch ?

snakefoot:
@304NotModified #2989 also includes the changes

PR #2991, opened 2018-11-04
SubstringLayoutRenderer - Fixed bug when Start is higher than actual length + improve tests

snakefoot:
See also https://travis-ci.org/NLog/NLog/builds/450439888


snakefoot:
Have now changed the NLog.UnitTests-project so LogManager.ThrowExceptions = true by default. Instead of being at random depending on the test-execution-order.

304NotModified:
> Have now changed the NLog.UnitTests-project so LogManager.ThrowExceptions = true by default. Instead of being at random depending on the test-execution-order.
> 
> No longer hiding exceptions like those in Substring-Unit-Tests.

good one!

PR #2996, opened 2018-11-06
Improve null handling and annotations for LogEventInfo

304NotModified:


snakefoot:
LoggerName is rarely null, unless access before having been sent to the Logger. See also #2917

304NotModified:
indeed, but there is not check in the ctor, so it could be. (unfortunately)

PR #2998, opened 2018-11-06
Added LoggingConfiguration.Configure(..)

304NotModified:
Fixes https://github.com/NLog/NLog/issues/1429

Issue #1429, opened 2016-05-02
Add Configuration.CombineConfig(config => config)


304NotModified:
Combining XML config with C# config is now tricky, as autoreload/load of the XML file could overwrite the C# config.

We can use the events for the on LogManager, but it will be more clear to create a method with receives a lambda. 

e.g.

``` c#
LogManager.Configuration.Combine(logConfig => 
                                  {
                                      logConfig.AddTarget(..);
                                  });
```

currently needed, the `ConfigurationReloaded` event

```c#
using System;
using NLog;
using NLog.Config;
using NLog.Targets;

class Program
{
    // Define a static logger variable so that it references the logger instanced named "Scribe"
    private static readonly Logger log = LogManager.GetCurrentClassLogger();

    static void Main(string[] args)
    {
        ExtendNLogConfig();

        LogManager.ConfigurationReloaded += LogManager_ConfigurationReloaded;

        log.Info("Entering Application.");

        Console.WriteLine("Press any key to exit ...");
        Console.Read();
    }


    private static void LogManager_ConfigurationReloaded(object sender, LoggingConfigurationReloadedEventArgs e)
    {
        ExtendNLogConfig();
    }

    /// <summary>
    /// Extend the logging rules in the nlog.config with programmically rules.
    /// </summary>
    private static void ExtendNLogConfig()
    {
        //don't set  LogManager.Configuration because that will overwrite the nlog.config settings
        var consoleTarget = new ColoredConsoleTarget();
        consoleTarget.Name = "console";
        LogManager.Configuration.AddTarget("console", consoleTarget);

        consoleTarget.Layout = @"${date:format=HH\:mm\:ss} ${logger} ${message} ${exception} ${event-context:item=MyValue}";
        var rule1 = new LoggingRule("*", LogLevel.Debug, consoleTarget);
        LogManager.Configuration.LoggingRules.Add(rule1);
        LogManager.ReconfigExistingLoggers();
    }
}

```

304NotModified:
Need good name for it

MikeFH:
A few remarks :
- Should the new method be on `LogManager` or `LogManager.Configuration` ?
- I suppose the overload would be applied on config reload but should it be also applied on config change ?
- Can the existing events on `LogManager` (`ConfigReloaded `and `ConfigChanged`) be used since we might want the overload to be applied before registered handlers to these events are executed ?
- Maybe `Overload` or `OverloadConfiguration` could be good names for the method
[...]

PR #3000, opened 2018-11-06
Added Fluent configuration API

snakefoot:
Alternative solution for #1429 (See also #2998)

It attempts to guide the user to configure the NLog configuration the right way.

Ex.

```C#
var factory = LogManager.LogFactory.BuildConfig(c =>
   c.FilterMinLevel(LogLevel.Debug,
      w => w.WriteToWithAsync(async =>
              async.WriteToFile(fileName: "file.txt")));
   c.FilterMinLevel(LogLevel.Info
      w => w.WriteToWithAsync(async =>
              async.WriteToConsole()));
);
var logger = factory.GetCurrentClassLogger();
```

### DIFF
--- a/src/NLog/ConfigurationBuilder.cs
+++ b/src/NLog/ConfigurationBuilder.cs
@@ -1,0 +1,359 @@
+﻿// 
+// Copyright (c) 2004-2018 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without 
+// modification, are permitted provided that the following conditions 
+// are met:
+// 
+// * Redistributions of source code must retain the above copyright notice, 
+//   this list of conditions and the following disclaimer. 
+// 
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution. 
+// 
+// * Neither the name of Jaroslaw Kowalski nor the names of its 
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission. 
+// 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE 
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE 
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE 
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR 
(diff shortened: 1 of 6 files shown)


304NotModified:
cool!

I'm not sure if this is the same as #2998. The goal of #2998 is that another config (e.g. xml) will be enhanced from C#. That's not always needed/preferred. 

Of course we could also tackle that with

```
var factory = LogManager.Configure(c =>
    c.LoadFromFile("nlog.config")
```


Some ideas:
---

 ```
var factory = LogManager.LogFactory.BuildConfig(c =>
```

I think I would prefer 
```
var factory = LogManager.Configure(c =>
```

---

```
 w => w.WriteToWithAsync(async =>
```
what about:
```
 w => w.WriteAsync(async =>
```
or would this also possible?

```
var factory = LogManager.LogFactory.BuildConfig(c =>
   c.FilterMinLevel(LogLevel.Debug,
      w => w.Async().WriteToFile(fileName: "file.txt")));
   c.FilterMinLevel(LogLevel.Info
      w => w.Async().WriteToConsole()));
);
```

snakefoot:
Yes adding a static `BuildConfig`-method to `LogManager` should not be difficult.

I like `WriteToWithAsync` because it matches the other `WriteTo` methods, that indicates that one is doing target-configuration. But I also don't like it because it is rather "wordy".

It would also be possible to use `w.async().WriteToFile`. Should be a simple change

304NotModified:
> But then it would be different from compound-targets in their API.

I don't think most users will think async as an compound thing

snakefoot:
> I don't think most users will think async as an compound thing. So OK to me :)

I was talking about compound-wrappers like the FallbackGroup. Like when interfaces act similar.

304NotModified:
> I was talking about compound-wrappers like the FallbackGroup. Like when interfaces act similar.

Yes, and I think this is OK: 

> But then it would be different from compound-targets in their API.

snakefoot:
> Yes, and I think this is OK:

So how should the WriteToWithFallback work then?

`WriteToWithFallback(f => f.WriteToConsole().WriteToDebug());`

`WriteToWithAsync(f => f.WriteToConsole());`

Also remember one can chain them together in any order, as NLog supports.

304NotModified:
> So how should the WriteToWithFallback work then?

Do you mean the API design, or how it should work technically?

snakefoot:
Everything in this PR is about API-design. Making a nice API is the hardest part.

But maybe it should cook a little longer, and not hold NLog 4.6 back ?

304NotModified:
> But maybe it should cook a little longer, and not hold NLog 4.6 back ? It was just something I thought could be nice and made an attempt.

I think that also wise. Only https://github.com/NLog/NLog/pull/2899 is really blocking 4.6, but unfortunately it should have more unit tests before merge.

304NotModified:
but what about https://github.com/NLog/NLog/pull/2998, that one will hurt this one in the future?

snakefoot:
> but what about #2998, that one will hurt this one in the future?

Yes it will be a little confusing with many different fluent-config-approaches. Else it should be a minimal builder-object that has LoggingConfiguration as property ? (Then it can grow from there).

304NotModified:
@snakefoot thanks for this PR, but gonna close this one for now as it isn't active and I try to focus the stuff of 4.6

snakefoot:
Created  #4128 that now extends the `LogFactory.Setup()`-logic

PR #3003, opened 2018-11-07
${json-encode} support for rawValue

### LINKED ISSUE #2856: RawValue should work with xml-encode en json-encode layout renderers
This should work if the #1332 is working:

```xml
<parameter name="@Request1" layout="${xml-encode:inner=${event-properties:value1}}" />
<parameter name="@Request2" layout="${json-encode:inner=${event-properties:value1}}" />
```

304NotModified:
Fixes #2856 

todo 

- [ ] xml-encode

### DIFF
--- a/src/NLog/LayoutRenderers/Wrappers/JsonEncodeLayoutRendererWrapper.cs
+++ b/src/NLog/LayoutRenderers/Wrappers/JsonEncodeLayoutRendererWrapper.cs
@@ -31,6 +31,8 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
+using NLog.Targets;
+
 namespace NLog.LayoutRenderers.Wrappers
 {
     using System;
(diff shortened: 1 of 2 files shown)


src/NLog/LayoutRenderers/Wrappers/JsonEncodeLayoutRendererWrapper.cs
```diff
@@ -70,9 +72,22 @@ public JsonEncodeLayoutRendererWrapper()
         [DefaultValue(true)]
         public bool EscapeUnicode { get; set; }
 
+        /// <summary>
+        /// Render with use of the Raw value, if possible.
+        /// </summary>
+        public bool Raw { get; set; }
```
304NotModified: not sure with this name

src/NLog/LayoutRenderers/Wrappers/JsonEncodeLayoutRendererWrapper.cs
```diff
@@ -70,9 +72,22 @@ public JsonEncodeLayoutRendererWrapper()
...
+
         /// <inheritdoc/>
         protected override void RenderInnerAndTransform(LogEventInfo logEvent, StringBuilder builder, int orgLength)
         {
+            if (Raw && Inner.TryGetRawValue(logEvent, out var rawValue))
+            {
+                ConfigurationItemFactory.Default.JsonConverter.SerializeObject(rawValue, builder); //todo missing EscapeUnicode
```
304NotModified: todo add warn for unsupported EscapeUnicode?

snakefoot:
What is the difference between this and `${event-properties:prop_key:format=@}` ?

304NotModified:
> Remember that I have now restricted RawValue to only be primitive types, and not full objects.

yes, that is indeed an issue with the original idea after this.

> ${event-properties:prop_key:format=@}

good one. Haven't think of that

304NotModified:
> What is the difference between this and `${event-properties:prop_key:format=@}` ?

after some thinking:

- '@' won't work for XML
- this should also work for GDC etc (but not now because of the change of IRawValue there)
- This is needed to serialize ${exception} to JSON?

snakefoot:
> '@' won't work for XML

Nope it is a special format-option from the message-template-world.

> this should also work for GDC

GdcLayoutRenderer also has support for the Format-parameter = '@'

> This is needed to serialize ${exception} to JSON?

ExceptionLayoutRenderer also has support for the Format-parameter = '@'

304NotModified:
close in favor of the current option (`@` format)

Issue #1948, opened 2017-02-10
Add aliases option for layout renderer


304NotModified:
For consistency and backwardscompatiblity

e.g.

* ` [LayoutRenderer("activityid")]` -> ` [LayoutRenderer("activity-id", "activityid")]`
* ` [LayoutRenderer("logger")]` -> ` [LayoutRenderer("logger", "loggername")]`

UgurAldanmaz:
Would we like to release this one with the next version?

304NotModified:
It would be nice to target 4.6 for this :)

PR #3004, opened 2018-11-07
Support for aliases for LayoutRenderer, targets, conditions etc. + Added some aliases

304NotModified:
fixes https://github.com/NLog/NLog/issues/1948

todo

- [x] test multiple aliases
- [x] add test for target and layout

### DIFF
--- a/src/NLog/Config/Factory.cs
+++ b/src/NLog/Config/Factory.cs
@@ -95,7 +95,7 @@ namespace NLog.Config
             IEnumerable<TAttributeType> attributes = type.GetCustomAttributes<TAttributeType>(false);
             if (attributes != null)
             {
-                foreach (TAttributeType attr in attributes)
+                foreach (var attr in attributes)
                 {
                     RegisterDefinition(itemNamePrefix + attr.Name, type);
                 }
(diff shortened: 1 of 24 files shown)


snakefoot:
Think it is "Aliases" (Instead of Aliasses)

304NotModified:
Thanks!

304NotModified:
@snakefoot 

please your opinion, I was thinking, maybe this is better:

```c#
    [Target("Database")]
    [Target("DB")]
    public class DatabaseTarget : Target, IInstallable
```

now in this PR:

```c#
   [Target("Database", Aliases = "DB")]
```

snakefoot:
Think I like this better:
```
    [Target("Database")]
    [Target("DB")]
```
But don't actually see a need for the "DB"-alias :)

But I can see this makes sense:
```
    [LayoutRenderer("event-properties")]
    [LayoutRenderer("event-property")]
```

304NotModified:
Yeah, it was primary for the layout renderers (also for ${document-uri})

304NotModified:
OK thanks for the feedback, done :)

snakefoot:
@304NotModified Please update the Wiki with whatever details you find necessary (And when done then add tag with docs completed)

PR #3013, opened 2018-11-11
Removed use of lock(this), removed dead code

snakefoot:
Removed use of lock(this) to make code analysis happy

### DIFF
--- a/src/NLog/Config/XmlLoggingConfiguration.cs
+++ b/src/NLog/Config/XmlLoggingConfiguration.cs
@@ -470,20 +470,20 @@ namespace NLog.Config
         /// <summary>
         /// Parses a single config section within the NLog-config
         /// </summary>
-        /// <param name="section"></param>
+        /// <param name="configSection"></param>
         /// <returns>Section was recognized</returns>
-        protected override bool ParseNLogSection(ILoggingConfigurationElement section)
+        protected override bool ParseNLogSection(ILoggingConfigurationElement configSection)
         {
-            if (section.MatchesName("include"))
+            if (configSection.MatchesName("include"))
             {
                 string filePath = _currentFilePath.Peek();
-                bool autoLoad = filePath != null ? _fileMustAutoReloadLookup[GetFileLookupKey(filePath)] : false;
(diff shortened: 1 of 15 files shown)


src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -470,20 +470,20 @@ private void ParseNLogElement(ILoggingConfigurationElement nlogElement, string f
         /// <summary>
         /// Parses a single config section within the NLog-config
         /// </summary>
-        /// <param name="section"></param>
+        /// <param name="configSection"></param>
         /// <returns>Section was recognized</returns>
-        protected override bool ParseNLogSection(ILoggingConfigurationElement section)
+        protected override bool ParseNLogSection(ILoggingConfigurationElement configSection)
```
304NotModified: unfortunately this is a binary breaking change

src/NLog/Targets/DefaultJsonSerializer.cs
```diff
@@ -432,26 +432,6 @@ private void SerializeCollectionObject(IEnumerable value, StringBuilder destinat
             destination.Append(']');
         }
 
-        private bool SerializeTypeCodeValue(object value, TypeCode objTypeCode, StringBuilder destination, JsonSerializeOptions options, SingleItemOptimizedHashSet<object> objectsInPath, int depth)
```
304NotModified: this is not needed anymore? Please do describe these changes in the PR title to make the review less of a surprise - thanks!

304NotModified: I will update it for now

src/NLog/Config/XmlLoggingConfiguration.cs
snakefoot: Pretty sure we have not released NLog 4.6 yet

src/NLog/Targets/DefaultJsonSerializer.cs
snakefoot: Thx

src/NLog/Config/XmlLoggingConfiguration.cs
304NotModified: ah yes, then it's OK indeed :)

PR #3017, opened 2018-11-13
Improve EventLogTarget installation logic - handle MaximumKilobytes for apps without elevated privileges

Coriolanuss:
Remove MaximumKilobytes setting from GetEventLog(), because when app is run without elevated privileges, the method spams exceptions.
Move MaximumKilobytes setting to installation (and initialization) phase, when the application is typically run with elevated privileges.
Add logic to not override a bigger value of MaximumKilobytes, because the Event log can be used by multiple Sources, each having its own space requirements.

### DIFF
--- a/src/NLog/Targets/EventLogTarget.cs
+++ b/src/NLog/Targets/EventLogTarget.cs
@@ -194,10 +194,8 @@ namespace NLog.Targets
         /// <param name="installationContext">The installation context.</param>
         public void Install(InstallationContext installationContext)
         {
-            var fixedSource = GetFixedSource();
-
-            //always throw error to keep backwardscomp behavior.
-            CreateEventSourceIfNeeded(fixedSource, true);
+            // always throw error to keep backwards compatible behavior.
+            CreateEventSourceIfNeeded(GetFixedSource(), true);
         }
 
         /// <summary>
@@ -244,20 +242,7 @@ namespace NLog.Targets
         {
             base.InitializeTarget();
 
-            var fixedSource = GetFixedSource();
-
-            if (string.IsNullOrEmpty(fixedSource))
-            {
-                InternalLogger.Debug("EventLogTarget(Name={0}): Skipping creation of event source because it contains layout renderers", Name);
-            }
-            else
-            {
-                var currentSourceName = EventLog.LogNameFromSourceName(fixedSource, MachineName);
-                if (!currentSourceName.Equals(Log, StringComparison.CurrentCultureIgnoreCase))
-                {
-                    CreateEventSourceIfNeeded(fixedSource, false);
-                }
-            }
+            CreateEventSourceIfNeeded(GetFixedSource(), false);
         }
 
         /// <summary>
@@ -383,11 +368,7 @@ namespace NLog.Targets
             {
                 eventLogInstance = new EventLog(Log, MachineName, renderedSource);
             }
-            if (MaxKilobytes.HasValue)
-            {
-                //you need more permissions to set, so don't set by default
-                eventLogInstance.MaximumKilobytes = MaxKilobytes.Value;
-            }
+
             return eventLogInstance;
         }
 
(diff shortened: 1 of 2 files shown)


Coriolanuss:
Hi again, @304NotModified, what are the expected actions to address the "codecov/patch" check?

304NotModified:
it warns us for the coverage on the new code, it's 17% instead of 50%. 

the code is never touched with an unit test, would be nice if we could do that:


image

Coriolanuss:
Hi, @304NotModified. Could you please check if the codecov/patch actually ran on 84319d6?

src/NLog/Targets/EventLogTarget.cs
```diff
@@ -98,6 +98,12 @@ public EventLogTarget(string name)
             Name = name;
         }
 
+        /// <summary>
+        /// Gets or sets the provider of Windows event log static methods to be utilized by <see cref="EventLogTarget"/>.
+        /// </summary>
+        /// <remarks>Introduced to improve the testability of the class.</remarks>
+        protected internal IEventLogStaticWrapper EventLogStaticMethods { get; set; } = EventLogStaticWrapper.Instance;
```
snakefoot: Think it would be nicer if you implemented an internal constructor that took an internal IEventLogWrapper-interface as input parameter. The default constructor would then call the new internal constructor with a private-class-implementation of the IEventLogWrapper-interface. Ofcourse the unit-tests will not be able to test initialization from NLog.config, but I think that is okay.

Coriolanuss: I get your point on injecting the implementation in the constructor instead of the property. But you mentioned an `IEventLogWrapper`, do you suggest merging the `IEventLogStaticWrapper` and `IEventLogInstanceWrapper`?

Also, you think these new interfaces and implementations should not fall under `protected internal`?

snakefoot: > do you suggest merging the IEventLogStaticWrapper and IEventLogInstanceWrapper?

Yes I would prefer just a single interface. And this interface and the constructor getting it as parameter are both marked as `internal`-only (Not protected).

Coriolanuss:
Hi, @snakefoot. I've addressed your comments in 5017749, but the build failed. Can you, or maybe @304NotModified, assist?

snakefoot:
Was thinking that all public constructors should call the unified internal constructor:

```c#
        private readonly IEventLogWrapper _eventLogWrapper;

        /// <summary>
        /// Initializes a new instance of the <see cref="EventLogTarget"/> class.
        /// </summary>
        public EventLogTarget()
            : this(null, null)
        {
        }

        /// <summary>
        /// Initializes a new instance of the <see cref="EventLogTarget"/> class.
        /// </summary>
        /// <param name="name">Name of the target.</param>
        public EventLogTarget(string name)
            : this(null, null)
        {
            Name = name;
        }

        /// <summary>
        /// Initializes a new instance of the <see cref="EventLogTarget"/> class.
        /// </summary>
        public EventLogTarget(IAppDomain appDomain)
            : this(null, appDomain)
        {
        }

        internal EventLogTarget(IEventLogWrapper eventLogWrapper, IAppDomain appDomain)
        {
            _eventLogWrapper = eventLogWrapper ?? new EventLogWrapper()
            appDomain = appDomain ?? LogFactory.CurrentAppDomain;
            Source = appDomain.FriendlyName;
            Log = "Application";
            MachineName = ".";
            MaxMessageLength = EventLogMaxMessageLength;
            OptimizeBufferReuse = GetType() == typeof(EventLogTarget);  // Class not sealed, reduce breaking changes
        }
```

snakefoot:
@Coriolanuss I know it is weird with the constructor `public EventLogTarget(IAppDomain appDomain)` but sadly enough it has become part of the public interface. Removing it would be a breaking change, and has to wait for NLog 5.0

We have to keep it until NLog ver. 5.0 where it can be removed. You are welcome to add the obsolete attribute. Ex:

```c#
        [Obsolete("This constructor will be removed in NLog 5. Marked obsolete on NLog 4.6")]
        public EventLogTarget(IAppDomain appDomain)
            : this(null, appDomain)
        {
        }
```

Coriolanuss:
Hi @snakefoot, I have brought back the constructor. Any other suggestions?

snakefoot:
No I think it looks great. Especially with the improved tests.

But I don't have the final verdict :)

Sent from my Sony Xperia


---- Vadim wrote ----


Hi @snakefoot<https://github.com/snakefoot>, I have brought back the constructor. Any other suggestions?

Coriolanuss:
It's nice to hear.
@304NotModified, do you have any comments?

304NotModified:
> No I think it looks great. Especially with the improved tests. But I don't have the final verdict :)

You approval says a lot anyway 👍

304NotModified:
> please move the new interfaces/classes (the interface to the "internal" folder and namespace) to new files (don't forget to copy the license header)

Maybe that isn't that easy with the multi projects now, so fine for now :)

304NotModified:
Merged, thanks!

PR #3022, opened 2018-11-19
LayoutRendererWrappers - Almost always support AppDomainFixedOutput

snakefoot:
When Inner-Layout allows it.

### DIFF
--- a/src/NLog/LayoutRenderers/MachineNameLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/MachineNameLayoutRenderer.cs
@@ -48,7 +48,7 @@ namespace NLog.LayoutRenderers
     [AppDomainFixedOutput]
     [ThreadAgnostic]
     [ThreadSafe]
-    public class MachineNameLayoutRenderer : LayoutRenderer, IStringValueRenderer
+    public class MachineNameLayoutRenderer : LayoutRenderer
     {
         internal string MachineName { get; private set; }
 
@@ -81,9 +81,6 @@ namespace NLog.LayoutRenderers
         {
             builder.Append(MachineName);
         }
-
-        /// <inheritdoc/>
-        string IStringValueRenderer.GetFormattedString(LogEventInfo logEvent) => MachineName;
     }
 }
 
--- a/src/NLog/LayoutRenderers/ThreadNameLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/ThreadNameLayoutRenderer.cs
@@ -59,6 +59,6 @@ namespace NLog.LayoutRenderers
 #endif
         }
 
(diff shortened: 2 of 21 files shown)


src/NLog/LayoutRenderers/MachineNameLayoutRenderer.cs
```diff
@@ -48,7 +48,7 @@ namespace NLog.LayoutRenderers
     [AppDomainFixedOutput]
     [ThreadAgnostic]
     [ThreadSafe]
-    public class MachineNameLayoutRenderer : LayoutRenderer, IStringValueRenderer
```
304NotModified: this is this removed? I don't understand in the context of this PR

src/NLog/LayoutRenderers/ThreadNameLayoutRenderer.cs
```diff
@@ -59,6 +59,6 @@ private static string GetStringValue()
 #endif
         }
 
-        string IStringValueRenderer.GetFormattedString(LogEventInfo _) => GetStringValue();
```
304NotModified: `_` is fine here?

src/NLog/LayoutRenderers/MachineNameLayoutRenderer.cs
snakefoot: There is no need for IStringValueRenderer when having the attribute `[AppDomainFixedOutput]`

src/NLog/LayoutRenderers/ThreadNameLayoutRenderer.cs
snakefoot: I know. But it is reported as warning by Sonar: 

https://sonarcloud.io/project/issues?branch=dev&id=nlog&open=AWb1jYMNW3Y69bz5TTB3&resolved=false

src/NLog/LayoutRenderers/MachineNameLayoutRenderer.cs
snakefoot: Have added unit-tests to verify compatibility between internal interfaces and attributes.

src/NLog/LayoutRenderers/ThreadNameLayoutRenderer.cs
304NotModified: mmm disagree on that. Unfortunately, I cannot set the rules in sonar for NLog now (need to move the repo to a new org)

src/NLog/LayoutRenderers/MachineNameLayoutRenderer.cs
304NotModified: great!

Issue #2925, opened 2018-09-20
FileLogger cannot create log file with NLog or UWP


JKGMS:
Hi,
we have set up a simple test Project where we try out NLog. It works fine with Android but fails to create Logfiles when doing so in an UWP or NUnit Test project. iOS is untested at this point.
Basic idea: .net Standard 2.0 Library provides a platform independent configuration and a wrapper because we want to use a single instance of the logger instead of calling LogManager.GetCurrentClassLogger() everywhere. The Android, UWP etc. projects just provide a filepath for log files and create a logger instance. (Code snippets and internal logs are attached to the end of this issue)

**NLog version**: 4.5.10

**Current NLog config**
logFilePath is depending on platform:

- Android: (public folder of the App, works fine)
 `Application.Context.GetExternalFilesDir(null).AbsolutePath `

- UWP: (somewhere in AppData, doesn't work)
` ApplicationData.Current.LocalFolder.Path `

- NUnitTest: (just in %temp%, doesn't work)
 `Path.Combine(Path.GetTempPath(), "NLog");`

````
fileTarget = new FileTarget("target2")
{
	FileName = Path.Combine(logFilePath,"log.txt"),
	CreateDirs = true,
	KeepFileOpen = true,
	AutoFlush = true,
	ArchiveEvery = FileArchivePeriod.Day,
	MaxArchiveFiles = 5,
	ArchiveNumbering = ArchiveNumberingMode.Date, 
	ArchiveDateFormat = "yyyy-MM-dd",
	Layout = @"${date:format=HH\:mm\:ss} [${level}] ${message} ${exception}"
};
config.AddTarget(fileTarget);
config.AddRule(LogLevel.Debug, LogLevel.Fatal, fileTarget);
````

- What is the current result?
No logfile is created, only console logging works. The internal Logfiles contain different exceptions (please see attached logs).
- What is the expected result?
Log file should be created and filled in the given Paths.
- Did you checked the [Internal log](https://github.com/NLog/NLog/wiki/Internal-Logging)?
Yes, please see attachments
[logInternal_NUnit.txt](https://github.com/NLog/NLog/files/2401029/logInternal_NUnit.txt)
[logInternal_UWP.txt](https://github.com/NLog/NLog/files/2401030/logInternal_UWP.txt)
- Please post full exception details (message, stacktrace, inner exceptions)
See Attachments, both contain a single run of our NUnit or UWP Test to create and write into the logfile 
- Are there any workarounds? No/Unknown
[...]

Issue #2824, opened 2018-07-31
Archive in UWP app throws exceptions


StrangerUA:
**NLog version**: (e.g. 4.5.7)
**Platform**: UWP
**Current NLog config**

```xml
<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd"
      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
      autoReload="true"
      internalLogLevel="Info">
  <targets>
    <target name="file"
            xsi:type="File"
            fileName="${var:LogPath}\*****-${shortdate}.log"
            layout="${date}|${logger}|${uppercase:${level}}|${message} ${exception}"
            archiveAboveSize="104857600"
            maxArchiveFiles="4"           
            />
  </targets>
  <rules>
    <logger name="*" minlevel="Info" writeTo="File" />
  </rules>
</nlog>

```

- What is the current result?
On first log to file app throws an 
System.UnauthorizedAccessException:

```
Access to the path 'Global\NLog-FileFileArchiveLock-c:_users_pavlo.holubovych_appdata_local_packages_*******_4sdaedqydyjdm_localstate_******-2018-07-31.log' is denied.'

   at System.IO.__Error.WinIOError(Int32 errorCode, String str)
   at System.Threading.Mutex.MutexTryCodeHelper.MutexTryCode(Object userData)
   at System.Runtime.CompilerServices.RuntimeHelpers.ExecuteCodeWithGuaranteedCleanup(TryCode code, CleanupCode backoutCode, Object userData)
   at System.Threading.Mutex.CreateMutexWithGuaranteedCleanup(Boolean initiallyOwned, String name, Boolean& createdNew, SECURITY_ATTRIBUTES secAttrs)
   at System.Threading.Mutex..ctor(Boolean initiallyOwned, String name, Boolean& createdNew, MutexSecurity mutexSecurity)
   at NLog.Internal.FileAppenders.BaseMutexFileAppender.CreateSharableMutex(String mutexNamePrefix)
   at NLog.Internal.FileAppenders.BaseMutexFileAppender.CreateArchiveMutex()

```

Log is working after I continue though. It throws an exception only once per file, SO the next one will be after first files reaches archiveAboveSize.

Problem is not reproduced without archive option, so the issue is somewhere there

Internal log file was not created

snakefoot:
Problems comes from UWP on NetCore2 doesn't support all features:

[...]

Issue #2742, opened 2018-05-25
ConcurrentWrites=true crashes on UWP without exception


mlischka:
**Type**: Bug
**NLog version**: 4.5.4.7700 (NLog for NetStandard 2.0)
**Platform**: UWP 6.1.4 on Windows 1803 (same behavior with managed and native)
**Current NLog config**: `FileTarget` with `KeepFileOpen=true`

- *What is the current result?*
While using a `FileTarget` with `KeepFileOpen=true` the app crashs immediately without an exception/stacktrace when the first log entry should be written. 

- *Did you checked the [Internal log](https://github.com/NLog/NLog/wiki/Internal-Logging)?*
Yes, the logger just shuts down after trying to create an appender.
```
...
2018-05-25 13:29:53.4018 Debug FileTarget(Name=): Process file 'C:\Users\Developer\AppData\Local\Packages\46eb4355-0396-4686-a706-046c3ba94973_4ca7txfpdd20p\LocalState\Log\log.txt' on startup
2018-05-25 13:29:53.4149 Debug FileTarget(Name=): Do archive file: 'C:\Users\Developer\AppData\Local\Packages\46eb4355-0396-4686-a706-046c3ba94973_4ca7txfpdd20p\LocalState\Log\log.txt'
2018-05-25 13:29:53.4451 Debug Creating file appender: C:\Users\Developer\AppData\Local\Packages\46eb4355-0396-4686-a706-046c3ba94973_4ca7txfpdd20p\LocalState\Log\log.txt
2018-05-25 13:29:53.5861 Info Shutting down logging...
2018-05-25 13:29:53.5957 Trace Flushing all targets...
...
```
- *Are there any workarounds?*
When setting `KeepFileOpen=false`, all works fine. Is this a bug or limitation when using NLog on UWP?

I traced this down the line and think the cause is that `MutexMultiProcessFileAppender` and `WindowsMultiProcessFileAppender` e.g. `BaseMutexFileAppender` creates a **global** mutex, but those are [not supported on UWP](https://msdn.microsoft.com/en-us/library/aa382954(v=vs.85).aspx) (*Note: The global namespace is not available for Windows Store apps.*) - [are they](https://github.com/dotnet/corefx/issues/20274#issuecomment-303874457)?

When I just execute `new Mutex(false, "Gloabl\anything")` on an UWP app, I get the same behavior, the app crashs.

Is there any way to use `KeepFileOpen=true` on UWP?

snakefoot:
[...]

PR #3048, opened 2018-12-20
FileTarget - ConcurrentWrites=false by default for better compatibility

snakefoot:
See also #2742 and #2824 and #2925 and #3059 (Xamarin mobile platforms also don't like global mutex) and #3893

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/3048)

### DIFF
--- a/src/NLog/Internal/FileAppenders/BaseFileAppender.cs
+++ b/src/NLog/Internal/FileAppenders/BaseFileAppender.cs
@@ -167,10 +167,10 @@ namespace NLog.Internal.FileAppenders
         /// <returns>A <see cref="FileStream"/> object which can be used to write to the file.</returns>
         protected FileStream CreateFileStream(bool allowFileSharedWriting, int overrideBufferSize = 0)
         {
-            int currentDelay = CreateFileParameters.ConcurrentWriteAttemptDelay;
+            int currentDelay = CreateFileParameters.FileOpenRetryDelay;
 
             InternalLogger.Trace("Opening {0} with allowFileSharedWriting={1}", FileName, allowFileSharedWriting);
-            for (int i = 0; i < CreateFileParameters.ConcurrentWriteAttempts; ++i)
+            for (int i = 0; i <= CreateFileParameters.FileOpenRetryCount; ++i)
             {
                 try
                 {
@@ -185,6 +185,7 @@ namespace NLog.Internal.FileAppenders
                         {
                             throw;
                         }
+
                         var directoryName = Path.GetDirectoryName(FileName);
                         try
                         {
@@ -195,13 +196,13 @@ namespace NLog.Internal.FileAppenders
                             //if creating a directory failed, don't retry for this message (e.g the ConcurrentWriteAttempts below)
                             throw new NLogRuntimeException("Could not create directory {0}", directoryName);
                         }
-                        return TryCreateFileStream(allowFileSharedWriting, overrideBufferSize);
 
+                        return TryCreateFileStream(allowFileSharedWriting, overrideBufferSize);
                     }
                 }
                 catch (IOException)
                 {
-                    if (!CreateFileParameters.ConcurrentWrites || i + 1 == CreateFileParameters.ConcurrentWriteAttempts)
+                    if (i + 1 >= CreateFileParameters.FileOpenRetryCount)
                     {
                         throw; // rethrow
                     }
--- a/src/NLog/Internal/FileAppenders/ICreateFileParameters.cs
+++ b/src/NLog/Internal/FileAppenders/ICreateFileParameters.cs
@@ -43,13 +43,13 @@ namespace NLog.Internal.FileAppenders
         /// <summary>
         /// Gets or sets the delay in milliseconds to wait before attempting to write to the file again.
         /// </summary>
-        int ConcurrentWriteAttemptDelay { get; }
+        int FileOpenRetryCount { get; }
 
         /// <summary>
         /// Gets or sets the number of times the write is appended on the file before NLog
         /// discards the log message.
         /// </summary>
-        int ConcurrentWriteAttempts { get; }
+        int FileOpenRetryDelay { get; }
 
         /// <summary>
         /// Gets or sets a value indicating whether concurrent writes to the log file by multiple processes on the same host.
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -163,7 +163,7 @@ namespace NLog.Targets
             ArchiveAboveSize = ArchiveAboveSizeDisabled;
             ArchiveOldFileOnStartupAboveSize = 0;
             ConcurrentWriteAttempts = 10;
-            ConcurrentWrites = true;
+            ConcurrentWrites = false;
             Encoding = Encoding.UTF8;
             BufferSize = 32768;
             AutoFlush = true;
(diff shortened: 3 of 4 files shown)


snakefoot:
> prefer testing the global mutex, before disabling it

I prefer not throwing exceptions as part of normal operation. The `ConcurrentWrites` is useful with IIS with multiple AppDomains. But on NetCore then there is no longer multiple AppDomains, so there will only be a single NLog instance.

Also the `ConcurrentWrites` introduces an overhead and complexity, that most NetCore-application do not need. Because `WindowsMultiProcessFileAppender` is not available on NetCore.

304NotModified:
> I prefer not throwing exceptions as part of normal operation.

Me neither, but there isn't a property to check. So I this case I think it's a better option than changing the default for all. 

Also note that frameworks like MVC also throw exceptions under the hood in normal operation. Last but not least, we are *preventing* a exception in this case (the mutex not available exception)

snakefoot:
> I think it's a better option than changing the default for all.

Well I think it is a better default to have `ConcurrentWrites = false` for all NetCore-platforms. Mostly because the time of the IIS and multiple AppDomains is over.

304NotModified:
> And having NetCore-platforms without support mutex, makes a good excuse to disable it by default (along with avoiding the extra complexity and overhead).

We could detect that (try-catch) and that isn't difficult? I don't like to move to default false, if there isn't another option. 

IMO smart defaults (detect it) are much better defaults than a black-white choice.

snakefoot:
The default is KeepFileOpen = false. No need to have ConcurrentWrites = true also.

Sent from my Sony Xperia


And having NetCore-platforms without support mutex, makes a good excuse to disable it by default (along with avoiding the extra complexity and overhead).

We could detect that (try-catch) and that isn't difficult? I don't like to move to default false, if there isn't another option.

IMO smart defaults (detect it) are much better defaults than a black-white choice.

snakefoot:
@304NotModified For NLog 5.0 as it is a minor breaking change.

snakefoot:
> @304NotModified prefer testing the global mutex, before disabling it

Well if you look at #2742 then the attempt of using the mutex takes down the application (UWP NetNative)

Still think that NetCore should have `ConcurrentWrites=false` by default, as #3079 will only help on certain NetCore platforms.

It will give the highest compatibility and the best performance on the NetCore-platform. IIS-users will not be affected as they will be using Net45.

304NotModified:
> IIS-users will not be affected as they will be using legacy Net45.


I'm not sure if that's true

snakefoot:
> @304NotModified I'm not sure if that's true

Please explain how changing the default for NetCore will affect legacy Net45 applications?

304NotModified:
I mean, IIS users aren't limited to non core .NET (I think)

snakefoot:
> @304NotModified I mean, IIS users aren't limited to non core .NET (I think)

You mean the new `In-process hosting model` ?

https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/iis/?view=aspnetcore-2.2

Looks like one have to wait for NetCore 3.0 to know the final level of isolation.

But until now it has been one process/appdomain for one website on NetCore. For legacy IIS then it is possible to spin up an extra AppDomain for the same website. This extra AppDomain for the same website required across-process file-synchronization.

304NotModified:
I think we should disable it for all platforms, or none.

What is the performance gain when ConcurrentWrites=false?

snakefoot:
From  #3490 then:

- KeepFileOpen=true, ConcurrentWrites=false = 5,387.220 ns
- KeepFileOpen=true, ConcurrentWrites=true  = 11,171.048 ns

But I guess this is with atomic-file-appender, and not with ForceMutex

304NotModified:
> I think we should disable it for all platforms, or none.

;)

304NotModified:
I think there are 3 audiences for NLog.

1. The ones that will just use NLog and it should be convenient and simple. Performance is important, but perfect is not needed.  
2. The ones that needs performance. They should have the controls (and good written advice) to get the most performance - a (small) feature loss is acceptable.
3. Benchmarks. Those are benchmarking unrealistic things. Nobody does `while(true) doLog`. 

I think it's more important to to focus on 1 and 2.  If we focus on 3, that could hurt 1 and 2. That we will lose on 3 is really pity (I've seen comparisons before) - but that is an explicit choice.

snakefoot:
> I think it's more important to compare when KeepFileOpen=false (the default).

When using `KeepFileOpen=false` then it doesn't change performance with `ConcurrentWrites=false`. The only difference is that it retries file-open-operation on failure (because of conflict with other application). Actually think that `KeepFileOpen=false` should ignore `ConcurrentWrites` and always perform retry on failure.

Should rename these parameters:
- concurrentWriteAttemptDelay -> FileOpenRetryDelay
- concurrentWriteAttempts -> FileOpenRetryCount

And change this line:

https://github.com/NLog/NLog/blob/899c3f51a681303855c21d03a17aa19cc5412e91/src/NLog/Internal/FileAppenders/BaseFileAppender.cs#L204

Into:
```C#
if ((!CreateFileParameters.ConcurrentWrites && !CreateFileParameters.KeepFileOpen) || i + 1 == CreateFileParameters.FileOpenRetryCount)
```

snakefoot:
My priority is this:
- Compatibility with most platforms
- Performance
So I don't mind that KeepFileOpen=false by default. But I do mind that ConcurrentWrites=true is enabled by default as it hurts performance and give compatibility issues with KeepFileOpen=true.

304NotModified:
so could we set ConcurrentWrites to false when unset and KeepFileOpen=true?

(sorry if this is proposed before)

snakefoot:
> so could we set ConcurrentWrites to false when unset and KeepFileOpen=true?

Same as having ConcurrentWrites=false by default. And perform the above rename of parameters (FileOpenRetryCount + FileOpenRetryDelay) so they apply for both ConcurrentWrites=true and KeepFileOpen=false.

snakefoot:
For NLog 5.0 then it would be nice if the default FileTarget didn't know about ConcurrentWrites (No Mutex-writes, No Atomic-writes. No FileWatchers. No Compression). Just being plain and simple without crazy logic.

Then one can make a ConcurrentFileTarget that inherits from FileTarget that adds Mutex + Atomic + FileWatchers

Then one can make a CompressArchiveFileWrapper that hooks into the wrapped FileTarget and reacts to archive-events.

snakefoot:
@304NotModified Have now added a commit that changes ConcurrentWrites=false for all platforms.

But since breaking change, then it will be for NLog 5.0. And since NLog 5.0 should breakup FileTarget so there is `FileTarget` + `ConcurrentFileTarget` then this PR might not be relevant?

snakefoot:
Created #3492 for NLog 4.6.6, while waiting for `ConcurrentFileTarget` to be ready for NLog 5.0

304NotModified:
> And since NLog 5.0 should breakup FileTarget so there is `FileTarget` + `ConcurrentFileTarget` then this PR might not be relevant?

Is the ConcurrentFileTarget , the slow  or the fast one?

snakefoot:
ConcurrentFileTarget uses mutex or atomic file-append together with file watchers.


But since NLog 5 has changed status to minimum breaking change, then I think such a split is for NLog 6.


Sent from my Sony Xperia


And since NLog 5.0 should breakup FileTarget so there is FileTarget + ConcurrentFileTarget then this PR might not be relevant?

Is the ConcurrentFileTarget , the slow or the fast one?

snakefoot:
Resolved merge conflicts for merging into DEV-branch.

snakefoot:
Any reason why this cannot be part of NLog 5.0 ?

304NotModified:
Sorry, I'm a bit lost in the discussion.

For me:

- I prefer the same defaults for .NET Core and .NET Framework (but it isn't a must)
- We should not have the defaults for the "lowest" platform
- I'm afraid of this change for users that will upgrade.
- Drop silverlight and Xamarin targets for NLog 5 is fine

snakefoot:
I have no trouble in making `ConcurrentWrites=false` for both NetCore + NetFramework

snakefoot:
> We should not have the defaults for the "lowest" platform

Then you should also have `KeepFileOpen=true` as default :). I prefer compatible over performance.

And if you notice the code in this PR, then I make given `KeepFileOpen=false` same behavior as if `ConcurrentWrites=true` (Performs file-open retries).

snakefoot:
Think this belongs to NLog 5.0, since the goal is to make NLog to be more slim and work on more platforms out-of-the-box.

snakefoot:

Issue #3049, opened 2018-12-21
Add IncludeGdc in JsonLayout


casperc89:
**NLog version**: 4.5

**Platform**: .Net 3.5 / .Net 4.0 / .Net 4.5 / Mono 4 / Silverlight 4 / Silverlight 5 / Xamarin Android / Xamarin iOs / .NET Core 1 / .NET Core 2

**Why do we need it?**
JsonLayout currently has support to log all properties from `MappedDiagnosticsContext` & `MappedDiagnosticsLogicalContext` through  `IncludeMdc` & `IncludeMdlc`. Unfortunately the feature to include properties from `GlobalDiagnosticsContext` is missing. 

Since we're using these properties in our application through a generic template. It would be awesome to have this feature for the `GlobalDiagnosticsContext` as well.

**An example of the XML config, if relevant.**
```xml
<layout type=""JsonLayout"" IncludeMdc='true' IncludeMdlc='true' IncludeGdc='true' >...</layout>
```

304NotModified:
good idea! PR welcome :)

please also check other targets/layout renderers with `IncludeMdc ` etc

304NotModified:
thanks! Fixed in #3050 and will be released n 4.6

PR #3050, opened 2018-12-21
Added IncludeGdc property in JsonLayout

casperc89:
Fix for #3049


304NotModified:
FYI, NLog 4.6 is online which includes this feature!

PR #3053, opened 2018-12-25
Added ${appsetting} - LayoutRenderer for easier access to app.config settings

### LINKED ISSUE #2335: Merge NLog.Extended appsettings to NLog
deprecate NLog.Extended and move to NLog

Move MSMQ to separate package

snakefoot:
After seeing that NetFramework will now enter maintenance mode, and NetCore will take over.

Then it doesn't make sense trying extract app.config-logic from NLog-core, as it will only be available for NetFramework (And not NetCore).

Also it is a little difficult to perform auto extension loading to ensure that nlog-config is automatically loaded from app.config.

This replaces #2919 and resolves #2335 

Other change included:
- Removed the optional baseDirectory-parameter from `CreateFromXmlString` as it doesn't make sense when it comes to autoreload config-file or use of include-files when giving of in-memory-config.
- Modified `RegisterExtendedItems` so it doesn't register non-existing class-types, and also prioritize the now builtin NLog-appsetting, instead of NLog.Extended-appsetting.

### DIFF
--- a/src/NLog/Config/ConfigurationItemFactory.cs
+++ b/src/NLog/Config/ConfigurationItemFactory.cs
@@ -560,19 +560,7 @@ namespace NLog.Config
 
                 // register types
                 string targetsNamespace = typeof(DebugTarget).Namespace;
-                _targets.RegisterNamedType("AspNetTrace", targetsNamespace + ".AspNetTraceTarget" + suffix);
                 _targets.RegisterNamedType("MSMQ", targetsNamespace + ".MessageQueueTarget" + suffix);
-                _targets.RegisterNamedType("AspNetBufferingWrapper", targetsNamespace + ".Wrappers.AspNetBufferingTargetWrapper" + suffix);
-
-                // register layout renderers
-                string layoutRenderersNamespace = typeof(MessageLayoutRenderer).Namespace;
-                _layoutRenderers.RegisterNamedType("appsetting", layoutRenderersNamespace + ".AppSettingLayoutRenderer" + suffix);
-                _layoutRenderers.RegisterNamedType("aspnet-application", layoutRenderersNamespace + ".AspNetApplicationValueLayoutRenderer" + suffix);
-                _layoutRenderers.RegisterNamedType("aspnet-request", layoutRenderersNamespace + ".AspNetRequestValueLayoutRenderer" + suffix);
-                _layoutRenderers.RegisterNamedType("aspnet-sessionid", layoutRenderersNamespace + ".AspNetSessionIDLayoutRenderer" + suffix);
-                _layoutRenderers.RegisterNamedType("aspnet-session", layoutRenderersNamespace + ".AspNetSessionValueLayoutRenderer" + suffix);
-                _layoutRenderers.RegisterNamedType("aspnet-user-authtype", layoutRenderersNamespace + ".AspNetUserAuthTypeLayoutRenderer" + suffix);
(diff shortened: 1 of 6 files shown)


src/NLog/Internal/IConfigurationManager.cs
```diff
@@ -31,7 +31,7 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
-#if !SILVERLIGHT && !NETSTANDARD
+#if !SILVERLIGHT && !__IOS__ && !__ANDROID__ && !NETSTANDARD
```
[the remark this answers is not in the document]
snakefoot: Pretty sure no one outside the NLog-codebase is using this interface. And the number of IOS and Android-users is very low. Just wanted to make things consistent, but I can introduce the interface if you like?

snakefoot:
> I'm doubting about that. While it's not likely, with this change we can't influence the relative paths?

The problem is that the filename-parameter is not just baseDirectory, but file-path to perform filewatch, and filepath to attempt to reload configuration from. Something that will not work well when providing nlog-config as xml-string.

src/NLog/Internal/IConfigurationManager.cs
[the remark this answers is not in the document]
304NotModified: > Pretty sure no one outside the NLog-codebase is using this interface.

I hope so, But there is no big reason to do this now (and not in NLog 5), so please keep it and mark it obsolete or something like that.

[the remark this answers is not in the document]
snakefoot: Have reverted the cleanup, so it can smell forever :)

src/NLog/LayoutRenderers/AppConfigLayoutRenderer.cs
```diff
@@ -0,0 +1,90 @@
...
+    /// </summary>
+    /// <remarks>
+    /// Use this layout renderer to insert the value of an application setting
+    /// stored in the application's App.config or Web.config file.
+    /// </remarks>
+    /// <code lang="NLog Layout Renderer">
+    /// ${appsetting:name=mysetting:default=mydefault} - produces "mydefault" if no appsetting
+    /// </code>
+    [LayoutRenderer("appsetting")]
+    public sealed class AppConfigLayoutRenderer : LayoutRenderer
```
304NotModified: I think it would be nice if those names are in-sync ;)

304NotModified: Do you mean a name clash in the layout renderer name or the class name?

snakefoot: Well I changed the class-name to avoid name-clash in class-name :)

304NotModified: what about a new namespace? And then in NLog 5 we could correct that.

snakefoot: Will move it into Internal-then, and then it can move out with NLog 5.0

304NotModified: Or AppSettingLayoutRenderer2 ;)

snakefoot: Done

304NotModified:
Thanks for the changes!

304NotModified:
@snakefoot 

guess we only need to document that the external package isn't needed anymore, correct?

snakefoot:
@304NotModified Updated wiki: https://github.com/NLog/NLog/wiki/AppSetting-Layout-Renderer (Created #3159)

PR #3061, opened 2018-12-29
TempDirLayoutRenderer - Avoid exception on static class initialization

snakefoot:


304NotModified:
Thanks! Exception in static handlers do result nasty error messages, so good to have less 👍  See also https://stackoverflow.com/q/34994470/201303

Issue #3059, opened 2018-12-28
Failed to create file appender on UWP


sakatuba:
Hi!

I found a little bug on FileTarget.

**NLog version**: 4.5.11

**Platform**: UniversalWindows (1803/17134, Windows10 64bit)

**Current NLog config**: I did it by programable way just below.

```C#
            FileTarget fileTarget = new FileTarget("fileTarget")
            {
                Encoding = Encoding.UTF8,
                Layout = "${date:format=yyyy-MM-dd HH\\:mm\\:ss.fff}\t${level}\t${logger}\t${message}\t${exception:format=toString,Data:maxInnerExceptionLevel=10}",
                FileName = @"${var:LogPath}\XXX\log-${date:format=yyyy-MM-dd}.log",
                WriteBom = false,
                LineEnding = LineEndingMode.CRLF,
                KeepFileOpen = true,
                ConcurrentWrites = true,
                OptimizeBufferReuse = true,
                CreateDirs = true
            };
            config.AddTarget(fileTarget);
```

- What is the current result?

No Log is logged. After making InnerLogger on, I found exceptions.

- What is the expected result?

Make it run and log.

- Did you checked the [Internal log](https://github.com/NLog/NLog/wiki/Internal-Logging)?

Yes, I did.

- Please post full exception details (message, stacktrace, inner exceptions)

```
NLog: 2018-12-28 10:18:40.1169 Warn Failed to create file appender: C:\Users\sakamoto\AppData\Local\Packages\~~~~~~~\LocalState\log-2018-12-28.log Exception: System.UnauthorizedAccessException: Access to the path 'Global\NLog-FileFileLock-c:_users_sakamoto_appdata_local_packages_~~~~~_localstate_log-2018-12-28.log' is denied.
   at System.IO.__Error.WinIOError(Int32 errorCode, String str)
   at System.Threading.Mutex.MutexTryCodeHelper.MutexTryCode(Object userData)
   at System.Runtime.CompilerServices.RuntimeHelpers.ExecuteCodeWithGuaranteedCleanup(TryCode code, CleanupCode backoutCode, Object userData)
   at System.Threading.Mutex.CreateMutexWithGuaranteedCleanup(Boolean initiallyOwned, String name, Boolean& createdNew, SECURITY_ATTRIBUTES secAttrs)
   at System.Threading.Mutex..ctor(Boolean initiallyOwned, String name, Boolean& createdNew, MutexSecurity mutexSecurity)
[...]

PR #3063, opened 2018-12-30
File target: Handle UnauthorizedAccessException for multi-process mutexes (e.g. Xamarin and UPW)

304NotModified:
alternative solution for #3048 

See #2742 and #2824 and #2925 and #3059


snakefoot:
- Bad idea to throw exception everytime creating an file-appender.
- Bad idea to fallback to normal Mutex, since it provides no protection. Instead it should fallback to KeepFileOpen=false.

Still think ConcurrentWrites = false should be the default on NetCore, since the default of KeepFileOpen = false works everywhere.

PR #3064, opened 2018-12-30
Improve EventLogTarget installation logic. Make public const values internal

snakefoot:
Small cleanup after #3017


304NotModified:
Thanks, good one!

Issue #3024, opened 2018-11-21
Machine name greater than 15 characters truncated


amitsaha:
Hi, looking at how the current `machineName` is derived [code](https://github.com/NLog/NLog/blob/3a2f1214700464e969e92384ea80dcaffec49b4e/src/NLog/Internal/EnvironmentHelper.cs#L64), I see it reads the `ComputerName` environment variable which as per this [article](https://blogs.msdn.microsoft.com/brada/2005/08/11/environment-machinename-truncates-after-the-first-15-characters/) is by default limited to 15 characters. This is a problem for AWS instance IDs for example or any custom hostname that exceeds 15 characters.

**NLog version**: (e.g. 4.4.13)

`master`

**Platform**: 

.NET Framework/Windows

**Current NLog config** (xml or C#, if relevant)

NA

- What is the current result?

Machine name logged as: `I-03FB71646161E` .

- What is the expected result?

I was expecting the machine name should be the same as hostname (in my case, AWS instance ID
`i-03fb71646161e8626`)

- Did you checked the [Internal log](https://github.com/NLog/NLog/wiki/Internal-Logging)?

Yes

- Please post full exception details (message, stacktrace, inner exceptions)

NA

- Are there any workarounds? yes/no

Not sure

- Is there a version in which it did work? 

No.

- Can you help us by writing an unit test?

Perhaps, if guided.

amitsaha:
Thanks, @snakefoot - will give it a shot. If I get some guidance, I could try creating a PR as well.

304NotModified:
> Thanks, @snakefoot - will give it a shot. If I get some guidance, I could try creating a PR as well.

See https://github.com/NLog/NLog/wiki/How-to-write-a-custom-layout-renderer

304NotModified:
Need any help on this @amitsaha ?

amitsaha:
hi @304NotModified - Plan to take a look at this in the next week.

304NotModified:
👍

amitsaha:
@snakefoot @304NotModified done. please take a look. I went ahead and implemented a new layout renderer, since it probably makes more sense IMO.

304NotModified:
Fixed by #3071. Thanks!

PR #3071, opened 2019-01-09
Added ${HostName} Layout Renderer

amitsaha:
This implements a new `hostname` layout renderer. This is slightly tangential to the original issue (#3024), but I think this makes more sense than "fixing" `machinename` since it's a Windows behavior that the machine name property is truncated to 15 characters.

PS: I wasn't sure what to do about the copyright header. so have left them as it is.


snakefoot:
What are the outputs of the 3 methods? 
All correctly contains more than 15 chars?
Do they contains domain-name? If so should it be stripped?

amitsaha:
@snakefoot i will have to verify on my real system. But considering i have seen HOSTNAME being set correctly on a Windows system, I don’t see any issue there.

For the question regarding the domain mame, I am not sure. Do we need to?

304NotModified:
> What are the outputs of the 3 lookup-methods for your machine?
> All correctly contains more than 15 chars?
> Do they contains domain-name? If so should it be stripped?

Good questions!

amitsaha:
@snakefoot some tests

Program.cs:

```
using System;

namespace test
{
    class Program
    {
        {
            Console.WriteLine("HOSTNAME env variable: " + Environment.GetEnvironmentVariable("HOSTNAME"));
            Console.WriteLine("DNS Get Hostname: " + System.Net.Dns.GetHostName());
            Console.WriteLine("COMPUTERNAME env variable: " + Environment.GetEnvironmentVariable("COMPUTERNAME"));
        }
    }
}
```

`Dockerfile`:

```
FROM microsoft/dotnet:2.2-sdk

RUN mkdir C:\app
ADD . C:/app/

WORKDIR C:/app
CMD ["dotnet", "run"]
```

Build the image and run:

```
C:\> docker run -e HOSTNAME="averylong-hostname.com" -h "averylong-hostname.com"  amitsaha/test

HOSTNAME env variable: averylong-hostname.com
DNS Get Hostname: averylong-hostname.com
COMPUTERNAME env variable: AVERYLONG-HOSTN
```

So, as expected both HOSTNAME (via env variable) and DNS get hostname call returns the complete hostname where as the COMPUTERNAME is limited to 15 characters.

I am not sure how you feel about the domain name being present in the hostname. But, I think it's fine to leave it as it is?

snakefoot:
Then I think you should replace the COMPUTERNAME fallback-lookup with `NLog.Internal.EnvironmentHelper.GetMachineName()`.

amitsaha:
@snakefoot done.

304NotModified:
Thanks!

amitsaha:
Thanks @304NotModified - i saw that the [wiki](https://github.com/NLog/NLog/wiki/HostName-Layout-Renderer) is updated as well.

304NotModified:
please also add it here :) (https://github.com/NLog/NLog.github.io/blob/master/config/layout-renderers.json) - for https://nlog-project.org/config/?tab=targets&search=hostname 

thanks!

amitsaha:
Sure, done, https://github.com/NLog/NLog.github.io/pull/89

304NotModified:
Thanks!

304NotModified:
FYI, NLog 4.6 RC is online including this change!

PR #3076, opened 2019-01-12
Fix crash on empty Continuation

304NotModified:


snakefoot:
Please don't merge. It is faulty usage of NLog, and one should get the exception

snakefoot:
Created #3077 for you instead

PR #3079, opened 2019-01-13
File target: Handle UnauthorizedAccessException for multi-process mutexes (e.g. Xamarin and UWP)

304NotModified:
detect creating Mutex also runtime.

alternative solution for #3048 

See #2742 and #2824 and #2925 and #3059

supersedes #3063

### DIFF
--- a/src/NLog/Internal/FileAppenders/BaseMutexFileAppender.cs
+++ b/src/NLog/Internal/FileAppenders/BaseMutexFileAppender.cs
@@ -136,7 +136,7 @@ namespace NLog.Internal.FileAppenders
         protected Mutex CreateSharableMutex(string mutexNamePrefix)
         {
             if (!PlatformDetector.SupportsSharableMutex)
-                throw new Exception("Creating Mutex not supported");
+                throw new NotSupportedException("Creating Mutex not supported");
 
             var name = GetMutexName(mutexNamePrefix);
 
--- a/src/NLog/Internal/FileAppenders/BaseMutexFileAppender.cs
+++ b/src/NLog/Internal/FileAppenders/BaseMutexFileAppender.cs
@@ -131,6 +131,11 @@ namespace NLog.Internal.FileAppenders
 
             var name = GetMutexName(mutexNamePrefix);
 
+            return ForceCreateSharableMutex(name);
+        }
+
+        internal static Mutex ForceCreateSharableMutex(string name)
+        {
 #if !NETSTANDARD
             // Creates a mutex sharable by more than one process
             var mutexSecurity = new MutexSecurity();
--- a/src/NLog/Internal/FileAppenders/BaseMutexFileAppender.cs
+++ b/src/NLog/Internal/FileAppenders/BaseMutexFileAppender.cs
@@ -43,6 +43,7 @@ namespace NLog.Internal.FileAppenders
     using System.Security;
     using System.Threading;
     using System.Text;
+    using JetBrains.Annotations;
 
 #if SupportsMutex
 #if !NETSTANDARD
@@ -50,8 +51,8 @@ namespace NLog.Internal.FileAppenders
     using System.Security.Principal;
 #endif
     using System.Security.Cryptography;
-    using Common;
 #endif
+    using Common;
 
     /// <summary>
     /// Base class for optimized file appenders which require the usage of a mutex. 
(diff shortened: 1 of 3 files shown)


304NotModified:
@snakefoot your input please :)

304NotModified:
> So when people configure KeepFileOpen on limited NetCore-platforms then they will continue to have KeepFileOpen=false. Think this is very confusing, and still prefer to have ConcurrentWrites = false by default on NetCore-platforms, instead of all this guess-magic.

I tried to *not* change the current behavior (the fallback when there is no mutex), but only change the detection, which wasn't good enough.

If the fallback is not good, IMO that's another issue and maybe more for NLog 5.

> Still don't like the testing with exceptions. Also you need to catch a lot more. See also
> 
> `if (ex is SecurityException || ex is UnauthorizedAccessException || ex is NotSupportedException || ex is NotImplementedException || ex is PlatformNotSupportedException)`

will fix that

snakefoot:
Please make `CreateSharableMutex()` return null instead of creating a dummy Mutex-object.

And places where null is not acceptable (Ex. `MutexMultiProcessFileAppender`) then throw exception.

304NotModified:
> Please make `CreateSharableMutex()` return null instead of creating a dummy Mutex-object.

that's existing code. Just moved around.

snakefoot:
Yes but now you have improve the detection, and the fallback is invalid, and now the fallback is exercised even more.

304NotModified:
Fair enough, will check that

snakefoot:
Btw. Title of PR should be UWP (and not UPW)

304NotModified:
WUPS!

304NotModified:
@snakefoot please review, thanks!

src/NLog/Internal/FileAppenders/BaseMutexFileAppender.cs
```diff
@@ -127,10 +136,15 @@ protected override void Dispose(bool disposing)
         protected Mutex CreateSharableMutex(string mutexNamePrefix)
         {
             if (!PlatformDetector.SupportsSharableMutex)
-                return new Mutex();
+                throw new Exception("Creating Mutex not supported");
```
snakefoot: Maybe a better Exception-Type. Ex. `System.NotSupportedException`

src/NLog/Internal/PlatformDetector.cs
```diff
@@ -76,19 +85,60 @@ public static bool SupportsSharableMutex
...
+                catch (Exception ex) 
+                {
+                    InternalLogger.Debug(ex, "Failed to create sharable mutex processes");
+                    runTimeSupportsSharableMutex = false;
+                }
+
+                return runTimeSupportsSharableMutex.Value;
+                }
+
+            }
```
snakefoot: Please re-format the cs-file, so curly-braces are indented properly.

src/NLog/Targets/FileTarget.cs
```diff
@@ -449,6 +449,12 @@ public bool ConcurrentWrites
             get
             {
 #if SupportsMutex
+
+                if (!PlatformDetector.SupportsSharableMutex)
+                {
+                    return _concurrentWrites ?? false;  // Better user experience for mobile platforms
```
snakefoot: Maybe change into `return _concurrentWrites ?? PlatformDetector.SupportsSharableMutex` ?

src/NLog/Targets/FileTarget.cs
```diff
@@ -1747,10 +1753,30 @@ private bool TryArchiveFile(string fileName, LogEventInfo ev, int upcomingWriteS
...
+                            {
+                                mutexFileAppender.ArchiveMutex.WaitOne();
+                            }
+                        }
                         else if (!KeepFileOpen || ConcurrentWrites)
+                        {
+                            mutexSupportedForArchive = false;
+                        }
+
+                        if (!mutexSupportedForArchive)
```
snakefoot: Maybe just change into:

```
if (archivedAppender is BaseMutexFileAppender mutexFileAppender && mutexFileAppender.ArchiveMutex != null)
{
     mutexFileAppender.ArchiveMutex.WaitOne();
}
else
{
     InternalLogger.Info("FileTarget(Name={0}): Archive mutex not available: {1}", Name, archiveFile);
}
```

304NotModified:
@snakefoot thanks for your suggestions. Those are applied now. Please let me know if this is ready to merge :)

snakefoot:
Looks better now. Still prefer just having global-mutex disabled by default for NetCore. Better performance and better compatibility.

304NotModified:
> Looks good now. Still prefer just having global-mutex disabled by default for NetCore. Better performance and better compatibility. Maybe for NLog 5.0 ?

I think that's OK, but we have to double check the pro and cons for it

PR #3082, opened 2019-01-14
WebService target: added support for custom SOAPAction property (Soap11+Soap12 protocol)

### LINKED ISSUE #3081: WebService target with Soap11 protocol does not work with custom action names.
**NLog version**: (e.g. 4.5.11)

**Platform**: .Net 3.5 / .Net 4.0 / .Net 4.5 / Mono 4 / Silverlight 4 / Silverlight 5 / Xamarin Android / Xamarin iOs / .NET Core 1 / .NET Core 2

```xml
<nlog>
  <targets>
    <target type="WebService"
            name="loggingService"
            url="http://localhost/logging/loggingservice.svc"
            encoding="UTF-8"
            protocol="Soap11"
            namespace="http://tempuri.org/"
            methodName ="Log"
            preAuthenticate="false" >
      <parameter name="level" layout="1" type="System.Int32"/>
      <parameter name="loggerType" layout="${logger}" type="System.String"/>
      <parameter name="message" layout="${message}" type="System.String"/>
    </target>
  </targets>
  <rules>
    <logger name="*" writeTo="loggingService" minlevel="debug" />
  </rules>
</nlog>

```
The actual action's name is ```http://tempuri.org/ILoggingService/Log```. The target makes action name ```http://tempuri.org/Log``` (namespace + methodName)

- What is the current result? Web service's method is not being invoked.
- What is the expected result? Web service's method is being invoked.
- Did you checked the [Internal log](https://github.com/NLog/NLog/wiki/Internal-Logging)? Yes.
```
2019-01-14 22:27:55.9415 Error WebServiceTarget(Name=loggingService): Error sending request Exception: System.Net.WebException: The remote server returned an error: (500) Internal Server Error.
   at System.Net.HttpWebRequest.EndGetResponse(IAsyncResult asyncResult)
   at NLog.Targets.WebServiceTarget.<>c__DisplayClass67_0.<CreateSendContinuation>b__1(IAsyncResult r)
```
- Is there a version in which it did work? No.
- Can you help us by writing an unit test? Yes.

AlexeyRokhin:
Fixes #3081 
Processing of existing header SoapAction has been added.
Unit tests for default SoapAction and custom SoapAction have been added.

### DIFF
--- a/src/NLog/Targets/WebServiceTarget.cs
+++ b/src/NLog/Targets/WebServiceTarget.cs
@@ -661,19 +661,18 @@ namespace NLog.Targets
 
             protected override void InitRequest(HttpWebRequest request)
             {
+                const string soapActionHeader = "SOAPAction";
                 base.InitRequest(request);
 
-                string soapAction;
-                if (Target.Namespace.EndsWith("/", StringComparison.Ordinal))
+                string soapAction = request.Headers[soapActionHeader];
+                if (string.IsNullOrEmpty(soapAction))
                 {
-                    soapAction = string.Concat(Target.Namespace, Target.MethodName);
-                }
-                else
-                {
-                    soapAction = string.Concat(Target.Namespace, "/", Target.MethodName);
+                    soapAction = Target.Namespace.EndsWith("/", StringComparison.Ordinal)
+                        ? string.Concat(Target.Namespace, Target.MethodName)
+                        : string.Concat(Target.Namespace, "/", Target.MethodName);
                 }
 
-                request.Headers["SOAPAction"] = soapAction;
+                request.Headers[soapActionHeader] = soapAction;
             }
         }
 
--- a/src/NLog/Targets/WebServiceTarget.cs
+++ b/src/NLog/Targets/WebServiceTarget.cs
@@ -670,9 +670,8 @@ namespace NLog.Targets
                     soapAction = Target.Namespace.EndsWith("/", StringComparison.Ordinal)
                         ? string.Concat(Target.Namespace, Target.MethodName)
                         : string.Concat(Target.Namespace, "/", Target.MethodName);
+                    request.Headers[soapActionHeader] = soapAction;
                 }
-
-                request.Headers[soapActionHeader] = soapAction;
             }
         }
 
--- a/src/NLog/Targets/WebServiceTarget.cs
+++ b/src/NLog/Targets/WebServiceTarget.cs
@@ -681,6 +681,14 @@ namespace NLog.Targets
             protected override string SoapEnvelopeNamespace => Soap12EnvelopeNamespaceUri;
 
             protected override string SoapName => "soap12";
+
+            protected override string ContentType => "application/soap+xml";
+
+            protected override void InitRequest(HttpWebRequest request)
+            {
+                base.InitRequest(request);
+                request.ContentType += $@"; action=""{GetSoapAction()}""";
+            }
         }
 
         private abstract class HttpPostSoapFormatterBase : HttpPostXmlFormatterBase
--- a/src/NLog/Targets/WebServiceTarget.cs
+++ b/src/NLog/Targets/WebServiceTarget.cs
@@ -137,6 +137,12 @@ namespace NLog.Targets
         public string Namespace { get; set; }
 
         /// <summary>
+        /// Gets or sets the Web service soap action name. Only used with Soap.
+        /// </summary>
+        /// <docgen category='Web Service Options' order='10' />
+        public string SoapAction { get; set; }
+
+        /// <summary>
         /// Gets or sets the protocol to be used when calling web service.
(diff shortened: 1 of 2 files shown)


304NotModified:
Thanks!


Example usage:

```xml
<target type='WebService> 
    ....
      <header name='SOAPAction' layout='http://tempuri.org/custom-namespace/Ping'/>
    ....
</target>
```

304NotModified:
> The diff coverage is `100%`.

👍  great! thanks!

304NotModified:
What about the Soap20?

AlexeyRokhin:
Regarding Soap 1.2: it will require a changing of ```content-type``` from ```text/xml``` to ```application/soap+xml``` (my soap12 binding does not work with ```text/xml```), adding ```action``` parameter to ```content-type``` and, probably, creating addition property ```SoapAction``` to use it in both ```Soap11``` and ```Soap12``` protocols.
Is this OK?

snakefoot:
I like the (ab)use of `<header>` better than adding SoapAction-property:

```
string soapAction = Target.Headers?.Count > 0 ? request.Headers[soapActionHeader] : string.Empty;
if (string.IsNullOrEmpty(soapAction))
{
   // Blah blah
}
```

AlexeyRokhin:
Yes, I initially I implemented ```soapaction``` as header but Soap1.2 does not have such header and I re-implemented it as a property with the thought of possible Soap 1.2 implementation.
What stops me from pushing soap1.2 implementation to GitHub is that it is kind of breaking change (because of changing ```content-type``'s value). However, I'm not sure whether current Soap1.2 implementation works at all for anyone.

snakefoot:
@AlexeyRokhin Think you should just let the Soap1.2 depend on `Target.Namespace` and `Target.MethodName`. But always nice with a test of Soap1.2 if it is easy to make.

AlexeyRokhin:
@snakefoot 
Do you mean that soap1.2 should not have ability of action name customization? Just ```Namespace```+```MethodName```? Such implementation will have the same problem as Soap1.1 had.

snakefoot:
@AlexeyRokhin I would just use the same `<header>`-trick as with soap11. 

But it seems you are right about the missing ContentType, and I'm pretty sure that it have never worked.

See also: https://www.ryadel.com/en/soap-web-service-request-asp-net-c-sharp-wsdl-proxy-soapclient/

AlexeyRokhin:
OK. I can add Soap1.2 implementation.
However, I have another idea: add Layout property to allow user set any request shape if it is required:
`
`
If it is set, we will use it. If it is not, we will use the 'standard' behavior.
Or we can create a new Target like ```HttpTarget``` with Layout for payload and properties for url, method, headers, etc.

snakefoot:
> If it is set, we will use it. If it is not, we will use the 'standard' behavior.

WebServiceTarget is already now a big ball of spaghetti. Having more secret settings not directly related to configuring the HttpRequest, and only used in few special cases is not going to make it better. 

But maybe I don't understand your meta-setting `Layout Layout`

AlexeyRokhin:
I mean that Layout property may be used in the following way:
```
<Layout><![CDATA[
<soap:Envelope xmlns:soap="http://www.w3.org/2003/05/soap-envelope" xmlns:tem="http://tempuri.org/">
  <soap:Header/>
  <soap:Body>
    <tem:Log>
      <tem:level>${level}</tem:level>
      <tem:loggerType>${logger}</tem:loggerType>
      <tem:message>${message}</tem:message>
    </tem:Log>
  </soap:Body>
</soap:Envelope>]]>
</Layout>
```

304NotModified:
Is this one ready for review?

AlexeyRokhin:
No new changes from me are planned.

304NotModified:
@snakefoot do you think this is ready for merge?

snakefoot:
I actually prefer the initial suggestion of providing a custom soap action using a custom Header. Instead of adding a new soapaction-property.


But I like the unit tests. So maybe just merge, and then I can create a followup PR that we can discuss.


Sent from my Sony Xperia


@snakefoot<https://github.com/snakefoot> do you think this is ready for merge?

304NotModified:
> I actually prefer the initial suggestion of providing a custom soap action using a custom Header. Instead of adding a new soapaction-property.

I think both are OK, slightly prefering the soapaction property

snakefoot:
> I think both are OK, slightly prefering the soapaction property

Think it is a little confusing to have. `namespace`  + `methodName`  + `soapaction`. The idea is to generate the soapaction-value from the two first. But if just setting the `soapaction` then it will not work correctly. Not very intuitive to me.

304NotModified:
Thanks for the changes. It's OK to me to tweak a bit further on this PR

snakefoot:
Created new PR #3100 as followup.

PR #3089, opened 2019-01-21
SonarLint - Refactoring to reduce code complexity (Fix performance)

snakefoot:
Was a little too quick, so lost an optimization.


304NotModified:
👍

PR #3094, opened 2019-01-30
Fix AutoReload (never released bug)

beppemarazzi:
autoReload does not work when using config files in default location (i.e. NLog.config or NLog.dll.nlog)

### DIFF
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -1092,6 +1092,18 @@ namespace NLog
 
                 return values;
             }
+            public void Reset()
+            {
+                _loggerCache.Clear();
+            }
+        }
+
+        /// <remarks>
+        /// Internal for unit tests
+        /// </remarks>
+        internal void ResetLoggerCache()
+        {
+            _loggerCache.Reset();
         }
 
         /// <summary>
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -209,6 +209,7 @@ namespace NLog
                         try
                         {
                             _config = config;
+                            _configLoader.Activated(this, _config);
                             _config.Dump();
                             ReconfigExistingLoggers();
                             LogConfigurationInitialized();
--- a/tests/NLog.UnitTests/Config/ReloadTests.cs
+++ b/tests/NLog.UnitTests/Config/ReloadTests.cs
@@ -111,20 +111,12 @@ namespace NLog.UnitTests.Config
             string tempPath = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString());
             Directory.CreateDirectory(tempPath);
 
-
             string configFilePath = Path.Combine(tempPath, "reload.nlog");
-
-            var pp = Path.GetDirectoryName(typeof(LogManager).Assembly.Location);
-            configFilePath = Path.Combine(pp, "NLog.dll.nlog");
-
             WriteConfigFile(configFilePath, config1);
 
-
             try
-            {
-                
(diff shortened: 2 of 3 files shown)


304NotModified:
Thanks for the PR! Any idea since when this is broken?

snakefoot:
@beppemarazzi Thank you for catching this. Could you change the unit-test to be an advanced version of `LogFactory.SetCandidateConfigFilePaths` ? (Instead of putting nlog.config-files everywhere).

snakefoot:
@304NotModified Bug was introduced with #3058

snakefoot:
@beppemarazzi Good idea with the parameter and reusing the existing auto-reloading tests.

Because NLog really like global static state, then to avoid too much problems with state flowing from one test to another could you do the logic like this:

```c#
                if (useExplicitFileLoading)
                {
                    LogManager.Configuration = new XmlLoggingConfiguration(configFilePath);
                }
                else
                {
                    LogManager.Configuration = null;   // Avoid state following from other tests
                    LogManager.LogFactory.SetCandidateConfigFilePaths(new string[] { configFilePath });
                }
```

And put this code into a separate private method (with configFilePath as parameter), and then change all the tests to call this method.

beppemarazzi:
@snakefoot `LogManager.Configuration = null;` is already done before each test in `NLogTestBase` constructor.
After some debugging, i discovered that `LogManager.Configuration = null;` is not sufficient to reset the `LoggerCache`. For this reason i put  `ResetLoggerCache` in `ReloadTests` constructor.

If `LogManager.Configuration = null;` is supposed to clear all the static status (LoggerCache included) there is a big problem with other tests: doing `ResetLoggerCache` when nulling `LogManager.Configuration` will invalidate about 500 tests!!! Is it a breaking change???
IMHO it is another issue.

snakefoot:
No it should never clear the LoggerCache.


But are you sure it calls the constructor between each test within the same test-class?


Sent from my Sony Xperia


---- beppemarazzi wrote ----


@snakefoot<https://github.com/snakefoot> LogManager.Configuration = null; is already done before each test in NLogTestBase constructor.
After some debugging, i discovered that LogManager.Configuration = null; is not sufficient to reset the LoggerCache. For this reason i put ResetLoggerCache in ReloadTests constructor.

If LogManager.Configuration = null; is supposed to clear all the static status (LoggerCache included) there is a big problem with other tests: doing ResetLoggerCache when nulling LogManager.Configuration will invalidate about 500 tests!!! Is it a breaking change???
IMHO it is another issue.

snakefoot:
And also calls the constructor for every theory-combination?


Sent from my Sony Xperia


---- beppemarazzi wrote ----


@snakefoot<https://github.com/snakefoot> LogManager.Configuration = null; is already done before each test in NLogTestBase constructor.
After some debugging, i discovered that LogManager.Configuration = null; is not sufficient to reset the LoggerCache. For this reason i put ResetLoggerCache in ReloadTests constructor.

If LogManager.Configuration = null; is supposed to clear all the static status (LoggerCache included) there is a big problem with other tests: doing ResetLoggerCache when nulling LogManager.Configuration will invalidate about 500 tests!!! Is it a breaking change???
IMHO it is another issue.

beppemarazzi:
Yes, the constructor is called for each test (also theory combination). Just tried with debugger. Also because my workaround was to call `ResetLoggerCache` in `ReloadTests `constructor, and it works!

304NotModified:
This is ready for merge, isn't?

snakefoot:
Yes good to go


Sent from my Sony Xperia


This is ready for merge, isn't?

304NotModified:
Thanks!

PR #3098, opened 2019-02-01
<rule> support for one or more '*' and '?' wildcards and in any position

beppemarazzi:
IMHO it may be very useful... (and in my use case i need it)

### DIFF
--- a/src/NLog/Config/LoggingRule.cs
+++ b/src/NLog/Config/LoggingRule.cs
@@ -52,9 +52,7 @@ namespace NLog.Config
         private readonly bool[] _logLevels = new bool[LogLevel.MaxLevel.Ordinal + 1];
 
         private string _loggerNamePattern;
-        private MatchMode _loggerNameMatchMode;
-        private string _loggerNameMatchArgument;
-        private Regex _loggerNameMatchRegex;
+        private LoggerNameMatcher _loggerNameMatcher = new LoggerNameMatcher.None();
 
         /// <summary>
         /// Create an empty <see cref="LoggingRule" />.
@@ -78,7 +76,7 @@ namespace NLog.Config
         /// <summary>
         /// Create a new <see cref="LoggingRule" /> with a <paramref name="minLevel"/> and  <paramref name="maxLevel"/> which writes to <paramref name="target"/>.
         /// </summary>
-        /// <param name="loggerNamePattern">Logger name pattern. It may include the '*' wildcard at the beginning, at the end or at both ends.</param>
+        /// <param name="loggerNamePattern">Logger name pattern. It may include one or more '*' or '?' wildcards at any position.</param>
         /// <param name="minLevel">Minimum log level needed to trigger this rule.</param>
         /// <param name="maxLevel">Maximum log level needed to trigger this rule.</param>
         /// <param name="target">Target to be written to when the rule matches.</param>
@@ -93,7 +91,7 @@ namespace NLog.Config
         /// <summary>
         /// Create a new <see cref="LoggingRule" /> with a <paramref name="minLevel"/> which writes to <paramref name="target"/>.
         /// </summary>
-        /// <param name="loggerNamePattern">Logger name pattern. It may include the '*' wildcard at the beginning, at the end or at both ends.</param>
+        /// <param name="loggerNamePattern">Logger name pattern. It may include one or more '*' or '?' wildcards at any position.</param>
         /// <param name="minLevel">Minimum log level needed to trigger this rule.</param>
         /// <param name="target">Target to be written to when the rule matches.</param>
         public LoggingRule(string loggerNamePattern, LogLevel minLevel, Target target)
@@ -107,23 +105,13 @@ namespace NLog.Config
         /// <summary>
         /// Create a (disabled) <see cref="LoggingRule" />. You should call <see cref="EnableLoggingForLevel"/> or see cref="EnableLoggingForLevels"/> to enable logging.
         /// </summary>
-        /// <param name="loggerNamePattern">Logger name pattern. It may include the '*' wildcard at the beginning, at the end or at both ends.</param>
+        /// <param name="loggerNamePattern">Logger name pattern. It may include one or more '*' or '?' wildcards at any position.</param>
         /// <param name="target">Target to be written to when the rule matches.</param>
         public LoggingRule(string loggerNamePattern, Target target)
             : this()
         {
             LoggerNamePattern = loggerNamePattern;
             Targets.Add(target);
-        }
-
-        internal enum MatchMode
-        {
-            All,
-            Equals,
-            StartsWith,
-            EndsWith,
-            Contains,
-            Regex,
         }
 
         /// <summary>
@@ -171,56 +159,47 @@ namespace NLog.Config
             set
             {
                 _loggerNamePattern = value;
-                _loggerNameMatchRegex = null;
+                if(_loggerNamePattern == null)
+                {
+                    _loggerNameMatcher = new LoggerNameMatcher.None();
+                    return;
+                }
 
                 int starPos1 = _loggerNamePattern.IndexOf('*');
                 int starPos2 = _loggerNamePattern.IndexOf('*', starPos1 + 1);
                 int questionPos = _loggerNamePattern.IndexOf('?');
                 if (starPos1 < 0 && questionPos < 0)
                 {
-                    _loggerNameMatchMode = MatchMode.Equals;
-                    _loggerNameMatchArgument = value;
+                    _loggerNameMatcher = new LoggerNameMatcher.Equals(_loggerNamePattern);
                     return;
                 }
                 if(_loggerNamePattern == "*")
                 {
-                    _loggerNameMatchArgument = string.Empty;
-                    _loggerNameMatchMode = MatchMode.All;
+                    _loggerNameMatcher = new LoggerNameMatcher.All();
                     return;
                 }
                 if(questionPos < 0)
                 {
                     if(starPos1 == 0 && starPos2 == _loggerNamePattern.Length-1)
                     {
-                        _loggerNameMatchArgument = _loggerNamePattern.Substring(1, _loggerNamePattern.Length - 2);
-                        _loggerNameMatchMode = MatchMode.Contains;
+                        _loggerNameMatcher = new LoggerNameMatcher.Contains(_loggerNamePattern.Substring(1, _loggerNamePattern.Length - 2));
                         return;
                     }
                     if(starPos2<0)
                     {
                         if(starPos1 == 0)
                         {
-                            _loggerNameMatchArgument = _loggerNamePattern.Substring(1);
-                            _loggerNameMatchMode = MatchMode.EndsWith;
+                            _loggerNameMatcher = new LoggerNameMatcher.EndsWith(_loggerNamePattern.Substring(1));
                             return;
                         }
                         if (starPos1 == _loggerNamePattern.Length - 1)
                         {
-                            _loggerNameMatchArgument = _loggerNamePattern.Substring(0, _loggerNamePattern.Length - 1);
-                            _loggerNameMatchMode = MatchMode.StartsWith;
+                            _loggerNameMatcher = new LoggerNameMatcher.StartsWith(_loggerNamePattern.Substring(0, _loggerNamePattern.Length - 1));
                             return;
                         }
                     }
                 }
-                _loggerNameMatchMode = MatchMode.Regex;
-                _loggerNameMatchArgument =
-                    '^' +
-                    Regex.Escape(_loggerNamePattern)
-                        .Replace("\\*", ".*")
-                        .Replace("\\?", ".")
-                    + '$';
-                _loggerNameMatchRegex = new Regex(_loggerNameMatchArgument, RegexOptions.CultureInvariant);
-
+                _loggerNameMatcher = new LoggerNameMatcher.MultiplePattern(_loggerNamePattern);
             }
         }
 
@@ -324,7 +303,7 @@ namespace NLog.Config
         {
             var sb = new StringBuilder();
 
-            sb.AppendFormat(CultureInfo.InvariantCulture, "logNamePattern: ({0}:{1})", _loggerNameMatchArgument, _loggerNameMatchMode);
+            sb.Append(_loggerNameMatcher.ToString());
             sb.Append(" levels: [ ");
             for (int i = 0; i < _logLevels.Length; ++i)
             {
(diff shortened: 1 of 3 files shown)


beppemarazzi:
With a very little effort is also possible to support '?' wildcard but I
doubt it may be useful....

Il giorno Sab 2 Feb 2019 10:30 Rolf Kristensen <notifications@github.com>
ha scritto:

snakefoot:
Think only the method is called on Logger-creation or on ReconfigExistingLoggers. Not for every LogEvent. 

And for people calling GetLogger constantly then they will instead hit the LoggerCache without calling this method.

304NotModified:
I think your right @snakefoot.

304NotModified:
Still think this can be done without regex, and that would be better

304NotModified:
image

143ms ("cold boot") ...

304NotModified:
I understand why everything is moved to regex, but replacing startswith en endsWith with a regex is a no-go for me....

beppemarazzi:
You are right: Regex is less efficient than Starts/EndsWith

I've done a little test on my machine, have you got 5 minutes to try this code?
My results:

```
new1: 670,73521982304 us
Str1: 1543,75040969143 us
new2: 61,9089270024227 us
Match: 21,8502095302668 us
StartsWith: 1,32425512304648 us
EndsWith: 4,96595671142428 us
Contains: 2,64851024609295 us
Tot: 2307,51455190848 us
```


However... i'm going to propose this solution:
1 - if no wildcards => Equals
2 - if == '*' => All
3 - if no '?'
  3.1 - if there is only one * at beginning => EndsWith
  3.2 - if there is only one * at end => StartsWith
  3.3 - if there are 2 *, one at the beginning and one at the end => Contains
4 - Regex

What do you think about this?


```
using System;
using System.Diagnostics;
using System.Text.RegularExpressions;

public class Example
{
    static double TicksTo_us(long t)
    {
        var ret = (t * 1e6) / Stopwatch.Frequency;
        return ret;
    }

    public static void Main()
    {
        var sTot = new Stopwatch();
        var sNew1 = new Stopwatch();
        var sNew2 = new Stopwatch();
        var sMatch = new Stopwatch();
        var sStr1 = new Stopwatch();
        var sStartsWith = new Stopwatch();
        var sEndsWith = new Stopwatch();
        var sContains = new Stopwatch();

        sTot.Start();

        string pattern = @"^server\[.*]\.connection\[.]\..*$";
        string input = @"server[1234].connection[3].reader";
        string starts = @"server[1234].";
        string ends = @"connection[3].reader";
        string cont = @"34].connect";

        sNew1.Start();
        new Regex("foobar"); //first regex use...
        sNew1.Stop();

        sStr1.Start();
        "foobar".StartsWith("foo"); //first string use...
        sStr1.Stop();

        sNew2.Start();
        var r = new Regex(pattern, RegexOptions.CultureInvariant);
        sNew2.Stop();

        sMatch.Start();
        var match = r.Match(input);
        sMatch.Stop();

        sStartsWith.Start();
        var startsWith = input.StartsWith(starts);
        sStartsWith.Stop();

        sEndsWith.Start();
        var endsWith = input.EndsWith(ends);
        sEndsWith.Stop();

        sContains.Start();
        var contains = input.Contains(cont);
        sContains.Stop();

        sTot.Stop();
        
        Console.WriteLine("new1: " + TicksTo_us(sNew1.ElapsedTicks) + " us");
        Console.WriteLine("Str1: " + TicksTo_us(sStr1.ElapsedTicks) + " us");
        Console.WriteLine("new2: " + TicksTo_us(sNew2.ElapsedTicks) + " us");
        Console.WriteLine("Match: " + TicksTo_us(sMatch.ElapsedTicks) + " us");
        Console.WriteLine("StartsWith: " + TicksTo_us(sStartsWith.ElapsedTicks) + " us");
        Console.WriteLine("EndsWith: " + TicksTo_us(sEndsWith.ElapsedTicks) + " us");
        Console.WriteLine("Contains: " + TicksTo_us(sContains.ElapsedTicks) + " us");
        Console.WriteLine("Tot: " + TicksTo_us(sTot.ElapsedTicks) + " us");
    }
}
```

snakefoot:
Could also add an extra property to the logging-rule whether it should apply regex (Ex. MatchType or some other new property)

Regex can do lots of wonderful things. But regex also recognizes special characters, so if using a logging-rule where `logger` contains special characters then it might give surprises.

beppemarazzi:
@snakefoot : Not full understand.
But look at the code and let me explain:
```
                _loggerNameMatchMode = MatchMode.Regex;
                _loggerNameMatchArgument =
                    '^' +
                    Regex.Escape(_loggerNamePattern)
                        .Replace("\\*", ".*")
                        .Replace("\\?", ".")
                    + '$';

```

i first enclosed the pattern in a pair '^'/'$' to match the start and the end of string
then there is Regex.Escape(userProvidedPattern) that escapes all regex special characters inserted by 
user
then i replace "\*" (the escaped 'star') with ".*" (0 or more occurrences of any char)
and "\?" (the escaped question mark) with "." (1 occurrence of any char)

So if you are scared by the fact that the user could insert regex specific sequences, Regex.Escape manages it!

snakefoot:
> So if you are scared by the fact that the user could insert regex specific sequences, Regex.Escape manages it!

No I was actually suggesting that the user could be allowed to use standard regex in matching. Ex.:

```xml
<logger name=".*" minlevel="Debug" matchType="RegEx" />
```

snakefoot:
Think I prefer the regex-mode-flag instead of special pattern. But it is just an idea. Think @304NotModified has the final decision.

304NotModified:
I think the regex flag isn't a bad idea, but if it's not needed (besides transparently) , I prefer without the flag. 

> However... i'm going to propose this solution:
> 1 - if no wildcards => Equals
> 2 - if == '*' => All
> 3 - if no '?'
> 3.1 - if there is only one * at beginning => EndsWith
> 3.2 - if there is only one * at end => StartsWith
> 3.3 - if there are 2 *, one at the beginning and one at the end => Contains
> 4 - Regex

I think this is best of both worlds, fast when possible, full (regex) support when needed. 

Maybe it's also nice to refactor the code before creating this feature, I think the matching should be moved outside this class and with usage of sub classes. The refactoring could be a separate PR.

e.g.  something like this

```c#

interface ILoggerNameMatcher 
{
   bool IsMatch(string name)
}

class StartsWithNameMatcher : ILoggerNameMatcher {}
class EndsWithNameMatcher : ILoggerNameMatcher {}
class RegexNameMatcher : ILoggerNameMatcher {} // or maybe something like MultiplePatternNameMatcher
```

What do guys think ?

beppemarazzi:
> > However... i'm going to propose this solution:
> > 1 - if no wildcards => Equals
> > 2 - if == '*' => All
> > 3 - if no '?'
> > 3.1 - if there is only one * at beginning => EndsWith
> > 3.2 - if there is only one * at end => StartsWith
> > 3.3 - if there are 2 *, one at the beginning and one at the end => Contains
> > 4 - Regex
> 
> I think this is best of both worlds, fast when possible, full (regex) support when needed.

With my last commit i implemented this algorithm (and some tests). Is it sufficiently readable? Need more comments?

beppemarazzi:
About full-regex support:
IMHO there is a problem with the flag solution proposed by @snakefoot 

`<logger name=".*" minlevel="Debug" matchMode="RegEx" />`

now the logic for select the matching algorithm is in `LoggerNamePattern` setter, so during deserialization from config it is an "atomical" operation.
Splitting it in two properties causes the dependency in props deserialization order... it smells

304NotModified:
> So you want to to eliminate the `switch (_loggerNameMatchMode)`

yes ditch it :)

> And what about ToString implementation? Now it allow testing!

testing ToString isn't that good, if we couldn't test it now we could do: add a public or internal method/property

> 
If you validate the WIP with new algorithm, i can go with sub-classes.... 

the sub classes are now there isn't? 

now reviewing...

src/NLog/Config/LoggingRule.cs
```diff
@@ -166,49 +159,47 @@ public string LoggerNamePattern
...
+                if(_loggerNamePattern == null)
                 {
-                    _loggerNameMatchMode = MatchMode.Equals;
-                    _loggerNameMatchArgument = value;
+                    _loggerNameMatcher = new LoggerNameMatcher.None();
                     return;
                 }
 
-                if (firstPos == lastPos)
+                int starPos1 = _loggerNamePattern.IndexOf('*');
```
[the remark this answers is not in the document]
beppemarazzi: Done

src/NLog/Config/LoggingRule.cs
```diff
@@ -354,29 +345,107 @@ public bool IsLoggingEnabledForLevel(LogLevel level)
         /// <returns>A value of <see langword="true"/> when the name matches, <see langword="false" /> otherwise.</returns>
         public bool NameMatches(string loggerName)
         {
-            switch (_loggerNameMatchMode)
+            return _loggerNameMatcher.NameMatches(loggerName);
+        }
+
+        abstract class LoggerNameMatcher
```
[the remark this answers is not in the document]
beppemarazzi: Done

src/NLog/Config/LoggingRule.cs
```diff
@@ -354,29 +345,107 @@ public bool IsLoggingEnabledForLevel(LogLevel level)
...
+                public Contains(string pattern) : base(pattern) { }
+                public override bool NameMatches(string loggerName)
+                {
+                    return loggerName.IndexOf(_pattern, StringComparison.Ordinal) >= 0;
+                }
+            }
+            internal class MultiplePattern : LoggerNameMatcher
+            {
+                private readonly Regex _regex;
+                private static string getRegex(string wildcardsPattern)
```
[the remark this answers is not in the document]
beppemarazzi: Done

src/NLog/Config/LoggingRule.cs
```diff
@@ -354,29 +345,107 @@ public bool IsLoggingEnabledForLevel(LogLevel level)
...
+                }
+                public override bool NameMatches(string loggerName)
+                {
+                    return _regex.IsMatch(loggerName);
+                }
+            }
+            internal class RegexPattern : LoggerNameMatcher
+            {
+                private readonly Regex _regex;
+                public RegexPattern(string pattern) : base(pattern)
```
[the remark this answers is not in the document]
beppemarazzi: The idea here is to define a syntax to provide inline Regex enclosed in a pair of separators. For a proof of concept look at 540691c

beppemarazzi:
@304NotModified  with 3c08362 i'm pretty happy... waiting your review

src/NLog/Config/LoggingRule.cs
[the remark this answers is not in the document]
304NotModified: isn't converting the pattern to regex a job for the "create" method?

src/NLog/Config/LoggingRule.cs
```diff
@@ -354,29 +345,107 @@ public bool IsLoggingEnabledForLevel(LogLevel level)
...
-                    return loggerName.StartsWith(_loggerNameMatchArgument, StringComparison.Ordinal);
-
-                case MatchMode.EndsWith:
-                    return loggerName.EndsWith(_loggerNameMatchArgument, StringComparison.Ordinal);
-
-                case MatchMode.Contains:
-                    return loggerName.IndexOf(_loggerNameMatchArgument, StringComparison.Ordinal) >= 0;
+                }
+            }
+            internal class All : LoggerNameMatcher
```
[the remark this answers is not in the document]
304NotModified: I think adding ` protected abstract string Name {get; }` (to `LoggerNameMatcher`) is better design then. More transparant and we can name those classes proper to the rest of the code base (MSDN name guidelines).

Also I prefer small code files, than larger ones. Those are easier to maintain and review etc.

tests/NLog.UnitTests/Config/ConfigApiTests.cs
```diff
@@ -304,5 +313,181 @@ public void ConfigLogRuleWithName()
...
+        public void LoggerNameMatcher_MultiplePattern_EscapedChars()
+        {
+            var matcher = LoggerNameMatcher.Create("a?b.c.foo.bar");
+            Assert.Equal("logNamePattern: (^a.b\\.c\\.foo\\.bar$:MultiplePattern)", matcher.ToString());
+        }
+
+        [Fact]
+        public void LoggerNameMatcher_RegexPattern_Valid()
+        {
+            var matcher = LoggerNameMatcher.Create("{ab.c}");
```
304NotModified: - this is another feature isn't? (Not described in the PR title)

- Is it's logical to wrap them in `{`? It's more common to mee to do `/ab.c/` to me (at least in Javascript ;))

304NotModified: maybe in this case, a Regex property is indeed better. I was under the impression first that we didn't needed it.

snakefoot: Can only repeat that I prefer a explicit regex-mode-flag instead of special pattern. Rather remove the feature than special pattern.

304NotModified: I think delimiter isn't really clear, so b) disable or future or c) property 

I think c isn't difficult, but also no problems with b) as I never saw complains about the missing regex in the past years.

304NotModified: > [8e417b6](https://github.com/NLog/NLog/commit/8e417b688b1e0c7d928bb7f81ae28fc509570ede) disables RegexPattern with a compiler macro `#if FEATURE_REGEXPATTERN`.
> If you prefer i can completely delete it.

I prefer deletion, in the future we can add it with another commit / branch

> Is there a place where i can see the builded documentation? I want to check how looks what i wrote....

Do you mean the sandcastle documentation? You need to run SHFB locally then (there is a config file in the repo)

304NotModified: > I prefer deletion, in the future we can add it with another commit / branch

If you could create a backup in a new branch with the regex option, I could help with adding it

beppemarazzi: Ok, i've build the doc. But there is a problem:
now `LoggerNameMatcher` is an internal class: it wraps an "internal detail" concept that isn't needed in the public scope.
But (as expected) all the documentation i wrote around `LoggerNameMatcher` class is not exposed...
Now i could:
a - publicize  `LoggerNameMatcher` with the sole purpose of publicize the doc
b - collapse all the doc in a suitable place in `LoggingRule` class (probably the `LoggerNamePattern` method with a reminder from constructors and `NameMatches` method

i prefer b. What do you think?

beppemarazzi: Done some refinements in doc and deleted the regex option. IMHO on this PR i'm done.

Done a [branch](/beppemarazzi/NLog/tree/feature/rule-regex-pattern) with b262f87  (actually is 1 commit behind my last commit 20c4440) with regex pattern stub commented out by `#if FEATURE_REGEXPATTERN`. Or do you prefer a new PR on this?

304NotModified:
great! Thanks also for the documentation and the very good coverage :)

tests/NLog.UnitTests/Config/ConfigApiTests.cs
304NotModified: > Or do you prefer a new PR on this?

PR would be nice, but I think I will focus first on NLog 4.6 (rc)

304NotModified:
@beppemarazzi NLog 4.6 RC is live! You you please help us with the docs? Thanks!

beppemarazzi:
Done some doc at https://github.com/NLog/NLog/wiki/Configuration-file

304NotModified:
thanks!

PR #3100, opened 2019-02-02
WebService target allow override of SoapAction through Headers

snakefoot:
Follow up to #3081

And reduce string-allocations by only doing string.Concat once for ContentType and SoapAction (Unless overriding using Headers-Layout)


304NotModified:
👍

snakefoot:
@304NotModified Updated Wiki: https://github.com/NLog/NLog/wiki/WebService-target (About possible override of SoapAction for Soap11)

PR #3103, opened 2019-02-04
StringBuilderPool - Reduce memory overhead until required

snakefoot:
No need to capture (and hold) a lot of memory without knowing if it will be used.

Ex running console program on 56 core machine allocates 400 StringBuilders of 16K. Now it will allocate 300 stringbuilders of 1K (6MB vs 0.3MB)


snakefoot:
Restores balance in the universe after multiplying the max-size with 10 for the fast-buffer in #2934

304NotModified:
Thanks!

PR #3107, opened 2019-02-06
Removed old todos and added missing docgen + Reduce code complexity of ParseRuleElement

snakefoot:
Last commit before vacation


304NotModified:
Thanks!

PR #3120, opened 2019-02-12
Improve build script

304NotModified:
easier to add new packages


304NotModified:
>   System.IO.IOException : The process cannot access the file 'NLog.dll' because it is being used by another process.

strange...  https://ci.appveyor.com/project/nlog/nlog/builds/22365221

PR #3131, opened 2019-02-20
DatabaseTarget - Removed logEvent-parameter from CreateDatabaseParameter

snakefoot:
Update the new NLog 4.6 method CreateDatabaseParameter, so it is open for batching support.

### DIFF
--- a/src/NLog/Targets/DatabaseTarget.cs
+++ b/src/NLog/Targets/DatabaseTarget.cs
@@ -92,7 +92,6 @@ namespace NLog.Targets
         /// </summary>
         public DatabaseTarget()
         {
-            Parameters = new List<DatabaseParameterInfo>();
             InstallDdlCommands = new List<DatabaseCommandInfo>();
             UninstallDdlCommands = new List<DatabaseCommandInfo>();
             DBProvider = "sqlserver";
@@ -267,7 +266,7 @@ namespace NLog.Targets
         /// </summary>
         /// <docgen category='SQL Statement' order='14' />
         [ArrayParameter(typeof(DatabaseParameterInfo), "parameter")]
-        public IList<DatabaseParameterInfo> Parameters { get; private set; }
+        public IList<DatabaseParameterInfo> Parameters { get; } = new List<DatabaseParameterInfo>();
 
 #if !NETSTANDARD
         internal DbProviderFactory ProviderFactory { get; set; }
@@ -284,6 +283,8 @@ namespace NLog.Targets
             set => _propertyTypeConverter = value;
         }
         private IPropertyTypeConverter _propertyTypeConverter;
+
+        SortHelpers.KeySelector<AsyncLogEventInfo, string> _buildConnectionStringDelegate;
 
         /// <summary>
         /// Performs installation which requires administrative permissions.
@@ -466,6 +467,7 @@ namespace NLog.Targets
             return providerName;
         }
 #endif
+
         /// <summary>
         /// Set the <see cref="ConnectionType"/> to use it for opening connections to the database.
         /// </summary>
@@ -533,7 +535,7 @@ namespace NLog.Targets
         {
             try
             {
-                WriteEventToDatabase(logEvent);
+                WriteEventToDatabase(logEvent, BuildConnectionString(logEvent));
             }
             catch (Exception exception)
             {
@@ -580,7 +582,10 @@ namespace NLog.Targets
         /// <param name="logEvents">Logging events to be written out.</param>
         protected override void Write(IList<AsyncLogEventInfo> logEvents)
         {
-            var buckets = logEvents.BucketSort(c => BuildConnectionString(c.LogEvent));
+            if (_buildConnectionStringDelegate == null)
+                _buildConnectionStringDelegate = (l) => BuildConnectionString(l.LogEvent);
+
+            var buckets = logEvents.BucketSort(_buildConnectionStringDelegate);
 
             try
             {
@@ -592,7 +597,7 @@ namespace NLog.Targets
 
                         try
                         {
-                            WriteEventToDatabase(ev.LogEvent);
+                            WriteEventToDatabase(ev.LogEvent, kvp.Key);
                             ev.Continuation(null);
                         }
                         catch (Exception exception)
@@ -625,29 +630,23 @@ namespace NLog.Targets
                 }
             }
         }
+
         /// <summary>
         /// Write logEvent to database
         /// </summary>
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Security", "CA2100:Review SQL queries for security vulnerabilities", Justification = "It's up to the user to ensure proper quoting.")]
-        private void WriteEventToDatabase(LogEventInfo logEvent)
+        private void WriteEventToDatabase(LogEventInfo logEvent, string connectionString)
         {
+            var commandText = RenderLogEvent(CommandText, logEvent);
+            InternalLogger.Trace("DatabaseTarget(Name={0}): Executing {1}: {2}", Name, CommandType, commandText);
+
             //Always suppress transaction so that the caller does not rollback logging if they are rolling back their transaction.
             using (TransactionScope transactionScope = new TransactionScope(TransactionScopeOption.Suppress))
             {
-                EnsureConnectionOpen(BuildConnectionString(logEvent));
+                EnsureConnectionOpen(connectionString);
 
-                using (IDbCommand command = _activeConnection.CreateCommand())
+                using (IDbCommand command = CreateDbCommandWithParameters(logEvent, CommandType, commandText, Parameters))
                 {
-                    command.CommandText = RenderLogEvent(CommandText, logEvent);
-                    if (command.CommandType != CommandType)
-                    {
-                        command.CommandType = CommandType;  // Some DbProviders will throw when trying to modify CommandType
-                    }
-
-                    InternalLogger.Trace("DatabaseTarget(Name={0}): Executing {1}: {2}", Name, command.CommandType, command.CommandText);
-
-                    AddParametersToCommand(command, Parameters, logEvent);
-
                     int result = command.ExecuteNonQuery();
                     InternalLogger.Trace("DatabaseTarget(Name={0}): Finished execution, result = {1}", Name, result);
                 }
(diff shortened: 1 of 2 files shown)


snakefoot:
Custom pre-release-build for those who wants to test (Just remove zip-extension):

[NLog.4.6.0-PR3131.nupkg.zip](https://github.com/NLog/NLog/files/2886704/NLog.4.6.0-PR3131.nupkg.zip)

snakefoot:
Custom pre-release-build for those who wants to test performance (Just remove zip-extension):

[NLog.4.6.0-PR3131.nupkg.zip](https://github.com/NLog/NLog/files/2886733/NLog.4.6.0-PR3131.nupkg.zip)

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -533,7 +539,18 @@ protected override void Write(LogEventInfo logEvent)
         {
             try
             {
-                WriteEventToDatabase(logEvent);
+                var connectionString = BuildConnectionString(logEvent);
+
+                //Always suppress transaction so that the caller does not rollback logging if they are rolling back their transaction.
+                using (TransactionScope transactionScope = new TransactionScope(TransactionScopeOption.Suppress))
```
304NotModified: this is another change then the PR suggests?

snakefoot: Nope just me moving code around to optimize the batching-logic.

304NotModified:
Thanks! moved it to https://github.com/NLog/NLog/pull/3132, so we could easy integrate it into dev and master.

src/NLog/Targets/DatabaseTarget.cs
304NotModified: OK thanks!

snakefoot:
Could this become NLog 4.6, because this is actually a breaking change if NLog 4.6 is released.

snakefoot:
Also ready to separate PR that just includes the breaking change in `CreateDatabaseParameter`. Just to prepare for this PR.

304NotModified:
please rebase on dev, thanks!

snakefoot:
> @304NotModified  please rebase on dev, thanks!

Done

D9001:
> Custom pre-release-build for those who wants to test performance (Just remove zip-extension):
> 
> [NLog.4.6.0-PR3131.nupkg.zip](https://github.com/NLog/NLog/files/2886733/NLog.4.6.0-PR3131.nupkg.zip)

> 2019-02-21 18:33:54.0713 Error DatabaseTarget(Name=wrapped_db): Error when writing to database. Exception: System.InvalidOperationException: SqlCommand.Prepare method requires all parameters to have a
n explicitly set type.
   at System.Data.SqlClient.SqlParameter.Prepare(SqlCommand cmd)
   at System.Data.SqlClient.SqlCommand.Prepare()
   at NLog.Targets.DatabaseTarget.WriteEventToDatabase(IDbCommand command, LogEventInfo logEvent, Layout commandLayout, IList`1 databaseParameterInfos, Boolean prepareDbCommand)
   at NLog.Targets.DatabaseTarget.Write(IList`1 logEvents)

I received this error. I'm going home tonight and will see if I'm missing an attribute on the parameter nodes, but I've essentially mirrored your example + added my connection string.

snakefoot:
> I received this error. I'm going home tonight and will see if I'm missing an attribute on the parameter nodes, but I've essentially mirrored your example + added my connection string.

Thanks for testing. Have updated the xml in the original issue-post with dbType for each parameter. Please update your test config and try again. See also: https://docs.microsoft.com/en-us/dotnet/api/system.data.dbtype

D9001:
> > I received this error. I'm going home tonight and will see if I'm missing an attribute on the parameter nodes, but I've essentially mirrored your example + added my connection string.
> 
> Thanks for testing. Have updated the xml in the original issue-post with dbType for each parameter. Please update your test config and try again. See also: https://docs.microsoft.com/en-us/dotnet/api/system.data.dbtype

I tried adding dbType earlier, but I used "int32" as the documentation for DatabaseTarget had an example in lower case. But maybe it's case sensitive. I will try it out with "Int32" instead.

snakefoot:
> I tried adding dbType earlier, but I used "int32" as the documentation for DatabaseTarget had an example in lower case. But maybe it's case sensitive. I will try it out with "Int32" instead.

Else try doing this instead:

`<parameter name="@level" layout="${level:format=ordinal}" dbType="SqlDbType.Int" />`

304NotModified:
> System.InvalidOperationException: SqlCommand.Prepare method requires all parameters to have a
> n explicitly set type.

PS: I think we won't do that by default, nor we can do because of backwards comp. So specify al DBtype attributes indeed

D9001:
> > I tried adding dbType earlier, but I used "int32" as the documentation for DatabaseTarget had an example in lower case. But maybe it's case sensitive. I will try it out with "Int32" instead.
> 
> Else try doing this instead:
> 
> `<parameter name="@level" layout="${level:format=ordinal}" dbType="SqlDbType.Int" />`

I tried both Int32 and SqlDbType.Int with no luck yet; the same error appears. Does this configuration appear to be correct?

```
  <targets>
    <target name="asyncDatabase" type="AsyncWrapper" timeToSleepBetweenBatches="10000" queueLimit="100" overflowAction="Grow">
      <target name="wrapped_db" connectionstring="${gdc:item=ConnectionString}" xsi:type="Database" prepareDbCommand="True" keepConnection="true">
        <commandText>
          INSERT INTO Logging ([Identifier]) VALUES (@Id);
        </commandText>
        <parameter name="@Id" dbType="SqlDbType.Int" layout="${event-properties:item=Id}"/>
      </target>
    </target>
  </targets>
```

304NotModified:
config looks valid. Could you please capture and post the internal log ? (trace level)

D9001:
> config looks valid. Could you please capture and post the internal log ? (trace level)

I forgot to post an update earlier, but this was due to a brain fart where I hadn't cleaned/rebuilt the solution so the nlog.config file being used was from my first attempt. I was able to bind the parameter in my example as well as dbTypes of SqlDbType.VarChar by specifying the size/length attributes as well.

304NotModified:
🎉 great news! Looking forward to the results!

304NotModified:
@kwaak maybe this feature is also nice for you :)

snakefoot:
@304NotModified Improved the logic when using non-string DbType where Layout-result is string.Empty. Returning default-value instead of errors about unable to convert empty string.

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -629,32 +664,46 @@ protected override void Write(IList<AsyncLogEventInfo> logEvents)
...
+            }
 
-                //not really needed as there is no transaction at all.
-                transactionScope.Complete();
+            if (!prepareDbCommand || command.CommandText != commandText)
+            {
+                command.CommandText = commandText;
+                if (prepareDbCommand)
+                {
+                    command.Prepare();
```
304NotModified: I'm not sure if we should prepare every sql query (worst case).

I would expect this in the init method.

snakefoot: Could make a cache of prepared DbCommands, and perform lookup using the key of ConnectionString + CommandText.

304NotModified:
PS ;)

>  would recommend against calling SqlCommand.Prepare() because the only potential benefit is saving network packets while the downside is taking up more server memory. 

https://dba.stackexchange.com/a/129721/32229

... to measure is to know... :)

snakefoot:
> ... to measure is to know... :)

Don't mind removing the Prepare, and just change the parameter to be `ReuseDbCommand`

Extra caching is needed for SqlClient as SqlCommand-constructor starts as unprepared.

snakefoot:
@D9001 Attached an updated build with caching of prepared DbCommands (See original post where new attachment has been added **NLog.4.6.0-PS3131.nupkg.zip**)

snakefoot:
@D9001 Btw NLog tries to perform correct type-conversion based on DbType. But when using "special" types like  `SqlDbType.Int` then it fails to guess the correct value-type. You can help NLog by doing this (Adding parameterType="System.Int32"):

`<parameter name="@ManagerId" layout="${event-properties:item=ManagerId}" dbType="SqlDbType.Int" parameterType="System.Int32" />`

NLog will only guess correctly for "special" DbTypes when having these names: Date, TimeStamp, Double, Decimal, Bool, Guid

D9001:
@snakefoot Thanks for the update. I've added the parameterType attributes for each parameter and have updated the NuGet package used in our test environment.

snakefoot:
@D9001 Thank you for the update about what happens on the Server-side. Could you activate the NLog InternalLogger on Trace-Level, so one could see what happens on the Client-side?

See also: https://github.com/NLog/NLog/wiki/Internal-logging (Might have to check the log-file for sensitive information before posting to github).

Btw. if performance is important then you should NOT use the BufferingWrapper. Instead use the AsyncWrapper as shown in the first post for this PR. Make sure that your logger-rules points to the wrapper-target and not the database-target or else the batching will not be activated.

D9001:
> @D9001 Thank you for the update about what happens on the Server-side. Could you activate the NLog InternalLogger on Trace-Level, so one could see what happens on the Client-side?
> 
> See also: https://github.com/NLog/NLog/wiki/Internal-logging (You might have to check the log-file for sensitive information before posting it here on github).
> 
> Btw. if performance is important then you should NOT use the BufferingWrapper (Mostly for throttle and to ensure high latency). Instead use the AsyncWrapper as shown in the first post for this PR. Make sure that your logger-rules points to the name of the wrapper-target and not the database-target or else the batching will not be activated.

Ah, thanks for explaining that. I thought I had read that BufferingWrapper provided the same functionality as AsyncWrapper as long as sliding timeouts were enabled. I figured it was better since I originally thought it would create one long insert statement. I'll swap to the AsyncWrapper and test that out with internal logging enabled also.

D9001:
Here's the internal logging from a regular run (not performance) with some sensitive information I've omitted in between: https://justpaste.it/20eml

snakefoot:
@D9001 That doesn't look good. Huge latency in the communication with the Database-instance. Guess a solution with sending an entire DataTable is needed for this case.

snakefoot:
@D9001 Would you mind trying out #3150 ? (Use `xsi:type="DataTable"` instead of `xsi:type="Database" and configure the parameter `updateBatchSize="200"`)

snakefoot:
@304NotModified The changes in this PR should still reduce client-side overhead for those using the DatabaseTarget with a local database-instance (without high latency)

304NotModified:
This can be closed in favor of #3150?

snakefoot:
@304NotModified Have now removed the PrepareDbCommand-logic and instead just made the change to the new NLog 4.6 method: CreateDatabaseParameter

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -884,7 +878,7 @@ protected virtual IDbDataParameter CreateDatabaseParameter(IDbCommand command, D
         /// </summary>
         /// <param name="logEvent">Current logevent.</param>
         /// <param name="parameterInfo">Parameter configuration info.</param>
-        protected internal virtual object GetParameterValue(LogEventInfo logEvent, DatabaseParameterInfo parameterInfo)
+        protected internal virtual object CreateDatabaseParameterValue(LogEventInfo logEvent, DatabaseParameterInfo parameterInfo)
```
304NotModified: IMO this is more a retrieve/get then a create. It won't create (`new`) things?

snakefoot:
It will create a value for all situations except when extracted rawvalue matches dbtype


Sent from my Sony Xperia


In src/NLog/Targets/DatabaseTarget.cs<https://github.com/NLog/NLog/pull/3131#discussion_r262997383>:

> @@ -884,7 +878,7 @@ protected virtual IDbDataParameter CreateDatabaseParameter(IDbCommand command, D
         /// </summary>
         /// <param name="logEvent">Current logevent.</param>
         /// <param name="parameterInfo">Parameter configuration info.</param>
-        protected internal virtual object GetParameterValue(LogEventInfo logEvent, DatabaseParameterInfo parameterInfo)
+        protected internal virtual object CreateDatabaseParameterValue(LogEventInfo logEvent, DatabaseParameterInfo parameterInfo)


IMO this is more a retrieve/get then a create. It won't create (new) things?

src/NLog/Targets/DatabaseTarget.cs
snakefoot: It will create a value for all situations except when extracted rawvalue matches dbtype

304NotModified: It's a getOrCreate ;)

Not sure if we should have a virtual getOrCreate (need to check that)

IMO a create method should never "get”, so "get" is s better name (I'm trying to keep it easy to implement this method by others as it's virtual)

snakefoot: Renamed to GetDatabaseParameterValue

PR #3137, opened 2019-02-23
XmlLayout - Renamed ElementName to NodeName

snakefoot:
Renamed ElementValue to NodeLayout. Renamed Elements to ChildNodes (more terse config and aligns attribute with node a little more):

```xml
<layout type='xmllayout' nodeName='log4j:event'>
  <attribute name='logger' layout='${logger}' includeEmptyValue='true' />
  <attribute name='level' layout='${uppercase:${level}}' includeEmptyValue='true' />
  <node nodeName='log4j:message' nodeLayout='${message}' />
  <node nodeName='log4j:throwable' nodeLayout='${exception:format=tostring}' />
  <node nodeName='log4j:locationInfo'>
    <attribute name='class' layout='${callsite:methodName=false}' includeEmptyValue='true' />
  </node>
</layout>
```

Old style:
```xml
<layout type='xmllayout' elementName='log4j:event'>
  <attribute name='logger' layout='${logger}' includeEmptyValue='true' />
  <attribute name='level' layout='${uppercase:${level}}' includeEmptyValue='true' />
  <element elementName='log4j:message' elementValue='${message}' />
  <element elementName='log4j:throwable' elementValue='${exception:format=tostring}' />
  <element elementName='log4j:locationInfo'>
    <attribute name='class' layout='${callsite:methodName=false}' includeEmptyValue='true' />
  </element>
</layout>
```

### DIFF
--- a/src/NLog/Layouts/XmlLayout.cs
+++ b/src/NLog/Layouts/XmlLayout.cs
@@ -48,7 +48,7 @@ namespace NLog.Layouts
     [ThreadSafe]
     public class XmlLayout : Layout
     {
-        private const string DefaultTopElementName = "logevent";
+        private const string DefaultTopNodeName = "logevent";
         private const string DefaultPropertyName = "property";
         private const string DefaultPropertyKeyAttribute = "key";
         private const string DefaultCollectionItemName = "item";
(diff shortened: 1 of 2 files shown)


304NotModified:
while XML is a tree-structure, the term "node" isn't in the domain of XML. There are only elements and attributes. (e.g. see here: https://www.w3.org/TR/2008/REC-xml-20081126/)

I was thinking we could do this: (but note: the root `<layout` uses elementName instead of Name


```xml
<layout type='xmllayout' elementName='log4j:event'>
  <attribute name='logger' layout='${logger}' includeEmptyValue='true' />
  <attribute name='level' layout='${uppercase:${level}}' includeEmptyValue='true' />
  <element name='log4j:message' value='${message}' />
  <element name='log4j:throwable' value='${exception:format=tostring}' />
  <element name='log4j:locationInfo'>
    <attribute name='class' layout='${callsite:methodName=false}' includeEmptyValue='true' />
  </element>
</layout>

```

304NotModified:
not sure if we could split the "name" of the `<layout` and the `<element`. I think the comments are wrong? " Name of the top level XML element" - it's not only the top-level?

snakefoot:
Think I will just step away from this. Not starting discussion about root, top or etc. naming.

Issue #3139, opened 2019-02-23
IRawValue - support for renderers that *could* have layouts


304NotModified:

Page-Not-Found:
@304NotModified - After about 2 years, I am ready to support nlog again. Is this up for grabs?

304NotModified:
Hey! Cool!

I think this is a hard one start with, but you could give it a try. :)

Page-Not-Found:
Could you add a little more detail?

304NotModified:
We like to get this working:

https://github.com/NLog/NLog/pull/3140/files#diff-b8f49ba5c2431083e904f0e421818acb

So one of the results of the ${when} is a Irawvalue layout, and the other isn't (of course both or none should also work).

I think this should work also in more renderers, but ${when} is an example.

related changes (after the test of #3140?) #3165, #3144, #2973

PR #3140, opened 2019-02-23
IRawvalue support in ${when}

304NotModified:
fixes https://github.com/NLog/NLog/issues/3139

todo 
- other conditional layouts

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/3140)

### DIFF
--- a/src/NLog/LayoutRenderers/Wrappers/WhenLayoutRendererWrapper.cs
+++ b/src/NLog/LayoutRenderers/Wrappers/WhenLayoutRendererWrapper.cs
@@ -38,6 +38,7 @@ namespace NLog.LayoutRenderers.Wrappers
     using NLog.Conditions;
     using NLog.Config;
     using NLog.Layouts;
+    using NLog.Internal;
 
     /// <summary>
     /// Only outputs the inner layout when the specified condition has been met.
@@ -46,7 +47,7 @@ namespace NLog.LayoutRenderers.Wrappers
     [AmbientProperty("When")]
     [ThreadAgnostic]
     [ThreadSafe]
-    public sealed class WhenLayoutRendererWrapper : WrapperLayoutRendererBuilderBase
+    public sealed class WhenLayoutRendererWrapper : WrapperLayoutRendererBuilderBase, IRawValue
     {
         /// <summary>
         /// Gets or sets the condition that must be met for the <see cref="WrapperLayoutRendererBase.Inner"/> layout to be printed.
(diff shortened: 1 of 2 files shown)


304NotModified:
@snakefoot could you please explain to me:

> NLog.UnitTests.ApiTests.TryGetRawValue_ThreadAgnostic_Attribute_Required [FAIL]
>      NLog.LayoutRenderers.Wrappers.WrapperLayoutRendererBase cannot implement IRawValue

snakefoot:
Erhh you have just added `IRawValue` to the base class, and suddenly changed the meaning of the interface.

Before it was not conditional. Now it is. Before it was not on base-classses. Now it is.

304NotModified:
> Erhh you have just added IRawValue to the base class, and suddenly changed the meaning of the interface

Not really, as the interface was about retrieving raw values. But ok, the implementation was (unfortunately?) more limited

> Before it was not conditional. Now it is. Before it was not on base-classses. Now it is.

> The unit-test verifies the old interface meaning. When changing the meaning, then the unit-test will probably also need to be changed.

Any ideas with this change? Change the unit test? Is this change a bad idea? Do you expect unwanted side effects of this change?

snakefoot:
> Any ideas with this change? Change the unit test? Is this change a bad idea? Do you expect unwanted side effects of this change?

I see both IRawValue and IStringValue as hacks to improve performance. As the Layout and LayoutRenderers have been born to output strings. They have to be implemented carefully as they can break the safety net need when doing async-logging.

If we want to open up for people setting up their own logic for returning actual objects, then I think we should consider looking at the Serilog-enrichers. Maybe getting inspiration from there?

304NotModified:
Clear, but that's more about go to public with the irawvalue 

What about implementing it like this for ${when}?
 I could merge the refactor (to try pattern) an do the  ${when} impl in a later version.

snakefoot:
I would probably just settle for `WhenEmptyLayoutRendererWrapper` since it is the standard coalesce-method in NLog. 

But it sounds like you are actually using the DatabaseTarget, so I guess you have a better idea of common use-cases. Prefer not to sprinkle performance hacks everywhere unless they are part of main-code-path.

304NotModified:
after #3144 this isn't a breaking change

snakefoot:
@304NotModified Could also make a special layoutrenderer-wrapper that support IRawValue and if the inner-Layout supports IRawValue-interface and it returns null/empty-string. Then it can coalesce to DbNull-value as raw-value.

snakefoot:
But after doing performance testing in #3150, then IRawValue-optimization of DateTime is dwarfed by the overhead in DatabaseTarget of building DbCommands.

So right now the IRawValue goal is to allow one to return dbnull, because one cannot render a magic-string value that can be converted by to dbnull.

304NotModified:
> But after doing performance testing in #3150, then IRawValue-optimization of DateTime is dwarfed by the overhead in DatabaseTarget of building DbCommands

Please elaborate, can't follow this unfortunately

snakefoot:
> Please elaborate, can't follow this unfortunately

Just saying that IRawValue is a very special hammer, and one have to look hard for nails where the hammer really shines.

304NotModified:
Lol I'm more lost now. Another day I think ;)

304NotModified:
TIME

304NotModified:
superseded by https://github.com/NLog/NLog/pull/3373

PR #3144, opened 2019-02-25
TryGetRawValue instead of GetRawValue

304NotModified:
part of #3140

### DIFF
--- a/src/NLog/Internal/IRawValue.cs
+++ b/src/NLog/Internal/IRawValue.cs
@@ -42,8 +42,8 @@ namespace NLog.Internal
         /// Get the raw value
         /// </summary>
         /// <param name="logEvent"></param>
-        /// <returns>null if not possible or unknown</returns>
-        object GetRawValue(LogEventInfo logEvent);
-
+        /// <param name="value">The value</param>
+        /// <returns>RawValue supported?</returns>
+        bool TryGetRawValue(LogEventInfo logEvent, out object value);
     }
 }
--- a/src/NLog/LayoutRenderers/CallSiteLineNumberLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/CallSiteLineNumberLayoutRenderer.cs
@@ -69,7 +69,11 @@ namespace NLog.LayoutRenderers
         }
 
         /// <inheritdoc />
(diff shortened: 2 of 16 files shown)


304NotModified:
@snakefoot merge or no merge for this one. What do you think?

snakefoot:
Well if you want the WhenEmpty to work, then you need something like this.

I guess I would change SimpleLayout, so when one rawvalue-lookup fails then it will stop trying to use the rawvalue-lookup-interface (set it to null).

304NotModified:
> Well if you want the WhenEmpty to work, then you need something like this.

Yes, I like that. (in the future)



> I guess I would change SimpleLayout, so when one rawvalue-lookup fails then it will stop trying to use the rawvalue-lookup-interface (set it to null).

Could you elaborate on that (or PR?) Thanks!

snakefoot:
> Could you elaborate on that (or PR?) Thanks!

You can just look at how `_stringValueRenderer` is assigned null. Same could be done for `_rawValueRenderer` when it fails (returns false).

304NotModified:
well, next time (e.g. with the ${when}, ) it could return true. See #3140

snakefoot:
That will only happen if the Layout suddenly contains a different layoutrenderer. Funky wonky logic is not good.

304NotModified:
no it doesn't? one part of the ${when} has a irawvalue-layout and the other a non-irawvalue layout. e.g. the example in the PR: `${when:${db-null}:when=logger=='logger':else=better}`

snakefoot:
Should only work when both layouts supports rawvalue


Sent from my Sony Xperia


no it doesn't? one part of the ${when} has a irawvalue-layout and the other a non-irawvalue layout. e.g. the example in the PR: ${when:${db-null}:when=logger=='logger':else=better}

snakefoot:
It is going to be strange logic with overhead, because you will be executing the when-clause multiple times if the resulting layout doesn't support rawvalue.

But as long you make sure that the rawvalue-probe-logic (casting and double-checking when-condition) only happens when having explicitly specified PropertyType / ParameterType / DbType different from string-type. Then can you add whatever you need. Hopefully this extra raw-code-path will have minimal impact on future code-maintenance.

Issue #3127, opened 2019-02-18
Incorrect ${logger:shortName=true} with Generic logger names


anton-gogolev:
Looking at the [code](https://github.com/NLog/NLog/blob/dev/src/NLog/LayoutRenderers/LoggerNameLayoutRenderer.cs#L61), this applies to pretty much any NLog version out there.

Imagine:

    public class FileConfigurationService<T> : 
        IFileConfigurationService<T>
        where T : class
    {
        private static readonly Logger Log =
            LogManager.GetLogger(typeof(FileConfigurationService<T>).FullName);

The `LogEventInfo.LoggerName` will now be equal to, e.g. 

> REDACTED.Platform.Impl.Configuration.Impl.FileConfigurationService`1[[REDACTED.Platform.Shared.ObjectModel.Configuration.HotelServiceOnlineCancellationConfiguration, REDACTED.Platform.Shared.ObjectModel, Version=3.19.7.3, Culture=neutral, PublicKeyToken=null]]

Which in turns makes `${logger:shortName=true}` render to

> 3, Culture=neutral, PublicKeyToken=null]]

That's even remotely not correct.

Should I tackle this with a PR?

304NotModified:
Not sure, but isn't the issue in `typeof(FileConfigurationService<T>).FullName`? 

GetLogger won't alter the name ;)

anton-gogolev:
If we dropped generics altogether:

public class FileConfigurationService
{
    private static readonly Logger Log =
        LogManager.GetLogger(typeof(FileConfigurationService).FullName);

(or even replace `GetLogger()` with `GetCurrentClassLogger()`), that would produce a perfectly reasonable (albeit long-ish) `LoggerName` of "REDACTED.Platform.Impl.Configuration.Impl.FileConfigurationService". This is something that NLog users are already well used to, but somehow classes being generic break this behavior.

snakefoot:
The problem is the Assembly-strong-name that is sometimes are included with FullName. What if you change to Type.ToString?

```c#
public class FileConfigurationService<T> : 
    IFileConfigurationService<T>
    where T : class
{
    private static readonly Logger Log =
        LogManager.GetLogger(typeof(FileConfigurationService<T>).ToString());
```

And what happens if you use `GetCurrentClassLogger` ?

```c#
public class FileConfigurationService<T> : 
    IFileConfigurationService<T>
    where T : class
{
    private static readonly Logger Log = LogManager.GetCurrentClassLogger();
```

[...]

PR #3157, opened 2019-03-05
Callsite class-name lookup should not include assembly-info

snakefoot:
Type.ToString() instead of Type.FullName. See also #3127


304NotModified:
this is hard to unit test (without mock), isn't?

snakefoot:
I'm trying to create a unit test. But actually the type-lookup when using StackFrame is rather limited.

snakefoot:
Think I will close this issue. Since generic-type-args are hard to extract from StackFrame, then there will not be an issue with assembly-info from type-args. Looking at example from #3127 then it is the generic-args that carries the assembly-info, and when not available on StackFrame, then they are not included in LoggerName.

Closing the issue until other evidence appears.

PR #3160, opened 2019-03-09
Refactor filetarget + target to reduce code complexity

snakefoot:


304NotModified:
Have you used a refactor tool for this? (So no need to review in depth)

snakefoot:
FileTarget.cs and Target.cs changes was done by hand.

304NotModified:
thanks!

I think it would be nice if we could move out some methods out of the FileTarget in the future. It's really too large.

snakefoot:
Yes the file archive logic would probably be a good candidate.

PR #3162, opened 2019-03-09
ContextProperty with IncludeEmptyValue means default value for ValueType

snakefoot:
Convert empty-string and null-value to default value when ValueType with IncludeEmptyValue=true


snakefoot:
@304NotModified Please include for NLog 4.6

304NotModified:
👍

PR #3165, opened 2019-03-10
Database target: dropped UseRawValue option on parameters, introduced ${myRenderer:noRawValue=true}

snakefoot:
NoRawValueLayoutRendererWrapper can be used by "many".

After seeing the trouble in making IRawValue work with wrappers. Then why not make a wrapper where the purpose is to disable the IRawValue-interface.

Another unreleased breaking change.

### DIFF
--- a/src/NLog/LayoutRenderers/Wrappers/NoRawValueLayoutRendererWrapper.cs
+++ b/src/NLog/LayoutRenderers/Wrappers/NoRawValueLayoutRendererWrapper.cs
@@ -1,0 +1,72 @@
+﻿// 
+// Copyright (c) 2004-2019 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without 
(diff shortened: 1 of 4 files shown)


snakefoot:
Guess I should modify the DatabaseTargetTests so the logger includes assignment of DbType and escapes string-value with string-quotes.

304NotModified:
> See the difference between "2" and ""2""

I'm a bit confused by the `""2""` 👼 

Good idea! Now we "lose" useRawValue=true with `string`, but I guess that's a good thing.

snakefoot:
> I'm a bit confused by the ""2""

Could also say see the difference between `2` (non-string) and `"2"` (string) in the MockDb-Log-Verification.

304NotModified:
> Could also say see the difference between `2` (non-string) and `"2"` (string) in the MockDb-Log-Verification.

I missed it was a verbatim string ;)

304NotModified:
could you please help me with the docs? I will release RC3 then

snakefoot:
Created https://github.com/NLog/NLog/wiki/NoRawValue-layout-renderer

304NotModified:
thanks!

PR #3187, opened 2019-03-19
AsyncTaskTarget - Fixed unwanted delay caused by slow writer

### LINKED ISSUE #3185: AsyncTaskTarget does not batch rows after first iteration
I'm trying to write a custom async target for NLog that pushes logs to Azure, and ran into an unexpected scenario. Basically this: when the async task that I am invoking is slow (in this example 1.5s), after the first batch, all remaining log entries are sent in single batches.  

This is reproducable with various values for TaskDelayMilliseconds, and various app delays. I would expect one "Write" operation every "TaskDelayMilliseconds" except in cases of exceptions / retry. I suspect this may be a bug in the AsyncTaskTarget, but I'd like a community review before I submit it. Any thoughts? Thanks in advance.

After further tracing, I suspect that this could be based on the dual role of `TaskStartNext` in that it cleans up '_previousTask', from TaskCompletion, and also works to empty the queue.  I hesitate to recommend changes because I'm not familiar enough with async, or the patterns here, but perhaps the `TaskCompletion` should kick off a new timer if the queue is not empty.

StackOverflow link:
https://stackoverflow.com/questions/55228585/nlog-asynctasktarget-does-not-batch-rows-after-first-iteration?noredirect=1#comment97196350_55228585

This is reproducible against 4.6.0-rc3, and I am running on .Net Framework, 4.7.2

```xml
<nlog internalLogLevel="Debug" internalLogToConsole="true">
    <extensions>
      <add assembly="ConsoleApp1"/>
    </extensions>

    <targets>
      <target type="console" name="console"/>
      <target type="AsyncTestTarget" name="test" />
    </targets>

    <rules>
      <logger name="*" minlevel="Trace" writeTo="console" />
      <logger name="*" minlevel="Trace" writeTo="test" />
    </rules>
</nlog>
```

Sample Consumer:
```
class Program
{
    static ILogger logger = LogManager.GetCurrentClassLogger();
    {

        for (var i = 0; i < 5000; i++)
        {
            logger.Info(i.ToString());
            System.Threading.Thread.Sleep(TimeSpan.FromMilliseconds(100));
        }

        logger.Info("Done");

        LogManager.Flush(TimeSpan.FromSeconds(15));
    }
}
```

Sample (buggy) Target:
```
[Target("AsyncTestTarget")]
public class AsyncTestTarget : AsyncTaskTarget
{
    {
        base.InitializeTarget();

        BatchSize = 10000;
        TaskDelayMilliseconds = 3000;
    }


    protected override Task WriteAsyncTask(LogEventInfo logEvent, CancellationToken cancellationToken)
    {
        InternalLogger.Info($"NLog.Extensions.AsyncTestTarget.WriteAsyncTask()");
        return WriteAsyncTask(new List<LogEventInfo> {logEvent}, cancellationToken);
    }

    protected override Task WriteAsyncTask(IList<LogEventInfo> logEvents, CancellationToken cancellationToken)
    {
        InternalLogger.Info($"NLog.Extensions.AsyncTestTarget.WriteAsyncTask(List) : {logEvents.Count}");
        if (!logEvents.Any())
            return Task.CompletedTask;
        // If this line is commented, the rows are batched as expected.  If this line is left as-is, then after the first batch, records are sent one at a time.
        System.Threading.Thread.Sleep(TimeSpan.FromMilliseconds(1500));

        return Task.CompletedTask;
    }
}
```

Current Result:
Records are batched initially, but seem to immediately transition to a sequential (non-batched) WriteAsyncTask implementation.  See the screenshot below.
Broken Result

Expected Result:
Records are batched consistently.  See the screenshot with the delay removed.
Expected Result

snakefoot:
Resolves #3185

### DIFF
--- a/src/NLog/Targets/AsyncTaskTarget.cs
+++ b/src/NLog/Targets/AsyncTaskTarget.cs
@@ -140,8 +140,6 @@ namespace NLog.Targets
             _taskCompletion = TaskCompletion;
             _taskCancelledToken = TaskCancelledToken;
             _taskTimeoutTimer = new Timer(TaskTimeout, null, Timeout.Infinite, Timeout.Infinite);
-            _cancelTokenSource = new CancellationTokenSource();
-            _cancelTokenSource.Token.Register(_taskCancelledToken);
 
 #if NETSTANDARD2_0
             // NetStandard20 includes many optimizations for ConcurrentQueue:
@@ -161,6 +159,9 @@ namespace NLog.Targets
         /// </summary>
         protected override void InitializeTarget()
         {
+            _cancelTokenSource = new CancellationTokenSource();
+            _cancelTokenSource.Token.Register(_taskCancelledToken);
+
             base.InitializeTarget();
 
             if (BatchSize <= 0)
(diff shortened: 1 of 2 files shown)


304NotModified:
thanks, will be 4.6 RTM, most likely released tonight :)

Issue #3197, opened 2019-03-25
MdlcLayoutRenderer not marked with ThreadAgnostic


SergiyBondarInvisible:
Hi! 
Any reason this class is not marked with ThreadAgnostic?
The Mapped Diagnostic Logical Context is async but without attribute ThreadAgnostic for class MdlcLayoutRenderer used as sync - as I understood.
Similar problem was here - https://github.com/NLog/NLog/issues/893

snakefoot:
The idea with ThreadAgnostic is that the output will be the same independent of what thread is logging.

The AsyncLocal is not the same for all threads. Just like ThreadId is not the same for all threads.

This means `MdlcLayoutRenderer` has to capture the AsyncLocal-state when trying to defer the actually log-writing to an async Timer-thread. Because the Timer-thread always sees an empty AsyncLocal-state.

Guess the work done with #1956 is not enough.

304NotModified:
> Guess the work done with #1956 is not enough.

Please let me know what to add, or PR ;)

SergiyBondarInvisible:
Few words about my problem - when my web app have started to response with delay about few minutes, I created dump from w3wp.exe and found that very huge scope of threads delayed on NLog.dll!NLog.Targets.Target.PrecalculateVolatileLayouts. So I have started to investigate and found that all layouts used attribute ThreadAgnostic, but MdlcLayoutRenderer not used it. In method PrecalculateVolatileLayouts if not all layouts with atribute ThreadAgnostic then used logic with lock - so we have bottleneck.

snakefoot:
Can you paste your Layout config?


NLog should not hold locks as long the layoutrenderers are marked as threadsafe.


Threadsafe was introduced with NLog 4.5.3 and NLog.Web.AspNetCore 4.7.1



Sent from my Sony Xperia


---- Sergiy Bondar2 wrote ----


Few words about my problem - when my web app have started to response with delay about few minutes, I created dump from w3wp.exe and found that very huge scope of threads delayed on NLog.dll!NLog.Targets.Target.PrecalculateVolatileLayouts. So I have started to investigate and found that all layouts used attribute ThreadAgnostic, but MdlcLayoutRenderer not used it. In method PrecalculateVolatileLayouts if not all layouts with atribute ThreadAgnostic then used logic with lock - so we have bottleneck.

SergiyBondarInvisible:
I used NLog 4.5.11

<target xsi:type="File"
            name="File"
[...]

PR #3201, opened 2019-03-26
${GDC} reading is now lockfree

### LINKED ISSUE #2960: NLog variables are not threadsafe
@snakefoot I see your recommendation for  gdc over var,

e.g.

> If needing to modify global variables at runtime, then it is recommended to use the GDC layout renderer instead. Because NLog variables are not threadsafe, and should be seen as readonly variables. Also GDC will behave correctly if LoggingConfiguration should change (or reloaded).

Is there something we could do to make the current nlog variables threadsafe? A lot of users find the variables more natural to use.

snakefoot:
GlobalDiagnosticsContext - Use copy-on-write and lockfree read

See also #3197 

Noticed that NLog 4.5.11 often showed threads in progress with GDC-Lookups. On NLog 4.6 one didn't see it because GDC is marked as ThreadAgnostic.

But if using GDC in a Layout with other LayoutRenderers that are not ThreadAgnostic, then the lock-punishment will be back.

Added GDC-Lookup to the file-target-layout and now they get stuck with NLog 4.6:
image

### DIFF
--- a/src/NLog/GlobalDiagnosticsContext.cs
+++ b/src/NLog/GlobalDiagnosticsContext.cs
@@ -65,7 +65,8 @@ namespace NLog
         {
             lock (_dict)
             {
-                GetWritableDict(_dictReadOnly != null && !_dict.ContainsKey(item))[item] = value;
+                bool requireCopyOnWrite = _dictReadOnly != null && !_dict.ContainsKey(item); // Overwiting existing value is ok (no resize)
+                GetWritableDict(requireCopyOnWrite)[item] = value;
             }
         }
 
@@ -130,7 +131,8 @@ namespace NLog
         {
             lock (_dict)
             {
-                GetWritableDict(_dictReadOnly != null && _dict.ContainsKey(item)).Remove(item);
+                bool requireCopyOnWrite = _dictReadOnly != null && _dict.ContainsKey(item);
+                GetWritableDict(requireCopyOnWrite).Remove(item);
             }
         }
 
@@ -141,7 +143,8 @@ namespace NLog
         {
             lock (_dict)
             {
-                GetWritableDict(_dictReadOnly != null && _dict.Count > 0, true).Clear();
+                bool requireCopyOnWrite = _dictReadOnly != null && _dict.Count > 0;
+                GetWritableDict(requireCopyOnWrite, true).Clear();
             }
         }
 
@@ -158,9 +161,9 @@ namespace NLog
             return readOnly;
         }
 
-        private static Dictionary<string, object> GetWritableDict(bool mustCreateNew, bool clearDictionary = false)
+        private static Dictionary<string, object> GetWritableDict(bool requireCopyOnWrite, bool clearDictionary = false)
         {
-            if (mustCreateNew)
+            if (requireCopyOnWrite)
             {
                 Dictionary<string, object> newDict = CopyDictionaryOnWrite(clearDictionary);
                 _dict = newDict;
--- a/src/NLog/GlobalDiagnosticsContext.cs
+++ b/src/NLog/GlobalDiagnosticsContext.cs
@@ -53,10 +53,7 @@ namespace NLog
         /// <param name="value">Item value.</param>
         public static void Set(string item, string value)
         {
-            lock (_dict)
-            {
-                GetWritableDict(_dictReadOnly != null && !_dict.ContainsKey(item))[item] = value;
-            }
+            Set(item, (object)value);
         }
 
         /// <summary>
--- a/src/NLog/GlobalDiagnosticsContext.cs
+++ b/src/NLog/GlobalDiagnosticsContext.cs
@@ -141,7 +141,7 @@ namespace NLog
         {
             lock (_dict)
             {
-                GetWritableDict(_dictReadOnly != null && _dict.Count > 0, false).Clear();
+                GetWritableDict(_dictReadOnly != null && _dict.Count > 0, true).Clear();
             }
         }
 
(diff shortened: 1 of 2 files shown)


snakefoot:
ConcurrentDictionary doesn't make the same "promise" of ordering as normal Dictionary. So JsonLayout.IncludeGdc is not producing identical output.

snakefoot:
The ordering is important in the expanding of variables that reference other variables. So reverted back to only using ConcurrentDictionary for GlobalDiagnosticsContext

304NotModified:
Isn't the concurrent queue much slower then the lock?

snakefoot:
> Isn't the concurrent queue much slower then the lock?

Not if you have two threads.

304NotModified:
Not sure it's coincidence this tests fails now :

NLog.UnitTests.Targets.ConcurrentFileTargetTests.SimpleConcurrentTest(numProcesses: 2, numLogs: 500, mode: "none|archive") [FAIL]
At C:\projects\nlog\run-tests.ps1:33 char:1
+ & ${env:xunit20}\xunit.console.x86.exe .\tests\NLog.UnitTests\bin\rel ...

From 
https://ci.appveyor.com/project/nlog/nlog/builds/23379399

tests/NLog.UnitTests/Layouts/JsonLayoutTests.cs
```diff
@@ -506,7 +506,7 @@ public void IncludeGdcJsonProperties()
             <targets>
                 <target name='asyncDebug' type='AsyncWrapper' timeToSleepBetweenBatches='0'>
                 <target name='debug' type='Debug'  >
-                 <layout type=""JsonLayout"" IncludeGdc='true' ExcludeProperties='Excluded1,Excluded2'>
+                 <layout type=""JsonLayout"" IncludeGdc='true' IncludeAllProperties='true' ExcludeProperties='Excluded1,Excluded2'>
```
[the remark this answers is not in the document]
snakefoot: Dictionary makes a very good promise to return items in the same order as inserted (Though not always). ConcurrentDictionary doesn't make this promise.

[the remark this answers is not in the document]
snakefoot: I have modified the test so GDC only contains the first property, and rest of the properties comes from the LogEvent.

snakefoot:
> Not sure it's coincidence this tests fails now :

Always been an unstable test. Concurrent processes writing to the same file test is not affected by this change.

snakefoot:
I think the trouble and the surprises with ConcurrentDictionary will outweight the gain in a truly multithreaded environment.

Even using readerwriterlockslim will probably have some side-effects:

https://blogs.msdn.microsoft.com/pedram/2007/10/07/a-performance-comparison-of-readerwriterlockslim-with-readerwriterlock/

snakefoot:
@304NotModified Now modified GDC to perform clone on write. This gives high concurrency for loggers but more allocation for writers (if they write a lot while others are reading).

304NotModified:
@snakefoot is this one safe for NLog 4.6.1? (So low impact)?

snakefoot:
I think it is safe for NLog 4.6.1. One could also consider using the same approach for NLog Configuration Variables.

src/NLog/GlobalDiagnosticsContext.cs
```diff
@@ -152,10 +142,40 @@ public static void Remove(string item)
...
+            }
+            return readOnly;
+        }
+
+        private static Dictionary<string, object> GetWritableDict(bool mustCreateNew, bool clone = true)
+        {
+            if (mustCreateNew)
             {
-                dict.Clear();
+                var newDict = new Dictionary<string, object>(clone ? _dict.Count + 1 : 0);
```
304NotModified: Is it correct that clear has the init capacity of 0, and the field has the default capacity (16?). It that on purpose?

304NotModified:
please check https://github.com/snakefoot/NLog/pull/5 :)

src/NLog/GlobalDiagnosticsContext.cs
```diff
@@ -69,7 +68,7 @@ public static void Set(string item, object value)
         {
             lock (dict)
             {
-                dict[item] = value;
+                GetWritableDict(dictReadOnly != null && !dict.ContainsKey(item))[item] = value;
```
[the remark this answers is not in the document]
snakefoot: Yes one could probably make one of the Setters call the other.

src/NLog/GlobalDiagnosticsContext.cs
snakefoot: Don't understand your question. But Clear (clone=false) calls with capacity = 0 that means do not allocate internal arrays until insert happens.

304NotModified: indeed, but the init doesn't set it to `0`

`private static Dictionary<string, object> _dict = new Dictionary<string, object>();`

snakefoot: Still don't understand your queustion. The default constructor for Dictionary uses capacity = 0. See also https://referencesource.microsoft.com/#mscorlib/system/collections/generic/dictionary.cs,85

snakefoot:
@304NotModified Have now extracted the clone-logic from GetWritableDict

src/NLog/GlobalDiagnosticsContext.cs
```diff
@@ -158,21 +158,27 @@ public static void Clear()
             return readOnly;
         }
 
-        private static Dictionary<string, object> GetWritableDict(bool mustCreateNew, bool clone = true)
+        private static Dictionary<string, object> GetWritableDict(bool mustCreateNew, bool clearDictionary = false)
```
304NotModified: This is a strange methode, with mustCreateNew and clear IMO. I tried to refactor that ;)

snakefoot: It is very identical to MDC + MDLC - GetLogicalThreadDictionary

304NotModified: well, clearDictionary  is only called once with `true`. Then it's strange to have that encapsulated in a method. It makes things a bit unnecessary complex.

snakefoot: Yes I'm terrible at making code complex. But trying to improve, though I often forget. But I think https://github.com/snakefoot/NLog/pull/5 now is showing boiler-plate code for the different write-operations.

snakefoot: Btw. the clearDictionary is an optimization-flag, so it knows it can skip the copying of the old dictionary.

304NotModified: > showing boiler-plate code for the different write-operations.

true, set, remove and clear has code duplication, but that was already the case (in 1 line) and now those are 2 lines of duplication. Against it, CloneDictionary is much more straight-forward.

304NotModified: Also GetWritableDict wasn't doing one thing, it was doing "get" but also "set" (but OK, that's also the case in MDC probably, but im trying to "fix the leak")

snakefoot: With copy-on-write semantics then get-operation will sometimes be create/set. That is the nature of copy-on-write.

304NotModified: from another view, do you think this line is easy to maintain?

`  GetWritableDict(_dictReadOnly != null && !_dict.ContainsKey(item))[item] = value;`

304NotModified: > Yes I'm terrible good at making complex code.

I focus on having maintainable code, and unfortunately that's not always performant code :(  I think the best code for NLog is in-between :)

snakefoot: Your points are very valid and valuable. Have now made another code-update for readability.

snakefoot:
Happy to do it. And also very happy about the result :)

PR #3253, opened 2019-03-28
Fix wrong warnings on <nlog> element + refactoring

### LINKED ISSUE #3251: "Skipping unknown 'NLog' property" after update to 4.6
Hi,

I have my NLog configs in separate files, normally named nlog.config. When an application starts, i load the config by calling LogManager.LoadConfiguration("nlog.config");

Here's a typical nlog.config:

```
<?xml version="1.0" encoding="utf-8"?>
<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" internalLogToConsole="true" internalLogLevel="Warn" autoReload="true"> 
    <!-- Actual configuration -->    
</nlog>
```
After updating to 4.6 i started seeing messages like these in the console upon an application start:

NLog: 2019-03-28 07:32:36.0827 Warn Skipping unknown 'NLog' property internalLogLevel=Warn
NLog: 2019-03-28 07:32:36.1380 Warn Skipping unknown 'NLog' property autoReload=true
NLog: 2019-03-28 07:29:05.0533 Warn Skipping unknown 'NLog' property xmlns=http://www.nlog-project.org/schemas/NLog.xsd
NLog: 2019-03-28 07:29:05.0533 Warn Skipping unknown 'NLog' property xsi=http://www.w3.org/2001/XMLSchema-instance

So, either there was some breaking change in 4.6 which i cannot find any info about or there is a bug. 

Any help is appreciated,
-D

snakefoot:
Resolves #3251

### DIFF
--- a/src/NLog/Config/LoggingConfigurationParser.cs
+++ b/src/NLog/Config/LoggingConfigurationParser.cs
@@ -74,28 +74,7 @@ namespace NLog.Config
             InternalLogger.Trace("ParseNLogConfig");
             nlogConfig.AssertName("nlog");
 
-            var dict = CreateNLogConfigDictionary(nlogConfig);
-
-            //check first exception throwing and internal logging, so that erros in this section could be handled correctly
-            SetThrowExceptions(dict);
-            SetThrowConfigExceptions(dict);
-            var internalLoggerEnabled = SetInternalLogLevel(dict);
-
-            SetNLogElementSettings(dict, out var parseMessageTemplates, out var internalLogFile);
-
-            if (internalLogFile != null)
-            {
-                internalLogFile = ExpandFilePathVariables(internalLogFile);
-                InternalLogger.LogFile = internalLogFile;
-            }
-
-            if (!internalLoggerEnabled && !InternalLogger.HasActiveLoggers())
-            {
-                InternalLogger.LogLevel = LogLevel.Off; // Reduce overhead of the InternalLogger when not configured
-            }
-
-            _configurationItemFactory = ConfigurationItemFactory.Default;
-            _configurationItemFactory.ParseMessageTemplates = parseMessageTemplates;
+            SetNLogElementSettings(nlogConfig);
 
             var children = nlogConfig.Children.ToList();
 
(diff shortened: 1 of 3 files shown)


src/NLog/Config/LoggingConfigurationParser.cs
```diff
@@ -132,14 +111,30 @@ protected void LoadConfig(ILoggingConfigurationElement nlogConfig, string basePa
...
         {
-            parseMessageTemplates = null;
-            internalLogFile = null;
+            var dict = CreateNLogConfigDictionary(nlogConfig);
+
+            bool? parseMessageTemplates = null;
+            bool internalLoggerEnabled = false;
             foreach (var configItem in dict)
             {
                 switch (configItem.Key.ToUpperInvariant())
```
304NotModified: PS, I don;t think this is needed because of the `StringComparer.OrdinalIgnoreCase` for the dictionary

snakefoot: It is used for the switch-statement. And also it returns a list, not a dict to promise ordering. Will fix the variable name.

snakefoot: Fixed the name

304NotModified:
While I like refactorings, this change is hard to review that no new bugs are introduced. But I you think this is better than #3254, I'm OK with taking this one.

src/NLog/Config/LoggingConfigurationParser.cs
```diff
@@ -184,78 +184,61 @@ private void SetNLogElementSettings(Dictionary<string, string> dict, out bool? p
...
             {
-                if (configItem.Key != null)
+                if (!string.IsNullOrEmpty(configItem.Key))
                 {
                     dict[configItem.Key.Trim()] = configItem.Value;
                 }
             }
 
-            return dict;
+            var sortedDict = new List<KeyValuePair<string, string>>(dict.Count);
```
304NotModified: List instead of dict?

src/NLog/Config/LoggingConfigurationParser.cs
```diff
@@ -184,78 +184,61 @@ private void SetNLogElementSettings(Dictionary<string, string> dict, out bool? p
...
+            AddHighPrioritySetting("InternalLogFile", dict, sortedDict);
+            AddHighPrioritySetting("InternalLogToConsole", dict, sortedDict);
+            foreach (var configItem in dict)
+            {
+                sortedDict.Add(configItem);
+            }
+            return sortedDict;
+        }
+
+        private static void AddHighPrioritySetting(string settingName, Dictionary<string, string> dict, List<KeyValuePair<string, string>> sortedDict)
```
304NotModified: Could we setup a type that encapsulates the list and dictionary? That type will be responsible for keeping things in sync.

src/NLog/Config/LoggingConfigurationParser.cs
snakefoot: The order is predictable for List, not always for Dictionary.

src/NLog/Config/LoggingConfigurationParser.cs
snakefoot: Not sure I understand the dictionary is temporary and is thrown away. The List is the final result.

src/NLog/Config/LoggingConfigurationParser.cs
304NotModified: I mean the name of the variable ;)

src/NLog/Config/LoggingConfigurationParser.cs
304NotModified: Hard to explain. I could refactor it when it's done (as PR)

src/NLog/Config/LoggingConfigurationParser.cs
snakefoot: Fixed

src/NLog/Config/LoggingConfigurationParser.cs
304NotModified: moved it to an inner method. To encapsulate the AddHighPrioritySetting. That's good enough for now.

(we need some refactor change in the future, as it's strange that we build up a dictionary, change it to list and then use a switch. I think the loop over the switch should be removed and replaced with direct dictionary access (so you can set the ordering easier)

snakefoot: I read your statement `type that encapsulates the list and dictionary` as creating some weird new container. But yes inner-method helps here, though they ruin readability especially on organic growth beyond 3-4 lines. I guess a better solution would be a Linq-expression or yield that returns IEnumerable.

304NotModified: > as creating some weird new container.

True, therefor I also skipped that. 

I think passing parameters to a method which should be in-sync (sorted list - dictionary) is a code smell and I fixed that now with the inner method. Inner methods should be avoided, but for this case it's perfect IMO (or at least, better then the previous situation and the wierd container). 

Another option was to create a new enumerator, but that's overkill for one situation I guess.

PR #3254, opened 2019-03-28
Fix wrong warnings on <nlog> element

304NotModified:
@snakefoot

I have somebit the same solution 

But both solutions has the same issue:

fixes

```xml
<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="somewhere" xsi:type="asa"
      internalLogToConsole="true" internalLogLevel="Warn">
</nlog>

```

but won't fix

```xml
<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xs2="http://www.w3.org/2001/XMLSchema-instance" xs2:schemaLocation="somewhere" xs2:type="asa"
      internalLogToConsole="true" internalLogLevel="Warn">
</nlog>
```

I'm thinking to fully drop the warning on the `<nlog>` element...

### DIFF
--- a/src/NLog/Config/LoggingConfigurationParser.cs
+++ b/src/NLog/Config/LoggingConfigurationParser.cs
@@ -184,6 +184,16 @@ namespace NLog.Config
                             ? (bool?)null
                             : ParseBooleanValue(configItem.Key, configItem.Value, true);
                         break;
+
+                    case "INTERNALLOGLEVEL":
+                        // skip, read later
+                        break;
+                    case "XMLNS":
+                    case "XSI":
+                    case "SCHEMALOCATION":
+                    case "TYPE":
+                        // We could ignore these
+                        break;
                     default:
                         InternalLogger.Warn("Skipping unknown 'NLog' property {0}={1}", configItem.Key, configItem.Value);
                         break;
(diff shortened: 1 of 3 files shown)


snakefoot:
My solution includes reduced code-complexity. And reduces the noise to info-level

304NotModified:
I think we should ignore the attribute (in NlogXmlElement), if it is prefixed by `xmlns:`. E.g. `xmlns:xsi`, `xmlns:xs2` etc.

304NotModified:
@snakefoot you could cherry-pick 288313acedabc1813b4700c397d4f0423bdebfdc

304NotModified:
> . And reduces the noise to info-level

I think it should be warn

snakefoot:
> I think it should be warn

Well it is new validation logic with NLog 4.6 and it seems to trigger to often.

snakefoot:
Really like the idea of using NLogXmlElement.cs to hide xml-specific tags. Good thinking.

304NotModified:
Thanks, indeed there could be more moved into the NLogXmlElement

> > I think it should be warn
> 
> Well it is new validation logic with NLog 4.6 and it seems to trigger to often.

Debug is also fine to me. I try to use Info for functional parts on a higher level (eg config file loaded)

snakefoot:
> Debug is also fine to me

Debug it becomes

304NotModified:
supersed by #3253

Issue #3200, opened 2019-03-26
EventSource Error (Incorrect Function) during shutdown w/ debugger


dwarburt:
**NLog version**: 4.6.0

**Platform**: .NET Core 2

**Current NLog config**

```csharp
            var config = new LoggingConfiguration();
            var consoleTarget = new ColoredConsoleTarget();
            config.AddTarget("console", consoleTarget);
            config.LoggingRules.Add(new LoggingRule("*", LogLevel.Debug, consoleTarget));
            LogManager.Configuration = config;
            LogManager.GetCurrentClassLogger().Info("message");
```

- What is the current result?
Error messages printed to console
- What is the expected result?
Not to print this.
- Did you checked the [Internal log](https://github.com/NLog/NLog/wiki/Internal-Logging)?
Yes, internal logging shows some messages before and after these messages when it's enabled.

- Please post full exception details (message, stacktrace, inner exceptions)
When run under the debugger in vscode, when the program exits it prints `EventSource Error: ERROR: Exception during construction of EventSource System.Threading.Tasks.TplEventSource: Incorrect function` to the console. With https://github.com/aws/aws-logging-dotnet installed and configured about a dozen similar messages are printed.

Example without internal logging

![image](https://user-images.githubusercontent.com/52614/55013539-ae1d1480-4fbf-11e9-844d-94f3acdb82d3.png)

Example with internal logging

![image](https://user-images.githubusercontent.com/52614/55013589-ca20b600-4fbf-11e9-8a22-e4971b2ac9d2.png)

```
Loaded 'C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.2.3\System.Diagnostics.Debug.dll'. Skipped loading symbols. Module is optimized and the debugger option 'Just My Code' is enabled.
2019-03-26 12:05:36.8890 Debug Targets for nlog_incorrect_function.Program by level:
2019-03-26 12:05:36.8929 Debug Trace =>
2019-03-26 12:05:36.8929 Debug Debug => 
2019-03-26 12:05:36.8929 Debug Info => 
2019-03-26 12:05:36.8929 Debug Warn => 
2019-03-26 12:05:36.8929 Debug Error => 
2019-03-26 12:05:36.8929 Debug Fatal => 
Loaded 'C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.2.3\System.Threading.Thread.dll'. Skipped loading symbols. Module is optimized and the debugger option 'Just My Code' is enabled.
[...]

PR #3255, opened 2019-03-28
Auto-flush on process exit improvements

snakefoot:
AsyncHelpers.StartAsyncTask defaults to ThreadPool.QueueUserWorkItem 

See also #3200 Tasks doesn't flush well on process-exit-event. Reverting back to NLog 4.4 behavior.

### DIFF
--- a/src/NLog/Common/AsyncHelpers.cs
+++ b/src/NLog/Common/AsyncHelpers.cs
@@ -53,12 +53,13 @@ namespace NLog.Common
 #endif
         }
 
-        internal static void StartAsyncTask(Action<object> action, object state)
+        internal static void StartAsyncTask(AsyncHelpersTask asyncTask, object state)
         {
-#if NET4_0 || NET4_5 || NETSTANDARD
-            System.Threading.Tasks.Task.Factory.StartNew(action, state, CancellationToken.None, System.Threading.Tasks.TaskCreationOptions.None, System.Threading.Tasks.TaskScheduler.Default);
+            var asyncDelegate = asyncTask.AsyncDelegate;
+#if NETSTANDARD1_0
+            System.Threading.Tasks.Task.Factory.StartNew(asyncDelegate, state, CancellationToken.None, System.Threading.Tasks.TaskCreationOptions.None, System.Threading.Tasks.TaskScheduler.Default);
 #else
-            ThreadPool.QueueUserWorkItem(new WaitCallback(action), state);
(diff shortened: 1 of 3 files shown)


src/NLog/Common/AsyncHelpers.cs
```diff
@@ -379,5 +386,17 @@ internal static bool WaitForDispose(this Timer timer, TimeSpan timeout)
             }
             return true;
         }
+
+        /// <summary>
+        /// Forward declare of system delegate type for use by other classes
+        /// </summary>
+        internal struct AsyncTask
```
304NotModified: Proposal, unnest and move to internal.

snakefoot: Well it really belongs to AsyncHelpers.

304NotModified: It's now inconsistent when it's a class in the internal namespace and when it's nested.

snakefoot: Fixed

304NotModified:
thanks!

Is this for NLog 4.6.1?

snakefoot:
Yes it should be safe.

304NotModified:
@snakefoot I try to rename the PRs to a more functional title, so it's easier to use in the changelog. Did I rename this PR correct?

snakefoot:
@304NotModified I like to describe the technical change, when not super sure whether it is actually the cure :)

304NotModified:
I think "improved" was better then, as it's functional and it's not strict (it's fuzzy ;)) 

I don't think anyone cases if we use threadpools or tasks ;)

304NotModified:
> I think "improved" was better then, as it's functional and it's not strict (it's fuzzy ;))
> 
> I don't think anyone cases if we use threadpools or tasks ;)

Is it OK if I would rename it back? Don't wanna have a PR title war ;)

snakefoot:
You just rename it as you want :)


Sent from my Sony Xperia


I think "improved" was better then, as it's functional and it's not strict (it's fuzzy ;))

I don't think anyone cases if we use threadpools or tasks ;)

Is it OK if I would rename it back? Don't wanna have a PR title war ;)

304NotModified:
> You just rename it as you want :)

It was tempting to rename it to "Julian is the best!1" 😄

snakefoot:
Would be more descriptive and correct, than random improvement to ProcessExit-logic :)

PR #3261, opened 2019-03-30
NLog config file loading: use process name (e.g. applicationname.exe.nlog) when app.config is not available

### LINKED ISSUE #3156: Configuration in applicationname.exe.nlog files do not load
**NLog version**: 4.5.11
**Platform**:  .NET Core 2

```xml
      throwExceptions="true" throwConfigExceptions="true">

  <targets>
    <target name="file" xsi:type="File"
            layout="${longdate} ${level} ${logger}: ${message} ${exception:format=ToString}"
            fileName="LogFiles/${shortdate}_LogFile.txt" keepFileOpen="true" encoding="utf-8"/>

    <target name="console" xsi:type="ColoredConsole"
            layout="${longdate} ${logger} ${level}: ${message} ${exception:format=ToString}"
            detectConsoleAvailable="true" />

  <rules>
    <logger name="*" minlevel="Debug" writeTo="file,console,viewer" />
  </rules>
</nlog>
```

According to https://github.com/nlog/nlog/wiki/Configuration-file it should be possible to have a configuration in a file called `program.exe.nlog` where `program.exe` is the executable for my program.

When the config file is in the `applicationname.exe.nlog` file, it does not work as expected.

Create a console application with the following `Program.cs`

```csharp
    class Program
    {
        {
            Console.WriteLine("Hello World!");

            foreach (var target in LogManager.Configuration.AllTargets)
                Console.WriteLine("Target: {0}", target.Name);
        }
    }
```

Publish the program with `dotnet publish -c Release -r win-x64` and run the executable. The program crashes with `Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.`.

Rename the file to `nlog.config` and it works as expected.

snakefoot:
Trying to resolve #3156 for NetCore

### DIFF
--- a/src/NLog/Config/LoggingConfigurationFileLoader.cs
+++ b/src/NLog/Config/LoggingConfigurationFileLoader.cs
@@ -37,6 +37,7 @@ namespace NLog.Config
     using System.Collections.Generic;
     using System.IO;
     using System.Security;
+    using System.Xml;
     using NLog.Common;
     using NLog.Internal;
     using NLog.Internal.Fakeables;
@@ -51,23 +52,23 @@ namespace NLog.Config
     /// </summary>
     internal class LoggingConfigurationFileLoader : ILoggingConfigurationLoader
     {
-        private static readonly FileWrapper DefaultFileWrapper = new FileWrapper();
-        private readonly IFile _file;
+        private static readonly AppEnvironmentWrapper DefaultAppEnvironment = new AppEnvironmentWrapper();
+        private readonly IAppEnvironment _appEnvironment;
 
         public LoggingConfigurationFileLoader()
-            :this(DefaultFileWrapper)
+            :this(DefaultAppEnvironment)
         {
         }
 
-        public LoggingConfigurationFileLoader(IFile file)
+        public LoggingConfigurationFileLoader(IAppEnvironment appEnvironment)
         {
-            _file = file;
+            _appEnvironment = appEnvironment;
         }
 
         public LoggingConfiguration Load(LogFactory logFactory, string filename)
         {
             var configFile = GetConfigFile(filename);
-            return LoadXmlLoggingConfiguration(logFactory, configFile);
+            return LoadXmlLoggingConfigurationFile(logFactory, configFile);
         }
 
         public virtual LoggingConfiguration Load(LogFactory logFactory)
@@ -86,7 +87,7 @@ namespace NLog.Config
             {
                 foreach (var path in GetDefaultCandidateConfigFilePaths(configFile))
                 {
-                    if (_file.Exists(path))
+                    if (_appEnvironment.FileExists(path))
                     {
                         configFile = path;
                         break;
@@ -108,7 +109,11 @@ namespace NLog.Config
                 {
                     if (stream != null)
                     {
-                        return LoadXmlLoggingConfiguration(logFactory, XmlLoggingConfiguration.AssetsPrefix + nlogConfigFilename);
+                        InternalLogger.Debug("Loading config from Assets {0}", nlogConfigFilename);
+                        using (var xmlReader = XmlReader.Create(stream))
+                        {
+                            return LoadXmlLoggingConfiguration(xmlReader, null, logFactory);
+                        }
                     }
                 }
             }
@@ -142,15 +147,20 @@ namespace NLog.Config
             {
 #if SILVERLIGHT && !WINDOWS_PHONE
                 Uri configFileUri = new Uri(configFile, UriKind.Relative);
-                if (Application.GetResourceStream(configFileUri) != null)
+                var streamResourceInfo = Application.GetResourceStream(configFileUri);
+                if (streamResourceInfo != null)
                 {
-                    config = LoadXmlLoggingConfiguration(logFactory, configFile);
-                    return true;    // File exists, and maybe the config is valid, stop search
+                    InternalLogger.Debug("Loading config from Resource {0}", configFileUri);
+                    using (var xmlReader = XmlReader.Create(streamResourceInfo.Stream))
+                    {
+                        config = LoadXmlLoggingConfiguration(xmlReader, null, logFactory);
+                        return true;
+                    }
                 }
 #else
-                if (File.Exists(configFile))
+                if (_appEnvironment.FileExists(configFile))
                 {
-                    config = LoadXmlLoggingConfiguration(logFactory, configFile);
+                    config = LoadXmlLoggingConfigurationFile(logFactory, configFile);
                     return true;    // File exists, and maybe the config is valid, stop search
                 }
 #endif
@@ -178,10 +188,18 @@ namespace NLog.Config
             return false;   // No valid file found
         }
 
-        private LoggingConfiguration LoadXmlLoggingConfiguration(LogFactory logFactory, string configFile)
+        private LoggingConfiguration LoadXmlLoggingConfigurationFile(LogFactory logFactory, string configFile)
         {
             InternalLogger.Debug("Loading config from {0}", configFile);
-            var xmlConfig = new XmlLoggingConfiguration(configFile, logFactory);
+            using (var xmlReader = _appEnvironment.LoadXmlFile(configFile))
+            {
+                return LoadXmlLoggingConfiguration(xmlReader, configFile, logFactory);
+            }
+        }
+
+        private LoggingConfiguration LoadXmlLoggingConfiguration(XmlReader xmlReader, string configFile, LogFactory logFactory)
+        {
+            var xmlConfig = new XmlLoggingConfiguration(xmlReader, configFile, logFactory);
             //problem: XmlLoggingConfiguration.Initialize eats exception with invalid XML. ALso XmlLoggingConfiguration.Reload never returns null.
             //therefor we check the InitializeSucceeded property.
             if (xmlConfig.InitializeSucceeded != true)
@@ -204,7 +222,7 @@ namespace NLog.Config
         {
             // NLog.config from application directory
             string nlogConfigFile = fileName ?? "NLog.config";
-            string baseDirectory = PathHelpers.TrimDirectorySeparators(LogFactory.CurrentAppDomain?.BaseDirectory);
+            string baseDirectory = PathHelpers.TrimDirectorySeparators(_appEnvironment.AppDomainBaseDirectory);
             if (!string.IsNullOrEmpty(baseDirectory))
                 yield return Path.Combine(baseDirectory, nlogConfigFile);
 
@@ -214,17 +232,16 @@ namespace NLog.Config
                 yield return Path.Combine(baseDirectory, nLogConfigFileLowerCase);
 
 #if !SILVERLIGHT && !NETSTANDARD1_3
-            var entryAssemblyLocation = PathHelpers.TrimDirectorySeparators(AssemblyHelpers.GetAssemblyFileLocation(System.Reflection.Assembly.GetEntryAssembly()));
-            if (!string.IsNullOrEmpty(entryAssemblyLocation))
-            {
-                if (!string.Equals(entryAssemblyLocation, baseDirectory, StringComparison.OrdinalIgnoreCase))
-                {
-                    yield return Path.Combine(entryAssemblyLocation, nlogConfigFile);
-                    if (!platformFileSystemCaseInsensitive)
-                        yield return Path.Combine(entryAssemblyLocation, nLogConfigFileLowerCase);
-                }
-            }
+            string entryAssemblyLocation = PathHelpers.TrimDirectorySeparators(_appEnvironment.EntryAssemblyLocation);
+#else
+            string entryAssemblyLocation = string.Empty;
 #endif
+            if (!string.IsNullOrEmpty(entryAssemblyLocation) && !string.Equals(entryAssemblyLocation, baseDirectory, StringComparison.OrdinalIgnoreCase))
+            {
+                yield return Path.Combine(entryAssemblyLocation, nlogConfigFile);
+                if (!platformFileSystemCaseInsensitive)
+                    yield return Path.Combine(entryAssemblyLocation, nLogConfigFileLowerCase);
+            }
 
             if (string.IsNullOrEmpty(baseDirectory))
             {
@@ -235,35 +252,13 @@ namespace NLog.Config
 
             if (fileName == null)
             {
-                // Current config file with .config renamed to .nlog
-                string configurationFile = LogFactory.CurrentAppDomain?.ConfigurationFile;
-                if (!StringHelpers.IsNullOrWhiteSpace(configurationFile))
-                {
-                    yield return Path.ChangeExtension(configurationFile, ".nlog");
-
-                    // .nlog file based on the non-vshost version of the current config file
-                    const string vshostSubStr = ".vshost.";
-                    if (configurationFile.Contains(vshostSubStr))
-                    {
-                        yield return Path.ChangeExtension(configurationFile.Replace(vshostSubStr, "."), ".nlog");
-                    }
-                }
+                // Scan for process specific nlog-files
+                foreach (var filePath in GetAppSpecificNLogLocations(entryAssemblyLocation))
+                    yield return filePath;
             }
 
-            IEnumerable<string> privateBinPaths = LogFactory.CurrentAppDomain.PrivateBinPath;
-            if (privateBinPaths != null)
-            {
-                foreach (var privatePath in privateBinPaths)
-                {
-                    var path = PathHelpers.TrimDirectorySeparators(privatePath);
-                    if (!StringHelpers.IsNullOrWhiteSpace(path) && !string.Equals(path, baseDirectory, StringComparison.OrdinalIgnoreCase))
-                    {
-                        yield return Path.Combine(path, nlogConfigFile);
-                        if (!platformFileSystemCaseInsensitive)
-                            yield return Path.Combine(path, nLogConfigFileLowerCase);
-                    }
-                }
-            }
+            foreach (var filePath in GetPrivateBinPathNLogLocations(baseDirectory, nlogConfigFile, platformFileSystemCaseInsensitive ? nLogConfigFileLowerCase : string.Empty))
+                yield return filePath;
 
 #if !SILVERLIGHT && !NETSTANDARD1_0
             if (fileName == null)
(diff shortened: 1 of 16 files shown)


src/NLog/Internal/ProcessIDHelper.cs
```diff
@@ -31,55 +31,60 @@
...
         /// <summary>
         /// Gets current process name (excluding filename extension, if any).
         /// </summary>
-        public abstract string CurrentProcessBaseName { get; }
+        public string CurrentProcessBaseName => _currentProcessBaseName ?? (_currentProcessBaseName = string.IsNullOrEmpty(CurrentProcessFilePath) ? UnknownProcessName : Path.GetFileNameWithoutExtension(CurrentProcessFilePath));
+
+        /// <summary>
+        /// Initializes the ThreadIDHelper class.
+        /// </summary>
+        private static ProcessIDHelper CreateInstance()
```
304NotModified: 👍 much better. Please rename it to just "Create" (it's a factory method) - and what other to create ;)

src/NLog/Config/LoggingConfigurationFileLoader.cs
```diff
@@ -248,6 +248,28 @@ public IEnumerable<string> GetDefaultCandidateConfigFilePaths(string fileName)
...
+                {
+                    string processFilePath = ProcessIDHelper.Instance.CurrentProcessFilePath;
+                    string processDirectory = PathHelpers.TrimDirectorySeparators(Path.GetDirectoryName(processFilePath));
+                    if (!string.IsNullOrEmpty(entryAssemblyLocation) && !string.Equals(entryAssemblyLocation, processDirectory, StringComparison.OrdinalIgnoreCase))
+                    {
+                        // Handle dotnet-process loading .NET Core-assembly, or IIS-process loading website
+                        string assemblyFileName = Path.GetFileName(System.Reflection.Assembly.GetEntryAssembly().Location);
+                        if (!string.IsNullOrEmpty(assemblyFileName))
+                            yield return Path.Combine(entryAssemblyLocation, assemblyFileName + ".nlog");
+                        // Handle unpublished .NET Core Application
```
304NotModified: Maybe a bit nitpicking, but I try to use those terms consistent :)

src/NLog/Internal/ProcessIDHelper.cs
```diff
@@ -31,56 +31,61 @@
...
         /// <summary>
         /// Gets current process name (excluding filename extension, if any).
         /// </summary>
-        public abstract string CurrentProcessBaseName { get; }
+        public string CurrentProcessBaseName => _currentProcessBaseName ?? (_currentProcessBaseName = string.IsNullOrEmpty(CurrentProcessFilePath) ? UnknownProcessName : Path.GetFileNameWithoutExtension(CurrentProcessFilePath));
+
+        /// <summary>
+        /// Initializes the ThreadIDHelper class.
+        /// </summary>
+        private static ProcessIDHelper Create()
```
304NotModified: FYI: I think "instance" in redundant, and most "factory methods" are called Create - so renamed it

src/NLog/Config/LoggingConfigurationFileLoader.cs
```diff
@@ -248,6 +248,28 @@ public IEnumerable<string> GetDefaultCandidateConfigFilePaths(string fileName)
...
+                        assemblyFileName = Path.GetFileNameWithoutExtension(assemblyFileName);
+                        if (!string.IsNullOrEmpty(assemblyFileName))
+                            yield return Path.Combine(entryAssemblyLocation, assemblyFileName + ".exe.nlog");
+                    }
+                    else if (!string.IsNullOrEmpty(processFilePath))
+                    {
+                        yield return processFilePath + ".nlog";
+                    }
```
304NotModified: Could you please elaborate the cases that here are implemented? Don't we need an unit test for this?

snakefoot: Thought it was obvious as the other case is for running inside "dotnet.exe" then this case is for NOT running inside "dotnet.exe".

snakefoot: Not that skilled with this build dynamic assemblies/processes and then running them. Find them very ugly and they only seem to be available for NetFramework (Not NetCore).

snakefoot:
Actually think that `Win32ProcessIDHelper` could be thrown away. Think the original code was this (Causing overhead by allocating Process-object on every lookup.

> Process.GetCurrentProcess().Id

Without any caching. Causing it to be much slower than this code:

> NativeMethods.GetCurrentProcessId()

But later caching was added so there no overhead at all for either method. Could be nice to remove those Win32-native-methods.

src/NLog/Config/LoggingConfigurationFileLoader.cs
snakefoot: Potato potato :)

src/NLog/Config/LoggingConfigurationFileLoader.cs
304NotModified: > Could you please elaborate the cases that here are implemented? Don't we need an unit test for this?

I meant all the 3? (not counted) test cases

src/NLog/Internal/ProcessIDHelper.cs
snakefoot: Think the whole instance thing should be removed. But not sure about the reason for the Win32-native-call. Since the values are cached, then I don't understand remark about Win32 is "without the overhead".

src/NLog/Config/LoggingConfigurationFileLoader.cs
304NotModified: > Not that skilled with this build dynamic assemblies/processes and then running them. Find them very ugly and they only seem to be available for NetFramework (Not NetCore).

I think we need to mock them, but I forgot that the code in NLog is not very mockable :(

304NotModified:
> Could be nice to remove some of those Win32-native-methods.

Agree on that! Less code is better

src/NLog/Config/LoggingConfigurationFileLoader.cs
snakefoot: Will try and create new internal interface that merges ProcessIDHelper and IAppDomain. But not easy since IAppDomain and AppDomainWrapper are public. Then it should become easier to mock and exercise the config-lookup.

304NotModified: > Will try and create new internal interface that merges ProcessIDHelper and IAppDomain

Don't think that is needed. We need to inject them both, or a factory 

I can give it a try if you like

snakefoot: Well to me IAppDomain is the hosting environment-interface. Don't need a new interface for every property in the hosting environment.

snakefoot: > I can give it a try if you like

Would rather prefer you review and merge: https://github.com/NLog/NLog.Extensions.Logging/pull/263 :)

304NotModified: > Would rather prefer you review and merge: [NLog/NLog.Extensions.Logging#263](https://github.com/NLog/NLog.Extensions.Logging/pull/263) :)

done

So now: https://github.com/snakefoot/NLog/pull/7 :) (for the general idea)

snakefoot: Already working on my own implementation of this mock. Your time would be better spent on all the other spider-issues in the hack-n-slash dungeon called NLog :)

304NotModified: What a spider issue? ❓

snakefoot: Guess you have never played Diablo or other hack-n-slash dungeon-games. They usually have spiders. And spiders are good at making cobwebs. For NLog it means all the issues detected by the code-analyzers you have been so nice to enable.

snakefoot:
@snakefoot Have squashed the commits, and ensured correct rename operations of the files.

src/NLog/Internal/Fakeables/IAppEnvironment.cs
```diff
@@ -33,14 +33,28 @@
...
 namespace NLog.Internal.Fakeables
 {
+    using System.Collections.Generic;
+    using System.Xml;
+
     /// <summary>
     /// Abstract calls to File
     /// </summary>
-    internal interface IFile
+    internal interface IAppEnvironment
```
304NotModified: Why is this interface so fat? It's hard to reuse and doesn't follow the SRP. :/

snakefoot: Like you said it is an internal interface and it is the properties needed to mock the loading of NLog-config.

src/NLog/Config/LoggingConfigurationFileLoader.cs
```diff
@@ -286,6 +338,7 @@ protected virtual void Dispose(bool disposing)
         public void Dispose()
         {
             Dispose(true);
+            GC.SuppressFinalize(this);
```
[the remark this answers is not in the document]
snakefoot: Just removing noise from SonarLint.

src/NLog/Internal/Fakeables/IAppEnvironment.cs
304NotModified: Please split it into separate interfaces. It's also important for internal code for a proper design

snakefoot: Have split into IFileSystem and IAppEnvironment

304NotModified: Thanks!

tests/NLog.UnitTests/ConfigFileLocatorTests.cs
```diff
@@ -48,9 +45,234 @@ namespace NLog.UnitTests
...
+            LayoutRenderer.Register("the-answer", (info) => "42");
+            Layout l = "${the-answer}";
+            var result = l.Render(LogEventInfo.CreateNullEvent());
+            Assert.Equal("42", result);
+        }
+
+        [Fact]
+        void FuncLayoutRendererRegisterTest1WithXML()
+        {
+            LayoutRenderer.Register("the-answer", (info) => "42");
```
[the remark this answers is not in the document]
snakefoot: Just me moving unit tests out of the `#if NETSTANDARD `-condition because of the exe-file-compile-logic. Next step will probably be to remove the exe-file-compile and only use mocks. But like to see that my other changes is not affecting the old tests before changing the tests also.

src/NLog/Config/LoggingConfigurationFileLoader.cs
```diff
@@ -51,23 +52,23 @@ namespace NLog.Config
     /// </summary>
     internal class LoggingConfigurationFileLoader : ILoggingConfigurationLoader
     {
-        private static readonly FileWrapper DefaultFileWrapper = new FileWrapper();
-        private readonly IFile _file;
+        private static readonly AppEnvironmentWrapper DefaultFileWrapper = new AppEnvironmentWrapper();
```
304NotModified: I think we should rename DefaultFileWrapper  here?

src/NLog/Config/LoggingConfigurationFileLoader.cs
```diff
@@ -235,22 +252,84 @@ public IEnumerable<string> GetDefaultCandidateConfigFilePaths(string fileName)
...
+
+            foreach (var filePath in GetPrivateBinPathNLogLocations(baseDirectory, nlogConfigFile, platformFileSystemCaseInsensitive ? nLogConfigFileLowerCase : string.Empty))
+                yield return filePath;
+
+#if !SILVERLIGHT && !NETSTANDARD1_0
+            if (fileName == null)
+            {
+                // Get path to NLog.dll.nlog only if the assembly is not in the GAC
+                var nlogAssembly = typeof(LogFactory).GetAssembly();
+                if (!string.IsNullOrEmpty(nlogAssembly?.Location) && !nlogAssembly.GlobalAssemblyCache)
```
304NotModified: Guess we should to do this one also (make mockable)

src/NLog/LayoutRenderers/ProcessIdLayoutRenderer.cs
```diff
@@ -61,9 +63,9 @@ bool IRawValue.TryGetRawValue(LogEventInfo logEvent, out object value)
             return true;
         }
 
-        private static int GetValue()
+        private int GetValue()
         {
-            return ThreadIDHelper.Instance.CurrentProcessID;
+            return _currentProcessId != 0 ? _currentProcessId : (_currentProcessId = ProcessIDHelper.Instance.CurrentProcessID);
```
304NotModified: just checking, this kind of a "cache"?

tests/NLog.UnitTests/ConfigFileLocatorTests.cs
```diff
@@ -48,9 +45,234 @@ namespace NLog.UnitTests
...
+                EntryAssemblyFileName = "Entry.dll"
+            };
+
+            var fileLoader = new LoggingConfigurationFileLoader(appEnvMock);
+
+            // Act
+            var result = fileLoader.GetDefaultCandidateConfigFilePaths().ToList();
+
+            // Assert base-directory + process-directory + nlog-assembly-directory
+            if (NLog.Internal.PlatformDetector.IsWin32)
```
304NotModified: please add some helpers for the test duplication (at least here in the assert part), thanks!

tests/NLog.UnitTests/Mocks/AppEnvironmentMock.cs
```diff
@@ -32,23 +32,43 @@
...
 
 using System;
+using System.Collections.Generic;
+using System.Xml;
 using NLog.Internal.Fakeables;
 
 namespace NLog.UnitTests.Mocks
 {
-    public class FileMock : IFile
+    internal class AppEnvironmentMock : IAppEnvironment
```
304NotModified: FYI, you could  use NSubsitute  for this when #3274 is merged

src/NLog/Config/LoggingConfigurationFileLoader.cs
snakefoot: Think I have the code-coverage I need. For another time.

src/NLog/Config/LoggingConfigurationFileLoader.cs
snakefoot: Good idea

src/NLog/Config/LoggingConfigurationFileLoader.cs
304NotModified: Fair enough

304NotModified:
maybe we should update the docs - and this unfortunately related: https://github.com/NLog/NLog/pull/959

src/NLog/LayoutRenderers/ProcessIdLayoutRenderer.cs
snakefoot: Yes but it is silly. Have remove it.

tests/NLog.UnitTests/ConfigFileLocatorTests.cs
snakefoot: Done

snakefoot:
> @304NotModified  maybe we should update the docs

Well this PR actually tries to ensure that NetCore follows the Wiki :)

tests/NLog.UnitTests/Mocks/AppEnvironmentMock.cs
snakefoot: Fine with implementing simple mocks by hand. It is more when meeting the monster-interface (or non-interfaces) that Mock-frameworks are helpful.

304NotModified:
so supersedes #959? Isn't this a breaking behavior change?

snakefoot:
> @304NotModified so supersedes #959? Isn't this a breaking behavior change?

No and No.

#959 actually implements the config loading so it happens in the correct order. It should always use the application-specific process.exe.nlog instead of a random Nlog.config. But this is a breaking change, so you have stalled this fix for NLog 5.0

This PR only fixes NetCore so it behaves like NetFramework. Because `AppDomain.SetupInformation.ConfigurationFile` is blank on NetCore-platform and needs special help.

304NotModified:
Ok thanks for explaining!

FYI, I like to start with merging 5.0 PRs after 4.6.3 is done.

snakefoot:
> FYI, I like to start with merging 5.0 PRs after 4.6.3 is done.

Think we should keep dev and master available for more hotfixes a few more weeks before starting up on NLog 5.0. But your decision.

304NotModified:
Hotfixes could be done to master then :)

(But features will be 5.0 instead of 4.6.*)

snakefoot:
@304NotModified Is the merge of this PR blocked by something?

304NotModified:
Yes my time. I try to do some developing next to reviewing. Only reviewing is boring and I'm losing some touch to NLog in that way.

src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -64,7 +64,7 @@ public class XmlLoggingConfiguration : LoggingConfigurationParser
         /// <summary>
         /// Prefix for assets in Xamarin Android
         /// </summary>
-        internal const string AssetsPrefix = "assets/";
+        private const string AssetsPrefix = "assets/";
```
304NotModified: 👍

src/NLog/Internal/Fakeables/IFileSystem.cs
```diff
@@ -33,14 +33,18 @@
...
 namespace NLog.Internal.Fakeables
 {
+    using System.Xml;
+
     /// <summary>
-    /// Abstract calls to File
+    /// Abstract calls to FileSystem
     /// </summary>
-    internal interface IFile
+    internal interface IFileSystem
```
304NotModified: FYI, this was not the direction I would take. I think it's better to have 10 small interfaces then 3 medium. 

Also, with the introduction of NSubstitute, really small interfaces are nice. 

But as it's internal I think this is more than good enough 👍

304NotModified:
> @304NotModified Is the merge of this PR blocked by something?

Anyway, pushing has some results ;)

304NotModified:
nice results!

image

snakefoot:
> @304NotModified  Yes my free time. I try to do some developing next to reviewing. Only reviewing is boring and I'm losing some touch to NLog in that way.

Very understandable. I also prefer developing over reviewing :) Will try and reduce the number of PRs for the NLog-project, so everything is not about my PRs :)

But thank you for the merge. Now there is some cleanup to do:
- Refactor the remaining unit tests in `ConfigFileLocatorTests.cs` to use the new mock-interface instead of building exe-file (Faster and can be tested on NetCore).
- Update #959 to handle the moving of code from `LogFactory` to `LoggingConfigurationFileLoader`

I don't mind fixing these two, unless they should be up-for-grabs?

PR #3262, opened 2019-03-30
AssemblyVersionLayoutRenderer - Tests should reuse the in-memory built assembly

snakefoot:


tests/NLog.UnitTests/LayoutRenderers/AssemblyVersionTests.cs
```diff
@@ -45,6 +44,11 @@ public class AssemblyVersionTests : NLogTestBase
     {
         private readonly ITestOutputHelper _testOutputHelper;
 
+#if !NETSTANDARD
+        private static Assembly TestAssembly => _testAssembly ?? (_testAssembly = GenerateTestAssembly());
```
304NotModified: Maybe we could use Lazy in the test project? :)

snakefoot: Fixed

PR #3263, opened 2019-03-31
SonarLint and CodeFactor warning reduction + better internal logging

snakefoot:


src/NLog/LayoutRenderers/NdcLayoutRenderer.cs
```diff
@@ -81,7 +81,7 @@ protected override void Append(StringBuilder builder, LogEventInfo logEvent)
         {
             if (TopFrames == 1)
             {
-                // Allows fast rendering of ${when:when='${ndc:topframes=1}' == '':inner=:else=${ndc}|}
+                // Allows fast rendering of ${ndc:topframes=1}
```
304NotModified: :+1: that's a easier example :)

304NotModified:
FYI 

> Targets/FileTarget.cs(730,51): error CS0103: The name '_lineEndingMode' does not exist in the current context [/home/travis/build/NLog/NLog/src/NLog/NLog.csproj]

src/NLog/LayoutRenderers/WindowsIdentityLayoutRenderer.cs
```diff
@@ -81,37 +81,26 @@ protected override void Append(StringBuilder builder, LogEventInfo logEvent)
...
             {
                 string output = string.Empty;
-
-                if (UserName)
+                if (UserName && Domain)
+                {
+                    // username && domain
+                    output = currentIdentity.Name;
+                }
+                else if (UserName)
```
304NotModified: I think this is a code smell (party same condition as previous condition)

snakefoot: I just wanted to reduce the complexity. If you are welcome to continue the work in a new PR :)

src/NLog/LayoutRenderers/NdcLayoutRenderer.cs
snakefoot: SonarLint complained about code in comments. Trying to get around that.

src/NLog/LayoutRenderers/WindowsIdentityLayoutRenderer.cs
304NotModified: Yes I like that, but we need to be careful not introducing more fragile code :)

snakefoot: Have you looked at the original code?

304NotModified: Only diff (on mobile), but I will first then look in depth (and I could also commit on this Branch ;)

src/NLog/LayoutRenderers/NdcLayoutRenderer.cs
304NotModified: another way is to move it to xml docs, e.g. `<example>` or `<remarks>`

src/NLog/LayoutRenderers/WindowsIdentityLayoutRenderer.cs
304NotModified: I checked it in-depth, and will try to explain better:

The current code had the structure:

```C#
if(Username)
{
    if (Domain) { }
}
else
{
    if (!Domain) { }
}
```

The new code has the structure:

```C#
if (UserName && Domain) { }
else if (UserName) { }
else if (Domain) { }
```

I think the first structure is better because:
- easier conditions (no combinations)
- no else condition that also is already check in the if condition
- 2 small code blocks are preferred over 1 larger code block.


Still we need to refactor the code, as it was too long indeed.

I propose an extract method for this. See also https://blog.codinghorror.com/flattening-arrow-code/

304NotModified:
@snakefoot thanks for doing refactorings 👍 

FYI: I really like this site, which has some nice lists how to refactor things: https://sourcemaking.com/refactoring

src/NLog/LayoutRenderers/WindowsIdentityLayoutRenderer.cs
snakefoot: Well I like my code. Removes 10 lines of unnecessary code. And is more readable to me. Will revert my improvement.

304NotModified: I have created: https://github.com/NLog/NLog/pull/3265

snakefoot: Heheh so instead of removing 10 lines, then you have added 20 lines more. Ohh well.

304NotModified: 👼  the code complexity of all those methods is very low.

snakefoot: Removed my changes to `WindowsIdentityLayoutRenderer.cs` from this PR

src/NLog/Config/LoggingConfigurationParser.cs
```diff
@@ -204,7 +204,12 @@ private void SetNLogElementSettings(ILoggingConfigurationElement nlogConfig)
             {
                 if (!string.IsNullOrEmpty(configItem.Key))
                 {
-                    dict[configItem.Key.Trim()] = configItem.Value;
+                    string key = configItem.Key.Trim();
+                    if (dict.ContainsKey(key))
+                    {
```
304NotModified: This is something else than the refactor, isn't?

304NotModified: Are there more of these changes in this PR?

snakefoot: Yes some debug-noise to help diagnose Xml-config issues.

Issue #3181, opened 2019-03-17
Code that could get some coverage


304NotModified:
- [ ] TargetWithContext.cs - [coverage](https://codecov.io/gh/NLog/NLog/src/a38d8bd891fa6c3e5793b75cea21b4b2fc1917f5/src/NLog/Targets/TargetWithContext.cs)
- [ ] NetworkSenders/ -  [coverage](https://codecov.io/gh/NLog/NLog/tree/a38d8bd891fa6c3e5793b75cea21b4b2fc1917f5/src/NLog/Internal/NetworkSenders)
- [ ] LogReceiverService/ -  [coverage](https://codecov.io/gh/NLog/NLog/tree/a38d8bd891fa6c3e5793b75cea21b4b2fc1917f5/src/NLog/LogReceiverService)
- [ ] (lower) AsyncTaskTarget.cs - [coverage](https://codecov.io/gh/NLog/NLog/src/a38d8bd891fa6c3e5793b75cea21b4b2fc1917f5/src/NLog/Targets/AsyncTaskTarget.cs)

aerotog:
The TypeConverter class, LogLevelTypeConverter, has 88% code coverage via the test LogLevelConvertTest in LogLevelTests. I don't see it in your coverage report so I'm wondering if it actually need more coverage added or if it could already be considered complete. I suppose it might benefit to move the test into its own dedicated test class to make the coverage more obvious.

Could you elaborate on what the these requested test cases mean:
* [ ]  (easy) ValueFormatter  [coverage](https://codecov.io/gh/NLog/NLog/src/a38d8bd891fa6c3e5793b75cea21b4b2fc1917f5/src/NLog/MessageTemplates/ValueFormatter.cs)
  * [ ]  render null
  * [ ]  handle cycle

304NotModified:
> The TypeConverter class, LogLevelTypeConverter, has 88% code coverage via the test LogLevelConvertTest in LogLevelTests. I don't see it in your coverage report so I'm wondering if it actually need more coverage added or if it could already be considered complete. I suppose it might benefit to move the test into its own dedicated test class to make the coverage more obvious.


I have to check that, maybe it's done already indeed. 

> Also, could you elaborate on what the these requested test cases mean:
> 
[...]

PR #3286, opened 2019-04-06
Added MessageTemplateParameter IsValidTemplateTest unit test

304NotModified:
see https://github.com/NLog/NLog/issues/3181

todo:

- [x] rebase/skip after merge #3291

@snakefoot 

got some interesting results here:

image

With an empty list as `object[] parameters` , is always says `IsValidTemplate=true` - independent of the message-template. Is this expected? It this for performance reasons?

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/3286)

### DIFF
--- a/tests/NLog.UnitTests/MessageTemplates/MessageTemplateParametersTests.cs
+++ b/tests/NLog.UnitTests/MessageTemplates/MessageTemplateParametersTests.cs
@@ -118,6 +118,45 @@ namespace NLog.UnitTests.MessageTemplates
             Assert.Equal(names, resultNames);
         }
 
+        [Theory]
+        [InlineData("", 0, true)] //empty OK
+        [InlineData("  ", 0, true)] //empty OK
+        [InlineData("", 1, false)]
+        [InlineData("{0}", 1, true)]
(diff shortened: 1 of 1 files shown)


snakefoot:
> @304NotModified With an empty list as object[] parameters , is always says IsValidTemplate=true - independent of the message-template. Is this expected? It this for performance reasons?

Have no special opinion about `MessageTemplateParameters` it was created for the Seq-target that wanted access to the internal parser engine of NLog.

By default NLog will ignore message-template when no parameters are provided. This is used by many that just writes raw Json without parameters. Whether there should be any correlation between the behavior of `MessageTemplateParameters` and the default parser/renderer behavior of `NLog` that I don't know.

Your own special fix with `GetHoleValueSafe` should probably also handle `isValidTemplate` :)

Any changes you make to `MessageTemplateParameters` should just be validated against the behavior of the Seq-Target.

304NotModified:
related: https://github.com/NLog/NLog/issues/2339

304NotModified:
closing for now, tracked here: https://github.com/NLog/NLog/issues/3375

PR #3298, opened 2019-04-10
Added WithProperty and SetProperty on Logger

### LINKED ISSUE #528: Feature request: define/set variable inside a logger

snakefoot:
Logger - WithProperty for adding Logger specific properties that are included for all LogEvents

Different attempt of implementing #2496

Still trying to resolve #528

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/3298)

### DIFF
--- a/src/NLog/Logger.cs
+++ b/src/NLog/Logger.cs
@@ -34,12 +34,13 @@
 namespace NLog
 {
     using System;
+    using System.Collections.Generic;
     using System.ComponentModel;
-    using NLog.Internal;
-
 #if NET4_5
     using System.Threading.Tasks;
 #endif
+    using JetBrains.Annotations;
+    using NLog.Internal;
 
     /// <summary>
     /// Provides logging interface and utility functions.
@@ -47,9 +48,10 @@ namespace NLog
     [CLSCompliant(true)]
     public partial class Logger : ILogger
     {
-        private readonly Type _loggerType = typeof(Logger);
-
-        private volatile LoggerConfiguration _configuration;
+        internal static readonly Type DefaultLoggerType = typeof(Logger);
+        private Logger _contextLogger;
+        private Dictionary<string, object> _contextProperties;
+        private LoggerConfiguration _configuration;
         private volatile bool _isTraceEnabled;
         private volatile bool _isDebugEnabled;
         private volatile bool _isInfoEnabled;
@@ -62,6 +64,35 @@ namespace NLog
         /// </summary>
         protected internal Logger()
         {
+            _contextLogger = this;
+        }
+
+        /// <summary>
+        /// Creates new logger that automatically appends the specified property to all log events
+        /// </summary>
+        /// <param name="propertyKey">Property Name</param>
+        /// <param name="propertyValue">Property Value</param>
+        /// <returns>New Logger that automatically appends specified property</returns>
+        public Logger WithProperty(string propertyKey, object propertyValue)
+        {
+            if (string.IsNullOrEmpty(propertyKey))
+                throw new ArgumentException(nameof(propertyKey));
+
+            Logger newLogger = Factory.CreateNewLogger(GetType()) ?? new Logger();
+            newLogger.Initialize(Name, _configuration, Factory);
+            var contextProperties = _contextProperties != null
+                ? new Dictionary<string, object>(_contextProperties)
+                : new Dictionary<string, object>();
+            contextProperties[propertyKey] = propertyValue;
+            newLogger._contextProperties = contextProperties;
+            newLogger._contextLogger = _contextLogger;
+            newLogger._isTraceEnabled = true;
+            newLogger._isDebugEnabled = true;
+            newLogger._isInfoEnabled = true;
+            newLogger._isWarnEnabled = true;
+            newLogger._isErrorEnabled = true;
+            newLogger._isFatalEnabled = true;
+            return newLogger;
         }
 
         /// <summary>
@@ -100,11 +131,12 @@ namespace NLog
         /// <param name="logEvent">Log event.</param>
         public void Log(LogEventInfo logEvent)
         {
-            if (IsEnabled(logEvent.Level))
+            var targetsForLevel = IsEnabled(logEvent.Level) ? GetTargetsForLevel(logEvent.Level) : null;
+            if (targetsForLevel != null)
             {
                 if (logEvent.LoggerName == null)
                     logEvent.LoggerName = Name;
-                WriteToTargets(logEvent);
+                WriteToTargets(logEvent, targetsForLevel);
             }
         }
 
@@ -115,11 +147,12 @@ namespace NLog
         /// <param name="logEvent">Log event.</param>
         public void Log(Type wrapperType, LogEventInfo logEvent)
         {
-            if (IsEnabled(logEvent.Level))
+            var targetsForLevel = IsEnabled(logEvent.Level) ? GetTargetsForLevel(logEvent.Level) : null;
+            if (targetsForLevel != null)
             {
                 if (logEvent.LoggerName == null)
                     logEvent.LoggerName = Name;
-                WriteToTargets(wrapperType, logEvent);
+                WriteToTargets(wrapperType, logEvent, targetsForLevel);
             }
         }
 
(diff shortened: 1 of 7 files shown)


304NotModified:
Thanks for the PR,

but i'm afraid it won't change the NLog API viewpoint, mentioned in https://github.com/NLog/NLog/pull/2496#issuecomment-410059676?

snakefoot:
The Logger currently has name that matches the context it represents. For every context you create a new logger.


This PR allows you to inject more detail into the context that the Logger represents (besides its name).


I don't like global context properties that affects all users of the same logger. Kind of ruins the idea of context.


Sent from my Sony Xperia


Thanks for the PR,

but i'm afraid it won't change the NLog API viewpoint, mentioned in #2496 (comment)<https://github.com/NLog/NLog/pull/2496#issuecomment-410059676>?

304NotModified:
If we also add SetProperty, next to WithProperty, do you think that is OK? (from viewpoint API design)

snakefoot:
@304NotModified Added a `SetProperty` method for updating the current object-instance.

304NotModified:
Cool!

304NotModified:
Need to check it in depth, but I think I will merge it soon :)

src/NLog/Logger.cs
```diff
@@ -94,17 +97,70 @@ public bool IsEnabled(LogLevel level)
...
+                : new Dictionary<string, object>();
+            contextProperties[propertyKey] = propertyValue;
+            newLogger._contextProperties = contextProperties;
+            newLogger._contextLogger = _contextLogger;
+            newLogger._isTraceEnabled = true;
+            newLogger._isDebugEnabled = true;
+            newLogger._isInfoEnabled = true;
+            newLogger._isWarnEnabled = true;
+            newLogger._isErrorEnabled = true;
+            newLogger._isFatalEnabled = true;
```
304NotModified: Why is everything enabled for the new logger?

src/NLog/Logger.cs
```diff
@@ -94,17 +97,70 @@ public bool IsEnabled(LogLevel level)
...
+        /// <param name="propertyValue">Property Value</param>
+        /// <returns>New Logger object that automatically appends specified property</returns>
+        public Logger WithProperty(string propertyKey, object propertyValue)
+        {
+            if (string.IsNullOrEmpty(propertyKey))
+                throw new ArgumentException(nameof(propertyKey));
+
+            Logger newLogger = Factory.CreateNewLogger(GetType()) ?? new Logger();
+            newLogger.Initialize(Name, _configuration, Factory);
+            var contextProperties = _contextProperties != null
```
304NotModified: please create a helper for creating contextProperties  - this is prevent to get the code complexity above the boundary

src/NLog/Logger.cs
```diff
@@ -94,17 +97,70 @@ public bool IsEnabled(LogLevel level)
...
+        /// </remarks>
+        /// <param name="propertyKey">Property Name</param>
+        /// <param name="propertyValue">Property Value</param>
+        public void SetProperty(string propertyKey, object propertyValue)
+        {
+            if (string.IsNullOrEmpty(propertyKey))
+                throw new ArgumentException(nameof(propertyKey));
+
+            var currentProperties = _contextProperties;
+            var contextProperties = new Dictionary<string, object>((currentProperties?.Count ?? 0) + 1);
```
304NotModified: Could we reuse the helper for contextProperties  here? (see earlier comment)

src/NLog/Logger.cs
```diff
@@ -557,58 +612,94 @@ internal void Initialize(string name, LoggerConfiguration loggerConfiguration, L
...
+            _isErrorEnabled = IsEnabled(LogLevel.Error);
+            _isFatalEnabled = IsEnabled(LogLevel.Fatal);
 
             OnLoggerReconfigured(EventArgs.Empty);
         }
 
         private TargetWithFilterChain GetTargetsForLevel(LogLevel level)
         {
-            return _configuration.GetTargetsForLevel(level);
+            if (ReferenceEquals(_contextLogger, this))
```
304NotModified: this is a performance tweak isn't? Please add a code comment  (before it gets refactored ;))

src/NLog/LoggerImpl.cs
```diff
@@ -51,15 +51,10 @@ internal static class LoggerImpl
         private const int StackTraceSkipMethods = 0;
 
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA2204:Literals should be spelled correctly", Justification = "Using 'NLog' in message.")]
-        internal static void Write([NotNull] Type loggerType, TargetWithFilterChain targets, LogEventInfo logEvent, LogFactory factory)
+        internal static void Write([NotNull] Type loggerType, [NotNull] TargetWithFilterChain targetsForLevel, LogEventInfo logEvent, LogFactory factory)
```
304NotModified: Resharper annotations, always +1

src/NLog/Logger.cs
snakefoot: Noise from my first attempt. Will remove it

src/NLog/Logger.cs
snakefoot: Will fix it

src/NLog/Logger.cs
snakefoot: Will fix it

src/NLog/Logger.cs
snakefoot: Nope avoid recursive loop.

src/NLog/Logger.cs
```diff
@@ -94,17 +97,62 @@ public bool IsEnabled(LogLevel level)
...
+
         /// <summary>
         /// Writes the specified diagnostic message.
         /// </summary>
         /// <param name="logEvent">Log event.</param>
         public void Log(LogEventInfo logEvent)
         {
-            if (IsEnabled(logEvent.Level))
+            var targetsForLevel = IsEnabled(logEvent.Level) ? GetTargetsForLevel(logEvent.Level) : null;
+            if (targetsForLevel != null)
```
304NotModified: is this a performance tweak?

src/NLog/Logger.cs
```diff
@@ -392,25 +441,23 @@ public void Log<TArgument>(LogLevel level, [Localizable(false)] string message,
             }
         }
 
-        private void WriteToTargets(LogLevel level, Exception ex, [Localizable(false)] string message, object[] args)
```
304NotModified: less overloads = 👍

snakefoot:
Think it is safe for NLog 4.6.3<tel:4.6.3>


Not sure what you are referring to regarding "performance tweak"


Sent from my Sony Xperia


@304NotModified approved this pull request.

Thanks!

Is this safe for 4.6.3?

In src/NLog/Logger.cs<https://github.com/NLog/NLog/pull/3298#discussion_r275955654>:

>          /// <summary>

         /// Writes the specified diagnostic message.

         /// </summary>

         public void Log(LogEventInfo logEvent)

         {

-            if (IsEnabled(logEvent.Level))

+            var targetsForLevel = IsEnabled(logEvent.Level) ? GetTargetsForLevel(logEvent.Level) : null;

+            if (targetsForLevel != null)



is this a performance tweak?

In src/NLog/Logger.cs<https://github.com/NLog/NLog/pull/3298#discussion_r275955760>:

> @@ -392,25 +441,23 @@ public void Log<TArgument>(LogLevel level, [Localizable(false)] string message,

             }

         }



-        private void WriteToTargets(LogLevel level, Exception ex, [Localizable(false)] string message, object[] args)



less overloads = 👍

304NotModified:
added here: https://github.com/NLog/NLog/wiki/EventProperties-Layout-Renderer

snakefoot:
Updated the docs a little so it the difference between `WithProperty` and `SetProperty` is clearer. So one will not be caught by the trap-door created by `SetProperty`.

304NotModified:
It's not s trap but feature ;)

304NotModified:
Updated it with a 2nd log. The set is really useful is some cases (e.g when you get more context along the way)

snakefoot:
> The set is really useful is some cases (e.g when you get more context along the way)

Not when having the global NLog Logger-cache where multiple locations are referencing the same Logger-object, and then one location suddenly decides to call SetProperty.

304NotModified:
It's not suddenly? It's by (logger) name.

304NotModified:
Buy anyway, I'm curious when we will implement WithProperties and ClearProperty :p

(Maybe nice with default interfaces - c# 8)

snakefoot:
> It's not suddenly? It's by (logger) name.

Not sure I understand. There is no guarantee that two isolated locations calls `GetLogger()` with the same name. When one of the isolated locations calls `SetProperty` then it will affect the other.

PR #3302, opened 2019-04-11
WIP: Support configuration that includes of remote URIs.

mmurrell:
This is an early WIP pull request to test the viability of including configuration via remote URIs in addition to local files.   

Admittedly there is some risk here, should the remote server be unavailable or there is other network interruptions, but coupled with a fallback configuration, and/or auto-reload, those could be mitigated.

Sample configuration:
```xml
<nlog>
    <include uri='https://www.remote.url/nlog.xml' />
</nlog>
```

Work Remaining: (in no particular order)
- [x] Proof of Concept
- [ ] Code Cleanup
- [ ] Validate other frameworks and environments
- [ ] Create test seam for mocking Http Request
- [ ] Covering Tests
- [ ] Supporting auto-reload by a background thread to poll the remote url.

There is plenty of work and test coverage remaining, but before I ventured too far into this one, I wanted to get a gut-check for viability.  Is this something you would consider bringing into the core library, sharing with the public, or should I continue implementation as a private extension?

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/3302)


snakefoot:
In hindsight then I would have preferred that `XmlLoggingConfiguration` had no knowledge about files. It should only receive a Stream. Then there should be extension methods for loading from files or raw-string-content.

To handle the advanced logic of include-files and autoreload, then it should be handled by a virtual-filesystem-interface (Lets call it `IFileSystem`). The `XmlLoggingConfiguration` would then massage this interface.

The implementation of the `IFileSystem` will receive a basedirectory, that all requested file-paths are a relative to. In the network-url situation then IFileSystem will receive a baseurl, that all requested urls-paths are relative to. Ofcourse it should still be possible to provide absolute paths.

Any changes you can make so `LoggingConfigurationParser` can make use of `IFileSystem` to perform autoreload and handle include-files will be an improvement.

Any changes you make that injects more hardcoded logic about loading streams from files or urls in `XmlLoggingConfiguration` (Or `LoggingConfigurationParser`) will make it worse.

mmurrell:
Perfect.  Thanks for the quick reply.  I think I can head that direction with the IFileSystem... Let me see how big of a refactor that will be and if it's something I feel comfortable undertaking.

snakefoot:
I guess you should also get input from @304NotModified since he has the final say.


I just gave my own personal opinion :)


Sent from my Sony Xperia


---- Matt Murrell wrote ----


Perfect. Thanks for the quick reply. I think I can head that direction with the IFileSystem... Let me see how big of a refactor that will be and if it's something I feel comfortable undertaking.

mmurrell:
Alright, here's another take.  I was able to pull the datapoints together into the interface `IXmlConfigurationSource` and update all the logic to reference the interface.  I currently have implementations of `Xml(File|Reader|String)ConfigurationSource`, although the reader one may eventually be rendered obsolete.  Hopefully in the future, `XmlRemoteConfigurationSource` would fall neatly in line here.

As a side effect, I think some of the logic in the `XmlLoggingConfiguration` is actually a bit cleaner now.    The Dictionary that was a reload list, is now a hashset of all loaded configuration sources.  I think some of the constructors can be cleaned up as well, but I did not want to introduce potentially breaking changes.

My only concern with the current approach is managing the disposables that come out of the sources.  There's no clear pattern for who "owns" the resource, and therefore should dispose.  (Does the consuming class dispose the file that was opened?  What about a passed in XmlReader? etc.)

I expect that the next step would be to move the reload logic out of `LoggingConfigurationWatchableFileLoader` and into the implementations of the various source classes.  

Can I get a real quick WIP code review?

snakefoot:
If you introduce the interface I was talking about, then I think all logic about file and and filenames and directories should be handled by the interface.

Anything about uri or filepath should not be something the XmlLoggingConfiguration should worry about (or recognize or parse).

304NotModified:
HI @mmurrell ,

Thanks for the PR, I really like the idea!

I

> Anything about uri or filepath should not be something the XmlLoggingConfiguration should worry about (or recognize or parse).

I agree on that one, but  currently I haven't checked in-depth how to fix that. I think snakefoot's suggestions are good 👍

mmurrell:
I've made another round of commits, hoping that this brings us closer to a consensus. 

A few thoughts:

- I strongly agree that there must be a separation of concerns, and that the `XmlLoggingConfiguration` has no need to know about File I/O.  In the two most recent commits, I've moved the Include file logic into `XmlFileConfigurationSource` to further separate those implementations.  The only references to File I/O that remain in the `XmlLoggingConfiguration` class are the filename based constructors, which may seem appropriate, and the `base.LoadConfig(nlogElement, config.LocalFolder);` on [line #460](https://github.com/NLog/NLog/pull/3302/files#diff-3eaaf2a7ce39e97333bb717020da7452R460), and the `(Get|Set)CandidateConfigFilePaths` methods.  How should I proceed with those?

- However, I'm not sure if `IFileSystem` is the semantically correct name / separation.  For example, sometimes logging configuration is parsed from Xml strings.  I don't see any "file system' related operations there.  Similarly, should we implement a pattern for loading configuration values remotely, I don't see how a remote web server would align with `IFileSystem`.  I expect you would not write any files remotely.  Would you accept the name `IXmlConfigurationSource` or another alternative in lieu of `IFileSystem` or would you still prefer I change it.  If you'd like me to change it, are there any other interface methods / properties that I've missed and should be supporting here?

 - The `XmlReader` constructor is proving problematic, especially when a forward-only XmlReader constructor is used in some reload scenarios.  As such, I've refactored a few tests to call other constructors, include the file constructor and the string `xmlContents` constructor.  It may be desirable to deprecate all the reader constructors and remove them outright, but that would be a breaking change that I don't feel comfortable implementing.  

Looking forward to your thoughts and continued conversation.

snakefoot:
- Yes to avoid breaking changes (in initial implementation) then the old public constructors must remain. But they should "just" create an instance of `IXmlConfigurationSource` and call the correct constructor taking the interface as parameter.

- `IFileSystem` was mostly to convey and idea of separation. You are completely correct that would one should not be able to write anything from `XmlLoggingConfiguration`. So `IXmlConfigurationSource`  is also good.

- Not sure I understand how XmlReader-constructor is problematic. I would guess this constructor would also create an instance of `IXmlConfigurationSource`  and call the correct constructor taking the interface as parameter.

`XmlLoggingConfiguration` is still very tightly integrated with `LogFactory`. And `LogFactory` is still very focused on loading from files (which is sad). I have tried to extract of the file-logic from `LogFactory` into `LoggingConfigurationFileLoader`. Again this file-logic is solely for handling `XmlLoggingConfiguration`, and should probably also live in the implementation of `IXmlConfigurationSource`, but that is probably for a future PR.

304NotModified:
> P.S. There is no harm done in marking unwanted constructors as obsolete

Except I need to bump the minor version for that (see http://semver.org) ;)

mmurrell:
Apologies for the delays.  I don't have the time to see this to completion at the moment, so I'm going to close for the moment.  When I can spend some more and bring this to completion, I'll reopen.  Thanks.

Issue #3308, opened 2019-04-14
NLog.NLogConfigurationException: Error when setting property 'Uppercase' on Layout Renderer: ${uppercase}


majumajid:
I tried to implement the NLog using this reference link (https://martynnw.wordpress.com/2016/10/16/logging-with-nlog-in-xamarin-forms/) and github(https://github.com/Martynnw/AndroidDemos/tree/master/LoggingDemo)

This error is encountering on this line 
**var config = new LoggingConfiguration();  //THIS LINE GENERATING ERROR**

```c#
 [assembly: Dependency(typeof(NLogManager))]
namespace LoggingDemo.Droid
{
    public class NLogManager : ILogManager
    {
        public NLogManager()
        {
             var config = new LoggingConfiguration();   
            var consoleTarget = new ConsoleTarget();
            config.AddTarget("console", consoleTarget);

            var consoleRule = new LoggingRule("*", LogLevel.Trace, consoleTarget);
            config.LoggingRules.Add(consoleRule);

            var fileTarget = new FileTarget();
            string folder = Environment.GetFolderPath(Environment.SpecialFolder.MyDocuments);
            fileTarget.FileName = Path.Combine(folder, "Log.txt");
            config.AddTarget("file", fileTarget);

            var fileRule = new LoggingRule("*", LogLevel.Warn, fileTarget);
            config.LoggingRules.Add(fileRule);

            LogManager.Configuration = config;
        }

        public ILogger GetLog([System.Runtime.CompilerServices.CallerFilePath] string callerFilePath = "")
        {
            string fileName = callerFilePath;

            if (fileName.Contains("/"))
            {
                fileName = fileName.Substring(fileName.LastIndexOf("/", StringComparison.CurrentCultureIgnoreCase) + 1);
            }

            var logger = LogManager.GetLogger(fileName);
            return new NLogLogger(logger);
        }
    }
}
```

ERROR :

```
{NLog.NLogConfigurationException: Error when setting property 'Uppercase' on Layout Renderer: ${uppercase}
---> System.MissingMethodException: string[] string.Split(char,System.StringSplitOptions)
 at NLog.Internal.PropertyHelper.SetPropertyFromString (System.Object obj, System.String propertyName, System.String value,
 NLog.Config.ConfigurationItemFactory configurationItemFactory) [0x000b1] in <b575d0eed9dd4c07998e108d5c7bb2ad>:0
  --- End of inner exception stack trace ---
[...]

PR #3311, opened 2019-04-14
Split string -  avoid allocation of object array. Added StringHelpers.SplitAndTrimTokens

snakefoot:
Triggered by #3308 to avoid params-object-array unless needed. But also to encapsulate:

> stringValue.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries);

In a simpler method:

> stringValue.SplitAndTrimTokens(';');

That also trims the tokens, and removes empty tokens.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/3311)

### DIFF
--- a/src/NLog/Config/LoggingConfigurationParser.cs
+++ b/src/NLog/Config/LoggingConfigurationParser.cs
@@ -610,8 +610,8 @@ namespace NLog.Config
                         break;
                     case "LEVELS":
                         {
-                            string[] tokens = CleanSpaces(childProperty.Value).Split(',');
-                            enableLevels = tokens.Where(t => !string.IsNullOrEmpty(t)).Select(LogLevelFromString);
+                            string[] tokens = childProperty.Value.SplitAndTrimTokens(',');
+                            enableLevels = tokens.Select(LogLevelFromString);
                             break;
                         }
                     case "MINLEVEL":
@@ -680,12 +680,8 @@ namespace NLog.Config
             if (string.IsNullOrEmpty(writeTargets))
                 return;
 
-            foreach (string t in writeTargets.Split(','))
+            foreach (string targetName in writeTargets.SplitAndTrimTokens(','))
             {
-                string targetName = t.Trim();
-                if (string.IsNullOrEmpty(targetName))
-                    continue;
-
                 Target target = FindTargetByName(targetName);
                 if (target != null)
                 {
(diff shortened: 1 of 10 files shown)


304NotModified:
👍 c# ext :)

Could you please add some test cases for SplitAndTrimTokens? If that one has a bug, that could have some serious consequences.

src/NLog/Config/LoggingConfigurationParser.cs
```diff
@@ -1227,19 +1223,7 @@ private static string StripOptionalNamespacePrefix(string attributeValue)
...
             return attributeValue.Substring(p + 1);
         }
-
-        /// <summary>
-        /// Remove all spaces, also in between text. 
-        /// </summary>
-        /// <param name="s">text</param>
-        /// <returns>text without spaces</returns>
-        /// <remarks>Tabs and other whitespace is not removed!</remarks>
-        private static string CleanSpaces(string s)
```
304NotModified: funny, this is something else then a trim. Not sure why this was introduced.

src/NLog/Internal/StringHelpers.cs
```diff
@@ -61,5 +59,28 @@ internal static bool IsNullOrWhiteSpace(string value)
...
+
+            if (value.IndexOf(delimiter) == -1)
+            {
+                return new[] { value.Trim() };
+            }
+
+            var result = value.Split(new char[] { delimiter }, StringSplitOptions.RemoveEmptyEntries);
+            for (int i = 0; i < result.Length; ++i)
+            {
+                if (char.IsWhiteSpace(result[i][0]) || char.IsWhiteSpace(result[i][result[i].Length - 1]))
```
304NotModified: this a "smart" trim, isn't?

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -409,7 +409,7 @@ protected virtual void AppendSerializeObject(StringBuilder sb, Exception ex)
         private static List<ExceptionRenderingFormat> CompileFormat(string formatSpecifier)
         {
             List<ExceptionRenderingFormat> formats = new List<ExceptionRenderingFormat>();
-            string[] parts = formatSpecifier.Replace(" ", string.Empty).Split(new string[] { "," }, StringSplitOptions.RemoveEmptyEntries);
+            string[] parts = formatSpecifier.SplitAndTrimTokens(',');
```
304NotModified: we've lost remove space here, not sure if this is an issue.

snakefoot: Yes it would support a format like `${exception:format=to string}`. But I think that is a buggy sideeffect of trying to support  `${exception:format=tostring }` (With trailing space).

src/NLog/Internal/StringHelpers.cs
snakefoot: Nope just my premature optimization (Always expecting the worst from Microsoft). But actually string.Trim() looks good so no need for this.

src/NLog/Config/LoggingConfigurationParser.cs
snakefoot: To remove trailing spaces in `Info, Warn `

snakefoot:
> @304NotModified  Could you please add some test cases for SplitAndTrimTokens? 

Added some unit-tests

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
304NotModified: I think also, thanks for checking

PR #3325, opened 2019-04-21
Fix Travis build

304NotModified:
See https://travis-ci.community/t/dotnet-core-2-2/1216/17


snakefoot:
@304NotModified Maybe consider to revisit the Mono / Travis unit-tests after the upgrade, to check if the current Skip-conditions are still necessary.

PR #3318, opened 2019-04-17
${event-properties}: Added evaluateAsNestedProperties option

304NotModified:
I was triggered by https://github.com/NLog/NLog/issues/3314,

example: 

    ${event-properties:key1.prop1:evaluateAsNestedProperties=true}

As (maybe naive) implementation to support this.

Choices to be made:

1. Is this a suitable feature and place? (${event-properties})
2. How should I name this? EvaluateAsNestedProperties is from NLog.Web
3. Is this config logical, or should we configure the path separately? e.g. `${event-properties:key1:path=prop1}`
4. Should be support arrays in the future? (e.g.`key1[prop1]`)
5. Do we need (fancy) caching (now or in the future)? Is the `ObjectReflectionCache` suitable for this?
6. What do you think about the case-sensitive-case-insensitive lookup? (best for configuration, not for performance, although we could improve that with 2 dictionaries)

@snakefoot you opinion please :)

PR #3329, opened 2019-04-22
${event-properties} - Added ObjectPath for extracting nested object property

snakefoot:
Alternative implementation of  #3318

`${event-properties:myOrder:objectpath=OrderId}`

Right now without fallback to case-insensitive lookup if the direct lookup fails.


304NotModified:
I think this is also better to have in ${event-properties} :)

304NotModified:
cool! Merged!

304NotModified:
docs -> https://github.com/NLog/NLog/wiki/EventProperties-Layout-Renderer

PR #3332, opened 2019-04-24
LogFactory - GetLogger should null-check logger name

snakefoot:
Throw `ArgumentNullException` instead of throwing `NullReferenceException` here (When Name is null):

https://github.com/NLog/NLog/blob/3c16e69ae0bd6e33d16d1580882dce354c02d51c/src/NLog/LogFactory.cs#L1015

Could also consider doing fallback to string.Empty.


304NotModified:
Thanks!

304NotModified:
> Could also consider doing fallback to string.Empty

Doubted also about this, but I think the change in this PR is good enough

PR #3306, opened 2019-04-14
XmlLoggingConfiguration - Respect KeepVariablesOnReload for manual reload

snakefoot:
Allow this to work:

```c#
LogManager.KeepVariablesOnReload = true;
LogManager.Configuration = LoggingConfiguration.Reload(LogManager.Configuration);
```

See also #3304

This is also useful for NLog.Extension.Logging.NLogLoggingConfiguration so it will also support `KeepVariablesOnReload`.

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/3306)

PR #3335, opened 2019-04-26
LoggingConfiguration - Rollback static Reload-method

snakefoot:
Rollback of #3306 until a better solution comes around for https://github.com/NLog/NLog.Extensions.Logging/pull/283

### DIFF
--- a/src/NLog/Config/LoggingConfiguration.cs
+++ b/src/NLog/Config/LoggingConfiguration.cs
@@ -48,8 +48,7 @@ namespace NLog.Config
     using NLog.Targets.Wrappers;
 
     /// <summary>
-    /// Keeps logging configuration and provides simple API
-    /// to modify it.
+    /// Keeps logging configuration and provides simple API to modify it.
     /// </summary>
     ///<remarks>This class is thread-safe.<c>.ToList()</c> is used for that purpose.</remarks>
     public class LoggingConfiguration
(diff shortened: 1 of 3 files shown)


304NotModified:
Hi, thanks for the PR,

but would it be more logical to have something like this:

```c#
LogManager.ReloadConfiguration();
// and/or
LogFactory.ReloadConfiguration();

```

I find the "active" a bit unusual

snakefoot:
> I find the "active" a bit unusual

I need a method on LoggingConfiguraion (or method that takes LoggingConfiguration as parameter), that allows one to reload the config with respect to KeepVariablesOnReload.

But maybe I should just remove the new introduced method completely. Since I can probably implement this variable copy manually in NLog.Extension.Logging.

snakefoot:
I need a method that I can call on an LoggingConfiguration object. That returns a reloaded config.


You are welcome to suggest a new name, but the method must be a member on LoggingConfiguration or take LoggingConfiguration as parameter


Sent from my Sony Xperia


Hi, thanks for the PR,

but would it be more logical to have something like this:

LogManager.ReloadConfiguration();
// and/or
LogFactory.ReloadConfiguration();


I find the "active" a bit unusual

304NotModified:
> But maybe I should just remove the new introduced method completely. Since I can probably implement this variable copy manually in NLog.Extension.Logging.

I'm also doubting this addition makes things easier, so for now I think 

     NLog.LogManager.Configuration = NLog.LogManager.Configuration.Reload();

Is good enough

snakefoot:
> Is good enough

But it will not support `KeepVariablesOnReload`

304NotModified:
> but would it be more logical to have something like this:
> 
> ```cs
> LogManager.ReloadConfiguration();
> // and/or
> LogFactory.ReloadConfiguration();
> ```

Can't be do `newConfig.CopyVariables(currentConfig.Variables);` in e.g. the LogManager to arrange this?

snakefoot:
> in e.g. the LogManager to arrange this?

Sounds like something for NLog 5.0. Lets just rollback #3306

304NotModified:
I'm confused. You forced pushed different content? Won't it be more logical to close the old PR and create a new one?

snakefoot:
Hate to waste a precious PulRequest-number :)

304NotModified:
This is still valid, isn't?

snakefoot:
Yes it removes an unnecessary public method.


Sent from my Sony Xperia


This is still valid, isn't?

PR #3337, opened 2019-04-27
ShortDateLayoutRenderer with support for IRawValue + IStringValueRenderer

snakefoot:
Makes it possible to get current date as raw-value (Without time-part)

Makes it possible to use `${ShortDate}` in `cacheKey` for https://github.com/NLog/NLog/wiki/Cached-Layout-Renderer with reduced string-allocations.


304NotModified:
Thanks!

PR #3353, opened 2019-05-02
StringBuilderPool - Added unit test of MaxCapacity checks

snakefoot:
Regression test for #3345

Figured out it was much easier to test the StringBuilderPool-directly, instead of trying to exercise the entire NLog-engine with many threads.

### DIFF
--- a/tests/NLog.UnitTests/Internal/StringBuilderPoolTests.cs
+++ b/tests/NLog.UnitTests/Internal/StringBuilderPoolTests.cs
@@ -1,0 +1,77 @@
+﻿// 
+// Copyright (c) 2004-2019 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without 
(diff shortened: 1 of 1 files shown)


304NotModified:
cool!

Can we maybe split this into 3 separate unit tests? We could use the ctor as test init for the large fields

snakefoot:
> Can we maybe split this into 3 separate unit tests? 

Do what you like. But the different steps in the test are actually slowly moving the StringBuilderPool into the broken-state (that was possible before the bugfix)

304NotModified:
the order of the RecursiveAcquirePoolItems calls is important, isn't?

snakefoot:
Yes the call order of the recursive method is important for testing this. Tried to explain this with my code comments.


Sent from my Sony Xperia


the order of the RecursiveAcquirePoolItems calls is important, isn't?

304NotModified:
So we can't move to separate tests then?

snakefoot:
You can make additional tests that are simpler than the initial one. But the initial test is important.


Sent from my Sony Xperia


So we can't move to separate tests then?

Issue #2536, opened 2018-01-22
FileTarget default encoding to UTF-8


304NotModified:
Move from Default to UTF-8

See also https://github.com/NLog/NLog/issues/2531#issuecomment-359414713

See also https://blogs.msdn.microsoft.com/shawnste/2005/03/15/dont-use-encoding-default/

PR #3360, opened 2019-05-03
Filetarget: default encoding changed to UTF8 - was Encoding.Default

304NotModified:
fixes https://github.com/NLog/NLog/issues/2536


snakefoot:

PR #3373, opened 2019-05-06
${when}, support for RawValue on Inner and Else

304NotModified:
fixes https://github.com/NLog/NLog/issues/3139

### DIFF
--- a/src/NLog/LayoutRenderers/Wrappers/WhenLayoutRendererWrapper.cs
+++ b/src/NLog/LayoutRenderers/Wrappers/WhenLayoutRendererWrapper.cs
@@ -100,24 +100,21 @@ namespace NLog.LayoutRenderers.Wrappers
         {
             if (ShouldRenderInner(logEvent))
             {
-                if (Inner == null)
-                {
-                    value = null;
-                    return false;
-                }
-
-                return Inner.TryGetRawValue(logEvent, out value);
+                return TryGetRawValueFromLayout(logEvent, Inner, out value);
             }
-            else 
+
+            return TryGetRawValueFromLayout(logEvent, Else, out value);
+        }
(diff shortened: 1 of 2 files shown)


304NotModified:
@snakefoot I checked your comments in the other PR. Your right. I think I have proper implemented it now.

src/NLog/LayoutRenderers/Wrappers/WhenLayoutRendererWrapper.cs
```diff
@@ -83,10 +84,40 @@ protected override void Append(StringBuilder builder, LogEventInfo logEvent)
...
+        {
+            if (ShouldRenderInner(logEvent))
+            {
+                if (Inner == null)
+                {
+                    value = null;
+                    return false;
+                }
+
+                return Inner.TryGetRawValue(logEvent, out value);
```
snakefoot: Could this also work:
```c#
value = null;
if (ShouldRenderInner(logEvent))
   return Inner?.TryGetRawValue(logEvent, out value) ?? false;
else
   return Else?.TryGetRawValue(logEvent, out value) ?? false;
```

304NotModified: I think it should work, but a bit hard to read. But I see the problem of the verbose/duplicate code. Will create a helper and let's see if it's then better.

304NotModified: PS: we need a helper


image

304NotModified: added helper. Let me know what you think

PR #3376, opened 2019-05-06
Scan for process specific nlog-config first

### LINKED ISSUE #223: Load order (precedence) of config files is broken

snakefoot:
Resolves #223 (Updated version of #959)

### DIFF
--- a/src/NLog/Config/LoggingConfigurationFileLoader.cs
+++ b/src/NLog/Config/LoggingConfigurationFileLoader.cs
@@ -220,6 +220,13 @@ namespace NLog.Config
         /// </summary>
         public IEnumerable<string> GetDefaultCandidateConfigFilePaths(string fileName)
         {
+            if (fileName == null)
+            {
+                // Scan for process specific nlog-files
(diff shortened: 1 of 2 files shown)


304NotModified:
Great! Thanks!

304NotModified:
Travis: ConfigFileLocatorTests.cs(258,98): error CS1503: Argument 3: cannot convert from 'System.StringComparer' to 'System.StringComparison' [/home/travis/build/NLog/NLog/tests/NLog.UnitTests/NLog.UnitTests.csproj]
The

304NotModified:
Thanks!

PR #3378, opened 2019-05-07
Implement FasterAccurateLocalTimeSource

Encamy:
This TimeSource is close to the performance of TimeSource.AccurateUtc but returns local time

Here is benchmark results using [BenchmarkDotNet](https://github.com/dotnet/BenchmarkDotNet)
[Source code of benchmark](https://gist.github.com/Encamy/51a789bf00450aa117596ecd1e1a1fa3)

``` ini

BenchmarkDotNet=v0.11.5, OS=Windows 10.0.17763.475 (1809/October2018Update/Redstone5)
Intel Core i5-3570 CPU 3.40GHz (Ivy Bridge), 1 CPU, 4 logical and 4 physical cores
  [Host]     : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.7.3324.0
  DefaultJob : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.7.3324.0


```
|               Method |       Mean |     Error |    StdDev |
|--------------------- |-----------:|----------:|----------:|
|            FastLocal |   5.784 ns | 0.0340 ns | 0.0318 ns |
|              FastUtc |   5.246 ns | 0.0303 ns | 0.0284 ns |
|        AccurateLocal | 215.149 ns | 0.8370 ns | 0.7830 ns |
|          AccurateUtc | 147.001 ns | 0.7035 ns | 0.6581 ns |
| AccurateAndFastLocal | 152.786 ns | 0.7137 ns | 0.6676 ns |

### DIFF
--- a/src/NLog/Time/FastAccurateLocalTimeSource.cs
+++ b/src/NLog/Time/FastAccurateLocalTimeSource.cs
@@ -76,22 +76,21 @@ namespace NLog.Time
 
         private DateTime GetTime()
         {
-            DateTime utc = DateTime.UtcNow;
+            long utcTicks = DateTime.UtcNow.Ticks;
 #if !SILVERLIGHT || WINDOWS_PHONE
-            long deltaTicks = utc.Ticks - Interlocked.Read(ref _lastUpdated);
+            long deltaTicks = utcTicks - Interlocked.Read(ref _lastUpdated);
 #else
-            long deltaTicks = utc.Ticks - Interlocked.CompareExchange(ref _lastUpdated, 0, 0);
+            long deltaTicks = utcTicks - Interlocked.CompareExchange(ref _lastUpdated, 0, 0);
 #endif
             long ticksOffset = _ticksOffset;
 
             if (deltaTicks > TimeSpan.TicksPerSecond || deltaTicks < -TimeSpan.TicksPerSecond)
             {
-                _ticksOffset = ticksOffset = (DateTime.Now - utc).Ticks;
-                _lastUpdated = utc.Ticks;
+                _ticksOffset = ticksOffset = (DateTime.Now.Ticks - utcTicks);
+                _lastUpdated = utcTicks;
             }
 
-            long ticks = utc.Ticks + ticksOffset;
-            return new DateTime(ticks, DateTimeKind.Local);
+            return new DateTime(utcTicks + ticksOffset, DateTimeKind.Local);
         }
     }
 }
--- a/src/NLog/Time/FastAccurateLocalTimeSource.cs
+++ b/src/NLog/Time/FastAccurateLocalTimeSource.cs
@@ -40,8 +40,8 @@ namespace NLog.Time
     /// <summary>
     /// Current local time retrieved from DateTime.UtcNow with local time offset
     /// </summary>
-    [TimeSource("FasterAccurateLocal")]
-    public class FasterAccurateLocalTimeSource : TimeSource
+    [TimeSource("FastAccurateLocal")]
+    public class FastAccurateLocalTimeSource : TimeSource
     {
         private long _ticksOffset;
         private long _lastUpdated;
@@ -49,7 +49,7 @@ namespace NLog.Time
         /// <summary>
         /// Initializes TimeSource with offset from utc to local time
         /// </summary>
-        public FasterAccurateLocalTimeSource()
+        public FastAccurateLocalTimeSource()
         {
             DateTime utc = DateTime.UtcNow;
             _ticksOffset = (DateTime.Now - utc).Ticks;
--- a/src/NLog/Time/FasterAccurateLocalTimeSource.cs
+++ b/src/NLog/Time/FasterAccurateLocalTimeSource.cs
@@ -1,0 +1,97 @@
+﻿// 
+// Copyright (c) 2004-2019 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without 
+// modification, are permitted provided that the following conditions 
+// are met:
+// 
+// * Redistributions of source code must retain the above copyright notice, 
+//   this list of conditions and the following disclaimer. 
+// 
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution. 
+// 
+// * Neither the name of Jaroslaw Kowalski nor the names of its 
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission. 
+// 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE 
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE 
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE 
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR 
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS 
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN 
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) 
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF 
+// THE POSSIBILITY OF SUCH DAMAGE.
+// 
+
+
+namespace NLog.Time
+{
+    using System;
+    using System.Threading;
+
+    /// <summary>
+    /// Current local time retrieved from DateTime.UtcNow with local time offset
+    /// </summary>
+    [TimeSource("FasterAccurateLocal")]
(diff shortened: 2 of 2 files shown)


304NotModified:
Cool! Thanks!

@snakefoot as you're better with performance wise code, any opinion on this?

snakefoot:
Benchmarks are usually the right way to measure performance. If they improve then it is usually good.

But when dealing with local-time and utc-time. Then one need to consider Daylight-Saving-Time (DST) this means that local-time can jump in time-difference compared to utc-time.

The hack with saving the time-difference in ticks between local-time and utc-time needs to be updated atleast every second.

Because the Time-source has to be highly concurrent then multiple threads will attempt to "refresh" the time-difference-cache. The refresh caching-logic needs to cater for this, and update two properties in one operations:

- Current Time Difference Ticks
- Last Updated Time Difference

The most optimal threadsafe way is probably using Interlocked-methods. These usually prefer primitive values (or references). So "Last Updated Time Difference" should probably be stored as DateTime.UtcNow.Ticks.

Encamy:
I understand that DST is very important, but what about countries where it [does not exist](https://en.wikipedia.org/wiki/Daylight_saving_time_by_country)? If cached time will be updated every second, the results will be as follows:

|              Method |       Mean |     Error |    StdDev |
|-------------------- |-----------:|----------:|----------:|
|           FastLocal |   5.965 ns | 0.0975 ns | 0.0912 ns |
|             FastUtc |   5.155 ns | 0.0494 ns | 0.0438 ns |
|       AccurateLocal | 210.057 ns | 1.4504 ns | 1.2857 ns |
|         AccurateUtc | 144.635 ns | 2.3186 ns | 2.1689 ns |
| FasterAccurateLocal_updateEachSecond | 197.228 ns | 2.1071 ns | 1.9710 ns |
| FasterAccurateLocal_updateEachMinute | 183.867 ns | 0.8148 ns | 0.7621 ns |

Seems like there is no point in new TimeSource if it will update cache data.

I did not create a commit because perfomance with updating cache data is not great, here is a [code](https://gist.github.com/Encamy/d736fffd11fba2e7bc8f03c6c3b9cf27)

What about creating TimeSource that will not check for DST? For example boolean flag in constructor (this will be fast for users without DST and a bit faster than AccurateLocal for users with DST) or rename this time source to AccurateLocal_NoDST?

304NotModified:
> What about creating TimeSource that will not check for DST? For example boolean flag in constructor (this will be fast for users without DST and a bit faster than AccurateLocal for users with DST) or rename this time source to AccurateLocal_NoDST?

I'm not sure if someone is looking for Local time with NoDST, then use UTC IMO

snakefoot:
> I did not create a commit because perfomance with updating cache data is not great, here is a code

You are doing many operations for checking if 1 second has passed. Why not this:

```c#
private DateTime GetTime()
{
  DateTime utc = DateTime.UtcNow;
  int ticksOffset = _ticksOffset;
  long lastUpdatedTicks = Interlocked.Read(ref _lastUpdated);
  long deltaTicks = utc.Ticks - lastUpdatedTicks;
  if (deltaTicks > TimeSpan.TicksPerSecond || deltaTicks < -TimeSpan.TicksPerSecond)
  {
      _ticksOffset = ticksOffset = (DateTime.Now - utc).Ticks;
      _lastUpdated = utc.Ticks;
  }
  long ticks = utc.Ticks + ticksOffset;
  return new DateTime(ticks);
}
```

Encamy:
Thanks, @snakefoot!
Your implementation is better, but why do we need local `long ticksOffset`?
Why we can't use updated cache offset? Is it because of a possible data race in multithreaded environment?

Here is new benchmark. It looks amazing

|              Method |       Mean |     Error |    StdDev |
|-------------------- |-----------:|----------:|----------:|
|           FastLocal |   5.205 ns | 0.0213 ns | 0.0189 ns |
|             FastUtc |   5.202 ns | 0.0266 ns | 0.0249 ns |
|       AccurateLocal | 213.456 ns | 0.5400 ns | 0.5051 ns |
|         AccurateUtc | 144.859 ns | 0.4159 ns | 0.3891 ns |
| FasterAccurateLocal | 158.777 ns | 0.4778 ns | 0.4469 ns |

304NotModified:
PS  On appveyor we build more platforms than locally (xamarin, silverlight).

snakefoot:
> Your implementation is better, but why do we need local long ticksOffset?

I just like local copies when concurrency is in play. Maybe more can be be squezed out with not using Ticks-property constantly.

```c#
private DateTime GetTime()
{
  long utcTicks = DateTime.UtcNow.Ticks;
  long deltaTicks = utcTicks - Interlocked.Read(ref _lastUpdated);
  long ticksOffset = _ticksOffset;
  if (deltaTicks > TimeSpan.TicksPerSecond || deltaTicks < -TimeSpan.TicksPerSecond)
  {
      _ticksOffset = ticksOffset = (DateTime.Now.Ticks - utcTicks);
      _lastUpdated = utcTicks;
  }
  return new DateTime(utcTicks + ticksOffset, DateTimeKind.Local);
}
```

Also ensures the DateTimeKind is correct for the Local-time (Instead of unspecified).

304NotModified:
Thanks for fixes!

Would it make sense to replace the AccurateLocal with FasterAccurateLocal?

snakefoot:
I would probably name it `FastAccurateLocal` it matches `FastLocal` (Instead of `FasterAccurateLocal`)

Encamy:
I think that it can't be named `AccurateLocal`, bacause we use some caching here. If users want to use really accurate and reliable time source, they should use `AccurateLocal` in the way in which it is implemented now.
I think we can name it `FastAccurateLocal` since it is both fast and accurate. Do you agree?
Should I add unit tests?

304NotModified:
Sorry to say, but if it isn't accurate, i'm doubting if we should add this to NLog. 

I'm not sure if the burden of another timesource is the improvement worth (~25% faster)

@snakefoot what do you think?

Encamy:
This time source is accurate. While I was using it I didn't notice any problems
But I think that not everyone will be happy when they notice that their time source has been changed.

snakefoot:
> @snakefoot what do you think?

I prefer not to replace the existing `AccurateLocal`. But `FastAccurateLocal` will never be faster than `AccurateUtc`. Have no opinion about adding another timesource-type. Normally use AccurateUtc myself.

304NotModified:
> This time source is accurate.

OK clear :)

FastAccurateLocal  is then fine to me. UTC is always faster than non-UTC 

@Encamy could you please update https://github.com/NLog/NLog/wiki/Time-Source? Thanks

304NotModified:
OK thanks, so next steps are tests :)

snakefoot:
@Encamy Would it be possible for you to apply the changes from my comment: https://github.com/NLog/NLog/pull/3378#issuecomment-491620318 (Avoid the Ticks-lookup)

Encamy:
I've applied changes from https://github.com/NLog/NLog/pull/3378#issuecomment-491620318 (sorry, I didn't notice it earlier)

I noticed that in [wiki page](https://github.com/NLog/NLog/wiki/Time-Source) results of AccurateUtc are different from my tests. Should I update this value or not?
Here is last test run:

|            Method |       Mean |     Error |    StdDev |     Median |
|------------------ |-----------:|----------:|----------:|-----------:|
|         FastLocal |   5.854 ns | 0.0216 ns | 0.0191 ns |   5.847 ns |
|           FastUtc |   5.839 ns | 0.0197 ns | 0.0165 ns |   5.834 ns |
|     AccurateLocal | 190.481 ns | 0.5205 ns | 0.4869 ns | 190.631 ns |
|       AccurateUtc | 137.258 ns | 0.4862 ns | 0.4310 ns | 137.423 ns |
| FastAccurateLocal | 150.147 ns | 0.1401 ns | 0.1311 ns | 150.150 ns |

I will make unit tests as soon as possible

snakefoot:
> Here is last test run:

Is that on 32 bit / 64 bit ? Net472 / NetCore22 ?

Encamy:
Previous comment is tests in 32bit environment:

BenchmarkDotNet=v0.11.5, OS=Windows 10.0.18362
AMD Ryzen 7 2700, 1 CPU, 16 logical and 8 physical cores
  [Host]     : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.8.3801.0
  DefaultJob : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.8.3801.0

**Results of 64 bit:**
``` ini

BenchmarkDotNet=v0.11.5, OS=Windows 10.0.18362
AMD Ryzen 7 2700, 1 CPU, 16 logical and 8 physical cores
  [Host]     : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.8.3801.0
  DefaultJob : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.8.3801.0


```
|            Method |      Mean |     Error |    StdDev |    Median |
|------------------ |----------:|----------:|----------:|----------:|
|         FastLocal |  3.546 ns | 0.0094 ns | 0.0088 ns |  3.542 ns |
|           FastUtc |  3.541 ns | 0.0015 ns | 0.0012 ns |  3.542 ns |
|     AccurateLocal | 85.927 ns | 1.0766 ns | 1.0070 ns | 85.286 ns |
|       AccurateUtc | 66.451 ns | 0.0659 ns | 0.0617 ns | 66.455 ns |
| FastAccurateLocal | 72.241 ns | 0.2385 ns | 0.2114 ns | 72.164 ns |

snakefoot:
Now things starts to get funny. "66.451 ns" (Slow) vs. "72.241 ns" (Fast).

That is very close in being the same :)

Encamy:
Yes, in 64bit difference between *72.241 ns* and *85.927 ns* (if we are comparing with AccurateLocal timesource) is hilarious. But in 32bit difference is noticeable

snakefoot:
> But in 32bit difference is noticeable

Yes that is funny behavior of the RyuJIT-engine. It mostly optimizes for 64 bit.

Before the RyuJIT-engine, then 32 bit was the fastest one :)

304NotModified:
@Encamy I you prefer to have this included for NLog 4.6.5 (NLog 5 could take some weeks), please rebase on master :)

304NotModified:
I'm closing this as it's inactive for a while. I like to merge this, but I also haven't time for adding the tests.

PR #3386, opened 2019-05-10
Move files into (sub)folders

304NotModified:
Some folders where pretty large.

Moved a lot into (sub) folders. Only moves (+ Resharper config), 0 code changes

for example:

image

and

image

and

image


ericnewton76:
Oh so awesome.  When I was working on the LogManager change I noted in my head it would be nice to do this.

Issue #3390, opened 2019-05-11
Update text of NLog config and schema packages


304NotModified:
Add note that it doesn't work when using package reference + links to example/xsd. 

(see also https://github.com/NLog/NLog/issues/2823)

304NotModified:
Fixed by #3393

PR #3393, opened 2019-05-12
Update package descriptions to note the issues with <PackageReference>

304NotModified:
fixes https://github.com/NLog/NLog/issues/3390


snakefoot:
Maybe just redirect to the tutorial:

https://github.com/NLog/NLog/wiki/Tutorial

Instead the heavy documents:

https://github.com/NLog/NLog/wiki/Configuration-file
https://github.com/NLog/NLog/wiki/Configuration-API

304NotModified:
Good point, but IMO the tutorial is on too high level and also references the nlog config package...

snakefoot:
Maybe the tutorial should not give high recommendation for using NLog.config, but just mention that it exists (and that NLog.config-package only works sometimes, and other times can be a world of hurt).

Issue #3372, opened 2019-05-06
${whenEmtpy} - support IRawValue


304NotModified:
Follow up of https://github.com/NLog/NLog/pull/3373,  see also https://github.com/NLog/NLog/issues/3139

304NotModified:
Fixed by https://github.com/NLog/NLog/pull/3395

PR #3395, opened 2019-05-12
Implement IRawValue on ${whenEmpty}

304NotModified:
fixes https://github.com/NLog/NLog/issues/3372

related: https://github.com/NLog/NLog/pull/3373 and https://github.com/NLog/NLog/issues/3139

### DIFF
--- a/src/NLog/LayoutRenderers/Wrappers/WhenEmptyLayoutRendererWrapper.cs
+++ b/src/NLog/LayoutRenderers/Wrappers/WhenEmptyLayoutRendererWrapper.cs
@@ -31,6 +31,8 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
+using NLog.Internal;
+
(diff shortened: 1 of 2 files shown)


304NotModified:
@snakefoot could you please help me getting this ready?

I doesn't work unfortunately, probably because `MutableUnsafe = true`. See this screenie when debugging (SimpleLayout.cs)

image

Any idea how to fix that? 

Also think this isn't the most efficient way, but dunno how to write a more efficient test.

```c#
   public bool TryGetRawValue(LogEventInfo logEvent, out object value)
        {
            var item = Inner.Render(logEvent);
            if (item.Length > 0)
                 ...

       }
```

snakefoot:
Don't do this:

`var item = Inner.Render(logEvent);`

Instead check if the raw-value-returned is null or empty-string.

snakefoot:
Created #3398

304NotModified:
Syoer

PR #3405, opened 2019-05-17
FilteringTargetWrapper: Add support for batch writing + PostFilteringTargetWrapper: performance optimizations

snakefoot:
And allow use of WhenRepeated as Filter:

```xml
      <target name="default" xsi:type="FilteringWrapper">
        <filter xsi:type="whenRepeated" layout="${message}" timeoutSeconds="30" action="Ignore" />
        <target name="console" xsi:type="Console" />
      </target>
```

Also made some optimizations to `PostFilteringTargetWrapper` so it also supports OptimizeBufferReuse = true.

And also "fixed" `PostFilteringTargetWrapper` so it also works when writing a single LogEvent, without using BufferingWrapper or AsyncWrapper (More intuitive and user-friendly)

### DIFF
--- a/src/NLog/Config/LoggingConfigurationParser.cs
+++ b/src/NLog/Config/LoggingConfigurationParser.cs
@@ -1021,6 +1021,11 @@ namespace NLog.Config
                 return;
             }
 
+            if (SetFilterFromElement(o, propInfo, element))
+            {
+                return;
+            }
+
             SetItemFromElement(o, propInfo, element);
         }
 
@@ -1093,6 +1098,30 @@ namespace NLog.Config
                 return null;
 
             return _configurationItemFactory.Layouts.CreateInstance(ExpandSimpleVariables(layoutTypeName));
+        }
+
+        private bool SetFilterFromElement(object o, PropertyInfo propInfo, ILoggingConfigurationElement layoutElement)
+        {
+            // Check if it is a Filter
+            if (!typeof(Filter).IsAssignableFrom(propInfo.PropertyType))
+                return false;
+
+            // Check if the 'type' attribute has been specified
+            string filterTypeName = GetConfigItemTypeAttribute(layoutElement);
+            if (filterTypeName == null)
+                return false;
+
+            Filter filter = _configurationItemFactory.Filters.CreateInstance(ExpandSimpleVariables(filterTypeName));
+            // and is a Filter and 'type' attribute has been specified
+            if (filter != null)
+            {
+                ConfigureObjectFromAttributes(filter, layoutElement, true);
+                ConfigureObjectFromElement(filter, layoutElement);
+                propInfo.SetValue(o, filter, null);
+                return true;
+            }
+
+            return false;
         }
 
         private void SetItemFromElement(object o, PropertyInfo propInfo, ILoggingConfigurationElement element)
--- a/src/NLog/Config/LoggingConfigurationParser.cs
+++ b/src/NLog/Config/LoggingConfigurationParser.cs
@@ -1070,15 +1070,15 @@ namespace NLog.Config
             return arrayItem;
         }
 
-        private bool SetLayoutFromElement(object o, PropertyInfo propInfo, ILoggingConfigurationElement layoutElement)
+        private bool SetLayoutFromElement(object o, PropertyInfo propInfo, ILoggingConfigurationElement element)
         {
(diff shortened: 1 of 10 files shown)


304NotModified:
Thanks!

Is this a Breaking change? (Removal of ConditionExpression Condition)

snakefoot:
> Is this a Breaking change? (Removal of ConditionExpression Condition)

Nope this property:

> public ConditionExpression Condition { get; set; }

Is now forwarding into the new `Filter`-property using `ConditionBasedFilter`

snakefoot:
@304NotModified Please include for ver. 4.6.4

src/NLog/Internal/ArrayHelper.cs
```diff
@@ -31,19 +31,34 @@
...
-        private static class EmptyArray<T>
+        /// <summary>
+        /// Create a partial list with the items from the parameter <paramref name="items"/>
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="items">items to include into the returning list.</param>
+        /// <param name="untilIndex">until index, not included.</param>
+        /// <returns>New list</returns>
+        /// <remarks>IList as input for performance reasons.</remarks>
+        public static IList<T> CreatePartialList<T>([NotNull] this IList<T> items, int untilIndex)
```
snakefoot: Maybe call this `Take` or `TakeList` ?

304NotModified:
It would be nice if we could deduplicate PostFilteringTargetWrapper.ApplyFilter and FilteringTargetWrapper.Write - any idea how?

src/NLog/Targets/Wrappers/FilteringTargetWrapper.cs
```diff
@@ -90,16 +89,32 @@ public FilteringTargetWrapper(Target wrappedTarget, ConditionExpression conditio
...
             Condition = condition;
-            OptimizeBufferReuse = GetType() == typeof(FilteringTargetWrapper);  // Class not sealed, reduce breaking changes
         }
 
         /// <summary>
         /// Gets or sets the condition expression. Log events who meet this condition will be forwarded 
         /// to the wrapped target.
         /// </summary>
         /// <docgen category='Filtering Options' order='10' />
+        public ConditionExpression Condition { get => (Filter as ConditionBasedFilter)?.Condition; set => Filter = CreateFilter(value); }
```
304NotModified: FYI I created a helper here, so the line would be more readable.

src/NLog/Config/LoggingConfigurationParser.cs
```diff
@@ -1095,6 +1100,30 @@ private Layout TryCreateLayoutInstance(ILoggingConfigurationElement element, Typ
             return _configurationItemFactory.Layouts.CreateInstance(ExpandSimpleVariables(layoutTypeName));
         }
 
+        private bool SetFilterFromElement(object o, PropertyInfo propInfo, ILoggingConfigurationElement layoutElement)
```
304NotModified: Please don't copy paste code.

I've added an abstraction method for this.

snakefoot:
> It would be nice if we could deduplicate PostFilteringTargetWrapper.ApplyFilter and FilteringTargetWrapper.Write - any idea how?

Made a commit

304NotModified:
cool,

what do you think of https://github.com/snakefoot/NLog/pull/8 ? Or are lambdas bad for performance?

snakefoot:
> Or are lambdas bad for performance?

Yes. Those that are doing capture. Was also doing it first. Until the other target comes with a dynamic-filter-object. Then it became a TState-parameter.

304NotModified:
OK! Added some docs and now this is perfect :)

snakefoot:
> OK! Added some docs and now this is perfect IMO :)

Also happy with the result :)

snakefoot:
@304NotModified Guess you need to restart the appveyor-build

304NotModified:
restarted 2nd time 👼

304NotModified:
Oops the idea was to merge this on to release/4.6.4. Anyway, canceled the build so no problem.

snakefoot:
Updated the documentation: https://github.com/NLog/NLog/wiki/FilteringWrapper-target and added link from https://github.com/NLog/NLog/wiki/WhenRepeated-Filter

304NotModified:
Ah this is also an xsd update

Issue #3377, opened 2019-05-07
Missing documentation on rules for NLog.config


mindisk:
I was looking at how to configure logger rules in NLog.config. I wanted to add a name (identifier) to rule. I tried following [Rules](https://github.com/nlog/NLog/wiki/Configuration-file#rules) section and I got confused about how to assign this name to a rule. 
After digging through the code I found a property named *RuleName*. It was not clear from the wiki how this property is assign using `<logger>` tag. I found out that this tag does not support such attribute.  You need to use `<rule>`  tag within  `<rules>` . There was no information in the wiki about this `<rule>` tag and difference between it and the `<logger>` tag.<br />
Example:
`<logger name="*" minlevel="Info" writeTo="defaultlogfile" />`
`<rule name="defaultlogfile" logger="*" minlevel="Info" writeTo="defaultlogfile"/>` 

I am not sure if these two tags can be used simultaneously pointing to different targets. 

All in all, I would ask to update the wiki about `<rule>` tag and the difference between it and the `<logger>`  (if this information is not located in some other place already).

snakefoot:
I created #2891 to resolve  #2394, so I could use it for NLog-config in appsettings.json.

Sadly enough the `<logger>`-tag uses the attribute `name=`, and telling everyone that they now should use `<rule>` just incase they wanted to give their logging-rules an identifier sounded like hard work without any winners.

So I settled for it being an easter-egg for those who need to fiddle with their logging-rules at runtime through code instead of nlog.config.

You are welcome to make a PR makes it possible to do this `<logger name="..." rulename="123" />`, and then make a nice Wiki-page about it.

304NotModified:
There was a mistake in the docs I see:

![image](https://user-images.githubusercontent.com/5808377/58054920-744e3180-7b5c-11e9-94c5-ac764b78d7b3.png)

fixed it:

![image](https://user-images.githubusercontent.com/5808377/58054973-a19adf80-7b5c-11e9-8c31-08d629af15f2.png)

I will also update the XSD

304NotModified:
OK I have to check what is now implemented. Not what I had in mind...

304NotModified:
fixed now:

![image](https://user-images.githubusercontent.com/5808377/58505078-64ea6c00-818c-11e9-97b2-34d0c4cdf9e6.png)

https://github.com/NLog/NLog/wiki/Configuration-file

304NotModified:
thanks for reporting!

Issue #2394, opened 2017-11-16
Support for `RuleName` property  on a `<Rule>`  (LoggingRule class)


qaqz111:
**Type**:

- Feature request

**NLog version**: 4.4.12

**Platform**: .Net 4.5

 - Why do we need it?
The `Id` property of class `LoggingRule` can be used to locate exact rule in code, otherwise, we have to code a lot to find the rule we want:
```C#
            var rule = config.LoggingRules
                .FirstOrDefault(r =>
                    "*" == r.LoggerNamePattern &&
                    false == r.Final &&
                    (null == r.ChildRules || 0 == r.ChildRules.Count) &&
                    (null == r.Filters || 0 == r.Filters.Count) &&
                    (null != r.Levels && 5 == r.Levels.Count) &&
                    (null != r.Targets && 2 == r.Targets.Count));
```
So, if there is an `Id` property, we can do this:
```C#
            var rule = config.LoggingRules
                .FirstOrDefault(r => "ruleYouWant" == r.Id);
```

 - An example of the XML config, if relevant. 
```xml
  <rules>
    <logger name="*" writeTo="trace"/>
    <logger id="ruleYouWant" name="*" minlevel="Debug" writeTo="con,file"/>
  </rules>
```

Not only for `LoggingRule` class, but all that can be accessed by code should have an `Id` property or something else which can be used to locate them in code like Linq calls.

304NotModified:
I like the idea! 

Currently you could use the index as a poor man's solution

qaqz111:
Haha, yes a poor man's solution indeed o(≧▽≦)o

In fact sometimes the index can be of help, but in a more complex application, the index way have a problem is that, we can not ensure that rules added by code always take the same indexes, so currently checking all properties is the most stable way.

I'm looking forward to this feature getting online ;)

304NotModified:
I like the idea, and it's easy to implement, but there is one issue:

```xml
    <logger name="myname" writeTo="trace"/>
```
this means: filter the logs on the name "myname"

```xml
    <logger id="myname" writeTo="trace"/>
```

And this won't filter on a id or something. It could be confusing. Yes the name of `name` attribute if bad, but we can't change it. What do you guys think?

stjephan:
I think id should be ok. If it's explained in the docu it should be enough.
Whats about "alias", "accessId" ,  "apiAccessId" or something like that?

304NotModified:
Im thinking of rule-name or rule-id

[...]

PR #3409, opened 2019-05-20
Various XSD improvements (NLog.Schema package)

304NotModified:
fixes https://github.com/NLog/NLog/issues/3377

- [x] fix LoggerName

related https://github.com/NLog/NLog/issues/2394 and https://github.com/NLog/NLog/pull/2891

### DIFF
--- a/tools/MakeNLogXSD/TemplateXSD.xml
+++ b/tools/MakeNLogXSD/TemplateXSD.xml
@@ -174,9 +174,9 @@
       </xs:annotation>
     </xs:attribute>
 
-    <xs:attribute name="enabled" type="xs:boolean" default="true">
+    <xs:attribute name="ruleName" type="xs:string" use="optional">
       <xs:annotation>
-        <xs:documentation>Enable or disable logging rule. Disabled rules are ignored.</xs:documentation>
+        <xs:documentation>Rule identifier to allow rule lookup with Configuration.FindRuleByName and Configuration.RemoveRuleByName.</xs:documentation>
       </xs:annotation>
     </xs:attribute>
   </xs:complexType>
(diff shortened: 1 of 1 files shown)


snakefoot:
@304NotModified Think you should also fix this method to recognize `RULENAME`:

```c#
        private LoggingRule ParseRuleElement(ILoggingConfigurationElement loggerElement)
        {
            IEnumerable<LogLevel> enableLevels = null;
            int minLevel = 0;
            int maxLevel = LogLevel.MaxLevel.Ordinal;

            string ruleName = null;
            string namePattern = null;
            bool enabled = true;
            bool final = false;
            string writeTargets = null;
            foreach (var childProperty in loggerElement.Values)
            {
                switch (childProperty.Key?.Trim().ToUpperInvariant())
                {
                    case "NAME":
                        if (loggerElement.MatchesName("logger"))
                            namePattern = childProperty.Value; // Legacy Style
                        else
                            ruleName = childProperty.Value;
                        break;
                    case "RULENAME":
                        ruleName = childProperty.Value; // Legacy Style
                        break;
```

304NotModified:
> @304NotModified Think you should also fix this method to recognize `RULENAME`:

This isn't used for the XML, isn't?

304NotModified:
I think we have to update the XSD for NLog 5 and recommend the new approach IMO

snakefoot:
> This isn't an issue in the XML now isn't?

Nope but it would make sense that `rulename` is recognized by the parser, when adding it to the xml-doc :)

304NotModified:
ow this is a mistake then. I'm confused why the names aren't in sync...

snakefoot:
Created #3411

PR #3411, opened 2019-05-21
 LoggingConfigurationParser - Recognize LoggingRule.RuleName property

snakefoot:
See also #3409 and #3377


304NotModified:
Thanks!

PR #3415, opened 2019-05-22
ConcurrentRequestQueue - Fixed deadlock caused by misunderstanding Monitor.Wait

### LINKED ISSUE #3414: Deadlock on ConcurrentRequestQueue.WaitForBelowRequestLimit()

QUESTIONS are preferred on StackOverflow. You could expect a faster response there (but do include all the info below). Use the tag "nlog" https://stackoverflow.com/questions/ask

**NLog version**: 4.6.3

**Platform**: .Net 4.7.2/ .NET Core 2

It happen when set `<default-wrapper xsi:type="AsyncWrapper" overflowAction="Block" forceLockingQueue="false" />`(or netcore2.0 without `forceLockingQueue`). And can be reproducted with the below code.
```cs
            var queue = new NLog.Targets.Wrappers.ConcurrentRequestQueue(10, NLog.Targets.Wrappers.AsyncTargetWrapperOverflowAction.Block);

            for (var j = 0; j < 3; ++j)
            {
                int cur = j;
                _ = Task.Run(() =>
                {
                    for (int i = 0; i < 20; ++i)
                    {
                        queue.Enqueue(default(NLog.Common.AsyncLogEventInfo));
                        Console.WriteLine($"Thread {cur}: PostEnqueue {i + 1}");
                    }
                });
            }
            Task.Run(() =>
            {
                Thread.Sleep(1000);
                int left = 20 * 3;
                while (left > 0)
                {
                    var tmp = queue.DequeueBatch(3);
                    left -= tmp.Length;
                    if (tmp.Length != 0)
                        Console.WriteLine($"DequeueBatch left={left}");
                }
                Console.WriteLine("Finished");
            }).Wait();
```

You can see some threads are block on line133(`Monitor.Enter(_logEventInfoQueue);`), but the queue is empty at that time. Seems `Monitor.Wait()` should not change the value of `lockTaken` at line141.

https://github.com/NLog/NLog/blob/610f36888dda1c5bf023fcb65b344fc10ac15701/src/NLog/Targets/Wrappers/ConcurrentRequestQueue.cs#L118-L160

- What is the current result?
Still block even if the queue is empty at that time.
No block after the queue be empty.
- Are there any workarounds? yes/no

snakefoot:
Resolves #3414

### DIFF
--- a/src/NLog/Targets/Wrappers/ConcurrentRequestQueue.cs
+++ b/src/NLog/Targets/Wrappers/ConcurrentRequestQueue.cs
@@ -127,21 +127,13 @@ namespace NLog.Targets.Wrappers
                 // If yield did not help, then wait on a lock
                 while (currentCount > RequestLimit)
                 {
+                    InternalLogger.Debug("Blocking because the overflow action is Block...");
                     if (!lockTaken)
-                    {
-                        InternalLogger.Debug("Blocking because the overflow action is Block...");
                         Monitor.Enter(_logEventInfoQueue);
-                        lockTaken = true;
-                        InternalLogger.Trace("Entered critical section.");
-                    }
                     else
-                    {
-                        InternalLogger.Debug("Blocking because the overflow action is Block...");
-                        if (!Monitor.Wait(_logEventInfoQueue, 100))
-                            lockTaken = false;
-                        else
-                            InternalLogger.Trace("Entered critical section.");
-                    }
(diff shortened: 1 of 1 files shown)


snakefoot:
Will have to figure out if we can reuse one of the existing unit-tests for stress-testing the async-queue to verify the bug is gone.

snakefoot:
Created  #3416 with unit-test. And also discovered a performance opportunity.

PR #3416, opened 2019-05-22
Async / buffering wrapper: Improve performance when blocking

snakefoot:
ConcurrentRequestQueue - Improved DequeueBatch performance when blocking

Avoid activating throttle logic, when timer-thread is able to keep up. The throttle logic also affects the timer-thread, so only perform throttle-logic when timer-thread is falling behind.

And unit test for deadlock issue

### DIFF
--- a/src/NLog/Targets/Wrappers/ConcurrentRequestQueue.cs
+++ b/src/NLog/Targets/Wrappers/ConcurrentRequestQueue.cs
@@ -93,9 +93,10 @@ namespace NLog.Targets.Wrappers
                                     InternalLogger.Debug("Discarding one element from queue");
                                     currentCount = Interlocked.Decrement(ref _count);
                                     OnLogEventDropped(lostItem.LogEvent);
-                                break;
+                                    break;
                                 }
-                                currentCount = Interlocked.Read(ref _count);
+                                Interlocked.Decrement(ref _count);
+                                currentCount = Interlocked.Increment(ref _count);
                             } while (currentCount > RequestLimit);
                         }
                         break;
@@ -106,13 +107,15 @@ namespace NLog.Targets.Wrappers
                         break;
                     case AsyncTargetWrapperOverflowAction.Grow:
                         {
+                            InternalLogger.Debug("The overflow action is Grow, adding element anyway");
                             OnLogEventQueueGrows(currentCount);
+                            RequestLimit *= 2;
                         }
                         break;
                 }
             }
             _logEventInfoQueue.Enqueue(logEventInfo);
-            return currentCount == 1;
+            return currentCount == 1;    // Inserted first item in empty queue
         }
 
         private long WaitForBelowRequestLimit()
@@ -127,6 +130,7 @@ namespace NLog.Targets.Wrappers
                 // If yield did not help, then wait on a lock
                 while (currentCount > RequestLimit)
                 {
+                    Interlocked.Decrement(ref _count);
                     InternalLogger.Debug("Blocking because the overflow action is Block...");
                     if (!lockTaken)
                         Monitor.Enter(_logEventInfoQueue);
@@ -134,12 +138,7 @@ namespace NLog.Targets.Wrappers
                         Monitor.Wait(_logEventInfoQueue);
                     lockTaken = true;
                     InternalLogger.Trace("Entered critical section.");
-                    currentCount = Interlocked.Read(ref _count);
-                }
-
-                if (lockTaken)
-                {
-                    Monitor.PulseAll(_logEventInfoQueue);
+                    currentCount = Interlocked.Increment(ref _count);
                 }
             }
             finally
@@ -158,6 +157,7 @@ namespace NLog.Targets.Wrappers
             SpinWait spinWait = new SpinWait();
             for (int i = 0; i <= 20; ++i)
             {
+                Interlocked.Decrement(ref _count);
                 if (spinWait.NextSpinWillYield)
                 {
                     if (firstYield)
@@ -166,7 +166,7 @@ namespace NLog.Targets.Wrappers
                 }
 
                 spinWait.SpinOnce();
-                currentCount = Interlocked.Read(ref _count);
+                currentCount = Interlocked.Increment(ref _count);
                 if (currentCount <= RequestLimit)
                     break;
             }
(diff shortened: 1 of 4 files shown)


src/NLog/Targets/Wrappers/ConcurrentRequestQueue.cs
```diff
@@ -223,7 +223,7 @@ public override void DequeueBatch(int count, IList<AsyncLogEventInfo> result)
 
             if (dequeueBatch)
             {
-                bool lockTaken = Monitor.TryEnter(_logEventInfoQueue);    // Try to throttle
+                bool lockTaken = result.Count == count && Monitor.TryEnter(_logEventInfoQueue);    // Only try throttle when falling behind
```
yyjdelete: What will happen if `result.Count` and `count` are not equals, should be below 2 lines 
```cs
for (int i = 0; i < result.Count; ++i)
    Interlocked.Decrement(ref _count);
```
be `Interlocked.Add(ref _count,-count);` instead?

snakefoot: Did not know about `Interlocked.Add` with negative delta. Yes that would generate much less overhead, independent of falling behind or not.

yyjdelete: @snakefoot 
Should it be `-count` instead of `-result.Count` here, if they can be different, to follow the same logic for `!dequeueBatch`

yyjdelete: I have a mistake, I see it as `i` and think it mean the input list is not empty before add here. So it's ok.

304NotModified: I think this a good improvement! Thanks!

tests/NLog.UnitTests/Targets/Wrappers/AsyncTargetWrapperTests.cs
```diff
@@ -108,35 +109,64 @@ private void AsyncTargetWrapperSyncTest_WhenTimeToSleepBetweenBatchesIsEqualToZe
...
-                    Assert.True(prevSequenceID < itemWrittenList[i]);
-                    prevSequenceID = itemWrittenList[i];
+                    if (writer1sequenceID < itemWrittenList[i])
+                    {
+                        Assert.True(writer1sequenceID < itemWrittenList[i]);
+                        writer1sequenceID = itemWrittenList[i];
+                    }
+                    else
+                    {
+                        Assert.True(writer2sequenceID < itemWrittenList[i]);
```
304NotModified: Please add a message, as we have multiple assert.trues in the test method (and if they fail, it will be unclear which one)

tests/NLog.UnitTests/Targets/Wrappers/AsyncTargetWrapperTests.cs
```diff
@@ -108,35 +109,64 @@ private void AsyncTargetWrapperSyncTest_WhenTimeToSleepBetweenBatchesIsEqualToZe
                 int flushCounter = 0;
                 AsyncContinuation flushHandler = (ex) => { ++flushCounter; };
 
-                List<KeyValuePair<LogEventInfo, AsyncContinuation>> itemPrepareList = new List<KeyValuePair<LogEventInfo, AsyncContinuation>>(500);
+                List<AsyncLogEventInfo> itemPrepareList = new List<AsyncLogEventInfo>(500);
```
304NotModified: `var` is also good to me ;)

src/NLog/Targets/Wrappers/ConcurrentRequestQueue.cs
```diff
@@ -223,11 +224,10 @@ public override void DequeueBatch(int count, IList<AsyncLogEventInfo> result)
...
```
304NotModified: I think this is OK, but maybe rename lockTaken? 

proposal: needsPulse

snakefoot: No it is `lockTaken`, because it needs to release it afterwards. Pulse is just a bonus-operation to be nice.

304NotModified: ow yes you're right!

tests/NLog.UnitTests/Targets/Wrappers/AsyncTargetWrapperTests.cs
snakefoot: Done

tests/NLog.UnitTests/Targets/Wrappers/AsyncTargetWrapperTests.cs
snakefoot: Made the code more readable.

snakefoot:
Apparently hard to make a stable unit-test with multi-threading on the build-engines. One more try.

snakefoot:
Hrmm. Don't like how the NetCoreApp2 unit test have become very unstable. Need to be checked.

304NotModified:
Yes please check. Unstable tests are really slowing down development and releases. Maybe we should partly mock the unit test.

304NotModified:
It's probably now more an integration test, like most "unit tests" in NLog.

snakefoot:
Actually think it is a real bug, that has been revealed by the optimization of the dequeue-operation for the timer-thread. But right now the appveyor-build-queue has become very long, so it will have to look at it another day.

304NotModified:
Maybe it's easier to run only on your appveyor account for this? Then you could stop/restart it yourself

snakefoot:
Think I have found the bug. After we have changed so blocking decrement/increments, then it means that timer-thread no longer automatically reschedules itself because it will not "see" producers by non-zero _count.

It is important now that the producers that wake up will remember to start the consumer-timer-thread, if needed.

Two producers that are blocked will get into `TrySpinWaitForLowerCount()` and it only performs Interlocked.Read. This means two producers will return with currentCount == 2. But one of them should have returned with currentCount = 1.

This causes both producers to say that they don't need to start the consumer-timer-thread

304NotModified:
makes sense!

Thanks for figuring out and fixing this!

PR #3423, opened 2019-05-24
${all-event-properties}: Added IncludeEmptyValues option

304NotModified:
suggested here: https://stackoverflow.com/questions/56298028/how-to-handle-empty-values-produced-by-nlogs-all-event-properties-layout-render


304NotModified:
docs done: https://github.com/NLog/NLog/wiki/All-Event-Properties-Layout-Renderer

Issue #3422, opened 2019-05-24
Feature: list contextProperties of Logger


304NotModified:
Make _contextProperties public on the Logger object.

Discussion: read-only or mutable?

asked here: https://stackoverflow.com/questions/56278360/nlog-list-property-values

snakefoot:
> Discussion: read-only or mutable?

I vote for readonly using `IReadOnlyDictionary` (Only have the property on platforms where interface is supported).

304NotModified:
I checked the code. I think IDictionary is flexible enough. If we need to know changes in the dictionary, we could implement this.

snakefoot:
Doesn't make sense. You can modify using the IDictionary-interface. It is way too flexible. You will get `Collection was modified`-exceptions

304NotModified:
Do we need also a clone then? (I think so)

snakefoot:
> Do we need also a clone then? (I think so)

Would be very confusing to get a mutable dictionary returned, and your changes doesn't have any effect.

I continue to vote for `IReadOnlyDictionary`

304NotModified:
That's clear ;)

But besides IReadOnlyDictionary we need a clone then. Otherwise your could cast it back. Unfortunately, you could always change and object living in the dictionary

snakefoot:
Trying to protect yourself from a developer is hard work, and seldom gives any bonus.

Developer can also do reflection and access the dictionary that way.

The goal is to provide an easy to understand interface.

304NotModified:
fixed by #3430

PR #3424, opened 2019-05-24
Expose logger properties on Logger

304NotModified:
fixes https://github.com/NLog/NLog/issues/3422 + refactor

### DIFF
--- a/src/NLog/Logger.cs
+++ b/src/NLog/Logger.cs
@@ -83,6 +83,11 @@ namespace NLog
         public LogFactory Factory { get; private set; }
 
         /// <summary>
+        /// Properties added with <see cref="WithProperty"/> or <see cref="SetProperty"/>
+        /// </summary>
+        public IDictionary<string, object> Properties => _contextProperties ?? new Dictionary<string, object>();
+
+        /// <summary>
         /// Gets a value indicating whether logging is enabled for the specified level.
         /// </summary>
         /// <param name="level">Log level to be checked.</param>
--- a/src/NLog/Logger.cs
+++ b/src/NLog/Logger.cs
@@ -82,10 +82,14 @@ namespace NLog
         /// </summary>
         public LogFactory Factory { get; private set; }
 
+#if !NET3_5 && !NET4_0 && !SILVERLIGHT && !WINDOWS_PHONE
+
         /// <summary>
         /// Properties added with <see cref="WithProperty"/> or <see cref="SetProperty"/>
         /// </summary>
-        public IDictionary<string, object> Properties => _contextProperties ?? new Dictionary<string, object>();
+        public IReadOnlyDictionary<string, object> Properties => CreateContextPropertiesDictionary(_contextProperties); //clone for immutable
+
+#endif
 
         /// <summary>
         /// Gets a value indicating whether logging is enabled for the specified level.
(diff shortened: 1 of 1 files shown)


src/NLog/Logger.cs
```diff
@@ -82,6 +82,11 @@ protected internal Logger()
         /// </summary>
         public LogFactory Factory { get; private set; }
 
+        /// <summary>
+        /// Properties added with <see cref="WithProperty"/> or <see cref="SetProperty"/>
+        /// </summary>
+        public IDictionary<string, object> Properties => _contextProperties ?? new Dictionary<string, object>();
```
snakefoot: You are opening a door to race-condition-hell by returning an unprotected dictionary.

304NotModified: And that race condition isn't there with WithProperty/SetProperty

snakefoot: I know ` WithProperty/SetProperty` are nice and sweet (because I made them so). But the evil user that starts calling the IDictionary-inteface to insert/remove elements will break everything.

304NotModified: Sorry, I don't see what it should break. Could you please give an example ?

snakefoot:
```c#
var logger = LogManager.GetLogger("Hello").WithProperty("Test", 0);
var thread = new Thread(() => {
for (int i = 1; i < 1000000; ++i)
{
   logger.Properties.Remove("Test");
   logger.Properties.Add("Test", i);
}
});
thread.Start();
for (int i = 0; i < 1000000; ++i)
      logger.Info("Hello World");  // Collection was modified
```

src/NLog/Logger.cs
snakefoot: Very confused that you don't know what happens if one enumerates a collection while another modifies the collection. Made some pseudo-code in new comment.

304NotModified:
We could change it to a concurrent dictionary? (Could, not saying we should)

snakefoot:
> We could change it to a concurrent dictionary? (Could, not saying we should)

That would be a bad trade, regarding logging-performance. Since the current reason for adding Properties-properties is to make it easier to see the current values.

Ofcourse if the IDictionary-logic just created for GDC was extracted then we would have a thread-safe idictionary that could be used here.

src/NLog/Logger.cs
```diff
@@ -82,10 +82,14 @@ protected internal Logger()
         /// </summary>
         public LogFactory Factory { get; private set; }
 
+#if !NET3_5 && !NET4_0 && !SILVERLIGHT && !WINDOWS_PHONE
```
snakefoot: Maybe just change to `NET4_5` ?

304NotModified: I prefer a blacklist, as that will work better for the future

src/NLog/Logger.cs
```diff
@@ -128,16 +138,15 @@ public void SetProperty(string propertyKey, object propertyValue)
             if (string.IsNullOrEmpty(propertyKey))
                 throw new ArgumentException(nameof(propertyKey));
 
-            _contextProperties = CopyOnWrite(propertyKey, propertyValue);
+            _contextProperties = CreateContextPropertiesDictionary(_contextProperties);
+            _contextProperties[propertyKey] = propertyValue;
```
snakefoot: This is not good. As you are modifying the active dictionary.

304NotModified: not really, CreateContextPropertiesDictionary made a new reference. But anyway, I see this PR is already superseded 👼

304NotModified:
superseded by https://github.com/NLog/NLog/pull/3430

Issue #2529, opened 2018-01-21
Remove obsolete properties for NLog 5.0


snakefoot:
**LoggerConfiguration.cs:**
```
        [Obsolete("This property marked obsolete before v4.3.11 and it will be removed in NLog 5.")]
        public bool ExceptionLoggingOldStyle { get; private set; }
```

**ConfigurationItemFactory.cs**
```
        [Obsolete("Use JsonConverter property instead. Marked obsolete on NLog 4.5")]
        public IJsonSerializer JsonSerializer
```

**LoggingConfigurationChangedEventArgs.cs**
```
        [Obsolete("This option will be removed in NLog 5. Marked obsolete on NLog 4.5")]
        public LoggingConfiguration OldConfiguration => ActivatedConfiguration;

        [Obsolete("This option will be removed in NLog 5. Marked obsolete on NLog 4.5")]
        public LoggingConfiguration NewConfiguration => DeactivatedConfiguration;
```

**FileTarget.cs**
```
        [Obsolete("This option will be removed in NLog 5. Marked obsolete on NLog 4.5")]
        public int maxLogFilenames { get; set; }
```

304NotModified:
Thanks for the summary!

PR #3429, opened 2019-05-24
Remove obsolete methodes, properties and classes

304NotModified:
fixes https://github.com/NLog/NLog/issues/2529, fixes https://github.com/NLog/NLog/issues/531


snakefoot:
If the Logger-interface is no longer sacred. Then I think LoggerV1-interface should be removed

304NotModified:
Good one, please create a new issue for that, we could check that later (for v5)

snakefoot:
Created  #3432

snakefoot:
Searched the Wiki for the deprecated properties that no longer exists in NLog 5.0, but found no traces in the Wiki.

PR #3431, opened 2019-05-25
Added support for StackTraceUsage With CallSite only (e.g. Caller Information attributes) + Improve CLScomplaint

### LINKED ISSUE #2059: Fluent API Method not found: 'Void NLog.Fluent.LogBuilder.Write()'.
**Type** Bug

**NLog version**: (e.g. 4.4.5)

**Platform**: .Net 4.0 / .Net 4.5 

In case of a BUG:

- What is the current result?
  System.MissingMethodException: Method not found: 'Void NLog.Fluent.LogBuilder.Write()'.
  success log write
  No
  N/A
  No
  No
  Dont know, that would depend on mocking platform, since this is related to usage with multiple target builds.

- steps to replicate the issue: 
 Have a lib compiled against NLog 4.4.5 net40, using 

have inside code 
_logger.Info()
       .Message("This is a test fluent message '{0}'.", DateTime.Now.Ticks)
       .Property("Test", "InfoWrite")
       .Write();

then use said lib with application compiled against NLog 4.4.5 net45, 
when running method 'Void NLog.Fluent.LogBuilder.Write()'. is produced.

- I believe that this is because of the signature of the method in NLog 4.4.5 net45 was change by using optional arguments, making it not able to find appropriate overload method. 

i propose one of 2 alternatives
 - remove usage of optional arguments for discrete specification, in  net45 ( this will MOST likely work) , 
 - Add overload in net40 will all possible arguments ( would have to try it out )

snakefoot:
Extracted the callsite optimization from #2527, but excluded the special LoggerCallSite. And changed `enum StackTraceUsage` into a Flags-enum.

This allows NLog to make optimal use of caller-member-attributes to provide Callsite-information. Just need to figure out how to make a nice way to provide the info besides the fluent LogBuilder-interface.

Resolves partly #532 when using LogBuilder and resolves #2059

### DIFF
--- a/src/NLog/Config/StackTraceUsage.cs
+++ b/src/NLog/Config/StackTraceUsage.cs
@@ -42,35 +42,36 @@ namespace NLog.Config
     public enum StackTraceUsage
     {
         /// <summary>
-        /// Stack trace should not be captured.
+        /// No Stack trace needs to be captured.
         /// </summary>
         None = 0,
 
         /// <summary>
-        /// Stack trace should be captured
+        /// Stack trace should be captured. This option won't add the filenames and linenumbers
         /// </summary>
         WithStackTrace = 1,
 
 #if !SILVERLIGHT
         /// <summary>
-        /// Source-level information should be capture (source-filename + linenumber)
+        /// Capture also filenames and linenumbers
         /// </summary>
         WithFileNameAndLineNumber = 2,
 #endif
 
         /// <summary>
-        /// Stack trace should only be captured if callsite details are missing
+        /// Capture the location of the call
         /// </summary>
         WithCallSite = 4,
 
         /// <summary>
-        /// Stack trace should be captured without source-level information.
+        /// Stack trace should be captured. This option won't add the filenames and linenumbers.
         /// </summary>
+        [Obsolete("Replace with `WithStackTrace`. Will be removed in NLog 6")]
         WithoutSource = WithStackTrace,
 
 #if !SILVERLIGHT
         /// <summary>
-        /// Stack trace should be captured including source-level information such as line numbers.
+        /// Stack trace should be captured including filenames and linenumbers.
         /// </summary>
         WithSource = WithStackTrace | WithFileNameAndLineNumber,
 
--- a/src/NLog/Config/StackTraceUsage.cs
+++ b/src/NLog/Config/StackTraceUsage.cs
@@ -55,7 +55,7 @@ namespace NLog.Config
         /// <summary>
         /// Source-level information should be capture (source-filename + linenumber)
         /// </summary>
-        WithSourceCode = 2,
+        WithFileNameAndLineNumber = 2,
 #endif
 
         /// <summary>
@@ -72,7 +72,7 @@ namespace NLog.Config
         /// <summary>
         /// Stack trace should be captured including source-level information such as line numbers.
         /// </summary>
-        WithSource = WithStackTrace | WithSourceCode,
+        WithSource = WithStackTrace | WithFileNameAndLineNumber,
 
         /// <summary>
         /// Capture maximum amount of the stack trace information supported on the platform.
--- a/src/NLog/Config/StackTraceUsage.cs
+++ b/src/NLog/Config/StackTraceUsage.cs
@@ -33,36 +33,56 @@
 
 namespace NLog.Config
 {
+    using System;
+
     /// <summary>
     /// Value indicating how stack trace should be captured when processing the log event.
     /// </summary>
+    [Flags]
     public enum StackTraceUsage
     {
         /// <summary>
         /// Stack trace should not be captured.
         /// </summary>
-        None = 0, 
+        None = 0,
+
+        /// <summary>
+        /// Stack trace should be captured
+        /// </summary>
+        WithStackTrace = 1,
+
+#if !SILVERLIGHT
+        /// <summary>
+        /// Source-level information should be capture (source-filename + linenumber)
+        /// </summary>
+        WithSourceCode = 2,
+#endif
+
+        /// <summary>
+        /// Stack trace should only be captured if callsite details are missing
+        /// </summary>
+        WithCallSite = 4,
 
         /// <summary>
         /// Stack trace should be captured without source-level information.
         /// </summary>
-        WithoutSource = 1,
+        WithoutSource = WithStackTrace,
 
 #if !SILVERLIGHT
         /// <summary>
         /// Stack trace should be captured including source-level information such as line numbers.
         /// </summary>
-        WithSource = 2,
+        WithSource = WithStackTrace | WithSourceCode,
 
         /// <summary>
         /// Capture maximum amount of the stack trace information supported on the platform.
         /// </summary>
-        Max = 2,
+        Max = WithSource,
 #else
         /// <summary>
         /// Capture maximum amount of the stack trace information supported on the platform.
         /// </summary>
-        Max = 1,
+        Max = WithoutSource,
 #endif
     }
 }
--- a/src/NLog/Fluent/LogBuilder.cs
+++ b/src/NLog/Fluent/LogBuilder.cs
@@ -285,19 +285,11 @@ namespace NLog.Fluent
         /// <param name="callerMemberName">The method or property name of the caller to the method. This is set at by the compiler.</param>
         /// <param name="callerFilePath">The full path of the source file that contains the caller. This is set at by the compiler.</param>
         /// <param name="callerLineNumber">The line number in the source file at which the method is called. This is set at by the compiler.</param>
+#if NET4_5
         public void Write(
-#if NET4_5
-            [CallerMemberName]
-#endif
-            string callerMemberName = null,
-#if NET4_5
-            [CallerFilePath]
-#endif
-            string callerFilePath = null,
-#if NET4_5
-            [CallerLineNumber]
-#endif
-            int callerLineNumber = 0)
+            [CallerMemberName]string callerMemberName = null,
+            [CallerFilePath]string callerFilePath = null,
+            [CallerLineNumber]int callerLineNumber = 0)
         {
             if (!_logger.IsEnabled(_logEvent.Level))
                 return;
@@ -306,7 +298,16 @@ namespace NLog.Fluent
 
             _logger.Log(_logEvent);
         }
-        
+#else
+        public void Write(
+            string callerMemberName = null,
+            string callerFilePath = null,
+            int callerLineNumber = 0)
+        {
+            _logger.Log(_logEvent);
+        }
+#endif
+
         /// <summary>
         /// Writes the log event to the underlying logger if the condition delegate is true.
         /// </summary>
@@ -314,20 +315,12 @@ namespace NLog.Fluent
         /// <param name="callerMemberName">The method or property name of the caller to the method. This is set at by the compiler.</param>
         /// <param name="callerFilePath">The full path of the source file that contains the caller. This is set at by the compiler.</param>
         /// <param name="callerLineNumber">The line number in the source file at which the method is called. This is set at by the compiler.</param>
+#if NET4_5
         public void WriteIf(
             Func<bool> condition,
-#if NET4_5
-            [CallerMemberName]
-#endif
-            string callerMemberName = null,
-#if NET4_5
-            [CallerFilePath]
-#endif
-            string callerFilePath = null,
-#if NET4_5
-            [CallerLineNumber]
-#endif
-            int callerLineNumber = 0)
+            [CallerMemberName]string callerMemberName = null,
+            [CallerFilePath]string callerFilePath = null,
+            [CallerLineNumber]int callerLineNumber = 0)
         {
             if (condition == null || !condition() || !_logger.IsEnabled(_logEvent.Level))
                 return;
@@ -336,6 +329,19 @@ namespace NLog.Fluent
 
             _logger.Log(_logEvent);
         }
+#else
+        public void WriteIf(
+            Func<bool> condition,
+            string callerMemberName = null,
+            string callerFilePath = null,
+            int callerLineNumber = 0)
+        {
+            if (condition == null || !condition() || !_logger.IsEnabled(_logEvent.Level))
+                return;
+
+            _logger.Log(_logEvent);
+        }
+#endif
 
         /// <summary>
         /// Writes the log event to the underlying logger if the condition is true.
@@ -344,20 +350,12 @@ namespace NLog.Fluent
         /// <param name="callerMemberName">The method or property name of the caller to the method. This is set at by the compiler.</param>
         /// <param name="callerFilePath">The full path of the source file that contains the caller. This is set at by the compiler.</param>
         /// <param name="callerLineNumber">The line number in the source file at which the method is called. This is set at by the compiler.</param>
+#if NET4_5
         public void WriteIf(
             bool condition,
-#if NET4_5
-            [CallerMemberName]
-#endif
-            string callerMemberName = null,
-#if NET4_5
-            [CallerFilePath]
-#endif
-            string callerFilePath = null,
-#if NET4_5
-            [CallerLineNumber]
-#endif
-            int callerLineNumber = 0)
+            [CallerMemberName]string callerMemberName = null,
+            [CallerFilePath]string callerFilePath = null,
+            [CallerLineNumber]int callerLineNumber = 0)
         {
             if (!condition || !_logger.IsEnabled(_logEvent.Level))
                 return;
@@ -366,6 +364,19 @@ namespace NLog.Fluent
 
             _logger.Log(_logEvent);
         }
+#else
+        public void WriteIf(
+            bool condition,
+            string callerMemberName = null,
+            string callerFilePath = null,
+            int callerLineNumber = 0)
+        {
+            if (!condition || !_logger.IsEnabled(_logEvent.Level))
+                return;
+
+            _logger.Log(_logEvent);
+        }
+#endif
 
         private void SetCallerInfo(string callerMemberName, string callerFilePath, int callerLineNumber)
         {
--- a/src/NLog/Fluent/LogBuilder.cs
+++ b/src/NLog/Fluent/LogBuilder.cs
@@ -279,7 +279,6 @@ namespace NLog.Fluent
             return this;
         }
 
-#if NET4_5
         /// <summary>
         /// Writes the log event to the underlying logger.
         /// </summary>
@@ -287,36 +286,27 @@ namespace NLog.Fluent
         /// <param name="callerFilePath">The full path of the source file that contains the caller. This is set at by the compiler.</param>
         /// <param name="callerLineNumber">The line number in the source file at which the method is called. This is set at by the compiler.</param>
         public void Write(
-            [CallerMemberName]string callerMemberName = null,
-            [CallerFilePath]string callerFilePath = null,
-            [CallerLineNumber]int callerLineNumber = 0)
+#if NET4_5
+            [CallerMemberName]
+#endif
+            string callerMemberName = null,
+#if NET4_5
+            [CallerFilePath]
+#endif
+            string callerFilePath = null,
+#if NET4_5
+            [CallerLineNumber]
+#endif
+            int callerLineNumber = 0)
         {
             if (!_logger.IsEnabled(_logEvent.Level))
                 return;
 
-            // TODO NLog ver. 5 - Remove these properties
-            if (callerMemberName != null)
-                Property("CallerMemberName", callerMemberName);
-            if (callerFilePath != null)
-                Property("CallerFilePath", callerFilePath);
-            if (callerLineNumber != 0)
-                Property("CallerLineNumber", callerLineNumber);
-
-            _logEvent.SetCallerInfo(null, callerMemberName, callerFilePath, callerLineNumber);
+            SetCallerInfo(callerMemberName, callerFilePath, callerLineNumber);
 
             _logger.Log(_logEvent);
         }
-#else
-        /// <summary>
-        /// Writes the log event to the underlying logger.
-        /// </summary>
-        public void Write()
-        {
-            _logger.Log(_logEvent);
-        }
-#endif
-
-#if NET4_5
+        
         /// <summary>
         /// Writes the log event to the underlying logger if the condition delegate is true.
         /// </summary>
@@ -326,39 +316,27 @@ namespace NLog.Fluent
         /// <param name="callerLineNumber">The line number in the source file at which the method is called. This is set at by the compiler.</param>
         public void WriteIf(
             Func<bool> condition,
-            [CallerMemberName]string callerMemberName = null,
-            [CallerFilePath]string callerFilePath = null,
-            [CallerLineNumber]int callerLineNumber = 0)
-        {
-            if (condition == null || !condition() || !_logger.IsEnabled(_logEvent.Level))
-                return;
-
-            if (callerMemberName != null)
-                Property("CallerMemberName", callerMemberName);
-            if (callerFilePath != null)
-                Property("CallerFilePath", callerFilePath);
-            if (callerLineNumber != 0)
-                Property("CallerLineNumber", callerLineNumber);
-
-            _logEvent.SetCallerInfo(null, callerMemberName, callerFilePath, callerLineNumber);
-
-            _logger.Log(_logEvent);
-        }
-#else
-        /// <summary>
-        /// Writes the log event to the underlying logger.
-        /// </summary>
-        /// <param name="condition">If condition is true, write log event; otherwise ignore event.</param>
-        public void WriteIf(Func<bool> condition)
-        {
-            if (condition == null || !condition() || !_logger.IsEnabled(_logEvent.Level))
-                return;
-
-            _logger.Log(_logEvent);
-        }
-#endif
-
 #if NET4_5
+            [CallerMemberName]
+#endif
+            string callerMemberName = null,
+#if NET4_5
+            [CallerFilePath]
+#endif
+            string callerFilePath = null,
+#if NET4_5
+            [CallerLineNumber]
+#endif
+            int callerLineNumber = 0)
+        {
+            if (condition == null || !condition() || !_logger.IsEnabled(_logEvent.Level))
+                return;
+
+            SetCallerInfo(callerMemberName, callerFilePath, callerLineNumber);
+
+            _logger.Log(_logEvent);
+        }
+
         /// <summary>
         /// Writes the log event to the underlying logger if the condition is true.
         /// </summary>
@@ -367,14 +345,31 @@ namespace NLog.Fluent
         /// <param name="callerFilePath">The full path of the source file that contains the caller. This is set at by the compiler.</param>
         /// <param name="callerLineNumber">The line number in the source file at which the method is called. This is set at by the compiler.</param>
         public void WriteIf(
-            bool condition, 
-            [CallerMemberName]string callerMemberName = null, 
-            [CallerFilePath]string callerFilePath = null, 
-            [CallerLineNumber]int callerLineNumber = 0)
+            bool condition,
+#if NET4_5
+            [CallerMemberName]
+#endif
+            string callerMemberName = null,
+#if NET4_5
+            [CallerFilePath]
+#endif
+            string callerFilePath = null,
+#if NET4_5
+            [CallerLineNumber]
+#endif
+            int callerLineNumber = 0)
         {
             if (!condition || !_logger.IsEnabled(_logEvent.Level))
                 return;
 
+            SetCallerInfo(callerMemberName, callerFilePath, callerLineNumber);
+
+            _logger.Log(_logEvent);
+        }
+
+        private void SetCallerInfo(string callerMemberName, string callerFilePath, int callerLineNumber)
+        {
+            // TODO NLog ver. 5 - Remove these properties
             if (callerMemberName != null)
                 Property("CallerMemberName", callerMemberName);
             if (callerFilePath != null)
@@ -382,23 +377,8 @@ namespace NLog.Fluent
             if (callerLineNumber != 0)
                 Property("CallerLineNumber", callerLineNumber);
 
-            _logEvent.SetCallerInfo(null, callerMemberName, callerFilePath, callerLineNumber);
-
-            _logger.Log(_logEvent);
+            if (callerMemberName != null || callerFilePath != null || callerLineNumber != 0)
+                _logEvent.SetCallerInfo(null, callerMemberName, callerFilePath, callerLineNumber);
         }
-#else
-        /// <summary>
-        /// Writes the log event to the underlying logger.
-        /// </summary>
-        /// <param name="condition">If condition is true, write log event; otherwise ignore event.</param>
-        public void WriteIf(bool condition)
-        {
-            if (!condition || !_logger.IsEnabled(_logEvent.Level))
-                return;
-
-            _logger.Log(_logEvent);
-        }
-#endif
-
     }
 }
--- a/src/NLog/Fluent/LogBuilder.cs
+++ b/src/NLog/Fluent/LogBuilder.cs
@@ -378,18 +378,18 @@ namespace NLog.Fluent
         }
 #endif
 
-        private void SetCallerInfo(string callerMemberName, string callerFilePath, int callerLineNumber)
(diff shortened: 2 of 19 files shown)


snakefoot:
Surprised by the special methods in `Log.cs`. I hope no one are using them as they must be crazy slow with disk-access for file-lookup on every LogEvent.

src/NLog/Config/StackTraceUsage.cs
```diff
@@ -41,28 +41,38 @@ public enum StackTraceUsage
...
         /// Stack trace should be captured without source-level information.
         /// </summary>
-        WithoutSource = 1,
+        WithoutSource = 3,
 
 #if !SILVERLIGHT
+        /// <summary>
+        /// Stack trace (with source) should only be captured if callsite details are missing
+        /// </summary>
+        WithCallSiteSource = 2,
```
304NotModified: please reoder the values. 

would it be better if we change this to a flags enum? e.g. WithCallSite | WithCallSiteSource

src/NLog/Fluent/LogBuilder.cs
```diff
@@ -279,17 +279,25 @@ public LogBuilder StackTrace(StackTrace stackTrace, int userStackFrame)
...
-            [CallerLineNumber]int callerLineNumber = 0)
+#if NET4_5
+            [CallerMemberName]
+#endif
+            string callerMemberName = null,
+#if NET4_5
+            [CallerFilePath]
+#endif
+            string callerFilePath = null,
+#if NET4_5
```
304NotModified: maybe it was to long ago, but why only .net 4.5?

src/NLog/Fluent/LogBuilder.cs
```diff
@@ -302,21 +310,12 @@ public LogBuilder StackTrace(StackTrace stackTrace, int userStackFrame)
             if (callerLineNumber != 0)
                 Property("CallerLineNumber", callerLineNumber);
 
-            _logEvent.SetCallerInfo(null, callerMemberName, callerFilePath, callerLineNumber);
+            if (callerMemberName != null || callerFilePath != null || callerLineNumber != 0)
```
304NotModified: if one of them isnot set, then we ignore them all?

it think `||` should be `&&`

src/NLog/Fluent/LogBuilder.cs
```diff
@@ -367,38 +363,35 @@ public void WriteIf(Func<bool> condition)
...
+            // TODO NLog 5.0 Remove these properties
             if (callerMemberName != null)
                 Property("CallerMemberName", callerMemberName);
             if (callerFilePath != null)
                 Property("CallerFilePath", callerFilePath);
             if (callerLineNumber != 0)
                 Property("CallerLineNumber", callerLineNumber);
 
-            _logEvent.SetCallerInfo(null, callerMemberName, callerFilePath, callerLineNumber);
+            if (callerMemberName != null || callerFilePath != null || callerLineNumber != 0)
```
304NotModified: please extract a method for this.

src/NLog/Fluent/LogBuilder.cs
```diff
@@ -367,38 +363,35 @@ public void WriteIf(Func<bool> condition)
...
+            string callerFilePath = null,
+#if NET4_5
+            [CallerLineNumber]
+#endif
+            int callerLineNumber = 0)
         {
             if (!condition || !_logger.IsEnabled(_logEvent.Level))
                 return;
 
+            // TODO NLog 5.0 Remove these properties
```
304NotModified: this is for NLog 5? So we could remove this?

src/NLog/Internal/TargetWithFilterChain.cs
```diff
@@ -50,6 +51,115 @@ internal class TargetWithFilterChain
...
+            }
+
+            /// <summary>
+            /// Determines if two objects of the same type are equal in value.
+            /// </summary>
+            /// <param name="other">Other object to compare to.</param>
+            /// <returns>True if objects are equal, false otherwise.</returns>
+            public bool Equals(CallSiteKey other)
+            {
+                return FileSourceLineNumber == other.FileSourceLineNumber
```
304NotModified: I think it's missing the null comparisons, See https://docs.microsoft.com/en-us/previous-versions/ms173147(v=vs.90)

PS: Resharper could generate equals and hashcode

src/NLog/Internal/TargetWithFilterChain.cs
```diff
@@ -50,6 +51,115 @@ internal class TargetWithFilterChain
...
+        internal bool TryCallSiteClassNameOptimization(StackTraceUsage stackTraceUsage, LogEventInfo logEvent)
+        {
+            if (stackTraceUsage != StackTraceUsage.WithCallSite
+#if !SILVERLIGHT
+             && stackTraceUsage != StackTraceUsage.WithCallSiteSource
+#endif
+               )
+                return false;
+
+            if (logEvent.CallSiteInformation == null)
```
304NotModified: we prefer to combine this with below, `  if (string.IsNullOrEmpty(logEvent.CallSiteInformation?.CallerFilePath))`

src/NLog/Internal/TargetWithFilterChain.cs
```diff
@@ -50,6 +51,115 @@ internal class TargetWithFilterChain
...
+            if (string.IsNullOrEmpty(className))
+                return false;
+
+            if (_callSiteClassNameCache == null)
+                return false;
+
+            string internClassName = logEvent.LoggerName == className ?
+                logEvent.LoggerName :
+#if !NETSTANDARD1_0
+                string.Intern(className);
```
304NotModified: why do we need this? Please add some code comments, as interning is really not necessary is most applications

src/NLog/Internal/TargetWithFilterChain.cs
```diff
@@ -50,6 +51,115 @@ internal class TargetWithFilterChain
...
+                return false;
+
+            if (string.IsNullOrEmpty(logEvent.CallSiteInformation.CallerFilePath))
+                return false;
+
+            string className = logEvent.CallSiteInformation.GetCallerClassName(null, true, true, true);
+            if (string.IsNullOrEmpty(className))
+                return false;
+
+            if (_callSiteClassNameCache == null)
```
304NotModified: I think this check should be earlier in this method (cheaper check)

src/NLog/Internal/TargetWithFilterChain.cs
```diff
@@ -50,6 +51,115 @@ internal class TargetWithFilterChain
...
+                return true;
+
+            if (_callSiteClassNameCache == null)
+            {
+                System.Threading.Interlocked.CompareExchange(ref _callSiteClassNameCache, new MruCache<CallSiteKey, string>(1000), null);
+            }
+            CallSiteKey callSiteKey = new CallSiteKey(logEvent.CallerMemberName, logEvent.CallerFilePath, logEvent.CallerLineNumber);
+            return _callSiteClassNameCache.TryGetValue(callSiteKey, out callSiteClassName);
+        }
+        private MruCache<CallSiteKey, string> _callSiteClassNameCache;
```
304NotModified: please add fields as first in the class

src/NLog/LayoutRenderers/CallSite and stacktraces/CallSiteLayoutRenderer.cs
```diff
@@ -130,12 +130,10 @@ StackTraceUsage IUsesStackTrace.StackTraceUsage
...
                 if (FileName)
-                {
-                    return StackTraceUsage.Max;
-                }
+                    return SkipFrames == 0 ? StackTraceUsage.WithCallSiteSource : StackTraceUsage.WithSource;
+                else
 #endif
-
-                return StackTraceUsage.WithoutSource;
+                    return SkipFrames == 0 ? StackTraceUsage.WithCallSite : StackTraceUsage.WithoutSource;
```
304NotModified: please create a helper for this

304NotModified: FYI,

I'm trying to follow https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming) these days in NLog

src/NLog/Fluent/LogBuilder.cs
snakefoot: Not sure I understand. NET4_5 means all newer versions (including NetStandard etc.)

src/NLog/Fluent/LogBuilder.cs
snakefoot: No it says only call `SetCallerInfo` if one of them is NOT null. Better handle NetFramework version without support for caller-member-attributes (NET3_5 +NET4_0)

src/NLog/Fluent/LogBuilder.cs
snakefoot: Well it is todo work. But I would like to do it in separate commit. Like to keep things apart, since removing these properties requires cleanup other places.

src/NLog/Internal/TargetWithFilterChain.cs
snakefoot: CallSiteKey is a struct, so no need for null-check

src/NLog/Config/StackTraceUsage.cs
snakefoot: StackTraceUsage is used everywhere as a level of details. The higher the number, the higher the detail. Not as a combination.

src/NLog/Fluent/LogBuilder.cs
snakefoot: Done

src/NLog/Internal/TargetWithFilterChain.cs
snakefoot: Done

src/NLog/Internal/TargetWithFilterChain.cs
snakefoot: Added comment. But the classname-extractor can perform substring, and other logic. Want to avoid allocating class-name-strings for every logging-location in a source-file.

src/NLog/Internal/TargetWithFilterChain.cs
snakefoot: Well this is just a safe-guard in case one has called in an expected order. So it will "never" catch anything.

src/NLog/Internal/TargetWithFilterChain.cs
snakefoot: Done

src/NLog/LayoutRenderers/CallSite and stacktraces/CallSiteLayoutRenderer.cs
snakefoot: Done

src/NLog/Config/StackTraceUsage.cs
304NotModified: good point, but what's higher? WithCallSiteSource  or WithoutSource ? ;)

src/NLog/Fluent/LogBuilder.cs
304NotModified: OK, but it wasn't there before. So why the change?

src/NLog/Fluent/LogBuilder.cs
304NotModified: OK 👍

src/NLog/Internal/TargetWithFilterChain.cs
304NotModified: ah missed that, thanks

src/NLog/Fluent/LogBuilder.cs
snakefoot: Fixing  #2059 by just moving `NET4_5` (Has been there all the time)

src/NLog/Config/StackTraceUsage.cs
snakefoot: Also a very good point. Have now changed it into enum-flags.

304NotModified:
> Surprised by the special methods in `Log.cs`. I hope no one are using them as they must be crazy slow with disk-access for file-lookup on every LogEvent.

Do you mean Path.GetFileNameWithoutExtension(callerFilePath)?

snakefoot:
Ah yes sorry. Thought it tried to find the filename on the disk. But it is "just' removing the file-extension. Not so bad besides doing string allocation and Logger-lookup on every logevent


Sent from my Sony Xperia


Surprised by the special methods in Log.cs. I hope no one are using them as they must be crazy slow with disk-access for file-lookup on every LogEvent.

Do you mean Path.GetFileNameWithoutExtension(callerFilePath)?

snakefoot:
Going on a 10 day vacation. So if more changes are needed before merge, then you are welcome to make commit to the branch :)

304NotModified:
Have a good holiday!

src/NLog/Config/StackTraceUsage.cs
```diff
@@ -33,36 +33,56 @@
...
+        /// <summary>
+        /// Stack trace should only be captured if callsite details are missing
+        /// </summary>
+        WithCallSite = 4,
 
         /// <summary>
         /// Stack trace should be captured without source-level information.
         /// </summary>
-        WithoutSource = 1,
+        WithoutSource = WithStackTrace,
```
304NotModified: what's the use of this alias?

src/NLog/Fluent/LogBuilder.cs
```diff
@@ -279,44 +279,34 @@ public LogBuilder StackTrace(StackTrace stackTrace, int userStackFrame)
...
-            [CallerLineNumber]int callerLineNumber = 0)
+#if NET4_5
+            [CallerMemberName]
+#endif
+            string callerMemberName = null,
+#if NET4_5
+            [CallerFilePath]
+#endif
+            string callerFilePath = null,
+#if NET4_5
```
304NotModified: I'm still a bit confused by this :(

In .net 4, why having this method with 3 parameters? What am I missing?

src/NLog/Fluent/LogBuilder.cs
```diff
@@ -279,44 +279,34 @@ public LogBuilder StackTrace(StackTrace stackTrace, int userStackFrame)
...
-            _logEvent.SetCallerInfo(null, callerMemberName, callerFilePath, callerLineNumber);
+            SetCallerInfo(callerMemberName, callerFilePath, callerLineNumber);
 
             _logger.Log(_logEvent);
         }
-#else
-        /// <summary>
-        /// Writes the log event to the underlying logger.
-        /// </summary>
-        public void Write()
```
304NotModified: maybe i'm missing it, but I think this was better

src/NLog/LoggerImpl.cs
```diff
@@ -55,19 +55,35 @@ internal static void Write([NotNull] Type loggerType, [NotNull] TargetWithFilter
         {
 #if !NETSTANDARD1_0 || NETSTANDARD1_5
             StackTraceUsage stu = targetsForLevel.GetStackTraceUsage();
-            if (stu != StackTraceUsage.None && !logEvent.HasStackTrace)
+            if (stu != StackTraceUsage.None)
             {
+                bool attemptCallSiteOptimization = targetsForLevel.TryCallSiteClassNameOptimization(stu, logEvent);
+                if (attemptCallSiteOptimization && targetsForLevel.TryLookupCallSiteClassName(logEvent, out string callSiteClassName))
+                {
+                    logEvent.CallSiteInformation.CallerClassName = callSiteClassName;
```
304NotModified: shouldn't we check ` logEvent.CallSiteInformation` for null?

304NotModified: or should we use `GetCallSiteInformationInternal()`?

src/NLog/Targets/Target.cs
```diff
@@ -521,9 +521,9 @@ private void FindAllLayouts()
             {
                 _allLayoutsAreThreadSafe = _allLayouts.All(layout => layout.ThreadSafe);
             }
-            StackTraceUsage = _allLayouts.DefaultIfEmpty().Max(layout => layout?.StackTraceUsage ?? StackTraceUsage.None);
-            if (this is IUsesStackTrace usesStackTrace && usesStackTrace.StackTraceUsage > StackTraceUsage)
-                StackTraceUsage = usesStackTrace.StackTraceUsage;
+            StackTraceUsage = _allLayouts.DefaultIfEmpty().Aggregate(StackTraceUsage.None, (seed,layout) => seed | (layout?.StackTraceUsage ?? StackTraceUsage.None));
```
304NotModified: I refactored the Aggregate, those are hard to read

snakefoot:
Replying from my phone. The moving of NET4_5 for caller member attributes. Is to make NET4_0 + NET3_5 binary compatible. One could also add 3 default parameters to the now removed Write-method with the exact same parameters name.



I have not removed the old enum names in attenpt to make it less breaking. But you are welcome to remove all the old aliases.



There is no need for null check as it needs the callerinfo-details to make the lookup (details equals null means no lookup and no assignment of class-name)




Sent from my Sony Xperia


Have a good holiday!

snakefoot:
Replying from my phone. The moving of NET4_5 for caller member attributes. Is to make NET4_0 + NET3_5 binary compatible. One could also add 3 default parameters to the now removed Write-method with the exact same parameters name.


I have not removed the old enum names in attenpt to make it less breaking. But you are welcome to remove all the old aliases.


There is no need for null check as it needs the callerinfo-details to make the lookup (details equals null means no lookup and no assignment of class-name)



Sent from my Sony Xperia


In src/NLog/Config/StackTraceUsage.cs<https://github.com/NLog/NLog/pull/3431#discussion_r289130220>:

>
         /// <summary>
         /// Stack trace should be captured without source-level information.
         /// </summary>
-        WithoutSource = 1,
+        WithoutSource = WithStackTrace,


what's the use of this alias?

In src/NLog/Fluent/LogBuilder.cs<https://github.com/NLog/NLog/pull/3431#discussion_r289131193>:

>          /// <summary>
         /// Writes the log event to the underlying logger.
         /// </summary>
         /// <param name="callerMemberName">The method or property name of the caller to the method. This is set at by the compiler.</param>
         /// <param name="callerFilePath">The full path of the source file that contains the caller. This is set at by the compiler.</param>
         /// <param name="callerLineNumber">The line number in the source file at which the method is called. This is set at by the compiler.</param>
         public void Write(
-            [CallerMemberName]string callerMemberName = null,
-            [CallerFilePath]string callerFilePath = null,
-            [CallerLineNumber]int callerLineNumber = 0)
+#if NET4_5
+            [CallerMemberName]
+#endif
+            string callerMemberName = null,
+#if NET4_5
+            [CallerFilePath]
+#endif
+            string callerFilePath = null,
+#if NET4_5


I'm still a bit confused by this :(

In .net 4, why having this method with 3 parameters? What am I missing?

In src/NLog/Fluent/LogBuilder.cs<https://github.com/NLog/NLog/pull/3431#discussion_r289131348>:

>
             _logger.Log(_logEvent);
         }
-#else
-        /// <summary>
-        /// Writes the log event to the underlying logger.
-        /// </summary>
-        public void Write()


maybe i'm missing it, but I think this was better

In src/NLog/LoggerImpl.cs<https://github.com/NLog/NLog/pull/3431#discussion_r289132775>:

>              {
+                bool attemptCallSiteOptimization = targetsForLevel.TryCallSiteClassNameOptimization(stu, logEvent);
+                if (attemptCallSiteOptimization && targetsForLevel.TryLookupCallSiteClassName(logEvent, out string callSiteClassName))
+                {
+                    logEvent.CallSiteInformation.CallerClassName = callSiteClassName;


shouldn't we check logEvent.CallSiteInformation for null?

In src/NLog/LoggerImpl.cs<https://github.com/NLog/NLog/pull/3431#discussion_r289133127>:

>              {
+                bool attemptCallSiteOptimization = targetsForLevel.TryCallSiteClassNameOptimization(stu, logEvent);
+                if (attemptCallSiteOptimization && targetsForLevel.TryLookupCallSiteClassName(logEvent, out string callSiteClassName))
+                {
+                    logEvent.CallSiteInformation.CallerClassName = callSiteClassName;


or should we use GetCallSiteInformationInternal()?

In src/NLog/Targets/Target.cs<https://github.com/NLog/NLog/pull/3431#discussion_r289134461>:

> @@ -521,9 +521,9 @@ private void FindAllLayouts()
             {
                 _allLayoutsAreThreadSafe = _allLayouts.All(layout => layout.ThreadSafe);
             }
-            StackTraceUsage = _allLayouts.DefaultIfEmpty().Max(layout => layout?.StackTraceUsage ?? StackTraceUsage.None);
-            if (this is IUsesStackTrace usesStackTrace && usesStackTrace.StackTraceUsage > StackTraceUsage)
-                StackTraceUsage = usesStackTrace.StackTraceUsage;
+            StackTraceUsage = _allLayouts.DefaultIfEmpty().Aggregate(StackTraceUsage.None, (seed,layout) => seed | (layout?.StackTraceUsage ?? StackTraceUsage.None));


I refactored the Aggregate, those are hard to read

304NotModified:
Thanks for the reply!

snakefoot:
Btw found this page where the wiki is used for  issue tracking


https://github.com/NLog/NLog/wiki/Unable-to-write-to-NLOG-in-kubernetes-containers


Sent from my Sony Xperia


Thanks for the reply!

304NotModified:
converted this to an issue -> https://github.com/NLog/NLog/issues/3446

snakefoot:
Regarding the new issue then absolute file path should be used that matches the mounted volumes configured. Kubernetes write permissions should of course be given.


Else good idea to upgrade to latest NLog and check the internal logging (maybe redirect to console if that is working)


Sent from my Sony Xperia


converted this to an issue -> #3446<https://github.com/NLog/NLog/issues/3446>

304NotModified:
> Regarding the new issue then absolute file path should be used that matches the mounted volumes configured. Kubernetes write permissions should of course be given. Else good idea to upgrade to latest NLog and check the internal logging (maybe redirect to console if that is working)

Is this related to this PR?

snakefoot:
Nope to the issue created from the rogue wiki page


Sent from my Sony Xperia


Regarding the new issue then absolute file path should be used that matches the mounted volumes configured. Kubernetes write permissions should of course be given. Else good idea to upgrade to latest NLog and check the internal logging (maybe redirect to console if that is working)

Is this related to this PR?

src/NLog/Config/StackTraceUsage.cs
304NotModified: backwards comp 👍

src/NLog/Fluent/LogBuilder.cs
304NotModified: If possible, please revert to this

src/NLog/Config/StackTraceUsage.cs
```diff
@@ -33,36 +33,56 @@
...
+        /// <summary>
+        /// Stack trace should be captured
+        /// </summary>
+        WithStackTrace = 1,
+
+#if !SILVERLIGHT
+        /// <summary>
+        /// Source-level information should be capture (source-filename + linenumber)
+        /// </summary>
+        WithSourceCode = 2,
```
304NotModified: The big reason why this isn't merged yet are the changed write method and the names of this enum.

I'm doubting if sourcecode is clear. But for now haven't better names

304NotModified: Maybe rename this to WithFileNameAndLineNumber?

src/NLog/Config/StackTraceUsage.cs
```diff
@@ -33,36 +33,56 @@
...
         /// <summary>
         /// Stack trace should not be captured.
         /// </summary>
-        None = 0, 
+        None = 0,
+
+        /// <summary>
+        /// Stack trace should be captured
+        /// </summary>
+        WithStackTrace = 1,
```
304NotModified: maybe: WithFullStackTrace?

src/NLog/Config/StackTraceUsage.cs
```diff
@@ -33,36 +33,56 @@
...
+        /// <summary>
+        /// Source-level information should be capture (source-filename + linenumber)
+        /// </summary>
+        WithSourceCode = 2,
+#endif
+
+        /// <summary>
+        /// Stack trace should only be captured if callsite details are missing
+        /// </summary>
+        WithCallSite = 4,
```
304NotModified: maybe WithOnlyCallSite?

src/NLog/Config/StackTraceUsage.cs
snakefoot: What is the difference between "full stacktrace" and "stacktrace" ?

src/NLog/Config/StackTraceUsage.cs
snakefoot: Please tell reason

src/NLog/Config/StackTraceUsage.cs
snakefoot: Fine with me

src/NLog/Fluent/LogBuilder.cs
snakefoot: Have partially reverted so your eyes will relax.

src/NLog/Fluent/LogBuilder.cs
snakefoot: Make Net45 and Net40 binary compatible.

src/NLog/LoggerImpl.cs
snakefoot: There is no need for null check as it needs the callerinfo-details to make the lookup (details equals null means no lookup and no assignment of class-name)

src/NLog/Targets/Target.cs
snakefoot: Thought you hated my for-loops :)

src/NLog/Config/StackTraceUsage.cs
304NotModified: I've difficulties to understand the difference between all the options now.  I'm afraid it's too difficult for the users.

src/NLog/Config/StackTraceUsage.cs
304NotModified: it's to make the difference between stacktrace and callsite more clear. But OK, this isn't a good proposal

src/NLog/Config/StackTraceUsage.cs
snakefoot: Only highly skilled people who knows about the NLog Callsite-Engine and will be creating their own special LayoutRenderers/Targets will have to know about these enum-values. Pretty sure no-one knows how to use them right now.

304NotModified: But maybe that should be in the docs. It isn't explaining now what it does (and so the difference between callsite and stackstrace is difficult)

Please note that in VS there is also a callstack window, which is 50% callsite  and 50% stackstrace  ;)

snakefoot: Well in NLog callsite means the calling method (aka caller information). and stacktrace means the callstack. Both of them have the option to include-source-information (filename + linenumber). But I'm pretty sure you know all this already.

304NotModified: When trying to keep things simple, I'm not the only audience ;)

snakefoot: NLog callsite-name comes from the `${callsite}` layoutrenderer. Keeping this name because I guess people using NLog, are used to the NLog-names.

snakefoot: > When trying to keep things simple, I'm not the only audience ;)

Well I guess there 4 people in total, when including me and you :)

src/NLog/Config/StackTraceUsage.cs
304NotModified: maybe we should `[Obsolete]` this one.

src/NLog/Config/StackTraceUsage.cs
snakefoot: I find the enum-names very simple. And have just kept the old enum-names to ensure backward-compatible. But made it very explicit that they are combination of the new enum-flag-values.

304NotModified: Updated docs, this is fine for me know. Thanks!

src/NLog/Fluent/LogBuilder.cs
```diff
@@ -294,36 +294,28 @@ public LogBuilder StackTrace(StackTrace stackTrace, int userStackFrame)
...
         }
 #else
-        /// <summary>
-        /// Writes the log event to the underlying logger.
-        /// </summary>
-        public void Write()
+        public void Write(
+            string callerMemberName = null,
+            string callerFilePath = null,
+            int callerLineNumber = 0)
```
304NotModified: I don't fully understand why the signature changed for `Write` for < net 45? 

(the docs outside is a good idea)

snakefoot: I want to make the Logger-interface binary compatible.

304NotModified: While that's a nice property, it's not really useful for now? We can't (runtime) replace the nlog.dll for .net4 with .net45 already. 

I prefer revert of this as it could confuse users. (the optional parameters are visible)

snakefoot: Not sure I understand the issue with confused users. Net45 users are also looking at default parameters.

If you have a net40 library compiled with NLog Net40 that uses fluent-interface, and then you attempt to use that library in an Net45 application, then you will get method-not-found. This is the issue I'm trying to fix. The NLog Logger-interface should not give such surprises.

304NotModified: > This is the issue I'm trying to fix

ah I would expect more issues, but OK you convinced me.  :)

src/NLog/LogEventInfo.cs
```diff
@@ -204,7 +204,13 @@ public LogEventInfo(LogLevel level, string loggerName, IFormatProvider formatPro
         /// <summary>
         /// Gets the callsite member function name
         /// </summary>
-        public string CallerMemberName => CallSiteInformation?.GetCallerMemberName(null, false, true, true);
+        [Obsolete("Use CallerMethodName instead. Marked obsolete on NLog 5.0")]
+        public string CallerMemberName => CallSiteInformation?.GetCallerMethodName(null, false, true, true);
```
304NotModified: Why the rename? The attribute is called CallerMemberName

Maybe method is better , but I really don't prefer obsoletes

snakefoot: The rename is because I find "CallerMethodName" easier to understand. And there are tools like Fody that can also be used to capture caller-context, where they don't use the special attributes.

304NotModified: Clear. But because of the obsolete I think we should not do this. It's not that important IMO and obsoletes should be prevented.

So please revert/remove the commit

snakefoot: Have reverted the rename of CallerMemberName

304NotModified: Thanks!

src/NLog/LogEventInfo.cs
```diff
@@ -551,12 +551,12 @@ public void SetStackTrace(StackTrace stackTrace, int userStackFrame)
         /// Sets the details retrieved from the Caller Information Attributes
         /// </summary>
         /// <param name="callerClassName"></param>
-        /// <param name="callerMemberName"></param>
+        /// <param name="callerMethodName"></param>
         /// <param name="callerFilePath"></param>
         /// <param name="callerLineNumber"></param>
-        public void SetCallerInfo(string callerClassName, string callerMemberName, string callerFilePath, int callerLineNumber)
+        public void SetCallerInfo(string callerClassName, string callerMethodName, string callerFilePath, int callerLineNumber)
```
304NotModified: I'm afraid this is confusing. Please membername everywhere. Using 2 names for one thing is confusing.

snakefoot: Fixed

304NotModified: Is still see many callerMethodNames?

snakefoot: Thought you agreed that `callerMemberName` was a confusing name for something that is a method-name. See no reason for a badly chosen property-name to obfuscate the rest of the internal code (But leaving the smell in the public interface). Notice that when extracting the method-name from stacktrace, then it is correctly called method-name.

304NotModified:
Sorry for the long delay!

PR #3434, opened 2019-05-25
ObjectReflectionCache - Skip property-reflection for IFormattable

snakefoot:
Avoid serializing JSON.NET `JValue`-properties when it contains a non-primitive object (Ex. like TimeSpan).


304NotModified:
Thanks!

PR #3435, opened 2019-05-26
Async / buffering wrapper: Improve performance when blocking

snakefoot:
ConcurrentRequestQueue - Always see queue as empty after reached RequestLimit (and resolved it)

When reaching the RequestLimit, then producers perform asymmetric update of queue-count.

Instead of trying to coordinate which of the producers should signal start-timer, then all producers will signal start-timer since it is "exceptional" situation.

### DIFF
--- a/src/NLog/Targets/Wrappers/ConcurrentRequestQueue.cs
+++ b/src/NLog/Targets/Wrappers/ConcurrentRequestQueue.cs
@@ -79,6 +79,7 @@ namespace NLog.Targets.Wrappers
         public override bool Enqueue(AsyncLogEventInfo logEventInfo)
         {
             long currentCount = Interlocked.Increment(ref _count);
+            bool queueWasEmpty = currentCount == 1;  // Inserted first item in empty queue
             if (currentCount > RequestLimit)
             {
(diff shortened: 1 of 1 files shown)


304NotModified:
Hi,

Is this a bug fix or an (performance)enhancement?

snakefoot:
It is continuation from #3416 where I tried to fix the "bug" of failing to handle a RequestLimit of 2 with Blocking-mode.

snakefoot:
It is a simplification of the previous commit. So instead of trying to be clever then it just takes a performance hit when blocking mode has been activated.


The simplication reduces the number of modifications of the _count variable, so less memory write barriers are needed when blocking mode is activated. Improves the concurrency of the producers as they are less chatty.


But now all producers will eagerly try to wake the timer-thread after blocking mode has been activated. This might kill any performance gain from improving producer concurrency.


But I like the simplicity of this change, and my simple performance tests still shows good performance


Sent from my Sony Xperia


Hi,

Is this a bug fix or an (performance)enhancement?

304NotModified:
thanks! Merged!

PR #3438, opened 2019-05-28
Removed AllowPartiallyTrustedCallers-attribute for NetCore1

### LINKED ISSUE #3436: App crash on UWP (.Net Standard 1.3) when built in release mode and submitted to Store
**NLog versions tested**: 4.6.3 and 4.5.0

_(Both versions were installed via NuGet)_

**Platform**: UWP with minimum sdk version set to 14393 ( = .Net Standard 1.3 )

**Steps to reproduce**:

I created a blank project with minimum sdk version set to 14393 and target sdk version set to 17763.

Then, I put some code in the constructor of `App` class, to initialize NLog and send a test log:

```c#
        public App()
        {

            var testTarget = new DebugTarget("testtarget")
            {
                Layout = "${logger}|${message}",
            };
            config.AddTarget(testTarget);
            config.AddRule(LogLevel.Debug, LogLevel.Fatal, testTarget);


            LogManager.Configuration = config;

            var logger = LogManager.GetCurrentClassLogger();

            logger.Warn("Test!");

            this.InitializeComponent();
            this.Suspending += OnSuspending;
        }
```

(I've also tested by adding a `FileTarget` instead of `DebugTarget`, and got exactly the same results)

When testing on debug mode, this code works fine. 

Then, I associated the app to an app in the Store and built it for uploading to the Store. The flow creates a appxbundle file for test, as well as an appxupload file for uploading to the Store.

The appxbundle test file also works fine (but Store does not accept it, saying it's a precompiled .Net Native binary and I should upload the appxupload file instead).

Then, I uploaded the appxupload to the Store and got it from the Store, it crashes on startup. I've also extracted the contents of the appxupload (there's an appxbundle there) and installed it manually, it also crashes.

I've attached Visual Studio debugger to it (from Debug->Other Debug Targets->Debug Installed App Package), and enabled exceptions. I saw these three exceptions being thrown:

```
1)

System.MethodAccessException: 'Attempt by security transparent method 'NLog.Internal.ReflectionHelpers.SafeGetTypes(System.Reflection.Assembly)' to access security critical method 'System.Reflection.AssemblyExtensions.GetTypes(System.Reflection.Assembly)' failed.

Assembly 'NLog, Version=4.0.0.0, Culture=neutral, PublicKeyToken=5120e14c03d0593c' is marked with the AllowPartiallyTrustedCallersAttribute, and uses the level 2 security transparency model.  Level 2 transparency causes all methods in AllowPartiallyTrustedCallers assemblies to become security transparent by default, which may be the cause of this exception.'

   at NLog.Internal.ReflectionHelpers.SafeGetTypes(Assembly assembly)





2)

System.ArgumentException: 'LayoutRenderer cannot be found: 'longdate''

   at NLog.Config.Factory`2.CreateInstance(String itemName)



3)

System.MethodAccessException: 'Attempt by security transparent method 'NLog.Internal.PropertyHelper.GetAllReadableProperties(System.Type)' to access security critical method 'System.Reflection.TypeExtensions.GetProperties(System.Type, System.Reflection.BindingFlags)' failed.

Assembly 'NLog, Version=4.0.0.0, Culture=neutral, PublicKeyToken=5120e14c03d0593c' is marked with the AllowPartiallyTrustedCallersAttribute, and uses the level 2 security transparency model.  Level 2 transparency causes all methods in AllowPartiallyTrustedCallers assemblies to become security transparent by default, which may be the cause of this exception.'

   at NLog.Internal.PropertyHelper.GetAllReadableProperties(Type type)
```

It seems that setting minimum sdk version to 16299 (.Net Standard 2.0) fixes the issue, but that's not always an option for apps.

(Please note that extracting appxupload and installing it manually does not work anymore when min sdk version is set to 16299, I assume something's changed in that version and Store does some processing on the package before making it available; but when submitted through the Store and downloaded from the Store, it works fine)

snakefoot:
Not supported by System.Reflection.TypeExtensions for NetStandard. Trying to resolve  #3436


304NotModified:
Not needed anymore :)

Thanks for the PR @snakefoot !

PR #3449, opened 2019-06-04
Decouple LoggingConfiguration from XmlLoggingConfiguration

304NotModified:


snakefoot:
Instead of taking a hack (to avoid making breaking change in xml-config) and making it a public interface.

Then I think this hack should be removed, and instead allow xml-config to throw exception not able to load a valid configuration.

Issue #3362, opened 2019-05-03
Better DI support


304NotModified:
Support DI providers. Difficult as logging is configured before the DI in our aspnet core example

304NotModified:
@snakefoot I think it's important to fix this, but I think we have to think of a design before implementing. I'm doubting to close https://github.com/NLog/NLog/pull/3128 for now, or do you think that's rude?

304NotModified:
What needs to be supported in this case:

1. Inject non NLog interfaces and classes into layout renderers and targets, e.g.
    

     ```c#

     class MyTarget{
             MyTarget(IMyTargetService service) { ...}

     }
2. Resolve NLog targets, layoutrs and layout renderers?
3. Resolve a custom target etc by DI (no need for "register"?)
3. 
3. Work correctly with NLog.Extensions.Logging, fix the issue that in the in main the injectables aren't ready (as configureService is later on)

- Currently when reading the XML config, the targets are "newed". Maybe that's too soon?

snakefoot:
I would settle for having a ServiceProvider that one can register/lookup services.

Trying to do funky stuff like normal DependencyInjection frameworks are doing like scanning all types in all assemblies, and trying and perform fluffy-matching of constructur-parameters are out scope. Along with assigning special lifetime-traits to objects (singletons/disposable/etc.) is also out-of-scope.

snakefoot:
> do you think that's rude?

You can do what you want. I just wanted a generic ServiceProvider than all these static properties. But I also wanted it to be very simple. People can then hook it up to more advanced DI-containers if neeeded.

304NotModified:
> People can then hook it up to more advanced DI-containers if needed.

>. People can then hook it up to more advanced DI-containers if needed.

Yes, that's the goal. 

FYI tried that with the current version (and some new code as extensions) for asp.net core 2 with MS DI, but failed. Got it working to create a new instance, but Write(LogEventInfo logEvent) was never called :(  (need to investigate that)

304NotModified:
Proposal:

```c#
interface IResolver{

    object Resolve(Type t);
}
```

Looks like: https://docs.microsoft.com/en-us/dotnet/api/system.iserviceprovider.getservice?view=netframework-4.8

304NotModified:
basic impl idea:

```c#
internal class DefaultResolver : IResolver, IRegisterDependecies{

Register<TInterface, TImplement>();
Register(Type from, Type to);

Resolve(Type t) {
[...]

PR #3450, opened 2019-06-04
Partly decouple LogFactory from FileLoaders

304NotModified:
this step 1

step 2:
-  Introduce Interface for factory
- Add to DI : needed: #3362

### DIFF
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -106,11 +106,7 @@ namespace NLog
         /// Initializes a new instance of the <see cref="LogFactory" /> class.
         /// </summary>
         public LogFactory()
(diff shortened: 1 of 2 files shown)


304NotModified:
Todo interface?

304NotModified:
interface done

snakefoot:
Instead of having the file-monitoring in the file-loader, then consider moving it somewhere else. The file-loader-logic was mostly a first step to remove all file-logic from LogFactory.

Maybe the file-monitoring-logic should be an interface that the xml-config can lookup (and activate), when it sees that autoreload=true.

Just like the NLogConfigurationConfig is able to able to perform autoreload without help from NLog.

Issue #1577, opened 2016-08-02
Feature Request: Support inner text in <variable> tags


anton-gogolev:
Currently, with complex expressions, `<variable>` syntax gets pretty unwieldy :

```xml
<!-- Adding linebreaks for readability -->
<variable name="mailBodyLayout" 
    value="${onexception:inner=${exceptionMailBodyLayout}:whenempty=${standardMailBodyLayout}}
    ${newline}${newline}---${newline}${application-name} v${application-version}
    ${newline}${newline}Server: ${machinename}${newline}Date: ${longdate}${newline}
    ${when:when=length('${iis-site-name}') > 0:inner=${iisApplicationInformation}}
    Base Directory: ${basedir}${newline}Identity: ${windows-identity}${newline}
    Process Time: ${processtime}" />
```

It would be nice if NLog could use inner text (and interpret `CR`/`LF` as `${newline}`) in `<variable>` tags (maybe wrapped in `<![CDATA[`/`]]>`) instead of `value` attribute. This would make things much easier on the eye:

```xml
<variable name="mailBodyLayout">
<![CDATA[${onexception:inner=${exceptionMailBodyLayout}:whenempty=${standardMailBodyLayout}}

---
${application-name} v${application-version}

Server: ${machinename}
Date: ${longdate}
${when:when=length('${iis-site-name}') > 0:inner=${iisApplicationInformation}}
Base Directory: ${basedir}
Identity: ${windows-identity}
Process Time: ${processtime}]]></variable>
```

304NotModified:
Thanks, I like the idea! 

I you could help in writing a unit test for it (check the current unit tests), that would boost the implementation!

anton-gogolev:
Absolutely. Is `NLog.UnitTests.Config.VariableTests` the correct place to start looking?

304NotModified:
Yes indeed

304NotModified:
Need any help on this?

anton-gogolev:
Welp, my bad for not following through. Please, do not close this just yet :)

snakefoot:
You can already do this with NLog:

```xml
<variable name="mailBodyLayout">
<value>
<![CDATA[${onexception:inner=${exceptionMailBodyLayout}:whenempty=${standardMailBodyLayout}}

---
${application-name} v${application-version}

Server: ${machinename}
Date: ${longdate}
${when:when=length('${iis-site-name}') > 0:inner=${iisApplicationInformation}}
[...]

Issue #1312, opened 2016-03-18
Allow: JsonLayout into a variable


mccabela:
NLog version 4.2.3 - .Net 4.5

Previously I used to store my layout in a variable and place that layout on the target.  This allowed me to setup the defaultLayout once and share it across multiple targets.

e.g.

``` xml
<!-- Setting the Variable -->
<variable name="defaultLayout" value="${iso8601dateformat}|${level:uppercase=true}|[${logger}]|[Thread-${threadid}]|[Machine-${machinename}]|${message}${onexception:${newline}EXCEPTION OCCURRED\:${exception:tostring:maxInnerExceptionLevel=5:innerFormat=tostring}}" />
```

``` xml
<!-- Assigning the layout -->
      <target xsi:type="File" fileName="${logFilePath}trace.txt" layout="${defaultLayout}"
              maxArchiveFiles="100"
              archiveNumbering="Sequence"
              archiveAboveSize="${archiveFileAboveBytes}"
              archiveFileName="${logFilePath}archive/trace-{#####}.txt" />
```

I'm now moving my logging to use JsonLayout.  Is it still possible to store the layout in a variable and assign it to a target?

If so can you please provide an example of how it can be done?

Thanks,

Lach

304NotModified:
I'm not sure if this would work, but maybe you can test (untested here)


```xml
<variable name="defaultLayout" />
  <layout type="JsonLayout">
    <attribute name="message" value="${messsage}">
  </layout>
</variable>
```

mccabela:
I tried and threw an exception "Expected value on <variable />" so I don't think it works :(

So it might need to be a feature request.

Thanks for the suggestion

304NotModified:
Checked the code, currently only SimpleLayouts are allowed. There was a good reason for this, not sure but I guess performance.

![image](https://cloud.githubusercontent.com/assets/5808377/14082974/5b356ac4-f513-11e5-8837-b18f537a00b2.png)


We should check if we could change this from SimpleLayout -> Layout.

PR #3459, opened 2019-06-07
Support layouts (e.g. JSON) and inner text in <variable>

304NotModified:
fixes https://github.com/NLog/NLog/issues/1577 and fixes https://github.com/NLog/NLog/issues/1312

### DIFF
--- a/src/NLog/Config/ILoggingConfigurationElement.cs
+++ b/src/NLog/Config/ILoggingConfigurationElement.cs
@@ -52,5 +52,10 @@ namespace NLog.Config
         /// Child config sections
         /// </summary>
         IEnumerable<ILoggingConfigurationElement> Children { get; }
+
+        /// <summary>
+        /// Gets the value of the element.
+        /// </summary>
+        string Value { get; }
     }
 }
--- a/src/NLog/Config/LoggingConfiguration.cs
+++ b/src/NLog/Config/LoggingConfiguration.cs
@@ -61,7 +61,7 @@ namespace NLog.Config
         /// <summary>
         /// Variables defined in xml or in API. name is case case insensitive. 
         /// </summary>
-        private readonly ThreadSafeDictionary<string, SimpleLayout> _variables = new ThreadSafeDictionary<string, SimpleLayout>(StringComparer.OrdinalIgnoreCase);
+        private readonly ThreadSafeDictionary<string, Layout> _variables = new ThreadSafeDictionary<string, Layout>(StringComparer.OrdinalIgnoreCase);
 
         /// <summary>
         /// Gets the factory that will be configured
@@ -88,7 +88,7 @@ namespace NLog.Config
         /// <summary>
         /// Gets the variables defined in the configuration.
         /// </summary>
-        public IDictionary<string, SimpleLayout> Variables => _variables;
+        public IDictionary<string, Layout> Variables => _variables;
 
         /// <summary>
         /// Gets a collection of named targets specified in the configuration.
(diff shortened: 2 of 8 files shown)


304NotModified:
@snakefoot I have the idea you don't like this change 👼

snakefoot:
Think it is good idea. Just need to do it right so the same variable is not used multiple places when not threadsafe.


Sent from my Sony Xperia


@snakefoot<https://github.com/snakefoot> I have the idea you don't like this change 👼

snakefoot:
The problem with not being threadsafe. Is NOT that multiple threads can get in trouble by "updating" the layout-value (no longer a issue) . But the problem is when multiple targets uses the same thread-unsafe layout, then you get  unexpected output when two targets shares the same layout object.


So your warning comment in variable-layout-renderer is misleading.


Sent from my Sony Xperia


Merged #3459<https://github.com/NLog/NLog/pull/3459> into dev.

304NotModified:
thanks for the feedback -> https://github.com/NLog/NLog/pull/3462

snakefoot:
After seeing you have implemented a breaking change in the config-element-interface that is not compatible with appsettings json then I must agree with you that I don't like this PR.


You have enforced XML-behavior by adding value-property. Please remove this change.


I like the idea of Json-Layout in variable. But not like this. Please fix


Sent from my Sony Xperia


@snakefoot<https://github.com/snakefoot> I have the idea you don't like this change 👼

304NotModified:
The value property could just return null in the case of JSON

snakefoot:
Then you suddenly have something that only works with XML-config


Sent from my Sony Xperia


The value property could just return null in the case of JSON

304NotModified:
'Value' isn't xml specific?

Anyway, do you have a proposal how to solve this different? One way or another, this needs a breaking change.

snakefoot:
Yes it is a breaking change for the Variables-container. Anyway created #3466 and none of your tests are breaking.

snakefoot:
Btw. could be nice if one could make a static reference to a JsonLayout. So one is not required to use `"${var:myjson}"` but could just write:

```
    <variable name='myJson'>
        <layout type='JsonLayout'>
            <attribute name='short date' layout='${shortdate}' />
        </layout>
    </variable>
    <targets>
         <target type='file' name='myfile' layout='${myJson}' />
    <targets>
```

Then one could skip a dictionary lookup for every logevent-write.

oskaremil:
Can I use 
```
"variables": {
      "varDefaultLayout": {
          "type": "JsonLayout",
          "Attributes": [
            {
              "name": "timestamp",
              "layout": "${date:format=o}"
            }
            ]
        }
    },
```

and then 
```
"targets": {
      "logfile": {
        "layout": "${varDefaultLayout}"
```

in my target when reading configuration from appsettings.json ?

snakefoot:
Maybe add example to wiki-page, when NLog 5 is released:

**NLog JsonLayout as NLog Config Variable Example**

```xml
<nlog>
  <variable name="defaultJsonLayout" >
    <layout type="JsonLayout">
       <attribute name="message" value="${messsage}">
    </layout>
  </variable>
  <targets>
    <target name="debug" type="Debug" layout="${var:defaultJsonLayout}" />
  </targets>
  <rules>
    <logger name="*" minlevel="Debug" writeTo="debug" />
  </rules>
</nlog>
```

### API

```c#
var newJsonLayout = new NLog.Layouts.JsonLayout() { IncludeAllProperties = true };
LogManager.Configuration.Variables["defaultJsonLayout"] = newJsonLayout;
```
```

snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/Configuration-file#variables

PR #3466, opened 2019-06-09
LoggingConfigurationParser - Moved attribute validation from Xml-Config

snakefoot:
Improves the validation for `appsettings-json` 

Removing `Value` from `ILoggingConfigurationElement`. Unless full support for checking the Value in all locations where possible (And not just for variables). And also avoid too much diversion from XML and JSON-config (Suddenly supporting unnamed-properties).

Instead one can do this, where one assigns the `value`-property (and allow `<layout>` as alias):
```xml
<variable name="mailBodyLayout">
<layout>
<![CDATA[${onexception:inner=${exceptionMailBodyLayout}:whenempty=${standardMailBodyLayout}}

---
${application-name} v${application-version}

Server: ${machinename}
Date: ${longdate}
${when:when=length('${iis-site-name}') > 0:inner=${iisApplicationInformation}}
Base Directory: ${basedir}
Identity: ${windows-identity}
Process Time: ${processtime}]]>
</layout>
</variable>
```

### DIFF
--- a/src/NLog/Config/LoggingConfigurationParser.cs
+++ b/src/NLog/Config/LoggingConfigurationParser.cs
@@ -76,19 +76,21 @@ namespace NLog.Config
 
             SetNLogElementSettings(nlogConfig);
 
-            var children = nlogConfig.Children.ToList();
+            var validatedConfig = ValidatedConfigurationElement.Create(nlogConfig, LogFactory); // Validate after having loaded initial settings
 
             //first load the extensions, as the can be used in other elements (targets etc)
-            var extensionsChilds = children.Where(child => child.MatchesName("extensions")).ToList();
-            foreach (var extensionsChild in extensionsChilds)
+            foreach (var extensionsChild in validatedConfig.ValidChildren)
             {
-                ParseExtensionsElement(extensionsChild, basePath);
+                if (extensionsChild.MatchesName("extensions"))
+                {
+                    ParseExtensionsElement(extensionsChild, basePath);
+                }
             }
 
-            var rulesList = new List<ILoggingConfigurationElement>();
+            var rulesList = new List<ValidatedConfigurationElement>();
 
             //parse all other direct elements
-            foreach (var child in children)
+            foreach (var child in validatedConfig.ValidChildren)
             {
                 if (child.MatchesName("rules"))
                 {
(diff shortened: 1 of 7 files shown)


src/NLog/Config/LoggingConfigurationParser.cs
```diff
@@ -1295,9 +1278,122 @@ private static string GetName(Target target)
...
+            IEnumerable<ValidatedConfigurationElement> YieldValidChildren()
+            {
+                foreach (var child in _element.Children)
+                {
+                    _validChildren = _validChildren ?? new List<ValidatedConfigurationElement>();
+                    var validChild = new ValidatedConfigurationElement(child, _throwConfigExceptions);
+                    _validChildren.Add(validChild);
+                    yield return validChild;
+                }
+                _validChildren = _validChildren ?? ArrayHelper.Empty<ValidatedConfigurationElement>();
```
304NotModified: This is a bit unexpected (according to the method name)

could we move this out of this method, or rename the method (YieldAndCacheValidChildren or something like that)

src/NLog/Config/LoggingConfigurationParser.cs
```diff
@@ -1295,9 +1278,122 @@ private static string GetName(Target target)
             return string.IsNullOrEmpty(target.Name) ? target.GetType().Name : target.Name;
         }
 
+        private class ValidatedConfigurationElement : ILoggingConfigurationElement
```
304NotModified: I which level this element is "Validated"? I don't fully understand the class name

304NotModified:
PS:

> And also avoid too much diversion from XML and JSON-config (Suddenly supporting unnamed-properties).

The unnamed property wasn't suddenly for me, and please note it's not unique for XML (yes, it isn't in JSON).

snakefoot:
> The unnamed property wasn't suddenly for me, and please note it's not unique for XML (yes, it isn't in JSON).

Lets say config-item in Nlog has 2 attributes (or more). How do you specify which attribute is configured by the unnamed-property-value?

src/NLog/Config/LoggingConfigurationParser.cs
snakefoot: It currently performs validation of attribute-names. But more validation can be added later.

src/NLog/Config/LoggingConfigurationParser.cs
snakefoot: Will rename the method. But the reason is to perform lazy validation of children.

304NotModified:
@snakefoot now it's my time for holiday :)
I'm a short week in Rome, so I'm limited with response and this PR will be probably after that week.

snakefoot:
Great idea. Rome is very nice. No hurry with the PRs. Enjoy the holiday :)


Sent from my Sony Xperia


@snakefoot<https://github.com/snakefoot> now it's my time for holiday :)
I'm a short week in Rome, so I'm limited with response and this PR will be probably after that week.

304NotModified:
-> https://github.com/snakefoot/NLog/pull/9

snakefoot:
> -> snakefoot#9

Seems you removal of my explicit cast has now broken the build:

https://ci.appveyor.com/project/nlog/nlog/builds/25516652

304NotModified:
> Seems your removal of my explicit cast has now broken the build:
> 
> https://ci.appveyor.com/project/nlog/nlog/builds/25516652

OK, I will check that. Locally it works without cast.

304NotModified:
Unfortunately the logs are always a bit unclear. I forgot .NET 3.5 doesn't support covariance interfaces :(

304NotModified:
did I broke this?

image

snakefoot:
> did I broke this?

No the validation introduced with this PR just makes sure you don't have invalid configuration. You have created the test after this PR was introduced.

snakefoot:
Have now fixed `Xml_configuration_variableWithInnerAndAttribute_attributeHasPrecedence` so it ignores that it uses an invalid config.

304NotModified:
Thanks!

PR #3469, opened 2019-06-10
LogLevel - Added support for TypeConverter

snakefoot:
Also fixes bug where LogLevel always converts to ordinal. Even when asked for string.

See also: https://www.hanselman.com/blog/TypeConvertersTheresNotEnoughTypeDescripterGetConverterInTheWorld.aspx

Sadly enough it doesn't fully solve this issue:

https://stackoverflow.com/questions/56454590/im-having-problems-deserializing-enumerations

Because Json.Net doesn't make use of TypeConverter when class implements `IConvertible`. Guess there is a price of implementing `IConvertible`-interface for custom objects. One could consider creating a PullRequest for Json.Net to improve support for TypeConverter for classes that implements `IConvertible`. (Probably not a an easy task for an outsider, since by design: https://github.com/JamesNK/Newtonsoft.Json/issues/676 )

https://github.com/JamesNK/Newtonsoft.Json/blob/4ab34b0461fb595805d092a46a58f35f66c84d6a/Src/Newtonsoft.Json/Serialization/JsonSerializerInternalReader.cs#L981 (See comment saying by design)

Alternative one could remove IConvertible-interface from LogLevel-class and only use TypeConverter-attribute.

Alternative one could change IConvertible-interface to return TypeCode.String for LogLevel. But it will probably have even more side-effects.

Alternative one could change LogLevel into an enum, instead of having it as class.

### DIFF
--- a/src/NLog/Attributes/LogLevelTypeConverter.cs
+++ b/src/NLog/Attributes/LogLevelTypeConverter.cs
@@ -1,0 +1,91 @@
+﻿// 
+// Copyright (c) 2004-2019 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without 
+// modification, are permitted provided that the following conditions 
+// are met:
+// 
+// * Redistributions of source code must retain the above copyright notice, 
+//   this list of conditions and the following disclaimer. 
+// 
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution. 
+// 
+// * Neither the name of Jaroslaw Kowalski nor the names of its 
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission. 
+// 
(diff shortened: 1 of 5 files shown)


304NotModified:
Thanks for the PR!

> Alternative one could remove IConvertible-interface from LogLevel-class and only use TypeConverter-attribute.

That's pity, as the DB target  / Type Converter is using that interface.


Could you please add some more tests? 

PS: I like the extension, saves some boilerplate stuff


image

https://marketplace.visualstudio.com/items?itemName=RandomEngy.UnitTestBoilerplateGenerator

snakefoot:
> That's pity, as the DB target / Type Converter is using that interface.

Should not be that difficult to also support the TypeConverter-attribute. Then IConvertible is not needed for LogLevel.

snakefoot:
> @304NotModified Could you please add some more tests?

Have improved the test-coverage.

304NotModified:
Thanks! 

Coverage is still not ideal, but good enough for now.

snakefoot:
@304NotModified You have added this to NLog 5.0-milestone. But it is included in the coming NLog 4.6.5

304NotModified:
Yes, I noticed after merge 👼

snakefoot:
> Yes, I noticed after merge

Well it is not a breaking change. Actually think it fixes a bug, when converting LogLevel to string.

PR #3474, opened 2019-06-11
FilteringTargetWrapper - Fix XSD for Filter-property

snakefoot:


304NotModified:
That's an easy solution :)

But do you think we should have the possibility to set the filter attribute (and also condition?) on the filter wrapper?

304NotModified:
Another option is to hide it

snakefoot:
I'm pretty blank when it comes to crafting XSD, so you are welxome to take over :)


Sent from my Sony Xperia


Another option is to hide it

304NotModified:
Ok, but I mean from the XML (NLog.config, not xsd)  for 

> But do you think we should have the possibility to set the filter attribute (and also condition?) on the filter wrapper?
> 
> Another option is to hide it

So functional, not technical. :)

304NotModified:
Eg

We have

    target xsi:type="FilteringWrapper" name="String" condition="Condition">

Does this makes sense? 

  target xsi:type="FilteringWrapper" name="String" filter="string">

And then what about

    target xsi:type="FilteringWrapper" name="String" condition="Condition" filter=string>

snakefoot:
Sounds like you are on top of things. If you can hide Filter as attribute, then please do so.

304NotModified:
thanks, I like the idea as failsafe, so added it here: https://github.com/NLog/NLog/pull/3475

snakefoot:
Please don't use NLogConfigurationIgnoreProperty


Sent from my Sony Xperia


thanks, I like the idea as failsafe, so added it here: #3475<https://github.com/NLog/NLog/pull/3475>

304NotModified:
Why not? I tested it, look ok to me

snakefoot:
> Why not? I tested it, look ok to me

I'm very confused. You said already that you know that `NLogConfigurationIgnoreProperty` will affect how configuration is applied. Then you add this attribute and say it works just fine.

Well updated #3476 with a unit test that will fail, when people adds unwanted attributes to random properties.

304NotModified:
Yes, and afterwards I tested it, included the code paths that are changed...

snakefoot:
But the updated unit-test in  #3476 fails when you add that attribute to the Filter-property. Please try.

304NotModified:
You're right. I didn't understand that my change was breaking things. 

Then my PR is a no-go :)

PR #3478, opened 2019-06-12
XSD: Support <value> in <variable>

304NotModified:
as alternative to variable attribute

resolves https://github.com/NLog/NLog/issues/1577


304NotModified:
unfortunalty I cannot add a validation rule, attribute value is needed when body is null. We need XSD 1.1 for that, which isn't supported in VS 2017 (and think also not VS 2019 - seen in my tests)

PR #3492, opened 2019-06-19
FileTarget - improvements when ConcurrentWrites=false + Extra test MessageTemplate.Render

snakefoot:
- Skips creation of archive mutex when ConcurrentWrites=false
- Skips call to PlatformDetector.SupportsSharableMutex when ConcurrentWrites=false

This is for NLog 4.6.6 No breaking changes. Just better compatibility when explicit specifying `ConcurrentWrites=false`


304NotModified:
Nice improvements!

PR #3494, opened 2019-06-21
GroupByTargetWrapper - Allow one to group by request-identifier

snakefoot:
Allow one to to take an entire buffer of logevents and group-by request-identifier. Sending individual emails / reports.

### DIFF
--- a/tests/NLog.UnitTests/Targets/Wrappers/GroupByTargetWrapperTests.cs
+++ b/tests/NLog.UnitTests/Targets/Wrappers/GroupByTargetWrapperTests.cs
@@ -42,7 +42,7 @@ namespace NLog.UnitTests.Targets.Wrappers
         [Fact]
         public void SimpleGroupByTest()
         {
-            // Stage
+            // Arrange
             var memoryTarget = new MemoryTarget("memory") { Layout = "${level}" };
             var groupByTarget = new GroupByTargetWrapper("groupby", memoryTarget, "${logger}");
             var bufferTarget = new BufferingTargetWrapper("buffer", groupByTarget);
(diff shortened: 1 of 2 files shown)


tests/NLog.UnitTests/Targets/Wrappers/GroupByTargetWrapperTests.cs
```diff
@@ -0,0 +1,76 @@
...
+    using NLog.Targets;
+    using NLog.Targets.Wrappers;
+    using Xunit;
+
+    public class GroupByTargetWrapperTests : NLogTestBase
+	{
+        [Fact]
+        public void SimpleGroupByTest()
+        {
+            // Stage
```
304NotModified: Arrange? ;)

snakefoot:
Still pondering on this one. Whether the GroupBy-logic should "just" be added to https://github.com/NLog/NLog/wiki/PostFilteringWrapper-target

304NotModified:
Cool! I like this feature!

> Still pondering on this one. Whether the GroupBy-logic should "just" be added to https://github.com/NLog/NLog/wiki/PostFilteringWrapper-target

I think separate would be better findable. What would be the benefit of combining with PostFilteringWrapper?

snakefoot:
I was thinking the following setup:

- AutoFlushWrapper (Delayed flush on error)
  - BufferingWrapper (Overflow = Discard)
    - PostFilteringWrapper (Only write groups that contains one or more errors)
       - MailTarget (Receives batch of messages for each group)

This would allow one to get (almost) all logevents (with same correlationid) for a requests that resulted in an error-message.

snakefoot:
> What would be the benefit of combining with PostFilteringWrapper?

I guess it is only the allocation issue that really matters. And I guess for most people this is a non-issue, since it only happens on error-reporting.

304NotModified:
> I guess it is only the reduced allocation that is a benefit But I guess for most people this is a non-issue, since it only happens on error-reporting.

agree on that. I think it's more clear to have this one separate. I'm also OK to merge it for 4.x

snakefoot:
Created wiki-page: https://github.com/NLog/NLog/wiki/GroupByWrapper-target

PR #3497, opened 2019-06-25
Print current building TargetFramework to log

304NotModified:


304NotModified:
Maybe bit confusing with parallel buids, but better than before

PR #3498, opened 2019-06-25
Deploy snupkg on AppVeyor without custom script

304NotModified:


304NotModified:
because broken again

image

PR #3505, opened 2019-07-08
Supporting contruction of  XmlLoggingConfiguration from XmlReader

dmitrynovik:
### Motivation

Every c-tor of `XmlLoggingConfiguration` accepts path. In other words, the `XmlLoggingConfiguration` only supports initialization from files. However, the XML configuration may reside in any place other than a file e.g. memory, network stream, a document data store etc.

### Solution
Introduce `public XmlLoggingConfiguration(XmlReader reader` constructor.

### DIFF
--- a/src/NLog/Config/XmlLoggingConfiguration.cs
+++ b/src/NLog/Config/XmlLoggingConfiguration.cs
@@ -115,6 +115,18 @@ namespace NLog.Config
         }
 
         /// <summary>
+        /// Initializes a new instance of the <see cref="XmlLoggingConfiguration" /> class.
+        /// </summary>
+        /// <param name="reader">XML reader to read from.</param>
+        /// <param name="ignoreErrors">Ignore any errors during configuration.</param>
+        /// <param name="logFactory">The <see cref="LogFactory" /> to which to apply any applicable configuration values.</param>
+        public XmlLoggingConfiguration(XmlReader reader, bool ignoreErrors, LogFactory logFactory)
+            : base(logFactory)
+        {
+            Initialize(reader, null, ignoreErrors);
+        }
+
+        /// <summary>
         /// Create XML reader for (xml config) file.
         /// </summary>
         /// <param name="fileName">filepath</param>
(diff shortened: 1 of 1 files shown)


304NotModified:
Hi,

there is already the ctor, 

```
    public XmlLoggingConfiguration(XmlReader reader, string fileName, bool ignoreErrors, LogFactory logFactory)
```

you could use empty string for the filename

dmitrynovik:
> Hi,
> 
> there is already the ctor,
> 
> ```
>     public XmlLoggingConfiguration(XmlReader reader, string fileName, bool ignoreErrors, LogFactory logFactory)
> ```
> 
> you could use empty string for the filename

The problem confusing the end user is that  `public XmlLoggingConfiguration(XmlReader reader, string fileName, bool ignoreErrors, LogFactory logFactory)` that it has `string fileName` in it, and one has to dig into `NLog` sources to find out that this argument is optional. I was actually the such user :)

304NotModified:
Yes, IMO I think improving the docs is better than adding (another) overload. What do you think?

dmitrynovik:
I think adding a constructor overload with single argument
Xmlloggingconfiguration (XmlReader) would remove all questions. It is one
line of code anyway. Docs are good, but self-explanatory API is better

304NotModified:
> Docs are good, but self-explanatory API is better

True, but which ctor should be invoked when you need to pass ignoreErrors and a custom LogFactory?

A consistent API is also important ;)

304NotModified:
But OK, I will accept this. :)

Could you rebase on master (and change the PR), or should I do that? Then we can release it with the 4.x

snakefoot:
>True, but which ctor should be invoked when you need to pass ignoreErrors and a custom LogFactory?

Curious if you could explain when you want to have `ignoreErrors=true` ? Why create an invalid NLog-config, that is broken halfway because of an exception?

304NotModified:
To ensure logging won't break your application.

snakefoot:
> @304NotModified To ensure logging won't break your application.

But if trying to load a filename, that doesn't exists then it will still throw exceptions (and break your application)

Would think a method that suppress the exceptions and just returns null as config would be better (and would also work when file doesn't exist).

304NotModified:
> In what situations does users create `XmlLoggingConfiguration` (instead of letting NLog load it automatically), but cannot figure out how to catch an exception and instead wants a broken config?

Yes, it's sloppy maybe, but also consistent. Anyway, the challenge was more for the LogFactory. I will rebase this on master and will add also an overload with a LogFactory so it's more consistent

304NotModified:
superseded by #3514

PR #3513, opened 2019-07-12
AutoFlushTargetWrapper - Added FlushOnConditionOnly property

### LINKED ISSUE #3512: Can AutoFlushWrapper be configured to not flush on config reload/shutdown?
**NLog version**: 4.6.5

**Platform**: .NET 4.7.2

**Current NLog config**

```xml
      autoReload="true"
      throwExceptions="false"
      internalLogLevel="Off" internalLogFile="c:\temp\nlog-internal.log">

  <targets>
     <target xsi:type="File" name="fileTarget" fileName="log_${cached:${date:format=yyyy-MM-dd HH_mm_ss}}.log" layout="${longdate}|${level:uppercase=true}|${logger}|${message} ${exception:format=ToString,StackTrace}" />		
		
     <target name="crashTraceTarget" xsi:type="AutoFlushWrapper" condition="level >= LogLevel.Fatal">
        <target xsi:type="BufferingWrapper" bufferSize="500" overflowAction="Discard">
           <target xsi:type="File" name="crashTraceFileTarget" fileName="crashtrace_${cached:${date:format=yyyy-MM-dd HH_mm_ss}}.log" layout="${longdate}|${level:uppercase=true}|${logger}|${message} ${exception:format=ToString,StackTrace}" />
        </target>
     </target>
  </targets>

  <rules>
    <logger name="*" minlevel="Info" writeTo="fileTarget" />
    <logger name="*" minlevel="Trace" writeTo="crashTraceTarget" />
  </rules>
</nlog>
```

I'm trying to use `AutoFlushWrapper` to automatically write a log file containing the last 500 `>= Trace` logs when a `Fatal` log is written. This works, however the log file is also flushed on config reload and application exit which is undesirable. Is there a way to set this up so that the `AutoFlushWrapper` does not flush on config reload or application exit, i.e. the buffered logs are discarded on clean exit?

I've had a look at the documentation for custom extensions/wrappers but can't see anything that would allow suppressing the flush on config reload/application exit.

snakefoot:
Skips flush on reload-config and application-shutdown. Resolve #3512 

To be used in combination with BufferingWrapper and `overflowAction=Discard`

### DIFF
--- a/src/NLog/Targets/Wrappers/AutoFlushTargetWrapper.cs
+++ b/src/NLog/Targets/Wrappers/AutoFlushTargetWrapper.cs
@@ -68,10 +68,17 @@ namespace NLog.Targets.Wrappers
         /// Delay the flush until the LogEvent has been confirmed as written
         /// </summary>
         /// <docgen category='General Options' order='10' />
-        public bool AsyncFlush { get => _asyncFlush ?? true;
+        public bool AsyncFlush
+        {
+            get => _asyncFlush ?? true;
             set => _asyncFlush = value;
         }
         private bool? _asyncFlush;
+
+        /// <summary>
+        /// Only flush when LogEvent matches condition. Ignore explicit-flush, config-reload-flush and shutdown-flush
+        /// </summary>
+        public bool FlushOnConditionOnly { get; set; }
 
         private readonly AsyncOperationCounter _pendingManualFlushList = new AsyncOperationCounter();
 
@@ -137,7 +144,7 @@ namespace NLog.Targets.Wrappers
                     AsyncContinuation wrappedContinuation = (ex) =>
                     {
                         if (ex == null)
-                            WrappedTarget.Flush((e) => { });
+                            FlushOnCondition();
                         _pendingManualFlushList.CompleteOperation(ex);
                         currentContinuation(ex);
                     };
@@ -147,7 +154,7 @@ namespace NLog.Targets.Wrappers
                 else
                 {
                     WrappedTarget.WriteAsyncLogEvent(logEvent);
-                    FlushAsync((e) => { });
+                    FlushOnCondition();
                 }
             }
             else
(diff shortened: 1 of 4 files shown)


snakefoot:
Notice PR is configured to be merged directly into master!

src/NLog/Targets/Wrappers/AutoFlushTargetWrapper.cs
```diff
@@ -68,11 +68,18 @@ public class AutoFlushTargetWrapper : WrapperTargetBase
...
+        {
+            get => _asyncFlush ?? true;
             set => _asyncFlush = value;
         }
         private bool? _asyncFlush;
 
+        /// <summary>
+        /// Ignore explicit flush on configuration reload or application shutdown
+        /// </summary>
+        public bool IgnoreExplicitFlush { get; set; }
```
304NotModified: Not sure about this name. I think "explicit" is also a bit difficult here (I think it should be implicit then)

Proposal: FlushOnEvents.

Or, It would be cool if we could split into 2 options, FlushOnShutdown and FlushOnReload, but I expect that's far more difficult to implement?

304NotModified:
> Notice PR is configured to be merged directly into master!

Thanks! This isn't visible on mobile view. Anyway, I expected it was on master (good choice) :)

src/NLog/Targets/Wrappers/AutoFlushTargetWrapper.cs
snakefoot: Not possible to split into more settings. It will ignore config-reload, application-shutdown and explicit-flush. All of them call the explicit `LogManager.Flush()`

304NotModified: What about, OnlyFlushOnCondition? 

(PS: LogManager.Flush() is explicit, Logmanager.ShutDown is implicit IMO)

snakefoot: To me everything is explicit, except the implicit flush performed by autoflushwrapper.

snakefoot: Renamed property to `FlushOnConditionOnly`

snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/AutoFlushWrapper-target

Issue #3519, opened 2019-07-19
Custom renderer for exceptions


Giorgi:
**NLog version**: 4.6.6

**Platform**: All

**Current NLog config** (xml or C#, if relevant)

```xml
<nlog>
  <targets>
    <target xsi:type="File" name="f" fileName="${basedir}/logs/${shortdate}.log"
            layout="${longdate} ${uppercase:${level}} ${message} ${exception:format=message}" />
  </targets>
  <rules>
    <logger name="*" minlevel="Debug" writeTo="f" />
  </rules>
</nlog>
```

Most of the exception subclasses have custom properties which are very useful when diagnosing the cause of the exception. For example `Win32Exception` has a `NativeErrorCode` property. If we want the property to get logged we need to add a catch block specifically for `Win32Exception` and log ` NativeErrorCode`. This can lead to useless and duplicated catch blocks.

I propose adding a way to create custom renderers for specific exception subclasses so the instances of that exception type will be formatted with that custom renderer. Here is an example for registering custom exception renderer for `Win32Exception` which will log `NativeErrorCode` in addition to default data logged for every exception:

```
ExceptionLayoutRenderer.RegisterExceptionRenderer<Win32Exception>(exception => $"NativeErrorCode: {exception.NativeErrorCode}");
```

With this renderer logging an instance of `Win32Exception` will result in this output:

`2019-07-19 23:44:53.9813 ERROR  The system cannot find the file specified NativeErrorCode: 2`

Giorgi:
I have a working code implemented in this commit: [Added custom renderer for exceptions.](https://github.com/Giorgi/NLog/commit/e367968f32c5e074060928270e8e4179934b1e21) and I can send a PR.

304NotModified:
Cool!  PR would be nice!

I think I see a small issue, but it's better to shoe that in a code review

Giorgi:
PR created

304NotModified:
related, In NLog 4.6.7 `${object-path}` has been added, so you could render properties of an (exception) object.

### Example:
Assume we catch new ExternalException("Exception with errorCode", 5000)
```
${exception:objectpath=HResult} // results in 5000
```

Giorgi:
That's nice but it's weakly typed and easy to make a typo. Strongly typed renderer defined in code would be better and probably faster as it would invoke user provided delegate instead of using reflection

snakefoot:
Work is now being made on opening the access to ObjectReflectionCache. See also #3606

[...]

PR #3520, opened 2019-07-19
Custom renderer for exceptions

Giorgi:
Implementation of custom renderer for exceptions for issue #3519

### DIFF
--- a/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
@@ -303,28 +303,31 @@ namespace NLog.LayoutRenderers
                 }
             }
 
-            var exceptionType = currentException.GetType();
-            while (exceptionType != typeof(Exception))
+            if (_customRenderingfunctions.Count > 0)
             {
-                if (_customRenderingfunctions.TryGetValue(exceptionType, out var renderer))
+                var exceptionType = currentException.GetType();
+                while (exceptionType != typeof(Exception))
                 {
-                    try
+                    if (_customRenderingfunctions.TryGetValue(exceptionType, out var renderer))
                     {
-                        builder.Append(Separator);
-                        var render = renderer(currentException);
-                        builder.Append(render);
-                    }
-                    catch (Exception exception)
-                    {
-                        var message = $"Exception in {typeof(ExceptionLayoutRenderer).FullName}.AppendException(): {exception.GetType().FullName}. Custom renderer for {currentException.GetType().Name} threw an exception";
+                        try
+                        {
+                            builder.Append(Separator);
+                            var render = renderer(currentException);
+                            builder.Append(render);
+                        }
+                        catch (Exception exception)
+                        {
+                            var message = $"Exception in {typeof(ExceptionLayoutRenderer).FullName}.AppendException(): {exception.GetType().FullName}. Custom renderer for {currentException.GetType().Name} threw an exception";
 
-                        InternalLogger.Warn(exception, message);
+                            InternalLogger.Warn(exception, message);
+                        }
+                        break;
                     }
-                    break;
-                }
-                else
-                {
-                    exceptionType = exceptionType.BaseType;
+                    else
+                    {
+                        exceptionType = exceptionType.BaseType;
+                    }
                 }
             }
         }
--- a/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
@@ -313,9 +313,8 @@ namespace NLog.LayoutRenderers
                 }
                 catch (Exception exception)
                 {
-                    var message = $"Exception in {typeof(ExceptionLayoutRenderer).FullName}.AppendException(): {exception.GetType().FullName}.";
-                    builder.Append("NLog message: ");
-                    builder.Append(message);
+                    var message = $"Exception in {typeof(ExceptionLayoutRenderer).FullName}.AppendException(): {exception.GetType().FullName}. Custom renderer for {currentException.GetType().Name} threw an exception";
+                    
                     InternalLogger.Warn(exception, message);
                 }
             }
--- a/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
@@ -303,19 +303,28 @@ namespace NLog.LayoutRenderers
                 }
             }
 
-            if (_customRenderingfunctions.TryGetValue(currentException.GetType(), out var renderer))
+            var exceptionType = currentException.GetType();
+            while (exceptionType != typeof(Exception))
             {
-                try
+                if (_customRenderingfunctions.TryGetValue(exceptionType, out var renderer))
                 {
-                    builder.Append(Separator);
-                    var render = renderer(currentException);
-                    builder.Append(render);
+                    try
+                    {
+                        builder.Append(Separator);
+                        var render = renderer(currentException);
+                        builder.Append(render);
+                    }
+                    catch (Exception exception)
+                    {
+                        var message = $"Exception in {typeof(ExceptionLayoutRenderer).FullName}.AppendException(): {exception.GetType().FullName}. Custom renderer for {currentException.GetType().Name} threw an exception";
+
+                        InternalLogger.Warn(exception, message);
+                    }
+                    break;
                 }
-                catch (Exception exception)
+                else
                 {
-                    var message = $"Exception in {typeof(ExceptionLayoutRenderer).FullName}.AppendException(): {exception.GetType().FullName}. Custom renderer for {currentException.GetType().Name} threw an exception";
-                    
-                    InternalLogger.Warn(exception, message);
+                    exceptionType = exceptionType.BaseType;
                 }
             }
         }
--- a/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
@@ -53,6 +53,7 @@ namespace NLog.LayoutRenderers
         private string _format;
         private string _innerFormat = string.Empty;
         private readonly Dictionary<ExceptionRenderingFormat, Action<StringBuilder, Exception>> _renderingfunctions;
+        private static readonly Dictionary<Type, Func<Exception, string>> _customRenderingfunctions = new Dictionary<Type, Func<Exception, string>>();
 
         private static readonly Dictionary<string, ExceptionRenderingFormat> _formatsMapping = new Dictionary<string, ExceptionRenderingFormat>(StringComparer.OrdinalIgnoreCase)
                                                                                                     {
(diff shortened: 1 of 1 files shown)


src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -285,11 +296,29 @@ private void AppendException(Exception currentException, IEnumerable<ExceptionRe
...
                 var currentRenderFunction = _renderingfunctions[renderingFormat];
                 currentRenderFunction(builder, currentException);
+
                 if (builder.Length == beforeRenderLength && builder.Length != orgLength)
                 {
                     builder.Length = orgLength;
                 }
             }
+
+            if (_customRenderingfunctions.TryGetValue(currentException.GetType(), out var renderer))
```
304NotModified: Is this isn't a issue for subclasses?

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -177,6 +178,16 @@ public List<ExceptionRenderingFormat> InnerFormats
...
         }
 
+        /// <summary>
+        /// Registers a custom exception renderer type
+        /// </summary>
+        /// <typeparam name="T">Type of the exception</typeparam>
+        /// <param name="renderer">The renderer for converting exception to string</param>
+        public static void RegisterExceptionRenderer<T>(Func<T, string> renderer) where T : Exception
+        {
+            _customRenderingfunctions.Add(typeof(T), exception => renderer((T) exception));
```
304NotModified: `_customRenderingfunctions[type] = ` is maybe better (no exception when duplicate)

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -285,11 +296,29 @@ private void AppendException(Exception currentException, IEnumerable<ExceptionRe
...
+                    builder.Append(Separator);
+                    var render = renderer(currentException);
+                    builder.Append(render);
+                }
+                catch (Exception exception)
+                {
+                    var message = $"Exception in {typeof(ExceptionLayoutRenderer).FullName}.AppendException(): {exception.GetType().FullName}.";
+                    builder.Append("NLog message: ");
+                    builder.Append(message);
+                    InternalLogger.Warn(exception, message);
```
304NotModified: Only log to internallog here?

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
Giorgi: @304NotModified Do you mean that if there is no custom renderer for this type it should use a renderer for base type?

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
Giorgi: Probably, I copied it from `AppendMessage` in the same class

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
Giorgi: Done

snakefoot:
One could also consider making use of ObjectReflectionCache, so it will automatically output any extra properties on the Exception-object (Without object-reference-recursion by default).

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
304NotModified: If you have the classes like this,

    Exception <- MyExceptionBase <- MyException

And registering for MyExceptionBase, it won't work for MyException?

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
304NotModified: Have to check this :)

snakefoot:
Making use of ObjectReflectionCache would also allow "automatic" support of generic exceptions and base-classes:

https://stackoverflow.com/questions/57147623/nlog-capture-faultexceptiontdetail-in-log

304NotModified:
Cool idea!

Giorgi:
The disadvantage of `ObjectReflectionCache` is that it handles only properties. For example in case of `WebException` you often need to read the stream by calling `WebException.Response.GetResponseStream` method which you can't do with ObjectReflectionCache

snakefoot:
> The disadvantage of ObjectReflectionCache is that it handles only properties. 

I you take a closer look then you will see it is possible to register custom-methods that can extract properties from an object of a certain type.

But the code is currently not public accessible, because I waited for a proper use-case to arrive (maybe this is the one?), but I also try to avoid static registration methods so not that keen on opening the door until we have proper DependencyInjection (DI) available in NLog. 

Maybe the registration of type-reflection-helpers should just be registration of a generic interface like ex. `IPropertyReflection<T>`. Then one could just register the interface in the global DI-container instead.

But I will try to show how to use `ObjectReflectionCache` then we can continue the discussion of how to register custom methods for extracting custom values. But my time is limited, so not able to work full time on NLog. Lots of stuff is missing for NLog 5.0

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
304NotModified: Only internal log please, thanks!

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
304NotModified: You could use `IsAssignableFrom` for this, unfortunately it's not working a dictionary then. (or you should write a fancy comparer?)

304NotModified:
> I you take a closer look then you will see it is possible to register custom-methods that can extract properties from an object of a certain type.

I think this is nice, but maybe to heavy for now?

304NotModified:
maybe do both, but I think objectpath in the exceptionrenderer isn't a bad idea (like that one in https://github.com/NLog/NLog/wiki/EventProperties-Layout-Renderer).

For those properties you don't need c# code for configuring

snakefoot:
I'm not thinking objectPath but more like format=@. But instead of json. Then it adds extra properties not included in default tostring output. Ex. format=tostring,data,properties. Where properties means starting the reflection cache but very restricted so it will not navigate a DbContext (will also skip Data, StackTrace, etc)


Sent from my Sony Xperia


maybe do both, but I think objectpath in the exceptionrenderer isn't a bad idea (like that one in https://github.com/NLog/NLog/wiki/EventProperties-Layout-Renderer).

For those properties you don't need c# code for configuring

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
304NotModified: I mean IsInstanceOfType

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
Giorgi: @304NotModified Done

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
Giorgi: @304NotModified I just loop thought the base types if there is no renderer for current exception type

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -303,19 +303,28 @@ private void AppendException(Exception currentException, IEnumerable<ExceptionRe
                 }
             }
 
-            if (_customRenderingfunctions.TryGetValue(currentException.GetType(), out var renderer))
+            var exceptionType = currentException.GetType();
+            while (exceptionType != typeof(Exception))
```
snakefoot: Please check if `_customRenderingfunctions.Count > 0` before starting all this logic.

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -285,11 +296,37 @@ private void AppendException(Exception currentException, IEnumerable<ExceptionRe
...
+                    {
+                        var message = $"Exception in {typeof(ExceptionLayoutRenderer).FullName}.AppendException(): {exception.GetType().FullName}. Custom renderer for {currentException.GetType().Name} threw an exception";
+
+                        InternalLogger.Warn(exception, message);
+                    }
+                    break;
+                }
+                else
+                {
+                    exceptionType = exceptionType.BaseType;
```
304NotModified: I think this doens't work for interface ;)

maybe a some test cases first?

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
snakefoot: How do you use registered `_customRenderingfunctions` alone without getting the output from the configured `format`-property ? (Default = `message`).

Giorgi: @snakefoot You can't do that currently. I don't think anyone would want to log specific properties and not message. I see custom renderer as an additional detail provider and if everyone needs a message NLog shouldn't force users to include message in every custom renderer

snakefoot: I was trying to nudge you in the direction that you should add new enum value to ExceptionRenderingFormat

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
Giorgi: @304NotModified When would exceptionType ever be an interface? `currentException.GetType()` will always return an exception type representing a class so accessing BaseType hierarchically should never be an interface type.

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
304NotModified: >  I don't think anyone would want to log specific properties and not message.

one use case is to log both, but not in the same (database) column/field.

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
304NotModified: ` public static void RegisterExceptionRenderer<IMyExceptionInterface>(...)`

Giorgi: RegisterExceptionRenderer is limited to Exception subclasses at the moment:

`public static void RegisterExceptionRenderer<T>(Func<T, string> renderer) where T : Exception`

304NotModified: not sure if that's a good idea. We could enumerate the list and use IsInstanceOfType? Then it also works if someone nice abstractions. :)

Giorgi: If someone registers a renderer for `MyExceptionBase` and than registers for `MyException` `IsInstanceOfType` will return true for both of them

Giorgi: @304NotModified Also, if the type implements two interfaces and someone registers custom renderer for both of them which one should be used ?

snakefoot:
Instead of registering custom formatter on the exception-layout-renderer. Then one could also consider injecting the custom rendering using the IValueFormatter-interface.

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
304NotModified: Both or closest path.

304NotModified:
I think this PR isn't going in the right direction as it doesn't support interfaces. I think we should do the design before implementing. 

Another option is to implement this is your own code (derive from the layout renderer) and maybe change something in NLog for the integration

Giorgi:
@304NotModified I can add support for interfaces if supporting them is such a top priority.

304NotModified:
@Giorgi  I think without interface, this will be a strange addition.

But I also added a new layout renderer, just for rendering properties. You could see it here, https://github.com/NLog/NLog/pull/3531 (needs review)

That will handle properties on more levels than ${exception} and don't need any configuration in code :)

304NotModified:
Closing this due to inactivity. Feel free to open a new PR in the future!

PR #3531, opened 2019-07-29
Added ${object-path} / ${exception:objectpath=PropertyName}, for rendering a property of an object (e.g. an exception)

304NotModified:
Examples:

```
${object-path:${exception}:path=PropertyName}
${exception:objectpath=PropertyName}
```

### DIFF
--- a/src/NLog/Internal/ObjectPropertyHelper.cs
+++ b/src/NLog/Internal/ObjectPropertyHelper.cs
@@ -1,0 +1,95 @@
+// 
+// Copyright (c) 2004-2019 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without 
+// modification, are permitted provided that the following conditions 
+// are met:
+// 
+// * Redistributions of source code must retain the above copyright notice, 
+//   this list of conditions and the following disclaimer. 
+// 
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution. 
+// 
+// * Neither the name of Jaroslaw Kowalski nor the names of its 
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission. 
+// 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE 
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE 
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE 
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR 
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS 
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN 
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) 
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF 
+// THE POSSIBILITY OF SUCH DAMAGE.
+// 
+
+namespace NLog.Internal
+{
+    /// <summary>
+    /// Helper for extracting propertyPath
+    /// </summary>
+    internal class ObjectPropertyHelper
+    {
+        private string[] _objectPropertyPath;
+        private ObjectReflectionCache _objectReflectionCache;
+        private ObjectReflectionCache ObjectReflectionCache => _objectReflectionCache ?? (_objectReflectionCache = new ObjectReflectionCache());
+
+        /// <summary>
+        /// Object Path to check
+        /// </summary>
+        public string ObjectPath
+        {
+            get => _objectPropertyPath?.Length > 0 ? string.Join(".", _objectPropertyPath) : null;
+            set => _objectPropertyPath = StringHelpers.IsNullOrWhiteSpace(value) ? null : value.SplitAndTrimTokens('.');
+        }
+
+        /// <summary>
+        /// Try get value from <paramref name="value"/>, using <see cref="ObjectPath"/>, and set into <paramref name="foundValue"/>
+        /// </summary>
+        /// <param name="value"></param>
+        /// <param name="foundValue"></param>
+        /// <returns></returns>
+        public bool TryGetObjectProperty(object value, out object foundValue)
+        {
+            foundValue = null;
+
+            if (_objectPropertyPath == null)
+            {
+                return false;
+            }
+
+            var objectReflectionCache = ObjectReflectionCache;
+            for (int i = 0; i < _objectPropertyPath.Length; ++i)
+            {
+                if (value == null)
+                {
+                    // Found null
+                    foundValue = null;
+                    return true;
+                }
+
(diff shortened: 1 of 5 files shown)


src/NLog/Internal/ObjectPropertyHelper.cs
```diff
@@ -0,0 +1,88 @@
...
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF 
+// THE POSSIBILITY OF SUCH DAMAGE.
+// 
+
+namespace NLog.Internal
+{
+    /// <summary>
+    /// Helper for extracting propertyPath
+    /// </summary>
+    internal class ObjectPropertyHelper
```
304NotModified: Note: extracted from EventPropertiesLayoutRenderer, changed TryGetObjectProperty (ref to out parameter)

src/NLog/Internal/ObjectPropertyHelper.cs
```diff
@@ -0,0 +1,88 @@
...
+    {
+        private string[] _objectPropertyPath;
+        private ObjectReflectionCache _objectReflectionCache;
+        private ObjectReflectionCache ObjectReflectionCache => _objectReflectionCache ?? (_objectReflectionCache = new ObjectReflectionCache());
```
304NotModified: Maybe `static`?

src/NLog/LayoutRenderers/Wrappers/ObjectPathRendererWrapper.cs
```diff
@@ -0,0 +1,143 @@
...
+namespace NLog.LayoutRenderers.Wrappers
+{
+    /// <summary>
+    /// Render a single property of a object
+    /// </summary>
+    [LayoutRenderer("Object-path")]
+    [AmbientProperty("path")]
+    [AppDomainFixedOutput]
+    [ThreadAgnostic]
+    [ThreadSafe]
```
304NotModified: Not sure about the attributes 
    [AppDomainFixedOutput]
    [ThreadAgnostic]
    [ThreadSafe]

src/NLog/LayoutRenderers/Wrappers/WrapperLayoutRendererBase.cs
```diff
@@ -49,14 +52,29 @@ namespace NLog.LayoutRenderers.Wrappers
...
+            {
+                _inner = value;
+                InnerChanged?.Invoke(this, EventArgs.Empty);
+            }
+        }
+
+        /// <summary>
+        /// Notify when <see cref="Inner"/> has been changed
+        /// </summary>
+        protected event EventHandler InnerChanged;
```
304NotModified: For performance, but maybe overdesign

src/NLog/Internal/ObjectPropertyHelper.cs
304NotModified: And broke it somehow 😂

src/NLog/LayoutRenderers/Wrappers/ObjectPathRendererWrapper.cs
snakefoot: It can only have `[ThreadSafe]`

src/NLog/LayoutRenderers/Wrappers/ObjectPathRendererWrapper.cs
```diff
@@ -0,0 +1,142 @@
...
+using NLog.Internal;
+using NLog.Layouts;
+
+namespace NLog.LayoutRenderers.Wrappers
+{
+    /// <summary>
+    /// Render a single property of a object
+    /// </summary>
+    [LayoutRenderer("Object-path")]
+    [AmbientProperty("path")]
```
snakefoot: Would think it would be better with AmbientProperty=object-path. Since "path" is very vague.

src/NLog/LayoutRenderers/Wrappers/WrapperLayoutRendererBase.cs
snakefoot: Alternative one could use ReferenceEquals in the specialized class. And if the cached parsing points to a different Layout, then refresh the cached parsing.

src/NLog/LayoutRenderers/Wrappers/ObjectPathRendererWrapper.cs
304NotModified: I'm thinking to rename (both!) it to property-path, what do you think?

304NotModified: that's right, event-properties has already "ObjectPath"

304NotModified: done :)

src/NLog/LayoutRenderers/Wrappers/ObjectPathRendererWrapper.cs
304NotModified: thanks!

src/NLog/LayoutRenderers/Wrappers/WrapperLayoutRendererBase.cs
```diff
@@ -49,14 +51,32 @@ namespace NLog.LayoutRenderers.Wrappers
...
+
+                if (changed)
+                    InnerChanged?.Invoke(this, EventArgs.Empty);
+            }
+        }
+
+        /// <summary>
+        /// Notify when <see cref="Inner"/> has been changed
+        /// </summary>
+        protected event EventHandler InnerChanged;
```
snakefoot: Could InnerChanged be made internal until needed by others? I was actually thinking that this ReferenceEquals code could just be in ObjectPathRendererWrapper.RenderInnerAndTransform (ReferenceEquals between Inner and _innerAsSimple). No need to have the caching-optimization for one renderer bleeding into the base-class. Not like the logic about caching disappears from ObjectPathRendererWrapper

304NotModified: Unfortunately c# 7.3 is still not working (private protected :( )

304NotModified: done!

src/NLog/LayoutRenderers/Wrappers/ObjectPathRendererWrapper.cs
```diff
@@ -0,0 +1,153 @@
...
+        private SimpleLayout _innerAsSimple;
+
+        /// <summary>
+        /// Gets or sets the object-property-navigation-path for lookup of nested property
+        ///
+        /// Shortcut for <see cref="ObjectPath"/>
+        /// </summary>
+        public string Path
+        {
+            get => ObjectPath;
```
304NotModified: @snakefoot not keen on this, but otherwise you need this:

```
${object-path:${exception}:objectpath=ParamName}
```

PS dashes won't work in ambient properties

snakefoot: Why is this not possible `${exception:objectpath=ParamName}` ?

304NotModified: yes that works!

This works now:

```
${object-path:${exception}:path=ParamName}
${exception:objectpath=ParamName}
```

But I don't prefer: (therefor the delegated property)

```
${object-path:${exception}:objectpath=ParamName}
```

and also this isn't possible: (dashed ambient property)

```
${exception:object-path=ParamName}
```

src/NLog/LayoutRenderers/Wrappers/ObjectPathRendererWrapper.cs
304NotModified: @snakefoot the tests says also threadadgnostic?

image

snakefoot: Yes I guess `[ThreadAgnostic]` is needed if you want `IRawValue`-support.

304NotModified: Any suggestions how to solve this?

snakefoot: Think I have to retract my statement about "It can only have [ThreadSafe]". Since this is a layout-wrapper then it will be automatically restricted by its Inner-Layout. So it should be okay to have both [ThreadAgnostic] + [ThreadSafe]

304NotModified:
docs -> https://github.com/NLog/NLog/wiki/objectpath-layout-renderer

PR #3534, opened 2019-07-30
MessageTemplateParameters.IsValidTemplate is behaving funky #3375

### LINKED ISSUE #3375: MessageTemplateParameters.IsValidTemplate is behaving funky
got some interesting results with `MessageTemplateParameters.IsValidTemplate`:

image

With an empty list as `object[] parameters` , is always says `IsValidTemplate=true` - independent of the message-template. 

See also unit test here: https://github.com/NLog/NLog/pull/3286

nishantgupta204:
Added the test case and fixed the funky issue, fixes #3375

### DIFF
--- a/src/NLog/MessageTemplates/MessageTemplateParameters.cs
+++ b/src/NLog/MessageTemplates/MessageTemplateParameters.cs
@@ -36,6 +36,7 @@ namespace NLog.MessageTemplates
     using System;
     using System.Collections;
     using System.Collections.Generic;
+    using System.Linq;
     using NLog.Common;
 
     /// <summary>
@@ -80,7 +81,7 @@ namespace NLog.MessageTemplates
             var hasParameters = parameters != null && parameters.Length > 0;
             bool isPositional = hasParameters;
             bool isValidTemplate = !hasParameters;
-            _parameters = hasParameters ? ParseMessageTemplate(message, parameters, out isPositional, out isValidTemplate) : Internal.ArrayHelper.Empty<MessageTemplateParameter>();
+            _parameters = ParseMessageTemplate(message, parameters, out isPositional, out isValidTemplate);
             IsPositional = isPositional;
             IsValidTemplate = isValidTemplate;
         }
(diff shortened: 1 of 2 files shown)


304NotModified:
Thanks!

304NotModified:
it looks like NormalStringFormatTest fails? See details of AppVeyor and/or Travis

snakefoot:
Glad someone wants to give `MessageTemplateParameters` some love. It was created as a quick fix for the Seq-Nlog-Target that wanted access to the NLog message-template-parser.

But please be aware of the following:

- NLog has always ignored the message-template when no parameters has been provided. This is both a performance optimization but also allow users to write "blob"-data without. This means the following line is wrong (should only parse when parameters provided):

`_parameters = ParseMessageTemplate(message, parameters, out isPositional, out isValidTemplate);`

- NLog was not the first framework to support message-templates. Actually think Microsoft-Extension-Logging (MEL) is the defacto-standard to follow. And pretty sure that when given the following input `{a} {a}` then MEL will see it as a faulty message-template with two parameters. I believe NLog should continue to return two parameters but signal that the message-template is not valid. This extra checking will ofcourse give a performance overhead.

- But NLog is special as it supports both string-format-template (positional) and message-template. So your idea with seeing `{0} {0}` as a positional string-format-template with a single parameter is correct. But how should the positional parameters be returned? I believe the logic is currently wrong as it should return parameters in the positional order, where MessageTemplateParameters returns string-format-template-parameters in their declared order (Ex. `{1} {0}` should return parameter `{0}` first, but currently it returns `{1}` first).

snakefoot:
After checking the code, then I actually think the current logic is valid for both positional-message-templates and structured-message-templates. Since one can provide different format-specifier for the same parameter positional-parameter (And `MessageTemplateParameter` includes the Format).

And the performance overhead introduced by checking if structed-message-template is returning duplicate names doesn't give any added bonus (Since MessageTemplateParameters is only used by Seq NLog Target that ignores `IsValidTemplate` completely)

304NotModified:
@snakefoot current logic, is that this PR or the master?

snakefoot:
> @snakefoot current logic, is that this PR or the master?

Current logic is the logic in master.

nishantgupta204:
@snakefoot Thanks for reviewing this PR. I have closed this PR. I believe that you are right in this context

Issue #3542, opened 2019-08-12
Custom target with async attribute get incorrect timestamp and order of logs


Kirya522:
**NLog version**: 4.6.5

**Platform**: WebApi 2.1 .net47 

**Current NLog config**

```xml
<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <targets async="true">
      <target name="graylogt" xsi:type="GraylogHttp" facility="WebApi" graylogServer="http://graylog" graylogPort="12201">
        <parameter name="current_user" layout="${aspnet-user-identity}" />
        <parameter name="exception_type" layout="${exception:Type}" />
        <parameter name="request" layout="${aspnet-request-method} ${aspnet-request-url}" />
        <parameter name="site" layout="${iis-site-name} ${aspnet-appbasepath}" />
        <parameter name="request_body" layout="${onexception:${aspnet-request-posted-body}:when=level>=LogLevel.Error}" />
      </target>
    </targets>
    <rules>
     <logger name="*" minlevel="debug" writeTo="graylogt" />
    </rules>
  </nlog>

```
Hi, I use a https://github.com/dustinchilson/NLog.Targets.GraylogHttp for graylog.
In order to don't block the thread for logging I use an async attribute. But logs have incorrect timestamps and order(compare to sync mode). I can't use "sync" logs, because every log operation takes 0.2 sec.
- What is the current result? 

> Incorrect order and timestamp of logs

- What is the expected result? 

> Correct order and timestamps

304NotModified:
Hi, 

Could you please test this with a logfile (file target)? Is there the same issue?

Kirya522:
I have tested file target before all the same, except in sync mode file is much faster than that target

304NotModified:
Please share your logger calls because wrong timestamp sounds unlikely

snakefoot:
@Kirya522 I think `GraylogHttp` is not guaranteed to send http-requests in-order. It only promises to throttle when http-exceptions occurs.

Could you try using [NLog.GelfLayout ver. 0.4](https://www.nuget.org/packages/NLog.GelfLayout) and try this instead:

```xml
<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <extensions>
    <add assembly="NLog.Web.AspNetCore" />
[...]

PR #3546, opened 2019-08-13
NetworkTarget - HttpNetworkSender no longer sends out-of-order

### LINKED ISSUE #3544: Network Target with Http can become out-of-order
- What is the current result? Network with TCP will ensure delivery in-order. Network with HTTP will allow multiple concurrent HTTP-requests out-of-order (to the same address).
- What is the expected result? Just like TCP is inorder, then HTTP should also be in-order
- Is there a version in which it did work? no

I guess the same issue can occur with WebServer-Target

snakefoot:
And will flush correctly (delays flush completed until request completed). Resolves #3544 (See also #3542)

### DIFF
--- a/src/NLog/Internal/NetworkSenders/HttpNetworkSender.cs
+++ b/src/NLog/Internal/NetworkSenders/HttpNetworkSender.cs
@@ -35,16 +35,15 @@ namespace NLog.Internal.NetworkSenders
 {
     using System;
     using System.Net;
-    using NLog.Common;
 
     /// <summary>
     /// Network sender which uses HTTP or HTTPS POST.
     /// </summary>
-    internal class HttpNetworkSender : NetworkSender
+    internal class HttpNetworkSender : QueuedNetworkSender
     {
-        readonly Uri _addressUri;
+        private readonly Uri _addressUri;
 
-        internal IWebRequestFactory WebRequestFactory { get; set; } = new WebRequestFactory();
+        internal IWebRequestFactory HttpRequestFactory { get; set; } = WebRequestFactory.Instance;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="HttpNetworkSender"/> class.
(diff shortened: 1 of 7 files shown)


src/NLog/Internal/NetworkSenders/HttpNetworkSender.cs
```diff
@@ -80,16 +77,17 @@ protected override void DoSend(byte[] bytes, int offset, int length, AsyncContin
...
                     catch (Exception ex)
                     {
+                        InternalLogger.Error(ex, "NetworkTarget: Error sending http request");
                         if (ex.MustBeRethrownImmediately())
                         {
                             throw; // Throwing exceptions here will crash the entire application (.NET 2.0 behavior)
                         }
 
-                        asyncContinuation(ex);
+                        base.EndRequest((e) => asyncContinuation(ex), null);    // pendingException = null to keep sender alive
```
304NotModified: Is it correct we're ignoring `e`. Maybe use `_` otherwise?

src/NLog/Internal/NetworkSenders/HttpNetworkSender.cs
```diff
@@ -107,12 +105,13 @@ protected override void DoSend(byte[] bytes, int offset, int length, AsyncContin
...
                     catch (Exception ex)
                     {
+                        InternalLogger.Error(ex, "NetworkTarget: Error starting http request");
                         if (ex.MustBeRethrown())
                         {
                             throw;
                         }
 
-                        asyncContinuation(ex);
+                        base.EndRequest((e) => asyncContinuation(ex), null);    // pendingException = null to keep sender alive
```
304NotModified: Idem:

> Is it correct we're ignoring `e`. Maybe use `_` otherwise?

src/NLog/Internal/NetworkSenders/HttpNetworkSender.cs
snakefoot: Fixed

src/NLog/Internal/NetworkSenders/HttpNetworkSender.cs
snakefoot: Fixed

snakefoot:
Added unit test that shows that it will not discard pending http-requests because one http-request has failed (NetworkTarget still discards active HttpNetworkSender on error, and will create new)

304NotModified:
Cool!

PR #3184, opened 2019-03-18
Added support for dynamic layout renderer in log level filters (e.g. minLevel, maxLevel) 

snakefoot:
Extending #2709 and resolves #1502

allowes:

```xml
<nlog>
   <variable name='myLevel' value='debug'/>
    <rules>
      <logger minLevel='${myLevel}'/>
    </rules>
</nlog>
```

And you do this:

```c#
LogManager.KeepVariablesOnReload = true;
LogManager.Configuration.Variables["myLevel"] = "Info";
LogManager.Configuration.Reload();
```

PR #3560, opened 2019-08-23
WhenMethodFilter - Support dynamic filtering using lambda

snakefoot:
Small bonus to #3184. Allows one to inject a delegate. In case one doesn't like Layout-variables.

Note merge to master (NLog ver. 4.6.7)

Small building block that one day will open the door for fluent-config-api.


304NotModified:
Cool!

IMO func <> delegate :angel:

Issue #3025, opened 2018-11-22
NLog fails silently with malformed config file


binaryunary:
**NLog version**: 4.5.11

**Platform**: .NET Core 2

**Current NLog config** (xml or C#, if relevant)

```xml
<?xml version="1.0" encoding="utf-8" ?>
<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd"
      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
      throwConfigExceptions="true"
      autoReload="true"
      internalLogLevel="Error"
      internalLogIncludeTimestamp="true" >>>>> THIS SHOULD NOT BE HERE <<<<<
      internalLogToConsole = "true"
      internalLogToConsoleError = "true"
      internalLogFile="log/internal-nlog.txt">

  <variable name="AppName" value="test"/>
  <!-- Note! Relative LogRoot for local runs -->
  <variable name="LogRoot" value="log"/>
  <variable name="DatePrefix" value="${date:format=yyyy-MM-dd:universalTime=true}"/>
  <variable name="CompleteLogFile" value="${LogRoot}/${DatePrefix}.${AppName}.complete.log"/>
  <variable name="ErrorLogFile" value="${LogRoot}/${DatePrefix}.${AppName}.error.log"/>
  <variable name="DefaultLayout" value="${longdate}|${event-properties:item=EventId.Id}|${logger}|${uppercase:${level}}|${message} ${replace-newlines:${exception:format=ToString}}"/>

  <targets>
    <target xsi:type="File" name="completefile" fileName="${CompleteLogFile}" layout="${DefaultLayout}" />
    <target xsi:type="Console" name="logconsole" layout="${DefaultLayout}"/>
  </targets>

  <rules>
    <logger name="*" minlevel="Debug" writeTo="completefile" />
    <logger name="*" minlevel="Trace" writeTo="logconsole" />
  </rules>
</nlog>
```

This initially came up when I noticed that I had accidentally added the `internalLogLevel` attribute twice:
```xml
internalLogLevel="Error"
internalLogLevel="Debug"
```

- What is the current result?
Nothing gets logged, no errors are shown neither in console nor in the internal (NLog) log file; no stack trace anywhere.

- What is the expected result?
By setting `throwConfigExceptions="true"` I'd expect to see an exception somewhere if parsing the config file fails.

- Can you help us by writing an unit test?
Minimal example highlighting the issue: https://github.com/binaryunary/nlog-failure

[...]

PR #3562, opened 2019-08-24
XmlLoggingConfiguration - Support ThrowConfigExceptions=true when invalid xml

### LINKED ISSUE #3178: Handle invalid nlog.config / XML better
> When the xml-parser fails to parse the input stream, then NLog should scan the configuration-blob-stream for these strings (without trying to parse as xml):

> throwConfigExceptions="true"
> throwExceptions="true"

this needs probably a regex. Search should be case insensitive and sinqle quote and spaces should also work.

Unit tests are needed for this, TTD recommended

snakefoot:
Resolves #3178 and #3025, so it will correctly throw exception when enabled in nlog.config (Even when invalid xml). Have the side effect that it will react to xml-comments that contains throwConfigExceptions=true (or throwExceptions=true), but only when invalid xml.

Unknown NLog-targets will no longer lead to broken NLog-config and nothing logged. Will still fail as expected when throwConfigExceptions=true (or throwExceptions=true)

Notice merge into master. Refactorings extracted from #3521

### DIFF
--- a/src/NLog/Config/LoggingConfigurationParser.cs
+++ b/src/NLog/Config/LoggingConfigurationParser.cs
@@ -278,16 +278,21 @@ namespace NLog.Config
                 var internalLogLevel = LogLevel.FromString(attributeValue?.Trim());
                 return internalLogLevel;
             }
-            catch (Exception e)
+            catch (Exception exception)
             {
-                const string message = "attribute '{0}': '{1}' isn't valid LogLevel. {2} will be used.";
-                var configException =
-                    new NLogConfigurationException(e, message, attributeName, attributeValue, @default);
-                if (configException.MustBeRethrown())
+                if (exception.MustBeRethrownImmediately())
                 {
                     throw;
                 }
 
+                const string message = "attribute '{0}': '{1}' isn't valid LogLevel. {2} will be used.";
+                var configException =
+                    new NLogConfigurationException(exception, message, attributeName, attributeValue, @default);
+                if (configException.MustBeRethrown())
+                {
+                    throw configException;
+                }
+                InternalLogger.Error(exception, configException.Message);
                 return @default;
             }
         }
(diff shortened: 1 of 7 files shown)


src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -77,7 +77,7 @@ public class XmlLoggingConfiguration : LoggingConfigurationParser
         /// Initializes a new instance of the <see cref="XmlLoggingConfiguration" /> class.
         /// </summary>
         /// <param name="fileName">Configuration file to be read.</param>
-        public XmlLoggingConfiguration([CanBeNull] string fileName)
+        public XmlLoggingConfiguration([NotNull] string fileName)
```
304NotModified: looks like a change which is not suitable for a path version? (breaking behavior change)

snakefoot: Well the `[CanBeNull]` is a bug. You cannot load an XmlLoggingConfiguration from a null-filename.

src/NLog/Config/LoggingConfigurationParser.cs
```diff
@@ -1303,15 +1273,29 @@ private static bool ParseBooleanValue(string propertyName, string value, bool de
...
             }
         }
 
+        private bool MustThrowConfigException(NLogConfigurationException configException)
+        {
+            if (configException.MustBeRethrown() || (LogFactory.ThrowConfigExceptions ?? LogFactory.ThrowExceptions))
+                return true;
+
+            if (configException.InnerException != null)
+                InternalLogger.Error(configException.InnerException, configException.Message);
```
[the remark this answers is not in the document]
snakefoot: Yes you are right. No need to log twice. Have removed the extra-logging.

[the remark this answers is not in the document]
snakefoot: Like the idea to pass local/active `LoggingConfiguration` / `LogFactory` as parameter to `MustBeRethrown()`. But think that is for NLog 5.0

src/NLog/Config/XmlLoggingConfiguration.cs
304NotModified: If we're sure we wont break stuff, I'm happy :) 👍

Issue #3548, opened 2019-08-16
Error when using password with semicolon


grungondola:
**NLog version**: 4.6.6

**Platform**: .Net 4.6.1 /  .NET Standard 2.0 (dual framework)

**Current NLog config**

```c#
// C# (Replaced some private/sensitive data with generic data)
LoggingConfiguration config = new LoggingConfiguration();
StringBuilder dbCommandBuilder = new StringBuilder();
dbCommandBuilder.AppendLine("INSERT INTO [dbo].[Log] ([ColumnA],[ColumnB],[ColumnC])");
dbCommandBuilder.Append("VALUES(@column_a, @column_b, @column_c)");
DatabaseTarget databaseTarget = new DatabaseTarget("database")
{
    DBDatabase = "DatabaseName",
    DBHost = "123.123.123.123",
    DBUserName = "Username",
    DBPassword = "passw;ord",
    CommandText = dbCommandBuilder.ToString(),
    DBProvider = "System.Data.SqlClient",
    KeepConnection = false
};
databaseTarget.Parameters.Add(new DatabaseParameterInfo("@exception", Layout.FromString("${exception:format=Message:innerFormat=Message:maxInnerExceptionLevel=10}"))); // Multiple of these for various parameters
config.AddTarget(databaseTarget);
config.AddRule(LogLevel.Warn, LogLevel.Fatal, databaseTarget);
LogManager.Configuration = config;

InternalLogger.LogToConsole = true;
InternalLogger.LogLevel = LogLevel.Error;
```

- What is the current result?
Not able to connect to the database because the password has a semicolon in it, which is being interpreted as the end of the password and the start of a new keyword
- What is the expected result?
I expect it to connect to the database using the password and log the errors (works in other environments with different passwords using the same code)
- Did you checked the [Internal log](https://github.com/NLog/NLog/wiki/Internal-Logging)?
Yes
- Please post full exception details (message, stacktrace, inner exceptions)
```
2019-08-16 12:11:03.0113 Error DatabaseTarget(Name=database): Error when writing to database. Exception: System.ArgumentException: Keyword not supported: 'ord;database'.
at System.Data.Common.DbConnectionOptions.ParseInternal(Dictionary`2 parsetable, String connectionString, Boolean buildChain, Dictionary`2 synonyms, Boolean firstKey)
at System.Data.Common.DbConnectionOptions..ctor(String connectionString, Dictionary`2 synonyms)
[...]

PR #3566, opened 2019-08-26
DatabaseTarget: escape specials chars password properly, handle empty username and database correctly

304NotModified:
Fixes https://github.com/NLog/NLog/issues/3548

### DIFF
--- a/src/NLog/Targets/DatabaseTarget.cs
+++ b/src/NLog/Targets/DatabaseTarget.cs
@@ -439,7 +439,7 @@ namespace NLog.Targets
                     InternalLogger.Warn(ex, "DatabaseTarget(Name={0}): DbConnectionStringBuilder failed to parse '{1}' ConnectionString", Name, ConnectionStringName);
                 else
 #endif
-                InternalLogger.Warn(ex, "DatabaseTarget(Name={0}): DbConnectionStringBuilder failed to parse ConnectionString", Name);
+                    InternalLogger.Warn(ex, "DatabaseTarget(Name={0}): DbConnectionStringBuilder failed to parse ConnectionString", Name);
             }
 
             return providerName;
@@ -723,10 +723,7 @@ namespace NLog.Targets
                 sb.Append(dbUserName);
                 sb.Append(";Password=");
                 var password = RenderLogEvent(DBPassword, logEvent);
-                if (password.Contains(";"))
-                {
-                    password = $"'{password}'";
-                }
+                password = EscapeValueForConnectionString(password);
                 sb.Append(password);
                 sb.Append(";");
             }
@@ -738,6 +735,47 @@ namespace NLog.Targets
             }
 
             return sb.ToString();
+        }
+
+        /// <summary>
+        /// Escape quotes and semicolons.
+        /// See https://docs.microsoft.com/en-us/previous-versions/windows/desktop/ms722656(v=vs.85)#setting-values-that-use-reserved-characters
+        /// </summary>
+        private static string EscapeValueForConnectionString(string value)
+        {
+            const string singleQuote = "'";
+
+            if (value.StartsWith(singleQuote) && value.EndsWith(singleQuote))
+            {
+                // already escaped
+                return value;
+            }
+            const string doubleQuote = "\"";
+            if (value.StartsWith(doubleQuote) && value.EndsWith(doubleQuote))
+            {
+                // already escaped
+                return value;
+            }
+
+            var containsSingle = value.Contains(singleQuote);
+            var containsDouble = value.Contains(doubleQuote);
+            if (value.Contains(";") || containsSingle || containsDouble)
+            {
+                if (!containsSingle)
+                {
+                    return $"{singleQuote}{value}{singleQuote}";
+                }
+                if (!containsDouble)
+                {
+                    return $"{doubleQuote}{value}{doubleQuote}";
+                }
+
+                // both single and double
+                var escapedValue = value.Replace(doubleQuote, doubleQuote + doubleQuote);
+                return $"{doubleQuote}{escapedValue}{doubleQuote}";
+            }
+
+            return value;
         }
 
         private void EnsureConnectionOpen(string connectionString)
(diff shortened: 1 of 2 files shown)


304NotModified:
@snakefoot doubting if we should escape all parameters.

Also the null check on the Layout have some issues. 

image

snakefoot:
Think it is fine just to auto-escape the password.

How does it work when using ConnectionString-Layout with `Password='passw;rd'` inside?

How Layout handles null is probably a different PR. Would prefer that assigned to null gives null, and assigned to `(string)null` gives empty-layout with FixedText = true.

snakefoot:
@304NotModified Was thinking. What is the Layout already includes escape-quotes?

Should they be added twice when password contains semi-colon?

304NotModified:
ow yes I have to check if it isn't already starting/ending with quotes?

snakefoot:
> ow yes I have to check if it isn't already starting/ending with quotes?

I guess so. Also are there other magic characters in Database-ConnectionStrings, that one need to check for besides semi-colon?

304NotModified:
> The keyword is "Authentication;Info".

If a keyword contains an equal sign (=), it must be preceded by an additional equal sign to indicate that it is part of the keyword.


Copy
Verification==Security=True
The keyword is "Verification=Security".

If multiple equal signs appear, each one must be preceded by an additional equal sign.

304NotModified:
But we need to escape more
https://docs.microsoft.com/en-us/previous-versions/windows/desktop/ms722656(v=vs.85)#setting-values-that-use-reserved-characters

304NotModified:
> How Layout handles null is probably a different PR. Would prefer that assigned to null gives null, and assigned to `(string)null` gives empty-layout with FixedText = true.

https://github.com/NLog/NLog/issues/3573

304NotModified:
Hi @snakefoot , could you please check? Thanks!

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -712,28 +726,83 @@ protected string BuildConnectionString(LogEventInfo logEvent)
...
+                    return $"{singleQuote}{value}{singleQuote}";
+                }
+                if (!containsDouble)
+                {
+                    return $"{doubleQuote}{value}{doubleQuote}";
+                }
+
+                // both single and double
+                var escapedValue = value.Replace(doubleQuote, doubleQuote + doubleQuote);
+                return $"{doubleQuote}{escapedValue}{doubleQuote}";
```
snakefoot: `string.Concat(doubleQuote, escapedValue, doubleQuote)` is faster.

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -712,28 +726,83 @@ protected string BuildConnectionString(LogEventInfo logEvent)
...
+            var containsDouble = value.Contains(doubleQuote);
+            if (value.Contains(";") || containsSingle || containsDouble)
+            {
+                if (!containsSingle)
+                {
+                    return $"{singleQuote}{value}{singleQuote}";
+                }
+                if (!containsDouble)
+                {
+                    return $"{doubleQuote}{value}{doubleQuote}";
```
snakefoot: string.Concat(doubleQuote, value, doubleQuote) is faster.

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -712,28 +726,83 @@ protected string BuildConnectionString(LogEventInfo logEvent)
...
+                return value;
+            }
+
+            var containsSingle = value.Contains(singleQuote);
+            var containsDouble = value.Contains(doubleQuote);
+            if (value.Contains(";") || containsSingle || containsDouble)
+            {
+                if (!containsSingle)
+                {
+                    return $"{singleQuote}{value}{singleQuote}";
```
snakefoot: `string.Concat(singleQuote, value, singleQuote)` is faster.

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -712,28 +726,83 @@ protected string BuildConnectionString(LogEventInfo logEvent)
...
             }
 
             return sb.ToString();
         }
 
+        private string GetRenderedAndEscapedPassword(LogEventInfo logEvent)
+        {
+            if (_fixedDbPassword != null)
+            {
+                return _fixedDbPassword;
```
snakefoot: Like this optimization :)

src/NLog/Targets/DatabaseTarget.cs
304NotModified: Any idea why the compiler rewrites it to `+`?

src/NLog/Targets/DatabaseTarget.cs
304NotModified: Thanks, I was also working on better approach (adapter instead of 2 fields), but that will be another PR

304NotModified: Thanks, I was also working on better approach (adapter instead of 2 fields), but that will be another PR

src/NLog/Targets/DatabaseTarget.cs
snakefoot: What is rewritten to `+` ? This is string.Concat -> https://referencesource.microsoft.com/#mscorlib/system/string.cs,3216

304NotModified: `$"{a}{b}"` -> `a+b` compile time? 

(anyway applied string.concat)

304NotModified: seems to be string.format or string.concat by the compiler 👼 


> Compilation of interpolated strings
If an interpolated string has the type string, it's typically transformed into a String.Format method call. The compiler may replace String.Format with String.Concat if the analyzed behavior would be equivalent to concatenation.

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated

snakefoot: $"{a}{b}" should become `string.Format("{0}{1}", a, b);`. Appends to StringBuilder from global-pool, and then performs ToString. `string.Concat` skips the StringBuilder-overhead.

snakefoot: > seems to be string.format or string.concat by the compiler

Interesting. Wonder what version of the compiler this was introduced.

304NotModified: Nevertheless, better to force it as join indeed!

PR #3574, opened 2019-09-01
Added adapter for layout transformations

304NotModified:


snakefoot:
One could also consider implementing this using the `Layout<T>`-idea. `Layout<T>` is a standard Layout with a type-converter.

I guess `Layout<T>`-idea could also be used to take type-converter that returns a string (after having encoded the string).

304NotModified:
> One could also consider implementing this using the `Layout<T>`-idea. `Layout<T>` is a standard Layout with a type-converter.

Good idea, but then it's logical this is moved to Layout?

Another option is to make it generic so a `Layout<T>` also fits for this.

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -736,7 +727,7 @@ protected string BuildConnectionString(LogEventInfo logEvent)
                 sb.Append("User id=");
                 sb.Append(dbUserName);
                 sb.Append(";Password=");
-                var password = GetRenderedAndEscapedPassword(logEvent);
+                var password = _dbPassword.Render(logEvent);
```
304NotModified: @snakefoot 
AFAIK I can't prevent calling `DbPassword.Render(logEvent);` here, expect by unit tests. Or do you know an method for that?

src/NLog/Internal/TransformedLayout.cs
```diff
@@ -0,0 +1,126 @@
...
+        #endregion
+
+        /// <summary>
+        /// Create TransformedLayout. If <paramref name="layout" /> is null, then the returns null
+        /// </summary>
+        /// <param name="layout">Layout</param>
+        /// <param name="transformation">Transformation method</param>
+        /// <param name="renderLogEvent">Optional renderer, otherwise <see cref="Layouts.Layout.Render" /> is used</param>
+        /// <returns>null if <paramref name="layout" /> is null</returns>
+        public static TransformedLayout Create([CanBeNull] Layout layout,
```
304NotModified: FYI passing the `renderLogEvent` is a bit pity, but guess it's needed.

snakefoot:
One could also implement `Layout<T>`-idea like this to avoid breaking changes:

```
Layout EventId { get => _eventId.Layout; set => _eventId = new Layout<T>(value); }
Layout<int> _eventId;
```

Same could be done for added password-encoding without breaking interfaces.

src/NLog/Internal/TransformedLayout.cs
snakefoot: I'm guessing that is the whole idea with this class to wrap a Layout with a post-processing-method (renderLogEvent). So not sure why it is a pity.

src/NLog/Targets/DatabaseTarget.cs
snakefoot: Not sure I understand. Are the unit-tests performing white-box testing?

304NotModified:
> Layout EventId { get => _eventId.Layout; set => _eventId = new Layout<T>(value); }
> Layout<int> _eventId;

True, but from the API you can't do 

```
eventLogTarget.EventId = 1;
```

OK, that's currently also. But don't like to refactor everything for NLog 5, as NLog 5 should be the next non-patch version

src/NLog/Internal/TransformedLayout.cs
304NotModified: because we have also `Layout.Render` and not also `renderLogEvent` . So 2 methods for the same thing (but with other caching)

src/NLog/Targets/DatabaseTarget.cs
304NotModified: Nope, but still I prefer compile time checks over unit tests ;)

Issue #3580, opened 2019-09-09
JSON: forward slashes (e.g. in URL) are getting escaped


kylelaverty:
Hello,

When I am trying to capture the url of my asp.net core application using JsonLayout the url of the site is being escaped.  

**NLog version**: 4.5.11

**NLog.Web.AspNetCore version**: 4.8.0

**Platform**: .NET Core 2.2

**Current NLog config** (xml or C#, if relevant)

```xml
<?xml version="1.0" encoding="utf-8" ?>
<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd"
      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
      autoReload="true"
      internalLogLevel="Info"
      internalLogFile="c:\temp\internal-nlog.txt">

  <!-- enable asp.net core layout renderers -->
  <extensions>
    <add assembly="NLog.Web.AspNetCore"/>
  </extensions>

  <!-- the targets to write to -->
  <targets async="true">
    <!-- Write logs as Json into a file -->
    <target name="json-file" 
      xsi:type="File" 
      fileName="${basedir}/application-logs.json"
      archiveAboveSize="50000000" 
      archiveEvery="Day" 
      maxArchiveFiles="9"
      concurrentWrites="False">
      <layout xsi:type="JsonLayout">
        <attribute name="date" layout="${longdate}" />
        <attribute name="level" layout="${level:upperCase=true}"/>
        <attribute name="logger" layout="${logger}"/>
        <attribute name="identity" layout="${aspnet-user-identity}"/>
        <attribute name="url" layout="${aspnet-request-url:IncludePort=true:IncludeQueryString=true}" encode="true"/>
        <attribute name="host" layout="${aspnet-Request-Host}"/>
        <attribute name="userAgent" layout="${aspnet-Request-UserAgent}"/>
        <attribute name="referrer" layout="${aspnet-Request-Referrer}"/>
        <attribute name="action" layout="${aspnet-mvc-action}"/>
        <attribute name="clientIp" layout="${aspnet-request-ip}"/>
        <attribute name="message" layout="${message}" />
        <attribute name="exception" layout="${onexception:|${exception:format=ToString}}" />
      </layout>
    </target>

  </targets>
  <rules>
    <!--Skip non-critical Microsoft logs and so log only own logs-->
    <logger name="Microsoft.*" maxlevel="Info" final="true" /> <!-- BlackHole without writeTo -->
    
[...]

PR #3586, opened 2019-09-17
JSON layout and ${json-encode}, added EscapeForwardSlash (default false). 

304NotModified:
fixes https://github.com/NLog/NLog/issues/3580

### DIFF
--- a/src/NLog/Targets/DefaultJsonSerializer.cs
+++ b/src/NLog/Targets/DefaultJsonSerializer.cs
@@ -816,7 +816,5 @@ namespace NLog.Targets
                 return false;
             }
         }
-
-
     }
 }
--- a/src/NLog/Targets/DefaultJsonSerializer.cs
+++ b/src/NLog/Targets/DefaultJsonSerializer.cs
@@ -104,7 +104,7 @@ namespace NLog.Targets
                     {
                         StringBuilder sb = new StringBuilder(str.Length + 4);
                         sb.Append('"');
-                        AppendStringEscape(sb, str, options.EscapeUnicode);
+                        AppendStringEscape(sb, str, options.EscapeUnicode, options.EscapeForwardSlash);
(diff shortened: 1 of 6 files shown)


snakefoot:
I guess in the premature optimization world, then it could be nice to skip the StringBuilder allocation when handling strings with http-address.

Maybe also fix RequiresJsonEscape:

https://github.com/NLog/NLog/blob/d7b33fddd5087afbfe71ccc1a9eb60b69d0767df/src/NLog/Targets/DefaultJsonSerializer.cs#L673-L682

src/NLog/Targets/DefaultJsonSerializer.cs
```diff
@@ -692,14 +691,19 @@ internal static void AppendStringEscape(StringBuilder destination, string text,
...
+
+        internal static bool RequiresJsonEscape(char ch, bool escapeUnicode, bool escapeForwardSlash)
         {
             if (!EscapeChar(ch, escapeUnicode))
             {
                 switch (ch)
                 {
                     case '"':
-                    case '\\':
+                    case '\\': return escapeForwardSlash;
```
snakefoot: Looks like you chose the backward-slash instead of the forward-slash?

snakefoot: Maybe something like this:
```c#
               switch (ch)
                {
                    case '/': return escapeForwardSlash;
                    case '"':
                    case '\\':
                        return true;
```

304NotModified: oops, thanks!

304NotModified:
note: default is `true` (backwards comp)

Fixed docs in https://github.com/NLog/NLog/wiki/JsonLayout
and https://github.com/NLog/NLog/wiki/JSON-Encode-Layout-Renderer

Issue #3591, opened 2019-09-20
Does OrdinalIgnoreCase works with (hash)Set if configured from nlog.config?


304NotModified:
from https://github.com/NLog/NLog.Web/pull/472

does the OrdinalIgnoreCase work when Exclude is set from nlog.config?

```
public class MyTarget {
    public ISet<string> Exclude { get; set; } = new HashSet<string>(new[] { "AUTH", "SESS_ID" }, StringComparer.OrdinalIgnoreCase);
}
```

If a new (hash)set is created, we lose the `OrdinalIgnoreCase`. So we should not create a new set if it isn't needed.

snakefoot:
The test failed when I tried. See also #3593

304NotModified:
fixed by https://github.com/NLog/NLog/pull/3593

PR #3593, opened 2019-09-21
AllEventPropertiesLayoutRenderer - Added support for Exclude with OrdinalIgnoreCase

snakefoot:
See also #3591

Note merge to master.

### DIFF
--- a/src/NLog/Internal/PropertyHelper.cs
+++ b/src/NLog/Internal/PropertyHelper.cs
@@ -366,17 +366,14 @@ namespace NLog.Internal
 
                 //note: type.GenericTypeArguments is .NET 4.5+ 
                 collectionItemType = collectionType.GetGenericArguments()[0];
-                var listType = isSet ? typeof(HashSet<>) : typeof(List<>);
-                var genericArgs = collectionItemType;
-                var concreteType = listType.MakeGenericType(genericArgs);
-                collectionObject = hashsetComparer != null ? Activator.CreateInstance(concreteType, hashsetComparer) : Activator.CreateInstance(concreteType);
+                collectionObject = CreateCollectionObjectInstance(isSet ? typeof(HashSet<>) : typeof(List<>), collectionItemType, hashsetComparer);
                 //no support for array
                 if (collectionObject == null)
                 {
                     throw new NLogConfigurationException("Cannot create instance of {0} for value {1}", collectionType.ToString(), valueRaw);
                 }
 
-                collectionAddMethod = concreteType.GetMethod("Add", BindingFlags.Instance | BindingFlags.Public);
+                collectionAddMethod = collectionObject.GetType().GetMethod("Add", BindingFlags.Instance | BindingFlags.Public);
                 if (collectionAddMethod == null)
                 {
                     throw new NLogConfigurationException("Add method on type {0} for value {1} not found", collectionType.ToString(), valueRaw);
@@ -389,6 +386,18 @@ namespace NLog.Internal
             collectionAddMethod = null;
             collectionItemType = null;
             return false;
+        }
+
+        private static object CreateCollectionObjectInstance(Type collectionType, Type collectionItemType, object hashSetComparer)
+        {
+            var concreteType = collectionType.MakeGenericType(collectionItemType);
+            if (hashSetComparer != null)
+            {
+                var constructor = concreteType.GetConstructor(new[] { hashSetComparer.GetType() });
+                if (constructor != null)
+                    return constructor.Invoke(new[] { hashSetComparer });
+            }
+            return Activator.CreateInstance(concreteType);
         }
 
         /// <summary>
--- a/src/NLog/Internal/PropertyHelper.cs
+++ b/src/NLog/Internal/PropertyHelper.cs
@@ -362,14 +362,14 @@ namespace NLog.Internal
             //not checking "implements" interface as we are creating HashSet<T> or List<T> and also those checks are expensive
             if (isSet || typeDefinition == typeof(List<>) || typeDefinition == typeof(IList<>) || typeDefinition == typeof(IEnumerable<>)) //set or list/array etc
             {
-                object[] constructorArgs = isSet ? CreateHashSetConstructorArgs(obj, propInfo) : ArrayHelper.Empty<object>();
+                object hashsetComparer = isSet ? ExtractHashSetComparer(obj, propInfo) : null;
 
                 //note: type.GenericTypeArguments is .NET 4.5+ 
                 collectionItemType = collectionType.GetGenericArguments()[0];
                 var listType = isSet ? typeof(HashSet<>) : typeof(List<>);
                 var genericArgs = collectionItemType;
                 var concreteType = listType.MakeGenericType(genericArgs);
-                collectionObject = constructorArgs?.Length > 0 ? Activator.CreateInstance(concreteType, constructorArgs) : Activator.CreateInstance(concreteType);
+                collectionObject = hashsetComparer != null ? Activator.CreateInstance(concreteType, hashsetComparer) : Activator.CreateInstance(concreteType);
                 //no support for array
                 if (collectionObject == null)
                 {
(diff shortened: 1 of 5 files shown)


304NotModified:
Cool! But won't it be easier to not creating a new instance? (So use clear and add). Then the original collection type is also preserved

snakefoot:
> Cool! But won't it be easier to not creating a new instance? (So use clear and add). Then the original collection type is also preserved

Then you need more checking if the original collection is readonly. And it will be very different from the current default behavior (always creating new collection and calling setter).

src/NLog/Internal/PropertyHelper.cs
```diff
@@ -393,6 +391,27 @@ private static bool TryCreateCollectionObject(Type collectionType, string valueR
             return false;
         }
 
+        /// <summary>
+        /// Attempt to reuse the HashSet.Comparer from the original HashSet-object (Ex. StringComparer.OrdinalIgnoreCase)
+        /// </summary>
+        private static object[] CreateHashSetConstructorArgs(object obj, PropertyInfo propInfo)
```
304NotModified: Proposal, rename to FindComparer as the array is never longer than 1

(And adjust body ;)

src/NLog/LayoutRenderers/AllEventPropertiesLayoutRenderer.cs
```diff
@@ -80,18 +84,50 @@ public AllEventPropertiesLayoutRenderer()
...
         /// </summary>
         /// <docgen category='Rendering Options' order='10' />
+        [Obsolete("Instead use the Exclude-property. Marked obsolete on NLog 4.6.8")]
         [DefaultValue(false)]
-        public bool IncludeCallerInformation { get; set; }
-
+        public bool IncludeCallerInformation
+        {
+            get => Exclude?.Contains(CallerInformationAttributeNames[0]) != true;
+            set
```
304NotModified: I think it's better to keep the separate boolean property/field and something like `!Exclude.Contains(..) && !IsUnwantedCallerPropery(..)`   . That makes things a lot easier and also better to remove in the future

src/NLog/Internal/PropertyHelper.cs
```diff
@@ -365,13 +362,14 @@ private static bool TryCreateCollectionObject(Type collectionType, string valueR
...
+                object[] constructorArgs = isSet ? CreateHashSetConstructorArgs(obj, propInfo) : ArrayHelper.Empty<object>();
+
                 //note: type.GenericTypeArguments is .NET 4.5+ 
                 collectionItemType = collectionType.GetGenericArguments()[0];
-
                 var listType = isSet ? typeof(HashSet<>) : typeof(List<>);
                 var genericArgs = collectionItemType;
                 var concreteType = listType.MakeGenericType(genericArgs);
-                collectionObject = Activator.CreateInstance(concreteType);
+                collectionObject = constructorArgs?.Length > 0 ? Activator.CreateInstance(concreteType, constructorArgs) : Activator.CreateInstance(concreteType);
```
304NotModified: Is this safe for all types? I think we should search for the proper ctor with reflection (and fallback to default ctor)

304NotModified: With this?

https://docs.microsoft.com/en-us/dotnet/api/system.type.getconstructor?view=netframework-4.8#System_Type_GetConstructor_System_Type___

Another option is a try-catch

src/NLog/Internal/PropertyHelper.cs
```diff
@@ -393,6 +391,27 @@ private static bool TryCreateCollectionObject(Type collectionType, string valueR
...
+        private static object[] CreateHashSetConstructorArgs(object obj, PropertyInfo propInfo)
+        {
+            var exitingValue = propInfo.CanRead && propInfo.GetIndexParameters().Length == 0 && propInfo.GetGetMethod() != null ? propInfo.GetGetMethod().Invoke(obj, null) : null;
+            if (exitingValue != null)
+            {
+                // Found original HashSet-object. See if we can extract the Comparer
+                var comparerPropInfo = exitingValue.GetType().GetProperty("Comparer", BindingFlags.Instance | BindingFlags.Public);
+                if (comparerPropInfo != null && comparerPropInfo.CanRead && comparerPropInfo.GetGetMethod() != null)
+                {
+                    var comparer = comparerPropInfo.GetGetMethod().Invoke(exitingValue, null);
```
304NotModified: Why not `comparerPropInfo.GetValue(existing Value)`? 

(Also above like this)

src/NLog/LayoutRenderers/AllEventPropertiesLayoutRenderer.cs
snakefoot: I want to remove IncludeCallerInformation and remove CallerAttributes-properties in NLog 5. See no reason to introduce new stuff to remove right after.

src/NLog/Internal/PropertyHelper.cs
snakefoot: The code for handling generic stuff is annoying, and this works.

src/NLog/Internal/PropertyHelper.cs
snakefoot: Fixed

src/NLog/Internal/PropertyHelper.cs
snakefoot: Fixed

src/NLog/Internal/PropertyHelper.cs
snakefoot: But please push your changes.

snakefoot: Now calling GetConstructor pretty sure the code path is the same as Activator.CreateInstance, but now a little more convoluted :)

src/NLog/Internal/ReflectionHelpers.cs
```diff
@@ -272,6 +272,20 @@ public static Assembly GetAssembly(this Type type)
             return typeInfo.Assembly;
 #else
             return type.Assembly;
+#endif
+        }
+
+        public static bool IsValidPublicProperty(this PropertyInfo p)
```
304NotModified: I like helpers 👍

src/NLog/Internal/ReflectionHelpers.cs
```diff
@@ -272,6 +272,20 @@ public static Assembly GetAssembly(this Type type)
...
+        {
+            return p.CanRead && p.GetIndexParameters().Length == 0 && p.GetGetMethod() != null;
+        }
+
+        public static object GetPropertyValue(this PropertyInfo p, object instance)
+        {
+#if NET45
```
304NotModified: this is a bit unexpected for me:

image

https://docs.microsoft.com/en-us/dotnet/api/system.reflection.propertyinfo.getvalue?view=netframework-4.8#System_Reflection_PropertyInfo_GetValue_System_Object_

Maybe it should be `!SILVERLIGHT` etc?

src/NLog/LayoutRenderers/AllEventPropertiesLayoutRenderer.cs
304NotModified: IMO this is introducing new stuff, but I see why it's introduced so let's keep it like this.

src/NLog/LayoutRenderers/AllEventPropertiesLayoutRenderer.cs
```diff
@@ -137,23 +173,26 @@ protected override void Append(StringBuilder builder, LogEventInfo logEvent)
...
-                {
-                    properties = properties.Where(p => !IsEmptyPropertyValue(p.Value));
-                }
-                foreach (var property in properties)
+                foreach (var property in logEvent.Properties)
                 {
+                    if (!IncludeEmptyValues && IsEmptyPropertyValue(property.Value))
+                        continue;
+
+                    if (checkForExclude && Exclude.Contains(property.Key as string))
```
304NotModified: Proposal: 

```
  if (checkForExclude && property.Key is string key && Exclude.Contains(key))
```
saves as lookup when ` property.Key` is `null`

src/NLog/Internal/ReflectionHelpers.cs
snakefoot: See https://docs.microsoft.com/en-us/dotnet/api/system.reflection.propertyinfo.getvalue?view=netframework-4.0#System_Reflection_PropertyInfo_GetValue_System_Object_

304NotModified: OK, missed that. Thanks!

src/NLog/Internal/PropertyHelper.cs
304NotModified: 👍

src/NLog/LayoutRenderers/AllEventPropertiesLayoutRenderer.cs
304NotModified: Let me know what you think, i'm also fine with keeping it like this. Then I will merge it.

snakefoot: Never thought you would suggest premature optimization, and increase code complexity. I can never say no to an optimization opportunity, even thought I think it will only help 1 out of 10 cases :)

304NotModified: I was trigger by the `null` to Contains, which crashes on dictionaries (well on ContainsKey), and not really by the optimization ;)

snakefoot: Optimization has been committed :)

304NotModified: 🎉

snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/All-Event-Properties-Layout-Renderer

Issue #3588, opened 2019-09-19
Is there a mechanism to filter exception serialization?


mmurrell:
### Summary
We recently encountered a scenario in NLog where we logged what happened to be an Entity Framework exception, and we did so  with full serialization.  The Entity Framework exception included references to the database entity in question,  and while serializing the object graph of the exception NLog began to evaluate every property which caused a 'Lazy Load' on all the related tables, their related tables, and their related tables.  This particular event was responsible for an application outage.

The particular exception that we attempted to log was a `UpdateException` ([link](https://docs.microsoft.com/en-us/dotnet/api/system.data.entity.core.updateexception?view=entity-framework-6.2.0)).  The serialization problem begins when serializing `exception.StateEntries[0].Entity`.  

As a consumer, it would be nice to blacklist certain types using a lambda, or another mechanism, or give the consumer the ability to subclass the serializer.  The classes would be trivial to exclude by lambda, because the class names all follow the pattern `System.Data.Entity.DynamicProxies.TableName_E2DC4A5B7B1B73EDD543E8CD945EC967B53FA31EA65D6F4D4F23EC13451D87FD`

### Details
**Version:** NLog 4.6.3 via Nuget
**Configuration:**
```
<nlog>
  <targets>
    <target name="console" type="Console" layout="${sequenceid}|${machinename}|${longdate}|${logger}|${level}|${message}|${exception:format=@}"/>
  </targets>
  <rules>
    <logger name="*" minlevel="Info" writeTo="console"/>
  </rules>
</nlog>
```

**Sample Code:**
```
try
{
    // Database configured to have multiple tables, lazy loading enabled, TableName referencing many child tables.
    var db = IoC.Container.GetInstance<EntityFrameworkDataContext>()
    var databaseObject = db.TableName.First();
    databaseObject.NonNullableField = null;
    db.SubmitChanges();
}
catch (Exception ex)
{
    // On execution, run Sql Server Profiler and view all tables in the database being scanned.
    Logger.Error(ex, "Error");
}
```
**Internal Log (Debug):** Available if you need it, but it doesn't show anything relevant
**Current Result:** When `Logger.Error(exception, "...")` is called with a Entity Framework exception, the entire object graph is traversed by the NLog Serializer, causing repeated table scans against the entire database.
**Expected Result:** A mechanism to blacklist certain types from being serialized.
**Workarounds:** 
- Remove exception from the error log: `Logger.Error("Something happened." + ex.Message);`,
[...]

Issue #3264, opened 2019-04-01
Serilog Destructure.ByTransforming equivalent?


greenmooseSE:
Hi,

Is there a way to transform the objects used in structured logging with NLog?
Similar to Serilog's
```csharp
            Log.Logger = new LoggerConfiguration()
                .Destructure.ByTransforming<ColorTransformedDto>(x => new {R = x.Red, G = x.Green})
```

Thanks.

snakefoot:
Not at the moment. But I guess it should not be that difficult to add an interface for injecting a custom destructure-method into the `ObjectReflectionCache`:

https://github.com/NLog/NLog/blob/dev/src/NLog/Internal/ObjectReflectionCache.cs

The cache is currently only being used by `JsonLayout` and `XmlLayout`. Could you tell what NLog target you are planning to use?

304NotModified:
I think this is for the message, so this is the `ValueFormatter`?

(If so, then you need to set ConfigurationItemFactory.ValueFormatter with your custom [IValueFormatter](https://nlog-project.org/documentation/v4.6.0/html/T_NLog_IValueFormatter.htm)

snakefoot:
@304NotModified After reading this link:

https://github.com/serilog/serilog/wiki/Structured-Data#customizing-the-stored-data

Then Serilog destructure sounds very much like NLog Precalculate. So you capture some properties from a volatile object and use those for the LogEvent. Instead of holding a reference to the volatile object that might become bad later.

I guess such a transformation-logic  would have to live on the logger, and replace the LogEvent-parameters on the fly with the transformed object.

snakefoot:
@greenmooseSE Special extraction of values from the LogEvent-parameters are usually done with laytoutrenderers. You can create your own:

https://github.com/NLog/NLog/wiki/How-to-write-a-custom-layout-renderer

greenmooseSE:
> Could you tell what NLog target you are planning to use?

I was just doing some comparison of the logging frameworks with regards to structural logging when noticing this feature in SeriLog (via [pluralsight](https://www.pluralsight.com/courses/modern-structured-logging-serilog-seq)).

I was just curious if NLog has similar "easy" option since the feature looked interesting since it is a one-liner to the configuration without the need of creating additional sinks/formatters/rendererse etc.

So my question has been answered so you may close this if you want, thanks :).

snakefoot:
NLog have now also added support for objectpath in `${event-properties}`:

[...]

PR #3606, opened 2019-09-28
(WIP) Adding exclusion list for object serialization.

mmurrell:
This is an PR with an attempt to implement the suggestion raised in #3588.  

Based on the suggestion in the Issue, I implemented a mechanism in the `ObjectReflectionCache` for excluding types based on a type argument, a generic type, or a function to evaluate based on other mechanisms (Namespace, Assembly, etc)

To provide an initial solution, I just added some static methods at the end of the `JsonSerializationOptions` class.  I don't think they belong there, but I could not find a better class and I assumed that making the `ObjectReflectionCache` public would be a bad idea.  If there's no other options, a singleton class might be a long term option.  A singleton with some testing seams may also help address 'To Do' item no. 2 below.

## To Do
- [x] Find permanent home for the configuration methods.
- [x] There is the potential of some concurrency issues in the unit tests due to the static configuration.
- [ ] Address any changes raised in the pull request.

### DIFF
--- a/src/NLog/Targets/JsonSerializeOptions.cs
+++ b/src/NLog/Targets/JsonSerializeOptions.cs
@@ -33,6 +33,8 @@
 
 using System;
 using System.ComponentModel;
+using System.Reflection;
+using NLog.Targets.SerializationInterceptors;
 
 namespace NLog.Targets
 {
@@ -88,16 +90,25 @@ namespace NLog.Targets
         {
             QuoteKeys = true;
             MaxRecursionLimit = 10;
+
+            SerializeAsToString<Assembly>();
+            SerializeAsToString<MemberInfo>();
+            SerializeAsToString<Uri>();
+            SerializeAsToString<IFormattable>();
         }
 
-        /// <summary>Registers a type<see cref="T:System.Type" /> to be excluded from serialization.</summary>
-        public static void RegisterSerializationExclusion(Type t) => NLog.Internal.ObjectReflectionCache.SerializationExclusionList.Add(x => x == t);
-        /// <summary>Registers a type<see cref="T:System.Type" /> to be excluded from serialization.</summary>
-        public static void RegisterSerializationExclusion<T>() => NLog.Internal.ObjectReflectionCache.SerializationExclusionList.Add(x => x == typeof(T));
-        /// <summary>Registers a custom evaluation function to determine if a given type should be serialized.  If the func returns true, the object's ToString() will be invoked.  If false, the object will be serialized.</summary>
-        public static void RegisterSerializationExclusion(Func<Type, bool> func) => NLog.Internal.ObjectReflectionCache.SerializationExclusionList.Add(func);
-        /// <summary>Resets the list for custom serialization.</summary>
-        public static void ResetSerializationExclusions() => NLog.Internal.ObjectReflectionCache.SerializationExclusionList.Clear();
+        /// <summary>
+        /// The interceptor to be called during the serialization process.
+        /// </summary>
+        internal ISerializationInterceptor SerializationInterceptor { get; private set; } = null;
 
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        public void SerializeAsToString<T>()
+        {
+            SerializationInterceptor = new ToStringSerializationInterceptor(SerializationInterceptor, typeof(T));
+        }
     }
 }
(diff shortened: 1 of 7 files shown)


src/NLog/Targets/JsonSerializeOptions.cs
```diff
@@ -89,5 +89,15 @@ public JsonSerializeOptions()
             QuoteKeys = true;
             MaxRecursionLimit = 10;
         }
+
+        /// <summary>Registers a type<see cref="T:System.Type" /> to be excluded from serialization.</summary>
```
mmurrell: I don't believe these methods belong in this class, but I couldn't find a better class. Any suggestions?

304NotModified:
Thanks! Will look at this indepth soon!

304NotModified:
PS. Related: https://github.com/NLog/NLog/issues/3264

snakefoot:
I would probably create an interface that can return true/false for a type.

Then have a default implementation of this interface that matches the current logic in `ObjectReflectionCache`

Then allow one to assign a new custom implementation of the interface. Maybe some helper methods that captures the original interface-implementation and chains it together with the new restriction.

Making the following boilerplate code pretty:

```c#
class SkipObjectReflection : IRestrictObjectReflection
{
Type _newType;
IRestrictObjectReflection _original;

public SkipObjectReflection(IRestrictObjectReflection original, Type newType)
{
    _original = original;
    _newType = newType;
}

public bool SkipReflection(Type type)
{
    if (_newType.IsAssignableFrom(type))
          return true;

   return _original.SkipReflection(type);
}
}

var original = NLog.Config.ConfigurationItemFactory.Default.RestrictObjectReflection;
NLog.Config.ConfigurationItemFactory.Default.RestrictObjectReflection = new SkipObjectReflection(original, skipNewType);
```

snakefoot:
Could also make a more advanced version of the interface, that returns an array like this:

```c#
interface IObjectTypeReflection
{
    IList<KeyValuePari<string,MemberInfo>> GetProperties(Type type);
}
```

Then make a contract:
- Returning null then it performs obejct.ToString(). 
- If it returns single item without any Key-name, then it will use the MemberInfo as custom object.ToString().

Then it would resolve both  #3519 and #3588

mmurrell:
So, I've spent a bit of time thinking about this and working through this in code, and I'm running into some stumbling blocks.  

It looks like the advanced interface should signal one of three serialization options:
- ToString
- Custom list (subset) of properties
- Custom serialization.

Regarding your proposal for custom serialization, I'm curious about your limitation for a single `MemberInfo`.  It would limit the customizations to a single (and direct accessed) member, which wouldn't quite address some of the concerns with #3519 (specifically the request for invoking `GetResponseStream` for `WebExceptions`).  It may be advantageous here to consider a Func<T, object> which could then be registered as a custom serializer.

```
config.RegisterCustomSerializer<Assembly>(); // Would register as 'ToString()'
config.RegisterCustomSerializer<CustomException>(new[] { "Property1", "Property2", "Property4" ); //Would register property list filter.
config.RegisterCustomSerializer<Win32Exception>(ex=> $"NativeErrorCode: {ex.NativeErrorCode}") //Would register custom func

//Also, this may be overkill, but the same pattern would allow a consumer to craft their own object:
config.RegisterCustomSerializer<CustomObject>(o=>new { o.Property1, Renamed = o.Child.GetCustomValue() });
```

Again, I'm not sure how this will conflict with some of the object path conversation on #3520. 

Thoughts?

snakefoot:
@mmurrell You are right. The interface should probably be like this:

```c#
interface IObjectTypeReflection
{
    IList<KeyValuePair<string,object>> GetProperties(object obj);
}
```

But that would require some heavy refactoring of `ObjectReflectionCache` to avoid killing performance.

mmurrell:
I know this is a bit of a different direction, but these commits could accomplish the goal without impacting the `ObjectReflectionCache`.  I did follow your advice with the object-chaining implementation.

In this case, the serialization is intercepted before it gets to the `ObjectReflectionCache`, allowing the object to be augmented or replaced before it even gets there.

If this different direction is acceptable, I would like to build out some at least a `Func<T, object>` interceptor, and maybe some others as need dictates.

snakefoot:
@mmurrell Would like to have this filtering to work also for XmlLayout, and other locations where ObjectReflectionCache is being used.

Have created an implementation of my thoughts here: #3610 (Stolen your unit-test to show how it works)

snakefoot:
Superseeded by #3610

PR #3609, opened 2019-10-01
FileTarget - Improve handling of relative path in ArchiveFileName

### LINKED ISSUE #3608: File archiving fails when using the same name
**NLog version**: 4.6.3.9764
**Platform**: .Net 4.5

I have NLog configured with file targets + archiving.
My configuration is so that the current file also includes the date, i.e. it has the same name as its future archive. -> This is a requirement because we don't want the file to change when archiving.

In principle it should work as this piece of code checks if the archive target is the same as current log, which results in a no-op:
https://github.com/NLog/NLog/blob/dev/src/NLog/Targets/FileTarget.cs#L1259

Now the bug seems to be that this comparaison uses non-normalized paths.
In my config I have a relative path `..\logs\fabrikam.txt`. 

When looking at NLog internal log, this results in this comparison:
```
Archiving C:\Fabrikam\logs\fabrikam-2019-10.txt to C:\Fabrikam\IIS\..\logs\fabrikam-2019-10.txt
```
As you can see they resolve to the same file, but the strings are different.
So the code continue into `ArchiveFileAppendExisting`, here:
`https://github.com/NLog/NLog/blob/dev/src/NLog/Targets/FileTarget.cs#L1274`
Which fails badly, probably because it tries to append to itself.

I think the paths need to be normalized before they are compared.

We run with `throwExceptions` so this prevents our app from starting.

For reference, here's the internal log:
```
2019-10-01 13:40:56.9385 Debug FileTarget(Name=logfile): Process file 'C:\Fabrikam\IIS\..\logs\Fabrikam.log_2019-10.txt' on startup
2019-10-01 13:40:56.9541 Info FileTarget(Name=logfile): Archive mutex not available: C:\Fabrikam\IIS\..\logs\Fabrikam.log_2019-10.txt
2019-10-01 13:40:56.9541 Debug FileTarget(Name=logfile): Do archive file: 'C:\Fabrikam\IIS\..\logs\Fabrikam.log_2019-10.txt'
2019-10-01 13:40:56.9697 Info FileTarget(Name=logfile): Archiving C:\Fabrikam\logs\Fabrikam.log_2019-10.txt to C:\Fabrikam\IIS\..\logs\Fabrikam.log_2019-10.txt
2019-10-01 13:40:56.9697 Info FileTarget(Name=logfile): Already exists, append to C:\Fabrikam\IIS\..\logs\Fabrikam.log_2019-10.txt
2019-10-01 13:40:56.9697 Warn FileTarget(Name=logfile): Failed to archive file 'C:\Fabrikam\IIS\..\logs\Fabrikam.log_2019-10.txt'. Exception: System.IO.IOException: The process cannot access the file 'C:\Fabrikam\logs\Fabrikam.log_2019-10.txt' because it is being used by another process.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share)
   at NLog.Targets.FileTarget.ArchiveFileAppendExisting(String fileName, String archiveFileName)
   at NLog.Targets.FileTarget.ArchiveFile(String fileName, String archiveFileName)
   at NLog.Targets.FileTarget.DoAutoArchive(String fileName, LogEventInfo eventInfo, DateTime previousLogEventTimestamp, Boolean initializedNewFile)
   at NLog.Targets.FileTarget.TryArchiveFile(String fileName, LogEventInfo ev, Int32 upcomingWriteSize, DateTime previousLogEventTimestamp, Boolean initializedNewFile)
```

snakefoot:
Trying to resolve #3608

Note merge to master-branch.

### DIFF
--- a/src/NLog/Targets/FileArchiveModes/FileArchiveModeDate.cs
+++ b/src/NLog/Targets/FileArchiveModes/FileArchiveModeDate.cs
@@ -86,6 +86,7 @@ namespace NLog.Targets.FileArchiveModes
             FileNameTemplate archiveFileNameTemplate = GenerateFileNameTemplate(archiveFilePath);
             string dirName = Path.GetDirectoryName(archiveFilePath);
             archiveFilePath = Path.Combine(dirName, archiveFileNameTemplate.ReplacePattern("*").Replace("*", archiveDate.ToString(_archiveDateFormat)));
+            archiveFilePath = Path.GetFullPath(archiveFilePath);    // Rebuild to fix non-standard path-format
             return new DateAndSequenceArchive(archiveFilePath, archiveDate, _archiveDateFormat, 0);
         }
     }
--- a/src/NLog/Targets/FileArchiveModes/FileArchiveModeDateAndSequence.cs
+++ b/src/NLog/Targets/FileArchiveModes/FileArchiveModeDateAndSequence.cs
@@ -91,6 +91,7 @@ namespace NLog.Targets.FileArchiveModes
                 $"{archiveDate.ToString(_archiveDateFormat)}.{paddedSequence}");
             string dirName = Path.GetDirectoryName(archiveFilePath);
             archiveFilePath = Path.Combine(dirName, archiveFileNameWithoutPath);
+            archiveFilePath = Path.GetFullPath(archiveFilePath);    // Rebuild to fix non-standard path-format
             return new DateAndSequenceArchive(archiveFilePath, archiveDate, _archiveDateFormat, nextSequenceNumber);
         }
 
--- a/src/NLog/Targets/FileArchiveModes/FileArchiveModeSequence.cs
+++ b/src/NLog/Targets/FileArchiveModes/FileArchiveModeSequence.cs
@@ -88,6 +88,7 @@ namespace NLog.Targets.FileArchiveModes
             string paddedSequence = nextSequenceNumber.ToString().PadLeft(minSequenceLength, '0');
             string dirName = Path.GetDirectoryName(archiveFilePath);
             archiveFilePath = Path.Combine(dirName, archiveFileNameTemplate.ReplacePattern("*").Replace("*", paddedSequence));
+            archiveFilePath = Path.GetFullPath(archiveFilePath);    // Rebuild to fix non-standard path-format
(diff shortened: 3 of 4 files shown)


src/NLog/Targets/FileArchiveModes/FileArchiveModeDate.cs
```diff
@@ -86,6 +86,7 @@ public override DateAndSequenceArchive GenerateArchiveFileName(string archiveFil
             FileNameTemplate archiveFileNameTemplate = GenerateFileNameTemplate(archiveFilePath);
             string dirName = Path.GetDirectoryName(archiveFilePath);
             archiveFilePath = Path.Combine(dirName, archiveFileNameTemplate.ReplacePattern("*").Replace("*", archiveDate.ToString(_archiveDateFormat)));
+            archiveFilePath = Path.GetFullPath(archiveFilePath);    // Rebuild to fix non-standard path-format
```
304NotModified: I'm afraid this could be slow. AFAIK GetFullPath is a really slow method.

That's one of the reasons FilePathLayout was build. See https://github.com/NLog/NLog/blob/0bfdad5213e0dccc6fc3ef2b192d5446411953b3/src/NLog/Internal/Files/FilePathLayout.cs#L233-L234

snakefoot: It only happens on archive operations. Archive-operations are by default slow. As correctly observed by #3608 then `Path.GetFullPath` happens for some archive-modes, but not all. This PR aligns them all.

snakefoot: The method you are referring to is called for every LogEvent. These methods are callled for every archive-operation, which is a different number.

304NotModified: OK 👍 thanks for checking/explaining!

PR #3610, opened 2019-10-01
ObjectReflectionCache extended to allow override of object reflection

snakefoot:
Alternative solution for #3588

Introduces the new interface:

```c#
    public interface IObjectTypeTransformer
    {
        /// <summary>
        /// Takes a dangerous (or massive) object and converts into a safe (or reduced) object
        /// </summary>
        /// <returns>
        /// Null if unknown object, or object cannot be handled
        /// </returns>
        object TryTransformObject(object obj);
    }
```

Then one can do this: (NLog 4.7+)

```c#
NLog.LogManager.Setup().SetupSerialization(s => 
   s.RegisterObjectTransformation<IExcludedInterface>(o => o.ToString()));
```

Note merge to master-branch.

### DIFF
--- a/src/NLog/Config/ConfigurationItemFactory.cs
+++ b/src/NLog/Config/ConfigurationItemFactory.cs
@@ -31,15 +31,14 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
-using System.IO;
-using System.Linq;
-using NLog.Common;
-
 namespace NLog.Config
 {
     using System;
     using System.Collections.Generic;
+    using System.IO;
+    using System.Linq;
     using System.Reflection;
+    using NLog.Common;
     using NLog.Conditions;
     using NLog.Filters;
     using NLog.Internal;
@@ -67,6 +66,7 @@ namespace NLog.Config
         private readonly Factory<TimeSource, TimeSourceAttribute> _timeSources;
 
         private IJsonConverter _jsonSerializer = DefaultJsonSerializer.Instance;
+        private IObjectTypeTransformer _objectTypeTransformer = ObjectReflectionCache.Instance;
 
         /// <summary>
         /// Called before the assembly will be loaded.
@@ -191,6 +191,15 @@ namespace NLog.Config
         {
             get => MessageTemplates.ValueFormatter.Instance;
             set => MessageTemplates.ValueFormatter.Instance = value;
+        }
+
+        /// <summary>
+        /// Gets or sets the custom object-type transformation for use in <see cref="JsonLayout"/>, <see cref="XmlLayout"/> or <see cref="NLog.LayoutRenderers.Wrappers.ObjectPathRendererWrapper"/>
+        /// </summary>
+        internal IObjectTypeTransformer ObjectTypeTransformer
+        {
+            get => _objectTypeTransformer;
+            set => _objectTypeTransformer = value ?? ObjectReflectionCache.Instance;
         }
 
         /// <summary>
--- a/src/NLog/Internal/ObjectReflectionCache.cs
+++ b/src/NLog/Internal/ObjectReflectionCache.cs
@@ -42,21 +42,48 @@ namespace NLog.Internal
     using System.Linq;
     using System.Reflection;
     using NLog.Common;
-    using NLog.Internal;
+    using NLog.Config;
 
     /// <summary>
     /// Converts object into a List of property-names and -values using reflection
     /// </summary>
-    internal class ObjectReflectionCache
+    internal class ObjectReflectionCache : IObjectTypeTransformer
     {
         readonly MruCache<Type, ObjectPropertyInfos> _objectTypeCache = new MruCache<Type, ObjectPropertyInfos>(10000);
-        static readonly Dictionary<Type, Func<object, IDictionary<string,object>>> _objectTypeOverride = new Dictionary<Type, Func<object, IDictionary<string, object>>>();
+        private IObjectTypeTransformer ObjectTypeTransformation => _objectTypeTransformation ?? (_objectTypeTransformation = ConfigurationItemFactory.Default.ObjectTypeTransformer);
+        private IObjectTypeTransformer _objectTypeTransformation;
+
+        public static IObjectTypeTransformer Instance { get; } = new ObjectReflectionCache();
+
+        object IObjectTypeTransformer.TryTransformObject(object obj)
+        {
+            return null;
+        }
 
         public ObjectPropertyList LookupObjectProperties(object value)
         {
             if (TryLookupExpandoObject(value, out var propertyValues))
             {
                 return propertyValues;
+            }
+
+            if (!ReferenceEquals(ObjectTypeTransformation, Instance))
+            {
+                var result = ObjectTypeTransformation.TryTransformObject(value);
+                if (result != null)
+                {
+                    if (result is IConvertible)
+                    {
+                        return new ObjectPropertyList(result, ObjectPropertyInfos.SimpleToString.Properties, ObjectPropertyInfos.SimpleToString.FastLookup);
+                    }
+
+                    if (TryLookupExpandoObject(result, out propertyValues))
+                    {
+                        return propertyValues;
+                    }
+
+                    value = result;
+                }
             }
 
             var objectType = value.GetType();
@@ -93,28 +120,6 @@ namespace NLog.Internal
                 }
                 objectPropertyList = new ObjectPropertyList(value, propertyInfos.Properties, propertyInfos.FastLookup);
                 return true;
-            }
-
-            if (_objectTypeOverride.Count > 0)
-            {
-                // User defined object reflection override
-                lock (_objectTypeOverride)
-                {
-                    if (_objectTypeOverride.TryGetValue(objectType, out var objectReflection))
-                    {
-                        var objectProperties = objectReflection.Invoke(value);
-                        if (objectProperties?.Count > 0)
-                        {
-                            objectPropertyList = new ObjectPropertyList(objectProperties);
-                            return true;
-                        }
-
-                        // object.ToString() since no properties
-                        propertyInfos = ObjectPropertyInfos.SimpleToString;
-                        objectPropertyList = new ObjectPropertyList(value, propertyInfos.Properties, propertyInfos.FastLookup);
-                        return true;
-                    }
-                }
             }
 
             if (TryExtractExpandoObject(objectType, out propertyInfos))
--- a/src/NLog/SetupSerializationBuilderExtensions.cs
+++ b/src/NLog/SetupSerializationBuilderExtensions.cs
@@ -1,0 +1,139 @@
+﻿// 
+// Copyright (c) 2004-2019 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without 
+// modification, are permitted provided that the following conditions 
+// are met:
+// 
+// * Redistributions of source code must retain the above copyright notice, 
+//   this list of conditions and the following disclaimer. 
+// 
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution. 
+// 
+// * Neither the name of Jaroslaw Kowalski nor the names of its 
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission. 
+// 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE 
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE 
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE 
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR 
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS 
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN 
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) 
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF 
+// THE POSSIBILITY OF SUCH DAMAGE.
+// 
+
+namespace NLog
+{
+    using System;
+    using System.Reflection;
+    using NLog.Common;
+    using NLog.Config;
+
+    /// <summary>
+    /// Extension methods to setup NLog extensions, so they are known when loading NLog LoggingConfiguration
+    /// </summary>
+    public static class SetupSerializationBuilderExtensions
+    {
+        /// <summary>
+        /// Overrides the active <see cref="IJsonConverter"/> with a new custom implementation
+        /// </summary>
+        public static ISetupSerializationBuilder RegisterJsonConverter(this ISetupSerializationBuilder setupBuilder, IJsonConverter jsonConverter)
+        {
+            ConfigurationItemFactory.Default.JsonConverter = jsonConverter ?? NLog.Targets.DefaultJsonSerializer.Instance;
+            return setupBuilder;
+        }
+
+        /// <summary>
+        /// Overrides the active <see cref="IValueFormatter"/> with a new custom implementation
+        /// </summary>
+        public static ISetupSerializationBuilder RegisterValueFormatter(this ISetupSerializationBuilder setupBuilder, IValueFormatter valueFormatter)
+        {
+            ConfigurationItemFactory.Default.ValueFormatter = valueFormatter ?? NLog.MessageTemplates.ValueFormatter.Instance;
+            return setupBuilder;
+        }
+
+        /// <summary>
+        /// Registers object Type transformation from dangerous (massive) object to safe (reduced) object
+        /// </summary>
+        public static ISetupSerializationBuilder RegisterObjectTransformation<T>(this ISetupSerializationBuilder setupBuilder, Func<T, object> transformer)
+        {
+            if (transformer == null)
+                throw new ArgumentNullException(nameof(transformer));
+
+            var original = ConfigurationItemFactory.Default.ObjectTypeTransformer;
+            ConfigurationItemFactory.Default.ObjectTypeTransformer = new ObjectTypeTransformation<T>(transformer, original);
+            return setupBuilder;
+        }
+
+        /// <summary>
+        /// Registers object Type transformation from dangerous (massive) object to safe (reduced) object
+        /// </summary>
+        public static ISetupSerializationBuilder RegisterObjectTransformation(this ISetupSerializationBuilder setupBuilder, Type objectType, Func<object, object> transformer)
+        {
+            if (objectType == null)
+                throw new ArgumentNullException(nameof(objectType));
+            if (transformer == null)
+                throw new ArgumentNullException(nameof(transformer));
+
+            var original = ConfigurationItemFactory.Default.ObjectTypeTransformer;
+            ConfigurationItemFactory.Default.ObjectTypeTransformer = new ObjectTypeTransformation(objectType, transformer, original);
+            return setupBuilder;
+        }
+
+        private class ObjectTypeTransformation<T> : IObjectTypeTransformer
+        {
+            private readonly IObjectTypeTransformer _original;
+            private readonly Func<T, object> _transformer;
+
+            public ObjectTypeTransformation(Func<T, object> transformer, IObjectTypeTransformer original)
+            {
+                _original = original;
+                _transformer = transformer;
+            }
+
+            public object TryTransformObject(object obj)
+            {
+                if (obj is T rawObject)
+                {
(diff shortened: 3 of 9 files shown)


304NotModified:
Hi, i'm sorry but I;m a bit sorry. I haven't checked the PR, but try to do that this week.

For me two things are important:

- it should work for sub classes
- it should work for interfaces

I hope this PR fulfills those things :)

snakefoot:
@304NotModified The unit-test I stole have been modified so it actually registers the interface-type as the "dangerous" type. And correctly skips objects implementing that interface.

304NotModified:
so that's a yes, isn't?

snakefoot:
One could also consider changing into this interface:

```
    public interface IObjectTypeTransform
    {
        /// <summary>
        /// Takes a dangerous object and transforms into a safe object
        /// </summary>
        /// <returns>
        /// Null if unknown object.
        /// </returns>
        object TryTransform(object obj);
    }
```

Those who wants to transform into string will just return the string-value.

mmurrell:
@snakefoot  - Looks good.  

Only concern I have at the moment is that my logging library (various helper methods, DI proxy class, etc) will require a strong reference / dependency to Entity Framework just to exclude this type... I guess that can be mitigated by adding a transform in the EF Context's Static Constructor, but that seems like it could be kind of hacky...

Any way to invoke `RegisterObjectTransformation` given a type name as a string?

snakefoot:
> @mmurrell Any way to invoke RegisterObjectTransformation given a type name as a string?

Can make a helper method that takes a type-string (Typename + AssemblyName), and then lazy loads the Type from the string.

snakefoot:
@mmurrell Added an extra unit-test showing how one can use Type.AssemblyQualifiedName to setup object-transformation as ToString().

304NotModified:
It would be nice if we could configure this from the (xml) config (in the future)

I Haven't found time yet to complete the review unfortunately :/

snakefoot:
Probably for a new PR after having fixed dependency injection with NLog 5.0


Sent from my Sony Xperia


It would be nice if we could configure this from the (xml) config (in the future)

I Haven't found time yet to complete the review unfortunately :/

mmurrell:
Looks good by me.

src/NLog/Internal/ObjectReflectionCache.cs
```diff
@@ -42,15 +42,23 @@ namespace NLog.Internal
...
 
     /// <summary>
     /// Converts object into a List of property-names and -values using reflection
     /// </summary>
-    internal class ObjectReflectionCache
+    internal class ObjectReflectionCache : IObjectTypeTransform
     {
         readonly MruCache<Type, ObjectPropertyInfos> _objectTypeCache = new MruCache<Type, ObjectPropertyInfos>(10000);
-        static readonly Dictionary<Type, Func<object, IDictionary<string,object>>> _objectTypeOverride = new Dictionary<Type, Func<object, IDictionary<string, object>>>();
+        private IObjectTypeTransform ObjectTypeReflection => _objectTypeReflection ?? (_objectTypeReflection = ConfigurationItemFactory.Default.ObjectTypeTransform);
```
304NotModified: I would be nice if we could decouple this from ConfigurationItemFactory.Default, but I guess that's a lot more work :(

tests/NLog.UnitTests/Targets/DefaultJsonSerializerClassTests.cs
```diff
@@ -0,0 +1,131 @@
...
+        public void IExcludedInterfaceSerializer_RegistersSerializeAsToString_InvokesToString()
+        {
+            try
+            {
+                var testObject = BuildSampleObject();
+
+                var sb = new StringBuilder();
+                var options = new JsonSerializeOptions();
+
+                LogManager.LogFactory.RegisterObjectTransformation<IExcludedInterface>(o => o.ToString());
```
304NotModified: Maybe nice if we could add a helper to skip, to make things more clear (and we could performance optimize it in the future)

304NotModified:
this also fixes https://github.com/NLog/NLog/issues/3264, isn't?

src/NLog/LogFactoryExtensions.cs
```diff
@@ -0,0 +1,133 @@
...
+{
+    /// <summary>
+    /// Extensions for NLog <see cref="LogFactory"/>.
+    /// </summary>
+    public static class LogFactoryExtensions
+    {
+        /// <summary>
+        /// Registers object Type transformation from dangerous (massive) object to safe (reduced) object
+        /// </summary>
+        public static LogFactory RegisterObjectTransformation<T>(this LogFactory logFactory, Func<T, object> transformer)
```
304NotModified: Not sure if the names are really clear or even to long, we could also do something like this:

`LogFactory.SerializationOptions.RegisterTransformation(...)`

(and also maybe `LogFactory.SerializationOptions.SkipSerialization(type)`)

304NotModified: it's nice grouped then. If we need it on LogFactory, NLog or maybe LogManager is also something we could think of

snakefoot: These a helper methods to hide away the dependency-injection chains. Suddenly making a bucket for certain interfaces called SerializationOptions will not be good.

304NotModified: Don't you think LogFactory gets to many methods which aren't related to each other?

tests/NLog.UnitTests/Targets/DefaultJsonSerializerClassTests.cs
snakefoot: Like to limit the number of extensions methods until we have a direction. Right now only two people have been knocking on this door for this feature. And only one of them have approved this direction.

src/NLog/Internal/ObjectReflectionCache.cs
snakefoot: Until I have completed my refactoring of NLog 5.0 dependency injection, then ConfigurationItemFactory is currently the goto-place for dependency injection stuff.

tests/NLog.UnitTests/Targets/DefaultJsonSerializerClassTests.cs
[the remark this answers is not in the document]
snakefoot: Like lipstick on a pig, since there is `ConfigurationItemFactory.Default`

src/NLog/LogFactoryExtensions.cs
```diff
@@ -0,0 +1,133 @@
...
+            public ObjectTypeTransformation(string objectTypeName, Func<object, object> transformer, IObjectTypeTransform original)
+            {
+                _original = original;
+                _transformer = transformer;
+                _objectTypeName = objectTypeName.Trim();
+            }
+
+            public object TryTransformObject(object obj)
+            {
+                if (_objectType == null && _objectTypeName != null)
```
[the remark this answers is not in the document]
snakefoot: If the Type-lookup fails at runtime, then it assigns _objectTypeName  to null. To avoid further issues.

src/NLog/IObjectTypeTransform.cs
```diff
@@ -0,0 +1,49 @@
...
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF 
+// THE POSSIBILITY OF SUCH DAMAGE.
+// 
+
+namespace NLog
+{
+    /// <summary>
+    /// Interface for handling object transformation
+    /// </summary>
+    public interface IObjectTypeTransform
```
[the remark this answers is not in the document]
snakefoot: Not sure I understand why?

src/NLog/Config/ConfigurationItemFactory.cs
```diff
@@ -193,6 +193,15 @@ public IValueFormatter ValueFormatter
             set => MessageTemplates.ValueFormatter.Instance = value;
         }
 
+        /// <summary>
+        /// Gets or sets the object-reflection serializer for use in <see cref="JsonLayout"/> or <see cref="XmlLayout"/>
+        /// </summary>
+        public IObjectTypeTransform ObjectTypeTransform
```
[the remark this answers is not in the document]
snakefoot: Not sure I understand why. This is how we add custom logic behavior in NLog

src/NLog/LogFactoryExtensions.cs
```diff
@@ -0,0 +1,80 @@
...
+    /// </summary>
+    public static class LogFactoryExtensions
+    {
+        /// <summary>
+        /// Registers object Type as being dangerous and should not probe its properties (Ex. Entity Framework)
+        /// </summary>
+        public static LogFactory SkipObjectTypeReflection(this LogFactory logFactory, Type objectType)
+        {
+            var original = ConfigurationItemFactory.Default.ObjectTypeReflection;
+            ConfigurationItemFactory.Default.ObjectTypeReflection = new SkipObjectTypeReflectionHelper(objectType, original);
```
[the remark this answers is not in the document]
snakefoot: Yes lets make it pink :). Can you elaborate?

tests/NLog.UnitTests/Targets/DefaultJsonSerializerClassTests.cs
```diff
@@ -0,0 +1,107 @@
...
+        public void ExcludedClassSerializer_RegistersSerializeAsToString_InvokesToString()
+        {
+            try
+            {
+                var testObject = BuildSampleObject();
+
+                var sb = new StringBuilder();
+                var options = new JsonSerializeOptions();
+
+                LogManager.LogFactory.SkipObjectTypeReflection(typeof(IExcludedClass));
```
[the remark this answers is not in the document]
snakefoot: Please elaborate

src/NLog/LogFactoryExtensions.cs
snakefoot: I hope LogFactory will have all the methods for configuring NLog and for creating Logger-objects.

src/NLog/LogFactoryExtensions.cs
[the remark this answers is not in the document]
snakefoot: Btw. you are commenting on previous commits, and the code no longer exists.

src/NLog/IObjectTypeTransform.cs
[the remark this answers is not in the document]
snakefoot: Ok. Made them internal. Like the idea of keeping the visibility to a minimum until there is a direction.

src/NLog/Config/ConfigurationItemFactory.cs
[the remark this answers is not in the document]
snakefoot: Ok. Made them internal. Like the idea of keeping the visibility to a minimum until there is a direction.

src/NLog/LogFactoryExtensions.cs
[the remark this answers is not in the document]
304NotModified: ow yes that's a intermediate comment for myself! (just a quick note to remember)

I started the review on phone, but then some things has changed.

Sorry for the blunt stuff - please ignore this. :)

snakefoot:
> @304NotModified this also fixes #3264, isn't?

Yes it will resolve that question also, as NLog will have something very similar to Serilog Transform.

More improvement can probably be made, as you have already recognized:
-  NLog is still missing the ability to setup user-defined static-methods-names in NLog.config to perform the transformation (together with object-type-name)
   - One of these static-methods-names could be a public static NLog-method that performs ToString (Allowing one to skip reflection for certain object-type-names in Nlog.config).

src/NLog/LogFactoryExtensions.cs
snakefoot: I have now added a new commit with `IConfigBuilder`. Then all extension methods related to updating LogFactory-configuration can be made as fluent extension methods for this interface. Is that an improvement?

src/NLog/ConfigBuilderExtensions.cs
```diff
@@ -0,0 +1,140 @@
...
+        {
+            var original = ConfigurationItemFactory.Default.ObjectTypeTransform;
+            ConfigurationItemFactory.Default.ObjectTypeTransform = new ObjectTypeTransformation<T>(transformer, original);
+            return configBuilder;
+        }
+
+        /// <summary>
+        /// Registers object Type transformation from dangerous (massive) object to safe (reduced) object
+        /// </summary>
+        public static IConfigBuilder RegisterObjectTransformation(this IConfigBuilder configBuilder, string objectTypeName, Func<object,object> transformer)
```
304NotModified: I'm not keen on this name, but I have to think about for a proposal. 

I don't think users would understand this name without ready docs. Maybe we should include something like destructure or serialize?

snakefoot:
I only introduced the IConfigBuilder interface because you didn't like the LogFactory extension methods. And I want to hide the ConfigurationItemFactory away.


The word destructure is even more confusing to me than transform (because I'm old and know about destructors). This feature allows one to override the default reflection for an object. The reflection is used to transform an object into a dictionary. This dictionary can then be used for serialization or other things.


Sent from my Sony Xperia


I think feature would make 4.6.8, a 4.7! :)

In tests/NLog.UnitTests/Targets/DefaultJsonSerializerClassTests.cs<https://github.com/NLog/NLog/pull/3610#discussion_r334678718>:

> +                Included = new IncludedClass { IncludedString = "serialized" }
+            };
+            return testObject;
+        }
+
+        [Fact]
+        public void IExcludedInterfaceSerializer_RegistersSerializeAsToString_InvokesToString()
+        {
+            try
+            {
+                var testObject = BuildSampleObject();
+
+                var sb = new StringBuilder();
+                var options = new JsonSerializeOptions();
+
+                LogManager.Config().RegisterObjectTransformation<IExcludedInterface>(o => o.ToString());


While I like the approach, I find this a bit confusing as we have also LogManager.Configuration.

In src/NLog/Internal/ObjectReflectionCache.cs<https://github.com/NLog/NLog/pull/3610#discussion_r334679026>:

>      {
         readonly MruCache<Type, ObjectPropertyInfos> _objectTypeCache = new MruCache<Type, ObjectPropertyInfos>(10000);
-        static readonly Dictionary<Type, Func<object, IDictionary<string,object>>> _objectTypeOverride = new Dictionary<Type, Func<object, IDictionary<string, object>>>();
+        private IObjectTypeTransform ObjectTypeReflection => _objectTypeReflection ?? (_objectTypeReflection = ConfigurationItemFactory.Default.ObjectTypeTransform);


good one!

In src/NLog/ConfigBuilderExtensions.cs<https://github.com/NLog/NLog/pull/3610#discussion_r334679782>:

> +        }
+
+        /// <summary>
+        /// Registers object Type transformation from dangerous (massive) object to safe (reduced) object
+        /// </summary>
+        public static IConfigBuilder RegisterObjectTransformation<T>(this IConfigBuilder configBuilder, Func<T, object> transformer)
+        {
+            var original = ConfigurationItemFactory.Default.ObjectTypeTransform;
+            ConfigurationItemFactory.Default.ObjectTypeTransform = new ObjectTypeTransformation<T>(transformer, original);
+            return configBuilder;
+        }
+
+        /// <summary>
+        /// Registers object Type transformation from dangerous (massive) object to safe (reduced) object
+        /// </summary>
+        public static IConfigBuilder RegisterObjectTransformation(this IConfigBuilder configBuilder, string objectTypeName, Func<object,object> transformer)


I'm not keen on this name, but I have to think about for a proposal.

I don't think users would understand this name without ready docs. Maybe we should include something like destructure or serialize?

304NotModified:
I really like this feature, but I think we should agree on an API before implementing for this.

snakefoot:
> I really like this feature, but I think we should agree on an API before implementing for this.

Think it is a good idea to post-pone it for NLog 4.7. Sadly enough it means @mmurrell and @giorgi will not be able to fully use #3611, until we have agreed on how to start on fluent-config-api.

Will be waiting until you get the time to present your own thoughts on how the config-api should work.

304NotModified:
Should the API on LogManager/LogFactory, or could we use a new class?

snakefoot:
> What about: (or on another class, see above)

That is what I'm suggesting by using `IConfigBuilder`. Think unregister doesn't make sense.

304NotModified:
Well it should be possible to undo the change? (maybe even list all registered changes)

snakefoot:
> Well it should be possible to undo the change? (maybe even list all registered changes)

My plan is that LogFactory.ServiceRepository contains all custom logic (See also #3616). So I just want a interface for registration of registration-interfaces, while waiting for this container to arrive in NLog 5.0 Then one just create a new LogFactory to get a clean scratch to start from.

304NotModified:
> just create a new LogFactory to get a clean plate to start from.

Sounds a bit unfriendly, but it could work.

304NotModified:
> Should the API on LogManager/LogFactory, or could we use a new class?

I really like to know of there is a technical reason for LogManager/LogFactory.

snakefoot:
> I really like to know of there is a technical reason for LogManager/LogFactory.

In this PR then I have now introduced IConfigBuilder as base-point for configuring LogFactory (Along with LogFactory.Configuration and LogFactory.ServiceRepository when it arrives).

304NotModified:
So its that a yes or a no ;)

snakefoot:
> So its that a yes or a no ;)

I have updated my reply above. Please re-read, and tell me what you want no/yes answer to.

tests/NLog.UnitTests/Targets/DefaultJsonSerializerClassTests.cs
```diff
@@ -0,0 +1,131 @@
...
+        public void IExcludedInterfaceSerializer_RegistersSerializeAsToString_InvokesToString()
+        {
+            try
+            {
+                var testObject = BuildSampleObject();
+
+                var sb = new StringBuilder();
+                var options = new JsonSerializeOptions();
+
+                LogManager.Config().RegisterObjectTransformation<IExcludedInterface>(o => o.ToString());
```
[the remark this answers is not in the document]
snakefoot: So we just need to find a different word than `Config` because it reminds too much of  `Configuration`. And your suggestion is `SerializationConfiguration`. Not sure if it is a better word, since LoggingConfiguration also controls serialization :) But I will try some brainstorming.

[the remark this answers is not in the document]
snakefoot: Renamed `Config()` to `Setup()` and updated code-example in original post.

src/NLog/ConfigBuilderExtensions.cs
snakefoot: Well the method being registered takes and object and transforms it into another object. Not sure it can be much simpler without removing important information.

snakefoot:
> So its that a yes or a no ;)

Have update the example to this:

`NLog.LogManager.Setup().RegisterObjectTransformation<IExcludedInterface>(o => o.ToString());`

src/NLog/LogFactoryExtensions.cs
snakefoot: Renamed from `Config()` to `Setup()` to avoid confusion with `LogManager.Configuration`

snakefoot:
Merged with master and now uses ISetupBuilder from #3686

src/NLog/SetupSerializationBuilderExtensions.cs
```diff
@@ -0,0 +1,149 @@
...
+        {
+            var original = ConfigurationItemFactory.Default.ObjectTypeTransformer;
+            ConfigurationItemFactory.Default.ObjectTypeTransformer = new ObjectTypeTransformation<T>(transformer, original);
+            return setupBuilder;
+        }
+
+        /// <summary>
+        /// Registers object Type transformation from dangerous (massive) object to safe (reduced) object
+        /// </summary>
+        public static ISetupSerializationBuilder RegisterObjectTransformation(this ISetupSerializationBuilder setupBuilder, string objectTypeName, Func<object, object> transformer)
```
304NotModified: objectTypeName is the full qualified type name, isn't? Would `type` not a better option?

snakefoot: Well it was actually done on requests, where the user wants to force tostring on certain objects without the project has to depend directly on the nuget-packages where object-types exists. See also https://github.com/NLog/NLog/pull/3610#issuecomment-537760596

304NotModified: Clear. But think this is a bad idea. Of course he could fix this by passing the builder to s project which has a reference to it.

snakefoot: If having the actual type, then one can use `RegisterObjectTransformation<T>`. I can add a bonus method that takes a Type as parameter?

snakefoot: Btw. you are talking with the guy that is not going to use this feature at all. So not sure I will put up the best fight :) But it sounded really important for @mmurrell that requested this entire feature the first place. Kind of funny to only give half of the feature.

304NotModified: Yeah sorry about that @mmurrell, but I think this weakens the API and it not strictly necessary.

snakefoot: @304NotModified I guessing that the use of `string objectTypeName` will also be handy, when one day supporting configuration from NLog.config. Btw. it is not super-duper fully qualified, as it is without assembly name.

snakefoot: @304NotModified Ignore my comment about it not being super-duper fully qualified. The `Type.GetType` is just lazy evaluated and it just disables the transformation if the type-name is wrong.

snakefoot:
@304NotModified So remove lazy evaluation of types, and require the caller have performed the `Type.GetType`-call upfront ? (Change `string objectTypeName` to `Type objectType`)

304NotModified:
yes, then I will merge it. We could always add the string version in the future, but I pretty sure it's really not needed if used in the right place.

src/NLog/SetupSerializationBuilderExtensions.cs
snakefoot: Change from `string objectTypeName` to `Type objectType`

snakefoot:
Have now committed, so it is without the string-parameter.

snakefoot:
And the build was green. Can almost smell the NLog 4.7

304NotModified:
🚀 merged!

snakefoot:
Updated wiki:

https://github.com/NLog/NLog/wiki/Configure-from-code#logfactory-initial-setup

PR #3611, opened 2019-10-02
${Exception} - Added Format options: HResult and Properties

### LINKED ISSUE #3519: Custom renderer for exceptions

snakefoot:
Trying to resolve #3519 together with #3610

```c#
NLog.LogManager.LogFactory.RegisterObjectTransformation<System.Net.WebException>(ex =>new {
    Type = ex.GetType().ToString(),
    Message = ex.Message,
    StackTrace = ex.StackTrace,
    Source = ex.Source,
    InnerException = ex.InnerException,
    Status = ex.Status,
    Response = ex.Response.ToString(),  // Call your custom method to render stream as string
});
```

Then you can do this in NLog.config:

```xml
${exception:format=Properties}
```

Then it will output `Status` and `Response`  like they were entries in the Exception.Data-dictionary.

Alternative one could just setup a conversion of the type `System.Net.WebResponse` so it becomes a string. Then any object logged by NLog that contains a `System.Net.WebResponse` will automatically output its string-value.

### DIFF
--- a/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
@@ -71,8 +71,20 @@ namespace NLog.LayoutRenderers
                                                                                                         {"PROPERTIES",ExceptionRenderingFormat.Properties},
                                                                                                     };
 
-        private static HashSet<string> ExcludeDefaultProperties => _excludeDefaultProperties  ?? (_excludeDefaultProperties  = new HashSet<string>(typeof(Exception).GetProperties().Where(p => p.IsValidPublicProperty()).Select(p => p.Name).Concat(new [] { "Type" }), StringComparer.OrdinalIgnoreCase));
-        private static HashSet<string> _excludeDefaultProperties;
+        private static readonly HashSet<string> ExcludeDefaultProperties = new HashSet<string>(new[] {
+            "Type",
+            nameof(Exception.Data),
+            nameof(Exception.HelpLink),
+            nameof(Exception.InnerException),
+            nameof(Exception.Message),
+            nameof(Exception.Source),
+            nameof(Exception.StackTrace),
+#if SILVERLIGHT || NETSTANDARD1_0
+            "TargetSite",
+#else
+            nameof(Exception.TargetSite),
+#endif
+        });
 
         private ObjectReflectionCache ObjectReflectionCache => _objectReflectionCache ?? (_objectReflectionCache = new ObjectReflectionCache());
         private ObjectReflectionCache _objectReflectionCache;
@@ -429,11 +441,10 @@ namespace NLog.LayoutRenderers
         protected virtual void AppendProperties(StringBuilder sb, Exception ex)
         {
             string separator = string.Empty;
-            var excludeProperties = ExcludeDefaultProperties;
             var exceptionProperties = ObjectReflectionCache.LookupObjectProperties(ex);
             foreach (var property in exceptionProperties)
             {
-                if (excludeProperties.Contains(property.Name))
+                if (ExcludeDefaultProperties.Contains(property.Name))
                     continue;
 
                 var propertyValue = property.Value;
--- a/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
@@ -68,16 +68,30 @@ namespace NLog.LayoutRenderers
                                                                                                         {"STACKTRACE", ExceptionRenderingFormat.StackTrace},
                                                                                                         {"DATA",ExceptionRenderingFormat.Data},
                                                                                                         {"@",ExceptionRenderingFormat.Serialize},
+                                                                                                        {"HRESULT",ExceptionRenderingFormat.HResult},
                                                                                                         {"PROPERTIES",ExceptionRenderingFormat.Properties},
                                                                                                     };
 
         private static readonly HashSet<string> ExcludeDefaultProperties = new HashSet<string>(new[] {
             "Type",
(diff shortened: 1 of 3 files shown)


snakefoot:
@Giorgi Any input whether this extra option would help solve your issue?

You would have to register a type-reflector using the interface from  #3610 that detects WebException. And converts it into a dictionary of name and value pairs. Ex:

```
"Type" => ex.GetType()
"Message" => ex.Message
"TargetSite" => ex.TargetSite
"Source" => ex.Source
"Data" => ex.Data
"InnerException" => ex.InnerException
"Response" => ConvertStreamToString(ex.Response.GetResponseStream())
```

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -66,8 +68,41 @@ public class ExceptionLayoutRenderer : LayoutRenderer, IRawValue
...
+                                                                                                        {"PROPERTIES",ExceptionRenderingFormat.Properties},
                                                                                                     };
 
+        private static readonly HashSet<string> ExcludeDefaultProperties = new HashSet<string>(new[] {
+            "Type",
+            nameof(Exception.Data),
+#if !SILVERLIGHT
+            nameof(Exception.HelpLink),
+#else
+            "HelpLink",
```
304NotModified: While I like `nameof`s, I think just "HelpLink" is more clear? The code is more complicated with all the `#if` (and more difficult to refactor, instead of easier to refactor with `nameof`)

304NotModified: same for below.

Please remove the `#if`s

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -382,6 +419,21 @@ protected virtual void AppendSource(StringBuilder sb, Exception ex)
...
 
+        /// <summary>
+        /// Appends the HResult of an Exception to the specified <see cref="StringBuilder" />.
+        /// </summary>
+        /// <param name="sb">The <see cref="StringBuilder"/> to append the rendered data to.</param>
+        /// <param name="ex">The Exception whose HResult should be appended.</param>
+        protected virtual void AppendHResult(StringBuilder sb, Exception ex)
+        {
+#if NET45
+            if (ex.HResult != 0 && ex.HResult != 1)
```
304NotModified: 0 are 1 are special? Is that somewhere documented? Maybe reference here? 

I only see 0 here: https://docs.microsoft.com/en-us/windows/win32/seccrypto/common-hresult-values

304NotModified: we could use reflection for older .NET :angel:

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -411,6 +463,30 @@ protected virtual void AppendSerializeObject(StringBuilder sb, Exception ex)
...
+        {
+            string separator = string.Empty;
+            var exceptionProperties = ObjectReflectionCache.LookupObjectProperties(ex);
+            foreach (var property in exceptionProperties)
+            {
+                if (ExcludeDefaultProperties.Contains(property.Name))
+                    continue;
+
+                var propertyValue = property.Value;
+                if (propertyValue == null || string.IsNullOrEmpty(propertyValue as string))
```
304NotModified: is this on purpose that we use here `as string` and we `ToString`? The latter could also return empty? (if so, please add a comment line, thanks!)

snakefoot: Fixed

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
snakefoot: Added const int documentation

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
snakefoot: Removed the `#if`

Giorgi:
@snakefoot Can you add couple of configuration examples to show how to use the new rendering options ?

Giorgi:
Looks good to me. Thanks!

304NotModified:
Thanks!!

304NotModified:
Updated https://github.com/NLog/NLog/wiki/Exception-layout-renderer !

PR #3613, opened 2019-10-03
TypeDescriptor.GetConverter can throw after linker tree shake

snakefoot:
See also: https://github.com/mono/linker/issues/153

### DIFF
--- a/src/NLog/Internal/PropertyHelper.cs
+++ b/src/NLog/Internal/PropertyHelper.cs
@@ -314,6 +314,18 @@ namespace NLog.Internal
                 return true;
             }
 
+            if (type == typeof(LineEndingMode))
+            {
+                newValue = LineEndingMode.FromString(value);
+                return true;
+            }
+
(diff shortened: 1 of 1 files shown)


src/NLog/Internal/PropertyHelper.cs
```diff
@@ -421,28 +433,25 @@ private static object ExtractHashSetComparer(object obj, PropertyInfo propInfo)
...
-                return true;
-            }
-#else
-            if (type == typeof(LineEndingMode))
+            try
             {
-                newValue = LineEndingMode.FromString(value);
-                return true;
+#if !SILVERLIGHT && !NETSTANDARD1_3
+                var converter = TypeDescriptor.GetConverter(type);
```
304NotModified: we could unit test this with NSubstitute and a wrapper arround `TypeDescriptor`. what do you think?

304NotModified:
👍 thanks

PR #3618, opened 2019-10-09
Faster reflection of NLog configuration items by skipping Type.IsDefined 

snakefoot:
Can load 100 NLog-configs within 3 secs (before 6 secs):

Test application:

```c#
        {
            for (int i = 0; i < 100; ++i)
            {
                NLog.Config.ConfigurationItemFactory.Default = null;
                NLog.LogManager.Configuration = null;
                var logger = NLog.LogManager.GetLogger(i.ToString());
                logger.Fatal("Testing");
            }
        }
```

Still not super duper fast, but still an improvement.

Note merge to master branch.

### DIFF
--- a/src/NLog/Config/Factory.cs
+++ b/src/NLog/Config/Factory.cs
@@ -31,14 +31,14 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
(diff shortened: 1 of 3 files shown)


304NotModified:
thanks!

Issue #3621, opened 2019-10-11
Threads deadlocking with AsyncWrapper


taion809:
**NLog version**: 4.6.4

**Platform**: .NET Core 2.2.7 Linux

**Current NLog config** (xml or C#, if relevant)

```F#
let configureLogging (logLevel: LogLevel) =
    LayoutRenderer.Register<HostSiteLayoutRenderer>("host-site");
    LayoutRenderer.Register<CorrelationLayoutRenderer>("corr");
    let hostSite = Environment.getEnvironmentVariable "HostSite" (Some "local")
    let logLayoutString = sprintf "${longdate}|${level:uppercase=true}|${logger}|${host-site:%s}${corr}|${message}" hostSite
    let config = new LoggingConfiguration()
    let cLayout = Layout.FromString(logLayoutString);
    let consoleTarget = new ConsoleTarget( Name = "logToConsole", Layout = cLayout)
    let bufferedTarget = new Targets.Wrappers.AsyncTargetWrapper(consoleTarget, 10000, Wrappers.AsyncTargetWrapperOverflowAction.Block)
    config.LoggingRules.Add(new LoggingRule("*", logLevel, bufferedTarget))
    LogManager.Configuration <- config
```

During a soak test we discovered that at a certain threshold memory would climb for some processes and OOM.

![oom](https://user-images.githubusercontent.com/5170979/66679523-5f258c80-ec3c-11e9-86c2-1e890fc5c934.png)

I was able to create a coredump of the process.  Upon inspection there were over 1100 blocked threads.  I dumped the stack for these and found the following:
```
00007F85D2FFA3A0 00007f86746a09f3 [HelperMethodFrame_1OBJ: 00007f85d2ffa3a0] System.Threading.Monitor.ObjWait(Boolean, Int32, System.Object)
00007F85D2FFA4D0 00007F85FE7F08B0 NLog.Targets.Wrappers.ConcurrentRequestQueue.WaitForBelowRequestLimit()
00007F85D2FFA500 00007F85FCF5B2AF NLog.Targets.Wrappers.ConcurrentRequestQueue.Enqueue(NLog.Common.AsyncLogEventInfo)
00007F85D2FFA570 00007F85FCF55E96 NLog.Targets.Wrappers.AsyncTargetWrapper.Write(NLog.Common.AsyncLogEventInfo)
[...]

PR #3636, opened 2019-10-20
AsyncTargetWrapper - Eject writers on close incase timer thread is blocked

snakefoot:
Trying to resolve the following issue:

- 10 Threads attacks the async-console-target. Async-background-writer falls behind because console is slow. OverflowAction block causes 10 threads to get stuck waiting for async-background-writer to catch up.
- NLog Configuration is reloaded. AsyncWrapper is closed (and stops async-background-writer), but never releases any of the threads being stuck inside waiting for queue. New async-console-target is created.
- 10 new threads attacks the new async-console-target. Story repeats itself until having 1100 threads stuck inside old async-console-target-objects.

See also #3621

Note merge to master


snakefoot:
Guess the issue is real. But need to improve the scenario.

PR #3639, opened 2019-10-20
WrapperLayoutRendererBuilderBase - Obsolete and no longer used

snakefoot:
Moving WrapperLayoutRendererBuilderBase one step closer to the edge of oblivion. The last remains of a failed optimization attempt, that should not be followed by anyone.

### DIFF
--- a/src/NLog/LayoutRenderers/Wrappers/WrapperLayoutRendererBuilderBase.cs
+++ b/src/NLog/LayoutRenderers/Wrappers/WrapperLayoutRendererBuilderBase.cs
@@ -41,6 +41,7 @@ namespace NLog.LayoutRenderers.Wrappers
     /// 
     /// This expects the transformation to work on a <see cref="StringBuilder"/>
     /// </summary>
+    [Obsolete("Inherit from WrapperLayoutRendererBase and override RenderInnerAndTransform() instead. Marked obsolete in NLog 5.0")]
     public abstract class WrapperLayoutRendererBuilderBase : WrapperLayoutRendererBase
     {
         /// <inheritdoc/>
@@ -91,7 +92,7 @@ namespace NLog.LayoutRenderers.Wrappers
         /// <returns></returns>
         protected sealed override string Transform(string text)
         {
-            throw new NotSupportedException("Use TransformFormattedMesssage");
+            throw new NotSupportedException("Inherit from WrapperLayoutRendererBase and override RenderInnerAndTransform()");
         }
 
         /// <summary>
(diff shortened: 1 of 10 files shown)


src/NLog/LayoutRenderers/Wrappers/WrapperLayoutRendererBuilderBase.cs
```diff
@@ -1,107 +0,0 @@
...
-using System.Text;
-
-namespace NLog.LayoutRenderers.Wrappers
-{
-    /// <summary>
-    /// Base class for <see cref="LayoutRenderer"/>s which wrapping other <see cref="LayoutRenderer"/>s. 
-    /// 
-    /// This expects the transformation to work on a <see cref="StringBuilder"/>
-    /// </summary>
-    public abstract class WrapperLayoutRendererBuilderBase : WrapperLayoutRendererBase
```
304NotModified, line 44: As this one was never obsolete, I propose to add the obsolete attribute and it's OK too remove it in NLog 6.

snakefoot: No one is using it. So see no reason to not cleanup as much as possible in NLog 5

304NotModified: How do we know that? It's public. We clean as much as possible by not using it?

snakefoot: Thought the whole idea with Major versions was the ability to make breaking changes. And not being required to make obsolete first. Especially when extremely rare features.

304NotModified: Yes breaking changes are allowed, but Obsolete make this decent.

I don't see the issue in let this class live for now. I won't influence other code. It's only ugly...

304NotModified: Please image your are a enduser, uses only public classes, not warnings with obsoletes and suddenly the type is gone. We can't even show a nice message then.

snakefoot: I'm guessing that a full chapter has be written where explaining all breaking changes, and new default values. This should be read careful when upgrading from old version to NLog 5. But I will add the class back and mark it as obsolete.

PR #3640, opened 2019-10-21
Add null and recursive tests for ValueFormatter

aerotog:
Helps with #3181. Includes some typo fixes in the same file.

I'm slightly concerned that the serialization result for the primitive 'null' is different for each CaptureType. Please verify the asserted values are expected for each CaptureType:
```
        [InlineData(CaptureType.Normal, "NULL")]
        [InlineData(CaptureType.Serialize, "null")]
        [InlineData(CaptureType.Stringify, "\"\"")]
```

### DIFF
--- a/tests/NLog.UnitTests/MessageTemplates/ValueFormatterTest.cs
+++ b/tests/NLog.UnitTests/MessageTemplates/ValueFormatterTest.cs
@@ -1,4 +1,4 @@
-﻿// 
+// 
 // Copyright (c) 2004-2019 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
 // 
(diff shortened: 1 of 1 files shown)


304NotModified:
Thanks!

I think you found a bug! Have to think about it what's expected output.

304NotModified:
>   [InlineData(CaptureType.Normal, "NULL")]

This is expected (in the docs)
>         [InlineData(CaptureType.Serialize, "null")]

Serializable should be parseable, so also expected.
>         [InlineData(CaptureType.Stringify, "\"\"")]

Doubting about this. Maybe it's nice you could control `""` or `null`, so maybe this is also OK

304NotModified:
Merge this one anyway, thanks!!

PR #3641, opened 2019-10-21
${CallSite} - Handle MoveNext More robust

snakefoot:
CallSite - NetCore3 Async uses MoveNext with AsyncMethodBuilderCore

NetCore can produce StackTrace with MoveNext from internal type System.Runtime.CompilerServices.AsyncMethodBuilderCore.

Instead of checking specific types, then just check if "MoveNext" comes from "System.Runtime.CompilerServices"-namespace.

This fixes tests that are failing when using NetCore3 on VS2019.

Note merge to master.

### DIFF
--- a/src/NLog/Internal/StackTraceUsageUtils.cs
+++ b/src/NLog/Internal/StackTraceUsageUtils.cs
@@ -77,6 +77,8 @@ namespace NLog.Internal
                 if (endIndex > 1)
                 {
                     methodName = callerClassType.Name.Substring(1, endIndex - 1);
+                    if (methodName.StartsWith("<"))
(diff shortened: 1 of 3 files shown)


snakefoot:
This change has the side-effect that it now also recognizes MoveNext for this inner internal class:

System.Runtime.CompilerServices.AsyncMethodBuilderCore+MoveNextRunner

snakefoot:
Have now "fixed" the unit-test `CleanNamesOfAsyncContinuationsIsFalse` by removing it. Since NLog now automatically activate MoveNext cleanup for `LogAfterAwait`

snakefoot:
Succeeded in creating a new tests where `${callsite}` fails when not having activated all options when using Task.Run:

`${callsite:cleanNamesOfAsyncContinuations=true:cleanNamesOfAnonymousDelegates=true}`

Issue #3632, opened 2019-10-18
Substring Layout Renderer cannot parse length=null from config file


evolve700:
Hi,
I'm trying to use the new [Substring Layout Renderer](https://github.com/NLog/NLog/wiki/substring-layout-renderer) (introduced in 4.6) to skip the first n characters of my log message and output the rest of the message until the last character.

It appears that the Length parameter is supposed to support `null`, but it fails during parsing of the config file. 

I don't get how we're supposed to configure this, since Length is a `RequiredParameter`, it's DefaultValue is null, but we cannot configure null via the configuration file. Seems like a bug to me.

**NLog version**: 4.6.7 (via NuGet)

**Platform**: .Net 4.5

**Current NLog config** (xml or C#, if relevant)

```xml
<nlog>
  <targets>
    <target xsi:type="File" name="xpo" fileName="${basedir}/logs/${shortdate}.xpo.log" layout="${longdate}|${uppercase:${level}}|${substring:${message}:start=22:length=null}" />
  </targets>
  <rules>
    <logger name="XPO" minlevel="Debug" writeTo="xpo" final="true" />
  </rules>
</nlog>

```

- What is the current result?
The current result is an error during parsing of the NLog configuration. It is also not possible to omit the Length parameter in the configuration because it is marked as a RequiredParameter.
- What is the expected result?
Since null is a supported value there should be a way to configure null via the app.config/nlog.config.
- Did you checked the [Internal log](https://github.com/NLog/NLog/wiki/Internal-Logging)?
Here is my internal log with the exception and stacktrace: [nlog-internal.log](https://github.com/NLog/NLog/files/3743033/nlog-internal.log)
- Are there any workarounds?
None, that I have found
- Is there a version in which it did work?
Not tested

304NotModified:
Thanks for reporting. 

Maybe we should remove the `null` think, as you could use ${left} or ${right} if one of the two values in empty (e.g. null?)

evolve700:
Thanks for getting back to me. However I fail to see how ${left} or ${right} are supposed to help here. I want to **skip** the first X characters of my output message, but I don't know how long the string is since it's variable length -> each message has a different length, the only thing I know is that I can always skip the first X=22 chars of my message.

Taking the ${right} part of my message would require me to know its exact length and subtract X from it. I don't know a way to do this with NLog. Taking a ${left} part is not my requirement.

I think allowing null is exactly what is required here for the length-parameter of the ${substring} renderer.

[...]

PR #3647, opened 2019-10-25
${substring}: length should not be mandatory

304NotModified:
fixes https://github.com/NLog/NLog/issues/3632


304NotModified:
updated https://github.com/NLog/NLog/wiki/Substring-layout-renderer

PR #3651, opened 2019-10-27
ConcurrentRequestQueue - DequeueBatch should handle that input list is non-empty

snakefoot:
Not really a bug, but just a very minor improvement of internal collection.

Note merge to master.


304NotModified:
I think this PR title is hard to understand for the end users. What do you think?

snakefoot:
> I think this PR title is hard to understand for the end users. What do you think?

I have no issue with keeping this issue our of the changelog :)

304NotModified:
OK clear, thanks!

PR #3653, opened 2019-10-28
Fixes NullReferenceException in SimpleLayout PreCalculate after TryGetRawValue

snakefoot:
Fixed bug introduced with #3144

Note merge to master

### DIFF
--- a/src/NLog/Layouts/SimpleLayout.cs
+++ b/src/NLog/Layouts/SimpleLayout.cs
@@ -308,7 +308,7 @@ namespace NLog.Layouts
                     {
                         // If raw value doesn't have the ability to mutate, then we can skip precalculate
                         var success = _rawValueRenderer.TryGetRawValue(logEvent, out var value);
-                        if (success && (Convert.GetTypeCode(value) != TypeCode.Object || value.GetType().IsValueType()))
+                        if (success && value != null && (Convert.GetTypeCode(value) != TypeCode.Object || value.GetType().IsValueType()))
                             return;
                     }
                 }
--- a/src/NLog/Layouts/SimpleLayout.cs
+++ b/src/NLog/Layouts/SimpleLayout.cs
@@ -297,22 +297,39 @@ namespace NLog.Layouts
         {
             if (_rawValueRenderer != null)
             {
-                if (!IsInitialized)
+                try
                 {
-                    Initialize(LoggingConfiguration);
-                }
+                    if (!IsInitialized)
+                    {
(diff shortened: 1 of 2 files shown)


src/NLog/Layouts/SimpleLayout.cs
```diff
@@ -306,13 +306,12 @@ public override void Precalculate(LogEventInfo logEvent)
                 {
                     // If raw value doesn't have the ability to mutate, then we can skip precalculate
                     var success = _rawValueRenderer.TryGetRawValue(logEvent, out var value);
-                    if (success && Convert.GetTypeCode(value) != TypeCode.Object || value.GetType().IsValueType())
+                    if (success && (Convert.GetTypeCode(value) != TypeCode.Object || value.GetType().IsValueType()))
```
304NotModified: proposal, check on `null` explicitly:

```
  if (success && value != null && (Convert.GetTypeCode(value) != TypeCode.Object || value.GetType().IsValueType()))
```

src/NLog/Layouts/SimpleLayout.cs
```diff
@@ -323,15 +322,15 @@ internal override void PrecalculateBuilder(LogEventInfo logEvent, StringBuilder
...
-
-            try
+            if (_rawValueRenderer != null)
             {
-                if (_rawValueRenderer != null)
+                try
                 {
+                    if (!IsInitialized)
+                    {
+                        Initialize(LoggingConfiguration);
```
304NotModified: Initialize on Precalculate hasn't a try-catch either?

src/NLog/Layouts/SimpleLayout.cs
snakefoot: `Convert.GetTypeCode(null)` returns [TypeCode.Empty](https://docs.microsoft.com/en-us/dotnet/api/system.typecode), so extra check not needed.

src/NLog/Layouts/SimpleLayout.cs
snakefoot: Fixed

src/NLog/Layouts/SimpleLayout.cs
304NotModified: it's needed for the readability ;)

It's a bit hidden that there isn't a nullref exception with `value.GetType()` in this case. Also the overhead of the method call is a bit of waste

304NotModified: added it :angel:

PR #3654, opened 2019-10-28
ObjectPathRendererWrapper - Refactoring to reduce code complexity

snakefoot:
Note merge to master

### DIFF
--- a/src/NLog/LayoutRenderers/Wrappers/ObjectPathRendererWrapper.cs
+++ b/src/NLog/LayoutRenderers/Wrappers/ObjectPathRendererWrapper.cs
@@ -33,11 +33,9 @@
 
 using System;
 using System.Globalization;
-using System.Linq;
 using System.Text;
 using NLog.Config;
 using NLog.Internal;
-using NLog.Layouts;
 
 namespace NLog.LayoutRenderers.Wrappers
 {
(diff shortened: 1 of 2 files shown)


304NotModified:
? So we remove the optimization for simplelayout? (most of the cases).

snakefoot:
> ? So we remove the optimization for simplelayout? (most of the cases).

Don't think skipping a virtual function-call is worth the additional code.

snakefoot:
Also think there is a possible bug in Layout-Precalculate with ObjectPathRendererWrapper.

The ObjectPathRendererWrapper should not implement `IRawValue`-interface.

Lets say we have the following scenario:
- Inner-Layout references a complex-object (usually the reason for using `Object-path`).
- ObjectPathRendererWrapper lives in a SimpleLayout as a single _rawValueRenderer.
- SimpleLayout Precalculate asks its _rawValueRenderer whether it is "stable".
- ObjectPathRendererWrapper will perform lookup into the complex-object and returns a string-value.
- SimpleLayout Precalculate will say that a string-value is very stable and cannot change, so skips Precalculate.
- AsyncWrapper takes over and now wants to perform serialization.
- SimpleLayout has skipped the Precalculate, and instead ask its _rawValueRenderer to render.
- _rawValueRenderer lookup the complex-object, but it has been disposed, so when doing `Object-path`-lookup then it returns null (instead of original string-value).

304NotModified:
> SimpleLayout Precalculate asks its _rawValueRenderer whether it is "stable".

Maybe this is the issue, as that is a bit unexpected. This is an assumption and not defined behaviour. Proposal: replace this assumption with an attribute or and extra method/property on IRawValue

I think it would be nice if we could drop this (from SimpleLayout):

```
var success = _rawValueRenderer.TryGetRawValue(logEvent, out var value);
                    if (success && (Convert.GetTypeCode(value) != TypeCode.Object || value.GetType().IsValueType()))
                        return;

```

as we don't need the value at all. (only the type).

snakefoot:
Yes new issue for that part is a good idea.

PR #3579, opened 2019-09-02
PerformanceCounterLayoutRenderer - Added Layout support for Instance

snakefoot:
Resolves #3569

Note merge to master

PR #3656, opened 2019-10-28
Added Typed layout

304NotModified:
Added `Layout<T>`, so we could have compile-time safe layoutables for e.g. integers.

```c#
Layout<int> = "..." // string, so will we rendered
Layout<int> =  1; // no string, so will be fast
Layout<int> =  true; // compile error :)
```


src/NLog/Layouts/Typed/EnumLayout.cs
```diff
@@ -0,0 +1,119 @@
...
+using System.Globalization;
+using System.Linq;
+using NLog.Common;
+
+namespace NLog.Layouts
+{
+    /// <summary>
+    /// Layout rendering to enum
+    /// </summary>
+    public class EnumLayout<TEnum> : TypedLayoutBase<TEnum?>
```
304NotModified: note: dunno if we need still this

src/NLog/Layouts/Typed/Layout.cs
```diff
@@ -0,0 +1,337 @@
...
+using NLog.Common;
+
+namespace NLog.Layouts
+{
+    /// <summary>
+    /// Layout with a simple value (e.g. int) or a layout which results in a simple value (e.g. ${counter})
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+  //todo  [CLSCompliant(false)] // IConvertible
+    public class Layout<T> : TypedLayoutBase<T>
```
304NotModified: note: maybe combine with TypedLayoutBase

src/NLog/Targets/EventLogTarget.cs
```diff
@@ -127,7 +126,7 @@ internal EventLogTarget(IEventLogWrapper eventLogWrapper, IAppDomain appDomain)
         /// Gets or sets the layout that renders event ID.
         /// </summary>
         /// <docgen category='Event Log Options' order='10' />
-        public Layout EventId { get; set; }
+        public Layout<int> EventId { get; set; }
```
304NotModified: changes for example usage

src/NLog/Targets/NetworkTarget.cs
```diff
@@ -143,62 +145,62 @@ public NetworkTarget(string name) : this()
         /// </summary>
         /// <docgen category='Connection Options' order='10' />
         [DefaultValue(true)]
-        public bool KeepConnection { get; set; }
+        public Layout<bool> KeepConnection { get; set; }
```
304NotModified: changes for example usage in this class

src/NLog/Targets/NetworkTarget.cs
```diff
@@ -104,6 +104,8 @@ public NetworkTarget()
             ConnectionCacheSize = 5;
             LineEnding = LineEndingMode.CRLF;
             OptimizeBufferReuse = GetType() == typeof(NetworkTarget);   // Class not sealed, reduce breaking changes
+            MaxQueueSize = 0;
+            KeepAliveTimeSeconds = 0;
```
304NotModified: note: because I had a null ref first (discussion later)

tests/NLog.UnitTests/Targets/EventLogTargetTests.cs
```diff
@@ -446,7 +445,7 @@ public void LogEntryWithDynamicEventIdAndCategory()
             int eventId = rnd.Next(1, short.MaxValue);
             int category = rnd.Next(1, short.MaxValue);
             var target = CreateEventLogTarget<EventLogTarget>("NLog.UnitTests" + Guid.NewGuid().ToString("N"), EventLogTargetOverflowAction.Truncate, 5000);
-            target.EventId = new SimpleLayout("${event-properties:EventId}");
+            target.EventId = "${event-properties:EventId}";
```
304NotModified: changes for example usage in this class

tests/NLog.UnitTests/Targets/NetworkTargetTests.cs
```diff
@@ -135,14 +137,15 @@ private void HappyPathTest(bool newLine, LineEndingMode lineEnding, params strin
...
             var target = new NetworkTarget();
-
-            Assert.True(target.KeepConnection);
-            Assert.False(target.NewLine);
-            Assert.Equal("\r\n", target.LineEnding.NewLineCharacters);
-            Assert.Equal(65000, target.MaxMessageSize);
-            Assert.Equal(5, target.ConnectionCacheSize);
-            Assert.Equal(0, target.MaxConnections);
-            Assert.Equal(0, target.MaxQueueSize);
+            //todo fix .ToValueSafe(logEventInfo)
```
304NotModified: todo

tools/MakeNLogXSD/XsdFileGenerator.cs
```diff
@@ -315,6 +315,9 @@ private static string GetXsdType(string apiTypeName, bool attribute)
             {
                 case "Layout":
                     return attribute ? "SimpleLayoutAttribute" : "Layout";
+             
+                case "NLog.Layouts.IntLayout":
```
304NotModified: todo

src/NLog/Targets/NetworkTarget.cs
```diff
@@ -289,12 +291,17 @@ protected override void Write(AsyncLogEventInfo logEvent)
 
             byte[] bytes = GetBytesToWrite(logEvent.LogEvent);
 
-            if (KeepConnection)
+            var networkTargetOverflowAction = OnOverflow.ToValueSafe(logEvent);
```
304NotModified: @snakefoot please your opinion on `ToValueSafe`. I don't like it.

But otherwise, I had null refs all the time.

I can't make `Layout<T>` as `struct` I guess - and also not sure if that would be a good idea if that was even possible

src/NLog/Layouts/Typed/Layout.cs
```diff
@@ -0,0 +1,337 @@
...
+            if (raw == null)
+            {
+                return true;
+            }
+
+            try
+            {
+                // todo could be better?
+                var objTypeCode = _typeCode;
+                switch (objTypeCode)
```
304NotModified: I dunno if we could improve this (less only separate method)

src/NLog/Layouts/Typed/EnumLayout.cs
304NotModified: don't think so

src/NLog/Targets/NetworkTarget.cs
304NotModified: idea:

- hide ToValue with explicit interface implementation, and rename to `.GetValue`. Need an extra interface for that.
- rename `ToValueSafe` to `GetValueOrDefault`

src/NLog/Layouts/Typed/Layout.cs
```diff
@@ -0,0 +1,337 @@
...
+                {
+                    //todo ugly :P
+                    case TypeCode.Boolean:
+                        {
+                            if (bool.TryParse(raw.ToString(), out var value1))
+                            {
+                                value = (T)(object)value1;
```
304NotModified: dunno how to fix this box-unbox operation. If T is a bool, then this is really meh

src/NLog/Targets/NetworkTarget.cs
304NotModified: @snakefoot any ideas on this?

snakefoot: Think the extension-method is a good way to handle that `Layout<T>` can be null. Maybe call it `RenderValue` and let the default value be optional-parameter so it is visible for the user. Should make it easy to guess what what will happen if null or unknown string-value.

304NotModified: Thanks!

src/NLog/Targets/EventLogTarget.cs
```diff
@@ -271,13 +271,7 @@ protected override void Write(LogEventInfo logEvent)
 
             EventLogEntryType entryType = GetEntryType(logEvent);
 
-            int eventId = 0;
-            string renderEventId = RenderLogEvent(EventId, logEvent);
```
304NotModified: @snakefoot please help! I removed this optimized RenderLogEvent call, and not sure how to replace it efficiently with `Layout<int>`,  etc

(I prefer to remove the whole RenderLogEvent method here and move the logic that works with `Layout<T>`)

304NotModified:
@snakefoot please review. I renamed a lot and I'm very happy with the result. I like to merge this soon

src/NLog/Targets/EventLogTarget.cs
snakefoot: RenderLogEvent skips allocation of StringBuilder for complex-Layout. In this situation performance is not that important. But if you plan to use it everywhere, then either performance-optimizations should be considered, or just put a warning-label on `Layout<T>` saying that it has a performance hit compared to `RenderLogEvent`

304NotModified: Yeah, so we have to make RenderLogEvent `Layout<T>` complaint. But I think that could be another PR, agree?

304NotModified: ow no, I will fix this!

src/NLog/Targets/Target.cs
```diff
@@ -719,6 +721,42 @@ protected string RenderLogEvent(Layout layout, LogEventInfo logEvent)
...
+                    return layout.RenderableToValue(logEvent, defaultValue);
+                }
+
+                if (TryGetCachedValue(layout, logEvent, out var value))
+                {
+                    return value;
+                }
+
+                return layout.RenderableToValue(logEvent, defaultValue); //todo
+                //using (var localTarget = ReusableLayoutBuilder.Allocate())
```
304NotModified: @snakefoot please your advise on this one! Thanks!

snakefoot: Well if you can dig out a raw-value then no need to render anything, but else you should provide the StringBuilder to RenderableToValue so it is available if needed.

304NotModified: I guess the performance issue is here:

```
 var text = _layout.Render(logEvent);
            if (TryParse(text, out var parsedValue))
            {
                return parsedValue;
            }
```

in TypedLayoutBase. But that code won't get called if the type supported IRawValue

snakefoot: Yes instead of calling `_layout.Render` then it should call the version that accepts a pre-allocated StringBuilder.

304NotModified: @snakefoot I think I got it. Please check (this PR or just 1714c3de57f4f41bd2b5765ad3a291cd90f9a532). 


I really like to merge this one. 

Thanks!

304NotModified:
I need I should drop the ValueConverter for the PropertyConverter

304NotModified:
should RenderableToValue => RenderToValue?

304NotModified:
//todo culture issue? See `//todo culture`

304NotModified:
adding tests, doesnt work well fro conversion between int and nullable int

304NotModified:
needs more work after some tests

304NotModified:
@snakefoot I think this is ready!

304NotModified:
Maybe I should use SetPropertyFromString? (so I could store "1,2,3" and render as HashSet<int>)?

src/NLog/Common/ValueConverter.cs
```diff
@@ -31,20 +31,21 @@
...
+using NLog.Internal;
 
+namespace NLog.Common
+{
     /// <summary>
-    /// Default implementation of <see cref="IPropertyTypeConverter"/>
+    /// Convert between types
     /// </summary>
-    internal class PropertyTypeConverter : IPropertyTypeConverter
+    internal class ValueConverter : IPropertyTypeConverter
```
snakefoot, line 43: Think it should keep its name `PropertyTypeConverter `

src/NLog/Layouts/Typed/IRenderable.cs
```diff
@@ -0,0 +1,48 @@
...
+// THE POSSIBILITY OF SUCH DAMAGE.
+// 
+
+namespace NLog.Layouts
+{
+    /// <summary>
+    /// Renderable to a typed value
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    public interface IRenderable<out T>
```
snakefoot, line 40: Is it needed as `public` ?

src/NLog/Layouts/Typed/Layout.cs
```diff
@@ -0,0 +1,315 @@
...
+        }
+
+        static Layout()
+        {
+            var type = typeof(T);
+            if (IsNullable(type))
+            {
+                var arg = type.GetGenericArguments()[0];
+
+#if !NETSTANDARD1_3 //todo fix
```
snakefoot, line 81: Want to use `TypeInfo` ?

src/NLog/Layouts/Typed/Layout.cs
```diff
@@ -0,0 +1,315 @@
...
+            {
+                _type = type;
+            }
+
+            _typeNamed = type.Name;
+        }
+
+        private static bool IsNullable(Type propertyType)
+        {
+            return propertyType.IsGenericType() && propertyType.GetGenericTypeDefinition() == typeof(Nullable<>);
```
snakefoot, line 97: Why not `Nullable.GetUnderlyingType(propertyType) ?? propertyType;` ?

src/NLog/Layouts/Typed/Layout.cs
```diff
@@ -0,0 +1,315 @@
...
+            {
+                value = default(T);
+                return false;
+            }
+
+            // We don't use DI here because of this will be called before DI setup and performance reasons
+            var cultureInfo = CultureInfo.CurrentCulture;
+            try
+            {
+                var convertedValue = ValueConverter.Instance.Convert(raw, _type, null, cultureInfo);
```
snakefoot, line 134: This doesn't look dependency injection friendly. Maybe fix it together with getting `Layout.FromString` working. Maybe extension method on LogFactory / ServiceRespository.

src/NLog/Layouts/Typed/Layout.cs
```diff
@@ -0,0 +1,315 @@
...
+
+            if (_layout == null)
+            {
+                rawValue = null;
+                return true;
+            }
+
+            if (_layout.TryGetRawValue(logEvent, out var raw))
+            {
+                var success = TryConvertRawToValue(raw, out var i);
```
snakefoot, line 217: NitPick. Maybe have a better variable-name than `i`.

src/NLog/Layouts/Typed/Layout.cs
```diff
@@ -0,0 +1,315 @@
...
+                rawValue = i;
```
snakefoot, line 218: Funny that value is also assigned when success = false.

src/NLog/Layouts/Typed/Layout.cs
```diff
@@ -0,0 +1,315 @@
...
+
+        /// <summary>
+        /// Render to value
+        /// </summary>
+        /// <param name="logEvent"></param>
+        /// <param name="reusableBuilder">if null, default layout render will be used</param>
+        /// <returns></returns>
+        internal T RenderToValueInternal(LogEventInfo logEvent, [CanBeNull] StringBuilder reusableBuilder)
+        {
+            if (_fixedValue)
```
snakefoot, line 243: Duplicate code with `TryGetRawValue`. Maybe call `TryGetRawValue` from this method, and only when that fails, then fallback to rendering.

src/NLog/Layouts/Typed/Layout.cs
```diff
@@ -0,0 +1,315 @@
...
+                return parsedValue;
+            }
+
+            InternalLogger.Warn("Parse {0} to {1} failed", text, _typeNamed);
+            return default(T);
+        }
+
+        private static bool TryConvertRawToValue(object raw, out T value)
+        {
+            if (raw == null)
```
snakefoot, line 275: Not sure I like that null-value automatically translate into default integer/enum-value. Might be working for reference-class-types.

src/NLog/Layouts/Typed/Layout.cs
```diff
@@ -0,0 +1,315 @@
...
+        private static bool TryGetFixedValue(Layout layout, out T value)
+        {
+            if (layout != null && layout is SimpleLayout simpleLayout && simpleLayout.IsFixedText)
+            {
+                if (!TryParse(simpleLayout.FixedText, out value))
+                {
+                    InternalLogger.Warn("layout with text '{0}' isn't an {1}", simpleLayout.FixedText, _typeNamed);
+                }
+
+                return true;
```
snakefoot, line 306: Strange that failing to parse a fixed-value, suddenly changes it into a valid type. Now `TryGetRawValue` will return `true`, but actually the fixed-value is invalid.

src/NLog/Targets/Target.cs
```diff
@@ -722,6 +724,39 @@ protected string RenderLogEvent(Layout layout, LogEventInfo logEvent)
...
+        /// <param name="layout">The layout.</param>
+        /// <param name="logEvent">The event info.</param>
+        /// <param name="defaultValue">The default value when <paramref name="layout"/> and/or <paramref name="logEvent"/> is <c>null</c></param>
+        /// <returns>Value representing log event.</returns>
+        protected T RenderLogEvent<T>([CanBeNull] Layout<T> layout, [CanBeNull] LogEventInfo logEvent, T defaultValue = default(T))
+        {
+            if (layout == null || logEvent == null)
+                return defaultValue;
+
+            if (OptimizeBufferReuse)
```
snakefoot, line 739: Think that `OptimizeBufferReuse` should be marked obsolete, and it should no longer be used anywhere.

src/NLog/Layouts/Typed/RenderableExtensions.cs
```diff
@@ -0,0 +1,74 @@
...
+    {
+        /// <summary>
+        /// Get the value, or if <paramref name="l"/>is <c>null</c>, the <paramref name="defaultValue"/>
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="l">The item to render the value from</param>
+        /// <param name="logEvent">Log event needed for rendering</param>
+        /// <param name="defaultValue">The default value when <paramref name="l"/>is <c>null</c></param>
+        /// <returns></returns>
+        public static T RenderToValueOrDefault<T>(this IRenderable<T> l, AsyncLogEventInfo logEvent, T defaultValue = default(T))
```
snakefoot, line 51: Think you should remove the version with `AsyncLogEventInfo`. And just keep for `LogEventInfo`.

src/NLog/Layouts/Typed/RenderableExtensions.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        /// <summary>
+        /// Get the value, or if <paramref name="l"/>is <c>null</c>, the <paramref name="defaultValue"/>
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="l">The item to render the value from</param>
+        /// <param name="logEvent">Log event needed for rendering</param>
+        /// <param name="defaultValue">The default value when <paramref name="l"/>is <c>null</c></param>
+        /// <returns></returns>
+        public static T RenderToValueOrDefault<T>(this IRenderable<T> l, LogEventInfo logEvent, T defaultValue = default(T))
```
snakefoot, line 64: Think the extension-method should be for `Layout<T>` (And not `IRenderable<T>`), then it is easier to understand for users.

304NotModified:
> I guess people will start doing `Layout<Encoding>` and `Layout<Uri>` how would that work?

That should work, but conversation from string to that type will be only in the xml reader. See also `Layout<Hashset<string>>` example. 

Maybe "only in the xml reader" isn't nice, there for thinking of https://github.com/NLog/NLog/pull/3656#issuecomment-635015441

src/NLog/Common/ValueConverter.cs
304NotModified: Well it's used for other Layouts and those aren't really properties. If I also create a IValueConverter, it makes maybe more sense. Anyway, it's internal so PropertyTypeConverter is also fine to me

src/NLog/Layouts/Typed/IRenderable.cs
304NotModified: for now yes, otherwise `RenderToValueOrDefault` won't work.

image

src/NLog/Layouts/Typed/Layout.cs
304NotModified: oops, that's was old. `#if` removed!

src/NLog/Layouts/Typed/Layout.cs
304NotModified: That's indeed better!

I made this first with TypeInfo, but switched to Type for .NET Standard 1.3

304NotModified: Fixed

src/NLog/Layouts/Typed/Layout.cs
304NotModified: This is indeed not DI friendly.

But altough I really like DI, im not sure if we need this here. Also this means we need to reference the LoggingConfiguration or LogFactory? We don't have those here (now).

src/NLog/Layouts/Typed/Layout.cs
304NotModified: names are important! Fixed!

src/NLog/Layouts/Typed/Layout.cs
304NotModified: Well we have to assign the out,. I assume here TryConvertRawToValue assigns the out correctly. Do you think this is wrong?

src/NLog/Layouts/Typed/Layout.cs
304NotModified: well this is needed if T is just a int, we can;t assign `null` to it?

src/NLog/Layouts/Typed/Layout.cs
304NotModified: this isn't good indeed. Will fix this

304NotModified: fixed

src/NLog/Layouts/Typed/RenderableExtensions.cs
304NotModified: removed

src/NLog/Layouts/Typed/RenderableExtensions.cs
304NotModified: good one!

304NotModified: That makes it also possible to make `IRenderable<T>` internal :)

src/NLog/Layouts/Typed/IRenderable.cs
304NotModified: fixed

src/NLog/Layouts/Typed/Layout.cs
304NotModified: Fixed,we needed an extra method because of T vs object

src/NLog/Targets/Target.cs
304NotModified: I don't know what to do here now :|

304NotModified:
@snakefoot added an example with `Layout<Encoding>`

src/NLog/Layouts/Typed/Layout.cs
304NotModified: ow there is already a LoggingConfiguration there, missed that

304NotModified: Fixed. What needs to be fixed in Layout.FromString?

304NotModified:
OK, we need "Fix specific conversion (taken from XML parser)"

src/NLog/Targets/Target.cs
snakefoot: I would stop checking the ``OptimizeBufferReuse`-property. And in a new PR change the property into being obsolete and by default always return true.

src/NLog/Layouts/Typed/Layout.cs
snakefoot: Trying to say that it should fail when converting null into an integer (Unless nullable).

src/NLog/Layouts/Typed/Layout.cs
snakefoot: Thinking something like this:

```
if (_layout.TryGetRawValue(logEvent, out var raw) && TryConvertRawToValue(raw, out var i))
{
    rawValue = i;
    return true;
}
```

Btw. might want to fix the variable name `i`.

src/NLog/Layouts/Typed/Layout.cs
snakefoot: I was thinking that SimpleLayout should delay the parsing until being initialized with a LoggingConfiguration (during Target-initialization). Same with this Layout. Should not be doing any parsing until having been initialized.

src/NLog/Targets/Target.cs
304NotModified: But why is that needed in this PR? Do I use OptimizeBufferReuse wrong here?

snakefoot: It is just because you are creating new code, that depends on a member-variable that is about to be marked as obsolete (and will default to always returning `true`). So just remove remove the `if`-statement.

304NotModified: As double check, with "remove the if-statement." - you mean remove the if condition and keep the body, isn't? (so essentially we will remove the else branch)

src/NLog/Common/ValueConverter.cs
304NotModified: reverted

src/NLog/Layouts/Typed/Layout.cs
304NotModified: fixed

304NotModified:
I think this is done!

src/NLog/Internal/Reflection/PropertyHelper.cs
```diff
@@ -103,21 +103,31 @@ internal static void SetPropertyFromString(object obj, string propertyName, stri
...
-                if (!TryNLogSpecificConversion(propertyType, value, configurationItemFactory, out var newValue))
+                if (!TryNLogSpecificConversion(unwrappedType, value, configurationItemFactory, out var newValue))
                 {
                     if (propInfo.IsDefined(_arrayParameterAttribute.GetType(), false))
                     {
                         throw new NotSupportedException($"Parameter {propertyName} of {objType.Name} is an array and cannot be assigned a scalar value.");
                     }
 
-                    propertyType = Nullable.GetUnderlyingType(propertyType) ?? propertyType;
+                    unwrappedType = Nullable.GetUnderlyingType(unwrappedType) ?? unwrappedType;
```
snakefoot, line 116: Maybe move this logic into `UnwrapTypedLayout()`

src/NLog/Internal/Reflection/PropertyHelper.cs
```diff
@@ -103,21 +103,31 @@ internal static void SetPropertyFromString(object obj, string propertyName, stri
...
+                          || TryTypeConverterConversion(unwrappedType, value, out newValue)))
+                    {
+                        newValue = Convert.ChangeType(value, unwrappedType, CultureInfo.InvariantCulture);
+                    }
+                }
+
+                // Wrap value again
+                if (isTypedLayout)
+                {
+                    newValue = WrapIntoTypedLayout(newValue, unwrappedType);
```
snakefoot, line 130: What happens when `unwrappedType` has changed from nullable to non-nullable ? Will the Layout always render non-nullable?

src/NLog/Layouts/Typed/Layout.cs
```diff
@@ -0,0 +1,350 @@
...
+                var convertedValue = converter.Convert(raw, Type, null, cultureInfo);
+                if (convertedValue is T goodValue)
+                {
+                    value = goodValue;
+                    return true;
+                }
+            }
+            catch (Exception e)
+            {
+                InternalLogger.Debug(e, "Conversion to type {0} failed", Type);
```
snakefoot, line 125: Maybe include the output of `raw`-value in the debug-output

src/NLog/Layouts/Typed/Layout.cs
```diff
@@ -0,0 +1,350 @@
...
+            }
+
+            var cultureInfo = CultureInfo.CurrentCulture;
+            try
+            {
+                var converter = ResolveService<IPropertyTypeConverter>() ?? PropertyTypeConverter.Instance;
+                var convertedValue = converter.Convert(raw, Type, null, cultureInfo);
+                if (convertedValue is T goodValue)
+                {
+                    value = goodValue;
```
snakefoot, line 119: Should cache the raw-value and the good-value. And when trying to convert the same raw-value again, then just return the cached good-value (instead of converting one more time). Will reduce the overhead of converting the same comma-seperated-string into a List-of-items.

snakefoot: To make it really nice. Then do like `FilePathLayout` where it skips StringBuilder.ToString() when it sees it contains the same raw-value as before.

src/NLog/Internal/Reflection/PropertyHelper.cs
304NotModified: The problem where are fixing here (this should be in code as comment)

When using reflection:

- default conversations in .net are problematic to nullables (so convert to int instead of int?)
- set a int value to a nullable int property is also problematic. So therefore the wrap again. 

Does this answer your questions?

src/NLog/Layouts/Typed/Layout.cs
304NotModified: I was afraid it could be large. But maybe that isn't important?

src/NLog/Layouts/Typed/Layout.cs
304NotModified: I like the cache idea, but I propose to do that in a follow-up PR. Im also convinced that's more your expertise :)

snakefoot: Like when people complete their task, and does some benchmarks. Remember you want to use this everywhere, so you should make sure the idea works.

src/NLog/Layouts/Typed/Layout.cs
snakefoot: Not when it fails.

src/NLog/Internal/Reflection/PropertyHelper.cs
snakefoot: No idea. I was just thinking that Nullable is a valid-return-value, but default-value of nullable-type is not the same.

src/NLog/Layouts/Typed/Layout.cs
304NotModified: If I add a benchmark, then to what should I compare to?

src/NLog/Layouts/Typed/Layout.cs
304NotModified: Will do otherwise a "left 1000" or something like that

src/NLog/Layouts/Typed/Layout.cs
304NotModified: Note, this isn't working yet "Will reduce the overhead of converting the same comma-seperated-string into a List-of-items.".  That's https://github.com/NLog/NLog/issues/3994

But agree, will look in the cache here

snakefoot:
Maybe a benchmark with something like this:

```c#
public class MyTarget : TargetWithLayout
{
       public bool LegacyMode { get; set; }
       public int LegacyInteger { get; set; }
       public Encoding LegacyEncoding { get; set; }
       public Layout<int> NewInteger { get; set; }
       public Layout<Encoding> NewEncoding { get; set; }

       protected override void Write(LogEventInfo logEvent)
       {
                 if (LegacyMode)
                 {
                       LegacyInteger.GetHashCode() + LegacyEncoding?.GetHashCode() ?? 0;
                 }
                 else
                 {
                       RenderLogEvent(NewInteger, logEvent).GetHashCode() + RenderLogEvent(NewEncoding, logEvent).GetHashCode() ?? +;
                 }
       }
}
```

With the following combinations in `NLog.config`:

| LegacyMode | LegacyInteger | LegacyEncoding | NewInteger | NewEncoding |
| -- | -- | -- | -- | -- |
| true | 1 | "UTF8" | |
| false |   |   | 1 | "UTF8" |
| false |   |   | "${gdc:integer}" | "${gdc:encoding}" |
| false |   |   | "${sequenceid}" | "UTF8" |

304NotModified:
Thanks! But what's the idea behind the hashcode?

snakefoot:
> Thanks! But what's the idea behind the hashcode?

I was hoping the compiler would not optimize away code in release-mode. But only trying.

src/NLog/Layouts/Typed/Layout.cs
```diff
@@ -0,0 +1,350 @@
...
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    public class Layout<T> : Layout, IRenderable<T>
+    {
+        // ReSharper disable StaticMemberInGenericType - this is safe, static ctor is called for every generic type
+        [NotNull] private static readonly Type Type;
+        [NotNull] private static readonly string TypeNamed;
+        // ReSharper restore StaticMemberInGenericType
+
+        [CanBeNull] private readonly Layout _layout;
```
snakefoot, line 57: Because this nested `Layout` cannot be discovered, then `Layout<T>` must announce (and check) whether it is ThreadAgnostic or MutableUnsafe.

snakefoot: Also need to override `PrecalculateBuilder` and call `_layout.PrecalculateBuilder` when necessary.

src/NLog/Internal/Reflection/PropertyHelper.cs
304NotModified: not possible, we need the prev value of unwrappedType above

src/NLog/Layouts/Typed/Layout.cs
304NotModified: Maybe we should add the same attributes as SimpleLayout? 
```c#
    [ThreadAgnostic]
    [ThreadSafe]
    [AppDomainFixedOutput]
    public class SimpleLayout : Layout, IUsesStackTrace
```

304NotModified: > Also need to override `PrecalculateBuilder` and call `_layout.PrecalculateBuilder` when necessary.

Could you please do a code proposal for this?

src/NLog/Layouts/Typed/Layout.cs
304NotModified: Added a cache. it as simple on (last used), and new update will overwrite old values. Hope you like it. Maybe need some locks for updating and reading the cache

src/NLog/Layouts/Typed/Layout.cs
snakefoot: Something like this:

```c#
       {
            base.InitializeLayout();
            ThreadAgnostic = _layout?.ThreadAgnostic ?? true;
            MutableUnsafe = _layout?.MutableUnsafe ?? false;
       }

       internal override void PrecalculateBuilder(LogEventInfo logEvent, StringBuilder target)
       {
           _layout?.PrecalculateBuilder(logEvent, target);
       }
```

src/NLog/Layouts/Typed/Layout.cs
```diff
@@ -0,0 +1,374 @@
...
+            return TryConvertTo(text, out value, defaultValue);
+        }
+
+        private object _cacheKey = null;
+        private T _cacheValue = default(T);
+        private bool _cacheResult = false;
+
+        private bool TryCache(object raw)
+        {
+            return _cacheKey != null && (ReferenceEquals(_cacheKey, raw) || _cacheKey.Equals(raw));
```
snakefoot, line 110: Maybe something like this:

```c#
        private bool TryGetCacheValue(object rawValue, out object resultValue, out bool resultValid)
        {
            var cacheKey = _cacheKey:
            var cacheValue = _cacheValue;
            var cacheValid = _cacheResult;
            if (!ReferenceEquals(cacheKey, null) && (ReferenceEquals(cacheKey, rawValue) || cacheKey.Equals(rawValue))
            {
                  resultValue = cacheValue;
                  resultValid = cacheValid;
                  return true;
            }
            resultValue = null;
            resultValid = false;
            return false;
       }
```

src/NLog/Layouts/Typed/Layout.cs
```diff
@@ -0,0 +1,374 @@
...
+            }
+
+            rawValue = default(T);
+            return false;
+        }
+
+
+        private bool TryConvertRawToValue(object raw, T defaultValue, out T value)
+        {
+            if (raw is T i)
```
snakefoot, line 280: Maybe a better variable-name than `i`

src/NLog/Layouts/Typed/Layout.cs
```diff
@@ -0,0 +1,374 @@
...
+        {
+            if (IsFixed)
+            {
+                rawValue = _fixedValue;
+                return true;
+            }
+
+            if (_layout == null)
+            {
+                rawValue = default(T);
```
snakefoot, line 257: Why not using `rawValue = defaultValue` ?

src/NLog/Layouts/Typed/Layout.cs
```diff
@@ -0,0 +1,374 @@
...
+                if (success)
+                {
+                    rawValue = value;
+                    return true;
+                }
+
+                InternalLogger.Warn("rawvalue isn't a {0} ", TypeNamed);
+            }
+
+            rawValue = default(T);
```
snakefoot, line 273: Why not using `rawValue = defaultValue` ?

src/NLog/Layouts/Typed/Layout.cs
```diff
@@ -0,0 +1,374 @@
...
+            if (_layout.TryGetRawValue(logEvent, out var raw))
+            {
+                var success = TryConvertRawToValue(raw, defaultValue, out var value);
+                if (success)
+                {
+                    rawValue = value;
+                    return true;
+                }
+
+                InternalLogger.Warn("rawvalue isn't a {0} ", TypeNamed);
```
snakefoot, line 270: Maybe include raw-value in the output (maybe truncated).

src/NLog/Layouts/Typed/Layout.cs
304NotModified: is !ReferenceEquals(cacheKey, null) better than `cacheKey != null`?

src/NLog/Layouts/Typed/Layout.cs
304NotModified: Fixed

src/NLog/Layouts/Typed/Layout.cs
snakefoot: Think it compiles to the same. Just thought it would match the other `RereferenceEquals` :)

src/NLog/Layouts/Typed/Layout.cs
304NotModified: yes will do :)

src/NLog/Layouts/Typed/Layout.cs
304NotModified: I could also rename the current TryCache to IsCacheHit. Or do you really prefer the (longer) TryGetCacheValue?

src/NLog/Targets/Target.cs
```diff
@@ -722,6 +724,39 @@ protected string RenderLogEvent(Layout layout, LogEventInfo logEvent)
...
+            {
+                if (layout.IsFixed)
+                {
+                    return layout.RenderToValueOrDefault(logEvent, defaultValue);
```
snakefoot, line 743: Should just call `layout.RenderToValue(logEvent, default);`

src/NLog/Targets/Target.cs
```diff
@@ -722,6 +724,39 @@ protected string RenderLogEvent(Layout layout, LogEventInfo logEvent)
...
+                    return value;
+                }
+
+                using (var localTarget = ReusableLayoutBuilder.Allocate())
+                {
+                    return layout.RenderToValueInternal(logEvent, localTarget.Result, default(T));
+                }
+            }
+
+            return layout.RenderToValueOrDefault(logEvent, defaultValue);
```
snakefoot, line 757: Should just call `layout.RenderToValue(logEvent, defaultValue);` (Have already checked layout is not-null)

src/NLog/Targets/Target.cs
```diff
@@ -722,6 +724,39 @@ protected string RenderLogEvent(Layout layout, LogEventInfo logEvent)
...
+                return defaultValue;
+
+            if (OptimizeBufferReuse)
+            {
+                if (layout.IsFixed)
+                {
+                    return layout.RenderToValueOrDefault(logEvent, defaultValue);
+                }
+
+                if (TryGetCachedValue(layout, logEvent, out var value))
```
snakefoot, line 746: This code-path handles layoutrenderer like `${threadid}` when target is wrapped in AsyncWrapperTarget. But `Layout<int>` is not allowed to cast the cached `string` value directly to int-value. Instead this logic should be moved into  `Layout<T>`-class where it can also test if TryGetRawValue can be used with needing help from `ReusableLayoutBuilder`.

src/NLog/Targets/Target.cs
```diff
@@ -722,6 +724,39 @@ protected string RenderLogEvent(Layout layout, LogEventInfo logEvent)
...
+                {
+                    return value;
+                }
+
+                using (var localTarget = ReusableLayoutBuilder.Allocate())
+                {
+                    return layout.RenderToValueInternal(logEvent, localTarget.Result, default(T));
```
snakefoot, line 753: When finally ending up at using `ReusableLayoutBuilder` to render into string-output, and then convert into the typed-value, then it would be nice if the caching logic could avoid performing StringBuilder.ToString() but instead check if the _cachedKey (as string) matched the contents of StringBuilder using the EqualTo-method (So returned the cached converted value without first allocating a temporary string-object)

src/NLog/Targets/Target.cs
304NotModified: good one, but then I also need `     IRenderable<T> renderable = l;`. This makes the code more complex. Is that really worth it?

src/NLog/Targets/Target.cs
304NotModified: Is this https://github.com/NLog/NLog/issues/3994?

src/NLog/Targets/Target.cs
304NotModified: I prefer to do nice to haves after merge. This PR is too long active

src/NLog/Targets/Target.cs
snakefoot: Nope it is bug. Try using `${threadid}` as EventId in the EventLog-target and wrap it in AsyncTargetWrapper. Then it should throw a nice cast-exception

src/NLog/Targets/Target.cs
snakefoot: Why is that cast necessary to call a method. You have the full type `Layout<T>`

src/NLog/Layouts/Typed/Layout.cs
snakefoot: I like TryGet-method-pattern (returning bool and having out-parameter).

src/NLog/Layouts/Typed/Layout.cs
```diff
@@ -145,7 +145,7 @@ private bool TryConvertTo(object raw, out T value, T defaultValue)
             }
             catch (Exception e)
             {
-                InternalLogger.Debug(e, "Conversion to type {0} failed. Value: {1}", Type, raw.ToString().Left(1000));
+                InternalLogger.Debug(e, "Conversion to type {0} failed. Value: '{1}'", Type, raw.ToString().Left(1000));
```
snakefoot, line 148: Maybe have conditional `ToString()` or can raw never be null ?

304NotModified: null check of raw is in the beginning of the method. :)

src/NLog/Targets/Target.cs
304NotModified: Hi @snakefoot,

I think this in the real blocker for this pr (others comments could be resolved without much hassle)

I don't really understand the issue here. Could you please help me with this by writing an unit test or if it's possible, the fix? I really like to merge this PR!

Feel free to push to this branch.

snakefoot: It is about correct handling of LayoutRenderers that are not threadagnostic (Like `${threadid}`). And using them with `Layout<int>` in a target that are wrapped with AsyncWrapper.

304NotModified: Please help me with an unit test for this. I'm very limited with NLog these days and haven't really time to puzzle to reproduce this.

src/NLog/Config/ServiceRepositoryExtensions.cs
```diff
@@ -40,19 +40,21 @@ namespace NLog.Config
...
-        internal static ServiceRepository GetServiceProvider(this LoggingConfiguration loggingConfiguration)
+        [NotNull]
+        internal static IServiceProvider GetServiceResolver([CanBeNull] this LoggingConfiguration loggingConfiguration)
         {
             return loggingConfiguration?.LogFactory?.ServiceRepository ?? LogManager.LogFactory.ServiceRepository;
         }
 
-        internal static T ResolveService<T>(this ServiceRepository serviceProvider, bool ignoreExternalProvider = true) where T : class
+        [CanBeNull]
+        public static T ResolveService<T>(this IServiceProvider serviceProvider) where T : class
```
304NotModified, line 50: unneeded change?

snakefoot: Nothing wrong with adding `[CanBeNull]`, but you should not remove the parameter `bool ignoreExternalProvider = true`

src/NLog/Layouts/Typed/Layout.cs
304NotModified: applied

src/NLog/Layouts/Typed/Layout.cs
304NotModified: done

src/NLog/Targets/Target.cs
304NotModified: Tis is not possible, as RenderToValue is on `IRenderable<T>` and `Layout<T> ` implements it explictly: ` T IRenderable<T>.RenderToValue(LogEventInfo logEvent, T defaultValue)`.

This is by design, so that others prefer `RenderToValueOrDefault` over `RenderToValue`

src/NLog/Targets/Target.cs
304NotModified: same as above, not possible (without cast).

src/NLog/Targets/Target.cs
304NotModified: @snakefoot please suggest which unit test and how, or feel free to add an commit to this branch :)

I don't see how I could change the current tests (which is using RenderToValueOrDefault), to the asyncwrapper

src/NLog/Targets/Target.cs
304NotModified: Feel free to send a PR for this after this one get merged

snakefoot:
Here is a unit-test that should work:

```c#
        [Fact]
        public void TypedLayoutTargetTest()
        {
            LoggingConfiguration c = XmlLoggingConfiguration.CreateFromXmlString(@"
            <nlog>
                <extensions>
                    <add type='" + typeof(MyTypedLayoutTarget).AssemblyQualifiedName + @"' />
                </extensions>

                <targets>
                    <target type='MyTypedLayoutTarget' name='myTarget'
                        byteProperty='42' 
                        int16Property='42' 
                        int32Property='42' 
                        int64Property='42000000000' 
                        stringProperty='foobar'
                        boolProperty='true'
                        doubleProperty='3.14159'
                        floatProperty='3.14159'
                        enumProperty='Value3'
                        flagsEnumProperty='Value1,Value3'
                        encodingProperty='utf-8'
                        cultureProperty='en-US'
                        typeProperty='System.Int32'
                        uriProperty='http://nlog-project.org'
                        lineEndingModeProperty='default'
                        />
                </targets>
            </nlog>");

            var nullEvent = LogEventInfo.CreateNullEvent();
            var myTarget = c.FindTargetByName("myTarget") as MyTypedLayoutTarget;
            Assert.NotNull(myTarget);
            Assert.Equal((byte)42, myTarget.ByteProperty.RenderToValueOrDefault(nullEvent));
            Assert.Equal((short)42, myTarget.Int16Property.RenderToValueOrDefault(nullEvent));
            Assert.Equal(42, myTarget.Int32Property.RenderToValueOrDefault(nullEvent));
            Assert.Equal(42000000000L, myTarget.Int64Property.RenderToValueOrDefault(nullEvent));
            Assert.Equal("foobar", myTarget.StringProperty.RenderToValueOrDefault(nullEvent));
            Assert.True(myTarget.BoolProperty.RenderToValueOrDefault(nullEvent));
            Assert.Equal(3.14159, myTarget.DoubleProperty.RenderToValueOrDefault(nullEvent));
            Assert.Equal(3.14159f, myTarget.FloatProperty.RenderToValueOrDefault(nullEvent));
            Assert.Equal(MyEnum.Value3, myTarget.EnumProperty.RenderToValueOrDefault(nullEvent));
            Assert.Equal(MyFlagsEnum.Value1 | MyFlagsEnum.Value3, myTarget.FlagsEnumProperty.RenderToValueOrDefault(nullEvent));
            Assert.Equal(Encoding.UTF8, myTarget.EncodingProperty.RenderToValueOrDefault(nullEvent));
            Assert.Equal("en-US", myTarget.CultureProperty.RenderToValueOrDefault(nullEvent).Name);
            Assert.Equal(typeof(int), myTarget.TypeProperty.RenderToValueOrDefault(nullEvent));
            Assert.Equal(new Uri("http://nlog-project.org"), myTarget.UriProperty.RenderToValueOrDefault(nullEvent));
            Assert.Equal(LineEndingMode.Default, myTarget.LineEndingModeProperty.RenderToValueOrDefault(nullEvent));
        }

        [Fact]
        public void TypedLayoutTargetAsyncTest()
        {
            // Arrange
            LogFactory logFactory = new LogFactory();
            LoggingConfiguration c = XmlLoggingConfiguration.CreateFromXmlString(@"
            <nlog>
                <extensions>
                    <add type='" + typeof(MyTypedLayoutTarget).AssemblyQualifiedName + @"' />
                </extensions>

                <variable name='value3' value='Value3' />

                <targets async='true'>
                    <target type='MyTypedLayoutTarget' name='myTarget'
                        byteProperty='42' 
                        int16Property='42' 
                        int32Property='${threadid}' 
                        int64Property='${sequenceid}' 
                        stringProperty='${processname}'
                        boolProperty='true'
                        doubleProperty='3.14159'
                        floatProperty='3.14159'
                        enumProperty='${var:value3}'
                        flagsEnumProperty='Value1,Value3'
                        encodingProperty='utf-8'
                        cultureProperty='en-US'
                        typeProperty='System.Int32'
                        uriProperty='http://nlog-project.org'
                        lineEndingModeProperty='default'
                        />
                </targets>

                <rules>
                    <logger minlevel='trace' writeTo='myTarget' />
                </rules>
            </nlog>", logFactory);
            logFactory.Configuration = c;

            // Act
            var logger = logFactory.GetLogger(nameof(TypedLayoutTargetAsyncTest));
            var logEvent = new LogEventInfo(LogLevel.Info, null, "Hello");
            logger.Log(logEvent);
            logFactory.Flush();

            // Assert
            Assert.Equal(System.Threading.Thread.CurrentThread.ManagedThreadId, logEvent.Properties[nameof(MyTypedLayoutTarget.Int32Property)]);
        }

        [Target("MyTypedLayoutTarget")]
        public class MyTypedLayoutTarget : Target
        {
            public Layout<byte> ByteProperty { get; set; }

            public Layout<short> Int16Property { get; set; }

            public Layout<int> Int32Property { get; set; }

            public Layout<long> Int64Property { get; set; }

            public Layout<string> StringProperty { get; set; }

            public Layout<bool> BoolProperty { get; set; }

            public Layout<double> DoubleProperty { get; set; }

            public Layout<float> FloatProperty { get; set; }

            public Layout<MyEnum> EnumProperty { get; set; }

            public Layout<MyFlagsEnum> FlagsEnumProperty { get; set; }

            public Layout<Encoding> EncodingProperty { get; set; }

            public Layout<CultureInfo> CultureProperty { get; set; }

            public Layout<Type> TypeProperty { get; set; }

            public Layout<Uri> UriProperty { get; set; }

            public Layout<LineEndingMode> LineEndingModeProperty { get; set; }

            protected override void Write(LogEventInfo logEvent)
            {
                logEvent.Properties[nameof(ByteProperty)] = RenderLogEvent(ByteProperty, logEvent);
                logEvent.Properties[nameof(Int16Property)] = RenderLogEvent(Int16Property, logEvent);
                logEvent.Properties[nameof(Int32Property)] = RenderLogEvent(Int32Property, logEvent);
                logEvent.Properties[nameof(Int64Property)] = RenderLogEvent(Int64Property, logEvent);
                logEvent.Properties[nameof(StringProperty)] = RenderLogEvent(StringProperty, logEvent);
                logEvent.Properties[nameof(BoolProperty)] = RenderLogEvent(BoolProperty, logEvent);
                logEvent.Properties[nameof(DoubleProperty)] = RenderLogEvent(DoubleProperty, logEvent);
                logEvent.Properties[nameof(FloatProperty)] = RenderLogEvent(FloatProperty, logEvent);
                logEvent.Properties[nameof(EnumProperty)] = RenderLogEvent(EnumProperty, logEvent);
                logEvent.Properties[nameof(FlagsEnumProperty)] = RenderLogEvent(FlagsEnumProperty, logEvent);
                logEvent.Properties[nameof(EncodingProperty)] = RenderLogEvent(EncodingProperty, logEvent);
                logEvent.Properties[nameof(CultureProperty)] = RenderLogEvent(CultureProperty, logEvent);
                logEvent.Properties[nameof(TypeProperty)] = RenderLogEvent(TypeProperty, logEvent);
                logEvent.Properties[nameof(UriProperty)] = RenderLogEvent(UriProperty, logEvent);
                logEvent.Properties[nameof(LineEndingModeProperty)] = RenderLogEvent(LineEndingModeProperty, logEvent);
            }
        }
```

304NotModified:
Thanks for the tests! There is another issue here, a lot of "try" methods in properyhelper.cs are throwing exceptions...

304NotModified:
fixed, but not really happy with the code

snakefoot:
Maybe always perform initial parse using SimpleLayout when property is type `Layout<T>`. 

Then modify the Layout<T>-constructor / conversion so it recognizes SimpleLayout, and when FixedText it converts it automatically the string to the fixed-type-value.

src/NLog/Targets/Target.cs
304NotModified: Fixed with the last commit (this is the same issue, right?)

Thanks, great I understand now the issue and it has been fixed :)

304NotModified:
> Maybe always perform initial parse using SimpleLayout when property is type `Layout<T>`.

This cannot be done, as #3994 is then needed. But it's a good idea to do with #3994. The parsing in propertyhelper is more enhanced than in `Layout<T>` (I will call `Layout<T>` for now TypedLayout)

> Then modify the Layout-constructor / conversion so it recognizes SimpleLayout,

To be clear, the TypedLayout contructor has support for SimpleLayouts?

```c#
    public class Layout<T> : Layout, IRenderable<T>
    {
        // ReSharper disable StaticMemberInGenericType - this is safe, static ctor is called for every generic type
        [NotNull] private static readonly Type Type;
        [NotNull] private static readonly string TypeNamed;
        // ReSharper restore StaticMemberInGenericType

        [CanBeNull] private readonly Layout _layout;
        private readonly T _fixedValue;

        /// <inheritdoc />
        public Layout(T value)
        {
            _fixedValue = value;
            IsFixed = true;
            _layout = null;
        }

        /// <inheritdoc />
        public Layout(Layout layout)
        {
            IsFixed = TryGetFixedValue(layout, out _fixedValue);
            _layout = layout;
        }


```

snakefoot:
I was thinking that Layout<T> should not make use of the PropertyHelper to perform value-conversion, but should do its own conversion from string / SimpleLayout / Layout. / etc.

304NotModified:
> I was thinking that `Layout<T>` should not make use of the PropertyHelper to perform value-conversion, but should do its own conversion from string / SimpleLayout / Layout. / etc.
> 
> Yes I can see `Layout<T>` already handles Fixed-value. Just also need to handle value-conversion when loading from config, instead of dependending on help from PropertyHelper to convert. 

The conversion needs to be extracted of the propertyhelper, but this is a big change and this is #3994


There is now also an issue, as the enum string value won't work `enumProperty='${var:value3}'`, as this currently needs the propertyhelper parse logic (so also   #3994)

> Would also make `Layout<T>` more predictable, that its conversion support doesn't change when loading from config (Compared to assigning from code).

True, but this PR is large enough and after merge, this is the next step.

I also see we missing a lot of asserts :'( for TypedLayoutTargetAsyncTest, so that one isn't completed

src/NLog/Internal/Abstractions/IStringValueRenderer.cs
```diff
@@ -36,6 +36,9 @@ namespace NLog.Internal
     /// <summary>
     /// Supports rendering as string value with limited or no allocations (preferred)
     /// </summary>
+    /// <remarks>
+    /// Implementors must have the [AppDomainFixedOutput] attribute
```
snakefoot, line 40: Implementors must **NOT** have the `[AppDomainFixedOutput]` attribute

tests/NLog.UnitTests/ApiTests.cs
```diff
@@ -183,7 +183,7 @@ public void IStringValueRenderer_AppDomainFixedOutput_Attribute_NotRequired()
                 if (typeof(NLog.Internal.IStringValueRenderer).IsAssignableFrom(type) && !type.IsInterface)
                 {
                     var appDomainFixedOutputAttribute = type.GetCustomAttribute<AppDomainFixedOutputAttribute>();
-                    Assert.True(ReferenceEquals(appDomainFixedOutputAttribute, null), $"{type.ToString()} should not implement IStringValueRenderer");
+                    Assert.True(ReferenceEquals(appDomainFixedOutputAttribute, null), $"{type.ToString()} cannot implement IStringValueRenderer as it isn't marked as [AppDomainFixedOutput]");
```
snakefoot, line 186: Should not implement `IStringValueRenderer` because it is marked as `[AppDomainFixedOutput]`

src/NLog/Targets/Target.cs
```diff
@@ -755,6 +790,18 @@ private static bool TryGetCachedValue(Layout layout, LogEventInfo logEvent, out
             return false;
         }
 
+        private static bool TryGetCachedValue<T>(Layout<T> layout, LogEventInfo logEvent, out T value)
+        {
+            if ((!layout.ThreadAgnostic || layout.MutableUnsafe) && logEvent.TryGetCachedLayoutValue(layout, out var value2))
```
snakefoot, line 795: There is a bug here, where it performs lookup using `Layout<T>`-reference, but the `Layout<T>.PrecalculateBuilder` stores the reference to the inner-layout in the `LogEventInfo._layoutCache`

snakefoot: Maybe move this method (and the above method) into the individual Layout-classes. Maybe consider making it a virtual-method and override it in `Layout<T>`-implementation.

snakefoot: Actually after further thought then this is a great idea. I have for a long time been worried about storing the raw-value in the LogEventInfo. But actually if `Layout<T>` promise that it will only store values of type `<T>` then it will be a great optimization. Ofcourse right now the current `PrecalculateBuilder´ is not up for the task, but that can be a different PR.

snakefoot:
Think it is interesting to see the side-effects on the code by your implementation of FileTarget-Encoding. The code doesn't get simpler by having everything Layoutable. And one also have to be careful, as it doesn't make sense to have different encodings for each LogEvent in the same file (Ex. for FileTarget then the encoding should only be "rendered" when initially creating the FileAppender, and be cached on the FileAppender).

Not sure I'm a big fan of making everything Layoutable, as the ability to provide a simple and predictable behavior suddenly becomes much harder. Still think that NLog Layouts can provide a lot of fancy and almost magical behavior, but it can be a difficult sword to wield (Ex. #3579)

304NotModified:
> Think it is interesting to see the side-effects on the code by your implementation of FileTarget-Encoding. The code doesn't get simpler by having everything Layoutable.

true, but it isn't that much harder. You have also compile time insurance. 



> as it doesn't make sense to have different encodings for each LogEvent in the same file

Agree on that, that it's maybe render too soon. But maybe this is a bad example indeed for dynamic. Anyway - layoutable doesn't mean dynamic, we could also render it on target initialize. 



> Not sure I'm a big fan of making everything Layoutable, as the ability to provide a simple and predictable behavior suddenly becomes much harder

There have been many questions about this, but maybe everthing is not needed. Also good to now, some are already layoutable but a bit hidden. 

> Maybe `Layout<T>` should operate in two modes:

I think the target should decide when to "refresh" the rendering, e.g. on writing log event, on a batch, or on target init.

snakefoot:
> I think the target should decide when to "refresh" the rendering, e.g. on writing log event, on a batch, or on target init.

I agree that is why I suggest that `Layout<T>` supports both modes out of the box:
- `Layout<T>.Value`-property when just want the value from targe-init
- `Layout<T>.RenderValue(LogEventInfo)`-method when wanting a refreshed value for every logevnt.

snakefoot:
> and isn't `Layout<T>.InitValue` just `Layout<T>.RenderValue(LogEventInfo.CreateNullEvent());` ?

Yep. But cached on the `Layout<T>`-object so only rendered once on target-init.

> I have difficulties to finalize this PR (too much discussion)

If you can get to a point where you can commit (without any breaking changes), then I don't mind implementing my ideas afterwards.

304NotModified:
@snakefoot what's the expected value here?



        [Fact]
        public void Equals_Dynamic_NotEqualTest()
        {
            // Arrange
            var layout1 = new Layout<int>("${event-properties:prop1}");
            var layout2 = new Layout<int>("${event-properties:prop1}");

            // Act
            var equal = layout1.Equals((object)layout2);

            // Assert
            // ?
        }

snakefoot:
Instead of using FileTarget.Encoding as demonstration of this feature. Could it not be `WebServiceTarget.Url` ? (Revert changes to FileTarget.Encoding)

snakefoot:
> @snakefoot what's the expected value here?

I expect the Equals to be impelmented like this:

```c#
        public override bool Equals(object obj)
        {
            return obj is Layout<T> other && object.Equals(_layout, other._layout) && object.Equals(_fixedValue, other._fixedValue);
        }
```

304NotModified:
I already reverted all filetarget stuff! :)

304NotModified:
> I expect the Equals to be impelmented like this

So true of false? ;)

snakefoot:
> So true of false? ;)

I expect "true" when SimpleLayout also adds override of object.Equals

304NotModified:
> Could it not be `WebServiceTarget.Url` ?

WebServiceTarget.Url as Layout/String is fine to me and it's maybe not the most stable target to test. What about UrlEncodeLayoutRendererWrapper?

304NotModified:
@snakefoot  I also sealed SimpleLayout, agree on that? (result for equality, Although there are other solutions)

snakefoot:
>@snakefoot I also sealed SimpleLayout, agree on that? 

No oppinion. I guess as many types should be sealed by default, and then one should open up on valid request.

snakefoot:
Btw. also think that having `Layout<ISet<string>>` is bad for the NLog-API. I like Layout<T> for simple-types, but for complex types it will generate confusion (Manually updating the string-set af having loaded from Config-Variable, will those changes disappear on next render value).

304NotModified:
> Btw. also think that having `Layout<ISet<string>>` is bad for the NLog-API. I like `Layout<T>` for simple-types (preferable immutable), but for complex types it will generate confusion (Ex. Manually updating the string-set after having loaded from Config-Variable, will those changes disappear on next render value?).

But how could we support lists/sets then?  I think `Layout<List<string>>` should be supported as least.

snakefoot:
> But how could we support lists/sets then? I think `Layout<List<string>>` should be supported as least.

For me the nightmare of trying to support mutable-types outweights the benefit. I'm fine with only supporting simple immutable types, and keeping thing simple and predictable.

snakefoot:
Maybe one could implement something fancy where `LayoutCollection<IList<string>>` is very aware that it owns a mutable collection type.

So RenderToValue will return an AsReadOnly-version that will explode when trying to modify. Instead `LayoutCollection` will have the modifies like `Add` / `Remove` / `Clear`. If calling the modifiers then the `LayoutCollection` will change its inner state to a fixed value (and stop being dynamic).

snakefoot:
Yet another alternative would be that `Layout<T>` sees that T is a collection, and instead of returning the real-collection, then it returns a special collection-wrapper. The collection-wrapper will then react to add / remove / clear-methods being called, and will explode with runtime error. Telling the user that to modify the inner value of `Layout<T>` then one must reassign a new collection-value. This would require one to create collection wrappers for the standard interfaces like ISet, IDictionary, IList etc.

snakefoot:
Regarding override Equals then I think SimpleLayout should not override Equals or GetHashCode.

And `Layout<T>` should only allow value comparison when holding Fixed-value. Else it should be object-reference.

This will ensure LogEventInfo.LayoutCache will work correctly, and the overhead for maintaining the cache is minimal.

PR #3660, opened 2019-10-31
ObjectHandleSerializer - GetObjectData should have SerializationFormatter=true

snakefoot:
See also: https://docs.microsoft.com/en-us/dotnet/api/system.runtime.serialization.iserializable.getobjectdata


304NotModified:
Thanks!

PR #3662, opened 2019-11-03
FileTarget - Extra logging when FileName Layout returns empty filename

snakefoot:
Note merge to master


304NotModified:
👍  thanks!

PR #3666, opened 2019-11-04
Refactor resolve & register for DI

304NotModified:
- Resolve easier
-  refactor registering, no unregister with `null`, `null` not allowed (update: partly now allowed)
- remove duplicate fields for IValueFormatter
 - Improve IServiceRepository injection into ConfigurationItemFactory
- note: added `TODO DI` comments

### DIFF
--- a/src/NLog/Config/ConfigurationItemFactory.cs
+++ b/src/NLog/Config/ConfigurationItemFactory.cs
@@ -31,6 +31,8 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
+using JetBrains.Annotations;
+
 namespace NLog.Config
 {
     using System;
@@ -56,7 +58,7 @@ namespace NLog.Config
     {
         private static ConfigurationItemFactory _defaultInstance;
 
-        private readonly IServiceResolver _serviceResolver;
+        [NotNull] private readonly IServiceRepository _serviceResolver;
         private readonly List<IFactory> _allFactories;
         private readonly Factory<Target, TargetAttribute> _targets;
         private readonly Factory<Filter, FilterAttribute> _filters;
@@ -80,9 +82,9 @@ namespace NLog.Config
         {
         }
 
-        internal ConfigurationItemFactory(IServiceResolver serviceResolver, ConfigurationItemFactory globalDefaultFactory, params Assembly[] assemblies)
+        internal ConfigurationItemFactory([NotNull] IServiceRepository serviceResolver, ConfigurationItemFactory globalDefaultFactory, params Assembly[] assemblies)
         {
-            _serviceResolver = serviceResolver;
+            _serviceResolver = serviceResolver ?? throw new ArgumentNullException(nameof(serviceResolver));
             _targets = new Factory<Target, TargetAttribute>(serviceResolver, globalDefaultFactory?._targets);
             _filters = new Factory<Filter, FilterAttribute>(serviceResolver, globalDefaultFactory?._filters);
             _layoutRenderers = new LayoutRendererFactory(serviceResolver, globalDefaultFactory?._layoutRenderers);
@@ -192,8 +194,8 @@ namespace NLog.Config
         [Obsolete("Instead use LogFactory.ServiceRepository.ResolveInstance(typeof(IJsonConverter)). Marked obsolete on NLog 5.0")]
         public IJsonConverter JsonConverter
         {
-            get => _serviceResolver?.ResolveService<IJsonConverter>();
-            set => (_serviceResolver as IServiceRepository)?.RegisterJsonConverter(value);
+            get => _serviceResolver.ResolveService<IJsonConverter>();
+            set => _serviceResolver.RegisterJsonConverter(value);
         }
 
         /// <summary>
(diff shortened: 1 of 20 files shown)


304NotModified:
@snakefoot please review, thanks!

src/NLog/Config/ConfigurationItemFactory.cs
```diff
@@ -192,8 +194,8 @@ internal LayoutRendererFactory GetLayoutRenderers()
         [Obsolete("Instead use LogFactory.ServiceRepository.ResolveInstance(typeof(IJsonConverter)). Marked obsolete on NLog 5.0")]
         public IJsonConverter JsonConverter
         {
-            get => _serviceResolver.ResolveJsonConverter();
-            set => (_serviceResolver as IServiceRepository)?.RegisterJsonConverter(value);
+            get => _serviceResolver.ResolveService<IJsonConverter>();
```
snakefoot: Missing fallback to static instance?

src/NLog/LayoutRenderers/LayoutRenderer.cs
```diff
@@ -304,5 +310,15 @@ public static void Register(string name, Func<LogEventInfo, LoggingConfiguration
             
             ConfigurationItemFactory.Default.GetLayoutRenderers().RegisterFuncLayout(name, layoutRenderer);
         }
+
+        /// <summary>
+        /// Resolve from DI
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <returns></returns>
+        protected T Resolve<T>() where T : class
```
snakefoot: Think I will skip this, until we have figured out a good way to do this.

src/NLog/Layouts/Layout.cs
```diff
@@ -390,5 +390,16 @@ internal virtual bool TryGetRawValue(LogEventInfo logEvent, out object rawValue)
             rawValue = null;
             return false;
         }
+
+        /// <summary>
+        /// Resolve from DI
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <returns></returns>
+        protected T Resolve<T>() where T : class
```
snakefoot: Think I will skip this, until we have figured out a good way to do this.

src/NLog/Targets/Target.cs
```diff
@@ -757,5 +757,15 @@ public static void Register(string name, Type targetType)
             ConfigurationItemFactory.Default.Targets
                 .RegisterDefinition(name, targetType);
         }
+
+        /// <summary>
+        /// Resolve from DI
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <returns></returns>
+        protected T Resolve<T>() where T : class
```
snakefoot: Think I will skip this, until we have figured out a good way to do this.

src/NLog/LayoutRenderers/LayoutRenderer.cs
```diff
@@ -49,12 +49,18 @@ public abstract class LayoutRenderer : ISupportsInitialize, IRenderable, IDispos
...
 
         /// <summary>
         /// Gets the logging configuration this target is part of.
         /// </summary>
         protected LoggingConfiguration LoggingConfiguration { get; private set; }
 
+        /// <summary>
+        /// Value formatter
+        /// </summary>
+        protected IValueFormatter ValueFormatter => _valueFormatter ?? (_valueFormatter = Resolve<IValueFormatter>());
```
snakefoot: Missing fallback to MessageTemplates.ValueFormatter.Instance

src/NLog/Config/ConfigurationItemFactory.cs
304NotModified: It's added already to the container?

src/NLog/LayoutRenderers/LayoutRenderer.cs
304NotModified: It's in the container, and you can register `null`

src/NLog/Config/ConfigurationItemFactory.cs
304NotModified: So intentional. No need for a default here

src/NLog/LayoutRenderers/LayoutRenderer.cs
304NotModified: What isn't good about this?

304NotModified: I was think to:

support Ctor injection and the Resolve method for Layouts/target etc. The user could choose. For the NLog components we will use the Resolve Method.

src/NLog/Config/ConfigurationItemFactory.cs
```diff
@@ -202,8 +204,8 @@ public IJsonConverter JsonConverter
         [Obsolete("Instead use LogFactory.ServiceRepository.ResolveInstance(typeof(IValueFormatter)). Marked obsolete on NLog 5.0")]
         public IValueFormatter ValueFormatter
         {
-            get => _serviceResolver.ResolveValueFormatter();
-            set => (_serviceResolver as IServiceRepository)?.RegisterValueFormatter(value);
+            get => _serviceResolver.ResolveService<IValueFormatter>() ?? MessageTemplates.ValueFormatter.Instance;
```
snakefoot: If resolver is always right, then you can remove the fallback.

304NotModified: good one, fixed

PR #3667, opened 2019-11-04
Tryout DI ctor injection (WIP)

304NotModified:


snakefoot:
Like the idea of being able to register a type to have constructor-dependency-injection.

But don't think it should be enabled by default. Instead it could be decided when explicitly registering the type, or the type has a special class-attribute.

Ex. we have constructors that takes Name as parameter.

Ex. We could have cyclic loop if type-constructor takes its own type as constructor-parameter.

304NotModified:
> But don't think it should be enabled by default. Instead it could be decided when explicitly registering the type, or the type has a special class-attribute.

I was also thinking that we should not resolve instances that aren't registered. 

> Ex. We could have cyclic loop if type-constructor takes its own type as constructor-parameter.

Yes we need to fix that. It's a todo. Cycles are forbidden in every DI framework I know

snakefoot:
> I was also thinking that we should not resolve instances that aren't registered.

Think you misunderstood. It should only resolve constructor for types where the type has been marked to support this.

304NotModified:
> Think you misunderstood. It should only auto-resolve constructor for types where the type has been marked to support this.

I think I understand what you mean. But why we need to mark things? I was thinking:

1. Resolve public ctor and if it hasn't parameters, it's fine
2. If it has parameters, resolve them also. 
3. If we can't resolve the parameters, crash or pass null (but the latter is most of the time not supported in other DI frameworks)

304NotModified:
> Again I feel advanced dependency injection features (with all their side-effects) are not something I would expect inside a logging-library.

People expect ctor injection nowadays. Saw it multiple times in code examples (that won't work).

Ctor injection is now "normal" with the changes to .NET Core IMO

304NotModified:
> Could be nice if one could setup NLog-config in code without it automatically starting reflection-circus (which fails on close-to-native platforms, and just generates overhead).

Good one, but ins't that the case of `Activator.CreateInstance(itemType);` already?

snakefoot:
> but ins't that the case of `Activator.CreateInstance(itemType)` already?

Don't think I said the current code-base is perfect. Lots of places to improve. But improving the ability to perform dependency-injection should not close doors so hard that they can never be opened.

Would be nice if NLog could be used as Blazor component without being a monster.

snakefoot:
> Ctor injection is now "normal" with the changes to .NET Core IMO.

Well not in my world. It is just the decision for Microsoft Dependency Injection. Other dependency injection framworks calls default constructor and assigns properties. And yet others stay away from dependency-injection as they are like religions. Trying to make 10 sub-components work together each with their own flavor of auto-magic dependency injection tricks (hell to me).

Also why should NLog care about constructor-parameter injection? If you need that then you just register your own funky service-resolver. No need to setup traps/overhead from the beginning for everyone.

And how to load a NLog.config where some targets requires interfaces that are not available at loading time. You want to call their constructor, but you just want to delay the initialization until they are ready (have loaded their special access-token from some secret-location). You want logging for simple targets to work (to log startup details), while application is still in progress of loading the interfaces needed for all targets.

The logging engine is not an application. It is the sub-component. It has to be dead simple, and always working.

304NotModified:
> Well not in my world. It is just the decision for Microsoft Dependency Injection. Other dependency injection framworks calls default constructor and assigns properties.

AFAIK ctor injection is by far the de facto standard with all DI containers. MS just took that. It's not new and assigns properties is for DI not recommend for years. 

Autofac - https://autofaccn.readthedocs.io/en/latest/register/prop-method-injection.html

> While constructor parameter injection is the preferred method of passing values to a component being constructed, you can also use property or method injection to provide values. 

Simple injector  - https://simpleinjector.readthedocs.io/en/latest/advanced.html#property-injection
> Simple Injector does not out-of-the-box inject any properties into types that get resolved by the container. In general there are two ways of doing property injection, and both are not enabled by default for reasons explained below.

.NET Core DI stackoverflow post

 https://stackoverflow.com/questions/38459625/property-injection-in-asp-net-core


PS: https://blog.ploeh.dk/2010/02/03/ServiceLocatorisanAnti-Pattern/
Also please note:

304NotModified:
superseded by https://github.com/NLog/NLog/pull/3672

PR #3669, opened 2019-11-05
Decouple LogEventInfo from other classes (WIP)

304NotModified:
@snakefoot I'm trying to get the DI working correctly for  IValueFormatter /IJsonConverter , that means, not hard dependencies. Currently this is broken, as `null` is passed and then ValueFormatter.Instance is used.

Unfortunately i'm now stuck. LogEventInfo is really complex coupled to a lot of classes, which isn't correct as IValueFormatter and IJsonConverter should be in the DI container.

This PR is a try, but failed. Any idea how to decouple this?

### DIFF
--- a/src/NLog/MessageTemplates/ValueFormatter.cs
+++ b/src/NLog/MessageTemplates/ValueFormatter.cs
@@ -47,11 +47,7 @@ namespace NLog.MessageTemplates
     /// </summary>
     internal class ValueFormatter : IValueFormatter
     {
-        public static IValueFormatter Instance
-        {
-            get => _instance ?? (_instance = new ValueFormatter(null)); //todo fix
-            set => _instance = value ?? new ValueFormatter(null);
-        }
+        public static IValueFormatter GetInstance(object serviceRepository) => _instance ?? (_instance = new ValueFormatter(serviceRepository));
         private static IValueFormatter _instance;
         private static readonly IEqualityComparer<object> _referenceEqualsComparer = SingleItemOptimizedHashSet<object>.ReferenceEqualityComparer.Default;
 
--- a/src/NLog/MessageTemplates/ValueFormatter.cs
+++ b/src/NLog/MessageTemplates/ValueFormatter.cs
@@ -37,7 +37,6 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Text;
 using JetBrains.Annotations;
-using NLog.Config;
 using NLog.Internal;
 
 namespace NLog.MessageTemplates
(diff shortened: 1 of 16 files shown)


304NotModified:
maybe we should move PreformatMessage/CalcFormattedMessage from LogEventInfo to LoggerImpl

304NotModified:
LogMessageTemplateFormatter/MessageFormatter should be outside LogEventInfo and then we could unstatic TemplateRenderer

src/NLog/MessageTemplates/ValueFormatter.cs
```diff
@@ -47,21 +46,13 @@ namespace NLog.MessageTemplates
...
-        private readonly IServiceRepository _serviceRepository;
-        private IJsonConverter JsonConverter => _jsonConverter ?? (_jsonConverter = _serviceRepository?.ResolveService<IJsonConverter>());
-        private IJsonConverter _jsonConverter;
+        private readonly IJsonConverter _jsonConverter;
 
-        internal ValueFormatter([NotNull] IServiceRepository serviceRepository)
+        public ValueFormatter([NotNull] IJsonConverter jsonConverter)
         {
-            _serviceRepository = serviceRepository ?? throw new ArgumentNullException(nameof(serviceRepository));
+            _jsonConverter = jsonConverter ?? throw new ArgumentNullException(nameof(jsonConverter));
```
snakefoot: The late resolve of IJsonConverter was actually on purpose, so allow late override.

snakefoot:
Again my idea was to make the Dependency-Injection-Resolver. And commit that.

Then start removing the static-references.

304NotModified:
I know, but both are really related. Injection in this target branch is now broken.

304NotModified:
We can't add di without decouple the logeventinfo I'm afraid

snakefoot:
Not sure I understand. My initial commit works just fine. It adds a service resolver. It is just not used every where.


It is the same as saying NLog is broken because all parameters do not support Layout. One step at the time, then we will get there. Hate eating elephants


Sent from my Sony Xperia


We can't add di without decouple the logeventinfo I'm afraid

304NotModified:
At some places it looks like a reference if from the servicerepo, but is isn't. It's always null and then it fallback to the initial. So it's kind of dead code

src/NLog/MessageTemplates/ValueFormatter.cs
304NotModified: With di there is no late resolve? Then we need a factory

snakefoot:
Well if you complete (or abandon) your active PR for the dependency-injection-branch. Then I don't mind looking at handling the static variable in LogEventInfo (Remember NLog has lots of static variables).

Don't like having multiple chefs standing around the kitchen-sink.

src/NLog/MessageTemplates/ValueFormatter.cs
snakefoot: Thought DI was all about late resolve.

snakefoot:
Serilog is having similar issue with dependency injection.


You want logging to work before dependency injection is fully initialized. So you cannot provide constructor parameters and initialize will fail explicit lookup of interfaces.


So I think magic properties that listens to service repository changes are better.


Then when magic property is marked as required then it will delay call of InitializeTarget  until it is ready. And the magic property will during the lifetime of the application also reference the latest registered interface.


Sent from my Sony Xperia


At some places it looks like a reference if from the servicerepo, but is isn't. It's always null and then it fallback to the initial. So it's kind of dead code

304NotModified:
As you may noticed, I like working together (e.g. in one branch). I think that makes the best software. 

But maybe this isn't a good case for it.

304NotModified:
Close for now.

PR #3672, opened 2019-11-06
DI Support ctor injection & added NLogResolveException

304NotModified:
- DI Support ctor injection 
-  added NLogResolveException

### DIFF
--- a/src/NLog/Config/ServiceRepository.cs
+++ b/src/NLog/Config/ServiceRepository.cs
@@ -65,7 +65,7 @@ namespace NLog.Config
                 ConfigurationItemFactory.Default = null;    //build new global factory
 
             this.RegisterDefaults();
-            CreateInstance = itemType => DefaultResolveInstance(itemType, new HashSet<Type>());
+            CreateInstance = itemType => DefaultResolveInstance(itemType, null);
             // Maybe also include active TimeSource ? Could also be done with LogFactory extension-methods
         }
 
@@ -90,7 +90,7 @@ namespace NLog.Config
                 return createInstance(itemType);
             }
 
-            return DefaultResolveInstance(itemType, new HashSet<Type>());
+            return DefaultResolveInstance(itemType, null);
         }
 
         private object DefaultResolveInstance(Type itemType, HashSet<Type> seenTypes)
@@ -112,6 +112,7 @@ namespace NLog.Config
                     return defaultConstructor.Invoke(ArrayHelper.Empty<object>());
                 }
 
+                seenTypes = seenTypes ?? new HashSet<Type>();
                 return CreateFromParameterizedConstructor(itemType, seenTypes);
             }
             catch (MissingMethodException exception)
@@ -126,7 +127,7 @@ namespace NLog.Config
 
         private object CreateFromParameterizedConstructor(Type itemType, HashSet<Type> seenTypes)
         {
-            var ctors = itemType.GetConstructors(); 
+            var ctors = itemType.GetConstructors();
 
             if (ctors.Length == 0)
             {
--- a/src/NLog/Config/ServiceRepository.cs
+++ b/src/NLog/Config/ServiceRepository.cs
@@ -31,9 +31,12 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
+using NLog.Common;
+
 namespace NLog.Config
 {
     using System;
+    using System.Linq;
     using System.Collections.Generic;
     using System.Reflection;
     using NLog.Internal;
@@ -62,7 +65,7 @@ namespace NLog.Config
                 ConfigurationItemFactory.Default = null;    //build new global factory
 
             this.RegisterDefaults();
-            CreateInstance = DefaultResolveInstance;
+            CreateInstance = itemType => DefaultResolveInstance(itemType, new HashSet<Type>());
             // Maybe also include active TimeSource ? Could also be done with LogFactory extension-methods
         }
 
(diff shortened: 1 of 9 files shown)


src/NLog/Config/ServiceRepository.cs
```diff
@@ -79,32 +82,97 @@ public void RegisterType(Type type, ConfigurationItemCreator objectResolver)
...
+        public object ResolveService(Type itemType)
         {
             var createInstance = CreateInstance;
             if (createInstance != null)
             {
                 return createInstance(itemType);
             }
 
-            return DefaultResolveInstance(itemType);
+            return DefaultResolveInstance(itemType, new HashSet<Type>());
```
snakefoot: Please don't allocate upfront.

src/NLog/Config/ServiceRepository.cs
```diff
@@ -62,7 +65,7 @@ internal ServiceRepository(bool resetGlobalCache = false)
                 ConfigurationItemFactory.Default = null;    //build new global factory
 
             this.RegisterDefaults();
-            CreateInstance = DefaultResolveInstance;
+            CreateInstance = itemType => DefaultResolveInstance(itemType, new HashSet<Type>());
```
snakefoot: Please don't allocate upfront.

src/NLog/Config/ServiceRepository.cs
```diff
@@ -79,32 +82,97 @@ public void RegisterType(Type type, ConfigurationItemCreator objectResolver)
...
             {
+                InternalLogger.Trace("Resolve {0} done", itemType.FullName);
                 return objectResolver(itemType);
             }
 
             try
             {
-                return Activator.CreateInstance(itemType);
+                //todo DI cache/find upfront
+                var defaultConstructor = itemType.GetConstructor(Type.EmptyTypes);
```
snakefoot: Can this lookup be cached?

src/NLog/Config/ServiceRepository.cs
304NotModified: fixed

src/NLog/Config/ServiceRepository.cs
304NotModified: fixed

(PS: I will always try to fix it everywhere)

src/NLog/Config/ServiceRepository.cs
304NotModified: Yes, it's a TODO ;)

Maybe another PR?

tests/NLog.UnitTests/Config/ServiceRepositoryTests.cs
```diff
@@ -44,51 +46,81 @@ public class ServiceRepositoryTests : NLogTestBase
         [Fact]
         public void SideBySideLogFactoryExternalInterfaceTest()
         {
-            // Stage
-            LogFactory logFactory1 = new LogFactory();
-            logFactory1.ServiceRepository.RegisterType(typeof(IMyPrettyInterface), (t) => new MyPrettyImplementation() { Test = nameof(logFactory1) });
+            // Arrange
+            var logFactory1 = new LogFactory();
+            const string name1 = "name1";
+            logFactory1.ServiceRepository.RegisterType(typeof(IMyPrettyInterface), (t) => new MyPrettyImplementation() { Test = name1 });
```
304NotModified: @snakefoot PS; the `nameof` confused me a lot here. So create a separate variable "name")

src/NLog/Config/ServiceRepository.cs
snakefoot: Sorry cannot do dependency injection without proper performance :)

304NotModified: This isn't to the master /dev yet.

;)

304NotModified: we could add it after this. I could do it, or maybe you if you like.

304NotModified: trying to keeping the PRS small

304NotModified: PS with the default ctor is as fast as before ;)

snakefoot: No this is your baby :)

304NotModified: For performance we need much more, e.g. another Register so we could do the reflection stuff on registering. :)

snakefoot: Not talking about the performance of calling. It was the `itemType.GetConstructor`-lookup. `Activator.CreateInstance` performs caching. You need to re-implement that here instead.

304NotModified: OK, so I should prefer Activator.CreateInstance here? 

> Not talking about the performance of calling actual constructor. It was the itemType.GetConstructor-lookup

I'm also not. You could do the lookup on register.

snakefoot: No when you have performed the constructor-lookup, then it doesn't make sense to ask `Activator.CreateInstance` to also perform constructor-lookup.

304NotModified: OK, will do in another PR

PS: We need at the end a final PR to dev, then we can soon/refactor also.

snakefoot: I'm still not sure why we have this feature-branch. Not like dev is released tomorrow. But I will wait until you are done. Then I will look into the LogEventInfo-part.

304NotModified: Because it pollutes dev and I try to keep PRs small

PR #3676, opened 2019-11-07
Fix setter ConfigurationItemFactory.Default

304NotModified:
This is a bug? Or not?


snakefoot:
Well since I want to remove this property completely, then I really don't care much.

The idea was to restart the initialization of the local ConfigurationItemFactory, when changing the global.

304NotModified:
OK, but it isn't removed now. OK will merge this for now.

snakefoot:
> OK, but it isn't removed now. OK will merge this for now.

Please don't instead remove the code completely if you don't understand the code.

PR #3679, opened 2019-11-11
If Load of Extension-Assembly fails, try AppDomain before rethrowing Exception

### LINKED ISSUE #3678: Check AppDomain before loading Extension Assembly
In [`AssemblyHelpers.LoadFromName`](https://github.com/NLog/NLog/blob/dev/src/NLog/Internal/AssemblyHelpers.cs#L105) please check in AppDomain, whether the file is already loaded. 

We are developing Addins for AutoCAD and have a custom Target in an extra Library, which would be in every of our Addins. Specifying `assemblyName` in `Extensions`-Section of nlog.config leads to a `FileNotFoundException`, if nlog is already loaded by another Addin (NLog is looking into directory of nlog.dll (the other addin's directory) and in AutoCAD-Installation-Directory, but not in my addin's directory).
We cannot foresee, where the Autodesk or the User installs the Addin, so specifying Extension via assemblyFile is not an option. But I am able to load my Assembly into AppDomain.

**NLog version**: 4.6.8
**Platform**: .Net 4.5

**Current NLog config:**
```xml
      internalLogLevel="Debug"
      internalLogFile="C:\Temp\nlog-internal.log"
      >

  <extensions>
    <add assembly="AutoCAD.Extensions.NLog" />
  </extensions>
  
  <targets>
    <target name="autocadConsole" xsi:type="AutoCADConsole" />
  </targets>

  <rules>
    <logger name="MyAddin" minlevel="Info" writeTo="autocadConsole" />
  </rules>
</nlog>
```

**Current Result**:
Error loading extensions. Exception: System.IO.FileNotFoundException: Could not load file or assembly 'AutoCAD.Extensions.NLog' or one of its dependencies.

**Expected Result**:
NLog is using the Assembly already loaded into AppDomain.

**Solution**:
Change [`AssemblyHelpers.LoadFromName`](https://github.com/NLog/NLog/blob/dev/src/NLog/Internal/AssemblyHelpers.cs#L105) as follows:
```C#
var name = new AssemblyName(assemblyName);
var loadedAssembly = AppDomain.CurrentDomain.GetAssemblies().FirstOrDefault(domainAssembly =>
{
    var domainAssemblyName = domainAssembly.GetName();
    if (name.Name != domainAssemblyName.Name)
        return false;
    if (name.Version != domainAssemblyName.Version)
        return false;
    if (name.CultureInfo != null && name.CultureInfo != domainAssemblyName.CultureInfo)
        return false;
    if (!name.GetPublicKeyToken().SequenceEqual(domainAssemblyName.GetPublicKeyToken()))
        return false;
    return true;
});
return loadedAssembly ?? Assembly.Load(assemblyName);
```

**Full Stacktrace**:
2019-11-07 10:42:21.5066 Error Error loading extensions. Exception: System.IO.FileNotFoundException: Could not load file or assembly 'AutoCAD.Extensions.NLog' or one of its dependencies. Das System kann die angegebene Datei nicht finden.
File name: 'AutoCAD.Extensions.NLog'
   at System.Reflection.RuntimeAssembly._nLoad(AssemblyName fileName, String codeBase, Evidence assemblySecurity, RuntimeAssembly locationHint, StackCrawlMark& stackMark, IntPtr pPrivHostBinder, Boolean throwOnFileNotFound, Boolean forIntrospection, Boolean suppressSecurityChecks)
   at System.Reflection.RuntimeAssembly.InternalLoadAssemblyName(AssemblyName assemblyRef, Evidence assemblySecurity, RuntimeAssembly reqAssembly, StackCrawlMark& stackMark, IntPtr pPrivHostBinder, Boolean throwOnFileNotFound, Boolean forIntrospection, Boolean suppressSecurityChecks)
   at System.Reflection.RuntimeAssembly.InternalLoad(String assemblyString, Evidence assemblySecurity, StackCrawlMark& stackMark, IntPtr pPrivHostBinder, Boolean forIntrospection)
   at System.Reflection.RuntimeAssembly.InternalLoad(String assemblyString, Evidence assemblySecurity, StackCrawlMark& stackMark, Boolean forIntrospection)
   at System.Reflection.Assembly.Load(String assemblyString)
   at NLog.Internal.AssemblyHelpers.LoadFromName(String assemblyName)
   at NLog.Config.XmlLoggingConfiguration.ParseExtensionWithAssembly(String assemblyName, String prefix)

WRN: Assembly binding logging is turned OFF.
To enable assembly bind failure logging, set the registry value [HKLM\Software\Microsoft\Fusion!EnableLog] (DWORD) to 1.
Note: There is some performance penalty associated with assembly bind failure logging.
To turn this feature off, remove the registry value [HKLM\Software\Microsoft\Fusion!EnableLog].

AchimStuy:
Fixes #3678


src/NLog/Internal/AssemblyHelpers.cs
```diff
@@ -102,9 +103,36 @@ public static Assembly LoadFromName(string assemblyName)
...
+            if (expected.Name != actual.Name)
+                return false;
+            if (expected.Version != null && expected.Version != actual.Version)
+                return false;
+#if !NETSTANDARD1_3 && !NETSTANDARD1_5
+            if (expected.CultureInfo != null && expected.CultureInfo.Name != actual.CultureInfo.Name)
+                return false;
 #endif
+            var expectedKeyToken = expected.GetPublicKeyToken();
+            if (expectedKeyToken != null && !expectedKeyToken.SequenceEqual(actual.GetPublicKeyToken()))
```
304NotModified: Minor, I prefer a variable and return that variable. 

(Instead of if true, return true/false)

src/NLog/Internal/AssemblyHelpers.cs
```diff
@@ -102,9 +103,36 @@ public static Assembly LoadFromName(string assemblyName)
...
+                var name = new AssemblyName(assemblyName);
+                var loadedAssembly = AppDomain.CurrentDomain.GetAssemblies().FirstOrDefault(domainAssembly => CheckAssembly(name, domainAssembly.GetName()));
+                if (loadedAssembly != null)
+                    return loadedAssembly;
+                else throw;
+            }
+#endif
+        }
+
+        static bool CheckAssembly(AssemblyName expected, AssemblyName actual)
```
304NotModified: Proposal name: IsSameAssembly

src/NLog/Internal/AssemblyHelpers.cs
```diff
@@ -102,9 +103,36 @@ public static Assembly LoadFromName(string assemblyName)
...
-            Assembly assembly = Assembly.Load(assemblyName);
-            return assembly;
+            try
+            {
+                Assembly assembly = Assembly.Load(AssemblyName.GetAssemblyName(assemblyName));
+                return assembly;
+            }
+            catch (FileNotFoundException)
+            {
+                var name = new AssemblyName(assemblyName);
```
304NotModified: Maybe log a debug/Trace to the internal logger?

AchimStuy: Yep.

src/NLog/Internal/AssemblyHelpers.cs
AchimStuy: I admit, my name is not good. But in my opinion 'IsSameAssembly' is also misleading, because it will also return true, if `expected` does not have a Version or Culture specified, but `actual` has. I was thinking of 'MeetsCriteria' or simply 'IsRequested'? During this I would also rename 'expected'-parameter to 'requested'.
Because we are in a static class, I could also implement it as an Extension Method, so we could call it like `domainAssembly.MeetsCriteria(name)`

src/NLog/Internal/AssemblyHelpers.cs
AchimStuy: That's a style, I don't like. :) But because it's your codebase I will change it. :)

src/NLog/Internal/AssemblyHelpers.cs
304NotModified: IsAssemblyMatch?

I use match for fuzzy matches. :) (Otherwise "equals")

src/NLog/Internal/AssemblyHelpers.cs
304NotModified: PS. http://wiki.c2.com/?ReturnBooleanEvaluations

304NotModified:
please check the test results:

```
 NLog.UnitTests.Config.ExtensionTests.Extensions_NLogPackageLoader_should_beCalled [FAIL]
      NLog.NLogConfigurationException : Exception when parsing . 
      ---- NLog.NLogConfigurationException : Error loading extensions: PackageLoaderTestAssembly
```

I think this test isn't called locally

src/NLog/Internal/AssemblyHelpers.cs
AchimStuy: I usually prefer the `return a > b`-Style, but in that case it would have led to an expression like 
```
return !(expected.Name != actual.Name
    || (expected.Version != null && expected.Version != actual.Version)
    || (expected.CultureInfo != null && expected.CultureInfo.Name != actual.CultureInfo.Name)
    || (expected.GetPublicKeyToken() != null && !expected.GetPublicKeyToken().SequenceEqual(actual.GetPublicKeyToken())));
```
but in my opinion this is less explicit than my version.

AchimStuy:
> please check the test results:
> 
> ```
>  NLog.UnitTests.Config.ExtensionTests.Extensions_NLogPackageLoader_should_beCalled [FAIL]
>       NLog.NLogConfigurationException : Exception when parsing . 
>       ---- NLog.NLogConfigurationException : Error loading extensions: PackageLoaderTestAssembly
> ```
> 
> I think this test isn't enabled locally

It is enabled, but for net452 it fails with FileNotFoundException. netcoreapp2.0 works.

How should I test my change? Create a new Project and manually load the assembly from it's build output path before creating the `XmlLoggingConfiguration`?

src/NLog/Internal/AssemblyHelpers.cs
304NotModified: No I don't advocate that. ;)

Just no 

```C#
if(true) return true else return false
```

This is OK
```c#
if(true) return true;
somethingelse()
```

304NotModified:
> It is enabled, but for net452 it fails with FileNotFoundException. netcoreapp2.0 works.
> 
> How should I test my change? Create a new Project and manually load the assembly from it's build output path before creating the `XmlLoggingConfiguration`?

Checking this now

304NotModified:
```
 Assembly assembly = Assembly.Load(assemblyName);
```
works, but 
```
 Assembly assembly = Assembly.Load(AssemblyName.GetAssemblyName(assemblyName));
```
doesn't. Why do we need that `AssemblyName.GetAssemblyName`?

The assemblyname is "PackageLoaderTestAssembly" and `AssemblyName.GetAssemblyName` throws the exception

304NotModified:
I think we need new unit tests for this. Otherwise it's hard to maintain. (I can't refactor this now without proper tests)

AchimStuy:
> ```
>  Assembly assembly = Assembly.Load(assemblyName);
> ```
> 
> works, but
> 
> ```
>  Assembly assembly = Assembly.Load(AssemblyName.GetAssemblyName(assemblyName));
> ```
> 
> doesn't. Why do we need that `AssemblyName.GetAssemblyName`?
> 
> The assemblyname is "PackageLoaderTestAssembly" and `AssemblyName.GetAssemblyName` throws the exception

D'oh! This was my mistake. It's still from Rolf's proposal to test `Assembly.Load(AssemblyName.GetAssemblyName(assemblyName));` Excuse me, please. Sure we can keep `Assembly.Load(assemblyName)`.

AchimStuy:
> I think we need new unit tests for this. Otherwise it's hard to maintain. (I can't refactor this now without proper tests)

Just tell my whether I should implement the tests as proposed by me or whether you have a better idea.

304NotModified:
> Just tell my whether I should implement the tests as proposed by me or whether you have a better idea.

Where is the proposal? I missed it I think

304NotModified:
@AchimStuy I changed some stuff here as help. 👼  (see pushed commits)

src/NLog/Internal/AssemblyHelpers.cs
304NotModified: See commit. Resolved now :)

src/NLog/Internal/AssemblyHelpers.cs
304NotModified: renamed

AchimStuy:
> 
> 
> @AchimStuy I changed some stuff here as help. 👼 (see pushed commits)

Oh, thanks. I also had implemented it locally, just not pushed it.

AchimStuy:
> > Just tell my whether I should implement the tests as proposed by me or whether you have a better idea.
> 
> Where is the proposal? I missed it I think

I [proposed](https://github.com/NLog/NLog/pull/3679#issuecomment-553785932) to create a new project and load the assembly manually from it's build output path before create the XmlLoggingConfiguration.

304NotModified:
> > > Just tell my whether I should implement the tests as proposed by me or whether you have a better idea.
> > 
> > 
> > Where is the proposal? I missed it I think
> 
> I [proposed](https://github.com/NLog/NLog/pull/3679#issuecomment-553785932) to create a new project and load the assembly manually from it's build output path before create the XmlLoggingConfiguration.

Sounds good to me!

304NotModified:
PS. I don't get a notification for a push to a pr (otherwise I will be spammed). So just noticed the test. Thanks!

304NotModified:
I don't know why Travis is keep failing. Have restarted it again

304NotModified:
Travis:

NLog.UnitTests.Config.ExtensionTests.LoadExtensionFromAppDomain [FAIL]
      System.IO.FileNotFoundException : 
      Stack Trace:
          at (wrapper managed-to-native) System.Reflection.Assembly.LoadFrom(string,bool)
          at System.Reflection.Assembly.LoadFrom (System.String assemblyFile) [0x00000] in <a6f810169ec746eeb2789b925e2858c0>:0 
          at NLog.UnitTests.Config.ExtensionTests.LoadExtensionFromAppDomain () [0x00068] in <0e8ac7cb3f7f4aadbbc3a498f099dc44>:0

AchimStuy:
> PS. I don't get a notification for a push to a pr (otherwise I will be spammed). So just noticed the test. Thanks!

No worries, I was also busy with other stuff.

I also don't understand, why Travis is failing. I do not know Travis, but it seems, my ManuallyLoadesExtension-Dll is not compiled before running the test. (Or built to another directory.) In my opinion NLog.UnitTests.csproj's
```
<ProjectReference Include="..\ManuallyLoadedExtension\ManuallyLoadedExtension.csproj">
  <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
</ProjectReference>
```
should handle this, but maybe I'm wrong.

304NotModified:
> I also don't understand, why Travis is failing.

It's Linux and Mono. Maybe wrong slashes somewhere?

304NotModified:
ow I think we have to update 

https://github.com/NLog/NLog/blob/dev/.travis.yml

There was a good reason for this (not compiling the SLN)...

AchimStuy:
Ah, okay. I tried only under Windows with Visual Studio.

In my opinion compiling the whole solution might not be required, because
the ManuallyLoadedExtension-Assembly should be built automatically because
of the ProjectReference.

But in the Test a Backslash is used during resolving the Assembly-File to
load. I will fix on monday in office.

Btw: I have seen, that also all ProjectReferences are using Backslashes,
but obvisiously this seems not to be a problem, otherwise all test should
be failing. 🤷‍♂️

304NotModified:
@AchimStuy I think you've reacted to an older comment of me (I've changed it). I don't think it's a slash problem, but the travis config 👼 . I think I fixed it now :)

304NotModified:
Hum it still fails. The DLL is created:

> ManuallyLoadedExtension -> /home/travis/build/NLog/NLog/tests/ManuallyLoadedExtension/bin/Release/net452/ManuallyLoadedExtension.dll

AchimStuy:
The FileNotFoundException led us on a wrong track:
My requested behavior to search AppDomain might be working in .NET-Core as well as with net452 under Mono. (Maybe because .NET-Core grew from Mono?) So searching the logs for Loading from AppDomain will fail on Travis.

304NotModified:
Thanks for testing and fixing!

PR #3680, opened 2019-11-12
DI: Optimize constructor calls

304NotModified:
@snakefoot I hope you like it :)

### DIFF
--- a/src/NLog/Config/ServiceRepository.cs
+++ b/src/NLog/Config/ServiceRepository.cs
@@ -48,7 +48,7 @@ namespace NLog.Config
     {
         private readonly Dictionary<Type, ConfigurationItemCreator> _creatorMap = new Dictionary<Type, ConfigurationItemCreator>();
         private readonly Dictionary<Type, CompiledConstructor> _lateBoundMap = new Dictionary<Type, CompiledConstructor>();
-        private readonly object _lateBoundMapLock = new object();
+        private readonly object _lockObject = new object();
         private ConfigurationItemFactory _localItemFactory;
         public event EventHandler<RepositoryUpdateEventArgs> TypeRegistered;
 
@@ -67,7 +67,7 @@ namespace NLog.Config
                 ConfigurationItemFactory.Default = null;    //build new global factory
 
             this.RegisterDefaults();
-            CreateInstance = itemType => DefaultResolveInstance(itemType, null);
+            CreateInstance = DefaultResolveInstanceTop;
             // Maybe also include active TimeSource ? Could also be done with LogFactory extension-methods
         }
 
@@ -85,7 +85,15 @@ namespace NLog.Config
                 return createInstance(itemType);
             }
 
-            return DefaultResolveInstance(itemType, null);
+            return DefaultResolveInstanceTop(itemType);
+        }
+
+        private object DefaultResolveInstanceTop(Type itemType)
+        {
+            lock (_lockObject)
+            {
+                return DefaultResolveInstance(itemType, null);
+            }
         }
 
         private object DefaultResolveInstance(Type itemType, HashSet<Type> seenTypes)
--- a/src/NLog/Config/ServiceRepository.cs
+++ b/src/NLog/Config/ServiceRepository.cs
@@ -47,6 +47,8 @@ namespace NLog.Config
     internal sealed class ServiceRepository : IServiceRepository
     {
         private readonly Dictionary<Type, ConfigurationItemCreator> _creatorMap = new Dictionary<Type, ConfigurationItemCreator>();
+        private readonly Dictionary<Type, CompiledConstructor> _lateBoundMap = new Dictionary<Type, CompiledConstructor>();
+        private readonly object _lateBoundMapLock = new object();
         private ConfigurationItemFactory _localItemFactory;
         public event EventHandler<RepositoryUpdateEventArgs> TypeRegistered;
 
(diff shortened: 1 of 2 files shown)


snakefoot:
I think I would replace Lookup-Dictionary with MruCache (with a very large size)

snakefoot:
I think it is okay that you only cache and precompile the parameter-less constructors.

304NotModified:
I cache both? I think that is OK to cache both?

304NotModified:
Please help me with gettings the locks right. I don't fully understand where you think those should be added.

snakefoot:
I was just thinking to reduce code-complexity (Can remove `LateBoundConstructor` and just use `LateBoundConstructor`)

src/NLog/Config/ServiceRepository.cs
```diff
@@ -95,30 +97,70 @@ private object DefaultResolveInstance(Type itemType, HashSet<Type> seenTypes)
                 return objectResolver(itemType);
             }
 
-            try
+            if (_lateBoundMap.TryGetValue(itemType, out var compiledConstructor))
```
snakefoot: Need try catch here with fallback to lookup with lock.

src/NLog/Config/ServiceRepository.cs
```diff
@@ -95,30 +97,70 @@ private object DefaultResolveInstance(Type itemType, HashSet<Type> seenTypes)
...
-                    return defaultConstructor.Invoke(ArrayHelper.Empty<object>());
-                }
-
-                seenTypes = seenTypes ?? new HashSet<Type>();
-                return CreateFromParameterizedConstructor(itemType, seenTypes);
+                return CreateNewInstance(compiledConstructor, seenTypes);
             }
-            catch (MissingMethodException exception)
+
+            //todo? lock (_lateBoundMapLock)
```
snakefoot: Uncomment the code and extract the locked-section into separate method. When having taken the lock, then re-check _lateBoundMap if Type has been inserted (TryGetValue-again)

304NotModified: done :)

please double check, thanks!

src/NLog/Config/ServiceRepository.cs
304NotModified: done :)

please double check, thanks!

snakefoot:
Missing try-catch around TryGetValue-lookup.


Sent from my Sony Xperia


In src/NLog/Config/ServiceRepository.cs<https://github.com/NLog/NLog/pull/3680#discussion_r346533872>:

> @@ -95,30 +97,70 @@ private object DefaultResolveInstance(Type itemType, HashSet<Type> seenTypes)
                 return objectResolver(itemType);
             }

-            try
+            if (_lateBoundMap.TryGetValue(itemType, out var compiledConstructor))


done :)

please double check, thanks!

304NotModified:
What kind of exception I could expect with the tryGetValue?

snakefoot:
Think the best option is to wrap the lookup and insert with lock. Since config loading is single threaded then it will not hurt performance.


TryGetValue value is not threadsafe and can throw. But having a lock around everything fixes that.


Sent from my Sony Xperia


What kind of exception I could expect with the tryGetValue?

304NotModified:
fixed!

snakefoot:
Don't like this. Please make the factory-lookup and factory-creation under the same lock. No need to optimize this. And please keep it to a single lock-object.

Just make sure to call the factory-method when outside the lock-object.

304NotModified:
> Don't like this. Please make the factory-lookup and factory-creation under the same lock.

Please elaborate. The lock is the same?

> And please keep it to a single lock-object.

So one lock object for two dictionaries?

304NotModified:
ow maybe the issue is that DefaultResolveInstance is called recursively

src/NLog/Internal/CollectionExtensions.cs
```diff
@@ -68,10 +68,30 @@ internal static class CollectionExtensions
...
                 }
             }
+
             return filterLogEvents ?? (hasIgnoredLogEvents ? ArrayHelper.Empty<TItem>() : items);
         }
+
+        /// <summary>
+        /// Thread-safe getting value. Locks <paramref name="lockObject"/> only when needed.
+        /// </summary>
+        internal static bool TryGetValueThreadSafe<TKey, TValue>(this IDictionary<TKey, TValue> dict, TKey key, out TValue value, object lockObject)
```
snakefoot: Please remove this method, and just do lookup and creation of factory method under the same lock

304NotModified: So we won't have lock free resolves? Please note that DI will quickly involve in a lot of resolve calls.

snakefoot: Only when loading config, and it will be performed with a single-thread.

304NotModified: Some resolves are lazy? So that's not guaranteed from a single thread. This method is also used in the Resolve method for (user) targets etc.

snakefoot: If the lookup happens for every LogEvent, then the Layout/Target is doing something wrong. Lock-free optimization should be implemented when it happens for every LogEvent.

304NotModified: good point. Will also think about this, I will call it a night for today :)

snakefoot:
I would prefer if it was just possible register an external IServiceResolver in the ServiceRepository.

Where NLog would use the external IServiceResolver as primary lookup, and then fallback to Activator.CreateInstance().

Then this code could be an example of how one could implement an external IServiceResolver .

304NotModified:
retry with the locks, but still not sure if this exactly what you think is best.

src/NLog/Config/ServiceRepository.cs
```diff
@@ -116,9 +144,42 @@ private object DefaultResolveInstance(Type itemType, HashSet<Type> seenTypes)
...
         private object CreateFromParameterizedConstructor(Type itemType, HashSet<Type> seenTypes)
+        {
+            seenTypes = seenTypes ?? new HashSet<Type>();
+
+            var ctor = GetParameterizedConstructor(itemType);
+            var constructorParameters = ctor.GetParameters();
+            var parameterValues = CreateCtorParameterValues(constructorParameters, seenTypes);
+            var compiledParameterizedCtor = ReflectionHelpers.CreateLateBoundConstructor(ctor);
+            _lateBoundMap.Add(itemType, new CompiledConstructor(compiledParameterizedCtor, constructorParameters));
+            return compiledParameterizedCtor.Invoke(parameterValues);
```
snakefoot: The invoke of the constructor should be outside the lock

304NotModified: OK, so needs a refactor

304NotModified: how important is this? I don't know how to refactor this because it's recursive. 

(no I don't want to drop this feature ;))

snakefoot: Don't think it is important, since it is only exercised for types that doesn't have a default-constructor. But please use `Activator.CreateInstance` when having registered that the type have a default-constructor (since it performs caching of constructor etc).

304NotModified: > But please use `Activator.CreateInstance` when having registered that the type have a default-constructor (since it performs caching of constructor etc).

But we do caching already? I create a `CompiledConstructor` and add it to `_lateBoundMap` for default constructors. If I replace it with `Activator.CreateInstance`, I can't add it to `_lateBoundMap` and also I can't do adding a upfront cache  (do reflection on register instead of runtime) (currently not there)

snakefoot: Not sure I understand. You want to remove your own caching of compiled-methods because it is too cumbersome to implement. And then I say that is okay with me, if you just fallback to `Activator.CreateInstance` when default constructor is available.

304NotModified: > . You want to remove your own caching of compiled-methods because it is too cumbersome to implement.

No I don't :). I like to merge (preferable without changes) this PR and ignore the fact the locks aren't optimal. :)

snakefoot: Then I don't understand why you don't implement the method so it just inserts the type in the _lateBoundMap without calling constructor. And change the code that performs the initial _lateBoundMap-lookup to call the create-constructor-method, and then performs one more _lateBoundMap-lookup after having called the create-constructor-method.

304NotModified: Because maybe i'm missing something, but that won't work as the call to "createInstance and add to latebound map" is recursive. (it's done for all parameters and all parameters of those parameters etc). I don't think we need createInstance is doing lock, unlock, lock, unlock etc.

But maybe we could merge this one and you could improve the locks? That would be really helpful for me, as I am a bit stuck here.

snakefoot: Hehe. Well you know my opinion. That I don't think NLog should perform dependency logic, but should just have support for using an external service-provider-interface. But if you want to commit, then commit.

304NotModified: > Hehe. Well you know my opinion. 

Yes I do 👼  And I think the current code is pretty good and really useful in a many cases. I hope you are agree with me on that.

Issue #3675, opened 2019-11-07
Support DI and DI containers 


304NotModified:
todo list

- [X] Register types with custom creator - https://github.com/NLog/NLog/pull/3616
- [X] Resolve NLog targets, layouts and layout renderers - yes it resolved all classes.
- [ ] Register all NLog targets, layouts and layout renderers ? For external libs
- [ ] Register types with interface, without static
- [x] Constructor injection - #3672
- [x] Easy resolve without Constructor injection - #3666
- [X] Support createIntance old style and make obsolete
- [ ] Obsolete ConfigurationItemCreator
- [ ] Optimize static registers (no lambda)
- [ ] Optimize  constructors precompile / cache
- [ ] Optimize  constructors precalculate  / find upfront
- [ ] Test / make package with another library (MS DI? - Not Unity!)
- [ ] Work correctly with NLog.Extensions.Logging, fix the issue that in the in main the injectables aren't ready (as configureService is later on)
- [ ] resolve nested statics dependencies lazy with storage (when using nested static resolves)

replaces https://github.com/NLog/NLog/issues/3362

working branch: https://github.com/NLog/NLog/tree/DI-Resolving

304NotModified:
@snakefoot you're working on 'Optimize constructors - cache and precalculate on register', isn't?

snakefoot:
Working on injecting IValueFormatter into the LogEventInfo.MessageFormatter

snakefoot:
Oops pushed directly to "DI-Resolving" without creating a pull-request first.

snakefoot:
Please review this commit:

https://github.com/NLog/NLog/commit/05fec46a02ab7feaec753a7ce5cd93cd2cb6a88e

snakefoot:
We should mark ConfigurationItemFactory.Default as obsolete in NLog 4.7 and introduce extension-methods to perform all config.

304NotModified:
working on performance/cache. Not sure if I finish it today.

304NotModified:
> Oops pushed directly to "DI-Resolving" without creating a pull-request first.

I prefer next time a git reset :angel:

snakefoot:
This static property should be marked as obsolete:

> ConfigurationItemFactory.Default

Beside the obvious usages for static registration methods, then there are missing cleanup in #3831:

- SimpleLayout - Parsing layout should happen through extension-method on LogFactory (that uses LogFactory ServiceRepository)
  - Remove the setter on the Text-property
[...]

Issue #2064, opened 2017-04-19
Deadlock when under load


natethegreat44:
Hi! Thanks for reporting this feature/bug/question! 

Please keep / fill in the relevant info from this template so that we can help you as best as possible.

For .NET Core users, please check the platform support: https://github.com/NLog/NLog/wiki/platform-support

**Type** (choose one):

- Bug

**NLog version**: 5.0.0-beta06 (and 5.0.0-beta03)

**Platform**: .NET Core 

**Current NLog config** (xml or C#, if relevant)

```xml
<?xml version="1.0" encoding="utf-8" ?>
<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd"
      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
      autoReload="true"
      internalLogLevel="Warn"
      internalLogFile="/tmp/internal-nlog.txt">
 
  <extensions>
    <add assembly="PlayAuth.Common" />
  </extensions>
 
  <variable name="correlationLayout" value="${when:when=length('${correlationId}') &gt; 0:inner= Id=${correlationId},}" />
  <variable name="messageLayout" value="Message=${replace-newlines:Message=${truncated:Limit=10000:inner=${message}}" />
  <variable name="exceptionLayout" value="${when:when=length('${exception}') &gt; 0:inner= Type=${exception:Format=Type} Exception=${exception:format=ToString,StackTrace}${newline}}" />
 
  <targets>
    <target name="splunkfileAsync" xsi:type="AsyncWrapper" queueLimit="20000" overflowAction="Discard">
      <target xsi:type="File" name="file"
              layout="[${date:universalTime=true:format=O}] Level=${level}, Category=${logger},${correlationLayout} ${messageLayout} ${exceptionLayout}"
              fileName="/var/log/starz/playauth.log" archiveFileName="/var/log/starz/log-own.log" archiveAboveSize="52428800" archiveNumbering="Sequence" maxArchiveFiles="2" />
    </target>
    <target xsi:type="Null" name="blackhole" />
  </targets>
 
  <rules>
    <logger name="Microsoft.*" minlevel="Trace" writeTo="blackhole" final="true" />
    <logger name="NewRelic.*" minlevel="Trace" writeTo="blackhole" final="true" />
    <logger name="Enyim.*" levels="Trace,Debug" writeTo="blackhole" final="true" />
    <logger name="*" minlevel="Trace" writeTo="splunkfileAsync" />
  </rules>
[...]

PR #3681, opened 2019-11-12
DI support (total) (WIP)

304NotModified:
https://github.com/NLog/NLog/issues/3675


snakefoot:
Think that the internal ServiceRepository in NLog should NEVER be used to register random user-types / interface-types. 

It should still be possible to override existing NLog-interface-implementations, so they become registered in the internal ServiceRepository. Ex. override the IJsonConverter without providing their custom dependency container.

If needing to register random types then one must register an external IServiceResolver that is used as primary lookup, and then fallback to internal ServiceRepository and then Activator.CreateInstance.

People using Dependency Injection already knows about Microsoft IServiceProvider and Autofac Container (or whatever favorite pet they have). They should just know that their DI-container becomes king when hooked together with NLog.

304NotModified:
@snakefoot I will try to merge this after #3815

304NotModified:
Those conflicts in MethodFactory and ConfigurationItemFactory are pretty hard

snakefoot:
Think we need to reduce the flexibility of the dependency injection system, so users can only do this:
- Register singleton object where providing the actual object-instance
- Register object-type where NLog will be responsible for calling the constructor.

Thus removing the ability to register a custom factory-function or external dependency-injection-container.

Suddenly remembered this case where deadlock occurred because NLog suddenly made use of Microsoft ServiceProvider while logging:

https://github.com/NLog/NLog/issues/2064#issuecomment-296813153

Think we should put a lot of effort in preventing deadlocks (external dependency injection taking locks in wrong order) and stackoverflows (external dependency injection system creating objects that causes logging).

304NotModified:
superseded by #3831

PR #3683, opened 2019-11-15
ObjectGraphScanner - Avoid holding list.SyncRoot lock while scanning

snakefoot:
Fixing regression from #3618 (Not super important, but just easier for the eye)

Note merge to master


304NotModified:
Thanks!

PR #3684, opened 2019-11-15
${message} renders Exception.GetType + Exception.Message when single exception parameter

### LINKED ISSUE #1034: Incorrect work of option "withException" / logger.Error(ex)
Hi
NLog 4.2.0
It seems that message layout renderer always outputs stack trace if "message" parameter has not been set.

Test code:

``` c#
{
    LogManager.GetCurrentClassLogger().Error(new Exception("test 1"));
    LogManager.GetCurrentClassLogger().Error(new Exception("test 2"), string.Empty);
}
```

NLog.config:

``` xml

    <targets>
        <target xsi:type="File" encoding="UTF-8"
            name="File"
            filename="Log.log"
            layout="${message:withException=true} ${exception:format=tostring}" />
    </targets>

    <rules>
        <logger name="*" minlevel="Trace" writeTo="File" />
    </rules>
</nlog>
```

Result with 'withException=false':

```
System.Exception: test 1 System.Exception: test 1
 System.Exception: test 2
```

Result with 'withException=true':

```
System.Exception: test 1
System.Exception: test 1 System.Exception: test 1

System.Exception: test 2 System.Exception: test 2
```

### LINKED ISSUE #1291: Add special handling for exceptions in Log<T>(...)
NLog should **never** render `${message}` as the return value of `Exception.ToString()`. Instead render `${message}` as the value of `Exception.Message` when no other message is specified by  the caller.

``` c#
catch(Exception e)
{
    // No message was passed -> render ${message} as e.Message instead of e.ToString()
    logger.Error(e);
}
```

Relates to #1290

snakefoot:
Instead of rendering Exception.ToString(). This PR resolves #1034 and resolves #1291 

Users that perform the following logging-operation will no longer get full stacktrace in `${message}`:

`logger.Error(ex);`

Users will still be able to get full stacktrace with `${message:withException=true}` or `${exception}`.

This PR will not affect users that perform the following logger-operation (Maintains the illusion of `string.Format`):

`logger.Error("{0}", ex);`

The new custom formatting is implemented with a special IFormatProvider, that is used when single-parameter-exception. This means `LogEventInfo.FormattedMessage` will also be affected (Will match the output of `${message}`)

### DIFF
--- a/tests/NLog.UnitTests/LayoutRenderers/MessageTests.cs
+++ b/tests/NLog.UnitTests/LayoutRenderers/MessageTests.cs
@@ -219,5 +219,31 @@ namespace NLog.UnitTests.LayoutRenderers
             logger.Debug(ex);
             AssertDebugLastMessage("debug", ex.ToString());
         }
+
+        [Fact]
+        public void SingleParameterException_OutputsSingleStackTrace()
+        {
+            // Arrange
+            var logFactory = new LogFactory();
+            var logConfig = new LoggingConfiguration(logFactory);
+            var logTarget = new NLog.Targets.DebugTarget("debug") { Layout = "${message}|${exception}" };
+            logConfig.AddRuleForAllLevels(logTarget);
+            logFactory.Configuration = logConfig;
+            var logger = logFactory.GetLogger("SingleParameterException");
+
+            // Act
+            try
+            {
+                logger.Info("Hello");
+                throw new ArgumentException("Holy Moly");
+            }
+            catch (Exception ex)
+            {
+                logger.Fatal(ex);
+            }
+
+            // Assert
+            Assert.StartsWith("System.ArgumentException: Holy Moly|System.ArgumentException", logTarget.LastMessage);
+        }
     }
 }
--- a/src/NLog/Internal/ExceptionMessageFormatProvider.cs
+++ b/src/NLog/Internal/ExceptionMessageFormatProvider.cs
@@ -1,0 +1,62 @@
+﻿// 
+// Copyright (c) 2004-2019 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without 
+// modification, are permitted provided that the following conditions 
+// are met:
+// 
+// * Redistributions of source code must retain the above copyright notice, 
+//   this list of conditions and the following disclaimer. 
+// 
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution. 
+// 
+// * Neither the name of Jaroslaw Kowalski nor the names of its 
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission. 
+// 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE 
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE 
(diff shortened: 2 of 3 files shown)


304NotModified:
Not checked in depth yet, but will ${exception} work with logger.fatal(exception) now? I think that was one of the issues.

tests/NLog.UnitTests/LayoutRenderers/MessageTests.cs
```diff
@@ -219,5 +219,34 @@ public void MessageWithExceptionAndCustomSeparatorTest()
             logger.Debug(ex);
             AssertDebugLastMessage("debug", ex.ToString());
         }
+
+        [Fact]
+        public void SingleParameterException_OutputsSingleStackTrace()
+        {
+            // Stage
```
304NotModified: Please use the term Arrange, -  AAA (Arrange-Act-Assert)

See https://medium.com/@pjbgf/title-testing-code-ocd-and-the-aaa-pattern-df453975ab80

tests/NLog.UnitTests/LayoutRenderers/MessageTests.cs
```diff
@@ -219,5 +219,34 @@ public void MessageWithExceptionAndCustomSeparatorTest()
...
+                throw new ArgumentException("Holy Moly");
+            }
+            catch (Exception ex)
+            {
+                logger.Fatal(ex);
+                argException = ex;
+            }
+
+            // Assert
+            Assert.NotNull(argException);
```
304NotModified: this is a strange assert. It asserts not our code. Proposal: remove it

tests/NLog.UnitTests/LayoutRenderers/MessageTests.cs
```diff
@@ -219,5 +219,34 @@ public void MessageWithExceptionAndCustomSeparatorTest()
             logger.Debug(ex);
             AssertDebugLastMessage("debug", ex.ToString());
         }
+
+        [Fact]
+        public void SingleParameterException_OutputsSingleStackTrace()
```
304NotModified: proposal: split the test for ${message} and ${exception}, then its better to read/maintain. 

(please create helpers for duplicate code)

Or use a parameterized test.

tests/NLog.UnitTests/LayoutRenderers/MessageTests.cs
```diff
@@ -219,5 +219,34 @@ public void MessageWithExceptionAndCustomSeparatorTest()
...
+            }
+            catch (Exception ex)
+            {
+                logger.Fatal(ex);
+                argException = ex;
+            }
+
+            // Assert
+            Assert.NotNull(argException);
+            Assert.StartsWith($"{argException.GetType()}: {argException.Message}|{argException.GetType()}", logTarget.LastMessage);
```
304NotModified: Maybe just write out the values? Those are fixed.

So

```
 Assert.StartsWith("System.ArgumentException: Holy Moly |System.ArgumentException", logTarget.LastMessage);
```

snakefoot:
It has been working since NLog 4.5.4


Sent from my Sony Xperia


Not checked in depth yet, but will ${exception} work with logger.fatal(exception) now? I think that was one of the issues.

tests/NLog.UnitTests/LayoutRenderers/MessageTests.cs
snakefoot: Fixed

tests/NLog.UnitTests/LayoutRenderers/MessageTests.cs
snakefoot: Fixed

tests/NLog.UnitTests/LayoutRenderers/MessageTests.cs
snakefoot: I prefer the unit-test like this

tests/NLog.UnitTests/LayoutRenderers/MessageTests.cs
snakefoot: Fixed

tests/NLog.UnitTests/LayoutRenderers/MessageTests.cs
304NotModified: Ok. It's now anyway much better to read :)

Issue #3433, opened 2019-05-25
LogBuilder skip allocation when no targets


snakefoot:
Could be nice if LogBuilder actually skipped the allocation of itself and LogEventInfo, when calling Logger with Level where no targets.

NLog.Fluent.LoggerExtensions  could instead return a dummy LogBuilder that doesn't do anything.

Would require one to remove the `Level`-method on the LogBuilder (Prevent it from dynamically changing level after initial level-assignment).

snakefoot:
Alternative one could consider dropping LogBuilder completely, and instead create new extension-methods for Logger-interface that returns a LogEventInfo, an then create extension-methods for LogEventInfo (Extension methods for LogEventInfo will handle that no LogEventInfo is returned if loglevel has no targets).

Ex.:

`logger.Info().Write("hello {0}", world);`

304NotModified:
With fluent APIs it's in my opinion always better to return on intermediate object. (Until last call). That's more flexible and the dependencies are also better. So last paragraph I guess?

Breaking change I think ;). We have to keep in my that we should win enough for those Breaking changes.

snakefoot:
Seems that `WithProperty` is working out nicely, and makes easier to apply structured logging to existing applications. I saw a nice trick where one did the following:

```
logger.LogProperties(new { OrderId = 42, UserId = 666, RequestId = Guid.NewGuid }).Info("Order Created");
```

Where the fluent-api performs reflection of the expando-object (can also just be `IDictionary<string,object>`). And harvest the property-names and values. This would also allow one to create a reusable expando-object with common-properties, where one just fill out what is available at the location:

```
logger.LogProperties(new LogContext() { OrderId = 42, UserId = 666 }).Info("Order Deleted");
```

This removes some of the magic-string logic from message-templates, and ensures the property-names are using the same standard everywhere. 

Maybe have the following ILogger extension methods (That returns a `LogBuilder`-struct)

- LogProperties(object value)
- LogProperty(string name, object value)
- LogException(Exception exception)
- LogCallsite() // Caller-member-attributes (not sure if they should happen automatically

The `LogBuilder`-struct only allocates (and performs reflection) when finally calling Debug/Info/Error-method, where it checks if the LogLevel is enabled.

snakefoot:
Resolved by #4325

PR #3685, opened 2019-11-16
LogBuilder - Removed Caller-properties since they are available with ${callsite}

### LINKED ISSUE #775: Fluent logger uses caller attributes but non-fluent not
from https://github.com/NLog/NLog/issues/768

> As a side remark, it's a bit weird that the fluent API includes caller info like this but the regular one doesn't. You might want to think about how to harmonize this behavior across the board

We should fix this, because it's inconsistent.

### LINKED ISSUE #2501: Fluent logger should not inject callsite as properties
This is for NLog 5.0 (new). 

Create a special setting that controls whether it should pollute the logevent-properties with its special callsite-properties (Default = false).

At the same time fix the interface problem when combining assemblies compiled with Net40 and Net45 that both makes use of the fluent-logger-interface. See also #2059

snakefoot:
Resolves #775 and resolves #2501

Also removes `IncludeCallerInformation`-property from `${all-event-properties}`, since it has no meaning.

Follow up to the work started with #3431. Still need to figure out a nice way to capture caller-member-variables when not using LogBuilder.  Maybe it should just be LogBuilder as a struct that performs no allocation, when LogLevel not active. Moving LogBuilder out of NLog.Fluent. See also #3433

### DIFF
--- a/src/NLog/Fluent/LogBuilder.cs
+++ b/src/NLog/Fluent/LogBuilder.cs
@@ -380,14 +380,6 @@ namespace NLog.Fluent
 
         private void SetCallerInfo(string callerMethodName, string callerFilePath, int callerLineNumber)
         {
-            // TODO NLog ver. 5 - Remove these properties
-            if (callerMethodName != null)
-                Property("CallerMemberName", callerMethodName);
-            if (callerFilePath != null)
-                Property("CallerFilePath", callerFilePath);
-            if (callerLineNumber != 0)
-                Property("CallerLineNumber", callerLineNumber);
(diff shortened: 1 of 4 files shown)


304NotModified:
Aren't caller attributes more stable and performant compared to ${callsite}?

snakefoot:
Notice the call to _logEvent.SetCallerInfo() ${callsite} will always use that information if provided without using StackTrace


Sent from my Sony Xperia


Aren't caller attributes more stable and performant compared to ${callsite}?

PR #3686, opened 2019-11-16
LogFactory.Setup to allow fluent configuration of LogFactory options

snakefoot:
Moving the discussion from #3610

Trying to guess what @304NotModified wants, and making an implementation of this.

You seem to prefer how Microsoft has created ConfiguringLogging. And with this PR you can do this:

Or you can do this:

```c#
LogManager.Setup().
   SetupExtensions(eb => {
          eb.AutoLoadAssemblies(false);
          eb.RegisterTarget<MyTarget>("MyTarget");
   });
```

This will mix the one-time setup together with runtime-adjustment of active LoggingConfiguration (When we start the journey for fluent configuration of LoggingConfiguration)


304NotModified:
MS is doing 

```
 .ConfigureLogging(logging =>
        {
            logging.ClearProviders();
            logging.AddConsole();
        })
        .ConfigureWebHostDefaults(webBuilder =>
        {
            webBuilder.UseStartup<Startup>();
        });

```

So the equivalent would be:

```
LogManager.Setup(builder => {
    builder.SkipAutoShutdown();
    builder.SkipAutoLoadNLogAssemblyExtensions();
});
```

snakefoot:
Also possible with this PR

304NotModified:
👍

304NotModified:
so what's the idea of setup? Only special options, or all config?

With the latter, I propose Configure instead of Setup. (not config as that is a abbreviation of configuration and the name should be in "active form")

snakefoot:
Only special LogFactory options independent of LoggingConfiguration. Like configuration of ServiceRepository etc.

Fluent configuration of LoggingConfiguration is very dangerous territory and we will most likely never agree there about the interface. So I will leave that one for you.

snakefoot:
My idea is to use the Setup()-method to obsolete `ConfigurationItemFactory.Default` in NLog 4.7 and have provided enough ISetupBuilder-extension methods, so one will never need it.

Then in NLog 5.0 `ConfigurationItemFactory.Default` can be removed completely and one can rewrite the ISetupBuilder-extension methods to modify the LogFactory.ServiceRepository.

304NotModified:
> Then in NLog 5.0 `ConfigurationItemFactory.Default` can be removed completely and one can rewrite the ISetupBuilder-extension methods to modify the LogFactory.ServiceRepository.

👍 

Not afraid that the Setup method will have all options and will be to broad? For example, MS has also ConfigureLogging and ConfigureServices, and not one Setup

snakefoot:
> Not afraid that the Setup method will have all options and will be to broad? For example, MS has also ConfigureLogging and ConfigureServices, and not one Setup

NLog is not ASP.NET-framework. NLog is "just" logging.

snakefoot:
@304NotModified Possible to get a "Yes" or "No" for the interface presented in this PR ? 

Will make unit-test to complete coverage, when having a "Go". Would like to get NLog 4.7 out, so work can continue on Dependency-Injection-Logic in NLog 5.0.

304NotModified:
I have difficulties with **one** configure method. It's too much different things in one method.  I will be large or too large in the future.

snakefoot:
There will be **one** fluent configure method one day for setting up LoggingConfiguration. Think it is idea with fluent-interfaces. Easy extensible.

304NotModified:
You can see in the long names (e.g. SkipAutoLoadNLogAssemblyExtensions), that the configure it too general. I don't think it's a good idea if there will be 30 options in the future in .Configure

snakefoot:
It is better to come with ideas, than saying "can it be done better ?" :)

I have a problem with all configuration is currently done with static methods, and they are being stuffed in every vacant location available. And it is not pretty.

Was thinking to start at some simple structure and as things grow, then make things obsolete and group them when having 5-7 options that are related.

Have no problem with grouping items related to recognizing config-types from the config-file. Just like with the IHostBuilder you can call the ILoggingBuilder.

In NLog case it would be ISetupBuilder gives access to IConfigItemBuilder ? (Or some better name).

304NotModified:
My proposal was to have more separate methods,

e..g

```
LogManager.ConfigureAutoLoading(builder =>
{
   builder.SkipNLogAssemblyExtensions();
});
```

but that's just https://github.com/NLog/NLog/pull/3686#issuecomment-554684199

snakefoot:
So instead of having 30 methods grouped nicely in separate-groups under Setup(). Then you will have 15 builders directly on LogManager ?

Remember that skipping assembly loading is part of `IConfigFactoryBuilder`.

304NotModified:
> So instead of having 30 methods grouped nicely in separate-groups under Setup().

I think 4-5 separate groups would be nice then. Max 7 options is a nice guide line.

snakefoot:
I think there should be these methods on LogFactory:

- Method for pre-setup
- Method for loading-config
- Method for creating logger
- Method for shutdown/close

Any groups (4-5 or more) that is related to pre-setup before having loaded the actual config should be kept under `Setup()`

Similar to this:

```
var builder = new HostBuilder()
    .ConfigureAppConfiguration((hostingContext, config) => { ... })
    .ConfigureServices((hostContext, services) =>
    {
        services.AddOptions();
        // configure services
    })
    .ConfigureLogging((hostingContext, logging) => { ... });
```
Where I imagine that Setup() returns a ISetupBuilder (similar to IHostBuilder above)

snakefoot:
I guess something like this:

```c#
LogManager.Setup().
    SetupConfigurationFactory(cf => 
             cf.SkipAutoLoadAssemblyExtensions();
             cf.RegisterTarget<MyCustomTarget>();
    ).
    SetupServiceInterfaces(di =>
             di.RegisterObjectTransformation<IExcludedInterface>(o => o.ToString());
    ).
    SetupTimeSourceAccurateUtc();
```

304NotModified:
What about:


```
LogManager.Configure().
    ConfigureExtensions(cf => 
             cf.AutoLoad(false);
             cf.RegisterTarget<MyCustomTarget>();
    ).
    ConfigureTransformations(cf =>
             cf.RegisterObjectTransformation<IExcludedInterface>(o => o.ToString());
    ).
    .ConfigureTiming(cf => 
       cf.SetupTimeSourceAccurateUtc()
    );
```

snakefoot:
I was thinking to reserve `Configure()` for the other elephant which is to fluently create a LoggingConfiguration. But was hoping to keep that elephant in the basement for now.

But if you really keen on using `Configure()` for pre-setup before loading the actual LoggingConfiguration. Then I'm ready, as I'm getting tired of walking around this topic :)

304NotModified:
> I was thinking to reserve `Configure()` for the other elephant which is to fluently create a LoggingConfiguration. But was hoping to keep that elephant in the basement for now.

I was under the impression that the LoggingConfiguration will be in this method. Isn't?

"Setup" is more for one-time configuration. It that the case here?

snakefoot:
> "Setup" is more for one-time configuration. It that the case here?

Newer heard about people changing their Time-Source every 10 minutes during the lifetime of their application. Or people who injects their own JsonFormatter every 1 hour. So yes this is the one-time-setup before loading the LoggingConfiguration.

snakefoot:
Force pushed so one can do this:

```c#
LogManager.Configure().
   AutoShutdownWithAppDomain(false).
   ConfigureExtensions(eb => {
          eb.AutoLoadAssemblies(false);
          eb.RegisterTarget<MyTarget>("MyTarget");
   });
```

304NotModified:
Sorry that this takes so long. I think a good API is very important - we have to live with it for years. And reviewing that takes more time then the spare time I have.

snakefoot:
Still not sure I like the name clash between `Configuration` and `Configure`. I find it confusing:
- The new method `Configure()` is for initial static setup of the NLog-engine before loading `Configuration`. 
- The existing 'Configuration` contains the dynamic configuration of targets + rules. Carved in stone and its meaning should not be changed.

I still think `Setup()` is a better name than `Configure()`. At least if it is going to live on LogManager/LogFactory.

304NotModified:
Ok setup it is!

snakefoot:
Back to "Setup()" and added some unit-tests.

snakefoot:
Would also like to add:

- SetupInternalLogging(LogLevel)
  - LogToFile(string)
  - LogToConsole(bool)
  - LogToTrace(bool)
  - LogToWriter(ITextWriter)
- SetupObjectSerializer()
  - RegisterJsonConverter(IJsonConverter)
  - RegisterValueConverter(IValueConverter)
  - RegisterObjectTransformation()
- SetupConfigLoader()
   - ThrowConfigExceptions(bool)
   - SetCandidateConfigFilePaths(string[])
   - ReloadConfigFileWatcher(bool)
- SetupConfigVariables()
   - KeepVariablesOnReload(bool)
   - SetVariable(string, object)

Curious where to put these:

- ActivateTimeSource<T>() : where T : TimeSource
- AutoShutdownWithAppDomain(bool)

304NotModified:
Proposal, but those "misc" / "single items" are hard

- SetupInternalLogging(LogLevel)
  - LogToFile(string)
  - LogToConsole(bool)
  - LogToTrace(bool)
  - LogToWriter(ITextWriter)
- SetupSerialization() // renamed
  - RegisterJsonConverter(IJsonConverter)
  - RegisterValueConverter(IValueConverter)
  - RegisterObjectTransformation()
- SetupConfigFileLoading()  // renamed
   - ThrowOnConfigExceptions(bool) // added on
   - SetConfigFilePaths(string[]) //renamed
   - ReloadConfigFileWatcher(bool) //todo I don't know now what this is doing ;)
- SetupConfigVariables()
   - KeepVariablesOnReload(bool)
   - SetVariable(string, object)
- SetupShutdown() //OR SetupLifecycle?
   - AutoShutdownWithAppDomain(bool)
- SetupTimings //not sure about this 
   - UseTimeSource&lt;T&gt;() : where T : TimeSource   //renamed   




PS: I'm a few days on holiday

snakefoot:
I will make another PR with the above extensions when this initial PR has been merged. I just wanted to present the playing field that I was seeing in my head.

Have now removed `AutoShutdownWithAppDomain` since we don't know where to put it.

snakefoot:
Guess we have to be careful with announcing that NLog now support fluent-configuration. Since everyone would think it means fluently creating the LoggingConfiguraion :wink:

snakefoot:
@304NotModified Possible you can kick the Azure-Build since it performed timeout?

snakefoot:
Updated wiki:

https://github.com/NLog/NLog/wiki/Register-your-custom-component#logfactory-setupextensions

https://github.com/NLog/NLog/wiki/Configure-from-code#logfactory-initial-setup

PR #3689, opened 2019-11-18
LoggingConfiguration - Moved CheckUnusedTargets to InitializeAll 

snakefoot:
To allow use of _configItems. Skip CheckUnusedTargets-checking when warnings disabled.

This also activates validation for NLog.Extension.Logging when loading NLog-config from appsettings.json

### DIFF
--- a/src/NLog/Config/LoggingConfiguration.cs
+++ b/src/NLog/Config/LoggingConfiguration.cs
@@ -765,7 +765,14 @@ namespace NLog.Config
 
         internal void InitializeAll()
         {
+            bool firstInitializeAll = _configItems.Count == 0;
+
             ValidateConfig();
+
+            if (firstInitializeAll && _targets.Count > 0)
+            {
+                CheckUnusedTargets();
+            }
 
             var supportsInitializes = GetSupportsInitializes(true);
             foreach (ISupportsInitialize initialize in supportsInitializes)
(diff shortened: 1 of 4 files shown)


neilmulhy:
Is this suggesting that I'll need to make a change to my appsettings to suppress all warnings or will the warning messages be fixed in a future update?

If it's having to disable warnings, ideally, it would be useful to still send warnings to the internal.txt file.

Would it be possible to suppress nested wrapper specific warnings, whether that be by:

1. automatically recognise that the targets being warned about are part of a nested target that is already being called by a logger
2. having a disable warn / ignore warn property on each nested wrapper target
3. general disable warn / ignore warn setting for nested targets in appsettings / nlog.config

snakefoot:
Warnings will be fixed in future version (NLog ver. 4.7)


Sent from my Sony Xperia


---- neilmulhy wrote ----


Is this suggesting that I'll need to make a change to my appsettings to suppress all warnings or will the warning messages be fixed in a future update?

If it's having to disable warnings, ideally, it would be useful to still send warnings to the internal.txt file.

Would it be possible to suppress nested wrapper specific warnings, whether that be by:

  1.  automatically recognise that the targets being warned about are part of a nested target that is already being called by a logger
  2.  having a disable warn / ignore warn property on each nested wrapper target
  3.  general disable warn / ignore warn setting for nested targets in appsettings / nlog.config

PR #3691, opened 2019-11-18
FileTarget - ConcurrentWrites should use FileStream.Length in archive-check

snakefoot:
FileStream.Length already resolves to API GetFileSize using existing handle. Avoid creating new file-handle with FileInfo-object for each logevent.

Also only performs single lookup of FileAppender GetFileLength when archive-on-size (instead of two)

Also skips lookup of FileAppender.GetFileLength when archive-on-size is disabled (only doing day-roll)

Instead added bonus Flush and FileInfo.Exists check when time to archive to verify the file-handle is still valid.

Removed FileCharacteristicsHelper since it is no longer needed.

Performance measurements shows that `KeepFileOpen=true + ConcurrentWrites=true` will get a nice improvement. Especially on NetCore-platform where it will be 250 times faster.

| Build          | KeepFileOpen=false | KeepFileOpen=true + ConcurrentWrites=true + ForceManaged=true |  KeepFileOpen=true + ConcurrentWrites=true | KeepFileOpen=true + ConcurrentWrites=false |
|------------------|------------|------|------|---------|
| NLog 4.6.8 |   291 msgs/sec        | 342 msgs/sec        |  93.745 msgs/sec | 1.152.752 msgs/sec |
| PR              |   307 msgs/sec        | 102.842 msgs/sec |  138.616 msgs/sec | 1.120.221 msgs/sec |

Test-Target-Config:

```xml
<target name="SyncFilePerfArchive" type="File" fileName="C:\Temp\Log\NLogPerformanceArchivePerf.txt" autoFlush="false" keepFileOpen="true" concurrentWrites="true" layout="${message}" ArchiveAboveSize="100000000000" MaxArchiveFiles="1000" ArchiveEvery="Year" ArchiveNumbering="Date" />
```


src/NLog/Internal/FileInfoExt.cs
```diff
@@ -55,7 +55,20 @@ public static DateTime GetCreationTimeUtc(this FileInfo fileInfo)
 #endif
         }
 
-       
+        public static DateTime? ValidateFileCreationTime<T>(T fileInfo, Func<T, DateTime?> primary, Func<T, DateTime?> fallback, Func<T, DateTime?> finalFallback = null)
```
304NotModified: The idea of the ext classes is to write c# extensions in it ;)

304NotModified:
Nice a lot of code has been removed :)

Maybe good idea to do a performance test?

src/NLog/Internal/FileInfoExt.cs
snakefoot: Refactored the code, so it now belongs :)

snakefoot:
> Maybe good idea to do a performance test?

Will make a test when the PR-nuget-package is ready.

snakefoot:
@304NotModified Updated the initial-post. Looks like it will help a lot on the NetCore-platform

src/NLog/Internal/FileInfoExt.cs
304NotModified: `LookupValidFileCreationTimeUtc<T>` is still no extension? 

Also am a bit confused by `T fileinfo`. How is T a fileinfo in the type system - what am I missing?

304NotModified: I think a helper is in this case better, as you should not make an extension on T here

Pity of the UnixFileInfo, but could live with that when this code is in a helper class.

snakefoot: So remove extensions, and change into helpers, or have both ?

snakefoot: Now introduced a static FileInfoHelper-class

src/NLog/Targets/FileTarget.cs
```diff
@@ -1884,37 +1881,68 @@ private string GetPotentialFileForArchiving(string fileName)
...
+                FileInfo fileInfo = new FileInfo(previousFileName);
+                if (fileInfo.Exists)
+                {
+                    shouldArchive = ShouldArchiveOnFileSize(fileInfo.Length, upcomingWriteSize);
+                    if (shouldArchive)
+                    {
+                        return previousFileName;
+                    }
+
+                    _fileAppenderCache.InvalidateAppender(previousFileName)?.Dispose(); // File changed behind our backs
```
304NotModified: Could we move this out of the if/else? (Same method invocation in if and else branch)

src/NLog/Targets/FileTarget.cs
```diff
@@ -1966,12 +1983,50 @@ private string GetArchiveFileNameBasedOnTime(string fileName, LogEventInfo logEv
...
+            DateTime? fallbackTimeSourceLinux = (previousLogEventTimestamp != DateTime.MinValue && IsSimpleKeepFileOpen) ? previousLogEventTimestamp : (DateTime?)null;
+            var creationTimeSource = _fileAppenderCache.GetFileCreationTimeSource(fileName, fallbackTimeSourceLinux);
+            var fileLength = creationTimeSource.HasValue ? _fileAppenderCache.GetFileLength(fileName) : null;   // Verifies file-handle by checking FileStream.Length
+            if (creationTimeSource.HasValue && fileLength.HasValue)
+            {
+                if (previousLogEventTimestamp != DateTime.MinValue && previousLogEventTimestamp < creationTimeSource)
+                {
+                    if (TruncateArchiveTime(previousLogEventTimestamp, FileArchivePeriod.Minute) < TruncateArchiveTime(creationTimeSource.Value, FileArchivePeriod.Minute) && PlatformDetector.IsUnix)
+                    {
+                        if (IsSimpleKeepFileOpen)
```
304NotModified: Could we please improve the nesting here? For every uncombined `if` will sonarqube complain;)

I think we need some helper methods/variables

src/NLog/Targets/FileTarget.cs
snakefoot: Fixed

src/NLog/Targets/FileTarget.cs
snakefoot: Removed some of the nesting

PR #3694, opened 2019-11-19
JsonSerializer - Write DateTime directly to StringBuilder instead of allocating string

snakefoot:
`System.Xml.XmlConvert.ToString(input, System.Xml.XmlDateTimeSerializationMode.Utc)` is a fixed format.

Reusing the same trick, that has been implemented for `${longdate}`-layout-renderer

In a performance test where creating logevent with below object and writing file:

```
    public class SomeData
    {
        public int MyProperty1 { get; set; } = 1;
        public double MyProperty2 { get; set; } = 1.4;
        public DateTime MyProperty3 { get; set; } = DateTime.Now;
        public bool MyProperty4 { get; set; } = true;
        public string MyProperty5 { get; set; } = "my value";
        public SomeData MyProperty6 { get; set; }
    }
```

Then writing object to file-target becomes 10 pct faster with this optimization, because most time is spent on DateTime-serialization (Now the only overhead is the conversion to UTC)

### DIFF
--- a/src/NLog/Internal/StringBuilderExt.cs
+++ b/src/NLog/Internal/StringBuilderExt.cs
@@ -103,6 +103,12 @@ namespace NLog.Internal
                 return;
             }
 
+            int digitCount = CalculateDigitCount(value);
+            ApppendValueWithDigitCount(builder, value, digitCount);
+        }
+
+        private static int CalculateDigitCount(uint value)
+        {
             // Calculate length of integer when written out
             int length = 0;
             uint length_calc = value;
@@ -113,14 +119,18 @@ namespace NLog.Internal
                 length++;
             }
             while (length_calc > 0);
+            return length;
+        }
 
+        private static void ApppendValueWithDigitCount(StringBuilder builder, uint value, int digitCount)
+        {
             // Pad out space for writing.
-            builder.Append('0', length);
+            builder.Append('0', digitCount);
 
             int strpos = builder.Length;
 
             // We're writing backwards, one character at a time.
-            while (length > 0)
+            while (digitCount > 0)
             {
                 strpos--;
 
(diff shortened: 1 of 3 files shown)


src/NLog/Internal/StringBuilderExt.cs
```diff
@@ -113,26 +119,73 @@ public static void AppendInvariant(this StringBuilder builder, uint value)
...
                 value /= 10;
-                length--;
+                digitCount--;
             }
         }
+
         private static readonly char[] charToInt = new char[] { '0', '1', '2', '3', '4', '5', '6', '7', '8', '9' };
 
+        /// <summary>
+        /// Convert DateTime into UTC and format to yyyy-MM-ddTHH:mm:ss.fffffffZ - DateTime.ToString("o")
```
304NotModified: Maybe note that this will render a ISO 6801 date?

src/NLog/Internal/StringBuilderExt.cs
```diff
@@ -113,26 +119,73 @@ public static void AppendInvariant(this StringBuilder builder, uint value)
...
+        /// </summary>
+        public static void AppendXsdDateTimeRoundTrip(this StringBuilder builder, DateTime dateTime)
```
304NotModified: Minor: xml instead of xsd please :)

tests/NLog.UnitTests/Internal/StringBuilderExtTests.cs
```diff
@@ -74,5 +75,58 @@ void TestAppendInvariant(int input)
             StringBuilderExt.AppendInvariant(sb, input);
             Assert.Equal(input.ToString(System.Globalization.CultureInfo.InvariantCulture), sb.ToString());
         }
+
+        [Theory]
+        [MemberData(nameof(TestAppendXsdDateTimeRoundTripCases))]
+        void TestAppendXsdDateTimeRoundTrip(DateTime input)
```
304NotModified: Proposal, 3 tests with same memberdata. 

(1 act is really prefered)

tests/NLog.UnitTests/Internal/StringBuilderExtTests.cs
```diff
@@ -74,5 +75,58 @@ void TestAppendInvariant(int input)
...
+
+        public static IEnumerable<object[]> TestAppendXsdDateTimeRoundTripCases()
+        {
+            yield return new object[] { DateTime.MinValue };
+            yield return new object[] { DateTime.MaxValue };
+            yield return new object[] { new DateTime(123, 1, 2) };
+            yield return new object[] { new DateTime(1234, 10, 20) };
+            yield return new object[] { new DateTime(1970, 1, 1, 1, 1, 1) };
+            yield return new object[] { new DateTime(1970, 1, 1, 1, 1, 1, 1) };
+            yield return new object[] { new DateTime(1970, 10, 10, 10, 10, 10) };
```
304NotModified: I think the values should be unique, e.g. not same day and month, otherwise you could switch those without triggering a test. 

(Maybe that's covered with addticks, but the addticks cases looks a bit  arbitrary and could be removed  (in the future))

src/NLog/Internal/StringBuilderExt.cs
snakefoot: Fixed

tests/NLog.UnitTests/Internal/StringBuilderExtTests.cs
snakefoot: Fixed

tests/NLog.UnitTests/Internal/StringBuilderExtTests.cs
snakefoot: The most exciting part is the removal of trailing zeros in ticks. So I like the AddTicks-cases.

src/NLog/Internal/StringBuilderExt.cs
snakefoot: Fixed

snakefoot:
Discovered a small nit to pick:  #3705

Issue #2796, opened 2018-07-06
target type="webService" url="${jsonEndpointAddress}" failed


papygourbi:
Hello
apparently, it's not possible to use a variable for the url of the webService. 
in the nlog log file i have the error : system.invalidOperatioException: This operation is not supported for a relative URI at system.Uri.get_absoluteUri()
if i put directly the address in litteral it's work fine.
as we will have different server for different environment it's a problem.

this is my Nlog.config, the variable to use is ${NeuronLogEndpoint}

?xml version="1.0" encoding="utf-8" ?>
<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd"
      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
      autoReload="true"
      throwExceptions="false" 
      throwConfigExceptions="false"
      internalLogToConsole="false"
      internalLogFile="d:\temp\nlog.txt" 
      InternalLogLevel="Trace">

  <variable name="baselocation" value="${environment:variable=NLogBaseLocation}" />
  <variable name="environment" value="${environment:variable=NLogEnvironment}" />
  <variable name="NeuronLogEndpoint" value="${environment:variable=NeuronLogEndpoint}" />
  <variable name="appName" value="BankServices" />
  <variable name="logDirectory" value="${baselocation}/${environment}/${appName}/${shortdate}" />
  
  <targets async="true">
    <target xsi:type="File"
            name="default"
            layout="${longdate} - ${level:uppercase=true}: ${message}${onexception:${newline}EXCEPTION\: ${exception:format=ToString}}"
            fileName="${logDirectory}/Debug.log"
            archiveFileName="${logDirectory}/Debug_${shortdate}.{##}.log"
            archiveNumbering="Sequence"
            archiveEvery="Day"
            maxArchiveFiles="30"
            keepFileOpen="false"
            />
    <target xsi:type="File"
            name="errorFile"
            layout="${longdate} - ${level:uppercase=true}: ${message}${onexception:${newline}EXCEPTION\: ${exception:format=ToString}}"
            fileName="${logDirectory}/Error.log"
            archiveFileName="${logDirectory}/Error_${shortdate}.{##}.log"
            archiveNumbering="Sequence"
            archiveEvery="Day"
            maxArchiveFiles="30"
            keepFileOpen="false"
            />
    
      <target xsi:type="EventLog"
            name="eventlog"
            source="${appName}"
            layout="${message}${newline}${exception:format=ToString}"
            />
[...]

PR #3699, opened 2019-11-21
#2796  Made Url a layout

Bjego:
Hey guys,

the issue is now older than one year. Here is a tiny update to handle variables in the url for the webservice target and it keeps the existing behaviour. 

With best regards

### DIFF
--- a/src/NLog/Targets/WebServiceTarget.cs
+++ b/src/NLog/Targets/WebServiceTarget.cs
@@ -43,6 +43,7 @@ namespace NLog.Targets
     using NLog.Common;
     using NLog.Config;
     using NLog.Internal;
+    using NLog.Layouts;
 
     /// <summary>
     /// Calls the specified web service on each log message.
@@ -122,7 +123,20 @@ namespace NLog.Targets
         /// Gets or sets the web service URL.
         /// </summary>
         /// <docgen category='Web Service Options' order='10' />
-        public Uri Url { get; set; }
+        /// 
+        public Layout Url { get; set; }
+        internal Uri _renderedUrl
+        {
+            get
+            {
+                var uri = RenderLogEvent(Url, LogEventInfo.CreateNullEvent());
+                return new Uri(uri);
+            }
+            set
+            {
+                Url = value.AbsoluteUri;
+            }
+        }
 
         /// <summary>
         /// Gets or sets the Web service method name. Only used with Soap.
@@ -483,7 +497,7 @@ namespace NLog.Targets
         {
             if (Protocol != WebServiceProtocol.HttpGet)
             {
-                return Url;
+                return _renderedUrl;
             }
 
             //if the protocol is HttpGet, we need to add the parameters to the query string of the url
@@ -496,7 +510,7 @@ namespace NLog.Targets
                 queryParameters = sb.ToString();
             }
 
-            var builder = new UriBuilder(Url);
+            var builder = new UriBuilder(_renderedUrl);
             //append our query string to the URL following 
             //the recommendations at https://msdn.microsoft.com/en-us/library/system.uribuilder.query.aspx
             if (builder.Query != null && builder.Query.Length > 1)
--- a/tests/NLog.UnitTests/Targets/WebServiceTargetTests.cs
+++ b/tests/NLog.UnitTests/Targets/WebServiceTargetTests.cs
@@ -250,6 +250,39 @@ Morbi Nulla justo Aenean orci Vestibulum ullamcorper tincidunt mollis et hendrer
             return WsAddress.Substring(0, WsAddress.Length - 5) + (9000 + portOffset).ToString() + "/";
         }
 
+        [Fact]
+        public void WebserviceTest_renderVariable()
+        {
+            var expected = "https://someurl.com/";
+            var configuration = XmlLoggingConfiguration.CreateFromXmlString($@"
+                <nlog throwExceptions='true'>
+                    <variable name='myService' value='{expected}'/>
+                     <targets>
+                        <target type='WebService'
+                                name='ws'
+                                url='${{var:name=myService}}'
+                                protocol='HttpPost'
+                                encoding='UTF-8'
+                               >
+                            <parameter name='param1' type='System.String' layout='${{message}}'/> 
+                            <parameter name='param2' type='System.String' layout='${{level}}'/>
+     
+                        </target>
+                    </targets>
+                    <rules>
+                      <logger name='*' writeTo='ws'>
+                       
+                      </logger>
+                    </rules>
+                </nlog>");
+
+
+            LogManager.Configuration = configuration;
+            var logger = LogManager.GetCurrentClassLogger();
+            var webserviceTarget = logger.Factory.Configuration.AllTargets.First() as WebServiceTarget;
+
+            Assert.Equal(expected,webserviceTarget._renderedUrl.AbsoluteUri);
+        }
         /// <summary>
         /// Test the Webservice with REST api - <see cref="WebServiceProtocol.HttpPost"/> (only checking for no exception)
         /// </summary>


src/NLog/Targets/WebServiceTarget.cs
```diff
@@ -122,7 +123,20 @@ public WebServiceTarget(string name) : this()
         /// Gets or sets the web service URL.
         /// </summary>
         /// <docgen category='Web Service Options' order='10' />
-        public Uri Url { get; set; }
+        /// 
+        public Layout Url { get; set; }
+        internal Uri _renderedUrl
```
snakefoot: This is a propery not a private member. Please remove the underscore.

snakefoot: Why having both Url and _renderedUrl ?

src/NLog/Targets/WebServiceTarget.cs
```diff
@@ -122,7 +123,20 @@ public WebServiceTarget(string name) : this()
...
+        {
+            get
+            {
+                var uri = RenderLogEvent(Url, LogEventInfo.CreateNullEvent());
```
snakefoot: No need to render every time when using `LogEventInfo.CreateNullEvent()`. Instead cache the rendering, and reset the cache in InitializeTarget.

snakefoot:
Thank you for the PR. The reason for not resolving the issue is because it is a breaking change, and requires NLog 5.0. Like this PR is also a breaking change, and cannot be included in NLog 4.7

Luckily NLog 5.0 is under way, but right now focus is on rewiring the logic for better interaction with external dependency injection systems (Ex. Microsoft Dependency Injection).

Bjego:
Okay - I've written my own custom target. I guess that is good enough for now. And I'll have a look at the official target again, when Nlog 5.0 is ready

snakefoot:
You might be interested in: https://nuget.org/NLog.Targets.Http


Sent from my Sony Xperia


---- Bjego wrote ----


Okay - I've written my own custom target. I guess that is good enough for now. And I'll have a look at the official target again, when Nlog 5.0 is ready

Bjego:
@snakefoot no that target isn't good as well. It's also using strings instead of layouts...
I just need a simple splunk target - and that should be able to read the configs from the app settings
my custom target looks like this now.
```csharp
using NLog;
using NLog.Common;
using NLog.Layouts;
using NLog.Targets;
using System;
using System.Collections.Generic;
using System.Linq;
using System.Net;
using System.Net.Http;
using System.Text;
using System.Threading;
using System.Threading.Tasks;

namespace CRM2.NlogTargets
{
  [Target("Splunk")]
  /// <summary>
  /// https://github.com/NLog/NLog/wiki/How-to-write-a-custom-async-target
  /// </summary>
  public class SplunkTarget : AsyncTaskTarget
  {
    /// <summary>
    /// https://github.com/NLog/NLog/wiki/How-to-write-a-custom-async-target
    /// </summary>
    public SplunkTarget()
    {
    }
    protected override void Dispose(bool disposing)
    {
      if (disposing)
      {
        _client.Dispose();
        _client = null;
      }
      base.Dispose(disposing);
    }
    private static SemaphoreSlim _semaphoreSlim;
    private string _url;
    private HttpClient _client;
    public Layout Url { get; set; }
    public Layout Token { get; set; }
    public bool OldSsl { get; set; }


    protected override async Task WriteAsyncTask(IList<LogEventInfo> logEvents, CancellationToken cancellationToken)
    {
      var localClient = _client ?? await InitHttpClient();

      var bodies = from e in logEvents select Layout.Render(e);
      var bodyToSend = string.Join(" ", bodies);

      InternalLogger.Trace($"Sending to Splunk ({logEvents.Count} Events): {bodyToSend}");
      var splunksResponse = await localClient.PostAsync(_url, new StringContent(bodyToSend, Encoding.UTF8, "application/json"), cancellationToken);
      splunksResponse.EnsureSuccessStatusCode();
    }

    private async Task<HttpClient> InitHttpClient()
    {
      try
      {
        await _semaphoreSlim.WaitAsync();
        if (_client == null)
        {
          var token = Token.Render(LogEventInfo.CreateNullEvent());
          InternalLogger.Info($"Using Splunk-Token: {token}");
          _client = new HttpClient();
          _client.DefaultRequestHeaders.Add("Authorization", $"Splunk {token}");
        }
        if (string.IsNullOrWhiteSpace(_url))
        {
          var splunkUrl = $"{Url.Render(LogEventInfo.CreateNullEvent())}/services/collector/event";
          InternalLogger.Info($"Using Splunk-Url:{splunkUrl}");
          _url = splunkUrl;
        }
        return _client;
      }
      finally
      {
        _semaphoreSlim.Release();
      }
    }

    {
      _semaphoreSlim = new SemaphoreSlim(1, 1);
      if (OldSsl)
      {
        ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls | SecurityProtocolType.Tls11 | SecurityProtocolType.Tls12 | SecurityProtocolType.Ssl3;
      }
      base.InitializeTarget();

    }

    protected override Task WriteAsyncTask(LogEventInfo logEvent, CancellationToken cancellationToken)
    {
    }
  }
}

```

and the nlog config is:
```xml
 <target xsi:type="Splunk"
         name="splunk"
         Url="${configsetting:name=Splunk.Server}"
         Token="${configsetting:name=Splunk.Token}"
         OldSsl="true"
         BatchSize="2000"
         TaskDelayMilliseconds="1000"
         OverflowAction="Grow"
         RetryCount="20"
         RetryDelayMilliseconds="2500" >
      <layout type='JsonLayout'>
        <attribute name='sourcetype' layout='_json' />
        <attribute name="event" encode="false">
            <attribute name="Host" layout="${machinename}" />
            <attribute name="App" layout="${appName}"/>
            <attribute name="Exception" encode="false" layout="${exception:format=@}" />
            <attribute name="Message" layout="${message}" />
            <attribute name="Thread" layout="${threadid}" />
            <attribute name="TraceIdentifier" layout="${aspnet-TraceIdentifier}"/>
            <attribute name="Logger" layout="${logger}" />
            <attribute name="Environment" layout="${aspnet-environment}"/>
            <attribute name="Level" layout="${level}"/>
            <attribute name="Date" layout="${longdate}"/>
            <attribute name="Controller" layout="${aspnet-mvc-controller}"/>
            <attribute name="Action" layout="${aspnet-mvc-action}"/>
            <attribute name="User" layout="${aspnet-user-identity}"/>
            <attribute name="Client Ip" layout="${aspnet-request-ip}"/>
            <attribute name="aspnet-request-url" layout="${aspnet-request-url}"/>
            <attribute name="aspnet-request-method" layout="${aspnet-request-method}"/>
            <attribute name="aspnet-request-referrer" layout="${aspnet-request-referrer}"/>
            <attribute name="aspnet-request-form" layout="${aspnet-request-form}"/>
            <attribute name="aspnet-request-host" layout="${aspnet-request-host}"/>
            <attribute name="aspnet-request-useragent" layout="${aspnet-request-useragent}"/>
            <attribute name="aspnet-request-querystring " layout="${aspnet-request-querystring}" />
            <attribute name="aspnet-request-posted-body" layout="${aspnet-request-posted-body}" />
            <attribute name="aspnet-request-contenttype" layout="${aspnet-request-contenttype}" />
            <attribute name="aspnet-user-authtype" layout="${aspnet-user-authtype}"/>
          </layout>
        </attribute>
      </layout>
    </target>
```

Is it worth it to share it in the official Nlog repo? Not sure could create a PR, if it's helpful

snakefoot:
@Bjego You don't need _semaphoreSlim, as AsyncTaskTarget will automatically wait for task to complete before starting new.

But happy to see that AsyncTaskTarget is working out for you. Making me consider to make a HttpClientTarget that inherits from AsyncTaskTarget.

Bjego:
@snakefoot your idea sounds good - maybe this could go into one of the next releases of Nlog? And you maybe deprecate the Webservice Target (for Json).. 
Please keep in mind that the Url and Headers (for authentication) should be layouts :). Makes it easier to stage an app between - dev, UAT and Prod.

Bjego:
@snakefoot I think a "Splunk" / "HttpClientTarget"- Target doesn't need more than the settings I've provided and looks like a simple implementation.
Plus you wouldn't need any external vendors to write the plugin - updating and fixing would be easier for everyone, because Nlog is a good maintained project, instead of those "lone wolfs", which provide specific extentions.
And in Nlog 5 you could inject the HttpClient via the Ihttpclient factory.. If a user wants to inject his/her own httpClient.

PR #3704, opened 2019-11-23
EventLogTarget - Improve diagnostics logging when using dynamic source

snakefoot:
- Added validation of LogName and Source when it changes. 
- Complain when empty Source. 
- Complain when Source doesn't exists.
- Complain when registered LogName for Source doesn't match expected Log.
- Skip Log-name lookup in Windows-registry for every logevent. ([EventLog.Log](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.eventlog.log)-property is expensive)
- Added dispose of previous EventLog-object when dynamically changing source. 
- Removed EventLogTargetMock + EventRecordMock and replaced with EventLogMock .
- Added unit-test that asserts EventLog-Source is recreated when LogName is wrong.

### DIFF
--- a/src/NLog/Targets/EventLogTarget.cs
+++ b/src/NLog/Targets/EventLogTarget.cs
@@ -41,7 +41,6 @@ namespace NLog.Targets
     using Internal.Fakeables;
     using NLog.Common;
     using NLog.Config;
-    using NLog.Internal;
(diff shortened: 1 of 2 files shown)


304NotModified:
great work! Thanks!

PR #3714, opened 2019-12-01
FileTarget - Added support for MaxArchiveDays

### LINKED ISSUE #311: File target: maxArchiveAge
I'd like to propose **maxArchiveAge** (or similar) attribute to File target. The idea is I want to keep the last x days worth of archives. I would expect this to work with **maxArchiveFiles**, whichever is hit first. If one is not specified, only the other should be used.

The syntax should default to days but for convenience might accept "3d" (for days), "1w" (for weeks), "8h" (for hours) etc.

snakefoot:
Resolves #311


304NotModified:
cool! Thanks!

snakefoot:

Issue #3715, opened 2019-12-02
Programmatic configuration of 'maxArchiveFiles' parameter


DirkDeclercq:
NLog 4.6.8

Just like I did with the 'fileName' parameter I'm trying to configure the 'maxArchiveFiles' (see config-below)
In code I have the following : (at startup)
'''
                using (var disableScope = LogManager.DisableLogging())
                {
                    var logFilePath = optionsService.GetOption<LogFilePath>();
                    LogManager.Configuration.Variables["logdir"] = logFilePath.Value;

                    var maximumFileAgeInDays = optionsService.GetOption<MaximumFileAge>();
                    LogManager.Configuration.Variables["maximumFileAgeInDays"] = ( (int) maximumFileAgeInDays.Value ).ToString();

                    LogManager.ReconfigExistingLoggers();
                    LogManager.KeepVariablesOnReload = true;
                }
'''

But with the given configuration (below) no logging is done.
If I replace maxArchiveFiles with a fixed integer-value, it works.
Any ideas ?

        target name="dynamicSchedulerModule"
                xsi:type="File"
                fileName="${var:logdir}/DynamicSchedulerModule-${shortdate}.log"
                archiveFileName="${var:logdir}/DynamicSchedulerModule-{#}.log"
                archiveEvery="Day"
                archiveNumbering="Date"
                archiveDateFormat="yyyy-MM-dd"
                maxArchiveFiles="${var:maximumFileAgeInDays}"
                layout="${defaultLayout}"

snakefoot:
Think you are waiting for this:  #1234

Right now the NLog-Config-Variables are not that easy to work with when modifying them at runtime. In the perfect world, then it should be possible to do this:

```c#
LogManager.Configuration.Variables["maximumFileAgeInDays"] = ( (int) maximumFileAgeInDays.Value ).ToString();
LogManager.Configuration = LogManager.Configuration.Reload();
```

But it will NOT work. I was thinking to fix NLog-Config-Variables in NLog 5.0 as it will be breaking change.

As a work-around could be this until NLog 5.0 is ready:

```c#
NLog.LogManager.Configuration?.AllTargets.OfType<FileTarget>().ToList().ForEach(f => f.MaxArchiveFiles = (int) maximumFileAgeInDays.Value);
```

snakefoot:
Closing as duplicate of #1234

PR #3729, opened 2019-12-06
Added LogManager.Reload

304NotModified:


snakefoot:
Careful Reload() returns null by default, so suddenly you deactivate NLog-config completely.

Reload only works for Xml-Config.

snakefoot:
I think this should be postponed until NLog 5.0 where KeepVariablesOnReload hopefully fixed so it actually works. See also https://github.com/NLog/NLog/issues/3715#issuecomment-560526749

Right now the config-reload is a special road, that one can only find using maps. The research of those maps makes the user aware of all the possible pitfalls on this road. When making it into an official highway with LogManager.Reload(), then it should not lead into a room full of knifes.

PR #3732, opened 2019-12-09
DatabaseTarget - DbType enum parsing should not be case-sensitive

### LINKED ISSUE #3695: Npgsql Database target parameterType docs are off
**NLog version**: 4.6.8

**Platform**: .NET Core 2.1

**Current NLog config**

```json
{
  "nlog": {
    "autoReload": true,
    "throwExceptions": true,
    "internalLogToConsole": true,
    "internalLogLevel": "Error",
    "variables": {
      "appName": "AbandonedAppEmailAutomation"
    },
    "targets": {
      "Console": {
        "type": "Console",
        "layout": "${level} ${event-properties:EventID:format=tostring} ${callsite:includeSourcePath=false:includeNamespace=false} ${message}",
        "detectConsoleAvailable": false
      },
      "Database": {
        "type": "Database",
        "dbProvider": "Npgsql.NpgsqlConnection, Npgsql",
        "connectionString": "Host=localhost;Database=postgres;Username=postgres;Password=postgres;",
        "commandText": "insert into event (event_type_id, level, exception, system, description, source, callsite, date_created) values (@EventId, @Level, @Exception, @System,  @Description, @Source, @Callsite, CAST(@Date AS timestamp));",
        "parameters": [
          {
            "name": "@System",
            "layout": "${machinename}"
          },
          {
            "name": "@Level",
            "layout": "${level}"
          },
          {
            "name": "@Description",
            "layout": "${message}"
          },
          {
            "name": "@Source",
            "layout": "${appName}"
          },
          {
            "name": "@Callsite",
            "layout": "${callsite}"
          },
          {
            "name": "@EventId",
            "layout": "${event-properties:item=EventId_Id:whenEmpty=6}",
            "dbType": "int64",
            "parameterType": "System.Int64"
          },
          {
            "name": "@Exception",
            "layout": "${exception:format=ToString:innerFormat=ToString}"
          },
          {
            "name": "@Date",
            "layout": "${date:format=yyyy-MM-dd HH\\:mm\\:ss.fff}"
          }
        ]
      }
    },
    "rules": {
      "0": {
        "writeTo": "Console",
        "minlevel": "Info",
        "logger": "*",
        "enabled": true
      },
      "1": {
        "writeTo": "Database",
        "minlevel": "Info",
        "logger": "*",
        "enabled": true 
      }
    }
  }
}
```

- The docs say parameterType should be Int64
- It should actually be System.Int64 like I have specified
- Are there any workarounds? Yes

snakefoot:
Resolves #3695

Cannot figure out why it was ignoreCase=false by default, but have now removed the unused `TryParse`, and instead changed to the public `TryParseEnum`-method that works as intended.

### DIFF
--- a/src/NLog/Common/ConversionHelpers.cs
+++ b/src/NLog/Common/ConversionHelpers.cs
@@ -42,7 +42,7 @@ namespace NLog.Common
     public static class ConversionHelpers
     {
         /// <summary>
-        /// Converts input string value into <see cref="System.Enum"/>
+        /// Converts input string value into <see cref="System.Enum"/>. Parsing is case-insensitive.
         /// </summary>
         /// <param name="inputValue">Input value</param>
         /// <param name="resultValue">Output value</param>
@@ -62,19 +62,6 @@ namespace NLog.Common
                 return false;
             }
             return true;
-        }
-
-        /// <summary>
-        /// Converts the string representation of the name or numeric value of one or more enumerated constants to an equivalent enumerated object. A parameter specifies whether the operation is case-sensitive. The return value indicates whether the conversion succeeded.
-        /// </summary>
-        /// <typeparam name="TEnum">The enumeration type to which to convert value.</typeparam>
-        /// <param name="value">The string representation of the enumeration name or underlying value to convert.</param>
-        /// <param name="result">When this method returns, result contains an object of type TEnum whose value is represented by value if the parse operation succeeds. If the parse operation fails, result contains the default value of the underlying type of TEnum. Note that this value need not be a member of the TEnum enumeration. This parameter is passed uninitialized.</param>
-        /// <returns><c>true</c> if the value parameter was converted successfully; otherwise, <c>false</c>.</returns>
-        /// <remarks>Wrapper because Enum.TryParse is not present in .net 3.5</remarks>
-        internal static bool TryParse<TEnum>(string value, out TEnum result) where TEnum : struct
-        {
-            return TryParse(value, false, out result);
         }
 
         /// <summary>
--- a/src/NLog/Targets/DatabaseParameterInfo.cs
+++ b/src/NLog/Targets/DatabaseParameterInfo.cs
@@ -184,7 +184,8 @@ namespace NLog.Targets
                     }
                     else
                     {
-                        if (ConversionHelpers.TryParse(dbTypeNames[dbTypeNames.Length - 1], out DbType dbType))
+                        dbTypeName = dbTypeNames[dbTypeNames.Length - 1];
+                        if (!string.IsNullOrEmpty(dbTypeName) && ConversionHelpers.TryParseEnum(dbTypeName, out DbType dbType))
                         {
                             _dbTypeValue = dbType;
                             ParameterType = TryLookupParameterType(dbType);
@@ -305,7 +306,7 @@ namespace NLog.Targets
                 bool IEnumTypeConverter.TryParseEnum(string value, out Enum enumValue)
                 {
                     TEnum enumValueT;
-                    if (ConversionHelpers.TryParse(value, out enumValueT))
+                    if (!string.IsNullOrEmpty(value) && ConversionHelpers.TryParseEnum(value, out enumValueT))
                     {
                         enumValue = enumValueT as Enum;
                         return enumValue != null;
--- a/tests/NLog.UnitTests/Internal/EnumHelpersTests.cs
+++ b/tests/NLog.UnitTests/Internal/EnumHelpersTests.cs
@@ -31,125 +31,46 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
-using System;
-using System.Collections.Generic;
-using System.Linq;
-using NLog.Common;
-using Xunit;
-
 namespace NLog.UnitTests.Internal
 {
+    using System;
+    using NLog.Common;
+    using Xunit;
+
     public class EnumHelpersTests : NLogTestBase
     {
-
         enum TestEnum
         {
             Foo,
             bar,
-
         }
-
-        #region tryparse - no ignorecase parameter
-
-        [Fact]
-        public void EnumParse1()
-        {
-            TestEnumParseCaseSentisive("Foo", TestEnum.Foo, true);
-        }
-        [Fact]
-        public void EnumParse2()
-        {
-            TestEnumParseCaseSentisive("foo", TestEnum.Foo, false);
-        }
-
-        [Fact]
-        public void EnumParseDefault()
-        {
-            TestEnumParseCaseSentisive("BAR", TestEnum.Foo, false);
-        }
-        [Fact]
-        public void EnumParseDefault2()
-        {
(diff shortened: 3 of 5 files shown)


snakefoot:
Before:
```c#
internal static bool TryParse<TEnum>(string value, out TEnum result) where TEnum : struct
{
   return TryParse(value, false, out result);
}
```
After:
```c#
internal static bool TryParse<TEnum>(string inputValue, out TEnum resultValue) where TEnum : struct
{
   return TryParse(inputValue, true, out resultValue);
}
```
Notice the **true**

carlin-q-scott:
Ah, I only looked at the overload that takes the case sensitive argument. Hopefully the parser code uses the default one.

src/NLog/Common/ConversionHelpers.cs
```diff
@@ -68,30 +68,30 @@ public static class ConversionHelpers
...
+        /// <param name="inputValue">The string representation of the enumeration name or underlying value to convert.</param>
+        /// <param name="resultValue">When this method returns, result contains an object of type TEnum whose value is represented by value if the parse operation succeeds. If the parse operation fails, result contains the default value of the underlying type of TEnum. Note that this value need not be a member of the TEnum enumeration. This parameter is passed uninitialized.</param>
         /// <returns><c>true</c> if the value parameter was converted successfully; otherwise, <c>false</c>.</returns>
         /// <remarks>Wrapper because Enum.TryParse is not present in .net 3.5</remarks>
-        internal static bool TryParse<TEnum>(string value, out TEnum result) where TEnum : struct
+        internal static bool TryParse<TEnum>(string inputValue, out TEnum resultValue) where TEnum : struct
```
304NotModified: What about input and result?

src/NLog/Common/ConversionHelpers.cs
```diff
@@ -68,30 +68,30 @@ public static class ConversionHelpers
...
         {
-            return TryParse(value, false, out result);
+            return TryParse(inputValue, true, out resultValue);
```
304NotModified: This is a bit unexpected, changing the default for the whole code?

src/NLog/Common/ConversionHelpers.cs
```diff
@@ -68,30 +68,30 @@ public static class ConversionHelpers
         /// Converts the string representation of the name or numeric value of one or more enumerated constants to an equivalent enumerated object. A parameter specifies whether the operation is case-sensitive. The return value indicates whether the conversion succeeded.
```
304NotModified: Guess we need to update these docs? 

> A parameter specifies whether the operation is case-sensitive

src/NLog/Common/ConversionHelpers.cs
304NotModified: (I missed the intro text, oops)

304NotModified: Enum.TryParse (.net 4+) is case sensitive by default?

src/NLog/Common/ConversionHelpers.cs
snakefoot: Just aligning with the public method,.

src/NLog/Common/ConversionHelpers.cs
snakefoot: Well when parsing then one wants to be case-insensitive, unless knowing the writer is strict. So ignoreCase=true is a better default for NLog. Just like for the public method

src/NLog/Common/ConversionHelpers.cs
snakefoot: Updated

snakefoot:
After putting on my thinking hat, then I have decided to completely remove the internal TryParse-method without ignoreCase-parameter. Instead changed to the public `TryParseEnum`-method. This will also make it easier to rip out DatabaseTarget from NLog-core in NLog 5.0.

304NotModified:
Good choice!

src/NLog/Common/ConversionHelpers.cs
```diff
@@ -42,7 +42,7 @@ namespace NLog.Common
     public static class ConversionHelpers
     {
         /// <summary>
-        /// Converts input string value into <see cref="System.Enum"/>
+        /// Converts input string value into <see cref="System.Enum"/>. Parsing is case-insensitive.
```
304NotModified: This is wrong now?

src/NLog/Common/ConversionHelpers.cs
```diff
@@ -68,30 +68,17 @@ public static class ConversionHelpers
...
+        /// <param name="resultValue">When this method returns, result contains an object of type TEnum whose value is represented by value if the parse operation succeeds. If the parse operation fails, result contains the default value of the underlying type of TEnum. Note that this value need not be a member of the TEnum enumeration. This parameter is passed uninitialized.</param>
         /// <returns><c>true</c> if the value parameter was converted successfully; otherwise, <c>false</c>.</returns>
         /// <remarks>Wrapper because Enum.TryParse is not present in .net 3.5</remarks>
-        internal static bool TryParse<TEnum>(string value, bool ignoreCase, out TEnum result) where TEnum : struct
+        internal static bool TryParse<TEnum>(string inputValue, bool ignoreCase, out TEnum resultValue) where TEnum : struct
```
304NotModified: I can't comment on the original comment, (due to force push),

But IMO inputValue and resultValue is double and more confusing than helping.  Note, the orignal method has following signature:

```
public static bool TryParse<TEnum> (string value, bool ignoreCase, out TEnum result) where TEnum : struct;
```

I don't know where it's aligned with now in this PR? Currently I still prefer the way before

src/NLog/Common/ConversionHelpers.cs
snakefoot: Not sure I understand. Thought you wanted me to update documentation to match actual behavior.

src/NLog/Common/ConversionHelpers.cs
snakefoot: No problem. Will revert this change. Just wanted to align with the method you introduced yourself TryParseEnum. But I see now this was a mistake.

src/NLog/Common/ConversionHelpers.cs
304NotModified: I'm confused I guess

>  I have decided to completely remove the internal TryParse-method without ignoreCase-parameter

So there is a still a tryparse without ignoreCase?

(and if we have the ignoreCase parameter, "Parsing is case-insensitive." is wrong)

PS: the diff made it here conflusing

snakefoot: Nope. Your existing public method `TryParseEnum` was case-insensitive already (see the source code). When I implemented parsing of DbType-enum I used an internal method TryParse that was case-sensitive. I have now removed this internal method TryParse, and now using public method `TryParseEnum` (Just like I wrote about putting on my thinking hat).

304NotModified: Ok thank for the clarification!

PR #3734, opened 2019-12-11
FileTarget - Write time details to InternalLogger when file is older than MaxArchiveDays

snakefoot:
Followup to  #3714

### DIFF
--- a/src/NLog/Targets/FileArchiveModes/FileArchiveModeBase.cs
+++ b/src/NLog/Targets/FileArchiveModes/FileArchiveModeBase.cs
@@ -195,7 +195,7 @@ namespace NLog.Targets.FileArchiveModes
                     var fileAgeDays = (currentDateUtc - fileDateUtc).TotalDays;
                     if (fileAgeDays > maxArchiveDays)
                     {
-                        InternalLogger.Debug("FileTarget: Detected old file in archive. FileName={0}, FileDate={1}, FileDateUtc={2}, CurrentDateUtc={3}, Age={4} days", existingArchiveFile.FileName, existingArchiveFile.Date.ToString("u"), fileDateUtc.ToString("u"), currentDateUtc.ToString("u"), Math.Round(fileAgeDays, 1));
+                        InternalLogger.Debug("FileTarget: Detected old file in archive. FileName={0}, FileDate={1:u}, FileDateUtc={2:u}, CurrentDateUtc={3:u}, Age={4} days", existingArchiveFile.FileName, existingArchiveFile.Date, fileDateUtc, currentDateUtc, Math.Round(fileAgeDays, 1));
                         return true;
                     }
                 }
--- a/src/NLog/Targets/FileArchiveModes/FileArchiveModeBase.cs
+++ b/src/NLog/Targets/FileArchiveModes/FileArchiveModeBase.cs
@@ -34,6 +34,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using NLog.Common;
 
 namespace NLog.Targets.FileArchiveModes
 {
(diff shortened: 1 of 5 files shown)


src/NLog/Targets/FileArchiveModes/FileArchiveModeBase.cs
```diff
@@ -111,7 +112,7 @@ public virtual List<DateAndSequenceArchive> GetExistingArchiveFiles(string archi
             }
 
             if (existingArchiveFiles.Count > 1)
-                existingArchiveFiles.Sort(FileSortOrderComparison);
+                existingArchiveFiles.Sort((x,y) => FileSortOrderComparison(x, y));
```
304NotModified: Why this change? Do you think it's more clear then?

src/NLog/Targets/FileArchiveModes/FileArchiveModeBase.cs
```diff
@@ -188,8 +189,16 @@ private bool ShouldDeleteFile(DateAndSequenceArchive existingArchiveFile, int re
...
                 var fileDateUtc = existingArchiveFile.Date.Date.ToUniversalTime();
-                if (fileDateUtc > MaxAgeArchiveFileDate && (NLog.Time.TimeSource.Current.Time.Date.ToUniversalTime() - fileDateUtc).TotalDays > maxArchiveDays)
-                    return true;
+                if (fileDateUtc > MaxAgeArchiveFileDate)
+                {
+                    var currentDateUtc = NLog.Time.TimeSource.Current.Time.Date.ToUniversalTime();
+                    var fileAgeDays = (currentDateUtc - fileDateUtc).TotalDays;
+                    if (fileAgeDays > maxArchiveDays)
+                    {
+                        InternalLogger.Debug("FileTarget: Detected old file in archive. FileName={0}, FileDate={1}, FileDateUtc={2}, CurrentDateUtc={3}, Age={4} days", existingArchiveFile.FileName, existingArchiveFile.Date.ToString("u"), fileDateUtc.ToString("u"), currentDateUtc.ToString("u"), Math.Round(fileAgeDays, 1));
```
304NotModified: PS I think you also could do `{2:u}` instead of `ToString("u")`. 

Maybe a bit easier to read, but not very important.

snakefoot: I prefer "u" or "o". But find "u" more readable.

src/NLog/Targets/FileArchiveModes/FileArchiveModeBase.cs
snakefoot: Just removes delegate allocation :)

src/NLog/Targets/FileArchiveModes/FileArchiveModeBase.cs
304NotModified: I mean,

instead of:

```c#
InternalLogger.Debug("Date={0},", existingArchiveFile.FileName, Date.ToString("u"))
```

A more reabable way would be:

```c#
InternalLogger.Debug("Date={0:u},", existingArchiveFile.FileName, Date)
```

:)

snakefoot: Well it still performs boxing etc. But if you really want it then ok.

src/NLog/Targets/FileArchiveModes/FileArchiveModeBase.cs
304NotModified: I really was under the impression that the "method group" call was better I'm perf (no anonymous method) , but have found a good article on that. Do you have something to read for me?

murrays1208049:
@snakefoot Hello, I have a stackoverflow question [https://stackoverflow.com/questions/73783383/nlog-maxarchivedays-delete-old-folders](https://stackoverflow.com/questions/73783383/nlog-maxarchivedays-delete-old-folders). This question troubled me a long time. Could you help me figure out

PR #3737, opened 2019-12-18
Added Layout.CreateFromMethod to allow custom layoutrenders using Config API

snakefoot:
Another building block to make it easier to use the Config API

- Added Layout.CreateFromString and marked Layout.FromString as obsolete
- Added Format-property to FuncLayoutRenderer 
- Added IStringValueRenderer-support to FuncLayoutRenderer

### DIFF
--- a/src/NLog/Layouts/Layout.cs
+++ b/src/NLog/Layouts/Layout.cs
@@ -130,9 +130,9 @@ namespace NLog.Layouts
         /// Implicits converts the specific lambda method into a <see cref="SimpleLayout"/>.
         /// </summary>
         /// <param name="layoutMethod">Method that renders the layout.</param>
-        /// <param name="threadSafe">Tell if method is safe for concurrent threading.</param>
+        /// <param name="options">Tell if method is safe for concurrent threading.</param>
         /// <returns>Instance of <see cref="SimpleLayout"/>.</returns>
-        public static Layout CreateFromMethod(Func<LogEventInfo, object> layoutMethod, bool threadSafe = false)
+        public static Layout CreateFromMethod(Func<LogEventInfo, object> layoutMethod, LayoutRenderOptions options = LayoutRenderOptions.None)
         {
             if (layoutMethod == null)
                 throw new ArgumentNullException(nameof(layoutMethod));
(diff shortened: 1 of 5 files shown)


304NotModified:
cool!

To fully understand, what is the use case for `Layout.FromLayoutMethod`? (as and end-user)?

snakefoot:
Lets imagine https://github.com/nlog/nlog/wiki/JsonLayout#from-the-api

```c#
var jsonLayout = new JsonLayout
{
    Attributes =
    {
        new JsonAttribute("myFunkyValue", Layut.FromLayoutMethod(logEvent => 42),
    }
}

```

304NotModified:
Proposal, rename to CreateFromMethod

304NotModified:
`.Create` would be also fine for me

src/NLog/Layouts/Layout.cs
```diff
@@ -114,6 +114,28 @@ public static Layout FromString(string layoutText, ConfigurationItemFactory conf
             return new SimpleLayout(layoutText, configurationItemFactory);
         }
 
+        /// <summary>
+        /// Implicits converts the specific lambda method into a <see cref="SimpleLayout"/>.
+        /// </summary>
+        /// <param name="layoutMethod">Method that renders the layout.</param>
+        /// <param name="threadSafe">Tell if method is safe for concurrent threading.</param>
+        /// <returns>Instance of <see cref="SimpleLayout"/>.</returns>
+        public static Layout FromLayoutMethod(Func<LogEventInfo, object> layoutMethod, bool threadSafe = false)
```
304NotModified: Consider replacing "threadSafe " with an object (e.g. LayoutCreationOptions). That would be easier to extend in the future)

snakefoot: Have introduced parameter LayoutRenderOptions instead of `bool threadSafe`

304NotModified:
Ow oops. Missed the fromString. Do think create.. is better? (I think so, but obsoletes and/or inconsistent apis aren't preferred!)

snakefoot:
Think Create is nice. But consistency is more important. Since I want to make ConfigurationItemFactory internal in NLog 5.0 then I want to obsolete one of Layout.FromString methods anyway.


Sent from my Sony Xperia


Ow oops. Missed the fromString. Do think create.. is better? (I think so, but obsoletes and/or inconsistent apis aren't preferred!)

304NotModified:
OK merged :)

PR #3742, opened 2019-12-28
TraceTarget - Introduced EnableTraceFail=false to avoid Environment.FailFast

snakefoot:
`EnableTraceFail = false` is a much better default, as it prevents unexpected `Environment.FailFast`

See also: https://github.com/aspnet/AspNetCore/issues/14547#issuecomment-569373496

See also: https://stackoverflow.com/questions/12639012/nlog-using-trace-output-makes-logger-error-throw-an-exception-how-to-fix

### DIFF
--- a/src/NLog/Targets/TraceTarget.cs
+++ b/src/NLog/Targets/TraceTarget.cs
@@ -65,9 +65,19 @@ namespace NLog.Targets
         /// <summary>
         /// Always use <see cref="Trace.WriteLine(string)"/> independent of <see cref="LogLevel"/>
         /// </summary>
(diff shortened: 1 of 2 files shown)


304NotModified:
Thanks! Looks good. Maybe we need another unit test? (Maybe I could also add it)

snakefoot:
> Thanks! Looks good. Maybe we need another unit test?

Extended the unit-test to cover both true/false

304NotModified:
Thanks a lot!

snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/Trace-target

PR #3743, opened 2019-12-28
JsonLayout - EscapeForwardSlash should be nullable

### LINKED ISSUE #3741: Fix EscapeForwardSlash on JsonLayout
The EscapeForwardSlash isn't proper implemented on JsonLayout.

Won't work:

```xml
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" autoReload="true" internalLogToConsoleError="true">
    <targets>
        <target xsi:type="Console" name="console">
            <layout xsi:type="JsonLayout" includeMdlc="true" escapeForwardSlash="false">
                <attribute name="@timestamp" layout="${date:universalTime=true:format=o}" />
                <attribute name="level" layout="${level:upperCase=true}" />
                <attribute name="exception" layout="${onexception:|${exception:format=ToString}}" />
                <attribute name="logger_name" layout="${logger}" />
                <attribute name="thread_name" layout="${threadid}" />
            </layout>
        </target>
    </targets>

    <rules>
        <logger name="Microsoft.*" maxlevel="Info" final="true" />
        <logger name="*" minlevel="Debug" writeTo="console" />
    </rules>
</nlog>
```

Does work: It _does_ work if I add `escapeForwardSlash="false"` to each JsonAttribute individually though:

```xml
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" autoReload="true" internalLogToConsoleError="true">
    <targets>
        <target xsi:type="Console" name="console">
            <layout xsi:type="JsonLayout" includeMdlc="true">
                <attribute name="@timestamp" layout="${date:universalTime=true:format=o}" escapeForwardSlash="false" />
                <attribute name="level" layout="${level:upperCase=true}" escapeForwardSlash="false" />
                <attribute name="message" layout="${message}" escapeForwardSlash="false" />
                <attribute name="exception" layout="${onexception:|${exception:format=ToString}}" escapeForwardSlash="false" />
                <attribute name="logger_name" layout="${logger}" escapeForwardSlash="false" />
                <attribute name="thread_name" layout="${threadid}" escapeForwardSlash="false" />
            </layout>
        </target>
    </targets>

    <rules>
        <logger name="Microsoft.*" maxlevel="Info" final="true" />
        <logger name="*" minlevel="Debug" writeTo="console" />
    </rules>
</nlog>
```

- Not sure if we could fix this easily, as EscapeForwardSlash on JsonEncodeLayoutRendererWrapper / JsonAttribute isn't nullable
- We need this for IncludeAllProperties / IncludeGdc etc
- Probable we also need to include JsonEncode / Encode  /  EscapeUnicode on layout level?

snakefoot:
Trying to resolve  #3741. But it will only apply to direct sub-attributes.

If a nested attribute embeds a JsonLayout using encode="false", then it will not inherit the configuration from the top-JsonLayout (There is no guaranteed of the order of initializing NLog Layouts)

### DIFF
--- a/src/NLog/LayoutRenderers/Wrappers/JsonEncodeLayoutRendererWrapper.cs
+++ b/src/NLog/LayoutRenderers/Wrappers/JsonEncodeLayoutRendererWrapper.cs
@@ -74,8 +74,13 @@ namespace NLog.LayoutRenderers.Wrappers
         /// <summary>
         /// Should forward slashes be escaped? If true, / will be converted to \/ 
         /// </summary>
-        [DefaultValue(true)] // todo NLog 5, default to false
-        public bool EscapeForwardSlash { get; set; } = true;
+        [DefaultValue(true)] // TODO NLog 5 change to nullable (with default fallback to false)
+        public bool EscapeForwardSlash
+        {
+            get => EscapeForwardSlashInternal ?? true;
+            set => EscapeForwardSlashInternal = value;
+        }
+        internal bool? EscapeForwardSlashInternal;
 
         /// <inheritdoc/>
         protected override void RenderInnerAndTransform(LogEventInfo logEvent, StringBuilder builder, int orgLength)
--- a/src/NLog/Layouts/JsonAttribute.cs
+++ b/src/NLog/Layouts/JsonAttribute.cs
@@ -31,10 +31,9 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
-using System.ComponentModel;
-
 namespace NLog.Layouts
 {
+    using System.ComponentModel;
     using NLog.Config;
 
     /// <summary>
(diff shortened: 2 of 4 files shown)


304NotModified:
Great!

304NotModified:
Great work thank!

chazt3n:
How is this done via API?

snakefoot:
@chazt3n Think you have to explain what you want to do, since I cannot guess what you want. `EscapeForwardSlash` is property you can assign on JsonLayout and JsonAttribute:

https://nlog-project.org/documentation/v4.7.0/html/T_NLog_Layouts_JsonLayout.htm
https://nlog-project.org/documentation/v4.7.0/html/T_NLog_Layouts_JsonAttribute.htm

See also: https://stackoverflow.com/a/61033465/193178

chazt3n:
@snakefoot my question was how would I set EscapeForwardSlash property via the fluent API. (i.e. code)

It is actually right on the JsonLayout object! Which is so super awesome. Thank you

chazt3n:
I see now that it may be a recent addition to the API's? My brand new things have it and older things the same code doesn't compile. Could be just doing something stupid.

snakefoot:
@chazt3n Do you have an error-message from the compiler? 

Remember to upgrade to the latest version of NLog in your older project to access the latest NLog-features.

304NotModified:
@chazt3n please open a new issue and provide more details (see explanation).

PR #3745, opened 2019-12-29
Renamed folder StringBuilderPools to ObjectPools

snakefoot:
And removed Async-folder since the items in the folder was very unrelated.


304NotModified:
The async folder was for all the async constructions used in NLog (continuation, helpers etc). Is that unclear?

304NotModified:
Good 2020 @snakefoot!

304NotModified:
The async folder was for all the async constructions used in NLog (continuation, helpers etc). Is that unclear?

snakefoot:
> Good 2020

Enjoyful New Year back to you :)

> The async folder was for all the async constructions used in NLog (continuation, helpers etc). Is that unclear?

Think Async is too wide a term (Reminds me of the Tasks system). AsyncContinuation and AsyncHelpersTask are not the same thing. I guess one could remove AsyncHelpersTask and rename Async-folder to AsyncContinuation-folder.

304NotModified:
> AsyncContinuation and AsyncHelpersTask are not the same thing.

Well all of them are for async support? (OK on different levels in targets). I like the async folder. It's on propose so wide as I prefer the 5-10 classes per folder/namespace.

But it makes things not clear (but even unclear), we should drop it.

snakefoot:
> Well all of them are for async support? 

`AsyncContinuation` are not related to async-tasks, but NLog self-invented future/promise logic, before async/await was invented.

304NotModified:
Yes indeed, so it's not MS.Async, but NLog.Async namespace (or "folder" for now) ;)

snakefoot:
But AsyncHelpersTask is Microsoft Async Task, so now unrelated things are stuffed into the same folder.

304NotModified:
They all have in common to support async in NLog. ;)

But OK, if not convinced, I will merge this one

snakefoot:
> But OK, if not convinced, I will merge this one

Well I'm not convinced. NLog AsyncContinuation are just advanced delegates. Nothing that resembles anything related to System-Tasks async/await.

304NotModified:
OK merged :)

PR #3748, opened 2020-01-01
AsyncTaskTarget - Skip AsyncWrapper since already having internal queue

snakefoot:


304NotModified:
Hi @snakefoot 

I think I asked this a long time before, but after a year with a lot of unit test, DI and mocking experience. 

I approved this PR

But I really like to keep the tests smaller (so less duplication). While this isn't really necessary, it makes changing/debugging/maintaining unit tests really easier. At work I'm now preferring max 10 lines in a unit test.

So that means:

- 1 Act
- helpers for arrange/assert
- setting up the SUT and mocking in the constructor

While that would be (probably) overkill for NLog, I really like to move a bit to it. So for example this PR we could do for `AsyncTaskTargetTest.cs`

1. using always the comments `// Arrange // Act // Assert`
1. ` ConfigurationItemFactory.Default.RegisterType(typeof(AsyncTaskTestTarget)` in ctor (arrange for all tests), and cleanup in class dispose (not sure, depends on the older tests)
1. create helper for creating the XML config (arrange part)
1. create helper for asserting. 


Please give me your opinion :)

snakefoot:
When the dependency injection rewrite has completed, and the LogFactory + ConfigurationItemFactory is completely isolated from other instances, then it would make sense to revisit all the unit-tests.

PR #3750, opened 2020-01-06
ExceptionLayoutRenderer - Improve Format-token parsing error message

snakefoot:
Instead of:

`2020-01-03 20:49:08.5829 Warn Unknown exception data target: String`

Then it becomes:

`2020-01-03 20:49:08.5829 Warn Exception-LayoutRenderer assigned unknown Format: String`


304NotModified:
Thanks!

PR #3754, opened 2020-01-07
LocalIpAddressLayoutRenderer - Higher priority when valid gateway

snakefoot:
Looks like the real-ip will have a default-gateway:

https://stackoverflow.com/questions/8089685/c-sharp-finding-my-machines-local-ip-address-and-not-the-vms

While the VMware Network Adapter will have none or just 0.0.0.0

### DIFF
--- a/src/NLog/LayoutRenderers/LocalIpAddressLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/LocalIpAddressLayoutRenderer.cs
@@ -93,9 +93,10 @@ namespace NLog.LayoutRenderers
                     if (networkScore == 0)
                         continue;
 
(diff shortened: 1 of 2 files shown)


snakefoot:
@jol5 Please check if the pre-release-build is able to resolved the expected `${local-ip}`:

https://ci.appveyor.com/project/nlog/nlog/builds/29958203/artifacts

304NotModified:
Polite bump @jol5

snakefoot:
@304NotModified Don't think jol5 doesn't have English as first language and not in the mood for playing with local nuget-package stores, so it will be an uphill challenge to get any confirmation.

You just have to take my word (or rather stackoverflows word) for that the solution will be working for VM-Ware-Nodes.

304NotModified:
Ok merged!

PR #3755, opened 2020-01-08
Add support for optional size threshold to archive log file on startup

Sam13:
With that feature it's possible to archive the old file on startup when a certain size threshold is reached.

By enabling this you can ensure that the current application session is always in the same file but still have some kind of size based archival. Could also be comined with other archival options.

### DIFF
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -89,7 +89,7 @@ namespace NLog.Targets
         /// <summary>
         /// List of the associated file appenders with the <see cref="FileTarget"/> instance.
         /// </summary>
-        private FileAppenderCache _fileAppenderCache;
+        private IFileAppenderCache _fileAppenderCache;
 
         IFileArchiveMode GetFileArchiveHelper(string archiveFilePattern)
         {
@@ -151,7 +151,11 @@ namespace NLog.Targets
         /// <remarks>
         /// The default value of the layout is: <code>${longdate}|${level:uppercase=true}|${logger}|${message}</code>
         /// </remarks>
-        public FileTarget()
+        public FileTarget() : this(FileAppenderCache.Empty)
+        {
+        }
+        
+        internal FileTarget(IFileAppenderCache fileAppenderCache)
         {
             ArchiveNumbering = ArchiveNumberingMode.Sequence;
             _maxArchiveFiles = 0;
@@ -181,7 +185,7 @@ namespace NLog.Targets
             ForceManaged = false;
             ArchiveDateFormat = string.Empty;
 
-            _fileAppenderCache = FileAppenderCache.Empty;
+            _fileAppenderCache = fileAppenderCache;
             CleanupFileName = true;
 
             WriteFooterOnArchivingOnly = false;
@@ -504,6 +508,8 @@ namespace NLog.Targets
         [Advanced]
         public int ConcurrentWriteAttemptDelay { get; set; }
 
+        private bool? _archiveOldFileOnStartup;
+
         /// <summary>
         /// Gets or sets a value indicating whether to archive old log file on startup.
         /// </summary>
@@ -513,13 +519,17 @@ namespace NLog.Targets
         /// </remarks>
         /// <docgen category='Output Options' order='10' />
         [DefaultValue(false)]
-        public bool ArchiveOldFileOnStartup { get; set; }
+        public bool ArchiveOldFileOnStartup
+        {
+            get => _archiveOldFileOnStartup ?? false;
+            set => _archiveOldFileOnStartup = value;
+        }
 
         /// <summary>
         /// Gets or sets a value of the file size threshold to archive old log file on startup.
         /// </summary>
         /// <remarks>
-        /// This option works only when the "ArchiveOldFileOnStartup" parameter is enabled.
+        /// This option won't work if <see cref="ArchiveOldFileOnStartup"/> is set to <c>false</c>
         /// Default value is 0 which means that the file is archived as soon as archival on
         /// startup is enabled.
         /// </remarks>
@@ -2293,16 +2303,16 @@ namespace NLog.Targets
         /// </summary>
         /// <param name="fileName">File name to be written.</param>
         /// <returns>Decision whether to archive or not.</returns>
-        private bool ShouldArchiveOldFileOnStartup(string fileName)
+        internal bool ShouldArchiveOldFileOnStartup(string fileName)
         {
-            if (!ArchiveOldFileOnStartup)
+            if (_archiveOldFileOnStartup.HasValue)
             {
-                return false;
+                return _archiveOldFileOnStartup.Value;
             }
             // No size threshold specified
             if (ArchiveOldFileOnStartupAboveSize <= 0)
             {
-                return true;
+                return false;
             }
             // Check whether size threshold exceeded
             var length = _fileAppenderCache.GetFileLength(fileName);
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -160,6 +160,7 @@ namespace NLog.Targets
             ConcurrentWriteAttemptDelay = 1;
             ArchiveEvery = FileArchivePeriod.None;
             ArchiveAboveSize = ArchiveAboveSizeDisabled;
+            ArchiveOldFileOnStartupAboveSize = 0;
             ConcurrentWriteAttempts = 10;
             ConcurrentWrites = true;
 #if SILVERLIGHT || NETSTANDARD1_0
@@ -513,6 +514,18 @@ namespace NLog.Targets
         /// <docgen category='Output Options' order='10' />
         [DefaultValue(false)]
         public bool ArchiveOldFileOnStartup { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value of the file size threshold to archive old log file on startup.
+        /// </summary>
+        /// <remarks>
+        /// This option works only when the "ArchiveOldFileOnStartup" parameter is enabled.
+        /// Default value is 0 which means that the file is archived as soon as archival on
+        /// startup is enabled.
+        /// </remarks>
+        /// <docgen category='Output Options' order='10' />
+        [DefaultValue(0)]
+        public long ArchiveOldFileOnStartupAboveSize { get; set; }
 
         /// <summary>
         /// Gets or sets a value specifying the date format to use when archiving files.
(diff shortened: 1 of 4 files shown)


304NotModified:
Thanks!

snakefoot:
Could this code:

```c#
var length = _fileAppenderCache.GetFileLength(fileName);
if (ArchiveOldFileOnStartup && length.HasValue && length.Value > ArchiveOldFileOnStartupAboveSize)
{
    // Do some stuff
}
```

Be changed into this code (Avoids using _fileAppenderCache unless necessary):

```c#
if (ArchiveOldFileOnStartup)
{
   if (ArchiveOldFileOnStartupAboveSize <= 0 || ArchiveOldFileOnStartupAboveSize < _fileAppenderCache.GetFileLength(fileName))
   {
        // Do some stuff
   }
}
```

Could you create an isolated unit-test instead of piggy-backing on an existing one? Let the unit-test-name include the setting `ArchiveOldFileOnStartupAboveSize`.

Sam13:
> Be changed into this code (Avoids using _fileAppenderCache unless necessary):

Ok will do

> Could you create an isolated unit-test instead of piggy-backing on an existing one? Let the new unit-test-name include the setting `ArchiveOldFileOnStartupAboveSize`.

That's what I initially wanted to do but I thought I'd conform to the existing style

snakefoot:
Please extract the new condition logic into a method called ShouldArchiveOldFileOnStartup


There is a lot of technical debt in the NLog unit tests for the FileTarget. But it is hard work to cleanup. But will be easier if not piling more on top :)


Sent from my Sony Xperia


---- Sam13 wrote ----


Be changed into this code (Avoids using _fileAppenderCache unless necessary):

Ok will do

Could you create an isolated unit-test instead of piggy-backing on an existing one? Let the new unit-test-name include the setting ArchiveOldFileOnStartupAboveSize.

That's what I initially wanted to do but I thought I'd conform to the existing style

Sam13:
> There is a lot of technical debt in the NLog unit tests for the FileTarget. But it is hard work to cleanup. But will be easier if not piling more on top :)

I fully agree - all findings addressed

Sam13:
https://travis-ci.org/NLog/NLog/builds/634731645?utm_source=github_status&utm_medium=notification

```
NLog.UnitTests.Targets.PlainFileTargetTests.DateArchive_SkipPeriod(timeKind: Local, archivePeriod: Hour, includeDateInLogFilePath: True, includeSequenceInArchive: True) [FAIL]
      Assert.Equal() Failure
                          ↓ (pos 10)
      Expected: 202001091310.0.txt
      Actual:   202001091309.0.txt
                          ↑ (pos 10)
```

seems to be a flaky unit test and not related to my changes

304NotModified:
FYI Restarted Travis

304NotModified:
Looks good to me :)   agreed @snakefoot ?

snakefoot:
I would probably have made it so setting ArchiveOldFileOnStartupAboveSize was enough alone. But since I will not be using it, then it is not that important. It looks good to me. Maybe include in NLog 4.7?


Sent from my Sony Xperia


Looks good to me :) agreed @snakefoot<https://github.com/snakefoot> ?

Sam13:
> 
> I would probably have made it so setting ArchiveOldFileOnStartupAboveSize was enough alone. 

I thought it's more intuitive to combine with the flag _ArchiveOldFileOnStartup_. Otherwise you can configure the behavior to archive on startup with two different options (setting flag to true or setting threshold >= 0) which may cause confusion.

snakefoot:
> I thought it's more intuitive to combine with the flag ArchiveOldFileOnStartup. 

We also have "archiveAboveSize" and "archiveEvery", and they work right away without first having configured an additional parameter called "ArchiveEnabled". But both our opinions are valid :)

304NotModified:
> I thought it's more intuitive to combine with the flag _ArchiveOldFileOnStartup_. Otherwise you can configure the behavior to archive on startup with two different options (setting flag to true or setting threshold >= 0) which may cause confusion.


I'm afraid it also confusing that ArchiveOldFileOnStartupAboveSize =2 is confusing, at it won't work without ArchiveOldFileOnStartup.

proposal, if we won't set ArchiveOldFileOnStartup, and we do set ArchiveOldFileOnStartupAboveSize , then the default for ArchiveOldFileOnStartup will be true (instead of false).

Or is this proposal making things to complicated?

Sam13:
> I'm afraid it also confusing that ArchiveOldFileOnStartupAboveSize =2 is confusing, at it won't work without ArchiveOldFileOnStartup.
 
You are right but I think it does not matter with proper documentation.

Another option I can think of is to leave _ArchiveOldFileOnStartup_ away since this is basically the same than _ArchiveOldFileOnStartupAboveSize = 0_
Disabling _ArchiveOldFileOnStartup_  then means to unset this option.

Since you are the maintainers please define on how it should behave and I'll try to implement that.

src/NLog/Targets/FileTarget.cs
```diff
@@ -2183,6 +2196,28 @@ private void WriteFooter(string fileName)
...
 
+        /// <summary>
+        /// Decision logic whether to archive logfile on startup.
+        /// <see cref="ArchiveOldFileOnStartup"/> and <see cref="ArchiveOldFileOnStartupAboveSize"/> properties.
+        /// </summary>
+        /// <param name="fileName">File name to be written.</param>
+        /// <returns>Decision whether to archive or not.</returns>
+        private bool ShouldArchiveOldFileOnStartup(string fileName)
+        {
+            if (!ArchiveOldFileOnStartup)
```
snakefoot: Change this to `if (!ArchiveOldFileOnStartup && ArchiveOldFileOnStartupAboveSize <= 0)`

304NotModified: isn't that conflicting with https://github.com/NLog/NLog/pull/3755/files/97b0d69e88637e378fba313af3ac42a3a0664e10#diff-ed0a3fc25ad7408a5b63501fa77dbe69R2212-R2215 ?

304NotModified: (PS @snakefoot I could edit the PR also from the Github GUI, I think that also counts for you)

snakefoot: I guess the pretty version would be this:

```c#
        private bool ShouldArchiveOldFileOnStartup(string fileName)
        {
            // No size threshold specified
            if (ArchiveOldFileOnStartupAboveSize <= 0)
            {
                 return ArchiveOldFileOnStartup;
            } 
```

304NotModified: it's the same behavior, just another version how to write, isn't?

I find it debatable which one is more readable. In terms of performance is no  (significant) difference.

Sam13: > I guess the pretty version would be this:
> 
> ```cs
>         private bool ShouldArchiveOldFileOnStartup(string fileName)
>         {
>             // No size threshold specified
>             if (ArchiveOldFileOnStartupAboveSize <= 0)
>             {
>                  return ArchiveOldFileOnStartup;
>             } 
> ```

Pretty but more complicated in my opinion. My version is straight forward and easy to understand that's why I did it with multiple return statements.

But as I said. Tell me what you want and I'll change it.

304NotModified: No need to change IMO

snakefoot:
Remember to rebase on master to get it on NLog 4.7

Sam13:
> 
> 
> Remember to rebase on master to get it on NLog 4.7
I assumed that all new features land on _dev_ first?
Should I rebase on master and change the PR's base to master?

snakefoot:
DEV branch is targeting NLog ver. 5. So small features and bugfixes are committed into master and merged back to dev on release.


Sent from my Sony Xperia


---- Sam13 wrote ----


Remember to rebase on master to get it on NLog 4.7
I assumed that all new features land on dev first?
Should I rebase on master and change the PR's base to master?

Sam13:
Change base branch to master

304NotModified:
close/reopen to trigger CI

304NotModified:
> You are right but I think every combination with two properties is not that complicated with proper documentation.
> 
> Another option I can think of is to leave _ArchiveOldFileOnStartup_ away since this is basically the same than _ArchiveOldFileOnStartupAboveSize = 0_
> Disabling _ArchiveOldFileOnStartup_ then means to unset this option.
> 
> Since you are the maintainers please define on how it should behave and I'll try to implement that.

Oops this is still open. It still propose: https://github.com/NLog/NLog/pull/3755#issuecomment-576874258

snakefoot:
> Oops this is still open. It still propose: #3755 (comment)

You just said that the PR was [fine as it is](https://github.com/NLog/NLog/pull/3755#discussion_r370291993):

304NotModified:
Well that was a refactor proposal and not a functional change?

snakefoot:
> Well that was a refactor proposal and not a functional change?

No it was a functional change, that implemented the proposed change

304NotModified:
Was it the code for https://github.com/NLog/NLog/pull/3755#issuecomment-576874258? I don't see that.

snakefoot:
Guess my proposal was too complex then :)

304NotModified:
I changed it to https://github.com/NLog/NLog/pull/3755#issuecomment-576874258

Also I made the FIleApenderCache mockable (we should really do more mocking). But a bit afraid of the performance concequence. (interface call vs class class). What do you think @snakefoot ?

snakefoot:
Think the introduction of `_archiveOldFileOnStartup` makes it even more confusing. Think my [suggestion](https://github.com/NLog/NLog/pull/3755#discussion_r369845004) is better.

304NotModified:
No opinion on the IFileApenderCache?

snakefoot:
Think it is the wrong place to mock. Lots of action happening in the file appender cache. Maybe a IFileAppender interface and then inject a factory that returns a mock of the interface


Sent from my Sony Xperia


No opinion on the IFileApenderCache?

304NotModified:
> Think it is the wrong place to mock. Lots of action happening in the file appender cache.

I also checked that, but the appendar cache is doing File stuff (FileInfo) and that is exactly what needs to be "mocked out".

304NotModified:
@snakefoot I you don't think we should performance test this, then I like to merge it. So please let me know if you're OK with that :)

snakefoot:
I don't use any of the features. But I still think your latest change makes the code even more confusing. Anyway it will not hurt performance.


Sent from my Sony Xperia


@snakefoot<https://github.com/snakefoot> I you don't think we should performance test this, then I like to merge it. So please let me know if you're OK with that :)

snakefoot:
Now if having configured archive at startup and archive at startup when above size. Then it will always archive independent of size. This will be confusing. Again my propasal is better.


Sent from my Sony Xperia


@snakefoot<https://github.com/snakefoot> I you don't think we should performance test this, then I like to merge it. So please let me know if you're OK with that :)

snakefoot:
I think there should be a IFileSystem interface that FileTarget should share with FileAppenderCache.


Sent from my Sony Xperia


Think it is the wrong place to mock. Lots of action happening in the file appender cache.

I also checked that, but the appendar cache is doing File stuff (FileInfo) and that is exactly what needs to be "mocked out".

snakefoot:
But maybe this PR is not the right PR for rewriting FileTarget to support mocking.


Sent from my Sony Xperia


Think it is the wrong place to mock. Lots of action happening in the file appender cache.

I also checked that, but the appendar cache is doing File stuff (FileInfo) and that is exactly what needs to be "mocked out".

304NotModified:
> But maybe this PR is not the right PR for rewriting FileTarget to support mocking.

Maybe, but we need tests. 

> Now if having configured archive at startup and archive at startup when above size. Then it will always archive independent of size. This will be confusing.

True, will fix that

304NotModified:
> Now if having configured archive at startup and archive at startup when above size. Then it will always archive independent of size. This will be confusing.

Fixed!

> But maybe this PR is not the right PR for rewriting FileTarget to support mocking.

We could refactor this in another PR. It's all internal.

snakefoot:
Find it strange to write mocking code that you know upfront needs to be rewritten right after.  The mock should cover all file-access and just half.


But yes now archive on startup when above size will work as expected.


Sent from my Sony Xperia


But maybe this PR is not the right PR for rewriting FileTarget to support mocking.

Maybe, but we need tests.

Now if having configured archive at startup and archive at startup when above size. Then it will always archive independent of size. This will be confusing.

True, will fix that

Sam13:
Any ETA when this gets merged and released?

304NotModified:
Will try to merge this weekend and release next week. Release is a bit dependent of the docs (if it's up-to-date already)

304NotModified:
> Find it strange to write mocking code that you know upfront needs to be rewritten right after.

Well I have no time now to make it perfect. 

So I will merge this for now.

PR #3761, opened 2020-01-13
Changed Log4JXmlEventLayoutRenderer to use the same sort order as used in the NdlcLayoutRenderer.

adanek:
``` 
NestedDiagnosticsLogicalContext.Push("foo");
NestedDiagnosticsLogicalContext.Push("bar");
```
rendered with the NdlcLayoutRenderer leads to `foo bar`.
rendered with the Log4JXmlEventLayoutRenderer leads to `bar foo`.

### DIFF
--- a/src/NLog/LayoutRenderers/Log4JXmlEventLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/Log4JXmlEventLayoutRenderer.cs
@@ -63,6 +63,12 @@ namespace NLog.LayoutRenderers
         private static readonly string dummyNLogNamespace = "http://nlog-project.org/dummynamespace/" + Guid.NewGuid();
         private static readonly string dummyNLogNamespaceRemover = " xmlns:nlog=\"" + dummyNLogNamespace + "\"";
 
+        private readonly NdcLayoutRenderer _ndcLayoutRenderer;
+
+#if !SILVERLIGHT
+        private readonly NdlcLayoutRenderer _ndlcLayoutRenderer;
+#endif
+
         /// <summary>
         /// Initializes a new instance of the <see cref="Log4JXmlEventLayoutRenderer" /> class.
         /// </summary>
@@ -75,8 +81,11 @@ namespace NLog.LayoutRenderers
         /// </summary>
         public Log4JXmlEventLayoutRenderer(IAppDomain appDomain)
         {
+            _ndcLayoutRenderer = new NdcLayoutRenderer();
             NdcItemSeparator = " ";
+
 #if !SILVERLIGHT
+            _ndlcLayoutRenderer = new NdlcLayoutRenderer();
             NdlcItemSeparator = " ";
 #endif
 
@@ -189,7 +198,10 @@ namespace NLog.LayoutRenderers
         /// </summary>
         /// <docgen category='Payload Options' order='10' />
         [DefaultValue(" ")]
-        public string NdlcItemSeparator { get; set; }
+        public string NdlcItemSeparator {
+            get => _ndlcLayoutRenderer.Separator;
+            set => _ndlcLayoutRenderer.Separator = value;
+        }
 #endif
 
         /// <summary>
@@ -209,7 +221,10 @@ namespace NLog.LayoutRenderers
         /// </summary>
         /// <docgen category='Payload Options' order='10' />
         [DefaultValue(" ")]
-        public string NdcItemSeparator { get; set; }
+        public string NdcItemSeparator {
+            get => _ndcLayoutRenderer.Separator;
+            set => _ndcLayoutRenderer.Separator = value;
+        }
 
         /// <summary>
         /// Gets or sets the log4j:event logger-xml-attribute (Default ${logger})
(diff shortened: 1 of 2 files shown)


src/NLog/LayoutRenderers/Log4JXmlEventLayoutRenderer.cs
```diff
@@ -340,12 +340,13 @@ private void AppendMdlc(XmlWriter xtw)
...
         }
 
-        private void AppendNdc(XmlWriter xtw)
+        private void AppendNdc(XmlWriter xtw, LogEventInfo logEvent)
         {
             string ndcContent = null;
             if (IncludeNdc)
             {
-                ndcContent = string.Join(NdcItemSeparator, NestedDiagnosticsContext.GetAllMessages());
+                var ndcRenderer = new NdcLayoutRenderer(){ Separator = NdcItemSeparator};
```
snakefoot: Instead of allocating a `new NdcLayoutRenderer` for every LogEvent. Could you put the `_ndcRenderer` as member-variable, or extract an internal method for rendering the NDC-string?

src/NLog/LayoutRenderers/Log4JXmlEventLayoutRenderer.cs
```diff
@@ -356,7 +357,8 @@ private void AppendNdc(XmlWriter xtw)
                     //extra separator
                     ndcContent += NdcItemSeparator;
                 }
-                ndcContent += string.Join(NdlcItemSeparator, NestedDiagnosticsLogicalContext.GetAllMessages());
+                var ndlcRenderer = new NdlcLayoutRenderer(){Separator = NdlcItemSeparator};
```
snakefoot: Instead of allocating a `new NdlcLayoutRenderer` for every LogEvent. Could you put the `_ndlcRenderer` as member-variable, or extract an internal method for rendering the NDLC-string?

tests/NLog.UnitTests/LayoutRenderers/Log4JXmlTests.cs
```diff
@@ -333,5 +334,50 @@ void BadXmlValueTest()
             Assert.Contains("abc", badString);
             Assert.Contains("abc", goodString);
         }
+
+        [Fact]
+        public void NestedDiagnosticContextInformation_is_stored_in_the_same_sort_order_as_used_by_the_nested_diagnostic_layout_renderer()
```
snakefoot: Don't think this test is relevant. As you have already asserted in the test above, that order is as expected.

src/NLog/LayoutRenderers/Log4JXmlEventLayoutRenderer.cs
adanek: Ok, I've choosen to use the member variable, because this formating is the main purpose of the NdlcLayoutRenderer and therefore it would be left as an empty wrapper for this method after extracting the logic.

Do you think, the initialization of the fields should be defered until the first use?

tests/NLog.UnitTests/LayoutRenderers/Log4JXmlTests.cs
adanek: I aggree. Technically they are testing the same thing. 
But I think semantically, the second test makes the desired behavior more clear for other developers.

The old test already had forced the inverse sorting and had to be modified. Perhaps this assert should be removed from the general test? 

In addition I suggest adding another explicit test, which specifies that if both ndc and ndlc are present, they are combined with a `ndcSeparator` (as implemented).

src/NLog/LayoutRenderers/Log4JXmlEventLayoutRenderer.cs
```diff
@@ -75,8 +81,11 @@ public Log4JXmlEventLayoutRenderer() : this(LogFactory.CurrentAppDomain)
         /// </summary>
         public Log4JXmlEventLayoutRenderer(IAppDomain appDomain)
         {
+            _ndcLayoutRenderer = new NdcLayoutRenderer();
```
snakefoot: Could the allocation be in the variable declaration since readonly ?

src/NLog/LayoutRenderers/Log4JXmlEventLayoutRenderer.cs
```diff
@@ -75,8 +81,11 @@ public Log4JXmlEventLayoutRenderer() : this(LogFactory.CurrentAppDomain)
...
             NdcItemSeparator = " ";
+
 #if !SILVERLIGHT
+            _ndlcLayoutRenderer = new NdlcLayoutRenderer();
```
snakefoot: Could the allocation be in the variable declaration since readonly ?

src/NLog/LayoutRenderers/Log4JXmlEventLayoutRenderer.cs
adanek: Done

src/NLog/LayoutRenderers/Log4JXmlEventLayoutRenderer.cs
adanek: Done

304NotModified:
(5.0 as this one is targeting dev)

snakefoot:
> @ snakefoot do you agree?

Yes I think it looks good.

304NotModified:
@adanek we could merge it, but then it scheduled for NLog 5 (unknown ETA). If you prefer NLog 4.7, would it be possible to rebase on dev? (maybe squash first). Of could we help you with that?

snakefoot:
> If you prefer NLog 4.7, would it be possible to rebase on dev? 

Should it not be rebase on master to get it on NLog 4.7 ?

304NotModified:
yes master for NLog 4.7 :angel:

304NotModified:
thanks! squased and rebased on master -> https://github.com/NLog/NLog/pull/3784

PR #3762, opened 2020-01-15
LogFactory - Flush should report to InternalLogger what targets produces timeout

snakefoot:


src/NLog/LogFactory.cs
```diff
@@ -580,9 +583,103 @@ public void Flush(AsyncContinuation asyncContinuation, TimeSpan timeout)
...
+                var flushContinuation = FlushAllTargetsAsync(oldConfig, (ex) =>
+                {
+                    if (ex != null)
+                        lastException = ex;
+                    flushCompletedEvent.Set();
+                }, null);
+
+                bool flushCompleted = flushCompletedEvent.WaitOne(timeout);
+                if (!flushCompleted)
+                    flushContinuation(new TimeoutException("Timeout."));
```
304NotModified: proposal for the TimeoutException message:

"Timeout when flushing all targets, timeout used: X sec"

tests/NLog.UnitTests/LogFactoryTests.cs
```diff
@@ -38,39 +38,49 @@ namespace NLog.UnitTests
...
-                        <logger name='*' minlevel='Debug' writeto='test'></logger>
-                    </rules>
-                </nlog>");
+            Logger logger = LogManager.GetCurrentClassLogger();
+            logger.Info("Prepare Timeout");
 
-                ILogger logger = LogManager.GetCurrentClassLogger();
-                logger.Factory.Flush(_ => { }, TimeSpan.FromMilliseconds(1));
-            }
+            Exception timeout = null;
```
304NotModified: please rename this variable, it's a strange name for an exception ;)

src/NLog/LogFactory.cs
snakefoot: Fine with me. Was just trying to ensure that the user would not see the impact of the change, unless using the InternalLogger.

snakefoot: But it will not change the `TimeoutException` returned by `TimeoutContinuation`. But I think most users will be calling the sync-`Flush()`.

tests/NLog.UnitTests/LogFactoryTests.cs
snakefoot: Fixed

PR #3150, opened 2019-02-26
WIP: DataTableTarget - Bulk-insert using DbDataAdapter and DataTable

snakefoot:
It is very similar to the existing DatabaseTarget in configuration. But with a different `xsi:type="DataTable"`  and a new parameter `UpdateBatchSize`.

```xml
<targets>
    <target name="db" type="AsyncWrapper" timeToSleepBetweenBatches="250" batchSize="200">
            <target name="wrapped_db" xsi:type="DataTable" updateBatchSize="200" keepConnection="True" prepareDbCommand="True">
                  <commandtext>INSERT INTO FooBar VALUES(@msg, @lvl)</commandtext>
                  <parameter name="@msg" layout="${message}" dbType="String" />
                  <parameter name="@lvl" layout="${level}"  dbType="String" />
            </target>
    </target>
</targets>
<rules>
   <logger name="*" minlevel="Debug" writeTo="db" />
</rules>
```

Pre-Release Nuget-package here: https://ci.appveyor.com/project/nlog/nlog/builds/22695687/artifacts (NLog.4.6.0.9175-PR3150.nupkg)

This should work better in cloud-enviroments when high network-latency to the database-instance.

For NetCore using Sql.Data.Client, then make sure to use ver. 4.7.0-preview:  https://www.nuget.org/packages/System.Data.SqlClient/4.7.0-preview.19073.11 (To fix bug https://github.com/dotnet/corefx/issues/29391)

---
This change is [Reviewable](https://reviewable.io/reviews/nlog/nlog/3150)

PR #3764, opened 2020-01-17
DatabaseTarget: Added IsolationLevel option - activates transactions for batch writes

snakefoot:
Alternative version of #3150

Improve performance with help from transactions. DbCommands will not be committed individually but will be committed as a single transaction-chunk. This will reduce the number of requests sent to the database-instance, so less network-latency overhead.

There is one downside. When a single LogEvent in a batch has problems, then it will also cause all other LogEvents in the same batch to fail..

### DIFF
--- a/src/NLog/Targets/DatabaseTarget.cs
+++ b/src/NLog/Targets/DatabaseTarget.cs
@@ -272,6 +272,12 @@ namespace NLog.Targets
         [ArrayParameter(typeof(DatabaseParameterInfo), "parameter")]
         public IList<DatabaseParameterInfo> Parameters { get; } = new List<DatabaseParameterInfo>();
 
+        /// <summary>
+        /// Configures isolated transaction batch writing. If supported by the database, then it will improve insert performance.
+        /// </summary>
+        /// <docgen category='Performance Tuning Options' order='10' />
+        public System.Data.IsolationLevel? IsolationLevel { get; set; }
+
 #if !NETSTANDARD
         internal DbProviderFactory ProviderFactory { get; set; }
 
@@ -555,20 +561,7 @@ namespace NLog.Targets
         {
             try
             {
-                WriteEventToDatabase(logEvent, BuildConnectionString(logEvent));
-            }
-            catch (Exception exception)
-            {
-                InternalLogger.Error(exception, "DatabaseTarget(Name={0}): Error when writing to database.", Name);
-
-                if (exception.MustBeRethrownImmediately())
-                {
-                    throw;
-                }
-
-                InternalLogger.Trace("DatabaseTarget(Name={0}): Close connection because of error", Name);
-                CloseConnection();
-                throw;
+                WriteLogEventToDatabase(logEvent, BuildConnectionString(logEvent));
             }
             finally
             {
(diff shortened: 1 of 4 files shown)


snakefoot:
Did a performance test with the following target:

```
      <target type="AsyncWrapper" name="db" overflowAction="Block" timeToSleepBetweenBatches="0">
        <target type="database" name="db_wrapped" keepConnection="True" isolationLevel="ReadCommitted">
          <connectionString>???</connectionString>
          <commandtext>INSERT INTO Messages VALUES(@ts, @msg, @lvl)</commandtext>
          <parameter name="@ts" layout="${longdate}" dbType="DateTime" />
          <parameter name="@msg" layout="${message}"  dbType="String" />
          <parameter name="@lvl" layout="${level}"  dbType="String" />
        </target>
      </target>
```

And the following table:
```
CREATE TABLE [dbo].[Messages] (
    [TimeStamp] DATETIME       NOT NULL,
    [Message]   NVARCHAR (MAX) NULL,
    [Level]     NCHAR (10)     NULL
);

```

Gave the following results:

| Test Name        | Time (ms) | Msgs/sec  | GC2 | GC1 | GC0 | CPU (ms) | Alloc (MB) |
| PR isolationLevel |    10.122 |     9.879 |   0 |   1 |   3 |    3.531 |      273,0 |
| NLog 4.6.8         |    23.847 |     4.193 |   1 |   2 |   5 |    3.296 |      298,7 |

I'm guessing the performance improvement will be ever better on a high-latency database.

snakefoot:
@D9001 + @aled Not sure if you are still in the NLog DatabaseTarget business for ex. Azure SQL Server. But this pre-release enables the NLog DatabaseTarget to write in batch-mode:

https://ci.appveyor.com/project/nlog/nlog/builds/30201933/artifacts

Just enable the attribute `isolationLevel="ReadCommitted"` for the DatabaseTarget. Any feedback are very welcome. The PR required very few changes, so a very good candidate for the NLog 4.7 release.

aled:
Had a quick poke around the source.

1. It would be nice to be able to configure the max size of the batch.

2. Can you share the performance testing code? I'd like to try it.

3. I find`isolationLevel="ReadCommitted"` less intuitive than, e.g. `UseBatchTransaction="true"` or `TransactionBatchSize="50"`. After all, which isolation level it is, has nothing to do with batching; it would be possible to write a batch of logs with some other isolation level.

snakefoot:
@aled The DatabaseTarget already has batching support. But it doesn't execute the batch using a transaction. The existing batching only had a performance bonus when using `KeepConnection="false"`.

1. Max size of the batch can be controlled by using the [AsyncWrapper](https://github.com/nlog/NLog/wiki/AsyncWrapper-target) with parameter `batchSize="500"`

```
      <target type="AsyncWrapper" name="db" batchSize="500" overflowAction="Block" timeToSleepBetweenBatches="0">
        <target type="database" name="db_wrapped" keepConnection="True" isolationLevel="ReadCommitted">
          <connectionString>...</connectionString>
          <commandtext>INSERT INTO Messages VALUES(@ts, @msg, @lvl)</commandtext>
          <parameter name="@ts" layout="${longdate}" dbType="DateTime" />
          <parameter name="@msg" layout="${message}"  dbType="String" />
          <parameter name="@lvl" layout="${level}"  dbType="String" />
        </target>
      </target>
```

2. The performance test is here https://github.com/NLog/NLog-Performance-tests . Just run the NLogPerformance-project after having adjusted the [NLog.config](url) for the project.

3. `isolationLevel="ReadCommitted"` is the input-parameter given to [DbConnection.BeginTransaction](https://docs.microsoft.com/en-us/dotnet/api/system.data.common.dbconnection.begintransaction). Again batching already exists in the DatabaseTarget this PR just adds parameter to tune performance.

snakefoot:
@aled Would `BatchIsolationLevel` or `BatchingIsolationLevel` give better meaning ?

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -272,6 +272,12 @@ public Layout DBPassword
         [ArrayParameter(typeof(DatabaseParameterInfo), "parameter")]
         public IList<DatabaseParameterInfo> Parameters { get; } = new List<DatabaseParameterInfo>();
 
+        /// <summary>
+        /// Configures isolated transaction batch writing. If supported by the database, then it will improve insert performance.
+        /// </summary>
+        /// <docgen category='Performance Tuning Options' order='10' />
+        public System.Data.IsolationLevel? IsolationLevel { get; set; }
```
304NotModified: Why is IsolationLevel fully qualified? System.Data is already in scope? What do I miss?

snakefoot: It conflicts with [System.Transactions.IsolationLevel](https://docs.microsoft.com/en-us/dotnet/api/system.transactions.isolationlevel)

304NotModified: Ah yes :)

304NotModified: Thanks!

snakefoot:
I'm guessing my question about `BatchIsolationLevel` or `BatchingIsolationLevel` might be a better name, has been answered by the merge? :)

304NotModified:
Oops. Sorry about that. I think the isolation level should also work if you insert a single item. (so name is good, but we don't need the count check)

snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/Database-target

PR #3767, opened 2020-01-18
Mark Condition Expressions as ThreadSafe. Improves concurrency

snakefoot:
Ex. For when-conditions:

`SimpleLayout l = @"${when:when=level<=LogLevel.Info:inner=Good:else=Bad}";`


304NotModified:
👍  thanks!

PR #3771, opened 2020-01-22
Added Layout.CreateFromString with throwConfigExceptions parameter

snakefoot:
Trying to prevent cases where people globally enabled LogManager.ThrowExceptions to perform Layout-validation. See also:

https://github.com/elmahio/elmah.io.nlog/commit/f26377ec38c1c77de73e82e917d0b21e866027f3

And moved FuncLayoutRenderer-tests to SimpleLayoutParserTests.

Could also be convinced to rename it to TryCreateFromString. But then I'm not allowed to throw exception with reason why (and where) it failed.

### DIFF
--- a/src/NLog/Layouts/Layout.cs
+++ b/src/NLog/Layouts/Layout.cs
@@ -127,6 +127,31 @@ namespace NLog.Layouts
         }
 
         /// <summary>
+        /// Implicitly converts the specified string to a <see cref="SimpleLayout"/>.
+        /// </summary>
+        /// <param name="layoutText">The layout string.</param>
+        /// <param name="throwConfigExceptions">Whether <see cref="NLogConfigurationException"/> should be thrown on parse errors (false = replace unrecognized tokens with a space).</param>
+        /// <returns>Instance of <see cref="SimpleLayout"/>.</returns>
+        public static Layout CreateFromString(string layoutText, bool throwConfigExceptions)
+        {
+            try
+            {
+                return new SimpleLayout(layoutText, ConfigurationItemFactory.Default, throwConfigExceptions);
+            }
+            catch (NLogConfigurationException)
+            {
+                throw;
+            }
+            catch (Exception ex)
+            {
+                if (!throwConfigExceptions || ex.MustBeRethrownImmediately())
+                    throw;
+
+                throw new NLogConfigurationException($"Invalid Layout: {layoutText}", ex);
+            }
+        }
+
+        /// <summary>
         /// Implicits converts the specific lambda method into a <see cref="SimpleLayout"/>.
         /// </summary>
         /// <param name="layoutMethod">Method that renders the layout.</param>
--- a/src/NLog/Layouts/LayoutParser.cs
+++ b/src/NLog/Layouts/LayoutParser.cs
@@ -50,7 +50,7 @@ namespace NLog.Layouts
     /// </summary>
     internal static class LayoutParser
     {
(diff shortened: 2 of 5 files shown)


304NotModified:
> Could also be convinced to rename it to TryCreateFromString. But then I'm not allowed to throw exception with reason why (and where) it failed.

I prefer the `TryCreateFromString` (bool return, out parameter) + `CreateFromString` (which throws always an exception - in an error case ;). 

It's really how MS is doing things and a "throw exception flag" is a known code smell AFAIK.

The only reason for not use the try-pattern is consistency which the current API (not the case I think).

So I vote for TryCreateFromString+CreateFromString :)

snakefoot:
Since `CreateFromString` is a new method, then we can change the contract so it throws on error. No problem with that. If that is what you mean?

But then I don't really need `TryCreateFromString` unless you suggest a contract where it can return false, but still return SimpleLayout-object?

304NotModified:
> Since `CreateFromString` is a new method, then we can change the contract so it throws on error. No problem with that. If that is what you mean?

Yes indeed. 



> But then I don't really need `TryCreateFromString` unless you suggest a contract where it can return false, but still return SimpleLayout-object?

Not for this case indeed. But I like it as it's easier to use from the client. (no need to do a catch).

snakefoot:
Ok what method should you call to get the default behavior where SimpleLayout replaces unknown renderers with literals ?

304NotModified:
> Ok what method should you call to get the default behavior where SimpleLayout replaces unknown renderers with literals ?

Ow forgot about that. (I wrote it myself years ago I think, lol).

Maybe we should keep it like this then :)

src/NLog/Layouts/Layout.cs
```diff
@@ -126,6 +126,31 @@ public static Layout CreateFromString(string layoutText)
             return new SimpleLayout(layoutText, ConfigurationItemFactory.Default);
         }
 
+        /// <summary>
+        /// Implicitly converts the specified string to a <see cref="SimpleLayout"/>.
+        /// </summary>
+        /// <param name="layoutText">The layout string.</param>
+        /// <param name="throwConfigExceptions">Whether <see cref="NLogConfigurationException"/> should be thrown on parse errors.</param>
```
304NotModified: Please add that if set to false, then invalid renderers will be replace with a space.

src/NLog/Layouts/LayoutParser.cs
```diff
@@ -431,9 +431,9 @@ private static LayoutRenderer GetLayoutRenderer(ConfigurationItemFactory configu
             }
             catch (Exception ex)
             {
-                if (LogManager.ThrowConfigExceptions ?? LogManager.ThrowExceptions)
+                if (throwConfigExceptions ?? LogManager.ThrowConfigExceptions ?? LogManager.ThrowExceptions)
                 {
-                    throw;
+                    throw;  // TODO throw NLogConfigurationException
```
304NotModified: Is this for NLog 5? Please make it `TODO NLog 5` then :)

src/NLog/Layouts/Layout.cs
snakefoot: Fixed

src/NLog/Layouts/LayoutParser.cs
snakefoot: Fixed

snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/Configure-from-code

PR #3774, opened 2020-01-26
Added missing docgen for public config items

snakefoot:
Fixed build warnings


304NotModified:
Thanks!

(I'm not sure if we are using these docgens. )

snakefoot:
>I'm not sure if we are using these `<docgen>`

https://github.com/NLog/NLog/blob/57d80c05f49dc5a191f2297777e9173eb4b50418/tools/MergeApiXml/NLogApiMerger.cs#L122-L30

Well it looks like it doesn't with the `order`-option, but if using `General Options` or `Layout Options` then they will move to the top in the XSD-Schema-File. But all other categories have priority 100 independent of their specified order.

PR #3779, opened 2020-01-29
`null` layout should marked as FixedText

### LINKED ISSUE #3573: Performance: (string) null to Layout isn't fixed
Given:
```c#
string s = null
Layout l2 =  s;
```

Expected:
- l2 , `isFixedText == true` , `isSimpleText == true`,  `threadsafe == true ` and ` isInitalized== true`

Current:
- l2 , the expected properties/fields has value `false`

image

304NotModified:
fixes #3573


304NotModified:
waiting for #3783

PR #3787, opened 2020-02-01
LogFactory.Setup with fluent registration of condition methods

### LINKED ISSUE #1934: New syntax for registering ConditionMethods via code?
**NLog Version:** 4.4.1
**Platform:** .NET 4.5

Not really sure if this would be considered a bug or a feature or just a question, but it looks to me like there's no suitable manual registration method for ConditionMethods that matches the new syntax style found in the examples for other extensions [here](https://github.com/NLog/NLog/wiki/Register%20your%20custom%20component). I'm trying to prototype some stuff in LINQPad where configuring NLog using code for now and I'm finding it would be nice to be able to point NLog at my static class decorated with ConditionMethods, rather than use the extensions configuration element (probably using the `typeof` format, since static classes can't be used as Generic Type Arguments). It looks like in the meantime, I can probably use the older style like the following example (haven't tested this yet, but it looks like it will probably work), but I thought I'd point out the discrepancy since it looks like there's some movement to provide a different mechanism for registering extensions with code.

```c-sharp
    ConfigurationItemFactory.Default.ConditionMethods.RegisterDefinition("has-diag-category", typeof(NLogConditionMethods).GetMethod("HasDiagnosticCategory", BindingFlags.Static));

### PR DESCRIPTION
**Pull request by snakefoot**
Trying to resolve #1934
```c#
LogManager.Setup().SetupExtensions(s => 
   s.RegisterConditionMethod("hasParameters", evt => evt.Parameters?.Length > 0)
);
```

And ofcourse the usual suspect:
```c#
LogManager.Setup().SetupExtensions(s =>
   s.RegisterConditionMethod("has-diag-category", typeof(NLogConditionMethods).GetMethod("HasDiagnosticCategory", BindingFlags.Static))
);
```

Also optimized the MethodFactory to only scan classes and abstract-classes for static-methods. Skipping value-types and interfaces.

### DIFF
--- a/src/NLog/Conditions/ConditionMethodExpression.cs
+++ b/src/NLog/Conditions/ConditionMethodExpression.cs
@@ -39,34 +39,40 @@ namespace NLog.Conditions
     using System.Reflection;
     using System.Text;
     using NLog.Common;
+    using NLog.Internal;
 
     /// <summary>
     /// Condition method invocation expression (represented by <b>method(p1,p2,p3)</b> syntax).
     /// </summary>
 	internal sealed class ConditionMethodExpression : ConditionExpression
     {
-        private readonly bool _acceptsLogEvent;
         private readonly string _conditionMethodName;
+        private readonly bool _acceptsLogEvent;
+        private readonly ConditionExpression[] _methodParameters;
+        private readonly ReflectionHelpers.LateBoundMethod _lateBoundMethod;
+        private readonly object[] _lateBoundMethodDefaultParameters;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="ConditionMethodExpression" /> class.
         /// </summary>
         /// <param name="conditionMethodName">Name of the condition method.</param>
         /// <param name="methodInfo"><see cref="MethodInfo"/> of the condition method.</param>
+        /// <param name="lateBoundMethod">Precompiled delegate of the condition method.</param>
         /// <param name="methodParameters">The method parameters.</param>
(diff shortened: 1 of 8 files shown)


304NotModified:
proposal: rename to RegisterConditionMethod

snakefoot:
Well `ConditionMethod` are used for filtering. That is why I suggest:

> RegisterFilter**ConditionMethod**

So it is easier for new users to recognize what `ConditionMethod` is used for.

304NotModified:
I understand that. But it's called ConditionMethod  and not FilterConditionMethod (unfortunately).  I prefer keeping one term for one thing and don't use other names in other APIs

snakefoot:
> But it's called ConditionMethod and not FilterConditionMethod (unfortunately).

I'm confused. We are trying to make a new Setup-API and mark the old `ConfigurationItemFactory.Default` as obsolete. Why should the new API continue the errors from something that will be obsolete (And some point removed) ?

304NotModified:
It's not only ConfigurationItemFactory.Default, it also the base class and all the docs on the web .

Also I don't thing adding Filter makes it really that more clear. Filters use conditions and those are not the same.

snakefoot:
Well a change has to start somewhere. But not today. `RegisterConditionMethod` it is.

304NotModified:
> Well a change has to start somewhere.

Yes this is difficult. When doubting I prefer to use the old terms and not adding obsoletes at all. If the name is a 6 out of 10, then if IMO still better dan 2 names.

snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/When-Filter#extensibility

PR #3790, opened 2020-02-03
AsyncTargetWrapper - Do not schedule lazy writer timer, when other active writer

### LINKED ISSUE #3789: AsyncTargetWrapper.ProcessPendingEvents use almost all threads from ThreadPool
**NLog version**: 4.6.7

**Platform**: .NET Framework 4.8

```xml
  <targets async="true">
    <target xsi:type="File"
            name="FatalsOnlyTextLog"
            fileName="${logDirectory}/fatals.txt"
            archiveFileName="${logDirectory}/old/fatals-{#}.zip"
            archiveEvery="Day"
            archiveAboveSize="5242880"
            maxArchiveFiles="450"
            encoding="UTF-8"
            autoFlush="false"
            layout="${fatalLogMessage}"
            enableArchiveFileCompression="true" />

    <target xsi:type="File"
            name="TextLog"
            fileName="${logDirectory}/log.txt"
            archiveFileName="${logDirectory}/old/log-{#}.zip"
            archiveEvery="Day"
            archiveAboveSize="5242880"
            maxArchiveFiles="450"
            encoding="UTF-8"
            autoFlush="false"
            layout="${when:when=(level == LogLevel.Fatal or level == LogLevel.Error):inner=${fatalLogMessage}:else=${logMessage}}"
            enableArchiveFileCompression="true"/>

    <target xsi:type="File"
            name="TraceLog"
            fileName="${logDirectory}/trace.txt"
            archiveFileName="${logDirectory}/old/trace-{#}.zip"
            archiveEvery="Day"
            archiveAboveSize="5242880"
            maxArchiveFiles="450"
            encoding="UTF-8"
            autoFlush="false"
            layout="${when:when=(level == LogLevel.Fatal or level == LogLevel.Error):inner=${fatalLogMessage}:else=${logMessage}}"
            enableArchiveFileCompression="true"/>

    <target xsi:type="ColoredConsole"
            name="ConsoleWriter"
            layout="(${level:uppercase=true})${time}(${threadid}): ${message}${exception:format=tostring:when=(level == LogLevel.Fatal || level == LogLevel.Error)}"
            useDefaultRowHighlightingRules="true">
    </target>
  </targets>
```

- What is the current result?
In very rare cases almost all threads are blocking by method AsyncTargetWrapper.ProcessPendingEvents

No, we didn't know about it

It's very weird bug which we got only twice for 6 months. Application unexpectedly use all threads from ThreadPool. We create dump and found, that most of threads are blocking by AsyncTargetWrapper.ProcessPendingEvents method: 

Dump

If it's important, we have 30-40 log entries per second.

- Are there any workarounds?
No

I think, no

I don't know how it is possible

snakefoot:
Trying to resolve  #3789. Hard to write an unit-test as it is hard to reproduce (Not even sure it is the issue).

### DIFF
--- a/src/NLog/Targets/Wrappers/AsyncTargetWrapper.cs
+++ b/src/NLog/Targets/Wrappers/AsyncTargetWrapper.cs
@@ -359,26 +359,37 @@ namespace NLog.Targets.Wrappers
         /// <returns>Returns true when scheduled a timer-worker-thread</returns>
         protected virtual bool StartInstantWriterTimer()
         {
+            return StartTimerUnlessWriterActive(true);
+        }
+
+        private bool StartTimerUnlessWriterActive(bool instant)
+        {
             bool lockTaken = false;
             try
             {
                 lockTaken = Monitor.TryEnter(_writeLockObject);
                 if (lockTaken)
                 {
-                    // Lock taken means no timer-worker-thread is active writing, schedule timer now
-                    lock (_timerLockObject)
+                    // Lock taken means no other timer-worker-thread is trying to write, schedule timer now
+                    if (instant)
                     {
-                        if (_lazyWriterTimer != null)
+                        lock (_timerLockObject)
                         {
-                            // Not optimal to schedule timer-worker-thread while holding lock,
-                            // as the newly scheduled timer-worker-thread will hammer into the writeLockObject
-                            _lazyWriterTimer.Change(0, Timeout.Infinite);
-                            return true;
+                            if (_lazyWriterTimer != null)
+                            {
+                                // Not optimal to schedule timer-worker-thread while holding lock,
+                                // as the newly scheduled timer-worker-thread will hammer into the writeLockObject
+                                _lazyWriterTimer.Change(0, Timeout.Infinite);
+                                return true;
+                            }
                         }
                     }
+                    else
+                    {
+                        StartLazyWriterTimer();
+                        return true;
+                    }
                 }
-
-                return false;
             }
             finally
             {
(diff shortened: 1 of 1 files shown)


304NotModified:
Could we reproduce it locally?

src/NLog/Targets/Wrappers/AsyncTargetWrapper.cs
```diff
@@ -358,27 +358,37 @@ protected virtual void StartLazyWriterTimer()
...
+                                // Not optimal to schedule timer-worker-thread while holding lock,
+                                // as the newly scheduled timer-worker-thread will hammer into the writeLockObject
+                                _lazyWriterTimer.Change(0, Timeout.Infinite);
+                                return true;
+                            }
                         }
                     }
+                    else
+                    {
+                        StartLazyWriterTimer();
```
304NotModified: shouldn't we return `true` here?

snakefoot:
> Could we reproduce it locally?

Not anywhere near the amount that the issue showed.

src/NLog/Targets/Wrappers/AsyncTargetWrapper.cs
snakefoot: The return value is whether it successfully scheduled an instant timer, but no harm in returning true here.

304NotModified:
OK, I think we should merge it as it won't hurt any way? Or do you like to wait for a confirmation?

snakefoot:
I think it should just be merged. Since all evidence was lost (would really have liked to see the internal logger output) and apparently not easy to re-produce.

304NotModified:
merged 👍  thanks!

PR #3793, opened 2020-02-04
Added ${dir-separator}

304NotModified:


src/NLog/LayoutRenderers/DirectorySeparatorLayoutRenderer.cs
```diff
@@ -0,0 +1,74 @@
...
+    using NLog.Internal;
+    using NLog.Config;
+
+    /// <summary>
+    /// The OS dependent directory separator
+    /// </summary>
+    [LayoutRenderer("dir-separator")]
+    [ThreadAgnostic]
+    [ThreadSafe]
+    public class DirectorySeparatorLayoutRenderer : LayoutRenderer, IStringValueRenderer, IRawValue
```
snakefoot: You can also add AppDomain-attribute

src/NLog/LayoutRenderers/DirectorySeparatorLayoutRenderer.cs
```diff
@@ -0,0 +1,74 @@
...
+        protected override void Append(StringBuilder builder, LogEventInfo logEvent)
+        {
+            builder.Append(SeparatorChar);
+        }
+
+        /// <inheritdoc/>
+        string IStringValueRenderer.GetFormattedString(LogEventInfo logEvent) => Separator;
+
+        /// <inheritdoc />
+        public bool TryGetRawValue(LogEventInfo logEvent, out object value)
```
snakefoot: Think the raw-value should return the string and not the char.

src/NLog/LayoutRenderers/DirectorySeparatorLayoutRenderer.cs
304NotModified: Thanks! I assume you mean `[AppDomainFixedOutput]` (forgot about that). 

Fixed!

src/NLog/LayoutRenderers/DirectorySeparatorLayoutRenderer.cs
304NotModified: The raw value **is** a char. Why should it be a string?

304NotModified:
I assume this is still valid advise, so remove that interface / implementation. Nice check :)

image

src/NLog/LayoutRenderers/DirectorySeparatorLayoutRenderer.cs
snakefoot: Because that is what people expect. No one uses single char.

snakefoot: How does rawvalue work with literallayoutrenderer? Does it return single char when only a single letter?

304NotModified: Now, and that because literallayoutrenderer isn't implementing IRawValue

304NotModified: anyway, you could use this in char columns in the database as well as strings. But OK, that's not really an use full case.

304NotModified: It's char because the underlining type is char. I think that is a good practice?

snakefoot: Oh well I guess we will never know since it will always be turned into a literallayoutrenderer :)

Issue #2710, opened 2018-05-09
Feature request: Add event handler in LogManager that is invoked in case of internal NLog/Extension exceptions


katdti:
Hi

If a logging target fails due to socket/http connection errors - it should be possible to detect this from  apps using NLog - this would allow the app to e.g. inform the user via the GUI,
The app should be able to attach to an event handler and get a callback first time problem is detected and a callback when the connection problem is solved and first successfully log to server was performed.

Optimal would be:
- if the internal logging stop logging the same error until the error disappears.
- any failed writes are cached until connection is back online

**Additional Information**:

NLog version: 4.5.3
Platform: Net 4.5

Sample configuration from PC that genrated an exception due to https sequity rules on local PC.
Exception comes at almost every logging.
```xml
<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
        throwExceptions="false" internalLogLevel="Warn" internalLogToConsole="false" autoReload="false"
        internalLogFile="C:/Temp/communications.log"
    >
    <extensions>
         <add assembly="NLog.Targets.Seq"/>
    </extensions>
<targets async="true">
      <target name="seq" xsi:type="Seq" serverUrl="https://localhost:5341" apiKey="">
      </target> 
</targets>
    <rules>
      <logger name="*" minlevel="Trace" writeTo="seq" />
    </rules>
</nlog>
```

The exception generated:
```
Error has been raised. Exception: System.Net.WebException: The underlying connection was closed: An unexpected error occurred on a send. ---> System.IO.IOException: The handshake failed due to an unexpected packet format.
   at System.Net.Security.SslState.StartReadFrame(Byte[] buffer, Int32 readBytes, AsyncProtocolRequest asyncRequest)
   at System.Net.Security.SslState.StartReceiveBlob(Byte[] buffer, AsyncProtocolRequest asyncRequest)
   at System.Net.Security.SslState.CheckCompletionBeforeNextReceive(ProtocolToken message, AsyncProtocolRequest asyncRequest)
   at System.Net.Security.SslState.StartSendBlob(Byte[] incoming, Int32 count, AsyncProtocolRequest asyncRequest)
   at System.Net.Security.SslState.ForceAuthentication(Boolean receiveFirst, Byte[] buffer, AsyncProtocolRequest asyncRequest)
   at System.Net.Security.SslState.ProcessAuthentication(LazyAsyncResult lazyResult)
[...]

Issue #1193, opened 2016-01-25
add onException to catch exceptions in targets etc


304NotModified:
add Configuration.OnException to handle exceptions async.

snakefoot:
Maybe introduce a InternalLogger-delegate that includes origin-target-type + exception-object. See also #2710

snakefoot:
Resolved by #3796 that added `InternalLoggerLogMessageReceived`-event.

Improved by 3823 that adds context with target-type + target-name. And improved further by #4256 to make it easier for all targets to provide proper context.

PR #3796, opened 2020-02-05
InternalLogger: added LogMessageReceived event

### LINKED ISSUE #3350: Add internallogger event(s)
Easier to subscribe than  ` InternalLogger.LogWriter`

304NotModified:
Fixes https://github.com/NLog/NLog/issues/2710, first steps for #3350 and https://github.com/NLog/NLog/issues/1193

@snakefoot please review, also the names. 

PS: I started with a delegate, but switched to EventHandler because of this (and also we use already in NLog (e.g. LogManager.ConfigurationChanged)

> For scenarios where the EventHandler and EventHandler<TEventArgs> delegates do not work, you can define a delegate. Scenarios that require you to define a delegate are very rare, such as when you must work with code that does not recognize generics. 

https://docs.microsoft.com/en-us/dotnet/standard/events/index?view=netframework-4.8

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ install:
   # Restore dependencies
   - dotnet restore src/NLog
   - dotnet restore tests/NLogAutoLoadExtension
+  - dotnet restore tests/ManuallyLoadedExtension
 
   - dotnet restore tests/SampleExtensions
 
--- a/src/NLog.sln
+++ b/src/NLog.sln
@@ -31,6 +31,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "TestAssets", "TestAssets", 
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Packages", "Packages", "{194AB4BD-C817-4C59-A86C-49D89B8C3A64}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ManuallyLoadedExtension", "..\tests\ManuallyLoadedExtension\ManuallyLoadedExtension.csproj", "{B9714BBB-A9F6-429B-ACC1-961C16643057}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
@@ -73,6 +75,10 @@ Global
 		{1A86A7FA-1C2E-4745-87FC-FD3F583DBAD2}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{1A86A7FA-1C2E-4745-87FC-FD3F583DBAD2}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{1A86A7FA-1C2E-4745-87FC-FD3F583DBAD2}.Release|Any CPU.Build.0 = Release|Any CPU
+		{B9714BBB-A9F6-429B-ACC1-961C16643057}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{B9714BBB-A9F6-429B-ACC1-961C16643057}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{B9714BBB-A9F6-429B-ACC1-961C16643057}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{B9714BBB-A9F6-429B-ACC1-961C16643057}.Release|Any CPU.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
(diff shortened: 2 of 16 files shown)


304NotModified:
again "No output has been received in the last 10m0s" in Travis

304NotModified:
@snakefoot please your opinion :)

snakefoot:
I was thinking to extend with context info. So having Target-Type and Target-Info (Target ToString output) as info. Then remove the need to manual write it every time when logging but just provide "this" in as parameter.


But I have been waiting for the Dependency Injection adventure to complete, along with all other major changes so it happens just before NLog 5 is released, because it will make any merging from master to dev conflict almost by guarantee (will include rewrite of all internal logging statements)


Sent from my Sony Xperia


@snakefoot<https://github.com/snakefoot> please your opinion :)

snakefoot:
Instead of calling it ReceivedInternalLogEventArgs then I just prefer InternalLoggerEventArgs


Sent from my Sony Xperia


@snakefoot<https://github.com/snakefoot> please your opinion :)

snakefoot:
Please move this for NLog 5.0

304NotModified:
> Please move this for NLog 5.0

Because of 

> I was thinking to extend with context info. 

?

Can't we add that to NLog 5 afterwards? Doesn't sound like a breaking change

snakefoot:
> > Please move this for NLog 5.0

>  Because of?

Well the issues this pull-request promise to resolve, do mention that it is important to have details of what target-type and target-name the event comes from. This is not included in this PR.

304NotModified:
true if fixes #2710, and are the first steps for #3350 and #1193

304NotModified:
> Instead of calling it `ReceivedLogEvent` and `ReceivedInternalLogEventArgs` then I prefer `OnLogEvent` and `InternalLoggerLogEventArgs`

- Renamed the ReceivedInternalLogEventArgs to InternalLoggerLogEventArgs
- I like also the On... name, but that's inconsistent with LogManager.ConfigurationChanged and LogManager.ConfigurationReloaded. We haven't anywhere a On... method in NLog currently. AFAIK MS is also inconsistent with On and not On ;). So I left it like this

304NotModified:
@snakefoot this is the latest for me for 4.7. If you think some things must be included for NLog 4.7, please let me know. Thanks!

snakefoot:
Problem is that it is an InternalLogger event and not a Logevent. Find received-logevent very confusing


Sent from my Sony Xperia


Instead of calling it ReceivedLogEvent and ReceivedInternalLogEventArgs then I prefer OnLogEvent and InternalLoggerLogEventArgs

  *   Renamed the ReceivedInternalLogEventArgs to InternalLoggerLogEventArgs
  *   I like also the On... name, but that's inconsistent with LogManager.ConfigurationChanged and LogManager.ConfigurationReloaded. We haven't anywhere a On... method in NLog currently. AFAIK MS is also inconsistent with On and not On ;). So I left it like this

304NotModified:
ReceivedInternalLogEvent then?

304NotModified:
but it's on the InternalLogger already?

snakefoot:
I like `ReceivedInternalLogEvent` :)

snakefoot:
Maybe call it `InternaLogger.RaiseEvent` or `InternaLogger.NotifyEvent` or `InternaLogger.NotifyEvent`

304NotModified:
Events should be in past tense?

304NotModified:
> Also think received is strange, since they are generated. Never seen "received" as event-notification, so it is actually the "received" part that I don't like.

https://docs.microsoft.com/en-us/dotnet/api/system.io.ports.serialport.datareceived

The internallogger received the generated event. But for sure LogEventRecieved is better than RecievedLogEvent

304NotModified:
EventWritten? (unfortunately written sounds like "written to file")

304NotModified:
related: https://stackoverflow.com/questions/724085/events-naming-convention-and-style

304NotModified:
I think the best candidates are LogEventReceived or LogMessageReceived. But im open for other suggestions

snakefoot:
I agree if InternalLogger received its events from another machine through a network connection like the serial cable then the name received is good.


But I don't think NLog receives any events. It has generated / raised them by it self.


Sent from my Sony Xperia


I think the best candidates are LogEventReceived or LogMessageReceived. But im open for other suggestions

snakefoot:
May call it LogWritten and ensure the event is raised after other destinations (Like LogWriter) have been notified.


Sent from my Sony Xperia


I think the best candidates are LogEventReceived or LogMessageReceived. But im open for other suggestions

304NotModified:
> Since the InternalLogger can write its messages to a lot of destinations (LogWriter + LogFile + Console + Trace). Then I don't think "Written" automatically means to file. See also: https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.eventlog.entrywritten

This is exactly why "written" is wrong here.

A target writes/calls to the internal logger. The internal logger writes to the console/file etc.

So these events would be written: ConsoleWritten, FileWritten. But that isn't the case here, as the event is trigger on receiving a log message and not after a write to console/file/db/eventlog whatever.

snakefoot:
Well to me `LogMessageReceived` is completely wrong, because it sounds like it is triggered when NLog receives a LogMessage from the application. Anyway agree to disagree.

Issue #1959, opened 2017-02-13
Change maxInnerExceptionLevel default


304NotModified:
https://github.com/NLog/NLog/wiki/Exception-Layout-Renderer

> maxInnerExceptionLevel - Maximum number of inner exceptions to include in the output. By default inner exceptions are not enabled for compatibility with NLog 1.0.

The default should be unlimited or something like that.

snakefoot:
Would be better if the default exception-layout was ToString().

304NotModified:
Or maybe toString,data

snakefoot:
@304NotModified Resolved with  #3458 ? (NLog 5.0)

304NotModified:
Have to check that, this is a bit more complicated

PR #3797, opened 2020-02-05
${Exception} Set default of MaxInnerExceptionLevel to 10

304NotModified:
fixes https://github.com/NLog/NLog/issues/1959

### DIFF
--- a/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
@@ -96,7 +96,7 @@ namespace NLog.LayoutRenderers
             Separator = " ";
             ExceptionDataSeparator = ";";
             InnerExceptionSeparator = EnvironmentHelper.NewLine;
-            MaxInnerExceptionLevel = 0;
+            MaxInnerExceptionLevel = 10;
 
             _renderingfunctions = new Dictionary<ExceptionRenderingFormat, Action<StringBuilder, Exception>>()
                                                                                                     {
@@ -167,10 +167,9 @@ namespace NLog.LayoutRenderers
 
         /// <summary>
         /// Gets or sets the maximum number of inner exceptions to include in the output.
-        /// By default inner exceptions are not enabled for compatibility with NLog 1.0.
         /// </summary>
         /// <docgen category='Rendering Options' order='10' />
-        [DefaultValue(0)]
+        [DefaultValue(10)]
         public int MaxInnerExceptionLevel { get; set; }
 
         /// <summary>
--- a/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
@@ -289,9 +289,20 @@ namespace NLog.LayoutRenderers
 
         private void AppendInnerException(Exception currentException, StringBuilder builder)
         {
-            // separate inner exceptions
+            var formats = InnerFormats;
+            if (formats == null)
+            {
+                //Don't flow toString to inner exceptions if used for out exception
+                formats = new List<ExceptionRenderingFormat>(Formats);
+                formats.Remove(ExceptionRenderingFormat.ToString);
+            }
+
+            if (formats.Count == 0)
+            {
+                return;
+            }
             builder.Append(InnerExceptionSeparator);
-            AppendException(currentException, InnerFormats ?? Formats, builder);
+            AppendException(currentException, formats, builder);
         }
 
         private void AppendException(Exception currentException, List<ExceptionRenderingFormat> renderFormats, StringBuilder builder)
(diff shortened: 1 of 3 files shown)


snakefoot:
Don't understand the need for this change. It will cause havoc if using `Format=ToString`

304NotModified:
It will fix the case of ${exception:format=message,stacktrace}, which won't show the innerexception now.

But maybe we should rename it to ${all-exceptions} and keep ${exception} untouched

snakefoot:
There is an external package that displays inner exceptions in nicely intended format


Sent from my Sony Xperia


It will fix the case of ${exception:format=message,stacktrace}, which won't show the innerexception now.

But maybe we should rename it to ${all-exceptions} and keep ${exception} untouched

snakefoot:
Please make a test that `format=tostring` doesn't go crazy.

304NotModified:
Good one. It was bad (but of course also when setting MaxInnerExceptionLevel yourself in combination with tostring)

Unfortunately, tostring is still a bit strange with MaxInnerExceptionLevel and other innerException settings.

I think it's better now, but not perfect. Also this is a bit strange (default) layout:

```c#
[Fact]
public void ExceptionWithInnerDataAndToStringAndData()
{
    // Arrange
    var renderer = new ExceptionLayoutRenderer { Format = "ToString,Data" };
    var innerException = new Exception("inner") { Data = { { "inner1", "value1" }, { "inner2", "value2" } } };
    Exception ex = new Exception("outer", innerException) { Data = { { "outer1", "value1b" }, { "outer2", "value2b" } } };
    var logEventInfo = LogEventInfo.Create(LogLevel.Error, "logger1", ex, null, "message1");

    // Act
    var result = renderer.Render(logEventInfo);

    //Assert
    Assert.Equal("System.Exception: outer ---> System.Exception: inner\r\n" +
                 "   --- End of inner exception stack trace --- outer1: value1b;outer2: value2b\r\n" +
                 "inner1: value1;inner2: value2", result);
}
```

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -289,9 +289,20 @@ private int AppendInnerExceptionTree(Exception currentException, int currentLeve
 
         private void AppendInnerException(Exception currentException, StringBuilder builder)
         {
-            // separate inner exceptions
+            var formats = InnerFormats;
+            if (formats == null)
+            {
+                //Don't flow toString to inner exceptions if used for out exception
+                formats = new List<ExceptionRenderingFormat>(Formats);
```
snakefoot, line 296: Think it is sad that you allocate a new list for every LogEvent. Could it not be created once, when parsing the Format?

304NotModified: Can do. Wasn't sure if the approach now is good enough

snakefoot: Ok, then I will wait until it is good enough :)

304NotModified: Do you think this approach is good enough?

snakefoot: I was thinking that if were going to dive into the Exception-object and display all inner-exceptions with details, then it would be nice it would be displayed nicely. Like here https://github.com/oleksabor/nlog.indentexception

Since the layout capability of the Format-specifier is rather limited, then it I think works best with a single exception-object. But besides `Format=ToString`, then you probably should also handle `Format=@` (Json)

snakefoot: But actually I think it is a bad approach to modify Format-specifier for inner exceptions. Instead MaxInnerExceptionLevel should be nullable. And when not configured explicitly then it should default to 10 (But if Format contains ToString or @ then it should default to 0)

snakefoot: Another example where prettify exception is better than Exception.ToString: https://codereview.stackexchange.com/questions/147768/exceptionprettifier-for-prettier-exception-strings

304NotModified: > Another example where prettify exception is better than Exception.ToString: https://codereview.stackexchange.com/questions/147768/exceptionprettifier-for-prettier-exception-strings

Agree on that one, but sounds like a scope creep to me

snakefoot: Yes. But trying to argue for that changing `MaxInnerExceptionLevel` gives more side-effects than wins. Maybe making this improvement by doing something new/different.

304NotModified: I agree this is difficult. I think ${exception} has at least 2 use cases, and that's impossible to handle in one renderer.

So maybe we need to split it to separate renderers, like ${fancy-exception} and ${all-exceptions).

Any way the issue here I try to solve is that people are lossing their inner exceptions. That's now partly covered with the ToString, but not for the exception data

snakefoot: There is a "gateway" in the `Format`-property for `${exception}`, which can be (ab)used like I did when adding `format=@` (automatically renders all inner-exceptions even when not requested).

304NotModified:
inactive so close for now

Issue #2871, opened 2018-08-24
Drop Silverlight and WinPhone8 support


304NotModified:

snakefoot:
Silvelight ver. 5 has end-of-life support until 10/12/2021:

https://support.microsoft.com/en-us/lifecycle/search/12905

Also the code cleanup will not be that impressive until Xamarin iOS and Xamarin Android start using NetStandard (Then the code cleanup will be real).

304NotModified:
> Silvelight ver. 5 has end-of-life support until 10/12/2021:
> 
> https://support.microsoft.com/en-us/lifecycle/search/12905
> 
> Windows Phone 8.1 has end-of-life support until July 1, 2023. But yes new WP8 Apps will not be accepted in Windows Store after 31 Oct. 2018

Thanks for the details!



> Don't hope NLog 5 is first in 3-5 years.

What do you mean with that?



> Also the code cleanup will not be that impressive until Xamarin iOS and Xamarin Android start using NetStandard (When those are also removed, then the code cleanup will be real).

True, but it also saves build time. For me the build time is too long (also because of the unstable tests).

snakefoot:
It takes 10 secs to build and package. So I don't think you will see much improvement there



Sent from my Samsung device


-------- Original message --------

snakefoot:
Starting to convince myself that we should remove Silverlight + Xamerin platforms with Nlog ver 5.0 (Maybe also remove ne40), and instead introduce net472 with NLog 5.0

snakefoot:
Another reason to drop Xamarin, Silverlight, etc.:

#3703

But we should make sure to use ConcurrentWrites=false by default, since Mutex-usages is not welcome on those platforms. See also #3048 + #2742

304NotModified:
:+1:

PR #3818, opened 2020-02-26
Added .NET Framework 4.6 as first-class citizen

### LINKED ISSUE #3817: Add net46 Target Framework
**NLog version**: 4.6.8

**Platform**: .NET Framework 4.8, .NET Core 3.1

I'm having an issue with a dual-targeting unit test project where the .NET Standard tests succeed, but the .NET Framework tests fail to run with the following error:

An exception occurred while invoking executor 'executor://mstestadapter/v2': Unable to find assembly 'NLog, Version=4.0.0.0, Culture=neutral, PublicKeyToken=5120e14c03d0593c'.

After this error no tests run. It appears this may be related to the use of `CallContext.LogicalSetData` (per [this issue](https://github.com/microsoft/testfx/issues/180)). We recently switched all our usage of `CallContext` to `AsyncLocal` which has alleviated a number of AppDomain-specific issues that don't apply to how we intend to use it and I figured your .NET Standard implementation was already using it, so I figured I'd go see what the code looked like when I noticed [this line](https://github.com/NLog/NLog/blob/86fba05e272a07d508b32475957b9310f607061d/src/NLog/Contexts/MappedDiagnosticsLogicalContext.cs#L390) which shows you were already prepared to make this migration, but TargetFrameworks didn't contain net46 so this isn't shipped in the NuGet package. 

I pulled the source locally and made the couple small modifications necessary to test this and when I build the package with a net46 target included it solves my issues so I'm hoping you'll consider including the net46 target in the 4.7 release.

Thanks kindly.

TheXenocide:
Fixes #3817 - added .NET Framework 4.6 (net46) Target Framework to switch from CallContext to AsyncLocal for contexts

Note that locally there are a few unit tests failing locally. I didn't think to run them before making my changes, so I'll try to circle back around to see if they were already broken locally, but this is the majority of the work either way.

For 4.5, 4.6, and .NET Standard 2.0 the following tests are broken locally:
```
NLog.UnitTests.LoggerTests.MixedStructuredEventsConfigTest(parseMessageTemplates: null, param1: "0", param2: "@Client")
NLog.UnitTests.LoggerTests.MixedStructuredEventsConfigTest(parseMessageTemplates: null, param1: "1", param2: "@Client")
```

For 4.5 and 4.6 the following test is also broken locally:
```
NLog.UnitTests.Internal.AppDomainPartialTrustTests.MediumTrustWithExternalClass
```

### DIFF
--- a/src/NLog/Abstractions/ISuppress.cs
+++ b/src/NLog/Abstractions/ISuppress.cs
@@ -34,7 +34,7 @@
 namespace NLog
 {
     using System;
-#if NET4_5
+#if NET4_5 || NET4_6
     using System.Threading.Tasks;
 #endif
 
@@ -68,7 +68,7 @@ namespace NLog
         /// <returns>Result returned by the provided function or fallback value in case of exception.</returns>
         T Swallow<T>(Func<T> func, T fallback);
 
-#if NET4_5
+#if NET4_5 || NET4_6
         /// <summary>
         /// Logs an exception is logged at <c>Error</c> level if the provided task does not run to completion.
         /// </summary>
--- a/src/NLog/Config/AdvancedAttribute.cs
+++ b/src/NLog/Config/AdvancedAttribute.cs
@@ -40,7 +40,7 @@ namespace NLog.Config
     /// default in generated documentation.
     /// </summary>
     [AttributeUsage(AttributeTargets.Property)]
-#if NET4_0 || NET4_5 && !NETSTANDARD1_0
+#if NET4_0 || (NET4_5 || NET4_6) && !NETSTANDARD1_0
     [System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage]
 #endif
     public sealed class AdvancedAttribute : Attribute
--- a/src/NLog/Config/ConfigurationItemFactory.cs
+++ b/src/NLog/Config/ConfigurationItemFactory.cs
@@ -566,7 +566,7 @@ namespace NLog.Config
         private void RegisterExternalItems()
         {
 
-#if NET4_5 || NETSTANDARD
+#if NET4_5 || NET4_6 || NETSTANDARD
             _layoutRenderers.RegisterNamedType("configsetting", "NLog.Extensions.Logging.ConfigSettingLayoutRenderer, NLog.Extensions.Logging");
 #endif
         }
--- a/src/NLog/Contexts/MappedDiagnosticsLogicalContext.cs
+++ b/src/NLog/Contexts/MappedDiagnosticsLogicalContext.cs
@@ -54,7 +54,7 @@ namespace NLog
         private sealed class ItemRemover : IDisposable
         {
             private readonly string _item1;
-#if NET4_5
+#if NET4_5 || NET4_6
             // Optimized for HostingLogScope with 3 properties
             private readonly string _item2;
             private readonly string _item3;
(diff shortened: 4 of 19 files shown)


TheXenocide:
Okay, I checked the tests and the same tests are failing locally before my changes, so I think this commit is passing all checks that were passing before.

snakefoot:
Think net46 should be added to the test-collection:

https://github.com/NLog/NLog/blob/dev/run-tests.ps1

But as repeated before, then it would be nice to remove other builds before adding net46. See also #2871

304NotModified:
> Also think the failing builds are caused by the new build-pipeline for VS2019.

Yes removed that for now.

@TheXenocide are there now with the PR features for .NET 4.6 that aren't in  the .NET 4.5 build

snakefoot:
>  If the features are the same, we could skip the .NET 4.6 tests for now.

Features should be the same, if NET4_6 defines have been applied properly. That is what tests confirms.

The important change between NET4_5 and NET4_6 is the use of AsyncLocal for NDLC and MDLC. Meaning NET4_6 works just like NET4_5 and NDLC + MDLC will no longer travel between AppDomains (good or bad).

304NotModified:
> Meaning NET4_6 works just like NET4_5 and NDLC + MDLC will no longer travel between AppDomains (good or bad).

Is this a breaking behavior change?

snakefoot:
I'm guessing it is breaking behavior change. But remoting is a thing of the past. So I'm for the change going forward. But I'm also for removing NET4_0 :)

304NotModified:
Yes ok, but this is for NLog 5 then

TheXenocide:
That's correct, it doesn't change the NLog-specific feature profile, but it does remove cross-`AppDomain` support from NDLC/MDLC by switching to `AsyncLocal`. It's worth noting that the precompiler symbols and conditions were already setup for that particular feature swap, I didn't change that when I turned the target on, though I did have to add NET4_6 to other conditions to keep other features. It may be worth considering if those other features are needed anymore or not, but I didn't go to that extent since this feature very specifically had conditions around it, whereas other features didn't seem to indicate one way or the other. As is, this pull request assumes all 4.5 features stay, but not all of them exist in .NET Standard 2.0. 

As for the concept of the feature difference more generally, unfortunately `CallContext` was not designed for ambient context, it's just what most of us have been using it for; it was designed specifically for Remoting and using it in a new world of async/await and Task can result in scenarios where `ExecutionContext` flow is polluted with `CallContext` data in a way that cannot be easily reclaimed. In my case this caused issues with MSTest v2 because the isolated parent domain of the test adapter could not load NLog types, but some portion of the `CallContext` required it to. In my code I am disposing all contexts I create and even tried clearing the MDLC and NDLC contexts during `AssemblyCleanup`, but unfortunately nothing I could get at prevented MSTest from trying to load NLog in the parent domain. Each test assembly runs in its own `AppDomain` and it explicitly intends to isolate itself from assemblies required by the test project (which is one of the reasons it's using `AppDomain`s in the first place). 

In honesty they could probably better isolate tests by controlling `ExecutionContext` flow in a way that prevents this type of pollution, but I think they probably also don't expect people to leave uncleaned logical thread affinitives around. We've had problems in MSTest in the past with our own uses of `CallContext` causing AppDomainDisposedException as well (even though all of our uses were applied via `using (SomeDisposableContext) { ... }`. It seems like not all TestMethods run in the same context and sometimes new empty contexts are introduced as well. Switching our stuff to AsyncLocal allowed us to keep all of our ambient context functionality without much compromise and we basically never need our contexts to flow across AppDomain boundaries (even though we actually do have an application that creates multiple domains). As an interesting side note, when using NLog in said multi-domain application we already had to go out of our way to configure it differently since each domain had different instances of the file targets (we now use the `AppDomain` `Id` in the filename).

snakefoot:
Created  #4223 where Travis-Build also works

PR #3819, opened 2020-02-27
Version 4.7.0

snakefoot:
Happy with the performance testing

https://github.com/maciejw/logger-perf-tests/issues/1#issuecomment-591066913


304NotModified:
We need to "sync" the csproj changelog I guess, maybe after #3821

PR #3824, opened 2020-03-01
Azure AD Access Token support for DatabaseTarget

anxkha:
Provide support for setting an Azure AD Access Token when creating a SqlConnection
Add a package reference to System.Data.SqlClient for the netstandard2.0 target
Update the System.Security.Principal.Windows package version in support of adding the System.Data.SqlClient package

### DIFF
--- a/src/NLog.WindowsIdentity/NLog.WindowsIdentity.csproj
+++ b/src/NLog.WindowsIdentity/NLog.WindowsIdentity.csproj
@@ -54,7 +54,7 @@
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
-    <PackageReference Include="System.Security.Principal.Windows" Version="4.4.0" />
+    <PackageReference Include="System.Security.Principal.Windows" Version="4.5.0" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/NLog/NLog.csproj
+++ b/src/NLog/NLog.csproj
@@ -304,6 +304,10 @@ For all config options and platform support, check https://nlog-project.org/conf
     <PackageReference Include="System.Reflection.TypeExtensions" Version="4.1.0" />
     <PackageReference Include="System.Xml.XmlDocument" Version="4.0.1" />
   </ItemGroup>
+
+	<ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
+		<PackageReference Include="System.Data.SqlClient" Version="4.6.0" />
+	</ItemGroup>
   
   <ItemGroup Condition=" '$(monobuild)' != '' ">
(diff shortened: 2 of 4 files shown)


snakefoot:
Sorry to disappoint you. But any solution that involves adding dependencies to the NLog-nuget-package is a complete no-go.

anxkha:
So what I don't understand about how it's currently built is for the netstandard 2.0 target (and the other netstandard targets), it's loading an assembly of "System.Data.SqlClient" in SetConnectionType. Without a package reference to System.Data.SqlClient either in NLog or the referencing application, this will blow up for netstandard x.x, as SqlClient is not included as part of the netstandard API. Is this intentional that you want a runtime-level exception to happen if the calling application does not have a reference to this package/DLL?

I can attempt to set the property via reflection if you absolutely cannot have a reference to that package. However, this feels to me just as dangerous as what I described above, for the same reason that it will cause a runtime exception instead of being caught at compile time.

snakefoot:
> I can attempt to set the property via reflection if you absolutely cannot have a reference to that package

snakefoot:
@anxkha Looked at the DbType-assignment code and it is rather complex with probe-logic and conversion logic from DbType-enum to System-Types.

Take a look at the class `PropertyHelper` with the method `SetPropertyFromString`. It is a helper method for assigning property-values using reflection. It is not the fastest method, but it can get you started.

304NotModified:
@anxkha

To elaborate on the references: NLog is currently dependency free. So if you don't need the database target, you don't need that package. Also the database targets works for a lot of database types with (runtime) dependencies, like the Oracle and MySQL, see https://github.com/NLog/NLog/wiki/Database-target#dbprovider-examples

Another option would be to make a new target (and optionally inherited from the database target) and a new nuget package and also a new git repo is preferred. For example see https://github.com/NLog/NLog.MailKit

304NotModified:
I've made an alternative approach, based on this PR! https://github.com/NLog/NLog/pull/3825 

Thanks @anxkha for the input!

anxkha:
@304NotModified You're welcome! Thank you for saving me from my still-limited knowledge of NLog internals ;)

I actually first started out by trying to make my own AzureDatabaseTarget that basically did the same thing as DatabaseTarget, except there are a couple of functions that are internal to NLog that seemed rather import to the operation of DatabaseTarget, so without those functions being made public or attempting a drastic, completely different approach, I wasn't able to go that route :(

304NotModified:
@anxkha no problem :) I think it's also a nice feature to have.

Would it be possible to test the pre-release package for us?

anxkha:
@304NotModified Absolutely (as you saw on the other PR). I assume it's safe to close this PR now that yours exists?

304NotModified:
yes I think also :)

Issue #3554, opened 2019-08-20
Configure and write to Azure SQL server with manage Identity token


Cook-code:
I want to connect with Managed Identity and write the logs in Azure SQL. How can I configure this. As manage identity login has an Access token.

**NLog version**: 4.6.6

**Platform**: . .NET Core 2

**Current NLog config** (4.6.6)

 - Why do we need it?
 - Need to log error with Managed Identity in Azure SQL

snakefoot:
NLog DatabaseTarget prefer that you provide the user-authentication using the ConnectionString.

Guess you have to ask Microsoft how to create a ConnectionString that supports "managed identity token".

See also: https://docs.microsoft.com/en-us/azure/storage/common/storage-configure-connection-string#configure-a-connection-string-for-an-azure-storage-account

snakefoot:
@Cook-code You can also try out my suggestion here for injecting Access Token in SqlConnection: https://stackoverflow.com/a/57597332/193178

304NotModified:
@snakefoot

Can't we add support for https://docs.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlconnection.accesstoken ? 

Doesn't seem difficult to delegate the accesstoken property, or am I missing something?

snakefoot:
> Doesn't seem difficult to delegate the accesstoken property, or am I missing something?

You are welcome to add an `IList<ConnectionProperty>` to NLog DatabaseTarget, that uses the NLog-reflection logic to assign different property values after having performed connectio-open.

tedmatuszewski:
I am experiencing this very issue. Has anybody found a solution or work-around for this?

Cook-code:
work around for me -  I had a blob storage so strong the logs in in BLOB now.

tedmatuszewski:
> work around for me - I had a blob storage so strong the logs in in BLOB now.

I could set it up to use BLOB storage, however I would rather not as consuming the logs from BLOB storage takes a little bit more doing that just reading from a database table.

anxkha:
Has anyone started on this or considered it for implementation? I have a potentially solid theory I can implement and add to the DatabaseTarget if not.

snakefoot:
@anxkha You are very welcome to make a pull-request. Only have 2 requirements:
- AccessToken value should come from a NLog Layout (Supporting `${configsetting}` etc.)
- AccessToken should only be assigned when opening the SqlConnection.

[...]

PR #3825, opened 2020-03-03
Azure AD Access Token support for DatabaseTarget 

304NotModified:
Based on #3824

fixes https://github.com/NLog/NLog/issues/3554

### DIFF
--- a/src/NLog/Internal/PropertySetter.cs
+++ b/src/NLog/Internal/PropertySetter.cs
@@ -110,6 +110,14 @@ namespace NLog.Internal
             _propertySetter = ReflectionHelpers.CreateLateBoundMethod(property.GetSetMethod());
         }
 
+        /// <summary>
+        /// Try get the property
+        /// </summary>
+        /// <param name="obj"></param>
+        /// <param name="propertyName"></param>
+        /// <param name="bindingFlags"></param>
+        /// <param name="property"></param>
+        /// <returns></returns>
         protected static bool TryGetProperty(T obj, string propertyName, BindingFlags bindingFlags, out PropertyInfo property)
         {
             if (obj == null)
--- a/src/NLog/Internal/PropertySetter.cs
+++ b/src/NLog/Internal/PropertySetter.cs
@@ -38,6 +38,9 @@ using NLog.Common;
 
 namespace NLog.Internal
 {
+    /// <summary>
+    /// Compiled version for setting a property value.
+    /// </summary>
     internal class PropertySetter : PropertySetter<object>
     {
         /// <inheritdoc />
@@ -84,9 +87,17 @@ namespace NLog.Internal
         }
     }
 
+    /// <summary>
+    /// Compiled version for setting a property value.
+    /// </summary>
+    /// <typeparam name="T">The type of the property. Not used for detection, but only for compile-time safe <see cref="SetValue"/></typeparam>
     internal class PropertySetter<T>
     {
         private readonly ReflectionHelpers.LateBoundMethod _propertySetter;
+
+        /// <summary>
+        /// array with 1 item. The property will always have one property
+        /// </summary>
         private readonly object[] _arrayWith1Item = new object[1];
 
         public PropertySetter([NotNull] PropertyInfo property)
--- a/src/NLog/Internal/PropertySetter.cs
+++ b/src/NLog/Internal/PropertySetter.cs
@@ -40,11 +40,11 @@ namespace NLog.Internal
 {
     internal class PropertySetter<T>
     {
-        private readonly PropertyInfo _property;
+        private ReflectionHelpers.LateBoundMethod _propertySetter;
 
         private PropertySetter([NotNull] PropertyInfo property)
         {
-            _property = property ?? throw new ArgumentNullException(nameof(property));
+            _propertySetter = ReflectionHelpers.CreateLateBoundMethod(property.GetSetMethod());
         }
 
         public static PropertySetter<T> TryCreate([NotNull] T obj, [NotNull] string propertyName, 
@@ -83,7 +83,7 @@ namespace NLog.Internal
                 throw new ArgumentNullException(nameof(obj));
             }
 
-            _property.SetValue(obj, value, null);
+            _propertySetter.Invoke(obj, new []{value});
         }
     }
 }
--- a/src/NLog/Internal/PropertySetter.cs
+++ b/src/NLog/Internal/PropertySetter.cs
@@ -101,7 +101,6 @@ namespace NLog.Internal
             }
 
             _arrayWith1Item[0] = value;
-
             _propertySetter.Invoke(obj, _arrayWith1Item);
         }
     }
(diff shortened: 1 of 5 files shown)


snakefoot:
Think Connection-Properties should be dynamic just like Command-Parameters.

Instead of DBAccessToken then it should be a list of connetion-property-items. Each with a property-name and property-value-layout.

304NotModified:
That's indeed also an option, but a bit more work

snakefoot:
> That's indeed also an option, but a bit more work

Well elegant generic solutions requires work.

Maybe your first suggestion of creating an MsSqlConnection-specific- uget-package is better. Maybe based on SqlBulkCopy.

anxkha:
@304NotModified Is there an easy way to get at a preview build of this on nuget so I can test drive it and put it in my project? Or only for the next RC of 4.7.0 or whatever comes next?

304NotModified:
> @304NotModified Is there an easy way to get at a preview build of this on nuget so I can test drive it and put it in my project? Or only for the next RC of 4.7.0 or whatever comes next?

Yes, it will be in the appveyor output when the build is done.

snakefoot:
Since you are walking down the path of creating yet another dynamic-setter-logic, then I think it is time to extract the logic from `class DbTypeSetter` into `class PropertyHelper` and reuse this logic for assigning DbConnection-properties.

304NotModified:
good idea for NLog 5. I will try not to refactor too much for NLog 4.x

snakefoot:
Adding new features are not refactoring.

304NotModified:
> Since you are walking down the path of creating yet another dynamic-setter-logic, then I think it is time to extract the logic from class DbTypeSetter into class PropertyHelper

This is refactoring?

snakefoot:
It is only this logic you should extract:

`PropertyInfo propInfo = dbParameterType.GetProperty(dbTypeNames[0], BindingFlags.IgnoreCase | BindingFlags.Public | BindingFlags.Instance);`

And this logic:

```c#
var dbTypeSetterLambda = ReflectionHelpers.CreateLateBoundMethod(_dbTypeSetter.GetSetMethod());
var dbTypeSetterParams = new object[] { _dbTypeValue };
_dbTypeSetterFast = (p) => dbTypeSetterLambda.Invoke(p, dbTypeSetterParams);
```

304NotModified:
I could create a more generic approach of AccessTokenSetter, e.g. PropertySetter. That will make this easier to set other properties on the Connection. What do you think?

snakefoot:
> I could create a more generic approach of AccessTokenSetter, e.g. PropertySetter

Sound good. If you can make a single ConnectionPropertySetter then it is super easy to iterate a list of them (Item with property-name + value-layout)

304NotModified:
there is a difference between 

DbTypeSetter: 
    _dbTypeSetterFast = (p) => dbTypeSetterLambda.Invoke(p, dbTypeSetterParams);

and the "setterFast" for this will have 2 parameters. I don't see how to combine those different approaches. Any suggestions?

snakefoot:
>  Any suggestions?

Of course the generic version should allow you to assign a dynamic-value using parameter-array. It is just `class DbTypeSetter` that knows the dynamic-value never changes and wraps the dynamic-value-setter in a lambda that captures the "static" array for reuse.

One could make an property-assigner-class that takes a value generator-func, and reuses the same parameter-array for each call (updating the first parameter before calling). Ofcourse it is restricted so it takes lock on the internal parameter-array while calling the setter-method.

snakefoot:
Actually the logic about reusing an internal parameter-array should just be placed in the value-generator-func

304NotModified:
I've create the PropertySetter, but haven't refactored the DbTypeSetter yet. Let me know what you think!

snakefoot:
You just need to 2 things:

- Add a `SetValue`-method that takes a pre-allocated parameter-array (object[]),
- Add new method for calling `ReflectionHelpers.CreateLateBoundMethod(_propertySetter);

Then it can be used by DbTypeSetter.

snakefoot:
If you are going to use it for AccessToken, then I guess you need to make an interface for PropertySetter called IPropertySetter. And create the propertysetter like this:

```
var propertySetter = (IPropertySetter)Activator.CreateInstance(typeof(PropertySetter<>).MakeGenericType(propInfo.PropertyType));
```

Alternative change `PropertySetter` to be non-generic and only get object-type and property-name.

304NotModified:
Or for non generic approach, pass object as T? I could even makes T optional (defaults to object)

304NotModified:
> Add a SetValue-method that takes a pre-allocated parameter-array

And then reuse the array every time?

snakefoot:
> Or for non generic approach, pass object as T? I could even makes T optional (defaults to object)

Right now there only two cases of types. Custom DbType and String. For both cases then the object-value will do just fine.

But I guess some day in the feature somebody wants to assign an integer (Either for the DbConnection, or maybe one day for the DbCommand). And then the ObjectSetter needs to inspet the PropertyType and attempt conversion from input-value-type (possible string or rawvalue) to the actual type.

snakefoot:
> > Add a SetValue-method that takes a pre-allocated parameter-array

> And then reuse the array every time?

Nope it is the caller that has to decide whether to reuse the same array every time. For DbType it is done by doing delegate-capture of the parameter-array.

src/NLog/Internal/PropertySetter.cs
```diff
@@ -60,8 +75,50 @@ private PropertySetter([NotNull] PropertyInfo property)
...
+    /// </summary>
+    /// <typeparam name="T">The type of the property. Not used for detection, but only for compile-time safe <see cref="SetValue"/></typeparam>
+    internal class PropertySetter<T>
+    {
+        private readonly ReflectionHelpers.LateBoundMethod _propertySetter;
+
+        /// <summary>
+        /// array with 1 item. The property will always have one property
+        /// </summary>
+        private readonly object[] _arrayWith1Item = new object[1];
```
snakefoot: Having the object-reuse here will give race-condition issues for those not aware. I would move this object-reuse out of the PropertySetter-class (Have SetValue-method that takes array)

snakefoot: The array-reuse is mostly important for custom DbType. For Connection-properties then a single-item-array allocation is nothing compare to the allocation performed by DbCommand.

304NotModified:
I like to do what you're asking, but it too hard and I don't want to burn my fingers on the DbTypeSetter . I think this good for now and I'm OK with refactoring this further in the future.

src/NLog/Internal/PropertySetter.cs
304NotModified: I think we should fix the array thing outside this PR. I think the latebound creator could be improved. Will create a new PR for that.

snakefoot:
Created #3829 to skip the journey with future refactoring and trying to mark DBAccessToken as obsolete (With lots of battles because the property has just been added, and it doesn't look good to mark it obsolete so quickly)

304NotModified:
Ok cool!

 I'm on a short holiday from tomorrow. Im back March 11 and will continue this and release 4.7 shortly. (I tried finishing 4.7 before my holiday ;) but quality of quantity :+1:)

304NotModified:
superseded by #3829

PR #3828, opened 2020-03-04
Skip need for Activator.CreateInstance in DbTypeSetter + fix Enum.TryParse for whitespace

304NotModified:
And removal of some code complexity in DbTypeSetter  with interfaces, explicit interface impl and MakeGeneric.

### DIFF
--- a/src/NLog/Common/ConversionHelpers.cs
+++ b/src/NLog/Common/ConversionHelpers.cs
@@ -44,15 +44,15 @@ namespace NLog.Common
         /// <summary>
         /// Converts input string value into <see cref="System.Enum"/>. Parsing is case-insensitive.
         /// </summary>
-        /// <param name="value">Input value</param>
-        /// <param name="result">Output value</param>
+        /// <param name="inputValue">Input value</param>
+        /// <param name="resultValue">Output value</param>
         /// <param name="defaultValue">Default value</param>
         /// <returns>Returns <paramref name="defaultValue"/> if the input value could not be parsed</returns>
-        public static bool TryParseEnum<TEnum>(string value, out TEnum result, TEnum defaultValue = default(TEnum)) where TEnum : struct
+        public static bool TryParseEnum<TEnum>(string inputValue, out TEnum resultValue, TEnum defaultValue = default(TEnum)) where TEnum : struct
         {
-            if (!TryParseEnum(value, true, out result))
+            if (!TryParseEnum(inputValue, true, out resultValue))
             {
-                result = defaultValue;
+                resultValue = defaultValue;
                 return false;
             }
             return true;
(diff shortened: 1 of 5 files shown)


snakefoot:
I like it.

304NotModified:
Also fix  Enum.TryParse for whitespace. It should not return success, that's nonsense.

src/NLog/Common/ConversionHelpers.cs
```diff
@@ -44,24 +44,39 @@ public static class ConversionHelpers
         /// <summary>
         /// Converts input string value into <see cref="System.Enum"/>. Parsing is case-insensitive.
         /// </summary>
-        /// <param name="inputValue">Input value</param>
-        /// <param name="resultValue">Output value</param>
+        /// <param name="value">Input value</param>
```
snakefoot: Not that I don't like the names, but changing parameter on public-methods are breaking change.

src/NLog/Common/ConversionHelpers.cs
```diff
@@ -44,24 +44,39 @@ public static class ConversionHelpers
...
+        /// <param name="result">Output value</param>
         /// <param name="defaultValue">Default value</param>
-        /// <returns>Returns failure if the input value could not be parsed</returns>
-        public static bool TryParseEnum<TEnum>(string inputValue, out TEnum resultValue, TEnum defaultValue = default(TEnum)) where TEnum : struct
+        /// <returns>Returns <paramref name="defaultValue"/> if the input value could not be parsed</returns>
+        public static bool TryParseEnum<TEnum>(string value, out TEnum result, TEnum defaultValue = default(TEnum)) where TEnum : struct
         {
-            if (string.IsNullOrEmpty(inputValue))
+            if (!TryParseEnum(value, true, out result))
```
snakefoot: Are you not changing behavior of public-method when removing the string.IsNullOrEmpty-check ?

src/NLog/Common/ConversionHelpers.cs
304NotModified: you're right, thanks

src/NLog/Common/ConversionHelpers.cs
304NotModified: yes, it's a bug IMO

snakefoot: Can see you have adjusted in both DatabaseParameterInfo.cs  and MailTarget.cs to cater for the new behavior. What about Mailkit ?

src/NLog/Common/ConversionHelpers.cs
```diff
@@ -47,38 +47,53 @@ public static class ConversionHelpers
...
-            if (!TryParse(inputValue, true, out resultValue))
+        /// <summary>
+        /// Converts input string value into <see cref="System.Enum"/>. Parsing is case-insensitive.
+        /// </summary>
+        /// <param name="inputValue">Input value</param>
+        /// <param name="enumType">The type of the enum</param>
+        /// <param name="resultValue">Output value. Null if parse failed</param>
+        internal static bool TryParseEnum(string inputValue, Type enumType, out object resultValue)
+        {
+            // Note: .NET Standard 2.1 added a public Enum.TryParse(Type)
```
snakefoot: Maybe add `if (StringHelpers.IsNullOrWhiteSpace(inputValue)) return false` since it is a common scenario.

src/NLog/Common/ConversionHelpers.cs
snakefoot: But must admit I also prefer that it returns false when empty-string. But one should just have special logging when parsing fails, and fallback is performed.

src/NLog/Targets/MailTarget.cs
```diff
@@ -650,7 +650,7 @@ private MailMessage CreateMailMessage(LogEventInfo lastEvent, string body)
             if (Priority != null)
             {
                 var renderedPriority = Priority.Render(lastEvent);
-                if (ConversionHelpers.TryParseEnum(renderedPriority, out MailPriority mailPriority))
+                if (!string.IsNullOrEmpty(renderedPriority) && ConversionHelpers.TryParseEnum(renderedPriority, out MailPriority mailPriority))
```
snakefoot: Should be `if (ConversionHelpers.TryParseEnum(renderedPriority, out MailPriority mailPriority)) { ... } else if (!string.IsNullOrEmpty(renderedPriority) { Log-warning }`

304NotModified: The parse is wrong if the value is empty?

snakefoot: Right now it logs a warning every time `string.IsNullOrEmpty(renderedPriority)==true`

304NotModified: Oops that's wrong of course. Thanks!

PR #3829, opened 2020-03-04
DatabaseTarget - Added support for Connection Properties (e.g. for Azure AD Access Token)

### LINKED ISSUE #3554: Configure and write to Azure SQL server with manage Identity token

snakefoot:
Based on  #3825

fixes #3554

```xml
<target name="db"
        xsi:type="Database"
        commandType="StoredProcedure"
        commandText="[dbo].[NLog_AddEntry_p]"
        >
  <connectionProperty name="AccessToken" layout="${gdc:AccessToken}" propertyType="System.String" />
  <parameter name="@logged"         layout="${date}" />
  <parameter name="@level"          layout="${level}" />
  <parameter name="@message"        layout="${message}" />
  <parameter name="@logger"         layout="${logger}" />
</target>
```

For SqlConnection AccessToken then one needs a special LayoutRenderer that can call `AzureServiceTokenProvider` as AccessToken can have a limited lifetime.

### DIFF
--- a/src/NLog/Internal/PropertySetter.cs
+++ b/src/NLog/Internal/PropertySetter.cs
@@ -37,18 +37,13 @@ namespace NLog.Internal
     using System.Reflection;
     using NLog.Common;
 
-    internal interface IPropertySetter
-    {
-        bool SetPropertyValue(object instance, object value);
-    }
-
-    internal class PropertySetter : IPropertySetter
+    internal class PropertySetter
     {
         private readonly Type _objectType;
         private readonly PropertyInfo _propertyInfo;
         private ReflectionHelpers.LateBoundMethod _propertySetter;
 
-        public static IPropertySetter GeneratePropertySetter(Type objectType, string propertyName)
+        public static PropertySetter CreatePropertySetter(Type objectType, string propertyName)
         {
             if (TryGetProperty(objectType, propertyName, out var propertyInfo))
             {
@@ -78,8 +73,9 @@ namespace NLog.Internal
             else
             {
                 // Generate compiled method if setter works without throwing exception
-                _propertyInfo.GetSetMethod().Invoke(instance, new[] { value });
-                _propertySetter = ReflectionHelpers.CreateLateBoundMethod(_propertyInfo.GetSetMethod());
+                var setterMethod = _propertyInfo.GetSetMethod();
+                setterMethod.Invoke(instance, new[] { value });
+                _propertySetter = ReflectionHelpers.CreateLateBoundMethod(setterMethod);
             }
 
             return true;
--- a/src/NLog/Internal/PropertySetter.cs
+++ b/src/NLog/Internal/PropertySetter.cs
@@ -1,0 +1,119 @@
+﻿// 
+// Copyright (c) 2004-2020 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without 
+// modification, are permitted provided that the following conditions 
+// are met:
+// 
+// * Redistributions of source code must retain the above copyright notice, 
+//   this list of conditions and the following disclaimer. 
+// 
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution. 
+// 
+// * Neither the name of Jaroslaw Kowalski nor the names of its 
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission. 
+// 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE 
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE 
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE 
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR 
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS 
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN 
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) 
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF 
+// THE POSSIBILITY OF SUCH DAMAGE.
+// 
+
+namespace NLog.Internal
+{
+    using System;
+    using System.Reflection;
+    using NLog.Common;
+
+    internal interface IPropertySetter
+    {
+        bool SetPropertyValue(object instance, object value);
+    }
+
+    internal class PropertySetter : IPropertySetter
+    {
+        private readonly Type _objectType;
+        private readonly PropertyInfo _propertyInfo;
+        private ReflectionHelpers.LateBoundMethod _propertySetter;
+
+        public static IPropertySetter GeneratePropertySetter(Type objectType, string propertyName)
+        {
+            if (TryGetProperty(objectType, propertyName, out var propertyInfo))
+            {
+                return new PropertySetter(objectType, propertyInfo);
+            }
+
+            return null;
+        }
+
+        private PropertySetter(Type objectType, PropertyInfo propertyInfo)
+        {
+            _objectType = objectType;
+            _propertyInfo = propertyInfo;
+        }
+
+        public bool SetPropertyValue(object instance, object value)
+        {
+            if (instance == null)
+            {
+                throw new ArgumentNullException(nameof(instance));
+            }
+
+            if (_propertySetter != null)
+            {
+                _propertySetter.Invoke(instance, new[] { value });
+            }
+            else
+            {
+                // Generate compiled method if setter works without throwing exception
+                _propertyInfo.GetSetMethod().Invoke(instance, new[] { value });
+                _propertySetter = ReflectionHelpers.CreateLateBoundMethod(_propertyInfo.GetSetMethod());
+            }
+
+            return true;
+        }
+
+        /// <summary>
+        /// Try get the property
+        /// </summary>
+        private static bool TryGetProperty(Type objectType, string propertyName, out PropertyInfo property)
+        {
+            if (objectType == null)
+            {
+                throw new ArgumentNullException(nameof(objectType));
+            }
+
+            if (propertyName == null)
+            {
+                throw new ArgumentNullException(nameof(propertyName));
+            }
+
+            property = objectType.GetProperty(propertyName, BindingFlags.Public | BindingFlags.Instance | BindingFlags.IgnoreCase);
+            if (property == null)
+            {
+                InternalLogger.Warn("Cannot set {0} on type {1}, property not found", propertyName, objectType.FullName);
+                return false;
+            }
+
+            if (!property.CanWrite)
+            {
+                InternalLogger.Warn("Cannot set {0} on type {1}, property not settable", propertyName, objectType.FullName);
+                return false;
+            }
+
+            return true;
+        }
+    }
+}
--- a/src/NLog/Targets/DatabaseConnectionInfo.cs
+++ b/src/NLog/Targets/DatabaseConnectionInfo.cs
@@ -36,7 +36,6 @@
 namespace NLog.Targets
 {
     using System;
-    using System.Collections.Generic;
     using System.ComponentModel;
     using System.Data;
     using System.Globalization;
@@ -45,7 +44,7 @@ namespace NLog.Targets
     using NLog.Layouts;
 
     /// <summary>
-    /// Information about database command + parameters.
+    /// Information about database connection property + value
     /// </summary>
     [NLogConfigurationItem]
     public class DatabaseConnectionInfo
@@ -58,37 +57,37 @@ namespace NLog.Targets
         }
 
         /// <summary>
-        /// Gets or sets the DbConnection Property to update
+        /// Gets or sets the property name for the database connection object (Not case-sensitive)
         /// </summary>
         /// <docgen category='Connection Options' order='10' />
         [RequiredParameter]
         public string Name { get; set; }
 
         /// <summary>
-        /// Gets or sets the value to assign on DbConnection
+        /// Gets or sets the value to assign on database connection object
         /// </summary>
         /// <docgen category='Connection Options' order='10' />
         [RequiredParameter]
         public Layout Layout { get; set; }
 
         /// <summary>
-        /// Gets or sets the type of the property
+        /// Gets or sets the type of the database connection property
         /// </summary>
         /// <docgen category='Connection Options' order='10' />
         [DefaultValue(typeof(string))]
         public Type PropertyType { get; set; } = typeof(string);
 
         /// <summary>
-        /// Gets or sets convert format of the property value .
+        /// Gets or sets convert format of the property value
         /// </summary>
-        /// <docgen category='Parameter Options' order='8' />
+        /// <docgen category='Connection Options' order='8' />
         [DefaultValue(null)]
         public string Format { get; set; }
 
         /// <summary>
         /// Gets or sets the culture used for parsing property string-value for type-conversion
         /// </summary>
-        /// <docgen category='Parameter Options' order='9' />
+        /// <docgen category='Connection Options' order='9' />
         [DefaultValue(null)]
         public CultureInfo Culture { get; set; }
 
@@ -96,17 +95,36 @@ namespace NLog.Targets
         {
             var dbConnectionType = dbConnection.GetType();
             var propertySetterCache = _propertySetter;
-            if (propertySetterCache.Key.Key != Name || propertySetterCache.Key.Value != dbConnectionType)
+            if (!propertySetterCache.Equals(Name, dbConnectionType))
             {
-                var propertySetter = PropertySetter.GeneratePropertySetter(dbConnectionType, Name);
-                propertySetterCache = new KeyValuePair<KeyValuePair<string, Type>, IPropertySetter>(new KeyValuePair<string, Type>(Name, dbConnectionType), propertySetter);
+                var propertySetter = PropertySetter.CreatePropertySetter(dbConnectionType, Name);
+                propertySetterCache = new PropertySetterCacheItem(Name, dbConnectionType, propertySetter);
                 _propertySetter = propertySetterCache;
             }
 
-            return propertySetterCache.Value?.SetPropertyValue(dbConnection, propertyValue) ?? false;
+            return propertySetterCache.PropertySetter?.SetPropertyValue(dbConnection, propertyValue) ?? false;
         }
 
-        KeyValuePair<KeyValuePair<string, Type>, IPropertySetter> _propertySetter;
+        private struct PropertySetterCacheItem
+        {
+            public string PropertyName { get; }
+            public Type ObjectType { get; }
+            public PropertySetter PropertySetter { get; }
+
+            public PropertySetterCacheItem(string propertyName, Type objectType, PropertySetter propertySetter)
+            {
+                PropertyName = propertyName;
+                ObjectType = objectType;
+                PropertySetter = propertySetter;
+            }
+
+            public bool Equals(string propertyName, Type objectType)
+            {
+                return PropertyName == propertyName && ObjectType == objectType;
+            }
+        }
+
+        PropertySetterCacheItem _propertySetter;
     }
 }
 
--- a/src/NLog/Targets/DatabaseConnectionInfo.cs
+++ b/src/NLog/Targets/DatabaseConnectionInfo.cs
@@ -1,4 +1,4 @@
-// 
+﻿// 
 // Copyright (c) 2004-2020 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
 // 
 // All rights reserved.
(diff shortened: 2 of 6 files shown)


304NotModified:
Could you please add a config example for this great feature for all future readers?

src/NLog/Internal/PropertySetter.cs
```diff
@@ -0,0 +1,119 @@
...
+// THE POSSIBILITY OF SUCH DAMAGE.
+// 
+
+namespace NLog.Internal
+{
+    using System;
+    using System.Reflection;
+    using NLog.Common;
+
+    internal interface IPropertySetter
```
304NotModified: Please move to a separate file

src/NLog/Internal/PropertySetter.cs
```diff
@@ -0,0 +1,119 @@
...
+        bool SetPropertyValue(object instance, object value);
+    }
+
+    internal class PropertySetter : IPropertySetter
+    {
+        private readonly Type _objectType;
+        private readonly PropertyInfo _propertyInfo;
+        private ReflectionHelpers.LateBoundMethod _propertySetter;
+
+        public static IPropertySetter GeneratePropertySetter(Type objectType, string propertyName)
```
304NotModified: Please CreatePropertySetter (it wont generate things IMO)

src/NLog/Internal/PropertySetter.cs
```diff
@@ -0,0 +1,119 @@
...
+
+            if (_propertySetter != null)
+            {
+                _propertySetter.Invoke(instance, new[] { value });
+            }
+            else
+            {
+                // Generate compiled method if setter works without throwing exception
+                _propertyInfo.GetSetMethod().Invoke(instance, new[] { value });
+                _propertySetter = ReflectionHelpers.CreateLateBoundMethod(_propertyInfo.GetSetMethod());
```
304NotModified: Maybe doing once `_propertyInfo.GetSetMethod()`? (sounds like an expensive method).

Maybe describe this class as "lazy compiled" (In class XML docs)

snakefoot:
@anxkha Looks like it is your lucky day. Getting the feature made for you almost on a silver plate. Links to nuget-package here:

https://ci.appveyor.com/project/nlog/nlog/builds/31250180/artifacts

Have added a new collection on DatabaseTarget called `ConnectionProperties` there you can add custom properties to assign on the SqlConnection.

```c#
var accessTokenLayout = NLog.Layouts.Layout.FromMethod(logEvent => "MyToken");
databaseTarget.ConnectionProperties.Add(new DatabaseConnectionInfo() { Name = "AccessToken", Layout = accessTokenLayout });
```

Hope it lives up to your expectations.

src/NLog/Targets/DatabaseConnectionInfo.cs
```diff
@@ -0,0 +1,113 @@
...
+        /// <docgen category='Connection Options' order='10' />
+        [RequiredParameter]
+        public string Name { get; set; }
+
+        /// <summary>
+        /// Gets or sets the value to assign on DbConnection
+        /// </summary>
+        /// <docgen category='Connection Options' order='10' />
+        [RequiredParameter]
+        public Layout Layout { get; set; }
```
304NotModified: Could we call this Value?

src/NLog/Targets/DatabaseConnectionInfo.cs
```diff
@@ -0,0 +1,113 @@
...
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DatabaseConnectionInfo"/> class.
+        /// </summary>
+        public DatabaseConnectionInfo()
+        {
+        }
+
+        /// <summary>
+        /// Gets or sets the DbConnection Property to update
```
304NotModified: please mention it's not case sensitive

src/NLog/Targets/DatabaseConnectionInfo.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        /// <summary>
+        /// Gets or sets the type of the property
+        /// </summary>
+        /// <docgen category='Connection Options' order='10' />
+        [DefaultValue(typeof(string))]
+        public Type PropertyType { get; set; } = typeof(string);
+
+        /// <summary>
+        /// Gets or sets convert format of the property value .
```
304NotModified: floating dot ;)

304NotModified: Please describe when we use this format in the XML docs.

src/NLog/Targets/DatabaseConnectionInfo.cs
```diff
@@ -0,0 +1,113 @@
...
+    using System.Globalization;
+    using NLog.Config;
+    using NLog.Internal;
+    using NLog.Layouts;
+
+    /// <summary>
+    /// Information about database command + parameters.
+    /// </summary>
+    [NLogConfigurationItem]
+    public class DatabaseConnectionInfo
```
304NotModified: Maybe name it DatabaseConnectionProperty?

src/NLog/Targets/DatabaseConnectionInfo.cs
```diff
@@ -0,0 +1,113 @@
...
+    using System.Collections.Generic;
+    using System.ComponentModel;
+    using System.Data;
+    using System.Globalization;
+    using NLog.Config;
+    using NLog.Internal;
+    using NLog.Layouts;
+
+    /// <summary>
+    /// Information about database command + parameters.
```
304NotModified: Wrong summary?

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -657,7 +692,7 @@ private void WriteLogEventBatchToDatabase(IList<AsyncLogEventInfo> logEvents, st
                 //Always suppress transaction so that the caller does not rollback logging if they are rolling back their transaction.
                 using (TransactionScope transactionScope = new TransactionScope(TransactionScopeOption.Suppress))
                 {
-                    EnsureConnectionOpen(connectionString);
+                    EnsureConnectionOpen(connectionString, logEvents.Count > 0 ? logEvents[0].LogEvent : null);
```
304NotModified: There was a reason for not using Linq, isn't?

src/NLog/Targets/DatabaseConnectionInfo.cs
```diff
@@ -0,0 +1,113 @@
...
+        public CultureInfo Culture { get; set; }
+
+        internal bool SetPropertyValue(IDbConnection dbConnection, object propertyValue)
+        {
+            var dbConnectionType = dbConnection.GetType();
+            var propertySetterCache = _propertySetter;
+            if (propertySetterCache.Key.Key != Name || propertySetterCache.Key.Value != dbConnectionType)
+            {
+                var propertySetter = PropertySetter.GeneratePropertySetter(dbConnectionType, Name);
+                propertySetterCache = new KeyValuePair<KeyValuePair<string, Type>, IPropertySetter>(new KeyValuePair<string, Type>(Name, dbConnectionType), propertySetter);
```
304NotModified: I really don't like to KeyValuePair (unreadable)... as we can't use c# 7 tuples, I propose:

- Creating a new type (class or struct) for ` KeyValuePair<KeyValuePair<string, Type>, IPropertySetter>`
- could be private nested
- Implement Equals for the check below.

e.g.

```c#
private class PropertySetterCacheItem
{
    [NotNull] public string Name { get; }
    [NotNull] public Type Type { get; }
    [NotNull] public IPropertySetter PropertySetter { get; }

    public PropertySetterCacheItem([NotNull] string name, [NotNull] Type type, [NotNull] IPropertySetter propertySetter)
    {
        Name = name ?? throw new ArgumentNullException(nameof(name));
        Type = type ?? throw new ArgumentNullException(nameof(type));
        PropertySetter = propertySetter ?? throw new ArgumentNullException(nameof(propertySetter));
    }


    /// <summary>
    /// Equals without IPropertySetter
    /// </summary>
    protected bool Equals(string name, Type type)
    {
        return string.Equals(Name, name, StringComparison.OrdinalIgnoreCase) && Type == type;
    }


}
```

src/NLog/Targets/DatabaseTarget.cs
```diff
@@ -1043,77 +1078,68 @@ protected virtual IDbDataParameter CreateDatabaseParameter(IDbCommand command, D
...
-            {
-                return RenderLogEvent(parameterInfo.Layout, logEvent) ?? string.Empty;
-            }
+            return RenderObjectValue(logEvent, parameterInfo.Name, parameterInfo.Layout, parameterInfo.ParameterType, parameterInfo.Format, parameterInfo.Culture);
+        }
 
-            IFormatProvider dbParameterCulture = GetDbParameterCulture(logEvent, parameterInfo);
+        private object GetDatabaseConnectionValue(LogEventInfo logEvent, DatabaseConnectionInfo connectionInfo)
+        {
+            return RenderObjectValue(logEvent, connectionInfo.Name, connectionInfo.Layout, connectionInfo.PropertyType, connectionInfo.Format, connectionInfo.Culture);
```
304NotModified: proposal: create an interface for DatabaseParameterInfo and DatabaseConnectionProperty, and pass that interface to RenderObjectValue.

snakefoot:
Guess it could evolve into also having the ability to assign "CommandProperty" :)

src/NLog/Targets/DatabaseTarget.cs
304NotModified: We need to add some internal readonly properties for the common interface on DatabaseParameterInfo and DatabaseConnectionProperty then, e.g.ValueType and ValueLayout

304NotModified: e.g. 


```c#
internal interface IDatabaseRenderable
{
    string Name { get; }
    Layout Layout { get; }
    Type ValueType { get; }
    string Format { get; }
    CultureInfo Culture { get; }
}

```

src/NLog/Targets/DatabaseConnectionInfo.cs
snakefoot: Nope Layout is the official standard in NLog.

src/NLog/Targets/DatabaseConnectionInfo.cs
snakefoot: Exotic scenario that is hard to explain.

src/NLog/Targets/DatabaseConnectionInfo.cs
snakefoot: I want it to match DatabaseParameterInfo

src/NLog/Targets/DatabaseConnectionInfo.cs
snakefoot: Lots of copy-paste

src/NLog/Targets/DatabaseTarget.cs
snakefoot: So many traps in Linq, so I try to avoid it, unless writing unit-testing code. Then allocation is free.

src/NLog/Targets/DatabaseConnectionInfo.cs
snakefoot: Sure just wanted to make it quick, since you were so eager on making a half solution. Now suddenly it has changed :)

src/NLog/Targets/DatabaseTarget.cs
snakefoot: Sounds like an mental exercise for you.

src/NLog/Targets/DatabaseConnectionInfo.cs
304NotModified: - multiple DatabaseParameterInfo means multiple parameters
- multiple DatabaseConnectionInfo doesn't mean multiple connections ;)

So the same name is a bit confusing

src/NLog/Targets/DatabaseTarget.cs
304NotModified: Could we add our own .FirstOrDefault (now or in the future)? It;s some much cleaner

src/NLog/Internal/PropertySetter.cs
snakefoot: Not expensive, but I can move it into a local variable for you.

src/NLog/Internal/PropertySetter.cs
snakefoot: Ok

src/NLog/Internal/PropertySetter.cs
304NotModified: I like locals :)

src/NLog/Internal/PropertySetter.cs
snakefoot: Removed the interface

src/NLog/Targets/DatabaseConnectionInfo.cs
snakefoot: Ok

src/NLog/Targets/DatabaseConnectionInfo.cs
304NotModified: LOL I had the same with #3825 ...

src/NLog/Targets/DatabaseTarget.cs
304NotModified: Yeah if you OK with this, I could push it to this branch

304NotModified:
> For SqlConnection AccessToken then one needs a special LayoutRenderer that can call `AzureServiceTokenProvider` as AccessToken can have a limited lifetime.

So after all, an AccessToken property is maybe easier to use?

src/NLog/Targets/DatabaseConnectionInfo.cs
304NotModified: Copy paste my private class is also quick? ;)

snakefoot: But now changed into struct.

304NotModified: OK, I was under the impression that a ref type in a struct, makes a struct a bit useless.

snakefoot:
> So after all, an AccessToken property is maybe easier to use?

Not sure I understand the question

304NotModified:
Well I like the generic/reusable approach. But it it will be a lot harder to use for an user, then i'm not sure if shouldn't just add the DbAccessToken property (and do the magic in there)

src/NLog/Targets/DatabaseConnectionInfo.cs
snakefoot: Not sure I understand. DatabaseConnectionInfo resides in a list. So does DatabaseParameterInfo

src/NLog/Targets/DatabaseTarget.cs
snakefoot: Feel free

snakefoot:
> add the DbAccessToken property (and do the magic in there)

What kind of magic will you do to wield AzureServiceTokenProvider without any dependencies ? Actually the reason why I wanted  @anxkha to perform this whole exercise, since the actual user. Never used an AccessToken in my life (Would allow us to learn something about how it used in different scenarios. AzureFunctions. AzureAspNetCore, etc. along with different lifetime-types).

src/NLog/Targets/DatabaseConnectionInfo.cs
snakefoot: I like immutable structs. They are clean and doesn't allocate. And provide nearness in memory.

src/NLog/Targets/DatabaseTarget.cs
snakefoot: Sure push away. Have committed the code-review changes agreed upon.

snakefoot:
One extra push was needed. Refactoring introduced a bug.

src/NLog/Targets/DatabaseTarget.cs
304NotModified: OK unresolve as reminder

304NotModified:
Yes I was already typing a comment and the commit was magically disappeared :P

src/NLog/Targets/DatabaseConnectionInfo.cs
304NotModified: - **Two** DatabaseParameterInfos will create/work on **two** database parameters. 
- **Two** DatabaseConnectionInfos work on **one single** DBconnection.

The same name won't work well if the don't behave the same. 

Dunno how I could explain this differently.

src/NLog/Targets/DatabaseConnectionInfo.cs
304NotModified: More like semiofficial ;)

src/NLog/Targets/DatabaseConnectionInfo.cs
snakefoot: Let me rephrase your sentence:

- **Two** DatabaseParameterInfos will assign two parameters to a DbCommand
- **Two** DatabaseConnectionInfos will assign two properties to a DbConnection

And both names makes sense to me.

304NotModified: Maybe it's DatabaseConnectionPropertyInfo then

304NotModified: > Let me rephrase your sentence:
> 
> * **Two** DatabaseParameterInfos will assign two parameters to a DbCommand 
> * **Two** DatabaseConnectionInfos will assign two properties to a DbConnection
> 

Don't you see the inconsistency? ParameterInfos/DbCommand  vs ConnectionInfos/DbConnection?

Any way, consistency aside, I find the name  ConnectionInfo confusing

snakefoot: Ok lets rename it to DatabasePropertyInfo. Then it can also be reused for DbCommand-properties.

304NotModified: Also the same in the target code:

c#
```
  public IList<DatabaseParameterInfo> Parameters { get; } = new List<DatabaseParameterInfo>();
 public IList<DatabaseConnectionInfo> ConnectionProperties { get; } = new List<DatabaseConnectionInfo>();
```

- **Parameter**s  - Database**Parameter**Info
- **ConnectionProperties**  - DatabaseConnectionInfo

anxkha:
@snakefoot @304NotModified Well, thank you both for taking this on. I've been super busy the last few days and you both moved so fast on this I didn't even have a good opportunity to sit down and fully review everything done in the PRs and your discussions.

> var accessTokenLayout = NLog.Layouts.Layout.FromMethod(logEvent => "MyToken");
> databaseTarget.ConnectionProperties.Add(new DatabaseConnectionInfo() { Name = "AccessToken", Layout = accessTokenLayout });

I haven't fully looked at the details of this PR, but based on the documentation example and this example, this "should" work correctly. The lambda for Layout.FromMethod would essentially look like the following:

` logEvent => new AzureServiceTokenProvider().GetAccessTokenAsync($"https://database.windows.net
/").Result;`

Technically `GetAccessTokenAsync` should be "awaited" and called from another async-capable function, but it appears from my looking that NLog does not have async piped through the entire stack at this point in time. The above approach will still work, and every time a connection is opened it would call `GetAccessTokenAsync` by rendering the layout, and get the latest token at the time it's called.

I'm happy to help as I am able are you are right, I initially was planning on providing an acceptable solution for this. My only concern, from a timing perspective, is I do not know the internals of NLog as thoroughly as you two do and it will take me a bit to come up to speed to where I feel comfortable not breaking anything. It sounds like you guys want to release 4.7.0 soon-ish.

src/NLog/Targets/DatabaseConnectionInfo.cs
304NotModified: > Ok lets rename it to DatabasePropertyInfo.

👍 that's a good idea

snakefoot: Ooops it became DatabaseObjectPropertyInfo

snakefoot:
@anxkha Would probably implement some fancy caching logic in your lambda, so you only refresh the AccessToken every 10-30 minutes (Maybe with an async-background-timer). So when calling the lambda it just returns the latest AccessToken.

snakefoot:
Rename also required an extra push

anxkha:
> Would probably implement some fancy caching logic in your lambda, so you only refresh the AccessToken every 10-30 minutes (Maybe with an async-background-timer). So when calling the lambda it just returns the latest AccessToken. Because the lambda will be called frequently (for every logevent unless doing batching with AsyncWrapper).

The `GetAccessTokenAsync` call actually does this for you. It will re-use the same token until it approaches expiration, then reach out to renew it for you so that you don't have to worry about it. I thought the same thing too, initially, but finally found a little more documentation on it.

snakefoot:
> The GetAccessTokenAsync call actually does this for you. It will re-use the same token until it approaches expiration,

Again I know nothing about AccessToken, but it looks like it can perform a http-request to lookup the next token. I'm guessing Microsoft have added logging incase AccessToken lookup fails. It is strictly forbidden to perform logging while inside a NLog-target (Causes deadlocks or stackoverflows, both issues you don't want).

So I think it is very important that the custom `accessTokenLayout` is not the one doing the actual http-request but just returns the latest AccessToken-Value, and then have a background-time responsible for refreshing the AccessToken-Value every 10 min.

snakefoot:
@anxkha Please tell if this custom pre-release solves you requirement. It will not enter the NLog 4.7 release, unless it actually meets the goal. It can take awhile before the next release train leaves the station.

anxkha:
@snakefoot Yeah I'll do my best to test it today as things are calmer. Regarding setting the access token, I hear you. I don't know what goes on inside their library to do it other than one or two HTTP calls to Azure. I'll come up with some process to refresh it outside the cycle.

anxkha:
@snakefoot I was able to test this and it appears to be working flawlessly! This is with build 4.6.0-11293.

snakefoot:
@anxkha Wow that was quick, and good news. Would it be possible for you to share how you implemented the AccessToken-Lookup logic with background-refresh?

anxkha:
@snakefoot Yep! In my first draft pass I have a `DatabaseConnectionFactory` class that has the following:

	public sealed class DatabaseConnectionFactory
	{
		private static readonly Timer _accessTokenTimer;
		...
		public static string AccessToken { get; private set; }

		static DatabaseConnectionFactory()
		{
			if (!UseMsiConnection)
				return;

			_accessTokenTimer = new Timer(60 * 1000)
			{
				AutoReset = true,
				Enabled = true
			};
			_accessTokenTimer.Elapsed += async (e, sender)
				=> AccessToken = await new AzureServiceTokenProvider().GetAccessTokenAsync($"https://{DatabaseHostSuffix}/");
			_accessTokenTimer.Start();

			AccessToken = new AzureServiceTokenProvider().GetAccessTokenAsync($"https://{DatabaseHostSuffix}/").Result;
		}
		...

And then for the NLog configuration:

	LogManager.LoadConfiguration("nlog.config");

	var databaseTarget = (DatabaseTarget)LogManager.Configuration.FindTargetByName("database");
	databaseTarget.ConnectionString = DatabaseConnectionFactory.SomeDatabase;
	databaseTarget.CommandType = CommandType.StoredProcedure;

	if (DatabaseConnectionFactory.UseMsiConnection)
	{
		var accessTokenLayout = Layout.FromMethod(logEvent => DatabaseConnectionFactory.AccessToken);
		databaseTarget.ConnectionProperties.Add(new DatabaseObjectPropertyInfo
		{
			Name = nameof(SqlConnection.AccessToken),
			Layout = accessTokenLayout
		});
	}

I still need to do some clean code refactoring, and it needs some form of semaphore or locking to protect the AccessToken interaction, but it works and it supports both Managed Service Identity and regular SQL Auth access scenarios.

snakefoot:
@anxkha If you have the time, then I would become very happy if you could can verify this nuget-package (NLog.Extensions.AzureAccessToken.2.0.0.nupkg) is working:

https://ci.appveyor.com/project/JDetmar/nlog-extensions-azurestorage/builds/31792097/artifacts

It can be used like this:

```
<extensions>
  <add assembly="NLog.Extensions.AzureAccessToken" /> 
</extensions>
<targets>
    <target type="database">
         <connectionProperty name="AccessToken" layout="${AzureAccessToken:ResourceName=${gdc:DatabaseHostSuffix}}"  />
    </target>
</targets>
```

And before logging then you should do this:

```c#
NLog.GlobalDiagnosticsContext.Set("DatabaseHostSuffix", $"https://{DatabaseHostSuffix}/");
```

snakefoot:
@anxkha Nuget-package is now available:

https://www.nuget.org/packages/NLog.Extensions.AzureAccessToken/

anxkha:
@snakefoot Thanks, I'll take a look today!

snakefoot:
@anxkha Any success in your adventures with [NLog.Extensions.AzureAccessToken](https://www.nuget.org/packages/NLog.Extensions.AzureAccessToken/) ?

anxkha:
@snakefoot My apologies, I am juggling 10 different projects and forgot to reply to you. Yes, this package using the instructions you put on the original feature request appears to be working great!

snakefoot:
@anxkha Excellent. Thank you for the testing and the confirmation.

PR #3830, opened 2020-03-04
Addded LateBoundMethodSingle

304NotModified:
Idea of skipping allocating the array.

I've tested this before with the PropertySetter (now not included)

Unfortunately there is some code duplication. Would be nice if we could skip that.

### DIFF
--- a/src/NLog/Internal/PropertySetter.cs
+++ b/src/NLog/Internal/PropertySetter.cs
@@ -41,7 +41,7 @@ namespace NLog.Internal
     {
         private readonly Type _objectType;
         private readonly PropertyInfo _propertyInfo;
-        private ReflectionHelpers.LateBoundMethod _propertySetter;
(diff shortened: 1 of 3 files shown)


snakefoot:
Not very skilled with expression trees. Try to stay away from them when possible :)

304NotModified:
I works, I think this is nice. This implementation is maybe not the best code (because of code duplication), but it makes things really easier (and better) from the outside.

snakefoot:
If 10000000 property assignments doesn't generate any memory-allocations with this new method, then I think it is great.

304NotModified:
AFAIK it does. But dunno how to benchmark (well)

304NotModified:
@snakefoot could please you perf test this?

snakefoot:
Performance looks great. Looking forward for it to be merged. Testing only showed this  #3843

PR #3831, opened 2020-03-05
DI support

snakefoot:
Merges DI-Resolving into dev after having performed rebase.

replaces https://github.com/NLog/NLog/pull/3681


snakefoot:
One could consider changing IServiceRepository into an abstract-class, thus making it easier to add properties / method without it being breaking change.

304NotModified:
> One could consider changing IServiceRepository into an abstract-class, thus making it easier to add properties / method without it being breaking change.

Smart one!

snakefoot:
I guess one goal for this PR is that NLog can support Microsoft ApplicationInsight. So the local singleton can be exposed (made available) to the NLog ApplicationInsight Target.

`NLog.config` will be loaded first (before ApplicationInsight `TelemetryConfiguration` is available), so the NLog ApplicationInsight Target needs to handle lazy initialization by itself. Meaning that it should retry the resolve of `TelemetryConfiguration` until available in NLog ServiceRepository.

It is dangerous for the NLog ServiceRepository to resolve services while handling a LogEvent, because the LogEvent can come from the service that we are trying to resolve (deadlock). This is why it is very important that NLog NEVER creates instances of outside objects, but only create NLog config-items (Trusting Layout/Target/LayoutRenderer will not be logging while being created).

It could be neat if NLog somehow could know upfront what service-types a Target / Layout / LayoutRenderer would like to make use of (Ex. like a class-attribute). Then when calling NLog UseNLog then it would know upfront what service-types (from the loaded `LoggingConfiguration`) to lookup and register in the NLog-ServiceRepository.

Ofcourse it should also be be possible to manually register/override the singleton-instance for a service-type in the NLog ServiceRepository.

snakefoot:
> Smart one!

Change IServiceRepository into abstract class. And dropped IServiceResolver and instead just use System.IServiceProvider.

snakefoot:
I think this is a good first step for improving the support for dependency injection. Still don't think it is completely safe to allocate random objects from inside the NLog-Logging-Engine (as they can trigger logging that causes deadlock), but I guess you (@304NotModified) have a use-case where this is needed (and know the custom objects will not be doing logging).

Think next step after this PR has been merged is to look into implementing the mention class-attribute, with support for lazy initialization. Next step again would be to handle buffering of LogEvents until lazy initialization has completed.

304NotModified:
> but I guess you (@304NotModified) have a use-case where this is needed (and know the custom objects will not be doing logging).

I'm not sure which use case I should describe.



> I guess one goal for this PR is that NLog can support Microsoft ApplicationInsight. So the `TelemetryConfiguration` singleton can be exposed (made available) to the NLog ApplicationInsight Target.

Yes that's a goal, and also these:

- Must: Targets could inject dependencies (not from NLog)
- Must: we remove static calls so we really could replace stuff inside NLog, like JSON.NET for JSON (currently not possible)
- Should: We should replace the DI with another DI container

 


> It could be nice if NLog could delay the initialization of a NLog Target until ready (inspecting the before mentioned class-attributes, and checking the NLog ServiceRepository). Instead of the NLog Target manually having to retry the resolve of needed services from the NLog ServiceRepository.

Yes indeed as this is also an issue with the setup when using MEL, to be clear:

```c#
public static void Main(string[] args)
{
    var logger = NLog.Web.NLogBuilder.ConfigureNLog("nlog.config").GetCurrentClassLogger(); // No DI possible...
    try
    {
        logger.Debug("init main");
        CreateHostBuilder(args).Build().Run(); // Somewhere here reload for DI
    }
    catch (Exception exception)
    {
        //NLog: catch setup errors
        logger.Error(exception, "Stopped program because of exception");
        throw;
    }
    finally
    {
        // Ensure to flush and stop internal timers/threads before application-exit (Avoid segmentation fault on Linux)
        NLog.LogManager.Shutdown();
    }
}
```

snakefoot:
> Should: We should replace the DI with another DI container

Maybe change this into that one can register an external `IServiceProvider` in the NLog ServiceRepository. And allow resolving it again by calling `ResolveService<IServiceProvider>()`.

snakefoot:
> Next step again would be to handle buffering of LogEvents until lazy initialization has completed.

Created PullRequest #4055 that implements this feature.

snakefoot:
Updated wiki-page: https://github.com/NLog/NLog/wiki/Dependency-injection-with-NLog

PR #3833, opened 2020-03-07
Added ${OnHasProperties} -  for easier custom output

snakefoot:
Can do this (Inspired by [OnException](https://github.com/nlog/NLog/wiki/OnException-Layout-Renderer)):

> `${onhasproperties:, Properties\: ${all-event-properties}}`

Instead of this (performance hit from double rendering):

> `${when:when=length('${all-event-properties}')>0:inner=, Properties\: ${all-event-properties}}`

I guess when #3830 has been merged, then one could also create a static condition-method called `HasProperties(LogEventInfo)`.

Note merge to master

### DIFF
--- a/src/NLog/LayoutRenderers/AllEventPropertiesLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/AllEventPropertiesLayoutRenderer.cs
@@ -171,44 +171,44 @@ namespace NLog.LayoutRenderers
         /// <param name="logEvent">Logging event.</param>
         protected override void Append(StringBuilder builder, LogEventInfo logEvent)
         {
-            if (logEvent.HasProperties)
+            if (!logEvent.HasProperties)
+                return;
+
+            var formatProvider = GetFormatProvider(logEvent);
+            bool checkForExclude = CheckForExclude(logEvent);
+            bool nonStandardFormat = _beforeKey == null || _afterKey == null || _afterValue == null;
+
+            bool first = true;
+            foreach (var property in logEvent.Properties)
             {
-                var formatProvider = GetFormatProvider(logEvent);
-                bool checkForExclude = CheckForExclude(logEvent);
-                bool nonStandardFormat = _beforeKey == null || _afterKey == null || _afterValue == null;
+                if (!IncludeEmptyValues && IsEmptyPropertyValue(property.Value))
+                    continue;
 
-                bool first = true;
-                foreach (var property in logEvent.Properties)
+                if (checkForExclude && property.Key is string propertyKey && Exclude.Contains(propertyKey))
+                    continue;
+
+                if (!first)
                 {
-                    if (!IncludeEmptyValues && IsEmptyPropertyValue(property.Value))
-                        continue;
+                    builder.Append(Separator);
+                }
 
-                    if (checkForExclude && property.Key is string propertyKey && Exclude.Contains(propertyKey))
-                        continue;
+                first = false;
 
-                    if (!first)
-                    {
-                        builder.Append(Separator);
-                    }
-
-                    first = false;
-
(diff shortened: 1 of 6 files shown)


304NotModified:
thanks!

I think "on" are event things.

What about "when-has-properties" or "has-properties"

snakefoot:
If going the when-lane, then I would probably just stick with `hasProperties` when you have resolved #3830

304NotModified:
> If going the when-lane, then I would probably just stick with `hasProperties`-condition-method, **when** you have resolved #3830

I'm missing something. I don't see the relationship to #3830 (also not in https://github.com/NLog/NLog/pull/3833#issue-385206099)). Please elaborate, thanks!

snakefoot:
Then you can do this without allocation:

`${when:when=HasProperties:inner=Properties:}`

Or this:

`${when:when=HasException:inner=Properties:}`

Using condition-methods with single parameter.

src/NLog/LayoutRenderers/Wrappers/OnHasPropertiesLayoutRendererWrapper.cs
```diff
@@ -0,0 +1,62 @@
...
+// THE POSSIBILITY OF SUCH DAMAGE.
+// 
+
+namespace NLog.LayoutRenderers.Wrappers
+{
+    using System.Text;
+    using NLog.Config;
+
+    /// <summary>
+    /// Outputs alternative layout when the inner layout produces empty result.
```
304NotModified: Please add the example to the code, thanks!

src/NLog/LayoutRenderers/AllEventPropertiesLayoutRenderer.cs
```diff
@@ -223,7 +223,7 @@ private bool CheckForExclude(LogEventInfo logEvent)
                 checkForExclude = !(logEvent.CallSiteInformation == null && Exclude.Count == CallerInformationAttributeNames.Length && Exclude.Contains(CallerInformationAttributeNames[0]));
             }
 #endif
-            return checkForExclude && logEvent.HasProperties;
```
304NotModified: double check, this check isn't needed? (probably check in the calling method?)

304NotModified:
> Then you can do this without allocation:
> 
> `${when:when=HasProperties:inner=Properties:}`
> 
> Or this:
> 
> `${when:when=HasException:inner=Exception:}`
> 
> Using condition-methods with single LogEventInfo-parameter.

Thanks, clear now! (the confusion was somewhere else, but resolved)

src/NLog/LayoutRenderers/AllEventPropertiesLayoutRenderer.cs
snakefoot: No it has already been checked before calling this method.

snakefoot:
Think I was too eager with plans for optimizing when-condition-methods. The method-factory will require some refactoring for this to occur. Think I will stick with:

> `${onhasproperties:, Properties\: ${all-event-properties}}`

src/NLog/LayoutRenderers/Wrappers/OnHasPropertiesLayoutRendererWrapper.cs
snakefoot: Added example code

304NotModified: Thanks!

snakefoot:
Created wiki-page: https://github.com/NLog/NLog/wiki/OnHasProperties-Layout-Renderer

snakefoot:
Create Doc-PR: https://github.com/NLog/NLog.github.io/pull/141

PR #3842, opened 2020-03-16
LogLevel - Compare operators should not throw, translate null into LogLevel.Off

snakefoot:
Minor change of Code Layout. See also #3840


304NotModified:
I don't think this PR will change the behavior?  👼

snakefoot:
Probably right. Just gave it the same labels as #3840

PR #3843, opened 2020-03-16
DatabaseTarget - RenderObjectValue should not lookup Type-name for InternalLogger.Trace

snakefoot:
Minor performance optimization


304NotModified:
Thanks!

PR #3853, opened 2020-03-22
LogFactory - Marked broken constructor as obsolete. It is a trap.

snakefoot:
See also: https://stackoverflow.com/questions/60803043/nlog-for-logging-using-nlog-api-into-application-insight-table-storage

### DIFF
--- a/src/NLog/Config/XmlLoggingConfiguration.cs
+++ b/src/NLog/Config/XmlLoggingConfiguration.cs
@@ -196,11 +196,22 @@ namespace NLog.Config
         /// <summary>
         /// Parse XML string as NLog configuration
         /// </summary>
-        /// <param name="xml">NLog configuration</param>
+        /// <param name="xml">The XML contents.</param>
         /// <returns></returns>
         public static XmlLoggingConfiguration CreateFromXmlString(string xml)
         {
-            return new XmlLoggingConfiguration(xml, string.Empty, LogManager.LogFactory);
+            return CreateFromXmlString(xml, LogManager.LogFactory);
+        }
+
+        /// <summary>
+        /// Parse XML string as NLog configuration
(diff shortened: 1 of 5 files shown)


src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -196,11 +196,22 @@ internal XmlLoggingConfiguration([NotNull] string xmlContents, [CanBeNull] strin
         /// <summary>
         /// Parse XML string as NLog configuration
         /// </summary>
-        /// <param name="xml">NLog configuration</param>
+        /// <param name="xml">The XML contents.</param>
```
304NotModified, line 199: proposal: "NLog configuration as XML string"

(I find The XML contents confusing)

304NotModified: same below

src/NLog/LogFactory.cs
```diff
@@ -114,6 +114,7 @@ public LogFactory()
         /// Initializes a new instance of the <see cref="LogFactory" /> class.
         /// </summary>
         /// <param name="config">The config.</param>
+        [Obsolete("Constructor with LoggingConfiguration as parameter should not be used. Instead provide LogFactory as parameter when constructing LoggingConfiguration.")]
```
304NotModified, line 117: add the text, obsolete in NLog 5? (or something like that)

snakefoot: Fixed

src/NLog/Config/XmlLoggingConfiguration.cs
snakefoot: Fixed

304NotModified:
Thanks!

Issue #3856, opened 2020-03-24
Allow AsyncWrapper to flush in batches instead of one giant flush


snakefoot:
Have an application that writes 8 GByte data using NLog as part of startup-rutine.

After startup-rutine has completed, and just about to go online, then it performs a blocking garbage collection.

It could be nice if one could flush NLog buffers before starting the garbage collection. The problem is that async-file-target might have a lot of data pending at this point.

If one calls NLog LogManager.Flush() then it will take the entire async-queue-buffer and write in one operation.

Lets say 200.000 items are still pending and each of them is 1000 bytes. Then it becomes close to 2 gigabyte which is the upper limit for a MemoryStream. Calling NLog LogManager.Flush() will then either lead to application crashing with out-of-memory exception, or discarding data because of stream-is-too-long.

It could be nice if AsyncWrapper had a MaxBatchSize (nullable) that could be configured instead of BatchSize. And it would also have effect during flush-operations. Instead of writing 200.000 in one operation, then it would do 200 write-operations of MaxBatchSize=1000 

Alternative solution would be to change FileTarget so it has a hardlimit of 1 MByte. So if the encoding-buffer has grown beyond this hardcoded max-size. Then it flushes the encoding-buffer to file, before continuing on processing the current batch (of maybe 200.000 items).

snakefoot:
Now implemented small work-around for FileTarget that ensures flushing will perform writing in batches.

PR #3857, opened 2020-03-24
FileTarget - Batch write to filestream in max chunksize of 100 times BufferSize

snakefoot:
Partial resolve of #3856, so LogManager.Flush() will not generate out-of-memory-exception in file-target.

Started with 1 MByte. But then recalled KeepFileOpen=false works a lot better with really huge buffers.

fixes https://github.com/NLog/NLog/issues/3889

### DIFF
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -1118,10 +1118,9 @@ namespace NLog.Targets
 
             var buckets = logEvents.BucketSort(_getFullFileNameDelegate);
 
-            using (var reusableStream = (OptimizeBufferReuse && logEvents.Count <= 1000) ? _reusableAsyncFileWriteStream.Allocate() : _reusableAsyncFileWriteStream.None)
-            using (var allocatedStream = reusableStream.Result != null ? null : new MemoryStream())
+            using (var reusableStream = OptimizeBufferReuse ? _reusableAsyncFileWriteStream.Allocate() : _reusableAsyncFileWriteStream.None)
             {
-                var ms = allocatedStream ?? reusableStream.Result;
+                var ms = reusableStream.Result ?? new MemoryStream();
 
                 foreach (var bucket in buckets)
                 {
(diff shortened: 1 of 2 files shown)


snakefoot:
Btw. curious what is the point of the `WrappedFileTargetTests` ? Cannot see how it adds any extra test-coverage (Besides running all tests twice).

snakefoot:
| Test Name        | Messages   | Size | Args | Threads |
|------------------|------------|------|------|---------|
| PerfLogger       |  2.000.000 | 1024 |    0 |       1 |

| Test Name        | Time (ms) | Msgs/sec  | GC2 | GC1 | GC0 | CPU (ms) | Alloc (MB) |
| NLog 4.7 |     7.906 |   252.947 |   1 |   2 |   5 |    8.515 |      414,1 |
| PR |     7.128 |   280.558 |   1 |   3 |   5 |    7.484 |      413,0 |

```xml
      <target name="AsyncFilePerf" type="AsyncWrapper" batchSize="500" overflowAction="Block" timeToSleepBetweenBatches="0">
        <target name="SyncFilePerf" type="File" fileName="C:\Temp\Log\NLogPerformancePerf.txt" autoFlush="false" keepFileOpen="true" concurrentWrites="false" layout="${message}" />
      </target>
```

304NotModified:
> Btw. curious what is the point of the `WrappedFileTargetTests` ? Cannot see how it adds any extra test-coverage (Besides running all tests twice).

I don't really know why we need it. I see it's added here: https://github.com/NLog/NLog/commit/8d9572b9009e0b01c02233ac49f965721ead58d7

snakefoot:
> @304NotModified  I don't really know why we need it. I see it's added here: 8d9572b

Think I have guessed why. It is because this method is not called by default:

`protected override void Write(IList<AsyncLogEventInfo> logEvents)`

So the duplication of all tests is to ensure this method is exercised. Think we could exercise this method without running ex. all archive-logic tests twice. Since the "end result" is calling this method:

`private void ProcessLogEvent(LogEventInfo logEvent, string fileName, ArraySegment<byte> bytesToWrite)`

304NotModified:
> hink I have guessed why. It is because this method is not called by default:
> 
> `protected override void Write(IList<AsyncLogEventInfo> logEvents)`

Maybe we should have one test for it, not all of them duplicated?

snakefoot:
> Maybe we should have one test for it, not all of them duplicated?

Was thinking the same. Maybe for another PR :)

src/NLog/Targets/FileTarget.cs
```diff
@@ -2203,12 +2219,6 @@ private void ConditionalFlushOpenFileAppenders()
         /// <param name="initializedNewFile">File has just been opened.</param>
         private void WriteToFile(string fileName, ArraySegment<byte> bytes, bool initializedNewFile)
         {
-            if (ReplaceFileContentsOnEachWrite)
```
304NotModified, line 2206: is this correct? This a behavior change? (WriteToFile calls this one, but ok, replace will be strange with a footer)

snakefoot: I think this is more correct. Since as you have also noticed, that WriteToFile is reused when writing Footer.

304NotModified: maybe create an issue and resolve it with this PR? It's a bit hidden now.

snakefoot: I'm ok it is hidden. Think `ReplaceFileContentsOnEachWrite` is something very occult and special.

snakefoot:
@304NotModified Changed the hardcoded 10 MegaByte to be 100 times the configured BufferSize. Added unit-test that verifies that AsyncContinuation is called in correct order.

src/NLog/Targets/FileTarget.cs
304NotModified: Issue #3889

304NotModified: I'm wrong here? There is no bug? See unit test ReplaceFileContentsOnEachWriteTest

snakefoot: There is a bug when using `ReplaceFileContentsOnEachWrite` together with  `Footer`. But both features are rather exotic.

304NotModified: but that case unit tested in ReplaceFileContentsOnEachWriteTest?

snakefoot: Again exotic features, so you have to read the code to get the documentation. Mind I didn't create `ReplaceFileContentsOnEachWrite` :). But footer is written by FinalizeFile, so if the ReplaceFileContentsOnEachWrite moved the FileTarget into a state where it performed Finalize then it would fail. I guess `WriteFooter`-method should return early when `ReplaceFileContentsOnEachWrite = true`.

snakefoot:
Created #4263 for removing `WrappedFileTargetTests`

PR #3861, opened 2020-03-25
NLog InternalLogger Setup: added AddLogSubscription and RemoveLogSubscription

snakefoot:
Fluent API for LogMessageReceived

```c#
                LogManager.Setup().SetupInternalLogger(s =>
                {
                    s.AddLogSubscription(eventHandler);
                });
```

### DIFF
--- a/src/NLog/SetupInternalLoggerBuilderExtensions.cs
+++ b/src/NLog/SetupInternalLoggerBuilderExtensions.cs
@@ -88,5 +88,23 @@ namespace NLog
             InternalLogger.LogWriter = writer;
             return setupBuilder;
         }
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


304NotModified:
I think it should have the term "internal" in the name, otherwise I would think this will be triggered when I log a message to a logger.

So  AddInternalLogSubscription and RemoveInternalLogSubscription?

snakefoot:
But we are already in the Setup-method called `SetupInternalLogger`. Just like LogToFile doesn't say InternalLogToFile.

304NotModified:
Ow yes indeed, missed that!

snakefoot:
Updated wiki-page: https://github.com/NLog/NLog/wiki/Fluent-Configuration-API

PR #3867, opened 2020-03-27
InternalLogger - Logging NLog version should include whether it is from GlobalAssemblyCache

snakefoot:


304NotModified:
Good idea!

PR #3868, opened 2020-03-27
SplitGroupTarget - Target write failure should not skip remaining targets

snakefoot:
The following bugs has been fixed:
- Writing a single LogEvent will not skip remaining targets, if first target fails with exception (Funny observation is that if writing a batch of LogEvents then it behaved correctly)
- Writing a batch of LogEvents without any configured targets should still ensure to call async continuation (Funny observation is that if writing a single LogEvent then it behaved correctly).

Have updated unit-tests to verify that both scenarios are working for single LogEvent and batch of LogEvents.

Have added minor optimization when writing batch with only a single LogEvent, then it skips array-allocation for each target destination.

Could consider marking `AsyncHelpers.ForEachItemSequentially` as obsolete, since no longer used.

### DIFF
--- a/src/NLog/Targets/Wrappers/SplitGroupTarget.cs
+++ b/src/NLog/Targets/Wrappers/SplitGroupTarget.cs
@@ -36,8 +36,7 @@ namespace NLog.Targets.Wrappers
     using System;
     using System.Collections.Generic;
     using System.Threading;
-    using Common;
-    using Internal;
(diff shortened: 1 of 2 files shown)


304NotModified:
> Could consider marking `AsyncHelpers.ForEachItemSequentially` as obsolete, since no longer used.

Good idea, but only allowed for a new minor version according to semver

PR #3884, opened 2020-04-07
InternalLogger - Include FilePath when logging config is being initialized

snakefoot:
Instead of writing:

> Found 28 configuration items

Then it will write:

> Validating config: TargetNames=f, c, ConfigItems=28, FilePath=C:\WindowsService\GasStationService\Nlog.config

### DIFF
--- a/src/NLog/Config/XmlLoggingConfiguration.cs
+++ b/src/NLog/Config/XmlLoggingConfiguration.cs
@@ -602,5 +602,11 @@ namespace NLog.Config
             return Path.GetFullPath(fileName);
 #endif
         }
+
+        /// <inheritdoc />
+        public override string ToString()
+        {
+            return $"{base.ToString()}, FilePath={_originalFileName}";
+        }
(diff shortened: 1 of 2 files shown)


src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -602,5 +602,11 @@ private static string GetFileLookupKey([NotNull] string fileName)
             return Path.GetFullPath(fileName);
 #endif
         }
+
+        /// <inheritdoc />
+        public override string ToString()
+        {
+            return $"{base.ToString()}, XmlFile={_originalFileName}";
```
304NotModified, line 609: proposal: "config file" or "xml config file path" instead of XmlFile

snakefoot: Changed to FilePath

304NotModified:
Thanks!

PR #3891, opened 2020-04-14
ColoredConsoleTarget - Allow using Condition-logic for word highlighting rules

snakefoot:
Condition-Logic is also faster than RegEx. Especially when combined with #3894

### DIFF
--- a/src/NLog/Targets/ColoredConsoleAnsiPrinter.cs
+++ b/src/NLog/Targets/ColoredConsoleAnsiPrinter.cs
@@ -62,7 +62,7 @@ namespace NLog.Targets
             }
         }
 
-        public ConsoleColor? ChangeForegroundColor(TextWriter consoleWriter, ConsoleColor? foregroundColor)
+        public ConsoleColor? ChangeForegroundColor(TextWriter consoleWriter, ConsoleColor? foregroundColor, ConsoleColor? oldForegroundColor = null)
         {
             if (foregroundColor.HasValue)
             {
@@ -71,7 +71,7 @@ namespace NLog.Targets
             return null;    // There is no "old" console color
         }
 
-        public ConsoleColor? ChangeBackgroundColor(TextWriter consoleWriter, ConsoleColor? backgroundColor)
+        public ConsoleColor? ChangeBackgroundColor(TextWriter consoleWriter, ConsoleColor? backgroundColor, ConsoleColor? oldBackgroundColor = null)
         {
             if (backgroundColor.HasValue)
             {
--- a/src/NLog/Targets/ConsoleWordHighlightingRule.cs
+++ b/src/NLog/Targets/ConsoleWordHighlightingRule.cs
@@ -152,7 +152,7 @@ namespace NLog.Targets
 
         internal MatchCollection Matches(LogEventInfo logEvent, string message)
         {
-            if (Condition != null && !true.Equals(Condition.Evaluate(logEvent)))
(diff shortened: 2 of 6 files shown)


snakefoot:
Before tuning:

image

After tuning (Faster handling of newlines and reduced number of console color changes):

image

304NotModified:
Cool!

Would be nice if we could cover these lines:

image


image

snakefoot:
> Would be nice if we could cover these lines:

Yes I'm missing unit-tests. Just playing around with comparing Microsoft AddConsole and NLog ColoredConsole. Noticed that Microsoft AddConsole is not cheap.

snakefoot:
@304NotModified Was just in a battle with code-smells, and needed some retries before it worked. Also the Azure-Build-Server sometimes like to timeout, so one needs an extra run.

304NotModified:
Well it's stuck on azure then,as the timeout is 60min and it should be < 20min.

src/NLog/Targets/ColoredConsoleAnsiPrinter.cs
```diff
@@ -51,7 +51,7 @@ public TextWriter AcquireTextWriter(TextWriter consoleStream, StringBuilder reus
             return new StringWriter(reusableBuilder ?? new StringBuilder(50), consoleStream.FormatProvider);
         }
 
-        public void ReleaseTextWriter(TextWriter consoleWriter, TextWriter consoleStream, ConsoleColor? oldForegroundColor, ConsoleColor? oldBackgroundColor, bool flush)
+        public void ReleaseTextWriter(TextWriter consoleWriter, TextWriter consoleStream, ConsoleColor? orgForegroundColor, ConsoleColor? orgBackgroundColor, bool flush)
```
304NotModified, line 54: please write out org, prev, etc

src/NLog/Targets/ConsoleWordHighlightingRule.cs
```diff
@@ -143,8 +150,13 @@ public bool IgnoreCase
         /// </summary>
         public Regex CompiledRegex => _regexHelper.Regex;
 
-        internal MatchCollection Matches(string message)
+        internal MatchCollection Matches(LogEventInfo logEvent, string message)
         {
+            if (Condition != null && !true.Equals(Condition.Evaluate(logEvent)))
```
304NotModified, line 155: false.Equals ;)?

snakefoot: No opinion. Will fix.

src/NLog/Targets/ColoredConsoleAnsiPrinter.cs
snakefoot: Uhh suddenly old is good again. Will change to old then

snakefoot:
Sigh. Now my refactoring revealed that we never really tested other control-characters than newline. New unit-test is needed.

304NotModified:
> Sigh. Now my refactoring revealed that we never really tested other control-characters than newline. New unit-test is needed.

So convert this PR to draft? Or is this fixed in the force push?

snakefoot:
>So convert this PR to draft? Or is this fixed in the force push?

Fixed as code-coverage is back at 100%

304NotModified:
ow indeed! NICE!

304NotModified:
Thanks! merged!

snakefoot:

PR #3895, opened 2020-04-14
Property type converter: fix converting to nullable value + lot of unit tests

304NotModified:


src/NLog/Config/PropertyTypeConverter.cs
```diff
@@ -74,13 +74,25 @@ public object Convert(object propertyValue, Type propertyType, string format, IF
...
+            var nullableType = Nullable.GetUnderlyingType(propertyType);
+            Type t = nullableType ?? propertyType;
+
+            var newValue = System.Convert.ChangeType(propertyValue, t, formatProvider);
+
+            if (nullableType != null)
+            {
+                switch (nullableType.Name)
+                {
+                    case "Int32": return (int?)newValue;
```
304NotModified, line 86: quickfix and won't even work well 🤣

src/NLog/Config/PropertyTypeConverter.cs
```diff
@@ -43,28 +44,40 @@ class PropertyTypeConverter : IPropertyTypeConverter
...
         public object Convert(object propertyValue, Type propertyType, string format, IFormatProvider formatProvider)
         {
-            if (propertyType == null || propertyValue == null || propertyValue.GetType() == propertyType || propertyType == typeof(object))
+            if (!NeedToConvert(propertyValue, propertyType))
             {
                 return propertyValue;
             }
 
+            bool? isNullable = null;
+            Type nullableType = null;
```
snakefoot, line 53: Should just be `Type nullableType = Nullable.GetUnderlyingType(propertyType);`, since you are doing the lookup no matter what. (Then you can also remove `bool? isNullable `).

304NotModified: indeed! Thanks!

snakefoot: Could consider having a conversion-dictionary. Where common types are registered along with their converter. Similar to what has been implemented in PropertyHelper. The conversion dictionary could have the type `DateTime`, `DateTime?`, `Guid`, `Guid?`, `int`, `int?`, `bool`, `bool?`, `string`. This would make the conversion path quicker.

src/NLog/Config/PropertyTypeConverter.cs
```diff
@@ -43,28 +44,35 @@ class PropertyTypeConverter : IPropertyTypeConverter
...
             {
                 propertyValue = propertyString = propertyString.Trim();
 
-                if (propertyType == typeof(DateTime))
+                if (nullableType != null && StringHelpers.IsNullOrWhiteSpace(propertyString))
+                {
+                    return null;
+                }
+                
+                if (type == typeof(DateTime))
```
snakefoot, line 63: Could consider having a conversion-dictionary. Where common types are registered along with their converter. Similar to what has been implemented in PropertyHelper. The conversion dictionary could have the type `DateTime`, `DateTime?`, `Guid`, `Guid?`, `int`, `int?`, `bool`, `bool?`, `string`. This would make the conversion path quicker.

snakefoot: I was think the latter, as one didn't have to lookup the underlying for common types, as one would just register both-types in the lookup-dictionary. Just a thought.

304NotModified:
Travis is green, not sure why it doesn't report back:

image

src/NLog/Config/PropertyTypeConverter.cs
304NotModified: Maybe nice to do this in another PR :)

PR #3896, opened 2020-04-14
Write internallog warning for unrecognized properties used in config on layout renderers

snakefoot:
LayoutParser - InternalLogger Warnings when skipping unrecognized properties for LayoutRenderers

### DIFF
--- a/src/NLog/Internal/SimpleStringReader.cs
+++ b/src/NLog/Internal/SimpleStringReader.cs
@@ -31,10 +31,10 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
-
-
 namespace NLog.Internal
 {
+    using System;
+
     /// <summary>
     /// Simple character tokenizer.
     /// </summary>
@@ -122,6 +122,23 @@ namespace NLog.Internal
         internal string Substring(int startIndex, int endIndex)
         {
             return _text.Substring(startIndex, endIndex - startIndex);
+        }
+
+        internal string ReadUntilMatch(Func<int, bool> charFinder)
+        {
+            int ch;
+            int startPosition = Position;
+            while ((ch = Peek()) != -1)
+            {
+                if (charFinder(ch))
+                {
+                    return Substring(startPosition, Position);
+                }
+
+                Read();
+            }
+
+            return Substring(startPosition, Position);
         }
     }
 }
--- a/src/NLog/Layouts/LayoutParser.cs
+++ b/src/NLog/Layouts/LayoutParser.cs
@@ -147,21 +147,7 @@ namespace NLog.Layouts
 
         private static string ParseLayoutRendererName(SimpleStringReader sr)
         {
-            int ch;
-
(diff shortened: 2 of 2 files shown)


304NotModified:
Hi @snakefoot 

Thanks for the PR. Unfortunately I don't fully understand what we here trying to achieve. Does this PR add a new warning to the internal log when something is skipped? What is an example of "something" here?

PS: you could create drafts now. Please do, because I get a notification for every push now (have converted to draft now). If you think it's ready for review, please undraft. Thanks!

snakefoot:
Let's say you spell something wrong in a property-Nam. Like this: tolowercase=true


Then it will not configure anything but just discard the unrecognized property value for the layoutrenderer.


I would like to have a warning when that happens. The PR is done, but playing with build server and exotic target frameworks.




Sent from my Sony Xperia


Hi @snakefoot<https://github.com/snakefoot>

Thanks for the PR. Unfortunately I don't fully understand what we here trying to achieve. Does this PR add a new warning to the internal log when something is skipped? What is an example of "something" here?

PS: you could create drafts now. Please do, because I get a notification for every push now (have converted to draft now). If you think it's ready for review, please undraft. Thanks!

snakefoot:
Oops. Had to fix some unnecessary Type-Name-Lookup when doing debug-logging in `SetPropertyFromString`

304NotModified:
> I would like to have a warning when that happens. The PR is now done.

Cool! I renamed the PR, maybe a bit long. Feel free to rename it (but I prefer to keep it "functional")

snakefoot:
Fine with me. Though it is only for LayoutRenderers

snakefoot:
Looked at the code for assigning default-parameter-value. I guess one is doing this:

`${exception }`

Then one will not get any output, as it would assign space ' ' to the default-parameter `Format`.

src/NLog/Internal/SimpleStringReader.cs
```diff
@@ -123,5 +123,22 @@ internal string Substring(int startIndex, int endIndex)
         {
             return _text.Substring(startIndex, endIndex - startIndex);
         }
+
+        internal string ReadUntilMatch(Func<int, bool> charFinder)
```
304NotModified, line 127: reuse 👍

src/NLog/Layouts/LayoutParser.cs
```diff
@@ -219,12 +205,24 @@ private static string ParseParameterName(SimpleStringReader sr)
...
+            return simpleValue;
+        }
+
+        private static void ParseParameterUnicodeValue(SimpleStringReader sr, StringBuilder nameBuf)
+        {
+            int ch;
             while ((ch = sr.Peek()) != -1)
             {
-                if (ch == ':' || ch == '}')
+                if (EndOfLayout(ch))
```
304NotModified, line 225: 👍

304NotModified:
Thanks!

PR #3898, opened 2020-04-15
ObjectGraphScanner - Reuse reflection cache from PropertyHelper

snakefoot:
Caches the lookup of `NLogConfigurationIgnorePropertyAttribute`, so object-scanning is cheaper.

Caches the lookup of `NLogConfigurationItemAttribute`, so object-scanning is cheaper.

Skips checking if required parameters are configured for Targets, since this check has already been performed when initializing LoggingConfiguration (Skip double checking)

Also skips checking for `ArrayParameterAttribute` when calling property-setter with standard type (Like string / int / etc.).

My sample app in #3618. Went from 3 sec down to 2 sec.

### DIFF
--- a/src/NLog/Targets/Target.cs
+++ b/src/NLog/Targets/Target.cs
@@ -417,7 +417,9 @@ namespace NLog.Targets
 
                 if (!IsInitialized)
                 {
-                    PropertyHelper.CheckRequiredParameters(this);
+                    if (configuration == null)
+                        PropertyHelper.CheckRequiredParameters(this);
+
                     try
                     {
                         InitializeTarget();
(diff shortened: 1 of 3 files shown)


src/NLog/Targets/Target.cs
```diff
@@ -417,7 +417,9 @@ internal void Initialize(LoggingConfiguration configuration)
 
                 if (!IsInitialized)
                 {
-                    PropertyHelper.CheckRequiredParameters(this);
+                    if (configuration == null)
```
304NotModified, line 420: Why do we need this here? We have already checked IsInitialized

snakefoot: Because some unit-tests still create and use targets without using the front-door. Notice how the code-coverage is all green.

PR #3901, opened 2020-04-18
MappedDiagnosticsLogicalContext - SetScoped with IReadOnlyList for Xamarin

snakefoot:
Curious if a workaround can be made for: https://github.com/NLog/NLog.Extensions.Logging/issues/415

For some "strange" reason then even if using NetCore / NetStandard on Xamarin, then it insists on using the Xamarin-builds instead of the NetStandard-build (But only when using the linker)

### DIFF
--- a/src/NLog/Internal/ReflectionHelpers.cs
+++ b/src/NLog/Internal/ReflectionHelpers.cs
@@ -349,7 +349,7 @@ namespace NLog.Internal
 
         public static object GetPropertyValue(this PropertyInfo p, object instance)
         {
-#if NET45
+#if NET4_5
             return p.GetValue(instance);
 #else
             return p.GetGetMethod().Invoke(instance, null);
--- a/src/NLog/Config/ConfigurationItemFactory.cs
+++ b/src/NLog/Config/ConfigurationItemFactory.cs
@@ -586,7 +586,7 @@ namespace NLog.Config
             }
 #endif
 
-#if NET4_5 || NETSTANDARD
+#if !SILVERLIGHT && !NET3_5 && !NET4_0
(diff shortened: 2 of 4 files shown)


snakefoot:
Hurray it builds. Guess this will fix: NLog/NLog.Extensions.Logging#415

304NotModified:
> Actually surprised how much functionality is included for Xamari

What do you mean? On what  target Framework?

snakefoot:
> What do you mean? 

I mean it is a good thing

> On what target Framework?

Xamarin Framework

snakefoot:
It has been confirmed that the PR is working. Think it should be included with NLog 4.7.1

NLog 5 will hopefully remove the Xamarin-Targets and only use NetStandard/NetFramework.

src/NLog/Internal/ReflectionHelpers.cs
```diff
@@ -349,7 +349,7 @@ public static bool IsValidPublicProperty(this PropertyInfo p)
 
         public static object GetPropertyValue(this PropertyInfo p, object instance)
         {
-#if NET45
+#if NET4_5
```
304NotModified, line 352: oops looks like a bug (before)

304NotModified:
Thanks!

src/NLog/Internal/ReflectionHelpers.cs
snakefoot: Yes was scanning for places where NET4_5 was used directly. I guess we should always opt-out, instead of having these opt-in situations that bites.

304NotModified: Yes opt out is preferred, also when removing platforms

PR #3902, opened 2020-04-18
LogEventInfo - PropertiesDictionary should not clone internal generated param list

snakefoot:
Should only protect itself from message-parameter-lists received as input-parameter (Ex. from NLog.Extensions.Logging)

The message-parameter-list received from NLog own internal parser is safe to store directly.


304NotModified:
nice! thanks!

PR #3903, opened 2020-04-19
Layout - Object scanning should only check class-attributes for LayoutRenderers

snakefoot:
Only Layouts and LayoutRenderers are relevant. Should ignore NLogConfigurationItem-objects.

It is confusing for NLog users that they need to add additional special attributes to custom-classes marked with NLogConfigurationItem (When creating their own custom targets/layouts). If forgetting then NLog-Layout-engine will see everything as "unsafe".

### DIFF
--- a/src/NLog/Layouts/JsonAttribute.cs
+++ b/src/NLog/Layouts/JsonAttribute.cs
@@ -40,9 +40,6 @@ namespace NLog.Layouts
     /// JSON attribute.
     /// </summary>
     [NLogConfigurationItem]
-    [ThreadAgnostic]
-    [ThreadSafe]
-    [AppDomainFixedOutput]
     public class JsonAttribute
     {
         /// <summary>
--- a/src/NLog/Conditions/ConditionExpression.cs
+++ b/src/NLog/Conditions/ConditionExpression.cs
@@ -31,20 +31,17 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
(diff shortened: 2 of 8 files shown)


src/NLog/Layouts/JsonAttribute.cs
```diff
@@ -40,9 +40,6 @@ namespace NLog.Layouts
     /// JSON attribute.
     /// </summary>
     [NLogConfigurationItem]
-    [ThreadAgnostic]
-    [ThreadSafe]
-    [AppDomainFixedOutput]
```
304NotModified, line 45: So this is not needed anymore for all attributes?

snakefoot: No objects that are just marked with `NLogConfigurationItem` (To signal they are safe for NLog to probe) doesn't need the extra attributes. The extra attributes are only for Layout and LayoutRenderers.

304NotModified: Could we please add that to the docs of those attributes? Then I will merge it 

(please no force push)

snakefoot: Not sure I understand. Should I write that attributes intended for Layout and LayoutRenderers should only need to be used for Layout and LayoutRenderers ?

304NotModified: Ah it was already there. Surprising!

PR #3904, opened 2020-04-19
ObjectReflectionCache - Skip reflection for Module objects

snakefoot:
They are "dangerous" like Assembly and MethodInfo


304NotModified:
Thanks!

PR #3915, opened 2020-04-22
CaptureStackTrace=false to disable implicit callsite capture

### LINKED ISSUE #3914: Need control over the always-on call stack
**NLog version**: 4.7
**NLog version**: .Net 4.5

The idea of logs is predictable and controllable data. However, current solution of `LoggerImpl.Write()` always saves callstack. What if not all log records need this, but we need configureate `{stacktrace}`?. All I could find about this question is this: [stackoverflow](https://stackoverflow.com/questions/53119784/customizing-when-to-capture-stack-trace-in-nlog)

That's about what I'd like to get:
Config file where I can configure will stack trace always enabled or not. Second thing - layout config for stack trace like for exceptions. Config shouldn't mean, that stack trace should always be enabled.

```xml
<nlog throwConfigExceptions="true" throwExceptions="true" xmlns="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <targets async="true">
    <target name="xmlFile" xsi:type="File" fileName="E:\log.xml" maxArchiveFiles="3" archiveNumbering="Sequence" archiveDateFormat="dd-mm-yyyy" archiveOldFileOnStartup="true" stacktraceAlwaysEnabled="false">
      <layout xsi:type="XmlLayout" indentXml="true" includeAllProperties="true" maxRecursionLimit="10">
        <element name="message" value="${message}" />
        <element name="exception" value="${exception:format=toString}" />
        <element name="stacktrace" value="${custom.stacktrace:format=Raw:separator=&#13;&#10;}" /> 
      </layout>
    </target>
  </targets>
  <rules>
    <logger name="*" minlevel="Debug" writeTo="xmlFile" />
  </rules>
</nlog>

```
Next two messeges should give different logs:
First
```C#
Log.Debug()
      .Message("Hello")
      .Write();
```
Would give
```xml
<logevent>
  <message>Hello</message>
</logevent>
```

Second:  
```C#
 Log.Debug()
       .Message("Hello")
       .StackTrace(new StackTrace(), 0)
       .Write();
```
Would give
```xml
<logevent>
  <message>Hello</message>
  <stacktrace>some stack </stacktrace>
</logevent>
```
The behaviour make be the same like with exceptions.

I didn't find how to configurate same behaviour right now.

snakefoot:
Added CaptureStackTrace-property to the following layoutrenderers:
- StackTraceLayoutRenderer
- CallSiteLayoutRenderer
- CallSiteFileNameLayoutRenderer
- CallSiteLineNumberLayoutRenderer

Resolves #3914

### DIFF
--- a/src/NLog/Internal/CallSiteInformation.cs
+++ b/src/NLog/Internal/CallSiteInformation.cs
@@ -44,12 +44,23 @@ namespace NLog.Internal
         /// </summary>
         /// <param name="stackTrace">The stack trace.</param>
         /// <param name="userStackFrame">Index of the first user stack frame within the stack trace.</param>
-        /// <param name="userStackFrameLegacy">Index of the first user stack frame within the stack trace.</param>
-        public void SetStackTrace(StackTrace stackTrace, int userStackFrame, int? userStackFrameLegacy)
+        /// <param name="loggerType">Type of the logger or logger wrapper. This is still Logger if it's a subclass of Logger.</param>
+        public void SetStackTrace(StackTrace stackTrace, int? userStackFrame = null, Type loggerType = null)
         {
             StackTrace = stackTrace;
-            UserStackFrameNumber = userStackFrame;
-            UserStackFrameNumberLegacy = userStackFrameLegacy != userStackFrame ? userStackFrameLegacy : null;
+            if (!userStackFrame.HasValue && stackTrace != null)
+            {
+                var stackFrames = stackTrace.GetFrames();
+                var firstUserFrame = loggerType != null ? FindCallingMethodOnStackTrace(stackFrames, loggerType) : 0;
+                var firstLegacyUserFrame = firstUserFrame.HasValue ? SkipToUserStackFrameLegacy(stackFrames, firstUserFrame.Value) : firstUserFrame;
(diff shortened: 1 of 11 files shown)


snakefoot:
Will probably be a challenge to merge from master into dev, and some extra work has to be done to make sure the special caller-member-optimization in dev-branch still works.

snakefoot:
Alternative property name could be `CaptureStackTrace`, then it more clear what the capture is :)

304NotModified:
I think that's a better name indeed :+1:

snakefoot:
> I think that's a better name indeed

Property renamed from `ForceCapture` to `CaptureStackTrace`

snakefoot:
Updated wiki pages:

https://github.com/NLog/NLog/wiki/Stack-Trace-Layout-Renderer
https://github.com/NLog/NLog/wiki/Callsite-layout-renderer
https://github.com/NLog/NLog/wiki/Callsite-file-name-layout-renderer
https://github.com/NLog/NLog/wiki/Callsite-line-number-layout-renderer

PR #3929, opened 2020-04-28
Support applyOnReload for LoadConfiguration

304NotModified:
Fixes https://github.com/NLog/NLog/issues/1429

Todo 

- [ ] Fix default
- [ ] Event should be before loading
- [ ] Check other PR for event

### DIFF
--- a/src/NLog/Config/XmlLoggingConfiguration.cs
+++ b/src/NLog/Config/XmlLoggingConfiguration.cs
@@ -272,7 +272,7 @@ namespace NLog.Config
         public override LoggingConfiguration Reload()
         {
             if (!string.IsNullOrEmpty(_originalFileName))
-                return LogFactory.CreateConfig(_originalFileName); //TODO NLog 5: should call LoadConfiguration which set LogFactory.Configuration
+                return LogFactory.CreateConfig(_originalFileName);
             else
                 return base.Reload();
         }
--- a/src/NLog/Config/XmlLoggingConfiguration.cs
+++ b/src/NLog/Config/XmlLoggingConfiguration.cs
@@ -272,7 +272,7 @@ namespace NLog.Config
         public override LoggingConfiguration Reload()
         {
             if (!string.IsNullOrEmpty(_originalFileName))
-                return LogFactory.CreateConfig(_originalFileName);
+                return new XmlLoggingConfiguration(_originalFileName, LogFactory);
             else
                 return base.Reload();
         }
--- a/src/NLog/Config/XmlLoggingConfiguration.cs
+++ b/src/NLog/Config/XmlLoggingConfiguration.cs
@@ -272,7 +272,7 @@ namespace NLog.Config
         public override LoggingConfiguration Reload()
         {
             if (!string.IsNullOrEmpty(_originalFileName))
-                return new XmlLoggingConfiguration(_originalFileName, LogFactory);
+                return LogFactory.CreateConfig(_originalFileName); //TODO NLog 5: should call LoadConfiguration which set LogFactory.Configuration
             else
                 return base.Reload();
         }
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -1175,7 +1175,7 @@ namespace NLog
         public LogFactory ReloadConfiguration()
         {
             InternalLogger.Info("Reloading configuration...");
-            Configuration = Configuration.Reload();
+            Configuration = Configuration.ReloadNewConfig();
 
 
             NotifyConfigurationReloaded();
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -562,7 +562,7 @@ namespace NLog
         /// <param name="timeout">Maximum time to allow for the flush. Any messages after that time will be discarded.</param>
         public void Flush(AsyncContinuation asyncContinuation, TimeSpan timeout)
         {
-            FlushInternal(timeout, asyncContinuation ?? (ex => { }));
+            FlushInternal(timeout, asyncContinuation ?? (ex => { }) );
         }
 
         private void FlushInternal(TimeSpan timeout, AsyncContinuation asyncContinuation)
@@ -1174,7 +1174,7 @@ namespace NLog
                 return this;    // Skip optional loading of default config, when config is already loaded
             }
 
-            var config = _configLoader.Load(this, configFile);
+            var config = CreateConfig(configFile);
             if (config == null)
             {
                 if (!optional)
@@ -1194,6 +1194,12 @@ namespace NLog
 
             Configuration = config;
             return this;
+        }
+
+        internal LoggingConfiguration CreateConfig(string configFile)
+        {
+            var config = _configLoader.Load(this, configFile);
+            return config;
         }
 
         private string CreateFileNotFoundMessage(string configFile)
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -84,12 +84,10 @@ namespace NLog
         /// </summary>
         public event EventHandler<LoggingConfigurationChangedEventArgs> ConfigurationChanged;
 
-#if !SILVERLIGHT && !__IOS__ && !__ANDROID__ && !NETSTANDARD1_3
         /// <summary>
         /// Occurs when logging <see cref="Configuration" /> gets reloaded.
         /// </summary>
         public event EventHandler<LoggingConfigurationReloadedEventArgs> ConfigurationReloaded;
-#endif
 
         private static event EventHandler<EventArgs> LoggerShutdown;
 
@@ -562,7 +560,7 @@ namespace NLog
         /// <param name="timeout">Maximum time to allow for the flush. Any messages after that time will be discarded.</param>
         public void Flush(AsyncContinuation asyncContinuation, TimeSpan timeout)
         {
-            FlushInternal(timeout, asyncContinuation ?? (ex => { }) );
+            FlushInternal(timeout, asyncContinuation ?? (ex => { }));
         }
 
         private void FlushInternal(TimeSpan timeout, AsyncContinuation asyncContinuation)
@@ -794,7 +792,6 @@ namespace NLog
             ConfigurationChanged?.Invoke(this, e);
         }
 
-#if !SILVERLIGHT && !__IOS__ && !__ANDROID__ && !NETSTANDARD1_3
         /// <summary>
         /// Raises the event when the configuration is reloaded. 
         /// </summary>
@@ -804,11 +801,16 @@ namespace NLog
             ConfigurationReloaded?.Invoke(this, e);
         }
 
-        internal void NotifyConfigurationReloaded(LoggingConfigurationReloadedEventArgs eventArgs)
+        /// <summary>
+        /// Notify success or failed, depending of <paramref name="exception"/>
+        /// </summary>
+        /// <param name="exception">success iif this parameter null</param>
+        internal void NotifyConfigurationReloaded(Exception exception = null)
         {
-            OnConfigurationReloaded(eventArgs);
+            var success = exception == null;
+
+            OnConfigurationReloaded(new LoggingConfigurationReloadedEventArgs(success, exception));
         }
-#endif
 
         private bool GetTargetsByLevelForLogger(string name, List<LoggingRule> loggingRules, TargetWithFilterChain[] targetsByLevel, TargetWithFilterChain[] lastTargetsByLevel, bool[] suppressedLevels)
         {
@@ -1166,6 +1168,20 @@ namespace NLog
             return LoadConfiguration(configFile, optional: false);
         }
 
+        /// <summary>
+        /// Reload and apply the current config
+        /// </summary>
+        /// <returns>LogFactory instance for fluent interface</returns>
+        public LogFactory ReloadConfiguration()
+        {
+            InternalLogger.Info("Reloading configuration...");
+            Configuration = Configuration.Reload();
+
+
+            NotifyConfigurationReloaded();
+            return this;
+        }
+
         internal LogFactory LoadConfiguration(string configFile, bool optional)
         {
             var actualConfigFile = string.IsNullOrEmpty(configFile) ? "NLog.config" : configFile;
@@ -1423,5 +1439,7 @@ namespace NLog
                 InternalLogger.Error(ex, "Logger failed to shut down properly.");
             }
         }
+
+
     }
 }
--- a/src/NLog/LogManager.cs
+++ b/src/NLog/LogManager.cs
@@ -173,6 +173,15 @@ namespace NLog
         }
 
         /// <summary>
+        /// Reload and apply the current config
+        /// </summary>
+        /// <returns>LogFactory instance for fluent interface</returns>
+        public static LogFactory ReloadConfiguration()
+        {
+            return factory.ReloadConfiguration();
+        }
+
+        /// <summary>
         /// Gets or sets the global log threshold. Log events below this threshold are not logged.
         /// </summary>
         public static LogLevel GlobalThreshold
--- a/src/NLog/SetupBuilderExtensions.cs
+++ b/src/NLog/SetupBuilderExtensions.cs
@@ -114,15 +114,13 @@ namespace NLog
             {
                 setupBuilder.LogFactory.Configuration = newConfig;
             }
-#if !NETSTANDARD1_3
             if (applyOnReload)
             {
-                setupBuilder.LogFactory.ConfigurationChanged += (sender, e) =>
+                setupBuilder.LogFactory.ConfigurationReloaded += (sender, e) =>
                 {
                     setupBuilder.LoadConfiguration(false, configBuilder);
                 };
             }
-#endif
 
             return setupBuilder;
         }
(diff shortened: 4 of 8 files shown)


src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -272,7 +272,10 @@ public override IEnumerable<string> FileNamesToWatch
         public override LoggingConfiguration Reload()
         {
             if (!string.IsNullOrEmpty(_originalFileName))
-                return new XmlLoggingConfiguration(_originalFileName, LogFactory);
+            {
+                LogFactory.LoadConfiguration(_originalFileName);
```
304NotModified, line 276: FYI, needed this otherwise IAppEnvironment will be ignored

src/NLog/SetupBuilderExtensions.cs
```diff
@@ -107,6 +115,16 @@ public static ISetupBuilder LoadConfiguration(this ISetupBuilder setupBuilder, A
                 setupBuilder.LogFactory.Configuration = newConfig;
             }
 
+            if (applyOnReload)
+            {
+#if !NETSTANDARD1_3
+                setupBuilder.LogFactory.ConfigurationChanged += (sender, e) =>
```
304NotModified, line 121: I'm always a bit confused by Changed and Reloaded :'(

Seems Reloaded only works for file reloads?

Not sure if I need a test for that case (file reload) then

src/NLog/Config/XmlLoggingConfiguration.cs
snakefoot: Breaking change as the Reload now assigns the Configuration. What happens if the LogFactory.LoadConfiguration fails?

src/NLog/SetupBuilderExtensions.cs
snakefoot: How do you ensure that you don't have multiple callbacks? Think users will expect that previous delegates will be reset, if calling LoadConfiguration twice.

src/NLog/Config/XmlLoggingConfiguration.cs
304NotModified: > Breaking change as the Reload now assigns the Configuration

Well I always was under the impression that was a bug. Or do you think it was useful to have that separated? 

> No longer possible to call explicit Reload() ,and then adjust the returned config and then assign the config.

Why is that not possible now?

snakefoot: Because `LogFactory.LoadConfiguration` will assign the configuration, so any adjustments are too late.

snakefoot: But I can certainly agree that the config-reload is full of pitfalls. But not sure if the fix should be implemented on XmlLoggingConfiguration directly.

src/NLog/SetupBuilderExtensions.cs
304NotModified: Could you give an example?
```
  logFactory.Setup().
                LoadConfigurationFromFile().
                LoadConfiguration(true, b =>
                {
                    b.Configuration.AddRuleForAllLevels("target1");
                })
                .LoadConfiguration(true, b =>
                {
                    b.Configuration.AddRuleForAllLevels("target2");
                });
```

snakefoot: I was just thinking it might confuse if doing Setup()-operation more than once. But lets see what happens. Btw. can you make sure that the callback is only called when a new config is loaded? (Avoid callback with null on shutdown / configuration assigned to null).

src/NLog/Config/XmlLoggingConfiguration.cs
304NotModified: > But maybe the improvement can be implemented without changing XmlLoggingConfiguration.Reload() ?

Good idea, will look into it. Will be tomorrow I guess

src/NLog/SetupBuilderExtensions.cs
304NotModified: > might confuse if doing Setup()-operation more than once

No this is great! Always wanted this! OK, maybe it needs a reset, but you could do that in LoadConfiguration

src/NLog/Config/XmlLoggingConfiguration.cs
304NotModified: Fixed, please double check 

PS I would appreciate an Approve  "flag"  :)

src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -272,7 +272,7 @@ public override IEnumerable<string> FileNamesToWatch
         public override LoggingConfiguration Reload()
         {
             if (!string.IsNullOrEmpty(_originalFileName))
-                return new XmlLoggingConfiguration(_originalFileName, LogFactory);
+                return LogFactory.CreateConfig(_originalFileName); //TODO NLog 5: should call LoadConfiguration which set LogFactory.Configuration
```
snakefoot, line 275: Still curious why you need to modify this method? Why not just have a method called `LogFactory.ReloadConfiguration()` instead of the config assigning itself?

src/NLog/SetupBuilderExtensions.cs
```diff
@@ -106,6 +114,15 @@ public static ISetupBuilder LoadConfiguration(this ISetupBuilder setupBuilder, A
             {
                 setupBuilder.LogFactory.Configuration = newConfig;
             }
+#if !NETSTANDARD1_3
+            if (applyOnReload)
+            {
+                setupBuilder.LogFactory.ConfigurationChanged += (sender, e) =>
```
snakefoot, line 120: ConfigurationChanged will give notifications on the following events:
- AutoReload - FileWatcher-logic
- ManualLoad - Assigning `LogFactory.Configuration = newConfig;`
- ManualUnload - Assigning `LogFactory.Configuration = null;`
- Shutdown

I think you should not perform the callback when there is no active LoggingConfiguration available. And maybe consider if the callback should also happen on `ManualLoad`.

tests/NLog.UnitTests/Config/LogFactorySetupTests.cs
```diff
@@ -568,5 +569,103 @@ public void SetupBuilderLoadConfigurationFromXmlPatchTest()
             Assert.False(logFactory.AutoShutdown);
             Assert.Single(logFactory.Configuration.Variables);
         }
+
```
snakefoot, line 572: Maybe include a test-method where you call LogFactory.Close() ? (Similar to Shutdown).

snakefoot:
Could also be interesting to verify that this works together with the autoreload-feature in MEL NLogLoggingConfiguration:

Notice how I abuse ConfigurationChanged to unhook from filewatch-notications from MEL-config-system:

https://github.com/NLog/NLog.Extensions.Logging/blob/master/src/NLog.Extensions.Logging/Config/NLogLoggingConfiguration.cs

src/NLog/Config/XmlLoggingConfiguration.cs
304NotModified: because `new XmlLoggingConfiguration(_originalFileName, LogFactory);` doens't use the _configLoader, and thus you well get this.

```c#
   var appEnv = new Mocks.AppEnvironmentMock( ...)
  var configLoader = new LoggingConfigurationFileLoader(appEnv);
  var logFactory = new LogFactory(configLoader);

  // Act
  logFactory.Setup().
  LoadConfigurationFromFile().
  // config loaded from configLoader 

  // crash:  System.IO.FileNotFoundException - Could not find file 'D:\nlog\NLog\tests\NLog.UnitTests\bin\Debug\net452\NLog.config'.
  logFactory.Configuration = logFactory.Configuration.Reload()
```

src/NLog/SetupBuilderExtensions.cs
304NotModified: I'm also doubting with ManualLoad , but that one is hard to prevent now isn't?

src/NLog/Config/XmlLoggingConfiguration.cs
snakefoot: Would it not be better to change `XmlLoggingConfiguration` to use the LogFactory-AppEnvironment for loading the XmlConfig-file ? (Improve mocking all around, instead of just Reload)

src/NLog/SetupBuilderExtensions.cs
snakefoot: > with ManualLoad , but that one is hard to prevent now isn't?

I guess for XmlLoggingConfiguration then one could check if config was assigned using LogFactory.Configuration-property-setter (And signal that through the ConfigurationChanged-EventArgs). But for NLogLoggingConfiguration then it is not easy, since it has to use the official setter-gateway.

So yes hard to distinguish AutoReload and ManulLoad.

src/NLog/Config/XmlLoggingConfiguration.cs
304NotModified: Good point

src/NLog/SetupBuilderExtensions.cs
304NotModified: Good to know is that KeepVariablesOnReload is also working for manual load.

I have some things to fix here, but we should have a clear direction on this. If manual load is a no-go, I can't fix this (or not sure how)

src/NLog/Config/XmlLoggingConfiguration.cs
snakefoot: Have created #3941 that improves the mocking of `XmlReader.Create` for `XmlLoggingConfiguration`.

src/NLog/Config/XmlLoggingConfiguration.cs
snakefoot: Have created #3941 that improves the mocking of `XmlReader.Create` for `XmlLoggingConfiguration`.

304NotModified: > Why not in the future just have a method called LogFactory.ReloadConfiguration() instead of the config assigning itself?

I think also this is better! I think using the setter was a mistake, also I really need to use the reload event, not the changed. So fixed that

304NotModified:
@snakefoot please your opinion on this approach.

I still have to do:

- [ ] verify that this works together with the autoreload-feature in MEL
- [ ] change LoggingConfiguration.Reload()? Dunno how, revert doesn't work yet.
- [ ]  LogFactory.Close() / LogFactory.Dispose() ? (Similar to Shutdown) - this isn't needed anymore?

src/NLog/Config/XmlLoggingConfiguration.cs
```diff
@@ -272,7 +272,7 @@ public override IEnumerable<string> FileNamesToWatch
         public override LoggingConfiguration Reload()
         {
             if (!string.IsNullOrEmpty(_originalFileName))
-                return new XmlLoggingConfiguration(_originalFileName, LogFactory);
+                return LogFactory.CreateConfig(_originalFileName);
```
snakefoot, line 275: Still curious why this change is needed?

src/NLog/LogFactory.cs
```diff
@@ -1166,6 +1168,20 @@ public LogFactory LoadConfiguration(string configFile)
...
         }
 
+        /// <summary>
+        /// Reload and apply the current config
+        /// </summary>
+        /// <returns>LogFactory instance for fluent interface</returns>
+        public LogFactory ReloadConfiguration()
+        {
+            InternalLogger.Info("Reloading configuration...");
+            Configuration = Configuration.ReloadNewConfig();
```
snakefoot, line 1178: Could this feature be made in a separate PullRequest?

src/NLog/LogManager.cs
```diff
@@ -172,6 +172,15 @@ public static LogFactory LoadConfiguration(string configFile)
             return factory;
         }
 
+        /// <summary>
+        /// Reload and apply the current config
+        /// </summary>
+        /// <returns>LogFactory instance for fluent interface</returns>
+        public static LogFactory ReloadConfiguration()
```
snakefoot, line 179: Could this feature be made in a separate PullRequest?

src/NLog/Config/XmlLoggingConfiguration.cs
304NotModified: Because the test still fails, reload will search on the disk and fails

src/NLog/LogFactory.cs
304NotModified: Will do

src/NLog/Config/XmlLoggingConfiguration.cs
snakefoot: Even when you merge/rebase with master, and take in #3941 ?

304NotModified: yes I did a rebase.

304NotModified: See https://github.com/NLog/NLog/compare/logFactory.ReloadConfiguration...master

snakefoot: Really strange that it tries to access the disk when it can only access the disk using the mocked AppEnvironment. Are you mocking AppEnvironment ?

304NotModified: I think the mocking is correct.

See branch loadconfiguration-applyOnReload_noLogFactory.CreateConfig

image

304NotModified: I see the issue, LogFactory needs IAppEnvironment, while it's already in ILoggingConfigurationLoader

304NotModified: Fixed

src/NLog/SetupBuilderExtensions.cs
```diff
@@ -91,6 +91,14 @@ public static ISetupBuilder SetupSerialization(this ISetupBuilder setupBuilder,
...
         /// </summary>
         public static ISetupBuilder LoadConfiguration(this ISetupBuilder setupBuilder, Action<ISetupLoadConfigurationBuilder> configBuilder)
+        {
+            return LoadConfiguration(setupBuilder, true, configBuilder);
+        }
+
+        /// <summary>
+        /// Loads NLog config created by the method <paramref name="configBuilder"/>
+        /// </summary>
+        public static ISetupBuilder LoadConfiguration(this ISetupBuilder setupBuilder, bool applyOnReload, Action<ISetupLoadConfigurationBuilder> configBuilder)
```
304NotModified, line 101: I'm doubting if adding a bool here is a good idea. Maybe it should be an option object? Otherwise it's hard to extend. 

What do you think @snakefoot ?

snakefoot: When creating the LoadConfiguration-methods here https://github.com/NLog/NLog.Web/pull/540, then I wanted them to look like `AddJsonFile`. https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.jsonconfigurationextensions.addjsonfile But the lambda makes it a little different. If you believe the number of options for the same method will be more than 3 then I think an options-object makes sense.

src/NLog/LogFactory.cs
```diff
@@ -1196,6 +1212,12 @@ internal LogFactory LoadConfiguration(string configFile, bool optional)
             return this;
         }
 
+        internal LoggingConfiguration CreateConfig(string configFile)
```
snakefoot, line 1215: Method extraction is no loading needed, and I like calling `_configLoader.Load` better than the name `CreateConfig`.

src/NLog/SetupBuilderExtensions.cs
```diff
@@ -91,6 +91,14 @@ public static ISetupBuilder SetupSerialization(this ISetupBuilder setupBuilder,
         /// Loads NLog config created by the method <paramref name="configBuilder"/>
         /// </summary>
         public static ISetupBuilder LoadConfiguration(this ISetupBuilder setupBuilder, Action<ISetupLoadConfigurationBuilder> configBuilder)
+        {
+            return LoadConfiguration(setupBuilder, true, configBuilder);
```
snakefoot, line 95: Don't think it should be `true` by default.

src/NLog/SetupBuilderExtensions.cs
```diff
@@ -106,6 +114,13 @@ public static ISetupBuilder LoadConfiguration(this ISetupBuilder setupBuilder, A
             {
                 setupBuilder.LogFactory.Configuration = newConfig;
             }
+            if (applyOnReload)
+            {
+                setupBuilder.LogFactory.ConfigurationReloaded += (sender, e) =>
+                {
+                    setupBuilder.LoadConfiguration(false, configBuilder);
```
snakefoot, line 121: Should only perform the call when `e.Succeeded == true`.

snakefoot: When using `ConfigurationReloaded` instead of `ConfigurationChanged`, then it will not work together with MEL-NLogLoggingConfiguration. As it doesn't make use of internal NLog file-watchers https://github.com/NLog/NLog.Extensions.Logging/blob/4bd3bcb9aceca3da3b76dee6cbe711f0cd0fb34b/src/NLog.Extensions.Logging/Config/NLogLoggingConfiguration.cs#L111

Think you should change from `ConfigurationReloaded` to `ConfigurationChanged`. Maybe mark `ConfigurationReloaded` as obsolete in NLog 5.0?

snakefoot: @304NotModified Created #3954 with example of how to use `ConfigurationChanged`

304NotModified:
inactive so close for now

PR #3930, opened 2020-04-28
Report the searched paths when the config file can't be found

304NotModified:
Always wanted this ❤️ 

image

### DIFF
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -1157,7 +1157,6 @@ namespace NLog
                 return this;    // Skip optional loading of default config, when config is already loaded
             }
 
-
             var config = _configLoader.Load(this, configFile);
             if (config == null)
             {
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -1165,7 +1165,7 @@ namespace NLog
 #if SILVERLIGHT
                     throw new System.IO.FileNotFoundException($"Failed to load NLog LoggingConfiguration from file {actualConfigFile}");
 #else
-                    var message = CreateFileNotFoundMessage(configFile);
+                    var message = CreateFileNotFoundMessage(_configLoader.GetDefaultCandidateConfigFilePaths(configFile));
                     throw new System.IO.FileNotFoundException(message, actualConfigFile);
 #endif
                 }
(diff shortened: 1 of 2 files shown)


snakefoot:
Great idea. Don't know how exception-safe the method `GetDefaultCandidateConfigFilePaths` is?

304NotModified:
Thanks!



> Don't know how exception-safe the method `GetDefaultCandidateConfigFilePaths` is?

Not sure due to the _appEnvironment stuff. But loading the config was already not (also documented) as non threadsafe? Good question by the way

304NotModified:
PS funny locations, not sure if this is expected


image

snakefoot:
> But loading the config was already not (also documented) as non **threadsafe**?

I was thinking of exception-safe.

> PS funny locations, not sure if this is expected

Looks as expected to me (Checking both entry-assembly-location and process-location).

304NotModified:
> I was thinking of exception-safe.

oops :)

304NotModified:
Updated it to make it exception-safe :)

src/NLog/LogFactory.cs
```diff
@@ -1165,7 +1165,7 @@ internal LogFactory LoadConfiguration(string configFile, bool optional)
 #if SILVERLIGHT
                     throw new System.IO.FileNotFoundException($"Failed to load NLog LoggingConfiguration from file {actualConfigFile}");
 #else
-                    var message = CreateFileNotFoundMessage(configFile);
+                    var message = CreateFileNotFoundMessage(_configLoader.GetDefaultCandidateConfigFilePaths(configFile));
```
snakefoot, line 1168: Not sure I understand. Why is `_configLoader.GetDefaultCandidateConfigFilePaths` no inside the try-catch-block?

304NotModified: Well I prefer static methods, and `_configLoader.GetDefaultCandidateConfigFilePaths(configFile)` is lazy. But maybe it's a bit too much coincidence

304NotModified: I was also doubting about this, so 2 hints = fix. :)

pushed!

304NotModified:
@snakefoot 

is maybe the text "The following locations were searched" better? (like MVC)

snakefoot:
Think `Searched the following locations: {0}` is great. Like when the parameter-values comes right after parameter-name :)

But I think both sentences are very easy to understand.

Issue #3555, opened 2019-08-20
Remove IDisposable from LayoutRenderer


snakefoot:
NLog has its own initialization system provided by `ISupportsInitialize`.

Instead of giving code-analysis warning to everyone about "Types that own disposable fields should be disposable".

Intended for NLog 5.0

304NotModified:
Sounds good!

PR #3932, opened 2020-04-29
XmlLayout - Avoid warning about IDisposable until we can remove it

snakefoot:
See also #3555


304NotModified:
Maybe I have to think about it, but for now I don't fully see why this works...

snakefoot:
> Maybe I have to think about it, but for now I don't fully see why this works...

It changes the code to match `JsonAttribute`, that is not registered as a bug. Work-around until IDisposable is removed from LayoutRenderer.

304NotModified:
OK thanks!

PR #3934, opened 2020-04-29
NLogViewerTarget - Fixed Log4JXmlEventLayout

snakefoot:


304NotModified:
nice solution, bit work around but sure why not :)

snakefoot:
Still a code-smell with an empty setter-method, but since it is intended, then I don't see it as a bug.

304NotModified:
yes, I will fix this in SQ, but the code is better as it looks now intended 👍

Issue #3939, opened 2020-05-02
Layout with {processid} will crash with Mono runtime on Android


jifang:
Hi! I use NLog with Unity developed android app using mono backend. It crashes if Layout contains {processid}

**NLog version**: (e.g. 4.4.13)
4.6.8

**Platform**: 
.Net 4.5 

**Current NLog config** (xml or C#, if relevant)

```C#
    var logfile = new NLog.Targets.FileTarget("logfile")
    {
        Layout = "${processid}",
    };
```

- What is the current result?
App crashes

- What is the expected result?
Don't crash.

- Did you checked the [Internal log](https://github.com/NLog/NLog/wiki/Internal-Logging)?
Yes. See logs below.
```
2020-05-01 22:33:46.8224 Debug ScanAssembly('NLog, Version=4.0.0.0, Culture=neutral, PublicKeyToken=5120e14c03d0593c')
2020-05-01 22:33:47.5014 Warn Ignoring assembly location because directory doesn't exists: '/data/app/redacted.app.package.name-TBeNq-QaKH-62e3uWEsJqA==/base.apk/assets/bin/Data/Managed' (NLog, Version=4.0.0.0, Culture=neutral, PublicKeyToken=5120e14c03d0593c)
2020-05-01 22:33:47.5763 Warn AppDomain.BaseDirectory Failed Exception: System.ArgumentException: The specified path is not of a legal form (empty).
  at System.IO.Path.InsecureGetFullPath (System.String path) [0x00025] in <e1319b7195c343e79b385cd3aa43f5dc>:0 
  at System.IO.Path.GetFullPath (System.String path) [0x00000] in <e1319b7195c343e79b385cd3aa43f5dc>:0 
  at System.AppDomainSetup.GetAppBase (System.String appBase) [0x00041] in <e1319b7195c343e79b385cd3aa43f5dc>:0 
  at System.AppDomainSetup.get_ApplicationBase () [0x00000] in <e1319b7195c343e79b385cd3aa43f5dc>:0 
  at System.AppDomain.get_BaseDirectory () [0x00006] in <e1319b7195c343e79b385cd3aa43f5dc>:0 
  at (wrapper remoting-invoke-with-check) System.AppDomain.get_BaseDirectory()
  at NLog.Internal.Fakeables.AppDomainWrapper.LookupBaseDirectory (System.AppDomain appDomain) [0x00000] in <05811f8762774917a895559f828b318a>:0 
[...]

PR #3921, opened 2020-04-23
BaseDirLayoutRenderer - Improve FixTempDir on Linux when TMPDIR is not set

snakefoot:
See GetTempPath on Linux (Uses `/tmp/` as only fallback):

https://github.com/dotnet/coreclr/blob/6ba74dc2a7194f8d6c86c3aeab572a074ef645c8/src/mscorlib/shared/System/IO/Path.Unix.cs#L153

See single file publish on Linux (Uses `/var/tmp/` as first fallback):

https://github.com/dotnet/core-setup/blob/074c3bece80ca300e2f9d053b18748b6537d6f67/src/corehost/common/pal.unix.cpp#L306

Everything is great when TMPDIR is configured, but if not then fallback-paths doesn't match.

Resolves #3899

PR #3941, opened 2020-05-02
Removed ProcessIDHelper and extended AppEnvironment with current process details

snakefoot:
Improves mocking. And allows rendering of CurrentProcessId without crashing on Unity. See also #3939

ProcessIDHelper have evolved over time. Initially lookup of process-id and process-executable was not cached. Instead lookup was done for every logevent. This had a performance penalty but it was improved with the optimized Win32-instance that didn't allocate a Process-object for every logevent. Later caching was added, so the lookup was only done at startup. Because of the caching then the optimized Win32 is not that important. And now the ProcessIDHelper is being removed.


src/NLog/Config/LoggingConfigurationParser.cs
```diff
@@ -247,7 +247,7 @@ private static string ExpandFilePathVariables(string internalLogFile)
                     internalLogFile = internalLogFile.Replace(tempDirToken, System.IO.Path.GetTempPath() + System.IO.Path.DirectorySeparatorChar.ToString());
 #if !NETSTANDARD1_3
                 if (ContainsSubStringIgnoreCase(internalLogFile, "${processdir}", out string processDirToken))
-                    internalLogFile = internalLogFile.Replace(processDirToken, System.IO.Path.GetDirectoryName(ProcessIDHelper.Instance.CurrentProcessFilePath) + System.IO.Path.DirectorySeparatorChar.ToString());
+                    internalLogFile = internalLogFile.Replace(processDirToken, System.IO.Path.GetDirectoryName(LogFactory.DefaultAppEnvironment.CurrentProcessFilePath) + System.IO.Path.DirectorySeparatorChar.ToString());
```
304NotModified, line 250: Should this be `CurrentAppEnvironment`? Or don't we have a LogFactory instance her?

src/NLog/LayoutRenderers/Log4JXmlEventLayoutRenderer.cs
```diff
@@ -94,7 +94,7 @@ public Log4JXmlEventLayoutRenderer(IAppDomain appDomain)
                 CultureInfo.InvariantCulture,
                 "{0}({1})",
                 appDomain.FriendlyName,
-                ProcessIDHelper.Instance.CurrentProcessID);
+                LogFactory.DefaultAppEnvironment.CurrentProcessId);
```
304NotModified, line 97: Would also expect CurrentAppEnvironment here

src/NLog/LogFactory.cs
```diff
@@ -142,25 +145,39 @@ internal LogFactory(ILoggingConfigurationLoader configLoader)
...
-            get
-            {
-                return currentAppDomain ??
-#if NETSTANDARD1_0
-                    (currentAppDomain = new FakeAppDomain());
-#else
-                    (currentAppDomain = new AppDomainWrapper(AppDomain.CurrentDomain));
-#endif
-            }
+            get => currentAppDomain ?? DefaultAppEnvironment.AppDomain;
```
304NotModified, line 148: Idem, why not CurrentAppEnvironment?

src/NLog/Config/LoggingConfigurationParser.cs
snakefoot: Change `ExpandFilePathVariables` from static-method to member-method. And adjusted.

src/NLog/LogFactory.cs
snakefoot: Because CurrentAppDomain is a static-member (Should be called DefaultAppDomain as it is global). And cannot reference a member-variable

src/NLog/LayoutRenderers/Log4JXmlEventLayoutRenderer.cs
snakefoot: Not possible to inject AppEnvironment into constructor without it becoming a breaking API change. Planning to make the constructor internal in NLog 5, and change parameter from AppDomain to AppEnvironment. But not possible with NLog 4.7.

snakefoot:
Resolved merge conflicts triggered by merge of #3921

PR #3942, opened 2020-05-02
Lazily get the process file path.

jifang:
Process.GetCurrentProcess().MainModule may crash on some certain
platforms, such as Android with mono runtime.

It allows the client can still use the ProcessID layout renderer.


jifang:
https://github.com/NLog/NLog/pull/3941

PR #3950, opened 2020-05-06
BufferingTargetWrapperTests - Make unit test more stable

snakefoot:
Waiting for events instead of short thread sleeps. Added event wait timeouts where missing (Fail instead of waiting forever)


304NotModified:
Thanks! Could we pass an message to assert.true? Maybe that's nice if it fails.

snakefoot:
> Could we pass an message to assert.true? Maybe that's nice if it fails.

Assert.True now with message

304NotModified:
Thanks!

PR #3952, opened 2020-05-10
Added logFactory.ReloadConfiguration and LogManager.ReloadConfiguration

### LINKED ISSUE #3938: logmanager/logFactory.Configuration.Reload() should update config and not only return it
Currently logFactory.Configuration.Reload() will only return a new instance and thus this is needed:

Same for LogManager

304NotModified:
fixes #3938

### DIFF
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -84,12 +84,10 @@ namespace NLog
         /// </summary>
         public event EventHandler<LoggingConfigurationChangedEventArgs> ConfigurationChanged;
 
-#if !SILVERLIGHT && !__IOS__ && !__ANDROID__ && !NETSTANDARD1_3
         /// <summary>
         /// Occurs when logging <see cref="Configuration" /> gets reloaded.
         /// </summary>
         public event EventHandler<LoggingConfigurationReloadedEventArgs> ConfigurationReloaded;
-#endif
 
         private static event EventHandler<EventArgs> LoggerShutdown;
 
@@ -794,7 +792,6 @@ namespace NLog
             ConfigurationChanged?.Invoke(this, e);
         }
 
-#if !SILVERLIGHT && !__IOS__ && !__ANDROID__ && !NETSTANDARD1_3
         /// <summary>
         /// Raises the event when the configuration is reloaded. 
         /// </summary>
@@ -804,11 +801,16 @@ namespace NLog
             ConfigurationReloaded?.Invoke(this, e);
         }
 
-        internal void NotifyConfigurationReloaded(LoggingConfigurationReloadedEventArgs eventArgs)
+        /// <summary>
+        /// Notify success or failed, depending of <paramref name="exception"/>
+        /// </summary>
+        /// <param name="exception">success iif this parameter null</param>
+        internal void NotifyConfigurationReloaded(Exception exception = null)
         {
-            OnConfigurationReloaded(eventArgs);
+            var success = exception == null;
+
+            OnConfigurationReloaded(new LoggingConfigurationReloadedEventArgs(success, exception));
         }
-#endif
 
         private bool GetTargetsByLevelForLogger(string name, List<LoggingRule> loggingRules, TargetWithFilterChain[] targetsByLevel, TargetWithFilterChain[] lastTargetsByLevel, bool[] suppressedLevels)
         {
@@ -1164,6 +1166,19 @@ namespace NLog
         {
             // TODO Remove explicit File-loading logic from LogFactory (Should handle environment without files)
             return LoadConfiguration(configFile, optional: false);
+        }
+
+        /// <summary>
+        /// Reload the current logging configuration and apply it.
+        /// </summary>
+        /// <returns>LogFactory instance for fluent interface</returns>
+        public LogFactory ReloadConfiguration()
+        {
+            InternalLogger.Info("Reloading configuration...");
+            Configuration = Configuration?.ReloadNewConfig();
+
+            NotifyConfigurationReloaded();
+            return this;
         }
 
         internal LogFactory LoadConfiguration(string configFile, bool optional)
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -1175,31 +1175,42 @@ namespace NLog
         public LogFactory ReloadConfiguration()
         {
             InternalLogger.Info("Reloading configuration...");
-
-            if (_config != null)
+            try
             {
-                var errorRaised = false;
-                try
+                if (_config == null)
                 {
-                    Configuration = Configuration.ReloadNewConfig();
-                }
-                catch (Exception e)
-                {
-                    if (e.MustBeRethrownImmediately() || e.MustBeRethrown())
+                    // Not loaded yet, so first load is a reload
+                    _config = Configuration;
+                    if (_config != null)
                     {
-                        throw;
+                        NotifyConfigurationReloaded();
                     }
-
-                    errorRaised = true;
-                    NotifyConfigurationReloaded(e);
                 }
-                
-                if (!errorRaised)
+                else
                 {
-                    NotifyConfigurationReloaded();
+                    var newConfig = _config.ReloadNewConfig();
+                    if (newConfig != null)
+                    {
+                        var oldConfig = _config;
+                        Configuration = newConfig;
+                        if (!ReferenceEquals(oldConfig, newConfig))
+                        {
+                            // Avoid applying reload-adjust-methods twice to the same config
+                            NotifyConfigurationReloaded();
+                        }
+                    }
+                }
+            }
+            catch (Exception e)
+            {
+                if (e.MustBeRethrownImmediately() || e.MustBeRethrown())
+                {
+                    throw;
                 }
 
+                NotifyConfigurationReloaded(e);
             }
+
 
             return this;
         }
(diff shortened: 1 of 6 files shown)


src/NLog/LogFactory.cs
```diff
@@ -804,11 +801,16 @@ protected virtual void OnConfigurationReloaded(LoggingConfigurationReloadedEvent
             ConfigurationReloaded?.Invoke(this, e);
         }
 
-        internal void NotifyConfigurationReloaded(LoggingConfigurationReloadedEventArgs eventArgs)
+        /// <summary>
+        /// Notify success or failed, depending of <paramref name="exception"/>
+        /// </summary>
+        /// <param name="exception">success iif this parameter null</param>
+        internal void NotifyConfigurationReloaded(Exception exception = null)
```
304NotModified, line 808: FYI: This is a small refactor, as both parameters where highly related, I made it to one.

src/NLog/LogFactory.cs
```diff
@@ -1166,6 +1168,19 @@ public LogFactory LoadConfiguration(string configFile)
...
         }
 
+        /// <summary>
+        /// Reload and apply the current config
+        /// </summary>
+        /// <returns>LogFactory instance for fluent interface</returns>
+        public LogFactory ReloadConfiguration()
+        {
+            InternalLogger.Info("Reloading configuration...");
+            Configuration = Configuration.ReloadNewConfig();
```
304NotModified, line 1178: Need a null check I guess

304NotModified: fixed

src/NLog/LogFactory.cs
```diff
@@ -1166,6 +1168,19 @@ public LogFactory LoadConfiguration(string configFile)
...
+        /// <summary>
+        /// Reload the current logging configuration and apply it.
+        /// </summary>
+        /// <returns>LogFactory instance for fluent interface</returns>
+        public LogFactory ReloadConfiguration()
+        {
+            InternalLogger.Info("Reloading configuration...");
+            Configuration = Configuration?.ReloadNewConfig();
+
+            NotifyConfigurationReloaded();
```
snakefoot, line 1180: Why not try-catch and call `NotifyConfigurationReloaded` with the Exception ?

snakefoot: Should the Reload not fail when Configuration == null ? (Call NotifyConfigurationReloaded with exception)

snakefoot: Instead of calling `Configuration?.ReloadNewConfig()` then check if `_config == null`. This means one should just call Configuration-property and it will automatically load the available NLog.config.

304NotModified: Partly done,

> Should the Reload not fail when Configuration == null ? (Call NotifyConfigurationReloaded with exception)

doubting about this.

snakefoot: Then at least not call `NotifyConfigurationReloaded` when no reload is performed

src/NLog/LogFactory.cs
```diff
@@ -1175,9 +1175,32 @@ public LogFactory LoadConfiguration(string configFile)
...
             InternalLogger.Info("Reloading configuration...");
-            Configuration = Configuration?.ReloadNewConfig();
+            var errorRaised = false;
 
-            NotifyConfigurationReloaded();
+            if (_config == null)
+            {
+                try
+                {
+                    Configuration = Configuration.ReloadNewConfig();
```
snakefoot, line 1184: I was actually thinking the other way. When _config == null then you just need to load Configuration. When _config != null then you should call `_config.ReloadNewConfig()`.

src/NLog/LogFactory.cs
```diff
@@ -1166,6 +1168,42 @@ public LogFactory LoadConfiguration(string configFile)
...
+        /// <summary>
+        /// Reload the current logging configuration and apply it.
+        /// </summary>
+        /// <returns>LogFactory instance for fluent interface</returns>
+        public LogFactory ReloadConfiguration()
+        {
+            InternalLogger.Info("Reloading configuration...");
+            var errorRaised = false;
+
+            if (_config != null)
```
snakefoot, line 1180: When `_config == null` then you should should just reference the `Configuration`-property to perform default load of NLog.config.

304NotModified: Too confusing and tricky now. Don't really get it why it's needs to be so complicated...

304NotModified: (well yes, I understand, the getters and setters are too complicated)

304NotModified: Anyway, dunno what I should do now

snakefoot: Something like this:

```c#
public LogFactory ReloadConfiguration()
{
  if (_config != null)
  {
       Configuration = _config.ReloadNewConfig();
       NotifyConfigurationReloaded();
  }
  else
  {
       _config = Configuration;
       if (_config != null)
       {
             NotifyConfigurationReloaded();
       }
  }
  return this;
}
```

src/NLog/LogFactory.cs
304NotModified: I'm doubting about this. We invoked a reload, but nothing has changed. If we have to listen for changes, then we have ConfigurationChanged event?

src/NLog/LogFactory.cs
304NotModified: Done but same as below, so resolve this one

src/NLog/LogFactory.cs
304NotModified: Done, I also added a try catch. 

Also I did a refactor of the property, as it's really confusing to me that it's doing so much. So made that clear with separate methods. Let me know if you like it. Otherwise I will just remove that commit.

304NotModified: I'm doubting, shouldn't we check _configLoaded instead of _config?

304NotModified:
Unit test fails, as how to handle (event, exception) is still a discussion when the config is null

src/NLog/LogFactory.cs
snakefoot: Think it is a good idea to extract logic `Configuration`-property into `ApplyNewConfig`. But I think it is a bad idea to change all places to call `ApplyNewConfig` directly. Instead all should just continue to assign the `Configuration`-property. (Only `Configuration`-property should call `ApplyNewConfig`)

snakefoot: `_configLoaded` is a special state variable for LogFactory to ensure it doesn't try to auto-load Configuration-property after Close/Shutdown.

snakefoot:
Please make a ReloadConfiguration that takes a LoggingConfiguration as parameter. Please make a new ConfigurationReloading-event that can be used in `applyOnReload`. Make sure the new event also has `IsNewConfiguration`.

304NotModified:
> Please make a new ConfigurationReloading-event

We can't have a new event? As it should also work when you reload things by changing the nlog.config (when autoreload is on)

snakefoot:
> We can't have a new event? As it should also work when you reload things by changing the nlog.config (when autoreload is on)

You should have the callback happening before initializing the targets. Not after. Which is why it should be `ConfigurationReloading` (BEFORE) and not `ConfigurationReloaded` (AFTER)

304NotModified:
I think you're right. Damn another event.

Funny, as this was always wrong then: https://github.com/NLog/NLog/wiki/Combine-XML-config-with-C%23-config

snakefoot:
Yes I know :). It only worked with "smart" targets like FileTarget that adapted to properties being changed after initialization. But very few target has that flexibility / complexity.

304NotModified:
Or does that work thanks to `    LogManager.ReconfigExistingLoggers();`? (I always find this call a bit strange)

304NotModified:
If we have a ConfigurationReloading, do we still need ConfigurationLoading? (or it is nice to have them both - but maybe it should be ConfigurationChanging then)

snakefoot:
`LogManager.ReconfigExistingLoggers()` will ensure that "new" targets are initialized, and new rules are applied in filtering.

But it doesn't help when making modifications to existing targets.

304NotModified:
> Please make a ReloadConfiguration that takes a LoggingConfiguration as parameter.

I get the idea, but I think it's a bit strange. Would it be more logical if we have `LoadConfigration(configuration, bool behaveAsReload)? (better name also OK)

snakefoot:
No you are misunderstanding me. This PR adds this method:

`LogFactory.ReloadConfiguration()`

I want also this method (That can be called by MEL-NLogLoggingConfiguration):

`LogFactory.ReloadConfiguration(LoggingConfiguration newConfig)`

I don't want new methods on the Setup()-API

304NotModified:
> want also this method (That can be called by MEL-NLogLoggingConfiguration):
> 
> `LogFactory.ReloadConfiguration(LoggingConfiguration newConfig)`

I don't understand this. A reload is always of the current config? It's now a replace-reload. What's na use case for that?

snakefoot:
How is MEL-NLogLoggingConfiguration going to signal that it is performing a reload?

304NotModified:
> How is MEL-NLogLoggingConfiguration going to signal that it is performing a reload?

By doing `LoadConfigration(configuration, bool behaveAsReload)`. 

I don't think we going agree with that. I could live with `LogFactory.ReloadConfiguration(newConfig);`

snakefoot:
> I could live with `LogFactory.ReloadConfiguration(newConfig);`

Excellent

304NotModified:
inactive so close for now

PR #3953, opened 2020-05-11
Added internal ExceptionMustBeRethrown helper to Target

304NotModified:
- easier to use
- better context capturing
- Better error config scoping (LogFactory instead of LogManager)

Discussion: should we make this internal? (as private protected is not yet possible)

if liked, could also make one for LayoutRenderer and Layout

### DIFF
--- a/src/NLog/Internal/ExceptionHelper.cs
+++ b/src/NLog/Internal/ExceptionHelper.cs
@@ -78,9 +78,9 @@ namespace NLog.Internal
         /// </summary>
         /// <param name="exception">The exception to check.</param>
         /// <param name="loggerContext">Target context of the exception.</param>
-        /// <param name="logFactory">For checking <see cref="LogFactory.ThrowExceptions"/> and <see cref="LogFactory.ThrowConfigExceptions"/>. Falls back to <see cref="LogManager"/> </param>
+
         /// <returns><c>true</c>if the <paramref name="exception"/> must be rethrown, <c>false</c> otherwise.</returns>
-        public static bool MustBeRethrown(this Exception exception, IInternalLoggerContext loggerContext = null, LogFactory logFactory = null)
+        public static bool MustBeRethrown(this Exception exception, IInternalLoggerContext loggerContext = null)
         {
             if (exception.MustBeRethrownImmediately())
             {
@@ -100,7 +100,7 @@ namespace NLog.Internal
                     InternalLogger.Log(exception, level, "Error has been raised.");
             }
 
-            logFactory = logFactory ?? LogManager.LogFactory;
+            var logFactory = loggerContext?.LogFactory ?? LogManager.LogFactory;
 
             //if ThrowConfigExceptions == null, use  ThrowExceptions
             var shallRethrow = isConfigError ? (logFactory.ThrowConfigExceptions ?? logFactory.ThrowExceptions) : logFactory.ThrowExceptions;
--- a/src/NLog/Internal/ExceptionHelper.cs
+++ b/src/NLog/Internal/ExceptionHelper.cs
@@ -100,10 +100,12 @@ namespace NLog.Internal
                     InternalLogger.Log(exception, level, "Error has been raised.");
             }
 
-            var logFactory = loggerContext?.LogFactory ?? LogManager.LogFactory;
+            var logFactory = loggerContext?.LogFactory;
 
-            //if ThrowConfigExceptions == null, use  ThrowExceptions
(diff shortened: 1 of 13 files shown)


src/NLog/Internal/ExceptionHelper.cs
```diff
@@ -78,8 +78,9 @@ public static bool IsLoggedToInternalLogger(this Exception exception)
         /// </summary>
         /// <param name="exception">The exception to check.</param>
         /// <param name="loggerContext">Target context of the exception.</param>
+        /// <param name="logFactory">For checking <see cref="LogFactory.ThrowExceptions"/> and <see cref="LogFactory.ThrowConfigExceptions"/>. Falls back to <see cref="LogManager"/> </param>
         /// <returns><c>true</c>if the <paramref name="exception"/> must be rethrown, <c>false</c> otherwise.</returns>
-        public static bool MustBeRethrown(this Exception exception, IInternalLoggerContext loggerContext = null)
+        public static bool MustBeRethrown(this Exception exception, IInternalLoggerContext loggerContext = null, LogFactory logFactory = null)
```
snakefoot, line 83: Maybe add LoggingConfiguration as property to `IInternalLoggerContext`, and then extract the LogFactory from that?

src/NLog/Internal/ExceptionHelper.cs
```diff
@@ -99,8 +100,10 @@ public static bool MustBeRethrown(this Exception exception, IInternalLoggerConte
                     InternalLogger.Log(exception, level, "Error has been raised.");
             }
 
+            logFactory = logFactory ?? LogManager.LogFactory;
```
snakefoot, line 103: Think to make it non-breaking then you should both check local `logFactory` AND global `LogManager.LogFactory`

src/NLog/Internal/ExceptionHelper.cs
304NotModified: OK will do!

src/NLog/Internal/ExceptionHelper.cs
304NotModified: OK, So I think

- logfactory.ThrowConfigExceptions, if null, then fallback to LogManager.ThrowConfigExceptions
- logfactory.ThrowExceptions, if false, then fallback to LogManager.ThrowExceptions? (unsure about this one)

snakefoot: If one of them is `true` then it should throw and explode.

304NotModified: I'm not sure if we should keep it the old behavior, is kinda bug that the properties on LogFactory aren't used.

snakefoot: Yes "bug" should be fixed with NLog 5.0 where lot of static behavior is being fixed.

304NotModified:
maybe move this to NLog 5 then for the breaking changes?

snakefoot:
Where is the breaking change?


Sent from my Sony Xperia


maybe move this to NLog 5 then for the breaking changes?

304NotModified:
this https://github.com/NLog/NLog/pull/3953#discussion_r423172916

snakefoot:
> this #3953 (comment)

As it is like this, then I don't see it as breaking:

```c#
if (logFactory?.ThrowExcetions==true || LogManager.ThrowExceptions)
{
}
```

304NotModified:
it. hurts. 

I've added this because in tests I still need to change the logmanager while I never use logmanager. But OK, that's a NLog 5 thing. Will add a fat todo

src/NLog/Internal/ExceptionHelper.cs
304NotModified: Updated it!

304NotModified:
@snakefoot please double check (the last commit should be enough)

snakefoot:
Looks good to me.

304NotModified:
Unfortunately Travis is still unstable sometimes:

image

snakefoot:
> Unfortunately Travis is still unstable sometimes:

Guess more of those Wait-forever calls needs to be removed, or maybe Travis-machines are just unstable.

304NotModified:
> Guess more of those Wait-forever calls needs to be removed, or maybe Travis-machines are just unstable.

Could be the latter, as .NET isn't officially supported. Maybe move to devops for Linux/Mono also

snakefoot:
Find that Azure Devops is worse than Travis and AppVeyor :)

304NotModified:
Why? It has tons of more features and really a better GUI

snakefoot:
Stability wise.

304NotModified:
True, we have some timeouts there also (60m).

PR #3954, opened 2020-05-11
Added LogFactory.ConfigurationLoading for better reload event

snakefoot:
Allow #3929 to use `ConfigurationLoading` (Because `ConfigurationReloaded` only works for xml-config. And `ConfigurationReloaded` happens AFTER target initialization):

```c#
            if (applyOnReload)
            {
                setupBuilder.LogFactory.ConfigurationLoading += (sender, e) =>
                {
                    if (e.IsNewConfiguration)
                    {
                         setupBuilder.Configuration = e.ActivatedConfiguration;
                         setupBuilder.LoadConfiguration(false, configBuilder);
                    }
```

This will make it also work for MEL-NLogLoggingConfiguration.

Maybe mark `ConfigurationReloaded` as obsolete in NLog 5.0? (Since that event-handler only works for xml-config)

### DIFF
--- a/src/NLog/Config/LoggingConfiguration.cs
+++ b/src/NLog/Config/LoggingConfiguration.cs
@@ -68,6 +68,8 @@ namespace NLog.Config
         /// </summary>
         public LogFactory LogFactory { get; }
 
+        internal bool IsConfigActivated { get; set; }
+
         /// <summary>
         /// Initializes a new instance of the <see cref="LoggingConfiguration" /> class.
         /// </summary>
--- a/src/NLog/Config/LoggingConfiguration.cs
+++ b/src/NLog/Config/LoggingConfiguration.cs
@@ -68,6 +68,9 @@ namespace NLog.Config
         /// </summary>
         public LogFactory LogFactory { get; }
 
+        /// <summary>
+        /// Is this configuration activated (so assigned at least once to a logFactory in the past)
+        /// </summary>
         internal bool IsConfigurationActivated { get; set; }
 
         /// <summary>
--- a/src/NLog/Config/LoggingConfiguration.cs
+++ b/src/NLog/Config/LoggingConfiguration.cs
@@ -68,7 +68,7 @@ namespace NLog.Config
         /// </summary>
         public LogFactory LogFactory { get; }
 
-        internal bool IsConfigActivated { get; set; }
+        internal bool IsConfigurationActivated { get; set; }
 
         /// <summary>
         /// Initializes a new instance of the <see cref="LoggingConfiguration" /> class.
--- a/src/NLog/Config/LoggingConfigurationLoadingEventArgs.cs
+++ b/src/NLog/Config/LoggingConfigurationLoadingEventArgs.cs
@@ -1,4 +1,4 @@
-// 
+﻿// 
 // Copyright (c) 2004-2020 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
 // 
 // All rights reserved.
@@ -36,50 +36,34 @@ namespace NLog.Config
     using System;
 
     /// <summary>
-    /// Arguments for <see cref="LogFactory.ConfigurationChanged"/> events.
+    /// Arguments for <see cref="LogFactory.ConfigurationLoading"/> events.
     /// </summary>
-    public class LoggingConfigurationChangedEventArgs : EventArgs
+    public class LoggingConfigurationLoadingEventArgs : EventArgs
     {
         /// <summary>
-        /// Initializes a new instance of the <see cref="LoggingConfigurationChangedEventArgs" /> class.
+        /// Initializes a new instance of the <see cref="LoggingConfigurationLoadingEventArgs" /> class.
         /// </summary>
-        /// <param name="activatedConfiguration">The new configuration.</param>
-        /// <param name="deactivatedConfiguration">The old configuration.</param>
-        public LoggingConfigurationChangedEventArgs(LoggingConfiguration activatedConfiguration, LoggingConfiguration deactivatedConfiguration)
+        /// <param name="activatingConfiguration">The new configuration.</param>
+        /// <param name="deactivatingConfiguration">The old configuration.</param>
+        public LoggingConfigurationLoadingEventArgs(LoggingConfiguration activatingConfiguration, LoggingConfiguration deactivatingConfiguration)
         {
-            ActivatedConfiguration = activatedConfiguration;
-            DeactivatedConfiguration = deactivatedConfiguration;
+            ActivatingConfiguration = activatingConfiguration;
+            DeactivatingConfiguration = deactivatingConfiguration;
         }
 
         /// <summary>
         /// Gets the old configuration.
         /// </summary>
-        /// <value>The old configuration.</value>
-        public LoggingConfiguration DeactivatedConfiguration { get; private set; }
+        public LoggingConfiguration DeactivatingConfiguration { get; private set; }
 
         /// <summary>
         /// Gets the new configuration.
         /// </summary>
-        /// <value>The new configuration.</value>
-        public LoggingConfiguration ActivatedConfiguration { get; private set; }
+        public LoggingConfiguration ActivatingConfiguration { get; private set; }
 
         /// <summary>
-        /// The <see cref="ActivatedConfiguration"/> has been loaded for the first time
+        /// The <see cref="ActivatingConfiguration"/> is being loaded for the first time
         /// </summary>
         public bool IsNewConfiguration { get; internal set; }
-
-        /// <summary>
-        /// Gets the new configuration
-        /// </summary>
-        /// <value>The new configuration.</value>
-        [Obsolete("This option will be removed in NLog 5. Marked obsolete on NLog 4.5")]
-        public LoggingConfiguration OldConfiguration => ActivatedConfiguration;
-
-        /// <summary>
-        /// Gets the old configuration
-        /// </summary>
-        /// <value>The old configuration.</value>
-        [Obsolete("This option will be removed in NLog 5. Marked obsolete on NLog 4.5")]
-        public LoggingConfiguration NewConfiguration => DeactivatedConfiguration;
     }
 }
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -264,7 +264,7 @@ namespace NLog
                         oldConfig.Close();
                     }
 
-                    var isNewConfiguration = value?.IsConfigActivated == false;
+                    var isNewConfiguration = value?.IsConfigurationActivated == false;
 
                     _config = value;
                     if (_config == null)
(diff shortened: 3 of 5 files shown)


304NotModified:
Hey! I was reviewing, but don't fully get it. Why has a *change* something to do with *activated*. ?

Could it be `e.IsInitialConfiguration` instead of `e.InitialConfigActivation`?

snakefoot:
> Could it be e.IsInitialConfiguration instead of e.InitialConfigActivation?

Lets say you do this:

```
LogManager.Configuration = config;  // InitialConfigActivation = true
LogManager.Configuration = config;  // InitialConfigActivation = false
LogManager.Configuration = config;  // InitialConfigActivation = false

var config2 = new LoggingConfiguration();
LogManager.Configuration = config2;  // InitialConfigActivation = true
LogManager.Configuration = config2;  // InitialConfigActivation = false
LogManager.Configuration = config;    // InitialConfigActivation = false
```

Think `IsInitialConfiguration` sounds like it is the very first assignment, but you can assign the same configuration multiple times. But you can also assign different configs for every time.

304NotModified:
> Think `IsInitialConfiguration` sounds like it is the very first assignment

Well, InitialConfigActivation doesn't sound as a boolean to me ;)

304NotModified:
I sounds more like a "FirstApplyOfThisConfig" boolean, but it's hard to understand/use!

snakefoot:
What about `IsInitialActivation` ?

snakefoot:
Btw. created this PR mostly as an example of how to use `ConfigurationChanged` instead of `ConfigurationReloaded`. Because it would make it more generic and work with MEL-NLogLoggingConfiguration.

But happy you are approving and we are just arguing about a property-names :)

304NotModified:
out-of-the-box thinking...

What about a new event, and leave this one unchanged? As now you capture events but are only interested in a subset which is a bit strange. 

Maybe don't add the event on logmanager, but only on logfactory.

snakefoot:
> What about a new event, and leave this one unchanged? As now you capture events but are only interested in a subset which is a bit strange.

I like the idea of using `ConfigurationChanged`. Hate where there are too many door-handles to the same door (Sickness that comes with some libraries where there is 3 standard ways to do the same thing).

304NotModified:
Well it sounds so unrelated to ConfigurationChanged to me. I find it hard to understand

btw, I think it should be onReloaded. Which is good, as I changed applyOnReload already to reloaded event (not merged yet of course)

snakefoot:
> Well it sounds so unrelated to ConfigurationChanged to me. I find it hard to understand

Think you have to explain this some more. I'm not changing `ConfigurationChanged`. I'm just adding more context-detail to the EventArg.

> btw, I think it should be onReloaded. Which is good, as I changed applyOnReload already to reloaded event (not merged yet of course)

Again we are back to having too many events. Think we should get rid of `ConfigurationReloaded` and focus on only recommending `ConfigurationChanged`

304NotModified:
> Think you have to explain this some more. I'm not changing `ConfigurationChanged`. I'm just adding more context-detail to the EventArg (Whether it is a new config assigned for the first time, or whether it is same config being re-assigned)

Yes that's clear. But of course the properties should be understandable. 

> Again we are back to having too many events. Think we should get rid of ConfigurationReloaded and focus on only recommending (and improving) ConfigurationChanged (Easier for us, and easier for the NLog users)

That's another discussion on which on don't agree. Reload could lead to change, but also not. And Change could be triggered by reload but also another way.

snakefoot:
> Yes that's clear. But of course the properties should be understandable.

Yes also want a clear and understandable name, that can signal that the change was done with a fresh/vanilla config, that has not yet been touched.

snakefoot:
I now see that there is a vital knob missing in this PR. the `ConfigurationChanged`-event is triggered AFTER the config has been assigned. `applyOnReload` wants to trigger BEFORE assigning (and initializing) the config.

snakefoot:
> @304NotModified  What about a new event, and leave this one unchanged?

Think you are right about this. Now introduced `ConfigurationChanging` (I know I have reused `LoggingConfigurationChangedEventArgs` as EventArgs and that will be changed, but right now working on the mental drawing board).

snakefoot:
@304NotModified How do you feel about having both `ConfigurationChanging` and `ConfigurationChanged` ?

304NotModified:
> @304NotModified How do you feel about having both `ConfigurationChanging` and `ConfigurationChanged` ?

I have to think about this.

304NotModified:
> Maybe mark ConfigurationReloaded as obsolete in NLog 5.0? (Since that event-handler only works for xml-config)

This is difficult as I see a manual assignment as change, but not as reload. 

Maybe stupid question, but what goes wrong if we always do LoadConfiguration? E.g.:


> ```cs
>             if (applyOnReload)
>             {
>                 setupBuilder.LogFactory.ConfigurationChanging += (sender, e) =>
>                 {
>                          setupBuilder.LoadConfiguration(false, configBuilder);
> ```

snakefoot:
Let's say the configBuilder method adds a new target to config and then re-assigns LogFactory.Configuration.


Sent from my Sony Xperia


Maybe mark ConfigurationReloaded as obsolete in NLog 5.0? (Since that event-handler only works for xml-config)

This is difficult as I see a manual assignment as change, but not as reload.

Maybe stupid question, but what goes wrong if we always do LoadConfiguration? E.g.:

            if (applyOnReload)
            {
                setupBuilder.LogFactory.ConfigurationChanging += (sender, e) =>
                {
                         setupBuilder.LoadConfiguration(false, configBuilder);

snakefoot:
Can also see that `ConfigurationReloaded` is not optimal for hooking `configBuilder`. Because the config-method will be called AFTER having loaded and initialized the configuration.

The idea with `LoadConfiguration` is that you adjust the LoggingConfiguration before assigning it as the new active LogFactory.Configuration. When using `ConfigurationReloaded`-event then it will be too late for adjusting Target-properties that must be configured before Target-initialization.

snakefoot:
Added unit-test that will only succeed when using `ConfigurationChanging`.

snakefoot:
Could also rename it from `ConfigurationChanging` to `ConfigurationLoading` and ensure it is only called when `if (!e.IsConfigAmend && e.ActivatedConfiguration != null)`. Like this:
```c#
            if (applyOnReload)
            {
                setupBuilder.LogFactory.ConfigurationLoading += (sender, e) =>
                {
                     setupBuilder.Configuration = e.NewConfiguration;
                     setupBuilder.LoadConfiguration(false, configBuilder);
```

304NotModified:
>  Could also rename it from `ConfigurationChanging` to `ConfigurationLoading`

If we gonna drop ConfigurationLoaded soon, then that's a bit strange? 

Anyway, I really like another name for amend, but don't have a proposal. (I tried thinking of a new name by asking questions, but unfortunately I don't have one)

snakefoot:
> If we gonna drop ConfigurationLoaded soon, then that's a bit strange?

Dropping `ConfigurationReloaded` for `ConfigurationLoading` doesn't sound strange to me. 

Also `ConfigurationReloaded` only works for xml-config, and the event is merely a left-over from when LogFactory was doing all the file-loading itself. I have now moved most of it out of LogFactory, but actually all logic about file-watching etc. should be controlled by XmlLoggingConfiguration (and not be visible in LogFactory).

snakefoot:
Instead of `IsConfigAmend` then it could also be `IsNewConfiguration`

snakefoot:
@304NotModified I'm happy with the PR now. Decided it is best to trigger the `ConfigurationLoading` after having flushed/closed the old-config, but just before initializing the new-config. Then the system should be somewhat "quiet".

304NotModified:
> Maybe mark ConfigurationReloaded as obsolete in NLog 5.0? (Since that event-handler only works for xml-config)

So for ApplyOnReload, we need to change it to ConfigurationChanged isn't? Too bad that will give a log of tests needed (shutdown etc)

snakefoot:
> So for ApplyOnReload, we need to change it to ConfigurationChanged isn't? 

Yes change it to [ConfigurationLoading](https://github.com/NLog/NLog/pull/3954#issue-416215701)

304NotModified:
and what about LogManager? Do you think we should include this also there?

304NotModified:
> Yes change it to [ConfigurationLoading](https://github.com/NLog/NLog/pull/3954#issue-416215701)

No that's not possible? It should be changed after loading? (hmmm or not, not sure)

304NotModified:
> No that's not possible? It should be changed after loading? (hmmm or not, not sure)

And also not sure if it works with assigning (it should not apply then). But we could see after merge. Or maybe I should test it first...

snakefoot:
The whole idea with LoadConfiguration-method is to afjust NLog before it is loaded. So the same should apply when enabling applyOnReload


Sent from my Sony Xperia


Yes change it to ConfigurationLoading<https://github.com/NLog/NLog/pull/3954#issue-416215701>

No that's not possible? It should be changed after loading? (hmmm or not, not sure)

304NotModified:
I will test this with #3929 now

304NotModified:
Doesn't work, as this is now a "reload".

 And while that's true, then you can reset stuff. 

This is expected (and maybe a better case)

    logFactory.Configuration = new ... // don't expect applyonreload here.

snakefoot:
Well to me there is no difference in loading a new configuration and reloading a new configuration. But I also think that applyOnReload should false by default, as users should opt-in to the side-effect.

304NotModified:
> Well to me there is no difference in loading a new configuration and reloading a new configuration. But I also think that applyOnReload should be false by default, as users should opt-in to the automagically side-effect.

Yes So I think we still need a reload event.

As this should not applyOnReload
```c#
logFactory.Configuration = something
```

As this should
```c#
logFactory.ReloadConfiguration()
```

304NotModified:
So I need an approval on https://github.com/NLog/NLog/pull/3952 ;)

snakefoot:
> Yes So I think we still need a reload event.

I don't agree. But I have now made some requests for #3952

304NotModified:
> I don't agree. But I have now made some requests for #3952

don't agree with https://github.com/NLog/NLog/pull/3954#issuecomment-628266967, or need for a reload event?

snakefoot:
I think that `ConfigurationLoading` will work just fine, and do the expected. But never too late to kill your darlings.

304NotModified:
> I think that `ConfigurationLoading` will work just fine, and do the expected. But never too late to kill your darlings.

I don't know which one you now mean  (in "don't agree with the functionality described in #3954 (comment), OR the need for a reload event?")

snakefoot:
> I don't know which one you now mean (in "don't agree with the functionality described in #3954 (comment), OR the need for a reload event?")

That you are welcome to close this PR (without implementing ConfigurationLoading), and instead implement the same logic #3952 (With IsNewConfiguration and allowing one to provide LoggingConfiguration as parameter to ReloadConfiguration).

snakefoot:
Closed as inactive for too long. See also #4250

PR #3957, opened 2020-05-13
Support applyOnReload for LoadConfiguration (V2)

304NotModified:


304NotModified:
@snakefoot this test fails now:

image


```c#
        [Theory]
        [InlineData(true, true)]
        [InlineData(false, true)]
        [InlineData(true, false)]
        [InlineData(false, false)]
        public void SetupBuilderMultipleLoadConfigurationApplyOnReload_manualShouldNotCountAsReload(bool applyOnReload1, bool applyOnReload2)
        {
            // Arrange

            var logFactory = ArrangeMultipleLoadConfigurationApplyOnReload(applyOnReload1, applyOnReload2);

            // Assert
            Assert.Equal(0, logFactory.Configuration.LoggingRules.Count);
        }
```

snakefoot:
@304NotModified Well it is your own test, so I guess you have to explain what you expect.

When you setup hook to apply-changes automatically whenever a new configuration is loaded, then it works as intended.

If you only want it to apply-changes automatically, then I suggest one just use applyOnReload=false, or call the default `LoadConfiguration`-method without the applyOnReload-parameter (so it defaults to applyOnReload=false)

PR #3962, opened 2020-05-15
XSD: Added Enabled attribute for <logger> + small description improvement

304NotModified:


tools/MakeNLogXSD/TemplateXSD.xml
```diff
@@ -174,6 +174,12 @@
       </xs:annotation>
     </xs:attribute>
 
+    <xs:attribute name="enabled" type="xs:boolean" default="true">
+      <xs:annotation>
+        <xs:documentation>Is this rule enabled?</xs:documentation>
```
304NotModified, line 179: Maybe a bit too simple text. Really like input.

snakefoot: Maybe "Is the rule enabled and should be included in the actual LoggingConfiguration?" (Since it will be discarded if enabled=false).

304NotModified: Good one about the discard, 

After thinking, I didn't like the question form. So made it now like this: 

`Enable this rule. Note: disabled rules aren't available from the API.`

PR #3965, opened 2020-05-16
Shutdown should not block on exceptions from targets

### LINKED ISSUE #3958: NLog Target that blocks in FlushAync can never be closed
This makes it impossible to complete `LogManager.Shutdown()`

It will correctly complain with timeout error on Flush using threadpool-threads. But afterwards when closing all targets then it will get stuck in Target.Close because there threadpool-thread is holding SyncRoot inside FlushAsync.

Ofcourse a NLog Target should not perform blocking operations inside `FlushAsync` and hold the SyncRoot-lock forever, but it could be nice that  `LogManager.Shutdown()` could still complete (with alerts) and not be stuck forever.

Maybe have a check that if the Flush-completes with timeout (because of missing cal to continuation), then it should also validate that all threadpool-threads have exited the FlushAync-method of the target. If threadpool-threads are stuck, then it should perform async-close-operation (or just skip target-close) instead of blocking shutdown forever.

snakefoot:
LogFactory Shutdown performs async close of configuration, when timeout on flush

Trying to resolve #3958

### DIFF
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -257,7 +257,10 @@ namespace NLog
                         if (!FlushAllTargetsSync(oldConfig, DefaultFlushTimeout, ThrowExceptions))
                         {
                             var manualResetEvent = new ManualResetEvent(false);
-                            AsyncHelpers.ForEachItemInParallel(new[] { oldConfig }, ex => manualResetEvent.Set(), (old, cont) => { old.Close(); cont(null); });
+                            AsyncHelpers.ForEachItemInParallel(new[] { oldConfig },
+                                ex => manualResetEvent.Set(),
+                                (old, cont) => { old.Close(); cont(null); }
+                            );
                             if (!manualResetEvent.WaitOne(DefaultFlushTimeout))
                             {
                                 InternalLogger.Info("Timeout closing old configuration.");
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -254,8 +254,19 @@ namespace NLog
                     if (oldConfig != null)
                     {
                         InternalLogger.Info("Closing old configuration.");
-                        Flush();
-                        oldConfig.Close();
+                        if (!FlushAllTargetsSync(oldConfig, DefaultFlushTimeout, ThrowExceptions))
+                        {
+                            var manualResetEvent = new ManualResetEvent(false);
+                            AsyncHelpers.ForEachItemInParallel(new[] { oldConfig }, ex => manualResetEvent.Set(), (old, cont) => { old.Close(); cont(null); });
+                            if (!manualResetEvent.WaitOne(DefaultFlushTimeout))
+                            {
+                                InternalLogger.Info("Timeout closing old configuration.");
+                            }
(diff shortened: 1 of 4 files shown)


src/NLog/LogFactory.cs
```diff
@@ -254,8 +254,19 @@ public LoggingConfiguration Configuration
                     if (oldConfig != null)
                     {
                         InternalLogger.Info("Closing old configuration.");
-                        Flush();
-                        oldConfig.Close();
+                        if (!FlushAllTargetsSync(oldConfig, DefaultFlushTimeout, ThrowExceptions))
+                        {
+                            var manualResetEvent = new ManualResetEvent(false);
+                            AsyncHelpers.ForEachItemInParallel(new[] { oldConfig }, ex => manualResetEvent.Set(), (old, cont) => { old.Close(); cont(null); });
```
304NotModified, line 260: Maybe we could make a new method in AsyncHelpers? `ForEachItemInParallel(new[] { oldConfig }` is a bit strange to read. 

(PS I prefer also to break this line with some newlines, e.g. something like this:

```c#
 AsyncHelpers.ForEachItemInParallel(new[] { oldConfig },
     ex => manualResetEvent.Set(), 
     (old, cont) => { old.Close(); cont(null); }
   );

```

src/NLog/LogFactory.cs
```diff
@@ -254,8 +254,19 @@ public LoggingConfiguration Configuration
                     if (oldConfig != null)
                     {
                         InternalLogger.Info("Closing old configuration.");
-                        Flush();
-                        oldConfig.Close();
+                        if (!FlushAllTargetsSync(oldConfig, DefaultFlushTimeout, ThrowExceptions))
```
304NotModified, line 257: Funny that the setter of Configuration has changed. it expected to see some changes in Flush(async) or Shutdown methods. 

Could this PR have more side effects?

src/NLog/LogFactory.cs
snakefoot: Added newlines.

src/NLog/LogFactory.cs
snakefoot: Well it has the side-effect that if flush fails with an exception, then it uses a threadpool thread for closing the configuration. And will continue after 15 secs (Even if a target requires 2 minutes to flush and close).

304NotModified:
@snakefoot Do you think it;s wise to release this one with NLog 4.7.2?

snakefoot:
> @snakefoot Do you think it;s wise to release this one with NLog 4.7.2?

I think something should be done about this blocking forever. Since it points fingers at NLog for not working. And I also think 30 seconds is reasonable long time to wait for flush and close.

Since the async closing of targets only happens when flush has failed with exception, then I was thinking it was safe.

snakefoot:
Think this PR needs some more thought, since unsure if there will be issues with holding LogFactory-lock and doing async Close on other thread.

Actually think the flush should be done outside the lock, and it should also do the async-close outside the lock if the flush fails. Maybe better for NLog5.

snakefoot:
Inactive for too long

PR #3974, opened 2020-05-25
${exception) - Improved logging AggregrateException's Data

chaos0307:
Current version of NLog is not able to log AggregrateException.Data as it will be lost after it was Flatten. Fixed it by assigned it to a temporary variable and add it to the newly flatten exception.

Unit test is include as well. 

Cheers.

### DIFF
--- a/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
@@ -36,8 +36,6 @@ namespace NLog.LayoutRenderers
     using System;
     using System.Collections.Generic;
     using System.ComponentModel;
-    using System.Linq;
-    using System.Reflection;
     using System.Text;
     using NLog.Common;
     using NLog.Config;
@@ -54,7 +52,7 @@ namespace NLog.LayoutRenderers
     {
         private string _format;
         private string _innerFormat = string.Empty;
-        private readonly Dictionary<ExceptionRenderingFormat, Action<StringBuilder, Exception>> _renderingfunctions;
+        private readonly Dictionary<ExceptionRenderingFormat, Action<StringBuilder, Exception, Exception>> _renderingfunctions;
 
         private static readonly Dictionary<string, ExceptionRenderingFormat> _formatsMapping = new Dictionary<string, ExceptionRenderingFormat>(StringComparer.OrdinalIgnoreCase)
                                                                                                     {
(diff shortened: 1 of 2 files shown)


src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -229,8 +230,12 @@ protected override void Append(StringBuilder builder, LogEventInfo logEvent)
 #if !NET3_5 && !SILVERLIGHT4
                 if (logEvent.Exception is AggregateException aggregateException)
                 {
+                    var aggregateExceptionData = aggregateException.Data;
                     aggregateException = aggregateException.Flatten();
                     primaryException = GetPrimaryException(aggregateException);
+                    foreach (var key in aggregateExceptionData.Keys)
+                        primaryException.Data.Add(key, aggregateExceptionData[key]);
```
snakefoot, line 237: Prefer that the logging-library only probes the Exception, and doesn't modify the Exceptions.

snakefoot:
Maybe something like this:

```c#
        private void AppendException(Exception currentException, List<ExceptionRenderingFormat> renderFormats, StringBuilder builder, Exception aggregateException)
        {
            int currentLength = builder.Length;
            foreach (ExceptionRenderingFormat renderingFormat in renderFormats)
            {
                int beforeRenderLength = builder.Length;
                var currentRenderFunction = _renderingfunctions[renderingFormat];
                if (renderingFormat == ExceptionRenderingFormat.Data)
                {
                     var dataException = (aggregateException?.Data?.Count > 0 && currentException?.Data?.Count == 0) ? aggregateException : currentException;
                     currentRenderFunction(builder, dataException);
                }
                else
                {
                   currentRenderFunction(builder, currentException);
                }
                if (builder.Length != beforeRenderLength)
                {
                    currentLength = builder.Length;
                    builder.Append(Separator);
                }
            }

            builder.Length = currentLength;
        }
```

Alternative one should perform the lookup of `primaryException` inside each ExceptionRenderingFormat-function. And then the `AppendData` could have the exotic logic of scanning the entire exception-tree.

src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
```diff
@@ -309,15 +305,27 @@ private void AppendInnerException(Exception currentException, StringBuilder buil
             builder.Append(InnerExceptionSeparator);
             AppendException(currentException, InnerFormats ?? Formats, builder);
         }
-
+#if !NET3_5 && !SILVERLIGHT4
+        private void AppendException(Exception currentException, List<ExceptionRenderingFormat> renderFormats, StringBuilder builder, AggregateException aggregateException = null)
```
snakefoot, line 309: If you change it to `Exception aggregateException = null`, then you can remove `#if !NET3_5 && !SILVERLIGHT4`

chaos0307: i think i'll keep the `#if` and input param as type AggregateException as it only available post NET3_5

304NotModified: I prefer no `#if` if possible. As that is hard with refactoring. 👼 (no refactoring tool will handle those fully correct when moving)

snakefoot:
Happy that the `#if` is now gone. Could one also remove the if-statement. So it becomes:

```c#
_renderingfunctions = new Dictionary<ExceptionRenderingFormat, Action<StringBuilder, Exception,Exception>>()
{
{ExceptionRenderingFormat.Message, (sb, ex, aggex) => AppendMessage(sb, ex)},
{ExceptionRenderingFormat.Type, (sb, ex, aggex) => AppendType(sb, ex)},
{ExceptionRenderingFormat.ShortType, (sb, ex, aggex) => AppendShortType(sb, ex)},
{ExceptionRenderingFormat.ToString, (sb, ex, aggex) => AppendToString(sb, ex)},
{ExceptionRenderingFormat.Method, (sb, ex, aggex) => AppendMethod(sb, ex)},
{ExceptionRenderingFormat.Source, (sb, ex, aggex) => AppendSource(sb, ex)},
{ExceptionRenderingFormat.StackTrace, (sb, ex, aggex) => AppendStackTrace(sb, ex)},
{ExceptionRenderingFormat.Data, (sb, ex, aggex) => AppendData(sb, ex, aggex)},
{ExceptionRenderingFormat.Serialize, (sb, ex, aggex) => AppendSerializeObject(sb, ex)},
{ExceptionRenderingFormat.HResult, (sb, ex, aggex) => AppendHResult(sb, ex)},
{ExceptionRenderingFormat.Properties, (sb, ex, aggex) => AppendProperties(sb, ex)},
};
```

And then have this:

```c#
        private void AppendException(Exception currentException, List<ExceptionRenderingFormat> renderFormats, StringBuilder builder, Exception aggregateException = null)
        {
            int currentLength = builder.Length;
            foreach (ExceptionRenderingFormat renderingFormat in renderFormats)
            {
                int beforeRenderLength = builder.Length;
                var currentRenderFunction = _renderingfunctions[renderingFormat];
                currentRenderFunction(builder, currentException, aggregateException );
                if (builder.Length != beforeRenderLength)
                {
                    currentLength = builder.Length;
                    builder.Append(Separator);
                }
            }

            builder.Length = currentLength;
        }
```

And then have this new method:

```c#
private void AppendData(StringBuilder builder, Exception ex, Exception aggregateException = null)
{
      if (aggregateException?.Data?.Count > 0)
      {
            AppendData(builder, aggregateException);
            builder.Append(Separator);
      }
      AppendData(builder, ex);
}
```

snakefoot:
Could also consider to get rid of the Flatten-logic by default:

```c#
private static Exception GetPrimaryException(AggregateException aggregateException)
{
   if (aggregateException.InnerExceptions.Count == 1)
   {
       var innerException = aggregateException.InnerExceptions[0];
       if (!(innerException is AggregateException))
            return innerException;  // Skip calling Flatten()
   }
 
   aggregateException = aggregateException.Flatten();
   if (aggregateException.InnerExceptions.Count == 1)
   {
       return aggregateException.InnerExceptions[0];
   }
 
   return aggregateException;
}
```

```c#
#if !NET3_5 && !SILVERLIGHT4
                if (logEvent.Exception is AggregateException aggregateException)
                {
                   primaryException = GetPrimaryException(aggregateException);
                   AppendException(primaryException, Formats, builder, aggregateException);
                   if (currentLevel < MaxInnerExceptionLevel)
                   {
                      currentLevel = AppendInnerExceptionTree(primaryException, currentLevel, builder);
                      if (currentLevel < MaxInnerExceptionLevel)
                      {
                          aggregateException = primaryException as AggregateException;
                          if (aggregateException?.InnerExceptions?.Count > 1)
                          {
                             AppendAggregateException(aggregateException, currentLevel, builder);
                          }
                      }
                   }
                }
```

snakefoot:
Could consider to rebase to "master", then it will not wait for NLog 5.0

snakefoot:
@chaos0307 Wow you are fast. Then just need to make the [code-smell-engine](https://sonarcloud.io/project/issues?id=nlog2&pullRequest=3974&resolved=false&types=CODE_SMELL) happy (by removing default parameter value `= null`)

`private void AppendData(StringBuilder builder, Exception ex, Exception aggregateException = null)`

> This method signature overlaps the one defined on line 467, the default parameter value can only be used with named arguments.Why is this an issue?

304NotModified:
Cool! I think this is done now, agree @snakefoot ?

snakefoot:
@304NotModified Yes I agree. Now just need to decide if for NLog 4.7.3 (`master`) or for NLog 5.0 (`dev`)

304NotModified:
I think 4.7.3 is better. Agree?

If so, @chaos0307 could you please rebase this on master? (maybe squash first). If you need help, let us know!

snakefoot:
Also think NLog 4.7.3 is good.

chaos0307:
@304NotModified should i rebase to dev branch and then to master branch? the current working branch is branch out from dev though

304NotModified:
I've squased and rebased on master :)

304NotModified:
(PS sonarcloud has issues, outside our control)

PR #3975, opened 2020-05-25
Capture CallSite ClassName only when WithCallSiteClassName enabled

snakefoot:
Extends #3431 to also support #3915 (CaptureStackTrace=false).

Making it possible to capture class-name for CallSite using CallerMember-Attributes alone, but only when available.

Minor optimization so it only attempts to capture class-name for CallSite, when explicitly requested (Similar to only capturing source-filename and source-linenumber for stacktrace).


304NotModified:
nice! Thanks!

Issue #962, opened 2015-10-16
LogFactory<T> should have new()  restriction


304NotModified:
Related https://github.com/NLog/NLog/issues/961

We try to create a new instance of T. That wont work if there is not a default ctor.

``` c#
   public class LogFactory<T> : LogFactory       where T : Logger
```

Should be

``` c#
   public class LogFactory<T> : LogFactory       where T : Logger, new()
```

PR #3986, opened 2020-05-29
Added new() contraint on LogFactory<T>

304NotModified:
Fixes https://github.com/NLog/NLog/issues/962

### DIFF
--- a/src/NLog/LogFactory-T.cs
+++ b/src/NLog/LogFactory-T.cs
@@ -42,7 +42,7 @@ namespace NLog
     /// <remarks>Use this only when a custom Logger type is defined.</remarks>
     /// <typeparam name="T">The type of the logger to be returned. Must inherit from <see cref="Logger"/>.</typeparam>
     public class LogFactory<T> : LogFactory 
-        where T : Logger
+        where T : Logger, new()
     {
         /// <summary>
         /// Gets the logger with type <typeparamref name="T"/>.


snakefoot:
Think #3990 and #3986 should be one PR. And the idea is to call the new-operator.

304NotModified:
superseded by https://github.com/NLog/NLog/pull/3990

304NotModified:
>  And the idea is to call the new-operator.

Do you mean we should make a new method internally that is generic? I don't think that will work well with _loggerCache etc. (not work well as in: lot of code duplication, and not really in improvement?)

snakefoot:
Think one could send a factory-lambda into the internal GetLogger-method. But the whole new-exercise is fruitless if you are not calling the new-operator


Sent from my Sony Xperia


And the idea is to call the new-operator.

Do you mean we should make a new method internally that is generic? I don't think that will work well with _loggerCache etc. (not work well as in: lot of code duplication, and not really in improvement?)

304NotModified:
> Think one could send a factory-lambda into the internal GetLogger-method. But the whole new-exercise is fruitless if you are not calling the new-operator

I think the new constraint with the Activtor is just fine. The benefit of new() is that it's clear how the instance is created (and that we need a public ctor, although not the case for our own Logger).

Is there a (performance?) benefit of using `new` compared to the activator?

Issue #3744, opened 2019-12-28
TraceTarget - Rename RawWrite to ForceTraceWriteLine with default = true


snakefoot:
Avoid the unexpected surprises from using other Trace-methods than Trace.WriteLine.

Should be the new default for NLog 5.0

P.S. Could probably introduce ForceTraceWriteLine with NLog 4.7 (Marking RawWrite as obsolete), but changing the default value in NLog 5.0

snakefoot:
See also #3742

snakefoot:
Starting to reconsider changing the default, since Azure Log Streaming actually listens to the LogLevel:

https://www.thebestcsharpprogrammerintheworld.com/2018/08/23/system-diagnostics-trace-application-logging-log-stream-on-azure-app-service-function-app/

Changing the default will actually affect applications depending on Azure Log Streaming. Maybe the setting should just be left alone, since better to have some output (and fixing it), than no output because everything suddenly changes to LogLevel-Verbose (Instead of Info, Warning, Error).

PR #3988, opened 2020-05-29
TraceTarget - Rename RawWrite to ForceTraceWriteLine with default = true

304NotModified:
Fixes https://github.com/NLog/NLog/issues/3744

### DIFF
--- a/src/NLog/Targets/TraceTarget.cs
+++ b/src/NLog/Targets/TraceTarget.cs
@@ -66,8 +66,8 @@ namespace NLog.Targets
         /// Always use <see cref="Trace.WriteLine(string)"/> independent of <see cref="LogLevel"/>
         /// </summary>
         /// <docgen category='Output Options' order='100' />
-        [DefaultValue(false)]
-        public bool RawWrite { get; set; }
+        [DefaultValue(true)]
+        public bool ForceTraceWriteLine { get; set; } = true;
 
         /// <summary>
         /// Forward <see cref="LogLevel.Fatal" /> to <see cref="Trace.Fail(string)" /> (Instead of <see cref="Trace.TraceError(string)" />)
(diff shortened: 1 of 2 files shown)


src/NLog/Targets/TraceTarget.cs
```diff
@@ -66,8 +66,8 @@ public sealed class TraceTarget : TargetWithLayout
         /// Always use <see cref="Trace.WriteLine(string)"/> independent of <see cref="LogLevel"/>
         /// </summary>
         /// <docgen category='Output Options' order='100' />
-        [DefaultValue(false)]
-        public bool RawWrite { get; set; }
+        [DefaultValue(true)]
+        public bool ForceTraceWriteLine { get; set; } = true;
```
snakefoot, line 70: Think you need to keep `RawWrite` around as obsolete (Property that just assigns `ForceTraceWriteLine`)

304NotModified:
Sonarqube broken...

image

src/NLog/Targets/TraceTarget.cs
304NotModified: Maybe just keep the name RawWrite then.

snakefoot: Like how `ForceTraceWriteLine` explains what is going to happen. And when making the breaking change of change the default-behavior (to use Trace WriteLine by default). Then might as well do it nicely.

snakefoot:
Starting to reconsider changing the default, since Azure Log Streaming actually listens to the LogLevel:

https://www.thebestcsharpprogrammerintheworld.com/2018/08/23/system-diagnostics-trace-application-logging-log-stream-on-azure-app-service-function-app/

Changing the default will actually affect applications depending on Azure Log Streaming. Maybe the setting should just be left alone, since better to have some output (and fixing it), than no output because everything suddenly changes to LogLevel-Verbose (Instead of Info, Warning, Error).

304NotModified:
yes, i'm fine without changing the default

Issue #3946, opened 2020-05-04
DirectorySeparatorLayoutRenderer - Remove public TryGetValue


snakefoot:
No need to have it public:

https://github.com/NLog/NLog/blob/8fb2a857f7f9e0b5980336ddec622c8a98236725/src/NLog/LayoutRenderers/DirectorySeparatorLayoutRenderer.cs#L64

Could also consider removing static-constructor and instead have lazy-evaluated `SeparatorChar`

Could also consider making the class **sealed**.

PR #3991, opened 2020-05-30
DirectorySeparatorLayoutRenderer - Removed public TryGetValue

304NotModified:
fixes https://github.com/NLog/NLog/issues/3946 and made SeparatorChar lazy

### DIFF
--- a/src/NLog/LayoutRenderers/DirectorySeparatorLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/DirectorySeparatorLayoutRenderer.cs
@@ -61,7 +61,7 @@ namespace NLog.LayoutRenderers
         }
 
         /// <inheritdoc />
-        public bool TryGetRawValue(LogEventInfo logEvent, out object value)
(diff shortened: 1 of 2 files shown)


src/NLog/LayoutRenderers/DirectorySeparatorLayoutRenderer.cs
```diff
@@ -47,11 +47,20 @@ namespace NLog.LayoutRenderers
     [AppDomainFixedOutput]
     public class DirectorySeparatorLayoutRenderer : LayoutRenderer, IRawValue
     {
-        private static readonly char SeparatorChar;
+        private static char? _separatorChar;
```
snakefoot, line 50: Think this is overly complex, when looking at the implementation for [Path.DirectorySeparatorChar](https://referencesource.microsoft.com/#mscorlib/system/io/path.cs,56)

Think non-static `private char _separatorChar = Path.DirectorySeparatorChar` will work just fine.

304NotModified: Fixed!

Issue #3992, opened 2020-05-30
Don't combine IRawValue with [AppDomainFixed]


304NotModified:
From @snakefoot 
> IRawValue doesn't work in combination with [AppDomainFixedOutput] (Will always be turned into a literal).
> 
> Actually there should be made a unit-test like this:
> 
>         [Fact]
>         public void IRawValueRenderer_AppDomainFixedOutput_Attribute_NotRequired()
>         {
>             foreach (Type type in allTypes)
>             {
>                 if (typeof(NLog.Internal.IRawValue).IsAssignableFrom(type) && !type.IsInterface)
>                 {
>                     var appDomainFixedOutputAttribute = type.GetCustomAttribute<AppDomainFixedOutputAttribute>();
>                     Assert.True(ReferenceEquals(appDomainFixedOutputAttribute, null), $"{type.ToString()} should not implement IStringValueRenderer");
>                 }
>             }
>         }
>

PR #3993, opened 2020-05-30
Don't combine IRawValue with [AppDomainFixed]

304NotModified:
Fixes https://github.com/NLog/NLog/issues/3992


304NotModified:
Another option is to make ConvertToLiteral better, and support for these cases IRawValue

snakefoot:
> Another option is to make ConvertToLiteral better, and support for these cases IRawValue. For example, create LiteralWithRawValueLayoutRenderer for these cases.

Yes when only having a single LayoutRenderer in a SimpleLayout, then one could change it so SimpleLayout becomes a FixedString but with RawValue support.

PR #3997, opened 2020-05-30
Support RawValue with [AppDomainFixedOutput]

### LINKED ISSUE #3992: Don't combine IRawValue with [AppDomainFixed]

304NotModified:
- plus some small code improvements in the file (Boy Scout Rule) 

Fixes #3992, alternative to https://github.com/NLog/NLog/pull/3993

### DIFF
--- a/src/NLog/Layouts/LayoutParser.cs
+++ b/src/NLog/Layouts/LayoutParser.cs
@@ -506,6 +506,12 @@ namespace NLog.Layouts
                 if (list[i] is LiteralLayoutRenderer lr1 && list[i + 1] is LiteralLayoutRenderer lr2)
                 {
                     lr1.Text += lr2.Text;
+
+                    // Combined literals don't support rawValue
+                    if (lr1 is LiteralWithRawValueLayoutRenderer lr1WithRaw)
+                    {
+                        list[i] = new LiteralLayoutRenderer(lr1WithRaw.Text);
+                    }
                     list.RemoveAt(i + 1);
                 }
                 else
--- a/src/NLog/Layouts/LayoutParser.cs
+++ b/src/NLog/Layouts/LayoutParser.cs
@@ -86,7 +86,7 @@ namespace NLog.Layouts
 
                     if (EndOfLayout(ch))
                     {
-                        //end of innerlayout. 
+                        //end of inner layout. 
                         // `}` is when double nested inner layout. 
                         // `:` when single nested layout
                         break;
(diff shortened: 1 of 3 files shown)


snakefoot:
Think this will break the whole literal optimization as merging stops working along with SimpleLayout fixed text.


Think you should just extend the existing LiteralLayoutRenderer and when merging two into one then reset raw value and mark it as raw value not available.


Sent from my Sony Xperia


@304NotModified<https://github.com/304NotModified> requested your review on: #3997<https://github.com/NLog/NLog/pull/3997> Improve ConvertToLiteral to handle RawValues.

304NotModified:
Good one indeed!

304NotModified:
@snakefoot all fixed now, please check :)

src/NLog/Layouts/LayoutParser.cs
```diff
@@ -494,13 +499,11 @@ private static bool CanBeConvertedToLiteral(LayoutRenderer lr)
...
+        private static void MergeLiterals(IList<LayoutRenderer> list)
         {
             for (int i = 0; i + 1 < list.Count;)
             {
-                var lr1 = list[i] as LiteralLayoutRenderer;
-                var lr2 = list[i + 1] as LiteralLayoutRenderer;
-                if (lr1 != null && lr2 != null)
+                if (list[i] is LiteralLayoutRenderer lr1 && list[i + 1] is LiteralLayoutRenderer lr2)
                 {
                     lr1.Text += lr2.Text;
```
snakefoot, line 508: You forgot to reset the ability to return raw-value when merging two literals into one.

304NotModified: Well when combining two you haven't a raw value? 

Dunno which test case I'm missing then. (I solved this TDD style)

snakefoot: What raw value does this Layout return: `${processid}-${processname}` ?

304NotModified: Fixed!

src/NLog/Layouts/LayoutParser.cs
```diff
@@ -512,9 +521,17 @@ private static void MergeLiterals(List<LayoutRenderer> list)
             }
         }
 
-        private static LayoutRenderer ConvertToLiteral(LayoutRenderer renderer)
+        private static LayoutRenderer ConvertToLiteral(IRenderable renderer)
```
snakefoot, line 524: Think it should stay as type `LayoutRenderer`. Less confusing.

src/NLog/Layouts/LayoutParser.cs
```diff
@@ -512,9 +521,17 @@ private static void MergeLiterals(List<LayoutRenderer> list)
...
         {
-            return new LiteralLayoutRenderer(renderer.Render(LogEventInfo.CreateNullEvent()));
+            var logEventInfo = LogEventInfo.CreateNullEvent();
+            var text = renderer.Render(logEventInfo);
+            if (renderer is IRawValue rawValueRender)
+            {
+                var success = rawValueRender.TryGetRawValue(logEventInfo, out var rawValue);
```
snakefoot, line 530: When not a success, then no need to change into `LiteralWithRawValueLayoutRenderer`

304NotModified: Well less tricky to keep it like this.

304NotModified: close as this is by design

src/NLog/Layouts/LayoutParser.cs
304NotModified: not really an opinion on this, so keep it LayoutRenderer

304NotModified:
> But looking good. Could consider re-base to master for ver. 4.7.3, since the change is actually very discrete.

I think we should only do bug fixes on 4.7.x for now

PR #3999, opened 2020-06-01
LogFactory Shutdown is public so it can be used from NLogLoggerProvider

snakefoot:
Allows this TODO to be fixed:

https://github.com/NLog/NLog.Extensions.Logging/blob/3c64a51328ace81efa2a6a0696926a29a53a7261/src/NLog.Extensions.Logging/Logging/NLogLoggerProvider.cs#L87

And also make NLog work better with isolated LogFactory-instances.

### DIFF
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -998,7 +998,10 @@ namespace NLog
             }
         }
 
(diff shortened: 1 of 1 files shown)


304NotModified:
Thanks!

PR #4005, opened 2020-06-07
LoggingConfiguration - AddRuleForBlackHole

snakefoot:
See also: https://stackoverflow.com/questions/62247678/how-do-i-configure-nlog-in-code-in-a-net-core-3-web-app-to-match-the-microsoft


304NotModified:
I was thinking to allow `null`  as a target

snakefoot:
Think this is easier to understand. Also for people who don't know about the null-value-hack.


Remember the original wiki by you also believed one should use a NullTarget


Sent from my Sony Xperia


I was thinking to allow null as a target

snakefoot:
Superseeded by #4128

PR #4017, opened 2020-06-17
Allow to change the RuleName of a LoggingRule

304NotModified:
Small enhancement for NLog 4.7.3

So we could 

- rename one
- create easier an instance

    e.g. do this:

    ```
    var target = new FileTarget("t1");
    var loggingRule = new LoggingRule("MyNamespace.*", LogLevel.Info, LogLevel.Error, target);
    loggingRule.RuleName = "name1";
    ```

    instead of
    ```
    var loggingRule = new LoggingRule("name1");
    loggingRule.EnableLoggingForLevels(LogLevel.Info, LogLevel.Error);
    loggingRule.Targets.Add(target);
    loggingRule.LoggerNamePattern = "MyNamespace.*";
    ```

### DIFF
--- a/src/NLog/Config/LoggingRule.cs
+++ b/src/NLog/Config/LoggingRule.cs
@@ -114,7 +114,7 @@ namespace NLog.Config
         /// <summary>
         /// Rule identifier to allow rule lookup
         /// </summary>
-        public string RuleName { get; }
+        public string RuleName { get; set; }
 
         /// <summary>
         /// Gets a collection of targets that should be written to when this rule matches.
--- a/tests/NLog.UnitTests/Config/ConfigApiTests.cs
+++ b/tests/NLog.UnitTests/Config/ConfigApiTests.cs
@@ -50,7 +50,7 @@ namespace NLog.UnitTests.Config
         public void AddTarget_testname()
         {
             var config = new LoggingConfiguration();
-            config.AddTarget("name1", new FileTarget {Name = "File"});
+            config.AddTarget("name1", new FileTarget { Name = "File" });
             var allTargets = config.AllTargets;
             Assert.NotNull(allTargets);
             Assert.Single(allTargets);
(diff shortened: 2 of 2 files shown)


304NotModified:
But maybe it's an issue as after reload (XML config) the rename is lost?

snakefoot:
I see the RuleName as a RuleId. And it should remain static throughout the object-lifetime.

Just like Target-Name-setter should explode in a rain-of-exceptions when trying to modify the Target-Name after having added to LoggingConfiguration. But throwing exceptions when calling setters is usually not nice, but dictionaries for lookup will not reflect the name-change.

If there is missing a constructor or fluent-api for setting up LoggingRules, then that should be implemented for NLog 5. Instead of opening of for public-setters and handling name-collisions etc.

304NotModified:
> And also allow dictionary-like lookup of LoggingRule on RuleName. But if suddenly allowing changes to LoggingRule.RuleName then this validation/lookup is very hard.

I'm aware of that, but I checked the code and there isn't a dictionary lookup? It's just a foreach loop.

snakefoot:
> I'm aware of that, but I checked the code and there isn't a dictionary lookup? It's just a foreach loop.

Because it was added at a late stage, so it had to handle unnamed rules. But could be nice with not adding duplicate rules (Maybe for NLog 5),  and enforcing that is hard when having a public-setter.

304NotModified:
Funny, i'm fine with non uniqueness. I don't see things break if we allow that. (as we don't use the name for anything functional)

snakefoot:
`FindRuleByName` kind of expects uniqueness. RuleName was just added in best-effort-style, because no-one else had the guts. And wanted to see in what direction the ball went. But making it more fragile with more unexpected behavior doesn't seem like a good direction.

304NotModified:
I don't see it like that. The name is optional and should not be used by NLog itself. So nothing goes wrong - it's just meta data. I think we should keep that, agree?

Maybe `FindRuleByName` should be renamed to FindFirstRuleByName in the future.

snakefoot:
> should not be used by NLog itself. 

Then please remove the methods FindRuleByName and RemoveRuleByName. If you say RuleName is not used by NLog.

NLog is a framework. This means its methods should be predictable and easy to use. Not having traps and pitfalls.

snakefoot:
Anyway just merge it. Not that important. Since I'm not gonna use it.

PR #4021, opened 2020-06-20
ObjectReflectionCache - Reduce initial memory allocation until needed

snakefoot:
Skips allocation of 300 KByte upfront. When actually being used, then it will  use start-capacity of 75 KByte (Just below LargeObjectHeap-treshold) and allow doubling capacity 2 times (150 KByte -> 300 KByte).

Before the initial memory allocation was only done when making use of structured-logging. But with #3610 this was changed to always (NLog 4.7)


snakefoot:
Closing and reopening to trigger rebuild.

304NotModified:
(build green on Travis)

304NotModified:
Thanks!

PR #4023, opened 2020-06-20
LoggingConfiguration with support for automatic target re-initialize

snakefoot:
Allow loading target from `NLog.config` that fails on InitializeTarget because of missing service-dependency. Without exploding in your face even with `ThrowConfigExceptions=true`.

When the missing service-dependency arrives in the ServiceRepostory, then it will automatically retry initialize of the target.

This should handle the case with Microsoft ApplicationInsight-Target where `TelemetryConfiguration.Active` is obsolete, and should now be acquired through dependency-injection.


304NotModified:
PS: build is green on Travis, reporting broken (again)

image

304NotModified:
cool! Thanks!

PR #4039, opened 2020-07-07
Removed obsolete DictionaryAdapter after LogEventInfo.Context was removed

snakefoot:
See also #3429


304NotModified:
Cool! Less is more :)

PR #4043, opened 2020-07-09
ObjectReflectionCache - Skip reflection for Stream objects

snakefoot:
Hard to serialize something useful from a Stream-object. Instead their properties often throw exceptions.


304NotModified:
Thanks! Good idea!

Issue #1148, opened 2016-01-10
Split code in more packages 


304NotModified:
+namespaces

304NotModified:
Packages:

Core (logging interface + config) 
Internal (internal logger, helpers) 
Xmlconfig 
Network (targets, LayoutRenderers) 
Classis asp
Webservices (WCF, rest) 
Database

304NotModified:
Silverlight renderers

Abstractions (interfaces)

snakefoot:
Should be resolved with the following issues:

#2582 - Move WCF targets to NLog.WCF package
#2676 - Move DatabaseTarget to NLog.DatabaseTarget package
#2680 - Move NetworkTarget to NLog.NetworkTarget package
#2786 - Move MailTarget to NLog.SmtpMailTarget package
#2787 - Move System.Configuration to NLog.AppSettings package

See also #268

304NotModified:
I think this should be 5.0, as 5.0 is later then planned IMO

304NotModified:
extensions packages should target NLog.Core then (or maybe NLog.Abstractions)

304NotModified:
> #2582 - Move WCF targets to NLog.WCF package
> #2676 - Move DatabaseTarget to NLog.DatabaseTarget package
> #2680 - Move NetworkTarget to NLog.NetworkTarget package
> #2786 - Move MailTarget to NLog.MailTarget package

Proposal: Nlog.Database, NLog.Network (including layout renderers), NLog.Mail

snakefoot:
Also good.

304NotModified:
@snakefoot what do you think the benefits should be of this change?

At first it was improving dependencies, but that isn't an issue with .NET standard 2 anymore (only one dependency). (of course we more platforms)

IMO the big improvement are

- smaller code base - so easier to maintain
- abstractions in separate package for custom implementations
- smaller build times
- maybe also:  changing a target won't break the core

But with this in mind, maybe another split would be better? (eg. no layout renderers in the core package).

What do you think?

snakefoot:
My focus is on limited NetCore platforms and compilation with NetCore Native. 

Reducing the amount of direct references from NLog.dll to system-libraries. Because with NetCore Native it will try to shake everything away not being used.

Must admit one of the strengths of NLog.dll is actually the release of unified NLog.dll where everything is in "sync". No need to chase around to find the perfect console-logger or file-logger.

[...]

PR #4044, opened 2020-07-12
Introduced ILog to replace ILogger. Introduced LogEventBuilder to replace LogBuilder

### LINKED ISSUE #532: Callsite Layout Renderer and Caller Attributes

snakefoot:
Right now `ILogger` has over 300 interface-members (Most are from NLog 1.0 before generics was invented). The new `ILog` has 25 interface-members. I guess one could reduce the number even further, but like that one can easily be notified on error-events when mocking.

It would be a major breaking change to make any changes to `ILogger`. Even adding a new interface-method like `WithProperty` and `PushScopeContext`.

Now marked `ILogger` as obsolete, and instead redirected people to `ILog`. The interfaces has same features, except that `ILog` has reduced the number of methods that takes `IFormatProvider`-input-parameter and removed all non-generic-overrides. The methods with `IFormatProvider`-input-parameter has been moved into LoggerExtensions-methods.

Have at the same time added several extension-methods to `ILog` so it has similar capabilities like actual `Logger`:
- Supports LogMessageGenerator
- Supports Swallow
- Supports SwallowAsync
- Supports ConditionalDebug
- Supports ConditionalTrace

One could consider in NLog 6.0 to remove `ILogger`-interface (and `ILoggerBase`) and then reintroduce it again where inheriting from `ILog`.

Have at the same time obsoleted the old fluent `LogBuilder` and introduced `LogEventBuilder` that is available without needing to write `using NLog.Fluent`. The interfaces are very similar, but the start method is a little different:

- **LogBuilder:** `logger.Info().Message("Hello").Write();`
- **LogEventBuilder:** `logger.ForInfoEvent().Message("Hello").Write();`

I guess one could change `LogEventBuilder` to be activated from using `Info()` but it would generate more noise about ambiguous extensions-methods. But actually I like `ForInfoEvent()` since it signals that one is starting to define a LogEvent.

Also unsure if I like the `Write()`-method. Think I prefer `.Log()` but then the transformation would not be that easy:

- **LogEventBuilder:** `logger.ForInfoEvent().Message("Hello").Log();`

The major improvement in `LogEventBuilder` is there is no allocation if the LogLevel is not enabled. And when the LogLevel is enabled, then it no longer allocates an additional `LogBuilder`-object (Because `LogEventBuilder` is a struct on the stack). Resolves #532


snakefoot:
Need to improve on the test-coverage with all the new LoggerExtensions-methods.

Also has some TODOs regarding improving single-parameter-logging, where the parameter is an Exception-object (Enable ExceptionMessageFormatProvider)

snakefoot:
Might consider adding caller-member-attributes-parameters to these ILog-extension-methods:

```c#
public static void Trace([NotNull] this ILog logger, [NotNull] LogMessageGenerator messageFunc)
public static void Trace([NotNull] this ILog logger, Exception exception, [NotNull] LogMessageGenerator messageFunc)
```

So it becomes:

```c#
public static void Trace([NotNull] this ILog logger, [NotNull] LogMessageGenerator messageFunc, 
                        [CallerMemberName]string callerMemberName = null,
                        [CallerFilePath]string callerFilePath = null,
                        [CallerLineNumber]int callerLineNumber = 0)
public static void Trace([NotNull] this ILog logger, Exception exception, [NotNull] LogMessageGenerator messageFunc
                        [CallerMemberName]string callerMemberName = null,
                        [CallerFilePath]string callerFilePath = null,
                        [CallerLineNumber]int callerLineNumber = 0)
```

Then it will also get the benefit of faster callsite-logic.

304NotModified:
I'm doubting if we should not wait for default interface implementations.

304NotModified:
I like the change, but I am very reluctant to change the logging API. There should be a big benefit. Sure, 25 interface is much better than 300, but I think #3656 , #1148  , #3675  and #1234 are far more important

snakefoot:
Yes this interface change is a preparation for doing default implementation with netstandard21


Sent from my Sony Xperia


I'm doubting if we should not wait for default interface implementations.

snakefoot:
All of those PullRequests are important, but any heavy refactoring to split NLog into multiple targets should wait for the `Layout<T>` to be committed. 

Also quite happy with the current state of DependencyInjection-support. There is a huge task with refactoring the NLog-UnitTests to stop using the global NLog.LogManager and instead use isolated LogFactory. Making it clear that isolated LogFactory is fully working.

I just saw that `ILog` was an isolated area that didn't overlap with those important PullRequests (Little change of merge conflicts). Besides putting a strain of the critical time-resource.

304NotModified:
> There is a huge task with refactoring the NLog-UnitTests to stop using the global NLog.LogManager and instead use isolated LogFactory. Making it clear that isolated LogFactory is fully working.

Yes this would be great. Also for performance and parallel testing. It started with this once, but replaced to much (so broke things). I think we should a few test classes per pr. Maybe we should mark then somehow that they're done

snakefoot:
Alternative interface than ILog could be `INLog` or `INLogger`, but maybe the name-collission with other Logger-interface is wanted.

snakefoot:
Given up on replacing `ILogger` and instead created #432

PR #4046, opened 2020-07-14
Set properties of ProcessStartInfo correctly

chucklu:
According the ProcessStartInfo.UseShellExecute property [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo.useshellexecute?view=netcore-3.1#remarks) 
Setting this property to false enables you to redirect input, output, and error streams.
**If you set the WindowStyle to ProcessWindowStyle.Hidden, UseShellExecute must be set to true.**

I think you would like to set UseShellExecute  to false to enable the redirect function. So, I removed code about set the [ProcessStartInfo.WindowStyle](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo.createnowindow?view=netcore-3.1) property.

### DIFF
--- a/tests/NLog.UnitTests/ProcessRunner.cs
+++ b/tests/NLog.UnitTests/ProcessRunner.cs
@@ -126,11 +126,9 @@ class C1
             proc.StartInfo.FileName = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Runner.exe");
 #endif
             proc.StartInfo.UseShellExecute = false;
-            proc.StartInfo.WindowStyle = ProcessWindowStyle.Normal;
-            proc.StartInfo.WindowStyle = ProcessWindowStyle.Hidden;
             proc.StartInfo.CreateNoWindow = true;
             // Hint:
-            // In case we wanna redirect stdout we should drain the redirected pipe continously.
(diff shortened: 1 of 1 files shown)


304NotModified:
Hi!

Do you think the old config would make the unit test unstable? (as this is only in the unit test code)

chucklu:
@304NotModified No, the test is stable. What I am concerned is, the code should be correct as the official document said.
Might somebody else will learn the bad code from this project.
There are lots of wrong usage on internet, such as https://stackoverflow.com/questions/1369236/how-to-run-console-application-from-windows-service/1369252#1369252, I did not have enough permission to add a comment for it.
https://stackoverflow.com/questions/1369236/how-to-run-console-application-from-windows-service/1369385#1369385

The point is, I would like to fix the wrong usage which violate "`If you set the WindowStyle to ProcessWindowStyle.Hidden, UseShellExecute must be set to true.`"

chucklu:
Although the test code is runnable, as the saying goes, 'Just because you can doesn't mean you should'.

304NotModified:
> Might somebody else will learn the bad code from this project.

I agree we don't want that :)

PR #4053, opened 2020-07-22
MessageTemplates - Changed Literal.Skip to be Int32 for long message templates

### LINKED ISSUE #4052: System.ArgumentOutOfRangeException when logging a message with unused parameters
I am running NLog Version 4.6.6, .NET Framework 4.7.1 on VS2017.
When logging a message which exceeds `short.MaxValue` characters might result in an `ArgumentOutOfRangeException`.
Try this:
```c#
public static void Bumm()
{
	string char32 = "ABCDEFÖ8901234567890123456789012";
	string a = "";
	for (int i = 0; i < 1023 * 1; i++)
	{
		a += char32;
	}
	var myLogger = NLog.LogManager.GetLogger("myLogger");
	LogEventInfo theEvent = new LogEventInfo(
		LogLevel.Debug, "muh", CultureInfo.InvariantCulture, a, new object[] { 42 });
	myLogger.Log(theEvent);  // OK (length < short.MaxValue)
	a += char32;
	theEvent = new LogEventInfo(
		LogLevel.Debug, "muh", CultureInfo.InvariantCulture, a, new object[] { 42 });
	myLogger.Log(theEvent); // Exception length = short.MaxValue + 1
}
```
The reason is an overflow when converting an `int32` (returned by `SkipUntil`) to an int16 (`Literal.Skip`) in the `TemplateEnumerator.ParseTextPart`.

What is the preferred way of fixing this?

- Change the `Literal.Skip` from `short` to `int`
- Ensure that no overflow happens when assigning the Skip Property (more or less: `(short)Math.Min(short.MaxValue, SkipUntil(...)`)
It depends on what the intention was when defining the `Skip` as a short in the firstplace.
If I read the history correct, the Skip and Print both have been `ushort` in the first place. Someone has changed `Print` to an `int32` but the sense of changing the position from `ushort` to `short` I do not get.

@snakefoot @304NotModified do you have any concerns when changing the type of `Skip` to an `int32`?

snakefoot:
Support message templates longer than short.MaxValue. Resolves #4052

Think the `short`-logic was a memory optimization, but since modern CPUs really like 32 bit (or even better 64 bit numbers), then I actually think it is a de-optimization to use `short`.

### DIFF
--- a/src/NLog/MessageTemplates/Literal.cs
+++ b/src/NLog/MessageTemplates/Literal.cs
@@ -43,6 +43,6 @@ namespace NLog.MessageTemplates
         public int Print;
         /// <summary>Number of characters to skip in the original template at the current position.</summary>
         /// <remarks>0 is a special value that mean: 1 escaped char, no hole. It can also happen last when the template ends with a literal.</remarks>
-        public short Skip;
+        public int Skip;
     }
 }
--- a/src/NLog/MessageTemplates/TemplateEnumerator.cs
+++ b/src/NLog/MessageTemplates/TemplateEnumerator.cs
@@ -140,7 +140,7 @@ namespace NLog.MessageTemplates
 
         private void ParseTextPart()
         {
-            _literalLength = (short)SkipUntil(TextDelimiters, required: false);
+            _literalLength = SkipUntil(TextDelimiters, required: false);
         }
 
         private void ParseOpenBracketPart()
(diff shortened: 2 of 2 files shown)


Oberndorfer:
@snakefoot please also change it in the `ParseTextPart` method (this is actually causing my problem).

snakefoot:
> please also change it in the `ParseTextPart` method (this is actually causing my problem).

Updated the PR. Thanks for noticing.

304NotModified:
Thanks!

I think a log message of that long is a bit strange, but OK, we should not crash on it.

PR #2989, opened 2018-11-03
ObjectReflectionCache - Extracted from JsonSerializer for use in XmlLayout

snakefoot:
Caches the TypeCode on initial reflection to enable faster serialization.

Improves support for objects implementing IDictionary<string,object>-interface (But not System.Collections.Dictionary. Ex Expando-objects).

Adds "Type" as implicit property for Exception-objects. See also https://github.com/NLog/NLog/issues/1311#issuecomment-435352148

PR #4057, opened 2020-07-24
ObjectReflectionCache - Reduce noise from properties that throws exceptions

snakefoot:
See also: https://stackoverflow.com/q/63064051/193178

The extra noise was introduced with #2989 (Before exceptions from properties was completely suppressed).


304NotModified:
Good idea!

PR #4058, opened 2020-07-24
FileTarget - Skip delegate capture in GetFileCreationTimeSource

snakefoot:
Skip unnecessary allocation for every logevent when using `archiveEvery="Day"`

Fallback only necessary when appender has been closed (Because NLog caches the creation-time on the appender).


304NotModified:
thanks!

PR #4066, opened 2020-08-02
Json serializer should always include decimal point to signal datatype

### LINKED ISSUE #4061: Incorrect serialization of float and double values in structured logging
Hello.
I faced with issue when using structured logging and trying to log some float or double values which have no fractional part. In this case they are serialized as integer.

For example: 
`float floatValue = 0.0f;
_log.WithProperty("myFloat", floatValue).Fatal("something");`

It's serialized as 0 instead of 0.0.

Is there some way of fixing this issue?

**NLog version**: (e.g. 4.7.2)

**Platform**:  .NET Core 3.1

**Current NLog config** 

```xml
<nlog
	xmlns="http://www.nlog-project.org/schemas/NLog.xsd"
	autoReload="true" throwConfigExceptions="true">

	<targets async="true">
		<target name="Console" xsi:type="Console" >
			<layout type="JsonLayout" renderEmptyObject="false">
				<attribute name="properties" encode="false">
					<layout type='JsonLayout' includeAllProperties="true" renderEmptyObject="false" MaxRecursionLimit="1"/>
				</attribute>
			</layout>
		</target>
	</targets>

	<rules>
		<logger name="*" writeTo="Console" minlevel="Info" />
	</rules>
</nlog>

```

snakefoot:
Resolves #4061

The validation of decimal point gives a performance hit, but have also added a performance optimization to skip string-allocation when decimal-datatype can be converted to integer without precision-loss (Ex. `0.0`)


snakefoot:
Some benchmarks numbers for doing `log.Info("{@Id} {@Age}", new { Value = 1D }, new { Value = 1D });`

| Value = 1D        | Time (ms) | Msgs/sec  | GC2 | GC1 | GC0 | CPU (ms) | Alloc (MB) |
| PR        |     8.106 |   616.821 |   0 |   1 | 210 |    8.296 |    3.051,8 |
| Master        |     9.604 |   520.562 |   0 |   1 | 223 |    9.687 |    3.242,6 |

| Value = 1000000D | Time (ms) | Msgs/sec  | GC2 | GC1 | GC0 | CPU (ms) | Alloc (MB) |
| PR        |     8.490 |   588.875 |   0 |   1 | 218 |    8.515 |    3.166,3 |
| Master        |     10.110 |   494.553 |   0 |   1 | 241 |   10.250 |    3.509,6 |

| Value = 1.1D        | Time (ms) | Msgs/sec  | GC2 | GC1 | GC0 | CPU (ms) | Alloc (MB) |
| PR         |     10.223 |   489.070 |   0 |   1 | 231 |   10.281 |    3.357,0 |
| Master        |     9.911 |   504.457 |   0 |   1 | 231 |    9.984 |    3.357,0 |

| Value = 31415926535897.9D       | Time (ms) | Msgs/sec  | GC2 | GC1 | GC0 | CPU (ms) | Alloc (MB) |
| PR        |    11.268 |   443.723 |   0 |   1 | 263 |   11.500 |    3.814,8 |
| Master        |     10.778 |   463.900 |   0 |   1 | 262 |   10.859 |    3.814,8 |

304NotModified:
Thanks for the PR and benchmark!

I haven't merged this one yet, as Im afraid of the consequences of this behavior change. IMO it's was a bug (it wasn't wrong but it wasn't perfect), so this change could breaks others unexpectedly

snakefoot:
Think it is a good change. Not so worried about breaking changes, since Json is a very forgiving format.


Sent from my Sony Xperia


Thanks for the PR and benchmark!

I haven't merged this one yet, as Im afraid of the consequences of this behavior change. IMO it's was a bug (it wasn't wrong but it wasn't perfect), so this change could breaks others unexpectedly

304NotModified:
I don't think that parsers will accept 0.0 if it's configured as int. (ok that's a bug in your config)

 I don't feel confident to merge this for a patch release. Maybe bump to 4.8

snakefoot:
Yes I guess when having a double that almost always converts to value without decimal, then it would be be indexed as integer in ElasticSearch.

When the default is `integer`-like values, then the inconsistency will only cause warnings once every hour when receiving a value with decimal-point.

This PR will suddenly make ElasticSearch generate warnings constantly, because now it will receive datatype decimal. Where it often would get `integer` like values.

snakefoot:
Random article about the painful adventures when the dynamic indexing is not working (Because float is confused as long):

http://www.dissmeyer.com/2020/02/13/index-templates-in-the-elastic-stack/

304NotModified:
mmm maybe this should be 5.0

304NotModified:
@snakefoot are you OK with nlog 5 for this?

snakefoot:
Yes it is fine with me.


Sent from my Sony Xperia


@snakefoot<https://github.com/snakefoot> are you OK with nlog 5 for this?

PR #4069, opened 2020-08-03
Fluent LogBuilder should suppress exception on invalid callerFilePath

### LINKED ISSUE #4067: LogBuilder from Fluent throws exception when used wrongly
**NLog version**: 4.7.2

**Platform**: .Net 4.0

The fluent API can be misused, and in that case it throws an exception. For example, I (wrongly) assumed that I should put the log text immediately in Info:

```csharp
Log.Info(jsonText);
```

I didn't read up on the documentation properly. But, this now failed with an exception:

```txt
System.ArgumentException: Illegal characters in path.
   at System.IO.Path.CheckInvalidPathChars(String path, Boolean checkAdditional)
   at System.IO.Path.GetFileName(String path)
   at System.IO.Path.GetFileNameWithoutExtension(String path)
   at NLog.Fluent.Log.Create(LogLevel logLevel, String callerFilePath)
```

I'm coming from log4net where exceptions are never thrown by the logging framework. This lack of error handling takes away quite a bit from NLog, unfortunately (IMO of course).
I can think of a simple error handling in this case. If the callerFilePath can't be parsed, use a default or omit entirely. But never throw an exception. Please 🙇‍♂️

snakefoot:
Resolves #4067 

Trying to make pretty logger-name based on `callerFilePath` should not lead to exception if path-value is invalid on the platform.

### DIFF
--- a/src/NLog/Fluent/Log.cs
+++ b/src/NLog/Fluent/Log.cs
@@ -30,13 +30,15 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF 
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
-using System;
-using System.IO;
-using System.Runtime.CompilerServices;
 
 namespace NLog.Fluent
 {
 #if NET4_5
+    using System;
+    using System.IO;
+    using System.Runtime.CompilerServices;
+    using NLog.Common;
+
     /// <summary>
     /// A global logging class using caller info to find the logger.
     /// </summary>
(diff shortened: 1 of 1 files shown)


src/NLog/Fluent/Log.cs
```diff
@@ -117,15 +117,26 @@ public static LogBuilder Fatal([CallerFilePath]string callerFilePath = null)
...
+        private static ILogger GetLogger(string callerFilePath)
+        {
+            try
+            {
+                string name = Path.GetFileNameWithoutExtension(callerFilePath ?? string.Empty);
+                return string.IsNullOrWhiteSpace(name) ? _logger : LogManager.GetLogger(name);
+            }
+            catch
+            {
+                return _logger;
```
304NotModified, line 137: Maybe an internal log (trace or debug)?

snakefoot: Fixed to match `InternalLogger.Warn(ex, "Error when formatting a message.");`

304NotModified:
Thanks!

PR #4073, opened 2020-08-06
FileTarget - Extra validation of the LogEvent-timestamp before checking time to archive

### LINKED ISSUE #4072: Why is Nlog archiving files every few minutes?
This is a continuation of the issue posted on Stack Overflow here https://stackoverflow.com/questions/62899627/why-is-nlog-archiving-files-every-few-minutes Moved discussion here so that more information can be provided.

We are using NLog 4.7.2 in an ASP.NET WebAPI app running on Windows Server 2016 under full .NET Framework v4.8. I have configured the archive settings in a way that I think is supposed to roll the logs slowly throughout the day when it hits 1GB in size. At our load this should produce maybe 1 or two log files. Instead I see log files archiving rapidly at unpredictable times 300 to 600 times a day, sometimes every few minutes, with unpredictable sizes ranging from 1 KB to 32,917 KB in size.

Using the settings below I look in our logging folder and see hundreds of files:

**NLog version**:  4.7.2
**Platform**: .Net 4.5
```xml
<?xml version="1.0"?>
      autoReload="true"
      internalLogLevel="Info"
      internalLogFile="C:\Logs\nlog-internal.log">

  <extensions>
    <add assembly="MyApp.Common.Web"/> <!-- adds the context-properties renderer -->
  </extensions>

  <targets async="true">
    <target
      name="logFile"
      xsi:type="File"
      layout="${longdate:universalTime=true}Z ${pad:padding=-7:inner=[${level:uppercase=true}]} [${machinename}] ${pad:padding=-5:inner=[${threadid}]} [${logger}] [${context-properties}] ${message} ${onexception:${exception:format=ToString}}"
      fileName="C:\Logs\My-Log-File.log"
      maxArchiveDays="2"
      archiveAboveSize="1000000000"
      archiveFileName="C:\Logs\My-Log-File.{#}.log"
      archiveEvery="Day"
      archiveDateFormat="yyyyMMdd"
      concurrentWrites="false"
      encoding="utf-8" />
  </targets>

  <rules>
    <!-- many rules here (omitted for clarity) --> 

    <logger name="*" minlevel="Info" writeTo="logFile" />
  </rules>

</nlog>
```

Current result - Files are getting archived every few minutes.
Expected result - Files should be archived according to size or time parameters instead of seemingly randomly.

Here is an example of the internal log at Debug level from the pre-release version Rolf Kristensen provided for me

[internal-log.txt](https://github.com/NLog/NLog/files/5029523/internal-log.txt)

snakefoot:
Very crude validation to a very special issue. Trying to resolve #4072

### DIFF
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -2063,8 +2063,9 @@ namespace NLog.Targets
             if (fileCreateTime != logEventTime)
             {
                 string formatString = GetArchiveDateFormatString(string.Empty);
+                var validLogEventTime = EnsureValidLogEventTimeStamp(logEvent.TimeStamp, creationTimeSource.Value);
                 string fileCreated = creationTimeSource.Value.ToString(formatString, CultureInfo.InvariantCulture);
-                string logEventRecorded = logEvent.TimeStamp.ToString(formatString, CultureInfo.InvariantCulture);
+                string logEventRecorded = validLogEventTime.ToString(formatString, CultureInfo.InvariantCulture);
                 var shouldArchive = fileCreated != logEventRecorded;
                 if (shouldArchive)
                 {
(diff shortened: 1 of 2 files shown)


src/NLog/Targets/FileTarget.cs
```diff
@@ -2298,12 +2299,32 @@ private DateTime InitializeFile(string fileName, LogEventInfo logEvent)
...
+                now = EnsureValidLogEventTimeStamp(now, lastTime);
                 _initializedFiles[fileName] = now;
             }
 
             return lastTime;
         }
 
+        DateTime EnsureValidLogEventTimeStamp(DateTime logEventTimeStamp, DateTime previousTimeStamp)
+        {
+            if (logEventTimeStamp < previousTimeStamp && logEventTimeStamp.Date < previousTimeStamp.Date)
```
304NotModified, line 2311: when only comparing dates `if(logEventTimeStamp.Date < previousTimeStamp.Date)`, doesn't give that the same results

snakefoot: Yes but `DateTime.Date` is a "slow" method. And by default then `logEventTimeStamp < previousTimeStamp` is never true. Thus skip calling the slow method.

304NotModified: Ok please add a code comment for that, than I will merge it :)

snakefoot: Added comment and added unit-test.

PR #4076, opened 2020-08-08
DatabaseTarget - Added AllowDbNull for easier support for nullable pa…

### LINKED ISSUE #4070: DB null layout render doesn't work in whenEmpty
**NLog version**: 4.7.2
**Platform**: .Net 4.8
**Current NLog config**

```xml
<nlog internalLogLevel="Warn"
  internalLogFile="C:\Temp\NLog.Internal.txt"
  internalLogToConsole="true"
  autoReload="true" 
  throwExceptions="true"
  xmlns="http://www.nlog-project.org/schemas/NLog.xsd"
  <targets async="true">
    <target xsi:type="Database" name="DbDefault"
            connectionStringName = "frameworkLogging"
            commandText = "INSERT INTO Logs ([Date], [Level], [Logger], [Message], [Exception], [LogGuid], [Username], [Source], [LogGroupingIdentifier]) VALUES (@log_date, @level, @logger, @message, @exception, @logGuid, @username, @source, @logGroupingIdentifier)" keepConnection="true">
      ...
      <parameter name="@exception" layout="${exception:format=tostring:whenEmpty=${db-null}}" />
      ...
    </target>
    ...
</nlog>
```

- What is the current result?
When exception is empty/null, an empty string.
When exception is empty/null, a NULL value is written to the database.
trace: `2020-08-04 10:27:20.6864 Trace   DatabaseTarget: Parameter: '@exception' = '' (String)`
Not that I know of
Not sure

snakefoot:
Instead of doing this:

`<parameter name="@exception" layout="${exception:format=tostring:whenEmpty=${db-null}}" parameterType="System.Object" 
/>`
`<parameter name="@correlationid" layout="${activityid:whenEmpty=${db-null}}" dbType="DbType.Guid" parameterType="System.Object" />`

Then one can do this:

`<parameter name="@exception" layout="${exception:format=tostring}" allowDbNull="true" />`
`<parameter name="@exception" layout="${activityid}" dbType="DbType.Guid" allowDbNull="true" />`

Trying to resolve #4070

### DIFF
--- a/src/NLog/Targets/DatabaseParameterInfo.cs
+++ b/src/NLog/Targets/DatabaseParameterInfo.cs
@@ -134,6 +134,13 @@ namespace NLog.Targets
         [DefaultValue(null)]
         public CultureInfo Culture { get; set; }
 
+        /// <summary>
+        /// Gets or sets whether empty value should translate into DbNull. Requires database column to allow NULL values.
+        /// </summary>
+        /// <docgen category='Parameter Options' order='8' />
+        [DefaultValue(false)]
+        public bool AllowDbNull { get; set; }
+
         internal bool SetDbType(IDbDataParameter dbParameter)
         {
             if (!string.IsNullOrEmpty(DbType))
--- a/tests/NLog.UnitTests/Targets/DatabaseTargetTests.cs
+++ b/tests/NLog.UnitTests/Targets/DatabaseTargetTests.cs
@@ -607,18 +607,23 @@ Dispose()
         [InlineData("${event-properties:almostAsIntProp}", DbType.Int32, 124)]
         [InlineData("${event-properties:almostAsIntProp}", DbType.Int64, (long)124)]
         [InlineData("${event-properties:almostAsIntProp}", DbType.AnsiString, " 124 ")]
+        [InlineData("${event-properties:emptyprop}", DbType.AnsiString, "")]
+        [InlineData("${event-properties:emptyprop}", DbType.AnsiString, "", true)]
         [InlineData("${event-properties:NullRawValue}", DbType.AnsiString, "")]
-        [InlineData("${event-properties:NullawValue}", DbType.Int32, 0)]
+        [InlineData("${event-properties:NullRawValue}", DbType.Int32, 0)]
         [InlineData("${event-properties:NullRawValue}", DbType.AnsiString, null, true)]
         [InlineData("${event-properties:NullRawValue}", DbType.Int32, null, true)]
+        [InlineData("${event-properties:NullRawValue}", DbType.Guid, null, true)]
         [InlineData("", DbType.AnsiString, null, true)]
         [InlineData("", DbType.Int32, null, true)]
+        [InlineData("", DbType.Guid, null, true)]
         public void GetParameterValueTest(string layout, DbType dbtype, object expected, bool allowDbNull = false, bool convertToDecimal = false)
         {
             // Arrange
             var logEventInfo = new LogEventInfo(LogLevel.Debug, "logger1", "message 2");
             logEventInfo.Properties["intprop"] = 123;
             logEventInfo.Properties["boolprop"] = true;
+            logEventInfo.Properties["emptyprop"] = "";
             logEventInfo.Properties["almostAsIntProp"] = " 124 ";
             logEventInfo.Properties["dateprop"] = new DateTime(2018, 12, 30, 13, 34, 56);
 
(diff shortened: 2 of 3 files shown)


src/NLog/Targets/DatabaseParameterInfo.cs
```diff
@@ -134,6 +134,13 @@ public DatabaseParameterInfo(string parameterName, Layout parameterLayout)
         [DefaultValue(null)]
         public CultureInfo Culture { get; set; }
 
+        /// <summary>
+        /// Gets or sets whether empty value should translate into DbNull
+        /// </summary>
+        /// <docgen category='Parameter Options' order='8' />
+        [DefaultValue(false)]
+        public bool AllowDbNull { get; set; }
```
304NotModified, line 142: What about EmptyToDbNull or Nullable?

snakefoot: Well in SQL language it is often called allow NULL. So that is why I like the **Allow**-part. But because in DotNet then it is not an actually null-reference but instead the artificial **DbNull**. This gives **AllowDbNull**

snakefoot:
> And is this difficult to unit test?

I seldom write unit-tests when making prototypes. I have never used the DatabaseTarget for anything myself. I was actually hoping that you or @leonardder would chime in and acknowledge / reject the direction.

snakefoot:
@304NotModified Now added unit-test that puts some more concrete around the name `AllowDbNull` :)

src/NLog/Targets/DatabaseParameterInfo.cs
304NotModified: Clear, but it does more than allow, it translates. I'm not sure how clear that is from the name (see also the description, it's using the term translate and not allow). But I also like to have feedback if others :)

snakefoot: Have updated to source-code comment about having allow null configured for the database-column :)

leonardder:
Would this pr cover the case where

1. The database column allows nulls and is of a string type
2. An empty string (so not a null) is written to the target?

snakefoot:
NLog does not know the difference between null and string.Empty:


- AllowDbNull = false will change null into string.Empty when DbType is string (else default of the expected value type, ex 0 when integer)

- AllowDbNull = true will change null and string.Empty into DbNull


Sent from my Sony Xperia


---- Leonard de Ruijter wrote ----


Would this pr cover the case where

  1.  The database column allows nulls and is of a string type
  2.  An empty string (so not a null) is written to the target?

tests/NLog.UnitTests/Targets/DatabaseTargetTests.cs
```diff
@@ -607,7 +607,13 @@ public void LevelParameterTest(DbType? dbType, bool noRawValue, string expectedV
         [InlineData("${event-properties:almostAsIntProp}", DbType.Int32, 124)]
         [InlineData("${event-properties:almostAsIntProp}", DbType.Int64, (long)124)]
         [InlineData("${event-properties:almostAsIntProp}", DbType.AnsiString, " 124 ")]
-        public void GetParameterValueTest(string layout, DbType dbtype, object expected, bool convertToDecimal = false)
+        [InlineData("${event-properties:NullRawValue}", DbType.AnsiString, "")]
+        [InlineData("${event-properties:NullawValue}", DbType.Int32, 0)]
```
leonardder, line 611:
```suggestion
        [InlineData("${event-properties:NullRawValue}", DbType.Int32, 0)]
```

tests/NLog.UnitTests/Targets/DatabaseTargetTests.cs
```diff
@@ -607,7 +607,13 @@ public void LevelParameterTest(DbType? dbType, bool noRawValue, string expectedV
...
+        [InlineData("${event-properties:NullRawValue}", DbType.AnsiString, null, true)]
+        [InlineData("${event-properties:NullRawValue}", DbType.Int32, null, true)]
```
leonardder, line 613: It would be nice if you could add tests for Guid as well:

```suggestion
        [InlineData("${event-properties:NullRawValue}", DbType.Int32, null, true)]
        [InlineData("${event-properties:NullRawValue}", DbType.Guid, null, true)]
```

tests/NLog.UnitTests/Targets/DatabaseTargetTests.cs
```diff
@@ -607,7 +607,13 @@ public void LevelParameterTest(DbType? dbType, bool noRawValue, string expectedV
...
+        [InlineData("", DbType.AnsiString, null, true)]
+        [InlineData("", DbType.Int32, null, true)]
```
leonardder, line 615:
```suggestion
        [InlineData("", DbType.Int32, null, true)]
        [InlineData("", DbType.Guid, null, true)]
```

snakefoot:
@leonardder Have fixed the spelling error, and added some testing of `DbType.Guid`

Introduced a small inconsistency, so if the `LogEventInfo.Properties`-dictionary contains `string.Empty` value, then it will respect the value and not convert into DbNull when using `AllowDbNull = true`.

304NotModified:
Merged! Great feature!

PR #4078, opened 2020-08-10
BufferingWrapper should use AsyncRequestQueue instead of LogEventInfo…

### LINKED ISSUE #4054: BufferingWrapper replace LogEventInfoBuffer with AsyncRequestQueue
Right now there are two containers for handling buffering:
- LogEventInfoBuffer  (Low performance and high allocation)
- AsyncRequestQueue (High performance and low allocation)

Should mark `LogEventInfoBuffer` as obsolete and instead use `AsyncRequestQueue` in `BufferingWrapper`.

menishmueli:
- Marked LogEventInfoBuffer as obsolete
- BufferingWrapper is using now AsyncRequestQueue instead of LogEventInfoBuffer
- Fixed bug in AsyncRequestQueue DequeueBatch method - count summery documented that passing -1 will result in taking everything from buffer when in reality it's caused aritmetic exception

fix #4054

### DIFF
--- a/src/NLog.Wcf/Targets/LogReceiverWebServiceTarget.cs
+++ b/src/NLog.Wcf/Targets/LogReceiverWebServiceTarget.cs
@@ -53,7 +53,9 @@ namespace NLog.Targets
     [Target("LogReceiverService")]
     public class LogReceiverWebServiceTarget : Target
     {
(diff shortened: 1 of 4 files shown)


304NotModified:
Thanks, looks good to me!

Issue #3995, opened 2020-05-30
Make IAppDomain internal


snakefoot:
- Remove constructor with IAppDomain or make them internal
- Make LogFactory.CurrentAppDomain internal - see https://github.com/NLog/NLog/pull/4304
- make IAppDomain internal

Instead make it internal and change it to receive `IAppEnvironment` (And fix `LogFactory.DefaultAppEnvironment.CurrentProcessId`)

snakefoot:
Closing issue as resolved, as IAppDomain is now obsolete. Will now be part of normal cleanup. See also #4433

Issue #3945, opened 2020-05-04
Log4JXmlEventLayoutRenderer - Remove Constructor with AppDomain


snakefoot:
Make this constructor internal:

https://github.com/NLog/NLog/blob/8fb2a857f7f9e0b5980336ddec622c8a98236725/src/NLog/LayoutRenderers/Log4JXmlEventLayoutRenderer.cs#L82

(And change parameter from IAppDomain to IAppEnvironment)

PR #4079, opened 2020-08-10
Log4JXmlEventLayoutRenderer should use be injected withIAppEnvironmen…

menishmueli:
Make Log4JXmlEventLayoutRenderer accept IAppEnvironment in ctor instead of IAppDomain and make that ctor internal as instructed in #3945

Also seems #3995 is a duplicate of #3945


304NotModified:
Thx!

PR #4102, opened 2020-08-29
LoggingConfiguration - ValidateConfig should only throw when enabled

### LINKED ISSUE #4099: Unit tests failing when referencing project using NLog 4.7
Hi all, some of our application's unit tests are failing after upgrading to NLog 4.7. We've done some troubleshooting and understand the reason, but we're unclear if this is unintended behavior in NLog 4.7 or if we should update our tests to work around it. The tests pass under NLog 4.6, but fail when using NLog 4.7.

We set the `ConfigurationItemFactory.Default.CreateInstance` delegate so we can inject services into a custom NLog target. The details of the custom NLog target aren't so important. We use it wrapped in a BufferingWrapper like so:

```xml
<target name="wrapperTarget" xsi:type="BufferingWrapper" slidingTimeout="true" bufferSize="${bufferSize}" flushTimeout="${flushTimeOut}">
    <target xsi:type="OurCustomTarget"
        BatchSize ="${bufferSize}"
        Path ="${customLogPath}"
        Protocol="${configsetting:item=Protocol}" />
</target>
```

In our unit tests, we don't bother setting `ConfigurationItemFactory.Default.CreateInstance` as historically we haven't needed (or wanted!) our custom targets to be used when running unit tests.

After upgrading to NLog 4.7 we get the following exceptions in our tests:

```
NLog.NLogConfigurationException: Required parameter 'WrappedTarget' on 'BufferingWrapper Target[wrapperTarget]()' was not specified.

at NLog.Internal.PropertyHelper.CheckRequiredParameters(Object o)
at NLog.Config.LoggingConfiguration.ValidateConfig()
at NLog.Config.LoggingConfiguration.InitializeAll()
at NLog.LogFactory.ReconfigExistingLoggers()
at NLog.LogFactory.GetLoggerThreadSafe(String name, Type loggerType)
at NLog.LogFactory.GetLogger(String name)
at NLog.LogManager.GetLogger(String name)
at OurCompany.Logging.LogService.GetLogger(String name) in E:\BuildAgent\...
at OurCompany.Service.cctor() in E:\BuildAgent\...
```

I think what's going on here is:

1. When our unit test project builds, it copies the tested project's NLog.config. We don't do this explicitly -- it just happens by us referencing the tested project.
1. Because we haven't set `ConfigurationItemFactory.Default.CreateInstance`, NLog does not know how to construct our custom target, so BufferingWrapper.WrappedTarget is not set.
1. Some new validation in NLog 4.7 is throwing a hard exception in this case.

I suspect the above steps 1 and 2 were also happening in NLog 4.6, but some new validation was added in NLog 4.7.

Potential fixes we could apply on our end are:

1. Prevent the copying of the NLog.config from the tested project to the unit test project. 
1. Set some sort of stub `ConfigurationItemFactory.Default.CreateInstance` in each unit test project

Any suggestions on the best way around this? Thanks!

snakefoot:
Trying to resolve #4099

For some reason then NLog has always thrown exceptions on missing / invalid required-parameters. Independent of throwException or throwConfigException.

I think NLog should continue even if having loaded invalid config (Logging should not break the host application).

Have adjusted the validation logic so it is expected that objects that implements `ISupportsInitialize` will implicit call `CheckRequiredParameters`. This ensures that Target-objects will become disabled and not call `InitializeTarget` (Just like now).

This has the side-effect that Layouts and LayoutRenderers will now also call `CheckRequiredParameters` when created and used outside the context of the `LoggingConfiguration`. Think this is an improvement.

### DIFF
--- a/src/NLog/Common/AsyncHelpers.cs
+++ b/src/NLog/Common/AsyncHelpers.cs
@@ -230,9 +230,10 @@ namespace NLog.Common
                 T itemCopy = item;
                 StartAsyncTask(new AsyncHelpersTask(s =>
                 {
+                    var preventMultipleCalls = PreventMultipleCalls(continuation);
                     try
                     {
-                        action(itemCopy, PreventMultipleCalls(continuation));
+                        action(itemCopy, preventMultipleCalls);
                     }
                     catch (Exception ex)
                     {
@@ -241,6 +242,8 @@ namespace NLog.Common
                         {
                             throw;  // Throwing exceptions here will crash the entire application (.NET 2.0 behavior)
                         }
+
+                        preventMultipleCalls.Invoke(ex);
                     }
                 }), null);
             }
--- a/src/NLog/Config/LoggingConfiguration.cs
+++ b/src/NLog/Config/LoggingConfiguration.cs
@@ -743,7 +743,18 @@ namespace NLog.Config
 
             foreach (object o in _configItems)
             {
-                PropertyHelper.CheckRequiredParameters(o);
+                try
+                {
+                    if (o is ISupportsInitialize)
+                        continue;   // Target + Layout + LayoutRenderer validate on Initialize()
+
+                    PropertyHelper.CheckRequiredParameters(o);
+                }
+                catch (Exception ex)
+                {
+                    if (ex.MustBeRethrown())
+                        throw;
+                }
             }
         }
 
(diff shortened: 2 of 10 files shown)


304NotModified (approved):
Thanks! Looks good!

PR #4106, opened 2020-09-01
FileTarget - New current file should write start header after archive of previous file

### LINKED ISSUE #4105: FileTarget not writing header when archiving is turned on
Edit: Since this may be relevant.  To test this behavior, I was running the program and changing the Date on my system clock / win 10.  I can also provide a simplified console program if necessary

**NLog version**: (e.g. 4.4.13)

**Platform**: .Net45

```xml
      autoReload="true"
      throwExceptions="true">
  <targets>
    <target name="console" xsi:type="Console" layout="${longdate} [${level}] ${callsite}: ${message}"/>
    <target name="eventlog"
            keepFileOpen="true"
            openFileCacheTimeout="15"
            concurrentWrites="true"
            archiveFileName="logs/archive-{#}.log"
            archiveNumbering="Date"
            maxArchiveFiles="30"
            archiveEvery="Day"
            xsi:type="File"
            fileName="logs/log-${shortdate}.log"
            layout="${longdate}: ${message}"/>
  </targets>

  <rules>
    <logger name="*" minlevel="Trace" writeTo="console" enabled="true"/>
    <logger name="eventlog" minlevel="Info" writeTo="eventlog" enabled="true"/>
  </rules>

</nlog>
```

I also manually configure the header via code

```csharp
            var LogConfig = LogManager.Configuration;
            var Target = LogConfig.FindTargetByName("eventlog") as FileTarget;
            var CurrentDate = DateTime.Now;
            var BuildString = $@"----------------------------------------
-----------  Information ---------------
- Version: {Configuration.Version}
- Build: {Configuration.Build}
- Date: {Configuration.TimeStamp}
- Today: {CurrentDate}
----------------------------------------";

            _headerString = BuildString;
            Target.Layout = new LayoutWithHeaderAndFooter() { Header = BuildString, Layout = Target.Layout };
```

- What is the current result?
I am using the FileTarget to log to a file with a filename based on the date.  At the start of my application, I modify the Header layout with specific info and it works great for the first write.  Once the first file is archived, no further files will render the header information.

I believe the header should be written first whenever a new file is created (daily in my case), even when archiving old files.  

Yep, no errors.  Behavior matches the code.

No exceptions, but upon cloning NLog, checking out 4.7.4 and stepping through I found that the FileTarget ProcessLogEvent method has the following happening ---
```csharp
        private void ProcessLogEvent(LogEventInfo logEvent, string fileName, ArraySegment<byte> bytesToWrite)
        {
            // min value as expected
            DateTime previousLogEventTimestamp = InitializeFile(fileName, logEvent);

            // this is true as previousLogTimeStamp is DateTime.MinValue
            bool initializedNewFile = previousLogEventTimestamp == DateTime.MinValue;


            if (initializedNewFile && fileName == _previousLogFileName && _previousLogEventTimestamp.HasValue)
                previousLogEventTimestamp = _previousLogEventTimestamp.Value;

            // this is true, file gets archived
            bool archiveOccurred = TryArchiveFile(fileName, logEvent, bytesToWrite.Count, previousLogEventTimestamp, initializedNewFile);
            
            if (archiveOccurred)
                // this makes intializedNewFile to false, as initializeFile was already called and sets a DateTime to the  `initialized files` Dictionary
                initializedNewFile = InitializeFile(fileName, logEvent) == DateTime.MinValue;

            if (ReplaceFileContentsOnEachWrite)
            {
                ReplaceFileContent(fileName, bytesToWrite, true);
            }
            else
            {
                // write to header does not get called inside WriteToFile as initializedNewFile is false
                WriteToFile(fileName, bytesToWrite, initializedNewFile);
            }
            ....
```

- Are there any workarounds?
Yes - disable archiving and the header will print to each newly created file.

I have only tried 4.7.4

Will follow up later, should be able to.

snakefoot:
Trying to resolve  #4105. Bug was introduced with #1993

### DIFF
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -1227,7 +1227,7 @@ namespace NLog.Targets
 
             bool archiveOccurred = TryArchiveFile(fileName, logEvent, bytesToWrite.Count, previousLogEventTimestamp, initializedNewFile);
             if (archiveOccurred)
-                initializedNewFile = InitializeFile(fileName, logEvent) == DateTime.MinValue;
+                initializedNewFile = InitializeFile(fileName, logEvent) == DateTime.MinValue || initializedNewFile;
 
             if (ReplaceFileContentsOnEachWrite)
             {
@@ -1834,8 +1834,8 @@ namespace NLog.Targets
                     // See: https://msdn.microsoft.com/en-us/library/system.threading.abandonedmutexexception.aspx
                 }
 #endif
-
-                return TryArchiveFileAfterCloseFileAppender(fileName, archiveFile, ev, upcomingWriteSize, previousLogEventTimestamp, initializedNewFile);
+                ArchiveFileAfterCloseFileAppender(archiveFile, ev, upcomingWriteSize, previousLogEventTimestamp);
+                return true;
             }
             finally
             {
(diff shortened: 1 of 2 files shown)


304NotModified (approved):
Thanks!

snakefoot:
Force-commit to remove code-smell, and improve InternalLogger.Debug-message.

304NotModified:
@snakefoot please no force push next time. I cannot check what have been changed as the review state has been reset. 

Ps: Nowadays I squash merge all pull requests if no explicit merge is needed after squash

PR #4113, opened 2020-09-13
Detect Microsoft.Data.SqlClient in DatabaseTarget

### LINKED ISSUE #4087: Detect Microsoft.Data.SqlClient in DatabaseTarget
NLog automatically tries to load `System.Data.SqlClient`-assembly on NetCore unless some other `DBProvider` has been specified.

Microsoft is now selling `Microsoft.Data.SqlClient` as the new default on NetCore (and soon .NET 5)

https://www.nuget.org/packages/Microsoft.Data.SqlClient/

Consider changing so it automatically loads `Microsoft.Data.SqlClient` and then have fallback to `System.Data.SqlClient`. See also #3792 and #3776

njqdev:
Fixes #4087 (hopefully)

But I wouldn't be surprised if I misunderstood the issue.

### DIFF
--- a/src/NLog/Targets/DatabaseTarget.cs
+++ b/src/NLog/Targets/DatabaseTarget.cs
@@ -523,51 +523,45 @@ namespace NLog.Targets
         {
             switch (DBProvider.ToUpperInvariant())
             {
-                case "MICROSOFT.DATA.SQLCLIENT":
-                    {
-#if NETSTANDARD
-                        var assembly = Assembly.Load(new AssemblyName("Microsoft.Data.SqlClient"));
-#else
-                        var assembly = typeof(IDbConnection).GetAssembly();
-#endif
-                        ConnectionType = assembly.GetType("Microsoft.Data.SqlClient.SqlConnection", true, true);
-                        break;
-                    }
-                case "SYSTEM.DATA.SQLCLIENT":
-                    {
-#if NETSTANDARD
-                        var assembly = Assembly.Load(new AssemblyName("System.Data.SqlClient"));
-#else
-                        var assembly = typeof(IDbConnection).GetAssembly();
-#endif
-                        ConnectionType = assembly.GetType("System.Data.SqlClient.SqlConnection", true, true);
-                        break;
-                    }
                 case "SQLSERVER":
                 case "MSSQL":
                 case "MICROSOFT":
                 case "MSDE":
-                    {
 #if NETSTANDARD
-                        Assembly assembly;
-                        string assemblyName;
+                    {
                         try
                         {
-                            assemblyName = "Microsoft.Data.SqlClient";
-                            assembly = Assembly.Load(new AssemblyName(assemblyName));
+                            var assembly = Assembly.Load(new AssemblyName("Microsoft.Data.SqlClient"));
+                            ConnectionType = assembly.GetType("Microsoft.Data.SqlClient.SqlConnection", true, true);
                         }
-                        catch (System.IO.IOException)
+                        catch (Exception ex)
                         {
-                            assemblyName = "System.Data.SqlClient";
-                            assembly = Assembly.Load(new AssemblyName(assemblyName));
+                            InternalLogger.Warn(ex, "DatabaseTarget(Name={0}): Failed to load assembly 'Microsoft.Data.SqlClient'. Falling back to 'System.Data.SqlClient.'", Name);
+                            var assembly = Assembly.Load(new AssemblyName("System.Data.SqlClient"));
+                            ConnectionType = assembly.GetType("System.Data.SqlClient.SqlConnection", true, true);
                         }
-#else
-                        var assemblyName = "System.Data.SqlClient";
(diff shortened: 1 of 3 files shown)


njqdev:
The travis build passed on my fork so I'm not sure why it failed here.

304NotModified:
Travis is sometimes a bit unstable. I restarted it and it's now green :)

304NotModified:
@snakefoot I think this is fine, do you agree? Unfortunately we can't really test the fallback, as in that case we need to mock Assembly.Load I guess

snakefoot:
Looks good to me.

snakefoot:
When having specified `case "SYSTEM.DATA.SQLCLIENT":` then it would be nice if it didn't first attempt `Microsoft.Data.SqlClient`.

Consider adding `case "MICROSOFT.DATA.SQLCLIENT":` where it doesn't perform fallback to `System.Data.SqlClient`.

njqdev:
@snakefoot If I only have that single case then the test below fails. Should I separate it into 2 tests with `#if NETSTANDARD` `#else`?

[https://github.com/NLog/NLog/blob/324f466b7f01d6ada01d8956a2692ad6101b333e/tests/NLog.UnitTests/Targets/DatabaseTargetTests.cs#L1249-L1251](url)

snakefoot:
I think it is fine to leave `public void SqlServerShorthandNotationTest()` untouched. It is testing these provider-names `"microsoft", "msde", "mssql", "sqlserver"`. And for NetStandard then those names should map to  `"MICROSOFT.DATA.SQLCLIENT`  and then fallback to `"SYSTEM.DATA.SQLCLIENT"`.

I'm talking about the cases where one explicitly specify `"SYSTEM.DATA.SQLCLIENT"` (and maybe `"MICROSOFT.DATA.SQLCLIENT`). Where it should not perform fallback logic.

njqdev:
I added a case for each, but I guess another option could be to have a case for `SYSTEM.DATA.SQLCLIENT` and then include Microsoft with the others.

`
                case "SQLSERVER":
                case "MSSQL":
                case "MICROSOFT":
                case "MSDE":
                case "MICROSOFT.DATA.SQLCLIENT":
`

snakefoot:
It is a little confusing that NetFramework will always load from System.Data, when having specified "MICROSOFT.DATA.SQLCLIENT".

Maybe something like this:
```c#
            switch (DBProvider.ToUpperInvariant())
            {
                case "SQLSERVER":
                case "MSSQL":
                case "MICROSOFT":
                case "MSDE":
#if NETSTANDARD
                    {
                         try
                         {
                                var assembly = Assembly.Load(new AssemblyName("Microsoft.Data.SqlClient"));
                                ConnectionType = assembly.GetType("System.Data.Microsoft.SqlConnection", true, true);
                         }
                         catch (Exception ex)
                         {
                                // TODO Some logging, about failing and doing fallback
                                var assembly = Assembly.Load(new AssemblyName("System.Data.SqlClient"));
                                ConnectionType = assembly.GetType("SqlClient.Data.Microsoft.SqlConnection", true, true);
                         }
                        break;
                    }
              case "SYSTEM.DATA.SQLCLIENT":
                    {
                         var assembly = Assembly.Load(new AssemblyName("System.Data.SqlClient"));
                         ConnectionType = assembly.GetType("System.Data.SqlClient.SqlConnection", true, true);
                         break;
                    } 
#else
             case "SYSTEM.DATA.SQLCLIENT":
                    {
                        var assembly = typeof(IDbConnection).GetAssembly();
                        ConnectionType = assembly.GetType("System.Data.SqlClient.SqlConnection", true, true);
                        break;
                    }
#endif
             case "MICROSOFT.DATA.SQLCLIENT":
                   {
                        var assembly = Assembly.Load(new AssemblyName("Microsoft.Data.SqlClient"));
                        ConnectionType = assembly.GetType("Microsoft.Data.SqlClient.SqlConnection", true, true);
                        break;
                   } 
```

njqdev:
I used your suggestion but with `case "MICROSOFT.DATA.SQLCLIENT":` under `#if NETSTANDARD`, and `InternalLogger.Warn` for logging.

snakefoot:
Any reason why `case "MICROSOFT.DATA.SQLCLIENT":` cannot work for .NetFramework ?

njqdev:
`Assembly.Load` doesn't work because of System.Reflection only being used for .Net Standard.

https://github.com/NLog/NLog/blob/324f466b7f01d6ada01d8956a2692ad6101b333e/src/NLog/Targets/DatabaseTarget.cs#L42-L44 

And if I understood you correctly `var assembly = typeof(IDbConnection).GetAssembly();` would load the System.Data assembly.

snakefoot:
> Assembly.Load doesn't work because of System.Reflection only being used for .Net Standard.

Then I would just change:

```
 #if NETSTANDARD 
     using System.Reflection; 
 #endif 
```

into (Remove  the `#if`):

```
     using System.Reflection;
```

snakefoot (approved):
Looks good to me

304NotModified:
Thanks!

PR #3142, opened 2019-02-24
upgrade to C# 7.2

304NotModified:
fails "MediumTrustWithExternalClass" test

works on: 7.0 and 7.1
fails on 7.2 and 7.3

most important exceptions:

```
System.Security.SecurityException
  HResult=0x8013150A
  Message=Request for the permission of type 'System.Security.Permissions.FileIOPermission, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089' failed.
  Source=mscorlib
  StackTrace:
   at System.Security.CodeAccessSecurityEngine.Check(Object demand, StackCrawlMark& stackMark, Boolean isPermSet)
   at System.Security.CodeAccessPermission.Demand()
   at System.AppDomainSetup.VerifyDir(String dir, Boolean normalize)
   at System.AppDomain.get_BaseDirectory()
   at NLog.Internal.Fakeables.AppDomainWrapper.LookupBaseDirectory(AppDomain appDomain) in D:\nlog\nlog\src\NLog\Internal\Fakeables\AppDomainWrapper.cs:line 100

System.Security.SecurityException
  HResult=0x8013150A
  Message=Request for the permission of type 'System.Security.Permissions.EnvironmentPermission, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089' failed.
  Source=mscorlib
  StackTrace:
   at System.Security.CodeAccessSecurityEngine.Check(Object demand, StackCrawlMark& stackMark, Boolean isPermSet)
   at System.Security.CodeAccessPermission.Demand()
   at System.Environment.GetEnvironmentVariable(String variable)
   at NLog.Internal.EnvironmentHelper.GetSafeEnvironmentVariable(String name) in D:\nlog\nlog\src\NLog\Internal\EnvironmentHelper.cs:line 83

System.Security.SecurityException
  HResult=0x8013150A
  Message=Request for the permission of type 'System.Security.Permissions.EnvironmentPermission, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089' failed.
  Source=mscorlib
  StackTrace:
   at System.Security.CodeAccessSecurityEngine.Check(Object demand, StackCrawlMark& stackMark, Boolean isPermSet)
   at System.Security.CodeAccessPermission.Demand()
   at System.Environment.GetEnvironmentVariable(String variable)
[...]

PR #3138, opened 2019-02-23
Split XML layout to XMLELement and XmlElementBase

304NotModified:
Counter proposal of #3137

fixes #3135

PR #4114, opened 2020-09-13
Fix VerificationException Operation could destabilize the runtime

snakefoot:
That was triggered by copying readonly structs in medium trust AppDomain. Not sure why it would destabilize, but lets not get into details.

See also #3142 and #3138

### DIFF
--- a/src/NLog/Internal/AppendBuilderCreator.cs
+++ b/src/NLog/Internal/AppendBuilderCreator.cs
@@ -49,7 +49,7 @@ namespace NLog.Internal
         /// </summary>
         public StringBuilder Builder => _builder.Item;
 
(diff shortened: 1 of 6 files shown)


304NotModified:
Looks good. Is it a good idea to add some comments that the `readonly` should not be readded to these places. E.g. R#/SonarQube will suggest to add these again

snakefoot:
Sure. Added some comments

304NotModified (approved):

304NotModified:
Thanks!

snakefoot:
See also: https://minidump.net/verificationexception-in-net-framework-when-using-structs-6269eb3df448

PR #4123, opened 2020-09-28
BucketSort - StringComparer.Ordinal should be faster than EqualityComparer.Default

snakefoot:


304NotModified (approved):

snakefoot:
Force push to restart Azure-Build, that was stuck in cancelled state. No changes made.

304NotModified (approved):

304NotModified:
Thanks for the info :)

snakefoot:
Starting to doubt if this is actually a performance improvement, after looking at the code for StringComparer.Ordinal.  It seems it is actually best when meeting many different long strings where ordinal-compare can shine.

But when meeting the same string again and again (ex. filetarget-filename) then `string.Equals` is faster as it performs  `Object.ReferenceEquals(this, value)` very early and it will usually return true (because of caching).

snakefoot:
Created PR #4125 to rollback this one. Guess I should test instead of guessing.

304NotModified:
This has been reverted by removing the last commit, 28ce43fff546df2204540bf1ec81a72a97836e8a from dev

PR #4124, opened 2020-09-30
ScopeContext to replace MDLC and NDLC

### LINKED ISSUE #4056: Merge NestedDiagnosticsLogicalContext and MappedDiagnosticsLogicalContext into one
Right now there are two async-scope-managers in NLog:

- MappedDiagnosticsLogicalContext
- NestedDiagnosticsLogicalContext 

Both operates like an async-stack, so one can push and pop state on the thread-execution-context.

Thinking that one could implement an unified stack that can contain both scope-values and scope-properties.

This would also make the NLog-implementation of MEL-BeginScope faster, since it will only update a single state instead of updating both NDLC-state and MDLC-state. Could also consider doing lazy evaluation/parsing of scope-properties (extracted from scope-value), so it is only done when actually needing to write LogEvent.

Maybe call the class `ThreadExecutionScopeContext` and the layout renderer `${scopecontext}` that matches `ILog.PushScopeContext`.

snakefoot:
Resolves #4056

ScopeContext has some small behavior adjustments:
- Dictionary is case-insensitive (Breaking change)
- Dictionary is stacked, so it can restore old values
   - Special legacy-mode activates for MappedDiagnosticsLogicalContext.Set, but can still restore
   - Special legacy-mode activates for NestedDiagnosticsLogicalContext.Pop, but can still restore
- ScopeContext-Timing only works on NetStandard / Net46 (Breaking change)
   - This is because legacy CallContext do not like unknown NLog types
   - Added unit-test `NdlcGetAllMessages` that explodes when putting non-Microsoft objects into the CallContext. If adding this unit-test to master-branch then it will fail because NLog own objects are not welcome in the CallContext. 

ScopeContext has better performance since it performs lazy allocation of Dictionary (To ensure unique properties) until it actually has to write the log-message. And NLog PR now only uses a single AsyncLocal-variable to capture the BeginScope-state (Instead of one for NDLC and one for MDLC).

Because MDLC (and NDLC) are well-known constructs in the NLog-universe, then I prefer that `${mdlc}` and `${ndlc}` continue to work when people using NLog.Extension.Logging upgrade to lastest NLog 5.0 (Along with IncludeMdlc on JsonLayout), Then over time people can be convinced to move to the new `${scopecontext}` (New PR). 

Very happy that all the existing unit-test for MDLC + NDLC are working without any modifications. ScopeContext is also able to handle the more exotic methods of adding / removing properties within an active scope.

### DIFF
--- a/src/NLog/Contexts/ScopeContext.cs
+++ b/src/NLog/Contexts/ScopeContext.cs
@@ -1,0 +1,994 @@
+﻿// 
+// Copyright (c) 2004-2020 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
+// 
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without 
+// modification, are permitted provided that the following conditions 
+// are met:
+// 
+// * Redistributions of source code must retain the above copyright notice, 
+//   this list of conditions and the following disclaimer. 
+// 
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution. 
+// 
+// * Neither the name of Jaroslaw Kowalski nor the names of its 
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission. 
+// 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE 
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE 
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE 
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR 
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS 
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN 
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) 
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF 
+// THE POSSIBILITY OF SUCH DAMAGE.
+// 
+
+using System;
+using System.Collections.Generic;
+using NLog.Internal;
+
+namespace NLog
+{
+    /// <summary>
+    /// <see cref="ScopeContext"/> allows one to store state in the thread execution context. All LogEvents created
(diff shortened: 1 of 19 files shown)


snakefoot:
After looking at https://github.com/NLog/NLog.Extensions.Logging/issues/445 then I think further building blocks can be made:

- Add the missing ScopeProperties-LayoutRenderer (MLDC) and ScopeOperations-LayoutRenderer (NDLC).
- Extend Logger-object with methods PushScopeProperty and PushScopeOperation
- Add DefaultScopeProperties-dictionary to LoggingConfiguration (`ThreadSafeDictionary<string, Layout>`).
   - Fix locations that support `IncludeScopeProperties` to retrieve the ScopeProperties from the LogFactory.
   - Extend LoggingConfigurationParser to handle default-scope-properties. Maybe in a way so it avoids all the pitfalls discovered with NLog-config-variables together with autoreload.

src/NLog/Contexts/ScopeContext.cs
```diff
@@ -0,0 +1,994 @@
...
+    /// <see cref="ScopeContext"/> allows one to store state in the thread execution context. All LogEvents created
+    /// within a scope can include the scope state when wanted. The logical operation scope state supports both
+    /// scope-properties and scope-operation-states.
+    /// </summary>
+    /// <remarks>
+    /// <see cref="ScopeContext"/> unifies <see cref="MappedDiagnosticsLogicalContext"/> (MDLC) and <see cref="NestedDiagnosticsLogicalContext"/> (NDLC).
+    /// 
+    /// .NetCore (and .Net46) uses AsyncLocal for handling the thread execution context. Older .NetFramework uses System.Runtime.Remoting.CallContext
+    /// </remarks>
+    public static class ScopeContext
```
304NotModified, line 50: I think it's wise to move to a non-static class with an interface and a Default property. We need that for unit testing/mocking/DI In the future I also like to add GlobalContext  (the same, non static and interfaces)

src/NLog/Contexts/ScopeContext.cs
```diff
@@ -0,0 +1,994 @@
...
+    {
+#if !NET35 && !NET40
+        /// <summary>
+        /// Creates logical operation scope that includes provided properties and stack-value
+        /// </summary>
+        /// <param name="operationState">Value to added to the scope operation stack</param>
+        /// <param name="properties">Properties being added to the scope dictionray</param>
```
304NotModified, line 57: dictionray ;)

304NotModified:
Maybe I missing something, but what's the difference between property and operationProperty?

src/NLog/Contexts/ScopeContext.cs
snakefoot: Both CallContext and AsyncLocal are static-variables, and will not disappear by living in a singleton. I was planning to let the Logger-object include the methods `PushScopeProperty` and `PushScopeOperation`, and do the mocking with Logger-object.

snakefoot:
ScopeProperty (MDLC) is name + value dictionary. Ex. "RequestId" = Guid.NewGuid().

ScopeOperationState (NDLC) is operation-description as stack. Ex.
   - "Processing order request"
       - "Saving order to database"

src/NLog/Contexts/ScopeContext.cs
snakefoot: Thanks. Fixed spelling. But Azure-build is now in cancelled state. Can you give it a kick? (Instead of me force-pushing)

snakefoot:
@304NotModified Anything one can do to get the ball rolling on this one. Would like to continue on with the [todo-list.](https://github.com/NLog/NLog/pull/4124#issuecomment-703072776)

snakefoot:
Updated wiki-pages:
- https://github.com/NLog/NLog/wiki/ScopeNested-Layout-Renderer
- https://github.com/NLog/NLog/wiki/ScopeProperty-Layout-Renderer

PR #4125, opened 2020-10-02
BucketSort - EqualityComparer.Default is always faster than StringComparer.Ordinal 

snakefoot:
Rollback of #4123

Starting to doubt if this is actually a performance improvement, after looking at the code for StringComparer.Ordinal. It seems it is actually best when meeting many different long strings where ordinal-compare can shine.

### DIFF
--- a/src/NLog/Internal/Collections/ThreadSafeDictionary.cs
+++ b/src/NLog/Internal/Collections/ThreadSafeDictionary.cs
@@ -41,7 +41,6 @@ namespace NLog.Internal
     internal class ThreadSafeDictionary<TKey, TValue> : IDictionary<TKey, TValue>
     {
         private readonly object _lockObject = new object();
-        private readonly IEqualityComparer<TKey> _comparer;
         private Dictionary<TKey, TValue> _dict;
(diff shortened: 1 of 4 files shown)


snakefoot:
Yes my worry was confirmed `StringComparer.Ordinal` performs the same operations as EqualityComparer.Default, but introduces an overhead. The only place where `StringComparer.Ordinal` has an edge is the implementation of `IComparer<string>` where it actually uses `String.CompareOrdinal`

snakefoot:
Created PR #4125 to rollback this one. Guess I should test instead of guessing.

304NotModified:
I could also remove the last commit of dev. :)

snakefoot:
That would also work. You choose :)


Sent from my Sony Xperia


I could also remove the last commit of dev. :)

304NotModified:
i've removed the last commit from dev, 28ce43fff546df2204540bf1ec81a72a97836e8a

We have to make sure other PRs won't introduce 28ce43fff546df2204540bf1ec81a72a97836e8a again :)

Issue #1169, opened 2016-01-16
Make API calls easier


304NotModified:
As discussed in https://github.com/NLog/NLog/issues/1164

Proposal:
1. Add maxLevel to contructor: 
   
   ``` cs
   public LoggingRule(string loggerNamePattern, 
                               LogLevel minLevel, 
                               LogLevel maxLevel, 
                              Target target)
   ```
2. create an `LoggingConfiguration.AddRule` and `LoggingConfiguration.RemoveRule`, analogous to the current `LoggingConfiguration.AddTarget` and `LoggingConfiguration.RemoveTarget` (targetname should be a `string`?)

Example

``` c#

var generalFileTarget = new FileTarget();
config.AddTarget("general", generalFileTarget);
config.AddRule("*",  LogLevel.Debug, generalFileTarget); //target type
config.AddRule("*",  LogLevel.Error, "target2"); // target name

```
1.  `RegisterDefinition` without name (use the attribute)  + generic version

now:

``` c#
//target
ConfigurationItemFactory.Default.Targets
    .RegisterDefinition("MyFirst", typeof(MyNamespace.MyFirstTarget));

    //layout renderer
ConfigurationItemFactory.Default.LayoutRenderers
    .RegisterDefinition("hello-world", typeof(MyNamespace.HelloWorldLayoutRenderer ));

```

proposal:

``` c#

//target
ConfigurationItemFactory.Default.Targets
    .RegisterDefinition<MyNamespace.MyFirstTarget>();

//layout renderer
ConfigurationItemFactory.Default.LayoutRenderers
    .RegisterDefinition<MyNamespace.HelloWorldLayoutRenderer>();

```

snakefoot:
Maybe resolved by  #3686 (NLog 4.7) and #4323 (NLog 5.0)  where once can do this:

```c#
LogManager.Setup().
   SetupExtensions(eb => {
          eb.AutoLoadAssemblies(false);
          eb.RegisterTarget<MyTarget>("MyTarget");
   });
```

snakefoot:
Closing as resolved with the new NLog 5.0 fluent-API with `LogManager.Setup()`

PR #4128, opened 2020-10-04
LogFactory.Setup().LoadConfiguration() with fluent creation of NLog-Config

snakefoot:
Alternative solution for #1169 and #1429 (See also #2998 and #3000)

Continuing the work started with NLog 4.7 and NLog 4.7.1 to improve fluent-configuration, by adding new extension-methods to the existing `ISetupLoadConfigurationBuilder`-interface.

You can now do this:

```c#
var logger = LogManager.Setup().LoadConfiguration(c =>
{
   c.ForLogger("*").FilterMinLevel(LogLevel.Info).WriteTo(new ConsoleTarget("logconsole")).WithAsync();
   c.ForLogger().WriteTo(new FileTarget("logfile") { FileName = "file.txt" }).WithAsync();
}).GetCurrentClassLogger();
```

Was thinking to add more helper-methods for the individual-target-types, but decided to skip it for simplicity (Only included `MethodCallTarget` as an example)

This allows one to combine file-config with runtime-config, where it loads from NLog.config, and then applies black-hole filtering:

```c#
var logger = LogManager.Setup().LoadFromFile().LoadConfiguration(c =>
{
   c.ForLogger("Microsoft.*").TopRule().WriteToNil(maxLevel: LogLevel.Info);
}).GetCurrentClassLogger();
```

### DIFF
--- a/src/NLog/Config/ISetupConfigurationLoggingRuleBuilder.cs
+++ b/src/NLog/Config/ISetupConfigurationLoggingRuleBuilder.cs
@@ -34,18 +34,9 @@
 namespace NLog.Config
 {
     /// <summary>
-    /// Interface for fluent setup of LoggingConfiguration for LogFactory 
+    /// Interface for fluent setup of LoggingRules for LoggingConfiguration
     /// </summary>
-    public interface ISetupLoadConfigurationBuilder
+    public interface ISetupConfigurationLoggingRuleBuilder : ISetupConfigurationWriteToTargetsBuilder
     {
-        /// <summary>
-        /// LogFactory under configuration
-        /// </summary>
-        LogFactory LogFactory { get; }
-
-        /// <summary>
-        /// LoggingConfiguration being built
-        /// </summary>
-        LoggingConfiguration Configuration { get; set; }
     }
 }
--- a/src/NLog/Config/ISetupConfigurationWriteToTargetsBuilder.cs
+++ b/src/NLog/Config/ISetupConfigurationWriteToTargetsBuilder.cs
@@ -34,18 +34,23 @@
 namespace NLog.Config
 {
     /// <summary>
-    /// Interface for fluent setup of LoggingConfiguration for LogFactory 
+    /// Interface for fluent setup of LoggingRules for LoggingConfiguration
     /// </summary>
-    public interface ISetupLoadConfigurationBuilder
+    public interface ISetupConfigurationWriteToTargetsBuilder
     {
         /// <summary>
-        /// LogFactory under configuration
+        /// LoggingRule being built
         /// </summary>
-        LogFactory LogFactory { get; }
+        LoggingRule LoggingRule { get; }
 
         /// <summary>
         /// LoggingConfiguration being built
         /// </summary>
-        LoggingConfiguration Configuration { get; set; }
+        LoggingConfiguration Configuration { get; }
+
+        /// <summary>
+        /// LogFactory under configuration
+        /// </summary>
+        LogFactory LogFactory { get; }
     }
 }
--- a/src/NLog/Internal/ReflectionHelpers.cs
+++ b/src/NLog/Internal/ReflectionHelpers.cs
@@ -355,5 +355,14 @@ namespace NLog.Internal
             return p.GetGetMethod().Invoke(instance, null);
 #endif
         }
+
+        public static MethodInfo GetDelegateInfo(this Delegate method)
+        {
+#if NETSTANDARD1_0
+            return System.Reflection.RuntimeReflectionExtensions.GetMethodInfo(method);
+#else
+            return method.Method;
+#endif
+        }
     }
 }
--- a/src/NLog/Internal/SetupConfigurationLoggingRuleBuilder.cs
+++ b/src/NLog/Internal/SetupConfigurationLoggingRuleBuilder.cs
@@ -31,21 +31,26 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
-namespace NLog.Config
-{
-    /// <summary>
-    /// Interface for fluent setup of LoggingConfiguration for LogFactory 
-    /// </summary>
-    public interface ISetupLoadConfigurationBuilder
-    {
-        /// <summary>
-        /// LogFactory under configuration
-        /// </summary>
-        LogFactory LogFactory { get; }
+using NLog.Config;
 
-        /// <summary>
-        /// LoggingConfiguration being built
-        /// </summary>
-        LoggingConfiguration Configuration { get; set; }
+namespace NLog.Internal
+{
+    internal class SetupConfigurationLoggingRuleBuilder : ISetupConfigurationLoggingRuleBuilder
+    {
+        public SetupConfigurationLoggingRuleBuilder(LogFactory logFactory, LoggingConfiguration configuration, string loggerNamePattern = null, string ruleName = null)
+        {
+            LoggingRule = new LoggingRule(ruleName) { LoggerNamePattern = loggerNamePattern ?? "*" };
+            Configuration = configuration;
+            LogFactory = logFactory;
+        }
+
+        /// <inheritdoc />
+        public LoggingRule LoggingRule { get; }
+
+        /// <inheritdoc />
+        public LoggingConfiguration Configuration { get; }
+
+        /// <inheritdoc />
+        public LogFactory LogFactory { get; }
     }
 }
--- a/src/NLog/SetupExtensionsBuilderExtensions.cs
+++ b/src/NLog/SetupExtensionsBuilderExtensions.cs
@@ -166,10 +166,9 @@ namespace NLog
             return setupBuilder;
         }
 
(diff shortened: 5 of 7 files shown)


snakefoot:
CLose and reopen to rebuild

snakefoot:
Close and reopen to rebuild

snakefoot:
Close and reopen to rebuild

304NotModified:
Cool! Thanks! 

I've tested this in a short time. I have some suggestions

- Make ForLogger optional and/or maybe order unimportant. e.g.  `c.FilterMaxLevel(Info).WriteTo...`  `c.FilterMaxLevel(Info).ForLogger("a").WriteTo```` and 
- ForLoggerWriteToNil,("a") I expected c.ForLogger("a").WriteToBlackhole() 
- FilterMaxLevel -> ForMaxLevel? (consistent with ForLogger)
- FilterDynamic -> WithFilter? Filter?
- FinalRule -> AsFinal or StopProcessingOtherRules ;)
- TopRule -> AsTopRule, or MoveAsTopRule, as we're changing the rule order?
- AddGlobalContextProperty -> SetGlobalContextProperty, as it sets there is no error when adding the same?
- UseTimeSource -> WithTimeSource?

>Was thinking to add more helper-methods for the individual-target-types, but decided to skip it for simplicity (Only included MethodCallTarget as an example)

Yes, I think this is also nice. I would recommend to create new extension classes for each target, otherwise there will be an extension class with many dependencies. Maybe add those for the most used targets? (guess file, database, eventlog

Also, could we combine FilterMinLevel and FilterMaxLevel? I'm not sure as I see the implementation

304NotModified:
I think this should be NLog 5

snakefoot:
> Make ForLogger optional and/or maybe order unimportant. 

I like that the fluent-interface matches how LoggingRules in NLog.config is defined:

```xml
<logger name="Microsoft.*" minLevel="Debug" writeTo="logConsole" />`
```

It nicely matches:

```c#
ForLogger("Microsoft.*").FilterMinLevel(LogLevel.Debug).WriteTo(new ConsoleTargert());
```

And I also like the default API ensures similar style of how to specify the NLog-config at runtime. So different developers with different solutions will use similar style for expressing their Nlog.config (Easier to help and and easier to recognize).

> ForLoggerWriteToNil,("a") I expected c.ForLogger("a").WriteToBlackhole()

Yes I like that. Then just have maxLevel + toprule as parameter to `WriteToNil()`. Will change that. Not sure if black-hole is still kosher in black-lives-matter-era.

> FilterMaxLevel -> ForMaxLevel? (consistent with ForLogger)

Because ForLogger is the entry-method **for** defining a new LoggingRule, then I prefer that there is only `ForLogger`.

> FinalRule -> AsFinal or StopProcessingOtherRules ;)
> TopRule -> AsTopRule, or MoveAsTopRule, as we're changing the rule order?

Because one is configuring both LoggingRule + LogLevels + Targets, then I like one can see it is for the rule. And `FinalRule` and `TopRule` are exoctic adjustments of the LoggingRule.

> FilterDynamic -> WithFilter? Filter?

I prefer FilterDynamic as it matches FilterMinLevel, FilterMaxLevel etc. All of them enables extra-filtering for the LoggingRule that is being defined.

> AddGlobalContextProperty -> SetGlobalContextProperty, as it sets there is no error when adding the same?

No problem with that. Will change to SetGlobalContextProperty,

> UseTimeSource -> WithTimeSource?

Will change it to SetTimeSource to match SetGlobalContextProperty

> Also, could we combine FilterMinLevel and FilterMaxLevel? I'm not sure as I see the implementation

There is `FilterLevels` that is the combination of the two. But pretty much everyone is using FilterMinLevel. I just included FilterMaxLevel for completion, and because it is used with `WriteToNil`.

> I think this should be NLog 5

I think NLog 4.7 is good for this, since the fluent NLog.LogManager.Setup()-interface was introduced with NLog 4.7. I like small steps, and NLog 5.0 is big bang, and these extension-methods (to an already existing interface) are very isolated. Any mistakes made can also be fixed in NLog 5.0. They cannot be fixed in NLog 5.0.1

snakefoot:
@ThomasArdal You have expressed interest for improvements in the fluent-api for the nlog-config.

Any input on the suggested api is very much appreciated. You can also try this [nuget-package](https://ci.appveyor.com/api/buildjobs/s2d95v4v61o5n9wu/artifacts/artifacts%2FNLog.4.7.5.12142-PR4128.nupkg) (pre-release for local-folder-repository).

ThomasArdal:
@snakefoot I remember that we talked about it. But totally forgot about in which context. Can you help me point me in the direction of an issue, email, something else? Would love to provide some feedback on the new API, but better be in the right context 👍

snakefoot:
This PR is about improving how the NLog users can setup targets for logging using a more fluent-api (See example in initial post for this PR)


Now imagine that you (or another NLog-user) wants to use this api for writing to the elmah.io-nlog-target (instead of the console-target as shown in the example in initial post for this PR)





Sent from my Sony Xperia


---- Thomas Ardal wrote ----


@snakefoot<https://github.com/snakefoot> I remember that we talked about it. But totally forgot about in which context. Can you help me point me in the direction of an issue, email, something else? Would love to provide some feedback on the new API, but better be in the right context 👍

ThomasArdal:
Ah yeah ok. I can see that I am currently recommending to configure the target like this:

```csharp
var elmahIoTarget = new ElmahIoTarget();
elmahIoTarget.Name = "elmahio";
elmahIoTarget.ApiKey = "API_KEY";
elmahIoTarget.LogId = "LOG_ID";
config.AddTarget(elmahIoTarget);
config.AddRuleForAllLevels(elmahIoTarget);
LogManager.Configuration = config;
```

I'll take a look at the new API and see how it works 👍 Thanks

ThomasArdal:
I tried it and I think the new API works great and is a great fluent alternative to the existing API. Here are my comments.

## AddRuleForAllLevels alternative

Maybe it's just me who cannot find it. But is there an alternative to `AddRuleForAllLevels`? Or should I simply leave out any level config if I want to log all levels? I tried this:

```csharp
.FilterMinLevel(LogLevel.AllLevels)
```

But it fails with a compile error since `AllLevels` is an enumerable.

## Target specific `WriteTo`

It would be nice with a more specific `WriteTo`. So rather than:

```csharp
.WriteTo(new ConsoleTarget(...))
```

You could do:

```csharp
.WriteTo(x => x.Console(...))
```

or:

```csharp
.WriteTo.Console(...)
```

snakefoot:
If you don't specify FilterMinLevel or any other level filter then you will get all levels.


I have implemented WriteToMethodCall as an example. I wanted to keep things simple from the start. Also leaving the door open to extract some targets into their own isolated nuget-package with NLog 5.


But it is just a matter of creating more extension methods to the writeto-target-interface. NLog target authors can also create their own extension methods for their special target.


Sent from my Sony Xperia


---- Thomas Ardal wrote ----


I tried it and I think the new API works great and is a great fluent alternative to the existing API. Here are my comments.

AddRuleForAllLevels alternative

Maybe it's just me who cannot find it. But is there an alternative to AddRuleForAllLevels? Or should I simply leave out any level config if I want to log all levels? I tried this:

.FilterMinLevel(LogLevel.AllLevels)

But it fails with a compile error since AllLevels is an enumerable.

Target specific WriteTo

It would be nice with a more specific WriteTo. So rather than:

.WriteTo(new ConsoleTarget(...))

You could do:

.WriteTo(x => x.Console(...))

or:

.WriteTo.Console(...)

ThomasArdal:
You're right. I was thinking about `WriteTo` as something that could take more parameters than just the target. I see that it doesn't, which makes sense. A `WriteToConsole` provided by the core package and accepting the required properties there could be awesome. And as you write, third-party vendors could provide their `WriteToMyAwesomeService`. I like it 👍 The advantage of having this as either a lambda or as an extension method to the result of an empty `WriteTo` would be that you are sure that no one will implement a `PersistToMyAwesomeService` and make it inconsistent. But maybe that's just a theoretical example and something that can be fixed through guidelines, issues, and PRs.

snakefoot:
@ThomasArdal Yes the idea is to create a basic building block. And including WriteToConsole is on the high priority-list (Along with WriteToColoredConsole). Hopefully people will play around with it, and from that experience further fine-tuning of guidelines and API can be made. Thank you for the feedback, much appreciated.

ThomasArdal:
Not sure how much I contributed here 😂 But I like what you have done so far 👍

snakefoot:
> @ThomasArdal Not sure how much I contributed here

It is nice to get confirmation that the suggested API-method-names are not too misleading at first sight. And the door for adding new custom targets is easy to use.

snakefoot:
@304NotModified Why not for NLog 4.7.6. Since NLog 5.0 is still a long way from completion? And it will be easy to merge back to `dev`-branch.

snakefoot:
Accepted that it was not approved for NLog 4.7. Created #4323

Issue #4032, opened 2020-07-02
An ability to bypass InternalLogger initialization


estrizhok:
**NLog version**: 4.7.2
**Platform**: .Net 3.5 / .Net 4.5

Setting up logging from a non-default AppDomain add a noticable time overhead, mainly due to initialization of InternalLogger. 
For a short-lived application, that has typical execution time of about 2 seconds, this represents 10-15%.

```
15.53%   ConfigurationItemFactory..ctor  •  348 ms  •  NLog.Config.ConfigurationItemFactory..ctor(params Assembly[])
  15.04%   RegisterItemsFromAssembly  •  336 ms  •  NLog.Config.ConfigurationItemFactory.RegisterItemsFromAssembly(Assembly, String)
    11.86%   [Native code]  •  265 ms
      11.86%   InternalLogger..cctor  •  265 ms  •  NLog.Common.InternalLogger..cctor
        11.86%   Reset  •  265 ms  •  NLog.Common.InternalLogger.Reset
          11.46%   GetSetting  •  256 ms  •  NLog.Common.InternalLogger.GetSetting(String, String, T)
            11.46%   GetSettingString  •  256 ms  •  NLog.Common.InternalLogger.GetSettingString(String, String)
              ►11.46%   GetAppSettings  •  256 ms  •  NLog.Common.InternalLogger.GetAppSettings(String)
    ►2.32%   SafeGetTypes  •  52 ms  •  NLog.Internal.ReflectionHelpers.SafeGetTypes(Assembly)
    ►0.45%   ScanTypes  •  10 ms  •  NLog.Config.MethodFactory.ScanTypes(Type[], String)
    ►0.40%   PreloadAssembly  •  9 ms  •  NLog.Config.ConfigurationItemFactory.PreloadAssembly(Type[])
```

snakefoot:
Never heard of anyone using the ability to configure the `InternalLogger` from `app.config`, besides embedding the entire `NLog.config` as section inside `app.config`.

Of course a breaking change, but could be included in the coming NLog ver. 5. You are welcome to create a pull-request for the NLog-project

estrizhok:
Not sure I understood correctly. 

I am not configuring in the internal logging in any way. I am just configuring the logger itself from a non-default appDomain, and run into a problem that:
a) ConfigurationManager subsystem has got to re-initialize even if there was not specific app.config set for it (and it, therefore, inherited the one from the default appDomain);
[...]

PR #4130, opened 2020-10-09
Move InternalLogger settings reading at initialization to a SetupBuilder extension

noamyogev84:
Trying to help with #4032 🌷 
@snakefoot please review and let me know if i'm anywhere near...
In case it's a good direction, i'll work on some tests if needed.
:beers:

### DIFF
--- a/src/NLog/Common/InternalLogger.cs
+++ b/src/NLog/Common/InternalLogger.cs
@@ -41,7 +41,6 @@ namespace NLog.Common
     using System.Reflection;
     using NLog.Internal;
     using NLog.Time;
-    using NLog.Targets;
 
     /// <summary>
     /// NLog internal logger.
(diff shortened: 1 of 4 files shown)


snakefoot:
Thank you for the pull-request. It is close to what I was thinking. But I would like some changes:

Instead of having one `SetupFromSettings` then it should be two different methods:

- SetupFromAppSettings (Not available for NetCore)
- SetupFromEnvironmentVariables

I can understand the beauty of having the abstraction `ISettingsReader`, but the lookup code is like this:

```c#
InternalLogger.LogToConsole = settingsReader.GetSetting("nlog.internalLogToConsole", "NLOG_INTERNAL_LOG_TO_CONSOLE", false);
```

The 2 setting-names makes the abstraction very leaky. I prefer that ISettingsReader and SettingsReader are removed, and the setting lookup are just simple like this:

**SetupFromEnvironmentVariables():**
```c#
// Injecting the original value `InternalLogger.LogToConsole` as default-value
InternalLogger.LogToConsole = LookupEnvironmentVariable("NLOG_INTERNAL_LOG_TO_CONSOLE", InternalLogger.LogToConsole);
```

**SetupFromAppSettings():** (`#if !NETSTANDARD`)
```c#
// Injecting the original value `InternalLogger.LogToConsole` as default-value
InternalLogger.LogToConsole = LookupAppSettings("nlog.internalLogToConsole", InternalLogger.LogToConsole);
```

snakefoot:
To handle the type conversion, then maybe something like this might work:

```c#
T LookupEnvironmentVariable<T>(string environmentVariabel, T defaultValue)
{
      try
      {
         string stringValue = LookupStringValue(environmentVariabel)?.Trim();
         if (string.IsNullOrEmpty(stringValue))
             return defaultValue;

         var typeConverter = System.ComponentModel.TypeDescriptor.GetConverter(typeof(T));
         var environmentValue = (T)typeConverter.ConvertFrom(stringValue);
         return environmentValue;
      }
      catch
      {
         return defaultValue;
      }
}
```

noamyogev84:
OK. I will take a look.
You want all this code to be implemented in the extension method?

snakefoot:
> You want all this code to be implemented in the extension method?

You are welcome put the two extension methods in their own individual static-class-file if you feel they grow wildly.

The NLog project is also very keen on putting reusable-methods into NLog.Internal-namespace. Ex. `GetSafeEnvironmentVariable`

snakefoot:
@noamyogev84 Should the pull-request be closed, or is it just a minor delay before completion?

noamyogev84:
i misht return to that soon :)

snakefoot:
Superseeded by #4184

Issue #4129, opened 2020-10-06
Excessive locking in Target.PrecalculateVolatileLayoutsWithLock


zvolkov:
Hi, we are on NLog 4.7.2 on .net core 3.1. We were profiling our asp.net app for some other issues and noticed that NLog's Target creates a lot of thread contention at PrecalculateVolatileLayoutsWithLock method when our apps runs under high load.

When we log we do our own formatting, so we never pass params to NLog, but we do populate the event context properties with quite a few items. We log to the async wrapper on top of the file target. Our layouts are pretty basic.

Can you give us some directions on where we should start looking in order to diagnose this. Why is PrecalculateVolatileLayoutsWithLock called a lot, to begin with. Why not its concurrent sibling? What co we do to optimize this?

Thanks a lot!

snakefoot:
@zvolkov PrecalculateVolatileLayoutsWithLock is activated when using LayoutRenders not marked with `[ThreadSafe]`-attribute.

Maybe you can show your NLog-config ?

zvolkov:
```
<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
      autoReload="true" throwExceptions="false">
    <targets>
        <target xsi:type="AsyncWrapper" name="logfile"
                batchSize="2000" timeToSleepBetweenBatches="100" queueLimit="20000" overflowAction="grow">
            <target name="applogjson" xsi:type="File"
                    fileName="${environment:variable=XYZ_APPLOGS}/${shortdate}.applogjson">
                <layout xsi:type="JsonLayout">
                    <attribute name="timestamp" layout="${date:universalTime=false:format=yyyy-MM-ddTHH\:mm\:ss.FFFFFFFK}" />
                    <attribute name="sequenceNumber" layout="${event-context:item=SequenceNumber}" />
                    <attribute name="level" layout="${level:upperCase=true}" />
                    <attribute name="saneId" layout="${event-context:item=SaneId}" />
                    <attribute name="traceId" layout="${event-context:item=TraceId}" />
                    <attribute name="requestId" layout="${event-context:item=RequestId}" />
                    <attribute name="logger" layout="${logger}" />
                    <attribute name="logId" layout="${event-context:item=LogId}" />
                    <attribute name="message" layout="${message}" />

                    <!-- stateful application -->
                    <attribute name="sessionId" layout="${event-context:item=SessionId}" />
[...]

PR #4132, opened 2020-10-13
Improve thead concurrency with Cache wrapper

snakefoot:
CachedLayoutRendererWrapper - Added ThreadSafe attribute

Improve thread concurrency issues reported with #4129

### DIFF
--- a/tests/NLog.UnitTests/LayoutRenderers/Wrappers/CachedTests.cs
+++ b/tests/NLog.UnitTests/LayoutRenderers/Wrappers/CachedTests.cs
@@ -119,11 +119,11 @@ namespace NLog.UnitTests.LayoutRenderers.Wrappers
         [Fact]
         public void CachedSecondsTimeoutTest()
         {
-            SimpleLayout l = "${guid:cachedSeconds=10}";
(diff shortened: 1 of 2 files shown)


tests/NLog.UnitTests/LayoutRenderers/Wrappers/CachedTests.cs
```diff
@@ -119,11 +119,11 @@ public void CacheKeyTest()
         [Fact]
         public void CachedSecondsTimeoutTest()
         {
-            SimpleLayout l = "${guid:cachedSeconds=10}";
+            SimpleLayout l = "${guid:cachedSeconds=60}";
```
304NotModified, line 122: why does the test need a change? 

Not that I'm against it, but don't see the reason for it.

snakefoot: It was easier to debug, when having 60 secs instead of 10 secs.

snakefoot:
Forced pushed because I discovered a loop-hole in the intiail cache-validation, where racing could lead to null-value.

PR #4138, opened 2020-10-19
ProcessNameLayoutRenderer - Use ProcessName instead of MainModule.FileName

snakefoot:
Most people will expect the ProcessName instead of the ProcessFileName.

I'm guessing for 99 out of 100, then the result is the same.

### DIFF
--- a/src/NLog/Internal/AppEnvironmentWrapper.cs
+++ b/src/NLog/Internal/AppEnvironmentWrapper.cs
@@ -57,7 +57,7 @@ namespace NLog.Internal.Fakeables
         /// <inheritdoc />
         public string CurrentProcessFilePath => _currentProcessFilePath ?? (_currentProcessFilePath = LookupCurrentProcessFilePath());
         /// <inheritdoc />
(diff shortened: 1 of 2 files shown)


304NotModified:
> Most people will expect the ProcessName instead of the ProcessFileName.

Agree on this, but what about an option to use the ProcessFileName? Then it will be less of a breaking change, which is always better?

snakefoot:
> Agree on this, but what about an option to use the ProcessFileName? 

Actually think that ProcessName is the ProcessFileName except when running on weird platforms:

https://referencesource.microsoft.com/#system/services/monitoring/system/diagnosticts/Process.cs,872

So I think my worry was unnecessary.

304NotModified:
> Actually think that ProcessName is the ProcessFileName except when running on weird platforms:
> 
> https://referencesource.microsoft.com/#system/services/monitoring/system/diagnosticts/Process.cs,872
 


👍 thx!

snakefoot:
#4182 Reported that using ProcessName (instead of MainModule.FileName) avoided first-chance exception, when running medium-trust.

PR #4171, opened 2020-11-12
ConditionLayoutExpression - Skip allocating StringBuilder for every condition check

snakefoot:
This filter will now only allocate a StringBuilder once (But will still allocate a string for every LogEvent with Exception):

```xml
        <logger name="*" minLevel="Info" writeTo="exceptionFile">
          <filters defaultFilterResult="Log">
            <when condition="'${exception:format=type}' == ''" action="Ignore" />
          </filters>
        </logger>
```


304NotModified:
looks good! Thanks!

PR #4172, opened 2020-11-12
AsyncTaskTarget - Flush when buffer is full should not block forever

snakefoot:
bug is about using `overflow="Block"` and flushing when the queue is completely full:


304NotModified:
thx!

Issue #3481, opened 2019-06-12
XSD, validation of wrappers


304NotModified:
XSD allows this, but it is nonsense?

```xml
<target xsi:type="FilteringWrapper" name="default" >
  <target xsi:type="Console"  />
  <target xsi:type="File"  />
</target>
```

related https://github.com/NLog/NLog/issues/3473

304NotModified:
If someone is good with XSD, help would be great!

VHellin:
Hi, 
Can you describe the non-sense ? Do you want your wrapper to contain only 1 target ? Is this applicable only for the FilteringWrapper ? Or do you want this for all kinds of wrappers extending WrapperTargetBase ?

snakefoot:
@VHellin Thanks for the heads up about the Wiki. Have now update it.

The `WrapperTargetBase` only supports a single wrapped-target. So it is non-sense to declare two targets.

304NotModified:
The specs of this issue are really not good, so gonna close this for now. I don't think this is easy to fix, as this should be valid with the XSD:


```xml
  <target xsi:type="FallbackGroup" name="String" returnToFirstOnSuccess="Boolean">
    <target xsi:type="wrappedTargetType" ... />
    <target xsi:type="wrappedTargetType" ... />
    ...
    <target xsi:type="wrappedTargetType" ... />
  </target>
```

PR #4173, opened 2020-11-13
Fix XSD issue #3481

VHellin:
Issue #3481 : Limit occurence of Target (or his child classes) to 1 into wrapper.


304NotModified:
Hi @VHellin!

Thanks for the code suggestion!

Unfortunately #3481 is poorly written (my bad) and this won't work as this is OK and should be accepted by the XSD:

```
  <target xsi:type="FallbackGroup" name="String" returnToFirstOnSuccess="Boolean">
    <target xsi:type="wrappedTargetType" ... />
    <target xsi:type="wrappedTargetType" ... />
    ...
    <target xsi:type="wrappedTargetType" ... />
  </target>
```

VHellin:
OK. No problem. Just remove my PR.

Issue #2732, opened 2018-05-21
Move FileTarget.FileCompressor to NLog.FileCompressorWrapper


snakefoot:
Remove the following dependencies from NLog-core:

`<Reference Include="System.IO.Compression" />`

See also #2729, #2296 and #1314

For NLog 6.0

snakefoot:
Maybe also add the option to flip the NTFS-compression flag on file after it has been archived:

https://stackoverflow.com/questions/624125/compress-a-folder-using-ntfs-compression-in-net

https://blogs.msdn.microsoft.com/oldnewthing/20110622-00/?p=10363

noamyogev84:
I'll take it :) @snakefoot

snakefoot:
@noamyogev84 I guess this up-for-graps is a little more tricky, since you probably have to implement this first:  #1249 (And it requires one to invent a great API that `FileCompressorWrapper` can hook into).

noamyogev84:
Got it.
Any other place I can contribute?

snakefoot:
Guess #4032 could be a good candidate for `first-issue`.

snakefoot:
Closed with #5673. When someone feels excited, then they are welcome to think/design a clever way of extending the new basic archive-logic with zip-compression (Using inheritance since it will not be part of NLog-core-package)

Issue #1249, opened 2016-02-18
Add File Archive Event (+ other file actions)


Sharpiro:
Type: Feature request/Question
NLog version: 4.2.3
Platform: .Net 4.5
- Why do I/We need it?

My requirement is that I need to be able to know exactly when that file has been successfully copied over to the network directory so that I can then either begin using it or perhaps move it to another directory on the same drive that holds completed archived files.  The goal is to ensure that no file is touched that is still in progress of being copied over, which is more likely in the event of an unusually large log file.

I was wondering if there is currently any way to determine if a file being archived to a remote directory has completed copying over, perhaps an event that could be subscribed to.  If not, is there a workaround, or would this be a good idea for a feature request?

304NotModified:
Hi,

there aren't events right now. Also the needed methods aren't `virtual`. So the only option for a work around would be: copy [the class](https://github.com/NLog/NLog/blob/master/src/NLog/Targets/FileTarget.cs) and add it.

We do like the events idea. I think it's then important to add a more events like when deleting a file, etc. Do you agree?

StevenLiekens:
File action events would also be great for when `LayoutWithHeaderAndFooter` is too limited. I would love to be able to write header and footer log messages from application code instead of having to statically define a header/footer layout.

304NotModified:
@StevenLiekens You can use (limited) layout renderers in footer /header. And I would write my own target wrapper in your case.

304NotModified:
If we like to boost this, please propose a list of events we need to implement.

farukhsharipov:
+Add event when file is being rolled.

304NotModified:
> if we like to boost this, please propose a list of events we need to implement.

;)

snakefoot:
Note that the pending #1993 includes a heavy refactoring of the FileTarget archiving logic.

adyshimony:
I can help with it, but we need first to decide if this is the best solution to our project.

I suggest the following events:

LogBeforeArchiveEvent - when the main log reach the size limit / day limit.
LogAfterArchiveEvent - when the main log is archived.
MaxArchiveReachedEvent - we have reach max archive number.
ArchiveFileDeleteEvent - when the first archive clear the space for the news one.

farukhsharipov:
@adyshimony events look good! Anyone else has any suggestions or comments?

304NotModified:
> +Add event when file is being rolled.

Which events will be triggered in this case?

[...]

PR #4189, opened 2020-12-03
Use the archive file name format for the compressed file when using compression

TalAloni:
This PR sets the entry name for the filename inside the zip (when compression is enabled) according to the archiveFileName format.

**Motivation:** I have recently enabled compression and found myself in a situation where I cannot mass-extract archived log files to the same directory because they are all using the same filename inside the archive. it seems to me that the rules for the archive filename should also apply to the entry name inside the archive.
Simply put, I expect that once I extract the zip files, I will have the same filenames that I had before enabling compression.

**Additional information:** 
I have worked with the following configuration
```
<target name="file" xsi:type="File" fileName="${basedir}/logs/log-latest.csv" archiveFileName="${basedir}/logs/log.{#####}.csv">
```
Now I wish to enable compression using the following configuration:
```
<target name="file" xsi:type="File" fileName="${basedir}/logs/log-latest.csv" archiveFileName="${basedir}/logs/log.{#####}.zip" enableArchiveFileCompression="true">
```


snakefoot:
This change makes a lot of sense, but is ofcourse a breaking change, and will have to wait for NLog 5.0

I guess the FileTarget can eat one more bullet, but I think it would be nice if the FileTarget was stripped from any copy/move/delete/compress-logic. Instead the file-manipulation-logic should be implemented using one (or more) FileTargetWrapper. See also #2732 and  #1249

snakefoot:
Closing and reopening to rebuild on Travis.

snakefoot:
Besides being a behavior breaking change, then it also breaks the `IFileCompressor`-interface.

I guess very few have made override of the `IFileCompressor`-interface, but would prefer if the upgrade-path was less breaking. Thus introducing a new interface, and make a wrapper for the old interface-property and mark the old interface as obsolete.

TalAloni:
Thanks,
There's already a FileCompressor property of type IFileCompressor.
It seems to me you're suggesting we add an "IArchiveFileCompressor" that inherits from IFileCompressor, and then check if we're dealing with IArchiveFileCompressor or the now obsolete IFileCompressor.

This seems a bit excessive, but I'll do it if you think it's necessary.
IMO given that not many inherited the interface maybe it's best to just reduce noise - it will be simple enough for implementing classes to add the new parameter to the implementation - given that it is targeted for a major version I think that's acceptable.
I don't think that it will be clearly understood that you can implement IArchiveFileCompressor if the property type is IFileCompressor - so I find the approach you're describing less appealing.

snakefoot:
I was thinking 3 items:
- IArchiveFileCompressor - New interface
- LegacyArchiveFileCompressor - Implements new interface and takes legacy IFileCompressor as input-parameter
- ZipArchiveFileCompressor - Implements new interface

The do the following:
- Mark IFileCompressor as obsolete
- Mark FileTarget FileCompressor-property as obsolete
- Change FileTarget FileCompressor-property implementation. Let the setter allocate LegacyArchiveFileCompressor and give the propery-value as constructor-parameter. Let the getter return the IFileCompressor inside the LegacyArchiveFileCompressor.

TalAloni:
I see,
We can also just change the implementation of ZipArchiveFileCompressor - so it would contain the `Path.GetFileNameWithoutExtension(archiveFileName) + Path.GetExtension(fileName)`
logic - this way the interface will remain the same - if that's important to you than maybe we should settle for this simple solution.
Let me know what you think is best. thanks!

snakefoot:
Like that work-around. Replacing interface and guiding users to the new-interface generates a lot of code-bloat.

TalAloni:
Great, I've made the change - the interface remain unchanged. this is tested and working as expected.

snakefoot:
Very elegant solution. I like it.

Tried to figure out whether there was any special reasons for the initial solution (with static filename). But it seems it was just hacked together until it worked, without much consideration about entry-filename. But still breaking change so waiting for NLog 5.0

snakefoot:
@TalAloni Could you extend one of the existing unit-test to verify that the entry-filename is as expected? (Then you will also have better gurantee that it doesn't break in the future).

snakefoot:
@TalAloniI would prefer that `AssertZipFileContents` was used for the unit-test to verify the contents of final zip-file (Having the expected the Entry-Name)

TalAloni:
This would mean adding an additional parameter to AssertZipFileContents (archiveFileName or entryName) and update a few dozens existing tests.
Instead we can add a new AssertZipFileEntryName which would be less obtrusive, which do you prefer?

snakefoot:
Was thinking you could do the exact same trick you did in `ZipArchiveFileCompressor.cs` to extract the expected Entry-Name from the input-filename. (Where EntryName has `.txt`-file-extension).

TalAloni:
Currently AssertZipFileContents only gets the zip file name - it does not have a way to check if the entry name (including the extension) inside the zip match the expected entry name without an additional parameter (the expected entry name or the log file name before archiving).
All we can do without adding an additional parameter is to check the entry name without the extension.

snakefoot:
> All we can do without adding an additional parameter is to check the entry name without the extension.

Pretty sure all test-cases use `.txt` as file-extension, so just go with that as hardcoded value. When someone needs a test with a different file-extension, then it can be improved.

snakefoot:
Curious with this failing test:

```
      Assert.Equal() Failure
      Expected: archive.txt
      Actual:   tmpE9E9.tmp
      Stack Trace:
        tests\NLog.UnitTests\NLogTestBase.cs(189,0): at NLog.UnitTests.NLogTestBase.AssertZipFileContents(String fileName, String contents, Encoding encoding)
        tests\NLog.UnitTests\Targets\FileTargetTests.cs(796,0): at NLog.UnitTests.Targets.FileTargetTests.ArchiveFileOnStartTests(Boolean enableCompression, Boolean customFileCompressor)
```

snakefoot:
Also looks like the rolling-archive-mode is not working well with this feature. When "rolling" the zip-files (renames it) then it doesn't roll the entry-name inside the zip-file. Guess the filename should remain static (the old way) when `ArchiveNumbering.Rolling`

TalAloni:
Unfortunately we no longer have access to ArchiveNumbering since we moved the logic to the ZipArchiveFileCompressor in order to avoid changing the interface.

TalAloni:
I can easily fix the failing test by appending .txt to Path.GetTempFileName()

TalAloni:
I think the original PR made more sense, especially given ArchiveNumbering.Rolling
and I also think it's OK to expand interfaces when releasing a major version, especially ones that are rarely implemented

snakefoot:
> I can easily fix the failing test by appending .txt to Path.GetTempFileName()

That will be good start.

> Unfortunately we no longer have access to ArchiveNumbering since we moved the logic to the ZipArchiveFileCompresso

Maybe something like this:

```c#
        public ArchiveNumberingMode ArchiveNumbering
        {
            get => _archiveNumbering;
            set
            {
                if (_archiveNumbering != value)
                {
                    _archiveNumbering = value;
                    ResetFileAppenders("ArchiveNumbering Changed"); // Reset archive file-monitoring
                    (FileCompressor as ZipArchiveFileCompressor)?.SetArchiveNumbering(value);
                }
            }
        }
        private ArchiveNumberingMode _archiveNumbering;
```

TalAloni:
The proposed change to ZipArchiveFileCompressor feels counter productive, as this would only be available to the internal ZipArchiveFileCompressor while IFileCompressor would be less flexible.
I stand by my original PR - I think that it's more appropriate to have the entry name come from the consumer of the interface.
I think that you put too much emphasis on maintaining the interface signature and you shouldn't let that from interfering with the right design choice - ask yourself how the interface would have looked like if you would have started from scratch, and that's what the interface should become. it would be best to avoid making one odd workaround after another.

snakefoot:
> and I also think it's OK to expand interfaces when releasing a major version,

Think my goal is to remove a lot of the logic from FileTarget, and instead have some lifetime events, that a wrapper can react to. Ex. archive-event that triggers compression in the background, instead of blocking the application. Or a wrapper that wraps newly create FileStream-objects so they actually creates a gzip-stream. Etc.

Would also like to split FileTarget into two, so it becomes a vanilla FileTarget without ConcurrentWrites-support (Thus no FileWatcher or global mutex, etc.). And a ConcurrentFileTarget in a seperate nuget-package.

Like to post-pone breaking changes until they are really needed or gives a huge benefit. And if possible provide an easy migration path without too many runtime-errors.

snakefoot:
> it would be best to avoid making one odd workaround after another.

The "problem" right now is that an unit-test has revealed that the pull-request is not working well with `ArchiveNumbering.Rolling`. You are ofcourse welcome to go back to my [original suggestion](https://github.com/NLog/NLog/pull/4189#issuecomment-761881813). But right now I prefer the path of just making `ZipArchiveFileCompressor` work, and if it needs a hint, then I'm okay with that.

TalAloni:
OK, I added a hint to ZipArchiveFileCompressor like you prefer,
AssertZipFileContents will need an additional parameter take into account ArchiveNumberingMode. this means updating a few dozens tests.

snakefoot:
Not sure how 3 locations turns into dozens, but yes having the ability to disable entry-name validation for `RollingArchiveTests` should be enough.

src/NLog/Targets/ZipArchiveFileCompressor.cs
```diff
@@ -50,17 +50,24 @@ internal class ZipArchiveFileCompressor : IFileCompressor
         /// </summary>
         public void CompressFile(string fileName, string archiveFileName)
         {
+            string entryName = IsEntryNameInferredFromArchiveFileName ? (Path.GetFileNameWithoutExtension(archiveFileName) + Path.GetExtension(fileName)) : fileName;
```
snakefoot, line 53: Think it should be (Missing `Path.GetFileName`):
`string entryName = IsEntryNameInferredFromArchiveFileName ? (Path.GetFileNameWithoutExtension(archiveFileName) + Path.GetExtension(fileName)) : Path.GetFileName(fileName);`

TalAloni: Of course - thank you

TalAloni:
Sorry, I miscounted the usages of AssertZipFileEntryName.
I have implemented the change in the way you preferred, all tests pass.

snakefoot:
Very happy with the updated unit tests. The unit tests for the FileTarget are rather gnarly. So thank you for your persistence on making them green again with the extra validation.


Sent from my Sony Xperia


---- TalAloni wrote ----


Sorry, I miscounted the usages of AssertZipFileEntryName.
I have implemented the change in the way you preferred, all tests pass.

snakefoot (approved):

snakefoot:
I can now see that I have given wrong advice. The FileCompressor-property is a static property, so having a single FileTarget changing properties on this static-instance will not work well, when using multiple FileTargets in the same NLog.config.

So the entry-name must be given as input-parameter when calling the IFileCompressor. To provide safe passage, then a V2-interface must be introduced (that includes the method with entry-name-parameter). And then call this people when the static-instance implements the V2-interface (else call the old interface-method).

TalAloni:
"So the entry-name must be given as input-parameter when calling the IFileCompressor"
Yes, this approach is healthier.

"To provide safe passage, then a V2-interface must be introduced"
Again, I think it's too cumbersome, I think it's extremely rare for users to implement this interface, we add a lot of noise for background compatibility that nobody really care for, and that could easily be updated to the new interface.
I will do it, but only to please you, I think that it's a messy and confusing approach. using the existing interface and property would be much cleaner and easier for most users, if not all.

snakefoot:
This will also do:

```c#
if (FileCompressor is ZipArchiveFileCompressor defaultCompressor)
   defaultCompressor.CompressFile(fileName, archiveFileName, entryFileName);
else 
   FileCompressor.CompressFile(fileName, archiveFileName);
```

TalAloni:
I've made the requested changes. I've added a new interface extending the existing interface to improve extendibility.

TalAloni:
Thanks,
The idea beyond making IArchiveFileCompressor public is to allow developers to set FileCompressor to an IArchiveFileCompressor implementation, this way they be able to have access to the 'entryName' argument.

snakefoot:
The NLog FileTarget has so many door-knobs that it is difficult for normal users to figure out how to use them. Several of the door-knobs are counter intuitive, and make normal users think the NLog FileTarget is broken. Adding more public door-knobs will not improve the situation.

Like I said initially "I guess the FileTarget can eat one more bullet, but I think it would be nice if the FileTarget was stripped" (down).

snakefoot:
Sorry for taking so long to get this pull-request merged. But very grateful for the unit-tests, they have been missing since the zip-feature was introduced.

Maybe one day a champion will arrive that will carve the NLog FileTarget into smaller pieces, so it will be easier to maintain and with improved platform compability.

TalAloni:
Thanks for the great and very useful project!

snakefoot:
Updated the wiki: https://github.com/NLog/NLog/wiki/File-target

PR #4196, opened 2020-12-09
Added a new ExcludeEmptyProperties to JsonLayout as to avoid GDC/MDC/MLDC properties with null or empty values from output

pruiz:
In order to reduce the size of generated json messages I've added a new parameter (ExcludeEmptyProperties) to JsonLayout, by which any GDC/MDC/MDLC property with a null or (string) empty value would be omitted from the generated json string.

A side from implementing our own JsonLayout, I've found no other way of excluding such properties, as the names of the properties which will be null are not known in advance.

Regards
Pablo

### DIFF
--- a/tests/NLog.UnitTests/Layouts/JsonLayoutTests.cs
+++ b/tests/NLog.UnitTests/Layouts/JsonLayoutTests.cs
@@ -45,6 +45,7 @@ namespace NLog.UnitTests.Layouts
     public class JsonLayoutTests : NLogTestBase
     {
         private const string ExpectedIncludeAllPropertiesWithExcludes = "{ \"StringProp\": \"ValueA\", \"IntProp\": 123, \"DoubleProp\": 123.123, \"DecimalProp\": 123.123, \"BoolProp\": true, \"NullProp\": null, \"DateTimeProp\": \"2345-01-23T12:34:56Z\" }";
+        private const string ExpectedExcludeEmptyPropertiesWithExcludes = "{ \"StringProp\": \"ValueA\", \"IntProp\": 123, \"DoubleProp\": 123.123, \"DecimalProp\": 123.123, \"BoolProp\": true, \"DateTimeProp\": \"2345-01-23T12:34:56Z\", \"NoEmptyProp4\": \"hello\" }";
 
         [Fact]
         public void JsonLayoutRendering()
@@ -442,6 +443,28 @@ namespace NLog.UnitTests.Layouts
         }
 
         [Fact]
+        public void ExcludeEmptyJsonProperties()
+        {
+            var jsonLayout = new JsonLayout()
+            {
+                IncludeEventProperties = true,
+                ExcludeEmptyProperties = true
+            };
+
+            jsonLayout.ExcludeProperties.Add("Excluded1");
+            jsonLayout.ExcludeProperties.Add("Excluded2");
+
+            var logEventInfo = CreateLogEventWithExcluded();
+            logEventInfo.Properties.Add("EmptyProp", "");
+            logEventInfo.Properties.Add("EmptyProp1", null);
+            logEventInfo.Properties.Add("EmptyProp2", new DummyContextLogger() { Value = null });
+            logEventInfo.Properties.Add("EmptyProp3", new DummyContextLogger() { Value = "" });
+            logEventInfo.Properties.Add("NoEmptyProp4", new DummyContextLogger() { Value = "hello" });
+
+            Assert.Equal(ExpectedExcludeEmptyPropertiesWithExcludes, jsonLayout.Render(logEventInfo));
+        }
+
+        [Fact]
         public void IncludeAllJsonPropertiesMaxRecursionLimit()
         {
             var jsonLayout = new JsonLayout()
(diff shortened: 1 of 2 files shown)


snakefoot:
Instead of doing this:

```c#
foreach (var prop in propertiesList)
{
         if (ExcludeEmptyProperties && string.IsNullOrEmpty(prop.Value?.ToString()))
             continue;

         AppendJsonPropertyValue(...)
}
```

Then I think it should be like this (Skip the ToString()-operation):

```c#
foreach (var prop in propertiesList)
{
         if (ExcludeEmptyProperties && IsValueNullOrEmpty(prop.Value))
             continue;

         AppendJsonPropertyValue(...)
}

private static bool IsValueNullOrEmpty(object propertyValue)
{
     return string.Empty.Equals(propertyValue) || ReferenceEquals(propertyValue, null);
}
```

pruiz:
Hi @snakefoot

We tend to use the pattern of overriding ToString on the objects attached to contextual logging scopes, like:

```csharp
                private class UserAddressLogContext
                {
                        public override string ToString()
                        {
                                return HttpContext.Current?.Request?.UserHostAddress;
                        }
                }
```

This requires invoking ToString to assert wether the context does have a value or is null/empty.

I could try to avoid the call to ToString(), by reflecting the object and inspect if ToString has been overriden, but this sounds more expensive than the actual ToString call.

snakefoot:
Ok then I think you need to implement something similar to what happens in `RenderAppendJsonPropertyValue` where return false means unwind. Since you are working with non-string-objects.

So when using ExcludeEmptyProperties=true then it should capture `int beforeAttribLength = sb.Length` and if the output in the StringBuilder is double-quotes `""` or `null`, then it should unwind back `sb.Length = beforeAttribLength`.

So it becomes like this:

```c#
         if (ExcludeEmptyProperties)
              AppendJsonPropertyValueWhenNotEmpty(...)
         else
              AppendJsonPropertyValue(...)

         private void AppendJsonPropertyValueWhenNotEmpty(...)
         {
               if (propertyValue == null)
                      return;

               int beforeAttribLength = sb.Length;
               AppendJsonPropertyValue(...)
      
               if (sb.Length > 2 && sb[sb.Length-1] == '\"' && sb[sb.Length-2] == '\"')
                    sb.Length = beforeAttribLength;
         }
```

pruiz:
@snakefoot I've reworked the code as to implement it the way you suggested, by rewinding the StringBuilder when an empty property was emitted.

pruiz:
Rebased branch.

tests/NLog.UnitTests/Layouts/JsonLayoutTests.cs
```diff
@@ -441,6 +442,28 @@ public void IncludeAllJsonProperties()
             Assert.Equal(ExpectedIncludeAllPropertiesWithExcludes, jsonLayout.Render(logEventInfo));
         }
 
+        [Fact]
+        public void ExcludeEmptyJsonProperties()
+        {
+            var jsonLayout = new JsonLayout()
+            {
+                IncludeAllProperties = true,
```
snakefoot, line 450: Please change from `IncludeAllProperties` (deprecated in NLog 5.0) to `IncludeEventProperties`

snakefoot:
Thank you for the update. Like this better than calling ToString upfron for all objects. But maybe you could remove the duplicated logic:

```c#
         if (ExcludeEmptyProperties && propertyValue == null)
             continue;
```

And move it here:

```c#
        private void AppendJsonPropertyValue(string propName, object propertyValue, string format, IFormatProvider formatProvider, MessageTemplates.CaptureType captureType, StringBuilder sb, bool beginJsonMessage)
        {
           if (ExcludeEmptyProperties && propertyValue == null)
                return;

            var initialLength = sb.Length;

            BeginJsonProperty(sb, propName, beginJsonMessage);
            if (MaxRecursionLimit <= 1 && captureType == MessageTemplates.CaptureType.Serialize)
            {
                // Overrides MaxRecursionLimit as message-template tells us it is safe
                JsonConverter.SerializeObjectNoLimit(propertyValue, sb);
            }
            else if (captureType == MessageTemplates.CaptureType.Stringify)
            {
                // Overrides MaxRecursionLimit as message-template tells us it is unsafe
                int originalStart = sb.Length;
                ValueFormatter.FormatValue(propertyValue, format, captureType, formatProvider, sb);
                PerformJsonEscapeIfNeeded(sb, originalStart, EscapeForwardSlash);
            }
            else
            {
                JsonConverter.SerializeObject(propertyValue, sb);
            }

            if (ExcludeEmptyProperties && (sb[sb.Length-1] == '"' && sb[sb.Length-2] == '"'))
            {
                sb.Length = initialLength;
            }
        }
```

pruiz:
@snakefoot applied changes as requested.

snakefoot:
Closing and reopening to trigger re-build with Travis.

304NotModified:
@snakefoot should I give you Travis and Azure access so you could retrigger the builds more easy?

snakefoot:
> @304NotModified should I give you Travis and Azure access so you could retrigger the builds more easy?

No it is ok. It is mostly Azure DevOps Environment that is failing, and I have access to retry that build. Travis and AppVeyor are usually stable.

snakefoot (approved):
Thank you for the updates. It looks good now.

snakefoot:
@304NotModified I guess it could be merged into master for faster release, since not the biggest change. Or should it wait for NLog 5.0 ?

304NotModified:
I'm fine for master for this :)

snakefoot:
@pruiz Interested in getting this included with NLog 4.7.7 or will it be okay that it is delayed until NLog 5.0 is released ? (Will probably take some months)

pruiz:
@snakefoot as we are using 4.x on a few projects, having it on next 4.x release would be fantastic.

304NotModified:
then we need to rebase it on master. Maybe squash those 4 commits into 1 first. :)

pruiz:
@304NotModified changes squashed into a single commit.

snakefoot:
@pruiz Guess you need to make this pull-request against the master-branch instead of the dev-branch.

pruiz:
@snakefoot IMHO, it will be easier if you merge this branch against dev, and I push another pull-req against master instead, otherwise there will be conflicts with stuff like IncludeEventProperties being named different on master vs dev.

PR #4207, opened 2020-12-17
Skip allocation of SingleCallContinuation when ThrowExceptions = false

snakefoot:
This should reduce NLog memory allocations a lot. Instead of allocating 3 objects (LogEventInfo + SingleCallContinuation +  SingleCallContinuation.Function-delegate-instance) then it will just be the actual LogEventInfo.

Also a small bonus speed optimization on many-core-machines, as it can skip an interlocked-operation for every logevent (skip unnecessary cpu-cache synchronization).

### DIFF
--- a/src/NLog/Internal/SingleCallContinuation.cs
+++ b/src/NLog/Internal/SingleCallContinuation.cs
@@ -42,6 +42,8 @@ namespace NLog.Internal
     /// </summary>
     internal class SingleCallContinuation
     {
+        internal static readonly AsyncContinuation Completed = new SingleCallContinuation(null).CompletedFunction;
+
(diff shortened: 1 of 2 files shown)


304NotModified:
cool!

snakefoot:
When I tested with NetCoreApp3.1 and compared NLog 4.6.8 with NLog 4.7.7 then it reduced memory-allocation with 20 pct.

When I tested with Net5.0 and compared NLog 4.6.8 with NLog 4.7.7 then it reduced memory-allocation with 10 pct. Somehow Net5.0 was able to optimize away some of the delegate-allocation in NLog 4.6.8, so the optimization in NLog 4.7.7 had less impact.

Issue #4220, opened 2020-12-29
Error Error has been raised. Exception: System.NullReferenceException: Object reference not set to an instance of an object.


AtaTunca:
Hi!

Type : Bug
NLog version: 4.5.11
Platform: .Net 4.5

I got the error below on my internal log and keeps writing same log continuously. And after geting this error it stopped writing logs.

**internal.log:**
```
2020-12-22 21:12:00.3284 Error Error has been raised. Exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at NLog.Internal.AppendBuilderCreator..ctor(StringBuilder appendTarget, Boolean mustBeEmpty)
   at NLog.Layouts.Layout.RenderAppendBuilder(LogEventInfo logEvent, StringBuilder target, Boolean cacheLayoutResult)
   at NLog.Targets.FileTarget.RenderFormattedMessage(LogEventInfo logEvent, StringBuilder target)
   at NLog.Targets.FileTarget.RenderFormattedMessageToStream(LogEventInfo logEvent, StringBuilder formatBuilder, Char[] transformBuffer, MemoryStream streamTarget)
   at NLog.Targets.FileTarget.Write(LogEventInfo logEvent)
   at NLog.Targets.Target.Write(AsyncLogEventInfo logEvent)
```

**Current NLog config**
```xml
<?xml version="1.0" encoding="utf-8" ?>
<!--<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" throwExceptions="false" autoReload="true" internalLogLevel="Warn">-->

<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <!--<extensions>
    <add assembly="NLog.Targets.ElasticSearch"/>
  </extensions>-->

  <targets async="true">
    <!--<target xsi:type="ElasticSearch"
            name="elastic"
            uri="http://localhost:aaa/"
            index="aaa-log-${date:format=yyyy.MM.dd}"
            includeAllProperties='true'>
      <field name='messageTemplate' layout='${message:raw=true}' />
    </target>-->

    <target xsi:type="File" name="f" fileName="./aaa-logs/${date:format=yyyy-MM}/${shortdate}.log" layout="${longdate} ${uppercase:${level}} ${message}" />
  </targets>

  <rules>
[...]

PR #4222, opened 2020-12-30
StringBuilderExt ClearBuilder with better handling of low memory

snakefoot:
See also #4220

### DIFF
--- a/src/NLog/Internal/StringBuilderExt.cs
+++ b/src/NLog/Internal/StringBuilderExt.cs
@@ -192,11 +192,22 @@ namespace NLog.Internal
         /// <param name="builder"></param>
         public static void ClearBuilder(this StringBuilder builder)
         {
+            try
+            {
 #if !SILVERLIGHT && !NET3_5
-            builder.Clear();
+                builder.Clear();
 #else
-            builder.Length = 0;
+                builder.Length = 0;
 #endif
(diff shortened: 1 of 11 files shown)


src/NLog/Internal/StringBuilderExt.cs
```diff
@@ -192,11 +192,22 @@ public static void AppendXmlDateTimeRoundTrip(this StringBuilder builder, DateTi
...
-            builder.Clear();
+                builder.Clear();
 #else
-            builder.Length = 0;
+                builder.Length = 0;
 #endif
+            }
+            catch
+            {
+                // Default StringBuilder Clear() can cause the StringBuilder to re-allocate new internal char-array
```
304NotModified, line 205: where could you see this behavior?

snakefoot: https://referencesource.microsoft.com/#mscorlib/system/text/stringbuilder.cs,498 <-- When doing `Length = 0` then it will drop all chained StringBuilder-buffers and instead allocate a single char-buffer with total capacity. This behavior doesn't occur when doing `Remove(0, Length - 1)`.

304NotModified (approved):

snakefoot:
Thanks for the review.

PR #4223, opened 2021-01-01
Introduced .NetFramework ver. 4.6 and removed ver. 4.0

### LINKED ISSUE #3817: Add net46 Target Framework

### LINKED ISSUE #3979: Proposal: Drop support .NET 4.0 (and add 4.6)
From https://docs.microsoft.com/en-us/lifecycle/faq/dotnet-framework

> .NET Framework 4: .NET Framework 4 did not ship in any operating system. Support for the .NET Framework 4 on Windows Server 2003 SP2 ended on July 14, 2015, and support on all other operating systems ended on January 12, 2016.

snakefoot:
Ensure ScopeContext also works on Net46 with AsyncLocal. And resolve #3979 and also resolves #3817 

Removed self-invented defines (NET3_5 + NET4_0 + NET5_0 + NETSTANDARD1_0) and instead replaced by official defines (NET35, NET40, NET45) and also removed the special use of NET4_5 that was also (ab)used for NETSTANDARD-builds. This should make it easier for outside developers to access the code.

Changed use of defines to be opt-out instead of opt-in, so without any defines then one will be using all bells and whistles. Instead of needing to configure special defines to get expected features. This should make it easier for outside developers to access the code.


snakefoot:
Would be great to extract MSMQ + DatabaseTarget + WCF into their own isolated repositories.

So outside developers on the Linux-platform will not be bothered by their Windows-platform requirements for unit-tests.

I would like to keep NLog.WindowsEventLog and NLog.WindowsIdentity in the repository (and keeping them in NLog-core-package for NetFramework), since they are default for everyone on NetFramework (Allow easier upgrade path from NLog 4).

snakefoot:
Looks like the "multi-targetframework" logic was not that perfect. When testing net35,net40,net45 then one was just doing 3 test-runs of net45. 

The problem is the use of `/p:OutputPath=.\bin\release\net35` that was applied to all projects, and not just `NLog.UnitTests`. So the NLog-project rebuilt using net45-targetframework and placed its result in the net35-folder. Which the `NLog.UnitTests`-project then used as it was told.

snakefoot:
Noticed that when testing "pure" NLog-builds on Mono, then the concurrentwrite-protection-logic is not working at all. It only works with `UnixMultiProcessFileAppender` that requires custom build.

snakefoot:
Checked with BitDiffer that Net35-build and Net45-build only has the expected suprises when comparing with nuget NLog 4.7.6

PR #4224, opened 2021-01-03
Added retry to unstable tests on Azure DevOps Pipeline

snakefoot:
SqlServer_NoTargetInstallException +  SqlServer_InstallAndLogMessage

Also noticed the default automatic retry of failing tests, then caused this [check to break](https://dev.azure.com/NLogLogging/NLog/_build/results?buildId=2002&view=results) because it only retried 2 tests:

> ##[error]The specified minimum number of tests 100 were not executed in the test run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: csharp
 
 mono: 5.8.0
 dotnet: 2.0.0
-dist: bionic
+dist: xenial
 
 
 install:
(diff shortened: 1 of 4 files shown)


snakefoot:
Funny or sad the Azure DevOps Builds fails with `WebserviceTest_restapi_httppost_checkingLost` that also has automatic retry :)

snakefoot:
Guess some Azure DevOps Instances are just broken, and retrying on a broken instance will not help anything. Will close this again.

PR #4225, opened 2021-01-03
ScopeContext - Logger.PushScopeState with covariance support

snakefoot:
NLog-Logger.PushScopeState now has extended support for extracting scope-properties. This matches the logic in MEL-ILogger BeginScope inside NLog.Extension.Logging. Besides making the behavior more similar, then it will also allow removal of scope-state-parser-logic from NLog.Extension.Logging.

It is ofcourse better and faster to use NLog-Logger.PushScopeProperties, that now supports both `KeyValuePair<string,object>[] `and `Dictionary<string,object>` as they matches `IReadOnlyCollection<KeyValuePair<string, object>>` as input-type.


snakefoot:
Not completely sure that I like this magic behavior of `NLog-Logger.PushScopeState`. Maybe a little too much probing and reflection, which is the result of the restrictions of MEL BeginScope. Not sure if the NLog-API will win by acquring the same magic behavior.

snakefoot:
Closed in favor of #4230, don't like to hurt kittens.

PR #4230, opened 2021-01-06
ScopeContext - Logger PushScopeProperties with covariance support

snakefoot:
Now you can do this:

```c#
var stringDictionary = new Dictionary<string, string>{ { "key1", "value1" } };
using (logger.PushScopeProperties(stringDictionary))
{
    // Tada
}
```

Or this:

```c#
var integerList = new [] { new KeyValuePair<string, int>("Key1", 42) };
using (logger.PushScopeProperties(integerList))
{
    // Tada
}
```

Instead of being restricted to `IList<KeyValuePair<string, object>>`

Think I prefer this over implementing covariance through `logger.PushScopeState`, as it will be easier for the user to understand (Than trying to guess the magic type needed to make the generic-state-value work as properties).


snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/ScopeProperty-Layout-Renderer (With Logger-example)

PR #4232, opened 2021-01-08
Disable automatic assembly loading by default

### LINKED ISSUE #3674: Skip auto-loading and auto-scanning of external assemblies by default
This for NLog 5.0

Auto-loading works extremely bad on NetCore (the future) and the auto-loading gives a performance hit and probably also a security-issue (loading rogue assemblies). When used in Azure functions, then any reduction in startup time is extremely important.

Maybe extract the auto-loading logic into its own assembly. So one has to explicit call it upfront to activate it (If still want it on old platforms)

### LINKED ISSUE #3798: Remove Auto loading assemblies
It won't work for .NET Core (https://github.com/NLog/NLog/issues/1091)

and extensions could be used.

snakefoot:
Added new NLog.config option `autoLoadExtensions='true'`. Resolves #3798 and resolves #3674 

Faster startup for light-weight application like cloud-functions, and improve security against code-injection.

### DIFF
--- a/src/NLog/Config/ConfigurationItemFactory.cs
+++ b/src/NLog/Config/ConfigurationItemFactory.cs
@@ -418,10 +418,16 @@ namespace NLog.Config
             var nlogAssembly = typeof(ILogger).GetAssembly();
             var factory = new ConfigurationItemFactory(LogManager.LogFactory.ServiceRepository, null, nlogAssembly);
             factory.RegisterExternalItems();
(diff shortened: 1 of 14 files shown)


snakefoot:
Updated wiki https://github.com/NLog/NLog/wiki/Register-your-custom-component about NLog 5.0 no longer autoloading assemblies with NLog-prefix.

PR #4236, opened 2021-01-12
Adding support for const NLog Config Variables

### LINKED ISSUE #3470: Add support for const variables with layouts
Btw. could be nice if one could make a static reference to a JsonLayout. So one is not required to use `"${var:myjson}"` but could just write:

```
    <variable name='myJson'>
        <layout type='JsonLayout'>
            <attribute name='short date' layout='${shortdate}' />
        </layout>
    </variable>
    <targets>
         <target type='file' name='myfile' layout='${myJson}' />
    <targets>
```

Then one could skip a dictionary lookup for every logevent-write. And one would be able to make use of optimizations like `ThreadAgnostic`.

_Originally posted by @snakefoot in https://github.com/NLog/NLog/pull/3459#issuecomment-500256839_

snakefoot:
Trying to resolve #3470 by skipping the need for dynamic lookup using `${var}`

Minor enhancement of #3459

### DIFF
--- a/src/NLog/Config/LoggingConfiguration.cs
+++ b/src/NLog/Config/LoggingConfiguration.cs
@@ -851,8 +851,15 @@ namespace NLog.Config
         [NotNull]
         internal string ExpandSimpleVariables(string input)
         {
+            return ExpandSimpleVariables(input, out var _);
+        }
+
+        [NotNull]
+        internal string ExpandSimpleVariables(string input, out Layout matchingLayout)
+        {
             string output = input;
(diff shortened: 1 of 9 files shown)


304NotModified (approved):

snakefoot:
Also discovered this PullRequest is also a a bugfix, since it now supports use of `${callsite}` in NLog Config Variables.

snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/Configuration-file#variables

PR #4240, opened 2021-01-14
LogFactory - KeepVariablesOnReload = true by default for config variables

### LINKED ISSUE #2509: keepVariablesOnReload enabled by default
For NLog ver. 5.0 

from: https://github.com/NLog/NLog/issues/2505

### LINKED ISSUE #2960: NLog variables are not threadsafe

snakefoot:
Resolves #2509 and resolves #2960

The old way triggered variable-restore AFTER having loaded the new config from file. It would restore ALL values from the old config (Thus discarding any updates made in the new config files). And at the same time prevents static-mode variables from working with config-variables assigned from API.

The new way triggers variable-restore BEFORE starting to load the new file. It will ONLY restore config-variables that has been assigned from the API by using `LoggingConfiguration.Variables`-property.
- This means changes to config-variables in the new config file, will no longer be discarded (unless the config-variable has been assigned from the API). Less surprises when having KeepVariablesOnReload=true by default.
- This means use of config-variables in static-mode (without NLog Layout) will now apply values assigned from the API.

The new way will work automatically when manually calling `XmlLoggingConfiguration.Reload()`, instead of only when file-watcher detects file-changes and triggers special reload.

The new way will also automatically add extra protection for NLog-Config-Variable-Layout that is not ThreadSafe by default (Rare case). Thus removing the [possible issue with thread-safety](https://github.com/NLog/NLog/wiki/Var-Layout-Renderer) and NLog-Config-Variables.
 - The ThreadSafe-wrapper is hidden from the user, and is only used by `${var}` or when assigned as const-Layout.

### DIFF
--- a/src/NLog/Config/LoggingConfiguration.cs
+++ b/src/NLog/Config/LoggingConfiguration.cs
@@ -54,17 +54,12 @@ namespace NLog.Config
     ///<remarks>This class is thread-safe.<c>.ToList()</c> is used for that purpose.</remarks>
     public class LoggingConfiguration
     {
-        private readonly IDictionary<string, Target> _targets =
-            new Dictionary<string, Target>(StringComparer.OrdinalIgnoreCase);
-
+        private readonly IDictionary<string, Target> _targets = new Dictionary<string, Target>(StringComparer.OrdinalIgnoreCase);
         private List<object> _configItems = new List<object>();
 
         private bool _missingServiceTypes;
 
-        /// <summary>
-        /// Variables defined in xml or in API. name is case case insensitive. 
-        /// </summary>
-        private readonly ThreadSafeDictionary<string, Layout> _variables = new ThreadSafeDictionary<string, Layout>(StringComparer.OrdinalIgnoreCase);
+        private readonly ConfigVariablesDictionary _variables;
 
         /// <summary>
         /// Gets the factory that will be configured
@@ -86,11 +81,13 @@ namespace NLog.Config
         {
             LogFactory = logFactory ?? LogManager.LogFactory;
             LoggingRules = new List<LoggingRule>();
+            _variables = new ConfigVariablesDictionary(this);
         }
 
         /// <summary>
-        /// Gets the variables defined in the configuration.
+        /// Gets the variables defined in the configuration or assigned from API
         /// </summary>
+        /// <remarks>Name is case insensitive.</remarks>
         public IDictionary<string, Layout> Variables => _variables;
 
         /// <summary>
@@ -180,6 +177,22 @@ namespace NLog.Config
                 var configTargets = new HashSet<Target>(_configItems.OfType<Target>().Concat(GetAllTargetsThreadSafe()), SingleItemOptimizedHashSet<Target>.ReferenceEqualityComparer.Default);
                 return configTargets.ToList().AsReadOnly();
             }
+        }
+
+        /// <summary>
+        /// Inserts NLog Config Variable without overriding NLog Config Variable assigned from API
+        /// </summary>
+        internal void InsertConfigFileVariable(string key, Layout value)
+        {
+            _variables.InsertConfigFileVariable(key, value, LogFactory.KeepVariablesOnReload);
+        }
+
+        /// <summary>
+        /// Look NLog Config Variable Layout without unwrapping <see cref="ConfigVariablesDictionary.ThreadSafeWrapLayout"/>
+        /// </summary>
+        internal bool TryLookupDynamicVariable(string key, out Layout value)
+        {
+            return _variables.TryLookupDynamicVariable(key, out value);
         }
 
         /// <summary>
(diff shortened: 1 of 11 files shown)


snakefoot:
@304NotModified Not the smallest PullRequest for changing a default-value. But I find it important that NLog Config Variables behaves in a predictable maner, where only NLog Config Variables that have been modified by API are kept on reload.

304NotModified:
The description sounds good!

304NotModified:
I'm fine with this enabled by default. I haven't checked the code yet.

snakefoot:
Close and reopening to trigger rebuild

304NotModified:
Restarted azure pipeline

304NotModified:
Green!

snakefoot:
Yeah! Looking forward to remove warnings an caveats from the docs.


Sent from my Sony Xperia


Green!

src/NLog/Config/LoggingConfiguration.cs
```diff
@@ -182,6 +179,22 @@ public ReadOnlyCollection<Target> AllTargets
             }
         }
 
+        /// <summary>
+        /// Inserts NLog Config Variable without overriding NLog Config Variable assigned from API
+        /// </summary>
+        internal void InsertConfigFileVariable(string key, Layout value)
```
304NotModified, line 185: While the config is most of the time a file, is file here really the right term?

src/NLog/Config/LoggingConfigurationParser.cs
```diff
@@ -1029,24 +1028,15 @@ private void SetPropertyValueFromString(object targetObject, string propertyName
         {
             try
             {
-                var propertyValueExpanded = ExpandSimpleVariables(propertyValue, out var matchingLayout);
-                if (matchingLayout != null && propertyValueExpanded == propertyValue)
+                var propertyValueExpanded = ExpandSimpleVariables(propertyValue, out var matchingVariableName);
+                if (matchingVariableName != null && propertyValueExpanded == propertyValue && TryLookupDynamicVariable(matchingVariableName, out var matchingLayout) && PropertyHelper.TryGetPropertyInfo(targetObject, propertyName, out var propInfo) && propInfo.PropertyType.IsAssignableFrom(matchingLayout.GetType()))
```
304NotModified, line 1032: Maybe exact condition to method for readability?

src/NLog/Internal/Collections/ConfigVariablesDictionary.cs
```diff
@@ -0,0 +1,251 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Text;
+using NLog.Config;
+using NLog.Layouts;
+
+namespace NLog.Internal
+{
+    [DebuggerDisplay("Count = {Count}")]
```
304NotModified, line 45: I like debugger displays but most of the time I prefix them with the name (e.g. config variables)

src/NLog/Internal/Collections/ConfigVariablesDictionary.cs
```diff
@@ -0,0 +1,251 @@
...
+    internal class ConfigVariablesDictionary : IDictionary<string, Layout>
+    {
+        private readonly ThreadSafeDictionary<string, Layout> _variables = new ThreadSafeDictionary<string, Layout>(StringComparer.OrdinalIgnoreCase);
+        private readonly LoggingConfiguration _configuration;
+        private ThreadSafeDictionary<string, Layout> _dynamicVariables;
+        private ThreadSafeDictionary<string, bool> _apiVariables;
```
304NotModified, line 51: Aren't those then the "const variables"?

src/NLog/Internal/Collections/ConfigVariablesDictionary.cs
```diff
@@ -0,0 +1,251 @@
...
+using System.Diagnostics;
+using System.Linq;
+using System.Text;
+using NLog.Config;
+using NLog.Layouts;
+
+namespace NLog.Internal
+{
+    [DebuggerDisplay("Count = {Count}")]
+    internal class ConfigVariablesDictionary : IDictionary<string, Layout>
```
304NotModified, line 46: Do we need this interface implementation? It add some needed code and also needed tests as result

304NotModified (approved):

src/NLog/Config/LoggingConfiguration.cs
snakefoot: Will rename it to `InsertParsedConfigVariable`

src/NLog/Internal/Collections/ConfigVariablesDictionary.cs
snakefoot: `_dynamicVariables` are those used by `${var}` and const-variables.  `_apiVariables` are those assigned using `LoggingConfiguration.Variables´-dictionary

src/NLog/Internal/Collections/ConfigVariablesDictionary.cs
snakefoot: Yes because of `LoggingConfiguration.Variables`-dictionary

src/NLog/Config/LoggingConfigurationParser.cs
snakefoot: Will extract into helper-method.

src/NLog/Internal/Collections/ConfigVariablesDictionary.cs
snakefoot: I like that it just works like a standard Dictionary. Less noise when looking at `LoggingConfigurarion.Variables`-propery. The property-name already reveals it is the configuration-variables.

snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/Configuration-file with KeepVariablesOnReload=true by default

PR #4246, opened 2021-01-17
Support for overriding AssemblyLoader for loading NLog extensions

### LINKED ISSUE #3981: Maybe NLog should use the same AssemblyLoadContext it loaded instead of the default one for netcore?
**NLog version**: 4.7.2

**Platform**: .NET Core 3.1

Maybe related to #3674

I use `NLog` and `NLog.Extensions.Logging` in an assembly, and load all of them into an new AssemblyLoadContext in another assembly as an plugin.

And then I found that when load and used NLog in an non-default AssemblyLoadContext, instead of use the current AssemblyLoadContext, it will load(or leak) another new instance of itself and all exts in default AssemblyLoadContext.

https://github.com/NLog/NLog/blob/86fba05e272a07d508b32475957b9310f607061d/src/NLog/Internal/AssemblyHelpers.cs#L80
Maybe `AssemblyLoadContext.GetLoadContext(typeof(AssemblyHelpers).Assembly)` should be used instead of `AssemblyLoadContext.Default` or [Assembly.LoadFrom](https://github.com/dotnet/runtime/blob/v5.0.0-preview.4.20251.6/src/libraries/System.Private.CoreLib/src/System/Reflection/Assembly.cs#L335)(which also call `AssemblyLoadContext.Default`) for netcore.

- What is the current result?
NLog load assemblies in the default AssemblyLoadContext.

NLog load assemblies in the same AssemblyLoadContext itself loaded.

- Are there any workarounds? 
use `LogManager.Setup().SetupExtensions(seb => seb.AutoLoadAssemblies(false));` to disable auto load, always use hardcode `SetupExtensionsBuilderExtensions.RegisterAssembly(this ISetupExtensionsBuilder, Assembly)` instead of the string version, not use `extensions` in `NLog.config`

snakefoot:
Resolves #3981

So you can do this:

```c#
NLog.LogFactory().Setup().SetupExtensions(ext => ext.RegisterAssemblyLoader(asmName => {
   var assemblyName = new AssemblyName() { Name = asmName };
   var assemblyContext = AssemblyLoadContext.GetLoadContext(typeof(NLog.LogFactory).Assembly);
   return assemblyContext.LoadFromAssemblyName(assemblyName);
}));
```

The AssemblyLoader will not be used for the NLog-Auto-Extension-Loader-Logic, but it also often fails and is disabled by default with NLog5.

Guess this could also be used by the application to restrict that assemblies are allowed to be loaded, when specified from NLog-configuration-file.

### DIFF
--- a/src/NLog/Config/ConfigurationItemFactory.cs
+++ b/src/NLog/Config/ConfigurationItemFactory.cs
@@ -142,6 +142,11 @@ namespace NLog.Config
         }
 
         /// <summary>
+        /// Gets or sets the assembly loader used for loading NLog extension-assemblies
+        /// </summary>
+        public Func<string, Assembly> AssemblyLoader { get; set; } = AssemblyHelpers.LoadFromName;
+
+        /// <summary>
         /// Gets the <see cref="Target"/> factory.
         /// </summary>
         /// <value>The target factory.</value>
--- a/src/NLog/Config/LoggingConfigurationParser.cs
+++ b/src/NLog/Config/LoggingConfigurationParser.cs
@@ -69,7 +69,7 @@ namespace NLog.Config
         /// Loads NLog configuration from provided config section
         /// </summary>
         /// <param name="nlogConfig"></param>
-        /// <param name="basePath"></param>
+        /// <param name="basePath">Directory where the NLog-config-file was loaded from</param>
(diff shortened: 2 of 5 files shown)


snakefoot:
Closing as not needed. See also #3981

Issue #4159, opened 2020-11-05
Remove LogEvent SequenceID or change to 64 bit ?


snakefoot:
Random thoughts after looking at #4139. And the overhead from synchronizing cpu-caches to atomically increment a counter that very few people uses, and unexpectedly can become negative after some days.

Lets say we have an application that handles 10.000 requests/sec. This application writes a single logevent for each request.

This means the application will write 36..000.000 logevents/hour. And 864.000.000 logevents/day.

Right now the SequenceID is an integer and can hold 2,147,483,647 events, before it becomes negative. So that is 3 days without restarts. Many applications often run 7 days without restarts.

Should https://github.com/NLog/NLog/wiki/SequenceId-layout-renderer be replaced by https://github.com/NLog/NLog/wiki/Counter-layout-renderer ?

304NotModified:
> Maybe the SequenceID-property should only increment when calling the propery-getter? (Skip on constructor)

Sounds good!

snakefoot:
> Sounds good for NLog 5!

Yes already implemented with #4247

But it only reduces the overhead. It doesn't really fix the issue with negative integer values

304NotModified:
Ow yes indeed. Well I guess we can close this one. Long is also nice, but not sure if needed

snakefoot:
NLog v6 now marks LogEventInfo.SequenceID as obsolete. See also #4159

304NotModified:
> NLog v6 now marks LogEventInfo.SequenceID as obsolete. See also [#4159](https://github.com/NLog/NLog/issues/4159)

I think its https://github.com/NLog/NLog/pull/5834 :)

PR #4247, opened 2021-01-17
LogEventInfo - Lazy initialize SequenceID when requested

snakefoot:
No need to synchronize all cpu-core-caches when creating a NLog LogEventInfo. Instead do it lazily when required.

See also #4159 (Will reduce the change of SequenceID becoming negative very fast)

### DIFF
--- a/src/NLog/LogEventInfo.cs
+++ b/src/NLog/LogEventInfo.cs
@@ -71,6 +71,7 @@ namespace NLog
         private LogMessageFormatter _messageFormatter;
         private IDictionary<Layout, object> _layoutCache;
         private PropertiesDictionary _properties;
+        private int _sequenceId;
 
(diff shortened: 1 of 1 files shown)


304NotModified (approved):

SylkoZ:
This change is breaking and needs documentation! The SequenceID is no more in the insert sequenze if you do not request it in the same order. This change broke our functionality because the SequenceID was no more in the correct order because of the lazy initialization.

snakefoot:
I'm sorry about the pain this change has caused you. I have updated the NLog v5 new-posts with breakdown of the merged-pull-requests into "Breaking Change" (compiler will fail) and "Breaking Behavior Change":

- https://nlog-project.org/2022/05/16/nlog-5-0-finally-ready.html

PR #4248, opened 2021-01-17
Fix NetworkTargetConnectionsOverflowAction to match normal OverflowAction-values

snakefoot:
Fixes typo in `AllowNewConnnection` by making it obsolete and replacing it with `Grow`.

Also replaced `DiscardMessage` with `Discard` to match the standard overflow-actions-names.


snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/Network-target with new enum values in NLog 5.0

PR #4252, opened 2021-01-18
TraceTarget - Replaced RawWrite with ReportLogLevel together with new default value

### LINKED ISSUE #3744: TraceTarget - Rename RawWrite to ForceTraceWriteLine with default = true

snakefoot:
Resolves #3744

### DIFF
--- a/src/NLog/Targets/TraceTarget.cs
+++ b/src/NLog/Targets/TraceTarget.cs
@@ -37,6 +37,7 @@
 
 namespace NLog.Targets
 {
+    using System;
     using System.ComponentModel;
     using System.Diagnostics;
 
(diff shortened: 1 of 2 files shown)


snakefoot:
Curious if this will have even more side-effects. Since Azure Log Streaming actually listens to the LogLevel:

https://www.thebestcsharpprogrammerintheworld.com/2018/08/23/system-diagnostics-trace-application-logging-log-stream-on-azure-app-service-function-app/

Changing the default will actually affect a lot of applications. Maybe the setting should just be left alone, since better to have some output (and fixing it), then no output before everything suddenly changes to LogLevel-Verbose.

PR #4264, opened 2021-01-25
Extracted NLog.MSMQ.Tests from NLog.UnitTests

snakefoot:
Prepare for moving NLog.MSMQ into its own source-repository. Part of decommissioning NLog.Extended.


snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/MSMQ-target about new nuget-package

PR #4268, opened 2021-01-26
Extracted NLog.SmtpClient.Tests from NLog.UnitTests

snakefoot:
Small step closer to completing #1148


snakefoot:
Guess it could be named NLog.Mail, but think it is confusing when also having NLog.MailKit.

snakefoot:
Partially reverted by #4298 (Kept refactoring of unit-tests). See also challenge #4297

PR #4275, opened 2021-02-01
Extracted NLog.WindowsIdentity.Tests from NLog.UnitTests

### LINKED ISSUE #2815: ImpersonatingWrapper for dotnetcore
Hi,
Do we know when ImpersonatingWrapper will be available for dotnetcore?
Thanks!

snakefoot:
Also resolves #2815 by adding support for ImpersonatingWrapper on NetCore.

Also removed wrong expectation that ImpersonatingWrapper would ensure WrappedTarget would be initialized using configured credentials. Only Write + Flush will have the expected user-credentials. Because it is the LoggingConfiguration that handles initializing / closing (not the target-wrapper)

Failed to implement proper unittest for NetCore, since I don't know how to create a new local-user with the NetCore-API.


snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/Windows-Identity-Layout-Renderer about new nuget-package

PR #4278, opened 2021-02-02
Extracted NLog.PerformanceCounter.Tests from NLog.UnitTests

snakefoot:
Also added support for NetCore to use Windows Performance Counters. 

One could consider changing the name from "NLog.PerformanceCounter" to "NLog.Windows.PerformanceCounters" (But now it matches "System.Diagnostics.PerformanceCounter"-package).


snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/PerformanceCounter-Layout-Renderer with new nuget-package

PR #4280, opened 2021-02-04
Make NLog.config bool attributes configurable through variables.

mmunozl:
With the current code is not possible to pass variables to set bool values in the nlog.config file. I want to be able to pass some config to file from a previously defined variable. For example:

`<variable name="LOGGER_ENABLED" value="${environment:LOGGER_ENABLED}" />`

`<logger name="*" minlevel="Trace" writeTo="myLogger"  enabled="${LOGGER_ENABLED}" >`

Added value expansion to allow method Parse Boolean to accept variables.

### DIFF
--- a/src/NLog/Config/LoggingConfigurationParser.cs
+++ b/src/NLog/Config/LoggingConfigurationParser.cs
@@ -1301,7 +1301,9 @@ namespace NLog.Config
         {
             try
             {
-                return Convert.ToBoolean(value?.Trim(), CultureInfo.InvariantCulture);
+                string expandedValue;
+                ExpandSimpleVariables(value, out expandedValue);
+                return Convert.ToBoolean(expandedValue?.Trim(), CultureInfo.InvariantCulture);
             }
             catch (Exception exception)
             {


snakefoot:
@mmunozl Thank you for this pull-request. It makes sense to have layout-behavior for enabled-property, when also have layout-behavior for the level-properties.

It is normal practise to add a unit-test to show that the code-change will give the wanted behavior. This will also ensure that this behavior is not broken by some future change (Think you will notice that your code-change will not solve the issue).

Notice that you can also use the LogLevel=Off to disable a logging-rule, and since the level-properties supports layouts, then you can do this:

```xml
<variable name="LOGGER_MINLEVEL" value="${environment:LOGGER_MINLEVEL:whenEmpty=Off}" />
<logger name="*" minlevel="${LOGGER_MINLEVEL}" writeTo="myLogger" />
```

mmunozl:
@snakefoot Thank you for your answer, you were really helpful :)

snakefoot:
@mmunozl Should the Pull-Request be closed, or will you continue to work on it?

maxleifer:
I actually would love to see this feature implemented to disable a logging-rule from an environment variable.
The enabled-property is evaluated during the initialization stage while the level-properties (minlevel, maxlevel, levels, etc.) are evaluated on everytime when log method is invoked.

thanks

snakefoot:
You are wrong. minlevel, maxlevel, level are only evaluated during initialization.


You can force evaluation using LogManager.ReconfigExistingLoggers()


Sent from my Sony Xperia


---- ml wrote ----


I actually would love to see this feature implemented to disable a logging-rule from an environment variable.
The enabled-property is evaluated during the initialization stage while the level-properties (minlevel, maxlevel, levels, etc.) are evaluated on everytime when log method is invoked.

thanks

PR #4285, opened 2021-02-06
Removed QueryPerformanceCounterLayoutRenderer ${qpc}

snakefoot:
The idea with QueryPerformanceCounter is to provide very high resolution timings to measure some small piece of code By suddenly including the overhead from the entire NLog-framework, then it will ruin all measurement results.

[StopWatch](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch) or `${scopetiming}` or `${processtime}` are much better and doesn't require native methods


snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/QPC-Layout-Renderer about deprecated for NLog 5.0

PR #4299, opened 2021-02-12
NLog.OutputDebugString - Extracted to own package

snakefoot:
But left behind DebugSystemTarget with alias "DebugSystem" and added alias "TraceSystem" to TraceTarget for consistency.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -38,6 +38,7 @@ function create-package($packageName)
 }
 
 create-package('NLog.MSMQ')
+create-package('NLog.OutputDebugString')
 create-package('NLog.PerformanceCounter')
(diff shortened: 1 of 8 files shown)


snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/OutputDebugString-target with new nuget-package

scottharman:
Hey team - can we get nuget updated with the full string reported as a result of the breaking change - i.e. tag the package with OutputDebugStringTarget  =- realise this is a zombie comment, but the error message doesn't match the package name:
`error CS0246: The type or namespace name 'OutputDebugStringTarget' could not be found`

PR #4300, opened 2021-02-13
LoggingConfigurationParser - Added support for using assembly-name in type-name

### LINKED ISSUE #2944: Extend NLog target extension type-name loading
Right now on NetCore one must remember to add extension targets to the extensions-section. See also #1091

But actually many extension targets have very good names, so could consider registering a target like this:

https://www.nuget.org/packages/NLog.Targets.Seq

Target-alias: **Seq**
Target-Assembly: **NLog.Targets.Seq**

Then one could support this declare of type:

`<target xsi:type="NLog.Targets.Seq" serverUrl="http://localhost:5341" apiKey="" />`

And the config-loader would then do the following:

- Check if any targets registered as `NLog.Targets.Seq`
- Check if any targets registered as `Seq`
- Attempt to resolve assembly `NLog.Targets.Seq` (Unless already tried)
- Perform the two previous checks again

Then it would not be necessary to have `NLog.Targets.Seq` in the `extensions`-section.

It would then just be a matter of updating the guide-lines for naming packages with NLog Target Extensions.

One could also consider to add full typename-support: **SeqTarget, NLog.Targets.Seq**

snakefoot:
Resolves #2944 by allowing you to do this:

```xml
<targets>
   <target name="evt" type="NLog.WindowsEventLog.EventLog" />
</targets>
```

Making it easier to specify types from NLog-extension-assemblies without also having to update `<extensions>`

### DIFF
--- a/src/NLog/Config/ConfigurationItemFactory.cs
+++ b/src/NLog/Config/ConfigurationItemFactory.cs
@@ -305,10 +305,14 @@ namespace NLog.Config
 
             InternalLogger.Debug("ScanAssembly('{0}')", assembly.FullName);
             var typesToScan = assembly.SafeGetTypes();
-            PreloadAssembly(typesToScan);
-            foreach (IFactory f in _allFactories)
+            if (typesToScan?.Length > 0)
             {
-                f.ScanTypes(typesToScan, itemNamePrefix);
+                var assemblyName = new AssemblyName(assembly.FullName).Name;
+                PreloadAssembly(typesToScan);
+                foreach (IFactory f in _allFactories)
+                {
+                    f.ScanTypes(typesToScan, assemblyName, itemNamePrefix);
+                }
             }
         }
 
--- a/src/NLog/Config/Factory.cs
+++ b/src/NLog/Config/Factory.cs
@@ -65,14 +65,15 @@ namespace NLog.Config
         /// Scans the assembly.
         /// </summary>
         /// <param name="types">The types to scan.</param>
-        /// <param name="prefix">The prefix.</param>
-        public void ScanTypes(Type[] types, string prefix)
+        /// <param name="assemblyName">The assembly name for the types.</param>
+        /// <param name="itemNamePrefix">The prefix.</param>
+        public void ScanTypes(Type[] types, string assemblyName, string itemNamePrefix)
         {
             foreach (Type t in types)
             {
                 try
                 {
-                    RegisterType(t, prefix);
+                    RegisterType(t, assemblyName, itemNamePrefix);
                 }
                 catch (Exception exception)
                 {
@@ -93,6 +94,17 @@ namespace NLog.Config
         /// <param name="itemNamePrefix">The item name prefix.</param>
         public void RegisterType(Type type, string itemNamePrefix)
         {
+            RegisterType(type, string.Empty, itemNamePrefix);
+        }
+
+        /// <summary>
+        /// Registers the type.
+        /// </summary>
+        /// <param name="type">The type to register.</param>
+        /// <param name="assemblyName">The assembly name for the type.</param>
+        /// <param name="itemNamePrefix">The item name prefix.</param>
+        public void RegisterType(Type type, string assemblyName, string itemNamePrefix)
+        {
             if (typeof(TBaseType).IsAssignableFrom(type))
             {
                 IEnumerable<TAttributeType> attributes = type.GetCustomAttributes<TAttributeType>(false);
@@ -100,7 +112,7 @@ namespace NLog.Config
                 {
                     foreach (var attr in attributes)
                     {
-                        RegisterDefinition(itemNamePrefix + attr.Name, type);
+                        RegisterDefinition(attr.Name, type, assemblyName, itemNamePrefix);
                     }
                 }
             }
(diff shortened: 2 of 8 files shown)


snakefoot:
@ThomasArdal Notice NLog will now attempt to extract Extension-Assembly-Name from the type-prefix.

This means funny error-message when having forgotten to include  `Elmah.Io.NLog` in `<extensions>`:

```xml
<targets>
  <target name="elmahio" type="elmah.io" apiKey="API_KEY" logId="LOG_ID"/>
</targets>
```

And if one tries to make use of this feature with Elmah.io, then it fail to recognize that "elmah.io" is the symbol-name. But instead think that "io" is the symbol-name and "Elmah.Io.NLog.elmah" is the assembly-name:

```xml
<targets>
  <target name="elmahio" type="Elmah.Io.NLog.elmah.io" apiKey="API_KEY" logId="LOG_ID"/>
</targets>
```

Btw. funny the [Elmah.io with NLog](https://elmah.io/features/nlog/) has no examples or documentation of how to setup NLog. Maybe include a link to the [Docs](https://docs.elmah.io/logging-to-elmah-io-from-nlog/)

ThomasArdal:
@snakefoot If I remember corretly, the `elmah.io` type references this, right?

```csharp
[Target("elmah.io")]
```

Would it change anything if the target is named `elmahio` or `elmah-io`? I guess having a dot in the name is bad for other reasons too.

snakefoot:
@ThomasArdal Yes instead of this:

```c#
    [Target("elmah.io")]
    public class ElmahIoTarget : AsyncTaskTarget
```

Then with NLog 5.0 it is possible to do this:


```c#
    [Target("elmah.io")]
    [Target("elmah-io")]
    [Target("ElmahIo")]
    public class ElmahIoTarget : AsyncTaskTarget
```

ThomasArdal:
@snakefoot That's nice. Will add that in time when we move the dependency to 5.0 for sure.

Maybe you should consider looking for exact target name matches before falling back to treating dots as part of an assembly name? This will break compatibility with all targets containing dots in the name. Being a major version change this is valid of course. But could still cause some confusion.

snakefoot:
> Maybe you should consider looking for exact target name matches before falling back to treating dots as part of an assembly name? 

The extension-loading-magic is only done when NLog detects the Type-Symbol-Name contains a dot `.`. And before activating this magic-loading then first checks if the Type-Symbol-Name is already registered. So this will still work without any changes:

```xml
<extensions>
  <add assembly="Elmah.Io.NLog"/>
</extensions>

<targets>
  <target name="elmahio" type="elmah.io" apiKey="API_KEY" logId="LOG_ID"/>
</targets>
```

ThomasArdal:
@snakefoot Makes sense 👍 I think we'll go for a name change, not including the dot. Thank you for the heads up.

ThomasArdal:
@snakefoot I have tested the elmah.io target with NLog 5 (dev branch) and written some more documentation. I just want to provide a few ideas that you can take or leave, but at least I have written down my thoughts somewhere.

I totally understand why you want to support fully qualified references of targets, since adding assemblies in extensions is a common source for confused users and non-working config. But I'm not sure I understand the "schema" you have chosen for the fully qualified name. As I understand it, you will need to reference targets on the following format:

```
assemblyname.targetname
```

Where a real example could be:

```
Custom.NLog.Targets.FancyFile
```

This will require a target implementation in an assembly named `Custom.NLog.Targets` with the target attribute and the name of `FancyFile`.

I think that this may be misleading for some users since this isn't a fully qualified class name. I could see something like this working:

```
Custom.NLog.Targets.FancyFileTarget
```

This would reference the file itself. And for target names it could be:

```
Custom.NLog.Targets, fancyfile
```

This includes the assembly name containing the target as well as the target name separated with a comma.

The purpose of this comment is not to force you to support:

```
Elmah.Io.NLog, elmah.io
```

This would be possible of course, but it's no problem for us to add more `Target` attributes once we upgrade the minimum version of NLog to v5.

Anyway, that is just my thoughts.

snakefoot:
I was thinking that NLog-users standing outside the source-code could easily see 2 things:
- The symbolic-type-name
- The nuget-package-name

So I guess that I could change from dottet format to comma-format, and support both:

- elmah.io, Elmah.Io.NLog
- ElmahIoTarget, Elmah.Io.NLog

Then instead of scanning for dot `.`, then it will scan for comma `,` and use that as seperator. But I kept away from actual fully-qualified-names as I find them very wordy with full namespaces etc.

ThomasArdal:
Agree. Real fully-qualified names are probably not preferred. I just think that it's a bit hard to see where the assembly/nuget name ends and where the target name starts with the current solution. The last dot, I know, but could be clearer with the comma for users not into the details.

Maybe you should get input from at least one other person that doesn't have a target with a dot in the name before making any changes :smile:

snakefoot:
> @ThomasArdal  Maybe you should get input from at least one other person that doesn't have a target with a dot in the name before making any changes

I kind of like having the assembly-name at the end, since the symblic-name is the most important part. So it should be first in case a nuget-package can contain multiple NLog-extensions.

ThomasArdal:
Yep agree. I think I had forgotten about the way to reference class names when writing the proposal. `targetname, assemblyname` is definitely better and follow how it's done in other places in .NET.

snakefoot:
@ThomasArdal Thank you for the feed, much appreciated. Have now created #4308

snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/Register-your-custom-component about how to use fully qualified name.

snakefoot:
It seems that using `Elmah.Io.NLog, elmah.io` in `xsi:type` is not valid format, when using XML-schema-validation.

It complains that `, ` (Hexidecimal value 0x2C) is not acceptable. See also #4978

ThomasArdal:
But it works on runtime, right? My memory around this is a bit vague but having a fully qualified name there instead of the add assembly extension should be possible in NLog 5. So, maybe the XSD for the `xsi:type` attribute inside the `target` element should be updated?

snakefoot:
`xsi:type` (And QName) is not under NLog-control, since part of the XML-standard.

ThomasArdal:
Oh, yeah of course. `xsi` namespace. I see that I am using a `type` attribute in the documentation but `xsi:type` in the samples. What would be the better option? `type` I guess to get the schema validation right.

snakefoot:
When using `xsi:type` then it allows intellisense and validation of config-options by using `NLog.xsd`-schema.

ThomasArdal:
Yeah ok. I see that IntelliSense no longer works in the sample. I think we have some URLs to update. Like www to non-www and https).

PR #4301, opened 2021-02-14
Update docs, remove out dated "ArrayList"

304NotModified:
The word "ArrayList" confused me in the docs, and it's also outdated (not only arraylist, but also the docs)


snakefoot (approved):
Looks great

PR #4304, opened 2021-02-14
 Added TypesInInternalNamespaceShouldBeInternalTest

304NotModified:
Currently the issues: "NLog.Internal.Xamarin.PreserveAttribute" + "NLog.Internal.Fakeables.IAppDomain"

move the namepaces for those?

proposal:

-  "NLog.Internal.Xamarin.PreserveAttribute" -> "NLog.Xamarin.PreserveAttribute" 
- "NLog.Internal.Fakeables.IAppDomain" ->  "NLog.IAppDomain"

### DIFF
--- a/src/NLog/Config/IUsesStackTrace.cs
+++ b/src/NLog/Config/IUsesStackTrace.cs
@@ -31,10 +31,8 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 // 
 
-namespace NLog.Internal
+namespace NLog.Config
 {
-    using Config;
-
     /// <summary>
     /// Allows components to request stack trace information to be provided in the <see cref="LogEventInfo"/>.
     /// </summary>
--- a/src/NLog/Internal/Abstractions/IAppDomain.cs
+++ b/src/NLog/Internal/Abstractions/IAppDomain.cs
@@ -77,7 +77,7 @@ namespace NLog.Internal.Fakeables
         /// Process exit event.
         /// </summary>
         event EventHandler<EventArgs> ProcessExit;
-        
+
         /// <summary>
         /// Domain unloaded event.
         /// </summary>
--- a/src/NLog/Internal/Abstractions/IConfigurationManager.cs
+++ b/src/NLog/Internal/Abstractions/IConfigurationManager.cs
@@ -40,7 +40,7 @@ namespace NLog.Internal
     /// <summary>
     /// Interface for the wrapper around System.Configuration.ConfigurationManager.
     /// </summary>
-    public interface IConfigurationManager
+    internal interface IConfigurationManager
     {
         /// <summary>
         /// Gets the wrapper around ConfigurationManager.AppSettings.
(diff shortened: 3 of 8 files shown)


304NotModified:
@snakefoot you mentioned somewhere you like to have this test, isn't?

snakefoot:
> you mentioned somewhere you like to have this test, isn't?

Yes I prefer types in NLog.Internal-namespace are not leaked by accident.

snakefoot:
> Currently the issues: "NLog.Internal.Xamarin.PreserveAttribute" + "NLog.Internal.Fakeables.IAppDomain"

I think there should be an opt-out-list for accepted types. Where `PreserveAttribute` is one of them. But I think `IAppDomain` should not be public.

Funny that it doesn't see `public static class XmlHelper`.

304NotModified:
> I think there should be an opt-out-list for accepted types. Where `PreserveAttribute` is one of them

But why should we have it in the internal namespace then?

304NotModified:
> Funny that it doesn't see `public static class XmlHelper`.

Well detecting if it's internal is a bit hard, but look into this

snakefoot:
> But why should we have it in the internal namespace then?

Similar to how Microsoft handles its API. Public classes in the Internal-namespace are subject to change and not guaranteed to be completely stable.

Ex. PreserveAttribute is not intended for others to use, but just a class used for tooling, so NLog-assembly works better on the Xamarin-platform. Internal for NLog alone, and should not be used by others. But have to be public or else the Xamarin-tooling will get sad.

304NotModified:
fixed, now we have these:


- NLog.Internal.IConfigurationManager 
- NLog.Internal.IUsesStackTrace 
- NLog.Internal.ConfigurationManager 
- NLog.Internal.StreamHelpers 
- NLog.Internal.StringHelpers 
- NLog.Internal.XmlHelper 
- NLog.Internal.Fakeables.IAppDomain

304NotModified:
IUsesStackTrace should be public?

snakefoot:
I think they should be made internal with NLog 5.0, and skipping the step with marking them "obsolete" first.

304NotModified:
yes there where a bit marked already by the namepace :)

304NotModified:
IAppDomain is hard to be internal, as LogFactory has CurrentAppDomain

snakefoot:
But it would make sense to move `IUsesStackTrace` into NLog.Layout-namespace.

snakefoot:
> IAppDomain is hard to be internal, as LogFactory has CurrentAppDomain

But why does LogFactory exposure CurrentAppDomain ?

304NotModified:
> But why does LogFactory expose CurrentAppDomain ? Some kind of special use-case ?

I really don't know. Doubting if we should obsolete it. Just make it internal sounds breaking. Maybe google (duckduckgo / github search) if someone uses it

304NotModified:
only forks, https://github.com/search?l=C%23&p=2&q=Factory.CurrentAppDomain&type=Code

but of course a bit sloppy search

snakefoot:
Also did a sloppy search for CurrentAppDomain + LogFactory and then CurrentAppDomain + LogManager.

Neither revealed any source-code outside the NLog-repository. But some callstacks where both where included.

Doesn't look like anyone is using CurrentAppDomain for anything exciting.

304NotModified:
I think we should handle

304NotModified:
superseded by https://github.com/NLog/NLog/pull/4331

PR #4308, opened 2021-02-15
LoggingConfigurationParser - Added support for using assembly-name in type-name (comma delimiter)

snakefoot:
Follow up to #4300, so instead you can do this:

```xml
<targets>
   <target name="evt" type="EventLog, NLog.WindowsEventLog" />
</targets>
```

Or like this:

```xml
<targets>
   <target name="elmah" type="elmah.io, Elmah.Io.NLog" />
</targets>
```

### DIFF
--- a/src/NLog/Config/Factory.cs
+++ b/src/NLog/Config/Factory.cs
@@ -158,7 +158,11 @@ namespace NLog.Config
             GetTypeDelegate typeLookup = () => itemDefinition;
             _items[itemNamePrefix + itemName] = typeLookup;
             if (!string.IsNullOrEmpty(assemblyName))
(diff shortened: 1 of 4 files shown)


snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/Register-your-custom-component with using fully qualified name.

snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/Configuration-file#extensions about using fully qualified type-name.

snakefoot:
It seems that using `Elmah.Io.NLog, elmah.io` in `xsi:type` is not valid format, when using XML-schema-validation.

It complains that `, ` (Hexidecimal value 0x2C) is not acceptable. See also #4978

PR #4323, opened 2021-02-27
LogFactory.Setup().LoadConfiguration() with fluent creation of NLog-LoggingConfiguration

snakefoot:
Same as #4128 but for dev-branch.

Continuing the work started with NLog 4.7 and NLog 4.7.1 to improve fluent-configuration, by adding new extension-methods to the existing `ISetupLoadConfigurationBuilder`-interface.

You can now do this:

```c#
var logger = LogManager.Setup().LoadConfiguration(c =>
{
   c.ForLogger("*").FilterMinLevel(LogLevel.Info).WriteTo(new ConsoleTarget("logconsole")).WithAsync();
   c.ForLogger().WriteTo(new FileTarget("logfile") { FileName = "file.txt" }).WithAsync();
}).GetCurrentClassLogger();
```

Was thinking to add more helper-methods for the individual-target-types, but decided to skip it for simplicity (Only included `MethodCallTarget` as an example)

This allows one to combine file-config with runtime-config, where it loads from NLog.config, and then applies black-hole filtering:

```c#
var logger = LogManager.Setup().LoadFromFile().LoadConfiguration(c =>
{
   c.ForLogger("Microsoft.*").TopRule().WriteToNil(maxLevel: LogLevel.Info);
}).GetCurrentClassLogger();
```

### DIFF
--- a/src/NLog/Config/ISetupConfigurationLoggingRuleBuilder.cs
+++ b/src/NLog/Config/ISetupConfigurationLoggingRuleBuilder.cs
@@ -1,5 +1,5 @@
 ﻿// 
-// Copyright (c) 2004-2020 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
+// Copyright (c) 2004-2021 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
 // 
 // All rights reserved.
 // 
--- a/src/NLog/Config/ISetupConfigurationLoggingRuleBuilder.cs
+++ b/src/NLog/Config/ISetupConfigurationLoggingRuleBuilder.cs
@@ -1,5 +1,5 @@
 ﻿// 
-// Copyright (c) 2004-2021 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
+// Copyright (c) 2004-2020 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
 // 
 // All rights reserved.
 // 
@@ -34,18 +34,9 @@
 namespace NLog.Config
 {
     /// <summary>
-    /// Interface for fluent setup of LoggingConfiguration for LogFactory 
+    /// Interface for fluent setup of LoggingRules for LoggingConfiguration
     /// </summary>
-    public interface ISetupLoadConfigurationBuilder
+    public interface ISetupConfigurationLoggingRuleBuilder : ISetupConfigurationWriteToTargetsBuilder
     {
-        /// <summary>
-        /// LogFactory under configuration
-        /// </summary>
-        LogFactory LogFactory { get; }
-
-        /// <summary>
-        /// LoggingConfiguration being built
-        /// </summary>
-        LoggingConfiguration Configuration { get; set; }
     }
 }
--- a/src/NLog/Config/ISetupConfigurationLoggingRuleBuilder.cs
+++ b/src/NLog/Config/ISetupConfigurationLoggingRuleBuilder.cs
@@ -36,7 +36,11 @@ namespace NLog.Config
     /// <summary>
     /// Interface for fluent setup of LoggingRules for LoggingConfiguration
     /// </summary>
-    public interface ISetupConfigurationLoggingRuleBuilder : ISetupConfigurationWriteToTargetsBuilder
+    public interface ISetupConfigurationLoggingRuleBuilder : ISetupConfigurationTargetBuilder
     {
+        /// <summary>
+        /// LoggingRule being built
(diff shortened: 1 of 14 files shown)


snakefoot:
@AndreaCuneo Thank you for the feedback. You are right about improving the support for re-using the same target in multiple logger-rules.

Right now `WriteTo` starts a Target-Builder, but makes sense that one can call WriteTo and give a pre-created target-builder. And re-use the same pre-created target-builder for several logging-rules.

And it also make sense that one can combine two Target-Builder together. Ex. WithFallback() should also take a pre-created target-builder.

Thank you for raising the bar, and showing that more advanced scenarios should be supported. And not just the simple example from the tutorials.

snakefoot:
@AndreaCuneo Now you can do this:

```c#
var logger = LogManager.Setup().LoadConfiguration(c =>
{
   var myTarget = c.ForTarget("MyTarget").WriteToConsole();
   var myTargetAsync = c.ForTarget("MyTarget_Async").WriteTo(myTarget).WithAsync(); 
   
   c.ForLogger("Microsoft*").FilterMaxLevel(LogLevel.Info).WriteToNil();
   c.ForLogger("*").FilterMinLevel(LogLevel.Error).FinalRule().WriteTo(myTarget);
   c.ForLogger("*").WriteTo(myTargetAsync);
}).GetCurrentClassLogger();
```

Guess NLog should be smarter, so a Logger-Configuration will never have the same target twice. Thus reducing the issue with duplicate output when not careful with the loggingrules. And if people really need duplicate output to the same target, then one can use other means.

AndreaCuneo:
@snakefoot looks grand, and I fully agree on the subscript.

One question: is loose reference by target name supported? 

My idea stem from the fact that often which Targets are available is an architecture decision predefined (ie by Architect / Governance ) across a fleet of Applications, for example in Enterprises.
Rules instead could change based on some peculiarities of the given Application.

I think would be useful if the scenario below would be supported out-of-the-box. 

```cs

// in library code

static ISetupBuilder SetupDefaults(this ISetupBuilder setup) => 
   setup.LoadConfiguration(c =>
   {
       var myTarget = c.ForTarget("Default").WriteToConsole().WriteToDatabase(...).WithAsync();
   
       c.ForLogger("Microsoft*").FilterMaxLevel(LogLevel.Info).WriteToNil();
       c.ForLogger("*").FilterMinLevel(LogLevel.Error).FinalRule().WriteTo(myTarget);    
   });

// in application code
LogManager.Setup().SetupDefaults().LoadConfiguration(c => {
    c.ForLogger("Microsoft.*").TopRule().FilterMinLevel(LogLevel.Error).WriteTo("Default");    
});

```

snakefoot:
Sounds like we have moved from "reviewing presented fluent api" to "investigating the wet dreams of a single user".

NLog has no direct concept of giving a name to a group of targets. But one could bring the [SplitGroup](https://github.com/NLog/NLog/wiki/SplitGroup-target)-target into the game like this:

```c#
static ISetupBuilder SetupDefaults(this ISetupBuilder setup) => 
   setup.LoadConfiguration(c =>
   {
       var myTarget = c.ForTarget("Default").WriteToSplitGroup().WriteToConsole().WriteToDatabase(...).WithAsync();
   
       c.ForLogger("Microsoft*").FilterMaxLevel(LogLevel.Info).WriteToNil();
       c.ForLogger("*").FilterMinLevel(LogLevel.Error).FinalRule().WriteTo(myTarget);    
   });

// in application code
LogManager.Setup().SetupDefaults().LoadConfiguration(c => {
    var myTarget = c.Configuration.FindTarget("Default");
    c.ForLogger("Microsoft.*").TopRule().FilterMinLevel(LogLevel.Error).WriteTo(myTarget);
});
```

But this is an advanced shotgun, that I seldom seen anyone use. The normal users that ask for fluent-api, are people with very simple config, and wants very simple API.

But thank you for all these special API-usage examples. It has revealed several places where more safe-guards must be added, to avoid users shooting themselves in the foot.

AndreaCuneo:
@snakefoot sorry, I've confused you with a copy-paste issue and made you lose time. Apologies.

I didn't meant to suggest the 'same name for multiple targets' case, but thanks for the hint nonethess on the SplitGroup target, I didn't know it existed.

The case was only for the the `.WriteTo("Default")` , by string. I've not seen the overload in the commits, so I wasn't sure if it was me missing it or could have been a small addition.

snakefoot:
I think `.WriteTo("Default")` is a sin because it involves magic-strings that can fail at runtime. It works against the whole idea with fluent-api, that it should be simple predictable with no surprises.

snakefoot:
Updated wiki-pages:
- https://github.com/NLog/NLog/wiki/Fluent-Configuration-API
- https://github.com/NLog/NLog/wiki/Configure-from-code

Issue #3366, opened 2019-05-04
NLog core package with bare minimum


304NotModified:
With bare minimum. For several goals:
- Using the NLog Logger in application-library should not invite many dependencies for the application-library 
- The NLog nuget-package should support platforms without console or file-system.
- The NLog nuget-package should produce a small binary when used in simple console-app with AOT

snakefoot:
Think NLog 6.0 should only have these classes and nothing else:

- ILogger
- Logger
- LogEventInfo
- LogFactory (Without GetCurrentClassLogger)
- LoggingConfiguration + LoggingRule
- InternalLogger (That has a subscription-system, but no Console-Stream / File-Stream / Trace-Writer)
- Layout
- LayoutRenderer
- Target

Then create NLog-meta-package that includes the standard-stuff:

- NLog.XmlLoggingConfiguration
- NLog.JsonLayout
- NLog.XmlLayout
- NLog.ServiceRepository
- NLog.CallSite
- NLog.FileTarget
- NLog.ConsoleTarget
- etc.

And another NLog-meta-package with the advanced stuff:

- NLog.AutoLoadConfiguration
- NLog.ConcurrentFileTarget
- NLog.ZipFileTarget
- NLog.DatabaseTarget
- NLog.SmtpTarget
- etc.

snakefoot:
When making the NLog 6.0 package, then these PublishTrimmed warnings should be isolated:
```
NLog\src\NLog\Internal\Reflection\PropertyHelper.cs(80,60): Trim analysis warning IL2057: NLog.Internal.PropertyHelper.<>c.<BuildPropertyConversionMapper>b__9_4(String,ConfigurationItemFactory): Unrecognized value passed to the parameter 'typeName' of method 'System.Type.GetType(String,Boolean)'. It's not possible to guarantee the availability of the target type. [C:\Users\snakefoot\source\repos\NLog\src\ConsoleApp1\ConsoleApp1.csproj]
NLog\src\NLog\Config\PropertyTypeConverter.cs(60,74): Trim analysis warning IL2057: NLog.Config.PropertyTypeConverter.<>c.<BuildStringConverterLookup>b__6_2(String,String,IFormatProvider): Unrecognized value passed to the parameter 'typeName' of method 'System.Type.GetType(String,Boolean)'. It's not possible to guarantee the availability of the target type. [C:\Users\snakefoot\source\repos\NLog\src\ConsoleApp1\ConsoleApp1.csproj]
[...]

PR #4325, opened 2021-02-27
Introduced LogEventBuilder instead of NLog.Fluent.LogBuilder

### LINKED ISSUE #532: Callsite Layout Renderer and Caller Attributes

snakefoot:
Introduces LogEventBuilder as replacement for Fluent.LogBuilder. Alternative to #4044 that skips obsolete of ILogger-interfaces

Obsoleted the old fluent `LogBuilder` and introduced `LogEventBuilder` that is available without needing to write `using NLog.Fluent`. The interfaces are very similar, but the start method is a little different:

- **LogBuilder:** `logger.Info().Message("Hello").Write();`
- **LogEventBuilder:** `logger.ForInfoEvent().Message("Hello").Write();`

I guess one could change `LogEventBuilder` to be activated from using `Info()` but it would generate more noise about ambiguous extensions-methods. But actually I like `ForInfoEvent()` since it signals that one is starting to define a LogEvent.

Also unsure if I like the `Write()`-method. Think I prefer `.Log()` but then the transformation would not be that easy:

- **LogEventBuilder:** `logger.ForInfoEvent().Message("Hello").Log();`

The major improvement in `LogEventBuilder` is there is no allocation if the LogLevel is not enabled. And when the LogLevel is enabled, then it no longer allocates an additional `LogBuilder`-object (Because `LogEventBuilder` is a struct on the stack). Resolves #532


snakefoot:
If #3366 becomes true, then it doesn't make sense to introduce a new interface, that will be removed with NLog 6.0

Guess I will just have just to live with the pain, and change the new LogBuilder to work with the very ugly `ILogger`-api.

snakefoot:
Created wiki-page: https://github.com/NLog/NLog/wiki/Fluent-Logger-API

PR #4331, opened 2021-02-28
Move types in the .Internal namespace to internal & move IUsesStackTrace to NLog.Config namespace

304NotModified:
- move all classes and interfaces in the internal namespace as internal. Including test!
- move IUsesStackTrace to NLog.Config namespace
- IAppDomain should be done in the future -> https://github.com/NLog/NLog/issues/3995
- NLog.Internal.Xamarin.PreserveAttribute is an exception. It isn't supported, but needed to be public

Follow up of https://github.com/NLog/NLog/pull/4304


snakefoot (approved):
Looks great. And excellent job with the unit-test.

PR #4344, opened 2021-03-08
ScopeContextNestedStatesLayoutRenderer - Support Format with FormatAsJson

### LINKED ISSUE #4315: JsonLayout should support rendering of ScopeContext Nested States
Something like this:

```xml
    <attribute name="Scopes" layout="${scopenested:format=@}" encode="false" />
</layout>
```

Then it should output the following:

```json
{
   "Scopes": [
    {
      "Message": "request 1",
      "RequestID": 1,
      "{OriginalFormat}": "request {RequestID}"
    },
    {
      "Message": "auth type ssl",
      "AuthType": "ssl",
      "{OriginalFormat}": "auth type {AuthType}"
    }
   ]
}
```

Special logic must be implemented so it will recognize when Nested-State-Object implements `IReadOnlyList<KeyValuePair<string,object>>`. It needs to render the list like a dictionary, but handle that the returned list might have multiple keys with the same name.

snakefoot:
So you can do this:

`${scopenested:format=@}`

And use it in JsonLayout like this:

```xml
<layout type="JsonLayout">
     <attribute name="scopes" layout="${scopenested:format=@}" encode="false" />
</layout>
```

Resolves #4315 (But without support for pretty indented mode)

### DIFF
--- a/src/NLog/Contexts/ScopeContext.cs
+++ b/src/NLog/Contexts/ScopeContext.cs
@@ -711,32 +711,19 @@ namespace NLog
 
             private IReadOnlyCollection<KeyValuePair<string, object>> EnsureCollectionWithUniqueKeys()
             {
-                if (_scopeProperties.Count > 10)
+                var propertyCount = _scopeProperties.Count;
+                if (propertyCount > 10)
                 {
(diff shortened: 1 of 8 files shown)


snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/ScopeNested-Layout-Renderer about `Format = @`

PR #4354, opened 2021-03-14
LoggerConfiguration - Avoid duplicate output when same target is reused

snakefoot:
Skip writing to the same target multiple times, eventhough configured (Since most likely by accident).

Reducing the chance of hair-loss or high-blood-pressue when using NLog.

But having this auto-repair might lead to situation where people can create very "messy" `NLog.config`-sample that works, but suddenly when going down the road of LoggingRule-filters, then everything explodes.

### DIFF
--- a/src/NLog/Config/LoggingRule.cs
+++ b/src/NLog/Config/LoggingRule.cs
@@ -50,6 +50,7 @@ namespace NLog.Config
     {
         private ILoggingRuleLevelFilter _logLevelFilter = LoggingRuleLevelFilter.Off;
         private LoggerNameMatcher _loggerNameMatcher = LoggerNameMatcher.Create(null);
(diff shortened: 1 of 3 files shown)


Zayden16:
Can confirm high-blood-pressure because of this.

304NotModified:
I like this idea! Nothing wrong with fixing silly mistakes. 

But would it be better if we ignore it one applying the config?

(Ps I expect a warning in the internal logger then. Maybe also log the concequense?)

snakefoot:
I guess the validation has to be done some where near `ReconfigureExistingLoggers` since I believe it is the only place where one looks at all the LoggingRules.

You have a better location?

snakefoot:
Have updated to `InternalLogger.Warn` when "fixing" duplicate target-ouput.

PR #4362, opened 2021-03-20
SimpleLayout - Faster parsing of simple values and reduced allocation

snakefoot:
When assigning simple value to Layout, then it now skips allocation of these objects:
- `SimpleStringReader`
- `StringBuilder`
- `List<LayoutRenderer>`
- `ReadOnlyCollection<LayoutRenderer>`


snakefoot:
Very tempted to mark `Renderer`-property as obsolete since it provides not real value, beside bloat and unnecessary allocation.

PR #4363, opened 2021-03-20
TargetWithLayout - Skip parsing of default Layout

### LINKED ISSUE #4033: Performance: Targets inheriting from NLog.Targets.TargetWithLayout can compile layout excessively
**NLog version**: 4.7.2
**Platform**: .Net 3.5 / .Net 4.5 / .NET Core 2

If one sets up the `Layout` property on a target inheriting from `NLog.Targets.TargetWithLayout`, `LayoutParser.CompileLayout` will be invoked at least twice: once for the default value set inside the constructor of `NLog.Targets.TargetWithLayout`, and the second time for the value set by the user.

Expected behavior: 
Only compile layout once and only for the value which is going to be used.

P.S. I understand it is pretty minor, but if that is not too much effort...

snakefoot:
Skip parsing of Layout in constructor that resolves #4033 (Removes unnecessary parsing that might be thrown away, when Layout is re-assigned right after)

### DIFF
--- a/src/NLog/LayoutRenderers/LevelLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/LevelLayoutRenderer.cs
@@ -34,6 +34,7 @@
 namespace NLog.LayoutRenderers
 {
     using System;
+    using System.Collections.Generic;
     using System.ComponentModel;
     using System.Text;
     using NLog.Config;
@@ -47,12 +48,31 @@ namespace NLog.LayoutRenderers
     [ThreadSafe]
     public class LevelLayoutRenderer : LayoutRenderer, IRawValue, IStringValueRenderer
     {
+        private static readonly Dictionary<int, string> _upperCaseMapper = new Dictionary<int, string>()
+        {
+            { LogLevel.Trace.Ordinal, LogLevel.Trace.ToString().ToUpperInvariant() },
+            { LogLevel.Debug.Ordinal, LogLevel.Debug.ToString().ToUpperInvariant() },
+            { LogLevel.Info.Ordinal, LogLevel.Info.ToString().ToUpperInvariant() },
+            { LogLevel.Warn.Ordinal, LogLevel.Warn.ToString().ToUpperInvariant() },
+            { LogLevel.Error.Ordinal, LogLevel.Error.ToString().ToUpperInvariant() },
+            { LogLevel.Fatal.Ordinal, LogLevel.Fatal.ToString().ToUpperInvariant() },
+            { LogLevel.Off.Ordinal, LogLevel.Off.ToString().ToUpperInvariant() },
+        };
+
         /// <summary>
         /// Gets or sets a value indicating the output format of the level.
         /// </summary>
         /// <docgen category='Rendering Options' order='10' />
         [DefaultValue(LevelFormat.Name)]
         public LevelFormat Format { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether upper case conversion should be applied.
+        /// </summary>
+        /// <value>A value of <c>true</c> if upper case conversion should be applied otherwise, <c>false</c>.</value>
+        /// <docgen category='Rendering Options' order='10' />
+        [DefaultValue(true)]
+        public bool Uppercase { get; set; }
 
         /// <inheritdoc/>
         protected override void Append(StringBuilder builder, LogEventInfo logEvent)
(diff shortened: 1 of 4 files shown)


304NotModified:
> Skip parsing of Layout in constructor that resolves #4033 (Removes unnecessary parsing that might be thrown away, when Layout is re-assigned right after)

I like that idea :)

But that's still possible with loglevel upper? I think a upper case loglevel is a better default from the perspective of an user.

snakefoot:
> But that's still possible with loglevel upper? I think a upper case loglevel is a better default from the perspective of an user.

The other is faster and simpler. Building a new SimpleLayout to embed in the Uppercase-Wrapper is ugly. 

Alternative one could add an UpperCase-property on LevelLayoutRenderer ?

304NotModified:
> Alternative one could add an UpperCase-property on LevelLayoutRenderer ?

Sounds good to me for this case

src/NLog/LayoutRenderers/LevelLayoutRenderer.cs
```diff
@@ -47,21 +48,40 @@ namespace NLog.LayoutRenderers
...
         /// <docgen category='Rendering Options' order='10' />
         [DefaultValue(LevelFormat.Name)]
         public LevelFormat Format { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether upper case conversion should be applied.
+        /// </summary>
+        /// <value>A value of <c>true</c> if upper case conversion should be applied otherwise, <c>false</c>.</value>
+        /// <docgen category='Rendering Options' order='10' />
+        [DefaultValue(true)]
```
304NotModified, line 74: Is the default really true?

src/NLog/LayoutRenderers/LevelLayoutRenderer.cs
```diff
@@ -71,15 +91,23 @@ protected override void Append(StringBuilder builder, LogEventInfo logEvent)
...
                     break;
             }
         }
 
+        private string GetUpperCaseString(LogLevel level)
+        {
+            if (_upperCaseMapper.TryGetValue(level.Ordinal, out var uppercaseString))
+                return uppercaseString;
+            else
+                return level.ToString().ToUpperInvariant();
```
304NotModified, line 108: When could this happen?

src/NLog/LayoutRenderers/LevelLayoutRenderer.cs
304NotModified: test is green, so I mis something?

src/NLog/LayoutRenderers/LevelLayoutRenderer.cs
snakefoot: I like fallbacks instead of exceptions.

src/NLog/LayoutRenderers/LevelLayoutRenderer.cs
snakefoot: It should be false. So it is strange that the test is green. Will take a look

src/NLog/LayoutRenderers/LevelLayoutRenderer.cs
304NotModified: me too, but this is impossible to reach? Dead code is also not nice (untested, could break in the future)

src/NLog/LayoutRenderers/LevelLayoutRenderer.cs
snakefoot: The test is not green. The test is located in AppVeyor build. Guess the `VerifyDefaultValues`-test should be extracted from SourceCodeTests.cs and be placed as normal unit-test.

src/NLog/LayoutRenderers/LevelLayoutRenderer.cs
snakefoot: If I had implemented some crazy fallback logic, then it would make sense to consider not having it. But this seem very harmless to me.

snakefoot:
Btw. found this nice trick, where all targets could just do this `<inheritdoc cref="DefaultLayoutText"/>` instead of verbatim-copy in XML-docs. See also: https://stackoverflow.com/a/65293201/193178

src/NLog/LayoutRenderers/LevelLayoutRenderer.cs
304NotModified: Well its confusing and it wont have a goal? You could argue this will prevent bugs in the future, but I think it will lead to bugs in the future as it's impossible to test.

PR #4372, opened 2021-03-23
 Fixed simple code warnings reported by Sonar

snakefoot:


src/NLog/Internal/StreamHelpers.cs
```diff
@@ -96,7 +95,7 @@ public static void CopyWithOffset(this Stream input, Stream output, int offset)
         {
             if (offset < 0)
             {
-                throw new ArgumentException("negative offset");
+                throw new ArgumentException("cannot be negative", nameof(offset));
```
304NotModified, line 98: Maybe this should be a ArgumentOutOfRangeException?

src/NLog/LogLevel.cs
```diff
@@ -251,7 +251,7 @@ public static LogLevel FromOrdinal(int ordinal)
                     return Off;
 
                 default:
-                    throw new ArgumentException("Invalid ordinal.");
+                    throw new ArgumentException($"Unknown loglevel: {ordinal.ToString()}.", nameof(ordinal));
```
304NotModified, line 254:
```suggestion
                    throw new ArgumentException($"Unknown log level ordinal: {ordinal.ToString()}.", nameof(ordinal));
```

304NotModified (approved):

304NotModified:
2 suggestions :)

src/NLog/Internal/StreamHelpers.cs
snakefoot: Just working on removing false positives. So real errors are not hidding in the noise.

src/NLog/LogLevel.cs
snakefoot: Well the Sonar Rule checks if the ArgumentException message contains the input-parameter in the text. So your suggestions wants this instead: `throw new ArgumentException($"Unknown log level {nameof(ordinal}: {ordinal.ToString()}.", nameof(ordinal));`

But I'm fine with just using ArgumentException-parameterName

PR #4360, opened 2021-03-18
Enable deterministic build

304NotModified:
See https://github.com/clairernovotny/DeterministicBuilds

related https://github.com/NLog/NLog.Web/pull/639

PR #4373, opened 2021-03-23
Enable deterministic build and SourceLink for all projects

304NotModified:
supersedes #4360

- Enabled sourcelink and deterministic build  for all projects
- Moved duplicate csproj properties to src/Directory.Build.props

Todo:
- [ ] test publish before merge!

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -1,0 +1,42 @@
+<Project>
+  <ItemGroup>
+    <SourceRoot Include="$(MSBuildThisFileDirectory)/"/>
+  </ItemGroup>
+  <PropertyGroup>
+    <RestoreProjectStyle>PackageReference</RestoreProjectStyle>
+
+    <Company>NLog</Company>
+    <Authors>Jarek Kowalski,Kim Christensen,Julian Verdurmen</Authors>
+    <CurrentYear>$([System.DateTime]::Now.ToString(yyyy))</CurrentYear>
+    <Copyright>Copyright (c) 2004-$(CurrentYear) NLog Project - https://nlog-project.org/ </Copyright>
+
+    <PackageIconUrl>https://nlog-project.org/N.png</PackageIconUrl>
+    <PackageProjectUrl>https://nlog-project.org/</PackageProjectUrl>
+    <PackageLicenseUrl>https://github.com/NLog/NLog/blob/master/LICENSE.txt</PackageLicenseUrl>
+    <PackageLicenseExpression>BSD-3-Clause</PackageLicenseExpression>
+    <RepositoryType>git</RepositoryType>
+    <RepositoryUrl>https://github.com/NLog/NLog.git</RepositoryUrl>
+    
+    <SignAssembly>true</SignAssembly>
+    <InformationalVersion>$(ProductVersion)</InformationalVersion>
+    <AssemblyVersion>4.0.0.0</AssemblyVersion>
+    <AssemblyOriginatorKeyFile>..\NLog.snk</AssemblyOriginatorKeyFile>
+    <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
+    <AssemblyTitle>$(Title)</AssemblyTitle>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <PackageIcon>N.png</PackageIcon>
+
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
+    <!-- EmbedUntrackedSources for deterministic build -->
+  </PropertyGroup>
+  <ItemGroup>
+    <None Include="N.png" Pack="true" PackagePath=""/>
+  </ItemGroup>
+  <Target Name="DownloadMissingContent" BeforeTargets="GenerateNuspec">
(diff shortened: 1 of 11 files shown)


snakefoot:
Notice that all projects are going to be moved into seperate repositories when NLog 5 has been released. No need to attach them together again.

304NotModified:
You could also move the directory.props and keep it easier in sync then?

snakefoot:
I find props-files to be a big unnecessary hammer, when only having 1 or 2 csproj in the solution.

304NotModified:
Maybe, but don't like to keep the csproj in sync between the repos? That's is far more easier with Directory.Build.props (and you could copy paste it fully)

src/Directory.Build.props
```diff
@@ -0,0 +1,42 @@
+<Project>
+  <ItemGroup>
+    <SourceRoot Include="$(MSBuildThisFileDirectory)/"/>
+  </ItemGroup>
+  <PropertyGroup>
+    <RestoreProjectStyle>PackageReference</RestoreProjectStyle>
```
snakefoot, line 6: This is not needed.

src/Directory.Build.props
```diff
@@ -0,0 +1,42 @@
...
+
+    <Company>NLog</Company>
+    <Authors>Jarek Kowalski,Kim Christensen,Julian Verdurmen</Authors>
```
snakefoot, line 9: This might not be true for all new nuget-packages

304NotModified: true, I will remove the directory.props solution

src/Directory.Build.props
```diff
@@ -0,0 +1,42 @@
...
+    <CurrentYear>$([System.DateTime]::Now.ToString(yyyy))</CurrentYear>
+    <Copyright>Copyright (c) 2004-$(CurrentYear) NLog Project - https://nlog-project.org/ </Copyright>
+
+    <PackageIconUrl>https://nlog-project.org/N.png</PackageIconUrl>
+    <PackageProjectUrl>https://nlog-project.org/</PackageProjectUrl>
+    <PackageLicenseUrl>https://github.com/NLog/NLog/blob/master/LICENSE.txt</PackageLicenseUrl>
+    <PackageLicenseExpression>BSD-3-Clause</PackageLicenseExpression>
```
snakefoot, line 16: Enforcing smae license all nuget-packages might not work with their dependencies.

src/Directory.Build.props
```diff
@@ -0,0 +1,42 @@
...
+    <RepositoryType>git</RepositoryType>
+    <RepositoryUrl>https://github.com/NLog/NLog.git</RepositoryUrl>
```
snakefoot, line 18: The repository url is not going to be the same

src/Directory.Build.props
```diff
@@ -0,0 +1,42 @@
...
+    
+    <SignAssembly>true</SignAssembly>
+    <InformationalVersion>$(ProductVersion)</InformationalVersion>
+    <AssemblyVersion>4.0.0.0</AssemblyVersion>
+    <AssemblyOriginatorKeyFile>..\NLog.snk</AssemblyOriginatorKeyFile>
+    <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
+    <AssemblyTitle>$(Title)</AssemblyTitle>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <PackageIcon>N.png</PackageIcon>
```
snakefoot, line 27: Think different nuget-packages should be allowed to have their own icon.

src/Directory.Build.props
```diff
@@ -0,0 +1,42 @@
...
+
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
+    <!-- EmbedUntrackedSources for deterministic build -->
+  </PropertyGroup>
+  <ItemGroup>
+    <None Include="N.png" Pack="true" PackagePath=""/>
+  </ItemGroup>
+  <Target Name="DownloadMissingContent" BeforeTargets="GenerateNuspec">
```
snakefoot, line 36: This is just me being lazy, and not knowing where to find the official icon for NLog.ExtensionLogging-nuget-package

But the correct way is having the PNG-file located in the repository, instead of downloading a random file from the internet on every build.

src/Directory.Build.props
```diff
@@ -0,0 +1,42 @@
...
+  </ItemGroup>
+  <PropertyGroup>
+    <RestoreProjectStyle>PackageReference</RestoreProjectStyle>
+
+    <Company>NLog</Company>
+    <Authors>Jarek Kowalski,Kim Christensen,Julian Verdurmen</Authors>
+    <CurrentYear>$([System.DateTime]::Now.ToString(yyyy))</CurrentYear>
+    <Copyright>Copyright (c) 2004-$(CurrentYear) NLog Project - https://nlog-project.org/ </Copyright>
+
+    <PackageIconUrl>https://nlog-project.org/N.png</PackageIconUrl>
```
snakefoot, line 13: These are now legacy and have been replaced by `<PackageIcon>`

src/Directory.Build.props
```diff
@@ -0,0 +1,42 @@
...
+    <PackageProjectUrl>https://nlog-project.org/</PackageProjectUrl>
```
snakefoot, line 14: Think different nuget-packages should be allowed to have their own project-page.

snakefoot (commented):
Added some comments to the contents of the props-file

PR #4399, opened 2021-04-15
Invalid JSON format when JProperty is null

vicentegnz:
Hi!!

We have noticed a strange behavior when using structured logs and the object that is passed to the NLog provider for logging is of type JObject, sometimes it generates a malformed JSON, which breaks the structured log.

An example of the log entry it generates is as follows.

`{"test":{"id": ,"description":"descriptionValue"}}`

The behavior I expect to have for these specific cases could be anything that does not generate an invalid JSON.

That's why I created this PR, I checked your development branch and it works correctly there, but in Master branch the change is not replicated yet. 

With the change I have applied the result of the JSON for these cases would be...

`{"test":{"id":"", "description":"descriptionValue"}}`

If you have any questions please let me know.

Thank you very much in advance!

### DIFF
--- a/src/NLog/Targets/DefaultJsonSerializer.cs
+++ b/src/NLog/Targets/DefaultJsonSerializer.cs
@@ -492,7 +492,7 @@ namespace NLog.Targets
             else
             {
                 string str = XmlHelper.XmlConvertToString(value, objTypeCode);
-                if (!forceToString && str != null && SkipQuotes(value, objTypeCode))
+                if (!forceToString && !string.IsNullOrEmpty(str) && SkipQuotes(value, objTypeCode))
                 {
                     destination.Append(str);
                 }
--- a/tests/NLog.UnitTests/Targets/DefaultJsonSerializerTestsBase.cs
+++ b/tests/NLog.UnitTests/Targets/DefaultJsonSerializerTestsBase.cs
@@ -37,6 +37,7 @@ using System.Collections.Generic;
 using System.Dynamic;
 using System.Globalization;
 using System.Text.RegularExpressions;
+using Newtonsoft.Json.Linq;
 using NLog.Config;
 using NLog.Internal;
 using NLog.Targets;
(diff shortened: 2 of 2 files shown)


snakefoot:
Thank you for reporting this bug, and also providing fix for the issue. Much appreciated.

Would it be possible that the unit-test can be changed to serialize a custom class, instead of using `JProperty` from `Newtonsoft.Json`? (Where the custom class behaves like the JProperty when given null-value).

We are working on splitting the NLog-project into smaller pieces, which also affects the NLog-Unit-Test-project as it will stop depending on the entire world. And then `Newtonsoft.Json`-dependency will no longer be available.

vicentegnz:
Hello again, thank you very much for your quick response.

Regarding the unit test, I tried not to use `Newtonsoft.Json` but when I tried to make "the mock class" I had no way to simulate the same scenario, the code gets complicated.

Do you think we could assume the change without adding the unit test? 

From my point of view, yes, the same change has already been added in the dev branch which will be the new NuGet project targeting version 5.... And of course, as there is no broken test, we can understand, that it is not a breaking change.

Anyway if you think of any other option, please let me know.


Thanks in advance

snakefoot:
When making bug-fixes then I prefer there is an unit-test to prevent regression.

I can understand from your reply, that you are not interested in spending time on improving the unit-test :)

snakefoot:
Created #4401 that implements unit-test without `Newtonsoft.Json` and also converts TypeCode.Empty to `null` instead of empty-string `""`.

vicentegnz:
Hi @snakefoot,

What I wanted to say is that I spent a lot of time trying to find the problem debugging the library and solve it in the best possible way, the best I could come up with was like this, I tried to follow the recommendation you suggested (unit test using mock class) but without luck (I guess because of low technical skill or library knowledge).

Anyway I appreciate your effort, and the quick solution you have provided, I have taken a look at it and it is much more accurate than mine  (`null` instead of `""`), 

If I can help you with anything else, just let me know.

Thank you very much again. :)

PS: As you have created another one with the new changes #4401, I will close this PR.

snakefoot:
> I tried to follow the recommendation you suggested (unit test using mock class) but without luck

Yes all the `Newtonsoft.Json`-types implements lots of different interfaces, so not easy to replicate. Also think it is more luck than skill, that NLog JsonSerializer is able to serialize the `Newtonsoft.Json`-types somewhat cleanly.

But it is very important that the NLog JsonSerializer works with all object-types, also the obscure ones like representing null with an object. So thank you again for reporting this issue.

snakefoot:
[NLog v4.7.10](https://www.nuget.org/packages/NLog/4.7.10) has been released.

PR #4401, opened 2021-04-16
JsonSerializer - Fixed bug when handling custom IConvertible returning TypeCode.Empty

snakefoot:
Alternative of #4399 to fix invalid JSON format when JProperty is null

### DIFF
--- a/src/NLog/Internal/StringHelpers.cs
+++ b/src/NLog/Internal/StringHelpers.cs
@@ -51,10 +51,7 @@ namespace NLog.Internal
         internal static bool IsNullOrWhiteSpace(string value)
         {
 #if NET3_5
-
-            if (value == null) return true;
-            if (value.Length == 0) return true;
-            return String.IsNullOrEmpty(value.Trim());
+            return value?.Length > 0 ? string.IsNullOrEmpty(value.Trim()) : true;
 #else
             return string.IsNullOrWhiteSpace(value);
 #endif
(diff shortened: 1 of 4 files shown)


snakefoot:
@vicentegnz No idea. It depends on when @304NotModified finds the optimal time-window.

src/NLog/Internal/StringHelpers.cs
```diff
@@ -51,10 +51,7 @@ public static class StringHelpers
         internal static bool IsNullOrWhiteSpace(string value)
         {
 #if NET3_5
-
-            if (value == null) return true;
-            if (value.Length == 0) return true;
-            return String.IsNullOrEmpty(value.Trim());
+            return value?.Length > 0 ? string.IsNullOrEmpty(value.Trim()) : true;
```
304NotModified, line 54: Do you find this really easier to read?

snakefoot: Yes I do, and the compiler also think it is better. But I can guess from your leading question, that I should change it back?

304NotModified: Hoe is this better for the compiler? I'm pretty sure that an equal comparison is better than a larger than in term of performance

snakefoot: Less instructions are usually better for the compiler, but I will revert. Since this is not important.

304NotModified (approved):

304NotModified:
> When will version 4.7.10 be released? will it come with this change, or is it postponed for later?

Will release it tomorrow :)

snakefoot:
[NLog v4.7.10](https://www.nuget.org/packages/NLog/4.7.10) has been released.

PR #4405, opened 2021-04-21
Logger OnLogEventInfoPrepared

valeriob:
Allow to customize each LogEventInfo independently

### DIFF
--- a/src/NLog/Logger.cs
+++ b/src/NLog/Logger.cs
@@ -73,6 +73,11 @@ namespace NLog
         public event EventHandler<EventArgs> LoggerReconfigured;
 
         /// <summary>
+        /// Occurs when logger has created the LogEventInfo
+        /// </summary>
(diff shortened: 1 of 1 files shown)


snakefoot:
Can you give an example how this should work? Lets say you have this class:
```c#
class MyClass
{
      private static readonly NLog.Logger Logger = NLog.LogManager.GetCurrentClassLogger();

      public MyClass(string contextId)
      {
             // What todo here ?
      }

      public void DoWork()
      {
            Logger.Info("Working Hard");
      }
}
```

And you have many threads allocating lots of `MyClass`-objects each object gets unique contextId as constructor-input-param.

How to ensure that each `MyClass`-object will not disturb the other instances of `MyClass`  when executing `DoWork`-concurrently.

valeriob:
Hi,
you are right, I rushed to it too quickly, I just moved the problem elsewhere, it's going to be impossible to subscribe to the event :D

PR #4421, opened 2021-05-06
LoggingRule - Added FinalMinLevel to quit further processing

snakefoot:
Instead of having to specify both `maxLevel="Warn"` + `final="true"` then it could be nice a combined-property, with a name that makes it easy to understand. 

FinalMinLevel signals that all following catch-all rules will also have this MinLevel applied when matching same logger:
 - Extended version of `Final=true` that only applies when LogLevel-severity is below the specified Level.

```xml
<rules>
    <!--Skip non-critical Microsoft logs and so log only own logs (BlackHole) -->
    <logger name="Microsoft.*" finalMinLevel="Warn" />
    <logger name="System.Net.Http.*" finalMinLevel="Warn" />

    <logger name="*" minlevel="Trace" writeTo="ownFile-web" />
</rules>
```

Other possible names:
- FinalLevel / FinalLogLevel (Similar to Microsoft Extension Logging)
- RestrictMinLevel (Similar to Serilog)
- FinalThreshold
- FinalStartLevel
- FinalAcceptLevel
- FinalBelowLevel
- SuppressMinLevel
- CapMinLevel
- FinalizeMaxLevel
- BlockMaxLevel
- TrapMaxLevel
- DrainMaxLevel
- SapMaxLevel
- TruncateMaxLevel
- FinalMaxLevel (Signal that both MaxLevel and Final is configured in one go)

The same property could also provide a more elegant solution for this simple application example:

```xml
<nlog>
    <targets>
        <target name="logconsole" xsi:type="Console" />
    </targets>

    <rules>
        <logger name="*" minlevel="Info" writeTo="logconsole" />
    </rules>
</nlog>
```

Where the application gets a new exciting sub-component is added, that is a little noisy. So you redirect to its own isolated target:

```xml
<nlog>
    <targets>
        <target name="logconsole" xsi:type="Console" />
        <target name="logfile" xsi:type="File" fileName="requests.txt" />
    </targets>

    <rules>
        <logger name="Requests" minlevel="Debug" writeTo="logfile" final="true" />
        <logger name="*" minlevel="Info" writeTo="logconsole" />
    </rules>
</nlog>
```

Then you discover that some of the noise is rather important, then you need to add it twice (Having to add your `catch-all` target several times):

```xml
<nlog>
    <targets>
        <target name="logconsole" xsi:type="Console" />
        <target name="logfile" xsi:type="File" fileName="requests.txt" />
    </targets>

    <rules>
        <logger name="Requests" minlevel="Warn" writeTo="logconsole" />
        <logger name="Requests" minlevel="Debug" writeTo="logfile" final="true" />
        <logger name="*" minlevel="Info" writeTo="logconsole" />
    </rules>
</nlog>
```

This pull-request add support for `finalThreshold`, that works like a conditional-final (Without enforcing MaxLevel for the LoggingRule, so both logging-rules will receive the error-logevents):

```xml
<nlog>
    <targets>
        <target name="logconsole" xsi:type="Console" />
        <target name="logfile" xsi:type="File" fileName="requests.txt" />
    </targets>

    <rules>
        <logger name="Requests" writeTo="logfile" minLevel="Debug" finalMinLevel="Warn" />
        <logger name="*" minlevel="Info" writeTo="logconsole" />
    </rules>
</nlog>
```

### DIFF
--- a/src/NLog/Config/LoggingConfigurationParser.cs
+++ b/src/NLog/Config/LoggingConfigurationParser.cs
@@ -582,6 +582,7 @@ namespace NLog.Config
         {
             string minLevel = null;
             string maxLevel = null;
+            string finalMinLevel = null;
             string enableLevels = null;
 
             string ruleName = null;
@@ -628,6 +629,9 @@ namespace NLog.Config
                     case "MAXLEVEL":
                         maxLevel = childProperty.Value;
                         break;
+                    case "FINALMINLEVEL":
+                        finalMinLevel = childProperty.Value;
+                        break;
                     case "FILTERDEFAULTACTION":
                         filterDefaultAction = childProperty.Value;
                         break;
@@ -659,6 +663,15 @@ namespace NLog.Config
                 LoggerNamePattern = namePattern,
                 Final = final,
             };
+
+            if (!string.IsNullOrEmpty(finalMinLevel))
+            {
+                rule.FinalMinLevel = LogLevelFromString(finalMinLevel);
+                if (string.IsNullOrEmpty(enableLevels) && string.IsNullOrEmpty(minLevel))
+                {
+                    minLevel = finalMinLevel;
+                }
+            }
 
             EnableLevelsForRule(rule, enableLevels, minLevel, maxLevel);
 
--- a/src/NLog/Config/LoggingRule.cs
+++ b/src/NLog/Config/LoggingRule.cs
@@ -141,6 +141,11 @@ namespace NLog.Config
         public bool Final { get; set; }
 
         /// <summary>
+        /// Gets or sets a LogLevel whether to quit processing any further rule for loglevels with lower severity.
+        /// </summary>
+        public LogLevel FinalMinLevel { get; set; }
+
+        /// <summary>
         /// Gets or sets logger name pattern.
         /// </summary>
         /// <remarks>
--- a/src/NLog/Filters/ConditionBasedFilter.cs
+++ b/src/NLog/Filters/ConditionBasedFilter.cs
@@ -33,8 +33,8 @@
 
 namespace NLog.Filters
 {
-    using Conditions;
-    using Config;
+    using NLog.Conditions;
+    using NLog.Config;
 
     /// <summary>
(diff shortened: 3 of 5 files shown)


304NotModified:
To be honest, I understand the reason, but IMO it's a too specific case for this attribute. It's already hard to explain "final"

snakefoot:
I'm fine with a different name than FinalMaxLevel.


Sent from my Sony Xperia


To be honest, I understand the reason, but IMO it's a too specific case for this attribute. It's already hard to explain "final"

304NotModified:
It's the use case, not the name.

snakefoot:
I think final is not working well and want to replace it.


Sent from my Sony Xperia


It's the use case, not the name.

304NotModified:
Maybe it would be nice if final could be a condition, but that's probably a lot slower and maybe also a lot more work.

snakefoot:
Think having it as Loglevel only is fine. I like the ease of Microsoft Logging Filters.


Right now one have to fiddle with Maxlevel + Final=true to solve a very common usecase.


By having a setting like DrainMaxLevel / FinalMaxLevel / RestrictMaxLevel then it is a single setting and it will have an added bonus feature (not needing 2 rules for handling a single logger)


Sent from my Sony Xperia


Maybe it would be nice if final could be a condition, but that's probably a lot slower and maybe also a lot more work.

snakefoot:
@304NotModified Renamed from `FinalMaxLevel` to `FinalThreshold` and also updated the original-post with better explanation for this change.

snakefoot:
Must admit I kind of like `FinalMinLevel` better than `FinalThreshold` (Though it ressembles [GlobalThreshold](https://nlog-project.org/documentation/v4.7.0/html/P_NLog_LogManager_GlobalThreshold.htm)).

- **FinalMinLevel** - Signals that all following catch-all rules will also have this MinLevel applied for the matching loggers

snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/Configuration-file#rules

Issue #2870, opened 2018-08-24
Move public classes from internal namespace


304NotModified:
Some classes are public, but in the internal namespace

snakefoot:
Guess it means duplicate classes. One obsolete as public, and new internal non-obsolete. Going to be confusing :)

304NotModified:
I don't think we need here obsolete attributes, as in general it's always not advised to use classes meant to be internal (so .Internal namespace or .Private etc).

snakefoot:
Happy we are back to grey in grey and not black and white :)

snakefoot:
Almost resolved by #4331 . Just need to figure out what to do about IAppDomain.

snakefoot:
Have created  #4433 to mark `LogFactory.CurrentAppDomain` as obsolete

snakefoot:
Closing issue as resolved by #4331. And now IAppDomain has become obsolete, and will be part of normal cleanup. See also #4433

PR #4433, opened 2021-05-18
Marked LogFactory.CurrentAppDomain as obsolete

snakefoot:
Tell users to leave the house, before it is burned down. See also #3995 and #2870


304NotModified (approved):
good idea!

snakefoot:
My refactorings introduced a bug in EventLogTarget, so it failed to initialize EventLog-Source with correct default-value.

Now added explicit unit-test, and also marked `Source`-property as `[RequiredParameter]`

PR #4306, opened 2021-02-14
Typed Layout for easy conversion from Layout to actual value

snakefoot:
Completed my own version of #3656. That resolves #1811 and resolves #3994

With full optimization of layout-precalculation and caching of previous value to reduce boxing.

Created my own version because I wasn't sure #3656 would be completed. Maybe steal the precalculate-logic from this PR ?

Notice that I have decided to always use the `Layout<T>` type-converter-abilities, instead of depending on `PropertyHelper`. This makes the behavior more predictable for users creating their own `Layout<T>`.

PR #4436, opened 2021-05-20
Layout Typed - Removed StaticValue and implicit conversion to avoid surprises

snakefoot:
It was a mistake to introduce the StaticValue-property (Caching stale data)

And implicit conversion of dynamic `Layout<T>` without FixedValue should not default to `default(T)`.

Follow up to #4306


snakefoot:
One might even consider removing `IEquatable<T>` + `Equals(T)` and comparison-operators. As they only works when `IsFixed`.

PR #4351, opened 2021-03-12
Extracted DatabaseTarget into NLog.Database with help from ValueTypeLayoutInfo

snakefoot:
Resolves #2676 and ensures NLog is not generating security vulnerability warnings #4051

ValueTypeLayoutInfo has also been included in:
- TargetWithContext
- MethodCallTarget

ValueTypeLayoutInfo is special as it uses `Layout<T>` to handle type-conversion, while ensuring layout-precalculate always happens on `Layout<T>`-instance (and not the SimpleLayout assigned from config). This makes performance optimizations in `Layout<T>` available from `NLog.config`.

Guess later PR can update JsonLayout and XmlLayout to also use ValueTypeLayoutInfo

Issue #3136, opened 2019-02-22
Expose irawvalue to public


304NotModified:

304NotModified:
And check namespace

304NotModified:
For NLog.web etc

snakefoot:
Could consider keeping IRawValue internal, and just reveal `Layout.TryGetRawValue()` since it is the only one needed for targets like DatabaseTarget.

snakefoot:
Also think the interface should be renamed to `IRenderRawValue` (And IStringValueRenderer should become `IRenderStringValue`).

304NotModified:
Maybe then `IRawRenderable` and `IStringRenderable`? We have already `IRenderable`

(Render is more for a method name than a class/interface name IMO)

snakefoot:
Then I think IRawValueRenderable and IStringValueRenderable are better.


Sent from my Sony Xperia


---- Julian Verdurmen wrote ----


Maybe then IRawRenderable and IStringRenderable? We have already IRenderable

(Render is more for a method name than a class/interface name IMO)

304NotModified:
OK let's do that (the rename). To public will be probably later than 5.0

snakefoot:
`Layout<T>` opens the access to `IRawValue`. And like how `Layout<T>` allows one to handle constant-values without boxing.

But since `Layout<T>` actually knows what type it wants, then it is a little sad that `IRawValue` always forces boxing. Think `Layout<T>` should have a different friend like `ILayoutRenderer<T>`.

Could also be nice to handle this scenario, where writing the `${date}` as Json-DateTime without constant boxing.

```c#
new JsonAttribute("date", "${date}") { ValueType = typeof(DateTime) },
```

But similar boxing-issue also exist for the object-reflection-logic, where the property-value-extractor also boxes all property-values.

304NotModified:
How would `ILayoutRenderer<T>` would look like?

>  it is a little sad that IRawValue always forces boxing

Agree. I don't know if we could create a new interface (E.g. IRawValueTypeValue) for values types?

snakefoot:
> How would ILayoutRenderer<T> would look like?

Maybe something like this:

```c#
interface ILayoutRenderer
{
}

interface ILayoutRenderer<T> : ILayoutRenderer
{
    T RenderValue(LogEventInfo logEvent);
}
```

PR #4469, opened 2021-06-19
JsonAttribute + XmlAttribute with ValueType support for encoding the raw-value

snakefoot:
Continue the work from #4351

- JsonAttribute can render the raw-value as Json.
- XmlAttribute can render the raw-value as Xml (but only simple values).

The performance is not super-hot, since the raw-value has to travel through the layout-logic. And if raw-value is struct-type then it will always be boxed. Also added my thoughts to #3136


snakefoot:
After trying to use ValueTypeLayoutInfo more places, then I think it should be implemented differently. I don't like that one is required to inherit from the ValueTypeLayoutInfo.

PR #4470, opened 2021-06-19
XmlLayout + JsonLayout - ExcludeProperties using StringComparer.OrdinalIgnoreCase

snakefoot:


304NotModified:
Nice. Maybe good idea to add unit tests?

snakefoot:
Yes more unit-tests should be made, especially since the unit-tests from the Database-Target might move into its own repository.

After trying to use ValueTypeLayoutInfo more places, then I think it should be implemented differently. I don't like that one is required to inherit from the ValueTypeLayoutInfo.

appveyor.yml
```diff
@@ -1,6 +1,6 @@
 version: 4.7.0-{build}
 image:
-  - Visual Studio 2019
+  - Previous Visual Studio 2019
```
304NotModified, line 3: Why do we need this?

snakefoot:
Because Microsoft broke Visual Studio in ver. 16.10. So rolling back to 16.9 until AppVeyor has updated their latest image with bugfix.


Sent from my Sony Xperia


In appveyor.yml<https://github.com/NLog/NLog/pull/4470#discussion_r655777374>:

> @@ -1,6 +1,6 @@
 version: 4.7.0-{build}
 image:
-  - Visual Studio 2019
+  - Previous Visual Studio 2019


Why do we need this?

Issue #4474, opened 2021-06-23
Set User-Agent header for WebServiceTarget


avdynut:
`WebServiceTarget` with 'User-Agent' header doesn't work.

**NLog version**: 4.7.10

```
var webTarget = LogManager.Configuration.FindTargetByName<WebServiceTarget>("targetName");
webTarget.Headers.Add(new MethodCallParameter("User-Agent", Layout.FromString("Custom user agent")));
```
There is an exception in the internal log
```
Error WebService Target[logRemote]: Error has been raised. Exception: System.ArgumentException: The 'User-Agent' header must be modified using the appropriate property or method.
Parameter name: name
   at System.Net.WebHeaderCollection.ThrowOnRestrictedHeader(String headerName)
   at System.Net.WebHeaderCollection.Set(String name, String value)
   at NLog.Targets.WebServiceTarget.DoInvoke(Object[] parameters, AsyncLogEventInfo logEvent)
   at NLog.Targets.MethodCallTargetBase.Write(AsyncLogEventInfo logEvent)
   at NLog.Targets.Target.WriteAsyncThreadSafe(AsyncLogEventInfo logEvent)
   at NLog.Targets.Target.WriteAsyncLogEvent(AsyncLogEventInfo logEvent)
```

As I see, `WebServiceTarget` should have 'UserAgent' property that will be set to appropriate `HttpWebRequest`'s property.

snakefoot:
Yes UserAgent-Header is special and needs to be assigned through [HttpRequest.UserAgent Property](https://docs.microsoft.com/en-us/dotnet/api/system.web.httprequest.useragent)

Created possible solution here: #4475

avdynut:
Thanks, it would be nice, if you add this fix to v4.7.11

snakefoot:
The NLog 4.7-branch is mostly for bug-fixes. But I guess if @304NotModified is in the mood for executing the whole release-exercise then I guess it could be done.

304NotModified:
Sure no problem for creating a 4.7.11

304NotModified:
Created #4489 for NLog 4

snakefoot:
Think I prefer to have UserAgent as Layout property instead of mixing with headers. It will also be easier explain in documentation

________________________________

snakefoot:
NLog v4.7.11 has been released - https://www.nuget.org/packages/NLog/4.7.11

PR #4475, opened 2021-06-23
WebServiceTarget - Added support for assigning UserAgent-Header

snakefoot:
Possible solution for #4474. So you can do this:

`<UserAgent>SecretAgent</UserAgent>`


snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/WebService-target

PR #4477, opened 2021-06-24
AsyncTargetWrapper - Obsoleted StartInstantWriterTimer as logic has grown more advanced

snakefoot:


snakefoot:
Hrm looks like the lock-throttle in blocking-mode hurts a lot more for the "new" ConcurrentRequestQueue, when constantly hitting the RequestLimit. Think I will close this issue until I find something that I like.

PR #4484, opened 2021-07-03
Fixed SimpleLayout TryGetRawValue() when LiteralWithRawValueLayoutRenderer

snakefoot:
Followup to #3997

### DIFF
--- a/src/NLog/Layouts/Layout.cs
+++ b/src/NLog/Layouts/Layout.cs
@@ -54,7 +54,7 @@ namespace NLog.Layouts
         /// Is this layout initialized? See <see cref="Initialize(NLog.Config.LoggingConfiguration)"/>
         /// </summary>
         internal bool IsInitialized;
-        private bool _scannedForObjects;
+        internal bool _scannedForObjects;
 
         /// <summary>
         /// Gets a value indicating whether this layout is thread-agnostic (can be rendered on any thread).
--- a/tests/NLog.UnitTests/Layouts/SimpleLayoutParserTests.cs
+++ b/tests/NLog.UnitTests/Layouts/SimpleLayoutParserTests.cs
@@ -813,14 +813,12 @@ namespace NLog.UnitTests.Layouts
             var layout = (SimpleLayout)Layout.FromString(input);
 
             // Assert
+            Assert.True(layout.IsFixedText);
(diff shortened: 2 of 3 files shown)


src/NLog/Layouts/Layout.cs
```diff
@@ -54,7 +54,7 @@ public abstract class Layout : ISupportsInitialize, IRenderable
         /// Is this layout initialized? See <see cref="Initialize(NLog.Config.LoggingConfiguration)"/>
         /// </summary>
         internal bool IsInitialized;
-        private bool _scannedForObjects;
+        internal bool _scannedForObjects;
```
304NotModified, line 57: @snakefoot why is internal needed?

tests/NLog.UnitTests/Layouts/SimpleLayoutParserTests.cs
```diff
@@ -813,14 +813,12 @@ public void Parse_AppDomainFixedOutputWithRawValue_ConvertSingleToLiteralAndKeep
...
 
             // Assert
+            Assert.True(layout.IsFixedText);
             var single = Assert.Single(layout.Renderers);
-            var singleRaw = Assert.IsType<LiteralWithRawValueLayoutRenderer>(single);
-            var succeeded = singleRaw.TryGetRawValue(LogEventInfo.CreateNullEvent(), out var rawValue);
+            Assert.IsType<LiteralWithRawValueLayoutRenderer>(single);
+            var succeeded = layout.TryGetRawValue(LogEventInfo.CreateNullEvent(), out var rawValue);
             Assert.True(succeeded);
-            var rawValueInt = Assert.IsType<int>(rawValue);
```
304NotModified, line 820: @snakefoot  Why is the assert of int type not needed anymore?

snakefoot: The cast will fail. But the main test is the output from `layout.TryGetRawValue()` instead of checking the result of the internal renderer.

src/NLog/Layouts/Layout.cs
snakefoot: Yes that was just me trying to skip the scanning of object-graph when seeing that it was "fixed"-value. Will improve on this, so the optimization will be true.

PR #4488, opened 2021-07-08
ConfigureAwait(false) for all awaits

304NotModified:


snakefoot (commented):
When using `ConfigureAwait(false)` then one says that async operation doesn't have to complete on the same execution-context.

Which is perfect for the methods that doesn't return their task. But for `SwallowAsync` then it might give new behavior when called from a GUI-/ASP-context, as the execution might continue on a new non-gui-thread.

But since the general advice for all libraries is to ALWAYS use `ConfigureAwait(false)`  then the change in behavior is probably for the better.

PR #4501, opened 2021-07-23
TargetWithFilterChain - FilterChain becomes empty array, when no filters

snakefoot:
Faster filter check when no filters. Another micro-optimization.


snakefoot:
Looks like appveyor stable and previous image are both pointing to a broken  .NET SDK 3.0.301. Waiting for the next update that will include  .NET SDK 3.0.302

PR #4502, opened 2021-07-24
RetryingWrapper - Added support for EnableBatchWrite

snakefoot:
Maybe consider having `EnableBatchWrite = true` by default ?

Also recognized that when calling `AsyncLogEventInfo.Continue(ex)` then one should be a in "stable" place. Not holding locks or resources. This is expecially important when using RetryingWrapper, that will perform recursive call on the same target. Imagine processing a single bucket from `BucketSort()` and suddenly perform try retry-logic while in the middle of handling one of the buckets. The target might not be in a state where it can handle a recursive call.

Similar issue can happen with FallbackWrapper, where one is deep-down into a callstack where logevent fails, and then FallbackWrapper causes one to jump further down the rabbit-hole to forward the logevent to another target.

### DIFF
--- a/src/NLog/Targets/Wrappers/RetryingTargetWrapper.cs
+++ b/src/NLog/Targets/Wrappers/RetryingTargetWrapper.cs
@@ -110,6 +110,13 @@ namespace NLog.Targets.Wrappers
         public int RetryDelayMilliseconds { get; set; }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


304NotModified:
> Maybe consider having EnableBatchWrite = true by default ?

Sounds good to me. :)

snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/RetryingWrapper-target

PR #4523, opened 2021-08-21
all-event-properties with IncludeScopeProperties

snakefoot:
Also `${all-event-properties}` now reacts to format-specifier like `@` (serialize value as json), instead of always doing `ToString()`


snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/All-Event-Properties-Layout-Renderer

PR #4524, opened 2021-08-21
Fixed LibLog integration that references obsolete methods

snakefoot:
LibLog explodes when failing to hookup with NLog. Hopefully when NLog 6 gets out, then LibLog is fully deprecated.

### DIFF
--- a/src/NLog/Config/ConfigurationItemFactory.cs
+++ b/src/NLog/Config/ConfigurationItemFactory.cs
@@ -228,9 +228,9 @@ namespace NLog.Config
                 {
                     return false;
                 }
-                else if (messageFormatter is LogMessageTemplateFormatter messageTemplateFormatter)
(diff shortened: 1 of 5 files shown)


304NotModified:
Isn't better to send a PR to liblog?

snakefoot:
> Isn't better to send a PR to liblog?

LibLog is obsolete, and will not receive any updates. Just like Common.Logging

304NotModified:
Ow yes indeed. Ok good call for this :)

304NotModified:
Another option would be to create a special package for liblog. E.g. NLog.LibLogIntegration

304NotModified:
(ps I'm a bit confused by the other changes. Are these on purpose?)

snakefoot:
> (ps I'm a bit confused by the other changes. Are these on purpose?)

I was doing a performance test with LibLog, and then several issues popped out. I just took worst one as issue-title.

304NotModified:
Ok!

PR #4535, opened 2021-08-27
Refactor: Cleanup get setting string

304NotModified:
small refactor, same results, less lines


snakefoot:
Right now it performs fallback to environment-variable, when appsettings fails.

With this change then when appsettings fail, then no fallback.

304NotModified:
Yes indeed, bad idea this change. Thanks!

PR #4536, opened 2021-08-27
Add aliases for layout renderers

304NotModified:
To make things consistent

But another way would be to ignore `-` when parsing.


304NotModified:
superseded by #4695

PR #4537, opened 2021-08-27
Ignore - when parsing layout renderers

304NotModified:
Just an idea, 

Because [the list](https://nlog-project.org/config/?tab=layout-renderers) is inconsistent. 

@snakefoot please share your opinion.

supersedes https://github.com/NLog/NLog/pull/4536

PS The replace could be more efficient, but not sure if that's is an problem. We aren't parsing when logging

### DIFF
--- a/src/NLog/Config/Factory.cs
+++ b/src/NLog/Config/Factory.cs
@@ -52,11 +52,13 @@ namespace NLog.Config
         private readonly Dictionary<string, GetTypeDelegate> _items = new Dictionary<string, GetTypeDelegate>(StringComparer.OrdinalIgnoreCase);
         private readonly ServiceRepository _serviceRepository;
         private readonly Factory<TBaseType, TAttributeType> _globalDefaultFactory;
+        private List<string> _ignoreCharacters;
 
-        internal Factory(ServiceRepository serviceRepository, Factory<TBaseType, TAttributeType> globalDefaultFactory)
+        internal Factory(ServiceRepository serviceRepository, Factory<TBaseType, TAttributeType> globalDefaultFactory, List<string> ignoreCharacters = null)
         {
             _serviceRepository = serviceRepository;
             _globalDefaultFactory = globalDefaultFactory;
+            _ignoreCharacters = ignoreCharacters;
         }
 
         private delegate Type GetTypeDelegate();
@@ -125,6 +127,7 @@ namespace NLog.Config
         /// <param name="typeName">Name of the type.</param>
         public void RegisterNamedType(string itemName, string typeName)
         {
+            itemName = NormalizeName(itemName);
             _items[itemName] = () => Type.GetType(typeName, false);
         }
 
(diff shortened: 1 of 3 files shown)


src/NLog/Config/Factory.cs
```diff
@@ -253,8 +274,10 @@ class LayoutRendererFactory : Factory<LayoutRenderer, LayoutRendererAttribute>
     {
         private Dictionary<string, FuncLayoutRenderer> _funcRenderers;
         private readonly LayoutRendererFactory _globalDefaultFactory;
+        private static readonly List<string> IgnoreCharacters = new List<string> { "-" };
```
snakefoot, line 277: Could this be `char[]` instead ?

src/NLog/Config/Factory.cs
```diff
@@ -146,6 +149,22 @@ public void RegisterDefinition(string itemName, Type itemDefinition)
...
         }
 
+        private string NormalizeName(string itemName)
+        {
+            if (_ignoreCharacters == null)
+            {
+                return itemName;
+            }
+
+            for (int i = 0; i < _ignoreCharacters.Count; i++)
```
snakefoot, line 159: Could this be `foreach` instead ?

src/NLog/Config/Factory.cs
```diff
@@ -253,8 +274,10 @@ class LayoutRendererFactory : Factory<LayoutRenderer, LayoutRendererAttribute>
     {
         private Dictionary<string, FuncLayoutRenderer> _funcRenderers;
         private readonly LayoutRendererFactory _globalDefaultFactory;
+        private static readonly List<string> IgnoreCharacters = new List<string> { "-" };
 
-        public LayoutRendererFactory(ServiceRepository serviceRepository, LayoutRendererFactory globalDefaultFactory) : base(serviceRepository, globalDefaultFactory)
+        public LayoutRendererFactory(ServiceRepository serviceRepository, LayoutRendererFactory globalDefaultFactory)
+            : base(serviceRepository, globalDefaultFactory, IgnoreCharacters)
```
snakefoot, line 280: Why only layout-renderers ?

snakefoot (approved):
Must admit I prefer just using the alias support, but if you feel this is needed, then lets go.

304NotModified:
> Must admit I prefer just using the alias support, but if you feel this is needed, then lets go.

No necessary. But it's hard to get things consistent and document it. We're already ignoring casing, so maybe this will also help.

src/NLog/Config/Factory.cs
304NotModified: sure, but AFAIK `for` was preferred over `foreach` (allocations?)

src/NLog/Config/Factory.cs
304NotModified: good question! Dunno!

src/NLog/Config/Factory.cs
304NotModified: sure :)

src/NLog/Config/Factory.cs
304NotModified: (will fix this)

src/NLog/Config/Factory.cs
snakefoot: `foreach` only allocates when using interfaces like `IEnumerable` or `IList`. Not when using actual implementation like `char[]` or `List<string>`

snakefoot:
When completing this, then also revert some the aliases added with #3004

304NotModified:
superseded by #https://github.com/NLog/NLog/pull/4695

PR #4538, opened 2021-08-28
LayoutParser with support for ${newline} in string-value

snakefoot:
Not super pretty, but it works


snakefoot:
Replaced by #4583

PR #4539, opened 2021-08-28
${Exception}: Support for ${newline} in separators

304NotModified:
Another poor man's solution... 👼 

Please share your thoughts @snakefoot :)

### DIFF
--- a/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/ExceptionLayoutRenderer.cs
@@ -40,6 +40,8 @@ namespace NLog.LayoutRenderers
     using NLog.Common;
     using NLog.Config;
     using NLog.Internal;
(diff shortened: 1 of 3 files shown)


snakefoot (commented):
You need to handle all `\`-logic, not just \r\n. And at the same time also handle unicode-character-logic.

There is a lot of features supported in the string-property, that is not supported for Layout-property.

And it might be hard to "repair" afterwards because of how the LayoutParser works.

snakefoot:
But I really like that the property-type becomes `Layout` instead of `string` (with all the extra complexity that gives), because it signals that Layout-logic is supported. Have updated  #4565 with an extra commit that adds the method `HandleMissingUnicodeStringEscape`.

snakefoot:
Created #4583 as alternative implementation.

snakefoot:
Replaced by  #4583

PR #4541, opened 2021-08-31
Implementation of TcpThreadNetworkSender

RVINAA:
Hi, I'm creating this pull request to add a new TCP network sender with the following reasons:

In our enviroment, with an average of <100k logs per minute; when using a network target with a tcp sender, we've found memory usage to go crazy.

The new proposed sender is based on a  dedicated thread making the actual network calls, as to avoid async callbacks which under high volume impose additional overhead/delays.

However, we understand this implementation might apply only to scenarios with high (network) logging volumes, and thus we thought it better to provide it as alternative to the actual TcpNetworkSender, by using a new uri prefix (tcp+thread://).

Thorough testing has been done against this implementation, trying different cases like, timeouts, disconnections in the middle of a run, in order to test sender recovery, etc.

Regarding performance, we can appreciate that (with high network output):
- The difference in memory is quite significant, with the original sender consuming a total of 2GB compared to this implementation, which consumed about 40MB in a stable way. 
- With our implementation, the total logs generated in one minute have been sent out without problems, but with the original sender only 1/3 of the generated logs have arrived (giving exception in the flush as they are still queued in the sender and cannot be dequeued in less than 15secs).

_These benchmarks have taken into consideration the tuning recommendations to be applied in the configuration file._ 

-------

### The image below is of the original sender in the course of one minute, where after having generated a total of **1.9M logs**, the **ncat has only received 440k lines and the flush cannot complete**.
image

### This bench was with our thread sender; **all the generated logs (1.8M) were received in ncat (no loss) and memory usage is constant** 
image

-------

## In conclusion
**In this scenario our implementation seems to fit well and we believe that both implementations should coexist.**

Also we provide unit tests for this new TcpSender which replicate the ones found for existing ones.

Any question or missing information.. feel free to ask.
Thanks for your time!

### DIFF
--- a/src/NLog/Internal/NetworkSenders/INetworkSenderFactory.cs
+++ b/src/NLog/Internal/NetworkSenders/INetworkSenderFactory.cs
@@ -47,9 +47,10 @@ namespace NLog.Internal.NetworkSenders
         /// <param name="maxQueueSize">The maximum queue size.</param>
         /// <param name="sslProtocols">SSL protocols for TCP</param>
         /// <param name="keepAliveTime">KeepAliveTime for TCP</param>
+        /// <param name="connectionTimeout">ConnectionTimeout for TcpThreadNetworkSender</param>
         /// <returns>
         /// A newly created network sender.
         /// </returns>
-        NetworkSender Create(string url, int maxQueueSize, System.Security.Authentication.SslProtocols sslProtocols, TimeSpan keepAliveTime);
+        NetworkSender Create(string url, int maxQueueSize, System.Security.Authentication.SslProtocols sslProtocols, TimeSpan keepAliveTime, TimeSpan connectionTimeout);
     }
 }
--- a/src/NLog/Internal/NetworkSenders/ISocket.cs
+++ b/src/NLog/Internal/NetworkSenders/ISocket.cs
@@ -34,6 +34,8 @@
 
 namespace NLog.Internal.NetworkSenders
 {
+    using System;
+    using System.Net;
     using System.Net.Sockets;
 
     /// <summary>
@@ -48,5 +50,17 @@ namespace NLog.Internal.NetworkSenders
         bool SendAsync(SocketAsyncEventArgs args);
 
         bool SendToAsync(SocketAsyncEventArgs args);
+
+#if !NET35 && !NETSTANDARD1_3 && !NETSTANDARD1_5
+        bool Connected { get; }
+
+        IAsyncResult BeginConnect(EndPoint remoteEP, AsyncCallback callback, object state);
+
+        void EndConnect(IAsyncResult asyncResult);
+
+        int Send(byte[] buffer, int offset, int size, SocketFlags socketFlags);
+#endif
+
+        void Dispose();
     }
 }
--- a/src/NLog/Internal/NetworkSenders/NetworkSender.cs
+++ b/src/NLog/Internal/NetworkSenders/NetworkSender.cs
@@ -69,6 +69,11 @@ namespace NLog.Internal.NetworkSenders
         public int LastSendTime { get; private set; }
 
         /// <summary>
+        /// Indicates wether this sender should be released on error.
+        /// </summary>
+        protected internal bool ReleaseOnError { get; protected set; } = true;
+
+        /// <summary>
         /// Initializes this network sender.
         /// </summary>
         public void Initialize()
--- a/src/NLog/Internal/NetworkSenders/NetworkSenderFactory.cs
+++ b/src/NLog/Internal/NetworkSenders/NetworkSenderFactory.cs
@@ -44,7 +44,7 @@ namespace NLog.Internal.NetworkSenders
         public static readonly INetworkSenderFactory Default = new NetworkSenderFactory();
 
         /// <inheritdoc />
-        public NetworkSender Create(string url, int maxQueueSize, System.Security.Authentication.SslProtocols sslProtocols, TimeSpan keepAliveTime)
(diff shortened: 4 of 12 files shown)


snakefoot:
I know the feeling of all children being ugly, and especially those not invented by one self.

You say that you see high memory allocation rate, when using the existing NetworkTarget. And you say that your new NetworkSender has lower memory allocation rate.

I can see that your new NetworkSender includes a BlockingCollection, so the producer is now blocked when consumer (network socket) is unable to keep up.

Maybe the issue is that you want the existing NetworkSender to block on overflow, instead of discarding?

Have you tried to perform profiling of what objects are being allocated by the existing NetworkTarget?

Maybe you could share the NLog.config that was used when testing the existing NetworkTarget?

RVINAA:
> I know the feeling of all children being ugly, and especially those not invented by one self.
> 
> You say that you see high memory allocation rate, when using the existing NetworkTarget. And you say that your new NetworkSender has lower memory allocation rate.
>
> I can see that your new NetworkSender includes a BlockingCollection, so the producer is now blocked when consumer (network socket) is unable to keep up.
>
> Maybe the issue is that you want the existing NetworkSender to block on overflow, instead of discarding?

**Our intention is to avoid losing logs unless necessary**, such as when the destination is not available.

- While sender is connected, logs aren't lost.
- If sender cannot connect.. this is where we clear the blocking collection (to avoid memory overload) the same as a FileAppender cannot write into a file (these logs are lost).
- This blocking collection doesn't fill the memory as we handle discarding on failure connection.
-  **The idea is to use this network target with the async wrapper as we've seen the thread is not blocked with the log call (similar as FileAppender + AsyncWrapper)**.

> Have you tried to perform profiling of what objects are being allocated by the existing NetworkTarget?
>

As mentioned above, with **high log traffic**, doing async writes on the socket.. serialized one per one.. with a lock.. causes saturation in the Queue<NetworkRequestArgs>.

> Maybe you could share the NLog.config that was used when testing the existing NetworkTarget?

Yes sir, will try to paste the related configuration for the Tcp sender + default async wrapper.
```
(...)
<time type="FastUTC" />
<variable name="Scope" value ="${when:when=length('${ndc}')>0:inner=${ndc}:else=${ndlc}}" />
<targets>
	<default-wrapper xsi:type="asyncwrapper" timetosleepbetweenbatches="0" overflowaction="grow" batchsize="1000" />

	<target xsi:type="Network" name="TcpNetworkSender" address="tcp://127.0.0.1:5001" newLine="true" lineEnding="LF">
		<layout xsi:type="JsonLayout" includeAllProperties="true" excludeEmptyProperties="true" excludeProperties="" includeGdc="true" includeMdc="true" includeMdlc="true" renderEmptyObject="false" escapeForwardSlash="false">
			<attribute name="@timestamp" layout="${date:format=yyyy-MM-ddTHH\:mm\:ss.fffZ}" />
			<attribute name="hostname" layout="${hostname}" />
			<attribute name="threadid" layout="${threadid}" />
			<attribute name="threadname" layout="${threadname}" />
			<attribute name="level" layout="${level:upperCase=true}" />
			<attribute name="logger" layout="${logger}" />
			<attribute name="scope" layout="${Scope}" />
			<attribute name="exception" layout="${exception:format=@}" />
		</layout>
	</target>
</targets>
(...)
```

snakefoot:
> As mentioned above, with high log traffic, doing async writes on the socket.. serialized one per one.. with a lock.. causes saturation in the Queue of NetworkRequestArgs (QueueNetworkSender).

Maybe we can change `ProcessNextQueuedItem` so `BeginRequest` is moved outside the lock-scope. Then there should not be any contention around `lock (_pendingRequests)`.

304NotModified:
(we should add N.png to the .gitignore ;))

snakefoot:
> @304NotModified  (we should add N.png to the .gitignore 😉 )

Could also remove the hacky download of NLog.png, and just have it as static resource.

RVINAA:
> (we should add N.png to the .gitignore 😉 )

Fail, only revised the diff instead of status to see; I was trying to solve a problem with compilation issue with defined constansts for test project 😂

Now should work ;)

RVINAA:
My bad, I've seen i forgot to implement the SSL socket connection for this network sender.. a pretty fixme (but for testing right now is not needed i think)

snakefoot:
Thank you so much for highlighting this issue with the NLog NetworkTarget exhausting the memory and stalling the logger-threads. The feedback is highly appreciated, and we should definitely fix this with NLog 5.0.

But I don't think changing to synchronous-socket-send is the right direction. Suddenly there will be a user with a bad network-connection, that is annoyed that too many threadpool-threads are stuck in synchronous-socket-send.

RVINAA:
> Thank you so much for highlighting this issue with the NLog NetworkTarget exhausting the memory and stalling the logger-threads. The feedback is highly appreciated, and we should definitely fix this with NLog 5.0.
> 
> But I don't think changing to synchronous-socket-send is the right direction. Suddenly there will be a user with a bad network-connection, that is annoyed that too many application-threads are stuck in synchronous-socket-send.

Small annotation, there is only one thread per target.. as we don't close it on failure (ie. when endpoint is down.. the thread will start to flush the blocking queue..).

Closing.. I would like to test your networksender in v5 release :)

snakefoot:
> Small annotation, there is only one thread per target.

There will be one thread for each NetworkSender (one for each connection). A single NetworkTarget has the ability to have many NetworkSenders.

snakefoot:
@RVINAA  Released ver. 5.0-preview-2 https://www.nuget.org/packages/NLog/5.0.0-preview.2 (Notice that OnQueueOverflow=Discard by default, and you probably want `Block`)

snakefoot:
NLog 5.0 RTM is now availble: https://www.nuget.org/packages/NLog (Better performance with discard-behavior instead of memory-growth)

PR #4552, opened 2021-09-06
LineEndingMode - Support TypeConverter on NETSTANDARD1_3

snakefoot:
#3469 added support for TypeConverter on NETSTANDARD1_3. Just forgot to enable this one.

### DIFF
--- a/src/NLog/Targets/LineEndingMode.cs
+++ b/src/NLog/Targets/LineEndingMode.cs
@@ -42,9 +42,7 @@ namespace NLog.Targets
     /// <summary>
     /// Line ending mode.
     /// </summary>
-#if !NETSTANDARD1_3
     [TypeConverter(typeof(LineEndingModeConverter))]
-#endif
     public sealed class LineEndingMode : IEquatable<LineEndingMode>
     {
         /// <summary>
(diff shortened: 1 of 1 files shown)


snakefoot:
Also tempted to remove `IConvertible` from the NLog `LogLevel`-type as it works better with Json.Net :)

304NotModified:
> Also tempted to remove `IConvertible` from the NLog `LogLevel`-type as it works better with Json.Net :)

Isn't the `IConvertible` helping the db target? 
I'm trying to determine what the result will be for removing 
 `IConvertible`.

Also any idea about system.text.json

snakefoot:
> Isn't the `IConvertible` helping the db target?

The db-target also supports `TypeConverter` so can still convert LogLevel into string / ordinal on destination-type.

snakefoot:
> Also any idea about system.text.json

`system.text.json` does not support `TypeConverter` or `IConvertible` out of the box, but just serialize all LogLevel-properties and fails to deserialize. Also the problem with IConvertible is that you can only convert in one direction. But you can provide your own System.Json.Net JsonSerializerOptions that includes a converter enables TypeConverter-support (so it can serialize and deserialize).

PR #4563, opened 2021-09-16
Added alias toupper and tolower

snakefoot:
Matching [String.ToUpper](https://docs.microsoft.com/en-us/dotnet/api/system.string.toupper) and [String.ToLower](https://docs.microsoft.com/en-us/dotnet/api/system.string.tolower) that are known in the DotNet-world.


304NotModified:
I guess I would expect just `lower` and `upper` (sql, azure pipelines).

snakefoot:
Well I mostly work with DotNet, and there we have `String.ToUpper` and `String.ToLower`. But just an idea.

PR #4565, opened 2021-09-16
ExceptionLayoutRenderer - Support layout for Separator

snakefoot:
Changed Separator-property from `string` to `Layout`, but added bonus-logic so the property still supports standard string-escape-logic. So it might be confusing for users that these Seperator-properties behaves a little differently from other Layout-properties.

It would ofcourse be nice if one could have both string-escape-logic and Layout-logic together. But because NLog FileTarget was the first real target with its FileName-option as Layout, then the Layout parsing was highly optimized for parsing Windows-File-Paths where single-slash does not mean "escape". But actually it is ONLY when Layout is used for File-Paths that this rules makes sense. I guess with NLog 6.0 then the `FilePathLayout` should become public and should be used for all filepath-related-properties.

### DIFF
--- a/src/NLog/Internal/Reflection/PropertyHelper.cs
+++ b/src/NLog/Internal/Reflection/PropertyHelper.cs
@@ -100,6 +100,11 @@ namespace NLog.Internal
                 throw new NLogConfigurationException($"Unknown property '{propertyName}'='{value}' for '{objType.Name}'");
             }
 
+            SetPropertyFromString(obj, propertyName, value, propInfo, configurationItemFactory);
+        }
+
+        internal static void SetPropertyFromString(object obj, string propertyName, string value, PropertyInfo propInfo, ConfigurationItemFactory configurationItemFactory)
+        {
             try
             {
                 Type propertyType = propInfo.PropertyType;
@@ -108,7 +113,7 @@ namespace NLog.Internal
                 {
                     if (propInfo.IsDefined(_arrayParameterAttribute.GetType(), false))
                     {
-                        throw new NotSupportedException($"Property {propertyName} on {objType.Name} is an array, and cannot be assigned a scalar value: '{value}'.");
+                        throw new NotSupportedException($"Property {propertyName} on {obj.GetType().Name} is an array, and cannot be assigned a scalar value: '{value}'.");
                     }
 
                     propertyType = Nullable.GetUnderlyingType(propertyType) ?? propertyType;
@@ -124,7 +129,7 @@ namespace NLog.Internal
             }
             catch (TargetInvocationException ex)
             {
-                throw new NLogConfigurationException($"Error when setting property '{propInfo.Name}'='{value}' on {objType.Name}", ex.InnerException);
+                throw new NLogConfigurationException($"Error when setting property '{propInfo.Name}'='{value}' on {obj.GetType().Name}", ex.InnerException);
             }
             catch (Exception exception)
             {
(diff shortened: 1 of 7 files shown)


snakefoot:
But also very tempted to allow string-properties to capture the ending `}` after having eaten `${`. Mostly because I have thought for a long time that one could have a string-property with a Layout-backing-field. And this idea works great when doing unit-test and assigning the string-value directly. But it fails hard when the config-parser assigns the partially parsed string-value.

snakefoot:
@304NotModified Any chance of a code-review of this pull-request? (It is pretty much like your own #4539, but with full layout support).

304NotModified:
Sorry, super busy. Will try tomorrow!

304NotModified:
> FileName-option as Layout, then the Layout parsing was highly optimized for parsing Windows-File-Paths where single-slash does not mean "escape". But actually it is ONLY when Layout is used for File-Paths that this rules makes sense. I guess with NLog 6.0 then the `FilePathLayout` should become public and should be used for all filepath-related-properties.

It's not only filepaths, but also newlines for layout renders?
 Anyway, makes sense to split them

304NotModified:
> So it might be confusing for users that these Seperator-properties behaves a little differently from other Layout-properties.

I'm doubting because of this. Maybe another way is to create another Layout type, e.g. CompatibeLayout. Or a new StringLayout. 

Please let me know what you think.

snakefoot:
> Maybe another way is to create another Layout type, e.g. CompatibeLayout. Or a new StringLayout.

Yes also considered introducing `UnicodeLayout` or `StringLayout`. But still I like that the property-type is `Layout` so any layout can be used, so it would only be the "backing-field" that would be different.

The only reason for introducing something like `UnicodeLayout` or `StringLayout` would be to make it easier for 3rd parties to have official Layout-properties that supports correct string-escape-handling (As they also can use it as "backing-field"). But usually 3rd parties just have a string-property, and then just have a normal Layout as "backing-field". This works as intended for 3rd-party NLog Target-options and NLog Layout-options, but just not NLog LayoutRenderer-options.

Again it is close to impossible to make something consistent here, because of the decisions made for FileTarget-FileName-property. And these seperator-properties are a very isolated part of NLog.

I see 2 directions:
1) Keep seperators as string-properties that magically are able to handle Layout, because Layout is used as "backing-field".
2) Change seperators as Layout-properties, but then they will behave differently than other Layout-properties. But at the same time it will be visible that layout-logic is possible.

I guess for consistency then I think that option `1)` is the way to go, so seperator-options remains string-properties but have Layout-backing-field (To signal that they are different from normal Layout-properties that are currently intended for FilePath)

304NotModified:
> The only reason for introducing something like UnicodeLayout or StringLayout would be to make it easier for 3rd parties to have official Layout-properties that supports correct string-escape-handling

Not sure if we are on the same page, but ideally I would propose:

- create a new type LayoutWithEscapes (better name would be great) that has the current Layout behaviour
- change the current Layout that I won't do the special escapes and that is compatible with a string
- change the FileTarget.FileName property (and maybe others) to LayoutWithEscapes 


OK, this a breaking changes, but it's easy to change in your config (replace `//n` to `/n` etc)

snakefoot:
> create a new type LayoutWithEscapes (better name would be great) that has the current Layout behaviour

Think you have confused the logic (or confused me). Layout is currently by default WITHOUT string-escape, because it has to support FilePaths.

Any change to the NLog Layout logic would be a huge breaking change, so not interested in going down that road.

The goal for me is to find the least ugly solution for making ExceptionLayoutRenderer-Seperator-property to accept `${newline}` as input. But while still supporting `\n` and `\t`.

snakefoot:
Replaced by #4583

Issue #147, opened 2012-10-07
Problem with Target FormControl


Xharze:
Reported in the forum at post: http://nlog-project.org/forum.html#nabble-tc1685764|a1685764

Hi,

 i have a little problem with the target FormControl. I wrote a NLog.config with 3 targets ( File, Debugger, FormControl ). The logger writes successfull all messages to the targets File and Debugger, the FormControl target is ignored by the logger. I attached my config, and the internal Nlog log.

The FormControl is a ToolStripStatusLabel Control.

 I hope somebody have an idea, thx.

 NLog.config

nlog.log

304NotModified:
Without more info why can't resolve this. Closing For now.

Fr33dan:
I wanted a logging output to show my most recent error in the status strip of my UI and it didn't work. I experienced this "bug" and was going to attempt to reactivate this/file a new one for the same thing. Although it may belong in the [NLog.Windows.Forms](https://github.com/NLog/NLog.Windows.Forms) project. In doing my due diligence to have any the information needed to diagnose it, I diagnosed it.

Turning on a higher internal logging level I can see it can't find my control on my form:
`2021-09-16 10:36:43.6446 Info Control programStatusLabel on Form ProgrammerForm not found`

I switched from a ToolStripStatusLabel to a regular Label on my form and it worked fine. I suspected this is related to the fact that ToolStripStatusLabel controls are actually parented to a StatusStrip rather than the form directly. In an effort to check I tried a ToolStripMenuItem and ToolStripTextBox which are similarly parented to a MenuStrip. All resulted in the same error.

So I did some investigating and the problem is that Both StatusStrip and MenuStrip extend ToolStrip which managed their children as Items instead of Controls. ToolStripStatusLabel isn't even an extension of Control.

Thus I ended up creating a target that I that could fill a ToolStripStatusLabel, resulting in pull request #4566 and pull request #87 on the NLog.Windows.Forms project.

Issue #87, opened 2012-10-07
EventLog target Throws FormatException @ ..Write(LogEventInfo logEvent)


ghost:
<p>It loots like it&#39;s trying to convert category &quot;INFO&quot; to a short. Not sure if I&#39;m doing something wrong. I&#39;ve attached the stack trace and the target configuration in the file. The code I&#39;m using to reproduce the issue is as follows:</p>

<p>&nbsp;</p>

<p>LogManager.GetLogger(&quot;Service&quot;).Log(LogLevel.Info, &quot;Starting service...&quot;);</p>

<p>&nbsp;</p>

<p>*** As I was writing this out, it kind of clicked that it is expecting category to be a number, however when I look for information on the layout for eventLog category target on the wiki I&#39;m not getting anything. Is there a way to format the loglevel as an short?</p>

Moved from http://nlog.codeplex.com/workitem/6571

PR #4566, opened 2021-09-16
Add example configuration for ToolStripItemTarget

Fr33dan:
After encountering the "bug" #147 . I implemented a target capable of logging to a ToolStripItem. That is in the NLog.Windows.Forms project under pull request [#87](https://github.com/NLog/NLog.Windows.Forms/pull/87). This pull incorporates the documentation for said target.


snakefoot:
Thank you for the pull-request. Think there is a pending cleanup of the example projects, that no longer directly belongs to the NLog-core-project (Like `NLog.Windows.Forms`)

Think it would makes most sense to have the example-project in the `NLog.Windows.Forms`-repository. Maybe even part of the solution, so it is easy to see it in action.

One could consider moving the existing [FormControl](https://github.com/NLog/NLog/tree/dev/examples/targets/Configuration%20API/FormControl)-sample together into the `NLog.Windows.Forms`-repository aswell.

Fr33dan:
That is fair, I mainly just copied what was already there for existing targets. Mainly I have never used inline XML documentation with linkage like FormControlTarget and figured if I copied from that it would work as intended. I can move example(s), but I have no idea what effect it will have wherever the XML documentation is used.

I figure its probably most important to have it documented in the wiki but it doesn't make sense to try to edit that until the new target is approved.

snakefoot:
Yes also think wiki-page-documenation is most important. Think the API-docs should just pickup a wiki-link, instead of having these example projects referenced directly. Instead the wiki-page should do the nice introduction, and if very needed then it can decide to throw a repository-link to an example project.

304NotModified:
@Fr33danis this PR still needed? I will try to release NLog.Windows.Forms  4.6 soon

Fr33dan:
As that I wrote the code for the ToolStripItem I should also write the documentation but where I should do so is a question for y'all as package maintainers. Based on this discussion I did write a page in the wiki. 

I think snakeroot is probably right in that it doesn't really belong here but I figure that is a call for you two. If you agree this PR can be closed.

snakefoot:
Thank you for creating the wiki-page:

https://github.com/NLog/NLog/wiki/ToolstripmenuItem-target

I was thinking example-projects in the solution, like done here:

https://github.com/NLog/NLog.Web/tree/master/examples

I will close this pull-request, and then one can consider creating pull-request for the `NLog.Windows.Forms`-project.

PR #4580, opened 2021-09-23
LayoutRenderer and Layout are now threadsafe by default

### LINKED ISSUE #4577: Obsolete ThreadSafe attribute and open for full concurrency
NLog 6.0 should open the gates, and should never have to worry about LayoutRenderers not being threadsafe:

https://github.com/NLog/NLog/blob/8a004e08fc36fda5d17c7a4a6e19b651c5ce7665/src/NLog/Config/ThreadSafeAttribute.cs#L45

Instead it should just be part of the release-notes that custom layoutrenderers might produce unpredictable results, and should be checked before updating to NLog 6.0

snakefoot:
Resolves #4577 so NLog becomes more predictable in its behavior (Original introduced with #2653 in NLog 4.5.3)


src/NLog/Layouts/Log4JXmlEventLayout.cs
```diff
@@ -48,8 +48,6 @@ namespace NLog.Layouts
     /// </remarks>
     [Layout("Log4JXmlEventLayout")]
     [ThreadAgnostic]
-    [ThreadSafe]
-    [AppDomainFixedOutput]
```
304NotModified, line 52: Is the removal of `[AppDomainFixedOutput]` on purpose? 

(looks like a copy-paste...euh..remove error)

304NotModified (commented):
I think this is a great improvement, especially for 3rd party layout renders

one double check question

src/NLog/Layouts/Log4JXmlEventLayout.cs
snakefoot: It is on purpose, because it makes no sense to have `[AppDomainFixedOutput]` for `Log4JXmlEventLayout`. As it would mean static output always.

304NotModified: ok!

snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/How-to-write-a-custom-layout-renderer

PR #4583, opened 2021-09-26
ExceptionLayoutRenderer - Separator with basic Layout support

### LINKED ISSUE #2249: ${exception} - separator=${newline} won't work

snakefoot:
Resolves #2249, so it is possible for users to specify `${newline}` in the NLog.config for the Separator-property.

Also changes property-parsing so it work for LayoutRenderer-string-options, so it is consistent with NLog-Target-options and NLog-Layout-options.

`ExceptionLayoutRenderer` is very special since it allows users to override the rendering-methods for the individual parts. So the backing-field remains a string to ensure predictable behavior for the users, that wants to override rendering-methods.

### DIFF
--- a/src/NLog/Internal/Reflection/PropertyHelper.cs
+++ b/src/NLog/Internal/Reflection/PropertyHelper.cs
@@ -100,6 +100,11 @@ namespace NLog.Internal
                 throw new NLogConfigurationException($"Unknown property '{propertyName}'='{value}' for '{objType.Name}'");
             }
 
+            SetPropertyFromString(obj, propertyName, value, propInfo, configurationItemFactory);
+        }
+
+        internal static void SetPropertyFromString(object obj, string propertyName, string value, PropertyInfo propInfo, ConfigurationItemFactory configurationItemFactory)
+        {
             try
             {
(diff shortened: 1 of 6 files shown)


snakefoot:
@304NotModified Unless you are completely against this, then I would like commit and proceed with adding basic layout support for the other seperators. So NLog 5.0-preview2 can get out.

304NotModified:
> with adding basic layout support for the other seperators

For the exception layout renderer? That sounds fine to me.

304NotModified (approved):
This PR is a bit hard too review (it's large), but I think its fine

snakefoot:
> For the exception layout renderer? That sounds fine to me.

Was thinking to also fix:

- StackTraceLayoutRenderer.Separator
- ScopeContextNestedStatesLayoutRenderer.Separator

304NotModified:
Ah OK, good idea :)

PR #4584, opened 2021-09-26
Layout - Render with StringBuilder target made public

snakefoot:
Allow 3rd party layouts to render complex output using sub-layouts with less memory-allocation of strings.

And now NLog `Layout.Precalculate` will call `RenderFormattedMessage` for 3rd party layouts, with help from NLog internal StringBuilder-pool. Before `Layout.Precalculate` would always call `GetFormattedMessage` directly.


snakefoot:
@RVINAA This should also make it easier for you to create a high-performance NLog-target without needing to perform method-reflection. As you are currently doing: https://github.com/evicertia/NLog.FastNetworkTarget

PR #4587, opened 2021-10-02
Enable LayoutRenderer to ResolveService during intialization

snakefoot:
Also avoid throwing exceptions during initialization, when dependency is available from external service-provider (NLog should attempt to be silent partner).

### DIFF
--- a/src/NLog/Config/ServiceRepository.cs
+++ b/src/NLog/Config/ServiceRepository.cs
@@ -53,7 +53,7 @@ namespace NLog.Config
         /// <remarks>Avoid calling this while handling a LogEvent, since random deadlocks can occur.</remarks>
         public abstract object GetService(Type serviceType);
 
-        internal abstract object TryGetService(Type serviceType);
+        internal abstract bool TryGetService<T>(out T serviceInstance) where T : class;
 
         internal abstract ConfigurationItemCreator ConfigurationItemCreator { get; set; }
 
--- a/src/NLog/Config/ServiceRepository.cs
+++ b/src/NLog/Config/ServiceRepository.cs
@@ -53,6 +53,8 @@ namespace NLog.Config
         /// <remarks>Avoid calling this while handling a LogEvent, since random deadlocks can occur.</remarks>
         public abstract object GetService(Type serviceType);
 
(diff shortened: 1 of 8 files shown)


src/NLog/Config/ServiceRepository.cs
```diff
@@ -53,6 +53,8 @@ public abstract class ServiceRepository : IServiceProvider
         /// <remarks>Avoid calling this while handling a LogEvent, since random deadlocks can occur.</remarks>
         public abstract object GetService(Type serviceType);
 
+        internal abstract object TryGetService(Type serviceType);
```
304NotModified, line 56: With A Try, I expected a bool resturn and a out param.

Proposal: rename to FindService

304NotModified: it's a common naming convention for me (find returns `null` when it's not there, get thows an exception for that case).

 See also https://softwareengineering.stackexchange.com/a/182120/16953

src/NLog/Config/ServiceRepositoryExtensions.cs
```diff
@@ -47,30 +47,50 @@ internal static ServiceRepository GetServiceProvider([CanBeNull] this LoggingCon
...
-            catch (NLogDependencyResolveException)
+            else
             {
-                if (ignoreExternalProvider)
-                    throw;
+                IServiceProvider externalServiceProvider;
+
+                try
+                {
+                    var service = serviceProvider.TryGetService(typeof(T)) as T;
```
304NotModified, line 60: generic method next, same as GetService, would be nice. 

(consistent API)

304NotModified (commented):
Looks good, some small notes

src/NLog/Config/ServiceRepository.cs
snakefoot: Yes thank you for spotting. Fixed TryGetService so it returns true / false

src/NLog/Config/ServiceRepositoryExtensions.cs
snakefoot: Yes thank you for spotting. Fixed TryGetService to be generic method.

Issue #4608, opened 2021-10-13
FileLoadException: The process cannot access the file because it is being used by another process


rgroenewoudt:
I am are running into this exception on production environment (~23 times in 3 months) but I cannot reproduce it.
I suspect that it might be anti-virus software related as we are running Windows Defender (Windows Server 2019) and SentinelOne software. In Windows Defender the C:\logs directory is already excluded.

```
System.IO.FileLoadException: The process cannot access the file because it is being used by another process. (Exception from HRESULT: 0x80070020)
   at System.Runtime.InteropServices.Marshal.ThrowExceptionForHRInternal(Int32 errorCode, IntPtr errorInfo)
   at NLog.Internal.FileAppenders.BaseFileAppender.WindowsCreateFile(String fileName, Boolean allowFileSharedWriting, Int32 overrideBufferSize)
   at NLog.Internal.FileAppenders.BaseFileAppender.TryCreateFileStream(Boolean allowFileSharedWriting, Int32 overrideBufferSize)
   at NLog.Internal.FileAppenders.BaseFileAppender.CreateFileStream(Boolean allowFileSharedWriting, Int32 overrideBufferSize)
   at NLog.Internal.FileAppenders.RetryingMultiProcessFileAppender.Write(Byte[] bytes, Int32 offset, Int32 count)
   at NLog.Targets.FileTarget.WriteToFile(String fileName, ArraySegment`1 bytes, Boolean initializedNewFile)
   at NLog.Targets.FileTarget.ProcessLogEvent(LogEventInfo logEvent, String fileName, ArraySegment`1 bytesToWrite)
   at NLog.Targets.FileTarget.Write(LogEventInfo logEvent)
   at NLog.Targets.Target.Write(AsyncLogEventInfo logEvent)
   at NLog.Targets.Target.WriteAsyncThreadSafe(AsyncLogEventInfo logEvent)
   at NLog.Targets.Target.WriteAsyncLogEvent(AsyncLogEventInfo logEvent)
   at NLog.LoggerImpl.WriteToTargetWithFilterChain(Target target, FilterResult result, LogEventInfo logEvent, AsyncContinuation onException)
   at NLog.LoggerImpl.Write(Type loggerType, TargetWithFilterChain targetsForLevel, LogEventInfo logEvent, LogFactory factory)
   at NLog.Logger.Info(String message, Int32 argument)
```

2 processes, one Windows Service, other IIS application pool are writing to the same logfile.

**NLog version**: 4.7.6
[...]

PR #4623, opened 2021-10-23
FileTarget - KeepFileOpen = true by default to avoid loosing file handle

### LINKED ISSUE #2758: FileTarget KeepFileOpen = true by default
This is for NLog 6.0. Instead of having FileTarget KeepFileOpen = false, then it should be true.

Also FileTarget ConcurrentWrites should only be true by default when able to use WindowsFileAppender or UnixFileAppender (Or if having configured KeepFileOpen = false).

snakefoot:
Resolves #2758. See also #4608

Applications will now fail with `System.IO.IOException: The process cannot access the file` when using `File.OpenText` or `File.ReadAllText` to read files created by NLog (Unless having explicitly configured `KeepFileOpen="false"`). But at the same time it will avoid failing application-logging caused by external application wrongly trying to use exclusive file-locking

### DIFF
--- a/src/NLog/Internal/ExceptionHelper.cs
+++ b/src/NLog/Internal/ExceptionHelper.cs
@@ -123,12 +123,12 @@ namespace NLog.Internal
 #if !NETSTANDARD1_3 && !NETSTANDARD1_5
             if (exception is StackOverflowException)
             {
-                return true;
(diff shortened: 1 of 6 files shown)


snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/File-target about KeepFileOpen = true by default

PR #4624, opened 2021-10-23
FileTarget - Only enable FileSystemWatcher when ConcurrentWrites = true

snakefoot:
Only enable Win32-API-call when overriding FileAttributes. Refactor MultiFileWatcher to better handle medium trust.

Better chance of NLog FileTarget working on low-permission-platforms (UWP / Xamarin Android / Xamarin iOS). But it will take longer time (up to 1 sec) for NLog to recover from the user deleting the active file (before 100 ms)


snakefoot:
Updated Wiki:  https://github.com/NLog/NLog/wiki/File-target (For enableFileDelete about FileSystem watchers being used when ConcurrentWrites=true)

PR #4626, opened 2021-10-24
PropertiesDictionary - Fixed threading issue in EventProperties

snakefoot:
The property allowed multiple threads to be reading and modifying the same collection. Maybe fixing #4496

The issue can be triggered by using structured-logging with message-templates, and having multiple targets doing logevent-property-lookup and also enumerating the logevent-properties.

If 2 targets has layouts where target1 uses `${event-properties}` and target2 uses `JsonLayout.IncludeAllProperties`. Then the following can happen:
- Target1 wants to lookup property, and starts building the lookup-dictionary.
- Target2 wants to enumerate properties, and sees lookup-dictionary is available and runs with it.
- Target1 continues to build the lookup-dictionary.
- Target2 enumerates the lookup-dictionary, and might get collection was modified.

Have not been able to implement an unit-test that reproduced the issue, but was able to get `Dictionary.Add` to fail with key already exists by introducing "artificial sleep" with the debugger (Two threads were working on building the same lookup-dictionary)

The issue has been there since NLog ver. 4.5 (4 years ago)

### DIFF
--- a/src/NLog/Internal/Collections/PropertiesDictionary.cs
+++ b/src/NLog/Internal/Collections/PropertiesDictionary.cs
@@ -103,18 +103,7 @@ namespace NLog.Internal
             {
                 if (_eventProperties is null)
                 {
-                    if (_messageProperties != null && _messageProperties.Count > 0)
-                    {
-                        _eventProperties = new Dictionary<object, PropertyValue>(_messageProperties.Count);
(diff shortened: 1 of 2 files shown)


snakefoot:
@304NotModified Thinking that I would like to have this bug-fix merged to master, and make a NLog ver. 4.7.12 release. Would that be ok with you?

PR #4633, opened 2021-10-25
PropertiesDictionary - Added support for case-insensitive property lookup

### LINKED ISSUE #4507: ${event-properties} should support case-insensitve lookup
This should be supported:
```c#
logger.Info("Hello {userName}", "Bob");
```
Where the following produces the output:
```xml
${event-properties:item=UserName}
```

snakefoot:
Resolves #4507


snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/EventProperties-Layout-Renderer with **IgnoreCase**-parameter.

Issue #4643, opened 2021-10-28
ScopeContext - Allow child task to propagate context properties to parent


snakefoot:
Maybe something like this, where it possible to do this:

```c#
using var propertyCollector = ScopeContext.PushPropertyCollector();
ScopeContext.CollectProperty("parentproperty", "value");
ScopeContext.CollectProperty("sharedproperty", "value");
await ChildTaskAsync();   // Pushes properties to ScopeContext
Logger.Info("Includes properties from child task");

async void ChildTaskAsync()
{
     ScopeContext.CollectProperty("childproperty", "value");
     await Task.Delay(1);
     ScopeContext.CollectProperty("sharedproperty", "value2");   // overwrites
}
```

The properties included in the property-collector should be accessible when using `${scopeproperty}` or `IncludeScopeProperties = true`.

Guess one could manually inject a ConcurrentDictionary as scope-property at the  parent-task-scopecontext, and then child-tasks can resolve that scope-property and insert their own propagating-properties into the ConcurrentDictionary. Thus parent-task-scopecontext will see that the scope-property with ConcurrentDictionary has the propagating-values from the child-tasks.

marcanpilami:
OK, thanks for the input. I had indeed not seen all the changes in version 5 and would have missed why you were taking about `ScopeContext`. I submit something this week I hope.

snakefoot:
Maybe the alternative solution is going for Activity.Current and its Baggage-Dictionary:
- https://www.nuget.org/packages/NLog.DiagnosticSource#readme-body-tab

PR #4652, opened 2021-11-06
#4643 - Added scoped async property collector.

marcanpilami:
For #4643, using the syntax initially proposed.

Since the existing contexts were immutable with automatic snapshots and we want here to create a mutable shared context with manual snapshots there is sadly very little code reuse here - obviously open to suggestions on how to reduce this.

A test seemingly unrelated fails. Added a test for the new context.

### DIFF
--- a/src/NLog/Contexts/ScopeContext.cs
+++ b/src/NLog/Contexts/ScopeContext.cs
@@ -104,7 +104,7 @@ namespace NLog
         {
 #if !NET45
             var parent = GetAsyncLocalContext();
-            
+
             var allProperties = ScopeContextPropertiesCollapsed.BuildCollapsedDictionary(parent, properties.Count);
             if (allProperties != null)
             {
@@ -207,6 +207,7 @@ namespace NLog
         {
 #if !NET35 && !NET40 && !NET45
             SetAsyncLocalContext(null);
+            ScopeContextAsyncPropertyCollector.Clear();
 #else
             ClearMappedContextCallContext();
             ClearNestedContextCallContext();
@@ -221,7 +222,8 @@ namespace NLog
         {
 #if !NET35 && !NET40 && !NET45
             var contextState = GetAsyncLocalContext();
-            return contextState?.CaptureContextProperties(0, out var _) ?? ArrayHelper.Empty<KeyValuePair<string, object>>();
+            var res = contextState?.CaptureContextProperties(0, out var _) ?? ArrayHelper.Empty<KeyValuePair<string, object>>();
+            return System.Linq.Enumerable.Union(res, ScopeContextAsyncPropertyCollector.GetAllProperties());
 #else
             var mappedContext = GetMappedContextCallContext();
             if (mappedContext?.Count > 0)
@@ -258,13 +260,13 @@ namespace NLog
             if (contextState != null)
             {
                 var mappedContext = contextState?.CaptureContextProperties(0, out var _);
-                if (mappedContext != null)
+                if (mappedContext != null && TryLookupProperty(mappedContext, key, out value))
                 {
-                    return TryLookupProperty(mappedContext, key, out value);
+                    return true;
                 }
             }
-            value = null;
-            return false;
+            // Property collectors - being less specific - have lower priority than other contexts.
+            return ScopeContextAsyncPropertyCollector.TryGetProperty(key, out value);
 #else
             var mappedContext = GetMappedContextCallContext();
             if (mappedContext != null && mappedContext.TryGetValue(key, out value))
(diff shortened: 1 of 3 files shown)


snakefoot:
Thank you for the pull-request and the unit-test.

Have been thinking whether the collector should behaves like an observer, so the normal scope-context-operations are just applied to the normal scope-context, but also on the scope-property-collector.

With the speciality that the scope-property-collector has its own "lifetime", so one can do this:

```c#
using var collector = ScopeContext.PushPropertyCollector();
using var parentProperty = ScopeContext.PushProperty("name", "bob");
await System.Threading.Tasks.Task.Run(() =>
{
     using var childProperty = ScopeContext.PushProperty("name", "alice"); // Also updates the active property-collector
}
ScopeContext.TryGetProperty("name", out var myNameIsAlice); // Will lookup in property-collector if available, with fallback to normal scope-context-lookup
```

Thus removing the need for the `CollectProperty`-method.

marcanpilami:
So basically the `ScopeContext` would be modal - either in `property collector` mode or in `normal` mode.

But there is a strong difference between the two modes when dealing with the 'using' pattern.

```
using (ScopeContext.PushPropertyCollector())
{
    using (ScopeContext.PushProperty("name", "bob"))
    {
        // Whatever
    }
    ScopeContext.TryGetProperty("name", out var myNameIsStillBob); // oops
}
```
In this case we would 'naturally' expect name to only exist inside the using block, but as we are in `property collector` mode it is not true. Would that be acceptable behaviour?

snakefoot:
> Would that be acceptable behaviour?

That is really up to you, since you are the main-consumer. I was just thinking one could create the property-collector inside ASP.NET-middleware at request-startup, and then on request-completion allow logging of all the collected properties (Ex. RequestId etc.)

I think it would be expected that when the parent-task dispses its property-collector, then it should change all child-task-instances that has inherited the property-collector. Thus one can have child-tasks with a "disabled" property-collector-asynclocal-state, because the parent has disposed the collector.

To reduce the side-effects on child-tasks if parent creates property-collector, then I guess one could also do this:

```c#
using var parentProperty = ScopeContext.PushProperty("name", "bob");
using var collector = ScopeContext.PushPropertyCollector(); // Will create new container with all existing scope-context-properties (includes `bob`)
await System.Threading.Tasks.Task.Run(() =>
{
     using var childProperty = ScopeContext.PushProperty("name", "alice"); // Also updates the active property-collector
}
ScopeContext.TryGetProperty("name", out var myNameIsAlice); // Will find "bob" (Will not use property-collector)
collector.PushCollectedProperties();  // Will push the collected properties to the scope-context
ScopeContext.PushCollectedProperties(collector);  // Will push the collected to the scope-context (alternative API where collector implements IDictionary<string, object>
ScopeContext.TryGetProperty("name", out var myNameIsAlice); // Will find "alice"
```

snakefoot:
After a little more thought, then I think this will work for my idea for ASP.NET Core Middleware:

```c#
using var parentProperty = ScopeContext.PushProperty("name", "bob");
// When collector is disposed, then any collected properties is pushed into the ScopeContext
using (var collector = ScopeContext.PushPropertyCollector())
{
     await System.Threading.Tasks.Task.Run(() =>
     {
          using var childProperty = ScopeContext.PushProperty("name", "alice"); // Also updates the active property-collector
     }

     ScopeContext.TryGetProperty("name", out var myNameIsBob); // Will find "bob" (Will not use property-collector)
}

ScopeContext.TryGetProperty("name", out var myNameIsAlice); // Will find "alice" (Property-collector has been applied)
```

Would this behavior also work for you?

marcanpilami:
I agree my use case better fits inside a middleware or filter. The current PR code is already compatible with this.

Your latest proposals would however not answer the use-case: I want the value to bubble-up from async contexts and be picked naturally when logging. So the TryGetProperty (being in this test the emulation of a log message) just outside the async block should give the value set in the task, therefore alice not bob.

My question about acceptable behaviour was not about answering to my use-case as it is already OK, it's more about creating an API which is weird - a `using`  pattern should clean up behind it, that's the justification of its existence. 

And whatever we choose, I believe it should remain simple - if we have a hard time explaining what we do to ourselves, I cannot begin to imagine the question tickets/bugs which may come later.
So we either 
* have two sets of APIs (current code - one API for scope, one for setting values) and values are simply merged on consumption.
* decide clear rules of priority and inheritance, like you proposed with the modal behaviour with collector > normal. Another way to do it would be to have the `PushPropertyCollector` API to simply block the creation of new normal `IScopeContextAsyncState` and force reuse of the top one when creating new values. As there would truly be only one stack of states, I believe it would be cleanest API. BUT: this would mean transforming a lot of code, not the least making everything mutable. When you have actually done a lot of efforts to make them immutable...

snakefoot:
My use-pattern would be like this:

```c#
        /// <summary>
        /// Executes the middleware.
        /// </summary>
        public async Task Invoke(HttpContext httpContext)
        {
            try
            {
                using (var collector = ScopeContext.PushPropertyCollector())
                    await _next(httpContext);
                LogHttpRequest(httpContext, null);  // Will see all properties collected
            }
            catch (Exception exception) when ()
            {
                LogHttpRequest(httpContext, exception);  // Will see all properties collected
            }
        }
```

It is fine for me that only parent-task sees the collected properties. If child-task also wants to collect properties, then it can also just use `PushPropertyCollector()`

snakefoot:
Guess one could add a parameter to ´PushPropertyCollector` that specifies whether properties are only applied to parent-scope on collector-exit, or whether collected-scope-properties should be leaky and sibling-tasks should be affecting each other (dirty read).

snakefoot:
Closing due to inactivity.

PR #4655, opened 2021-11-09
JsonLayout - ExcludeProperties should also handle IncludeScopeProperties

### LINKED ISSUE #4287: JsonLayout excludeProperties does not work with includeMdlc
Hi,

When I'm using mdlc with a layout and trying to exclude this property from being added twice - It does not work and the
 excludeProperties of the eventProperties is not working.

for example:
 {
"name": "using-mdlc"
"layout": "${mdlc:hi:dusty=true}"
},
{
"name": eventProperties"
"layout": { "excludeProperties": dusty }
}

I will still see the property twice - one with the layout and another one without it.

example:
image

snakefoot:
To match #4642 and resolves #4287


304NotModified:
Close/reopen to trigger CI (Azure Devops)

PR #4695, opened 2021-12-03
Ignore dash (-) when parsing (e.g. layouts, layout renderers, targets)

304NotModified:
### Dashes (-) will be ignored when parsing names

* **Impact:** Dashes in names in targets, layout renderers, layouts, filters are ignored. For example: `${aspnet-sessionid}` could be written als `${aspnetsessionid}`, `${aspnet-session-id}` etc and `ColoredConsole` could be written als `Colored-Console`

* **Reason:** It's hard to maintain consistency between the names. Also, when registering custom items, it is confusing to have names with only dashes as difference. For example: `${activityid}` and `${activity-id}`

- [x] add extra test


304NotModified:
Hi

snakefoot:
Make sure not to break support for fully qualified type with assembly-name. See also #4308

Maybe rename one of the unit-test-assemblies to have "-" in its name to verify that it continues to work.

304NotModified:
> Make sure not to break support for fully qualified type with assembly-name. See also #4308

I think it works, as I don't normalize the assemblyName:

```
     private void RegisterDefinition(string itemName, Type itemDefinition, string assemblyName, string itemNamePrefix)
        {
            GetTypeDelegate typeLookup = () => itemDefinition;
            itemName = NormalizeName(itemName);
            _items[itemNamePrefix + itemName] = typeLookup;
            if (!string.IsNullOrEmpty(assemblyName))
            {
                _items[itemName + ", " + assemblyName] = typeLookup;
                _items[itemDefinition.Name + ", " + assemblyName] = typeLookup;
                _items[itemDefinition.ToString() + ", " + assemblyName] = typeLookup;
            }
        }
```

But will try to test it

304NotModified:
pushed! Green!

snakefoot:
Can this test be added to `ExtensionTests.cs`: 

```c#
        [Fact]
        public void FullyQualifiedExtensionTest()
        {
            // Arrange + Act
            var logFactory = new LogFactory().Setup().LoadConfigurationFromXml(@"<nlog throwConfigExceptions='true'>
                <targets>
                    <target name='t' type='Manually-Loaded-Target, Manually-Loaded-Extension' />
                </targets>
            </nlog>").LogFactory;

            // Assert
            Assert.NotNull(logFactory.Configuration.FindTargetByName("t"));
        }
```

src/NLog/Config/Factory.cs
```diff
@@ -246,7 +251,7 @@ public virtual TBaseType CreateInstance(string itemName)
                 message += ". Extension NLog.Web not included?";
 #endif
             }
-            else if (itemName?.StartsWith("database", StringComparison.OrdinalIgnoreCase)==true)
+            else if (itemName?.StartsWith("database", StringComparison.OrdinalIgnoreCase) == true)
             {
                 //common mistake and probably missing NLog.Web
```
snakefoot, line 256: See some incorrect, copy-paste-comments. Maybe remove those?

304NotModified: Maybe on a new PR? This hasn't changed here?

304NotModified: all aren't that needed anymore. So removed them :)

304NotModified:
> Can this test be added to `ExtensionTests.cs`:
> 
> ```cs
>         [Fact]
>         public void FullyQualifiedExtensionTest()
>         {
>             // Arrange + Act
>             var logFactory = new LogFactory().Setup().LoadConfigurationFromXml(@"<nlog throwConfigExceptions='true'>
>                 <targets>
>                     <target name='t' type='Manually-Loaded-Target, Manually-Loaded-Extension' />
>                 </targets>
>             </nlog>").LogFactory;
> 
>             // Assert
>             Assert.NotNull(logFactory.Configuration.FindTargetByName("t"));
>         }
> ```

Ah missed that comment before

304NotModified:
Added and fixed the code. Refactored the normalize and added some unit tests

src/NLog/Internal/NameNormalizer.cs
```diff
@@ -0,0 +1,63 @@
...
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) 
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF 
+// THE POSSIBILITY OF SUCH DAMAGE.
+// 
+
+namespace NLog.Internal
+{
+    internal static class NameNormalizer
+    {
+        internal static string NormalizeName(string itemName)
```
snakefoot, line 38: Would prefer that internal helper methods, that are only used at one place, stays at that place (instead of being extracted into isolated source-file). But I guess you have your reasons.

src/NLog/LayoutRenderers/Contexts/TraceActivityIdLayoutRenderer.cs
```diff
@@ -46,7 +46,6 @@ namespace NLog.LayoutRenderers
     /// A renderer that puts into log a System.Diagnostics trace correlation id.
     /// </summary>
     [LayoutRenderer("activity-id")]
-    [LayoutRenderer("activityid")]
```
snakefoot, line 49: I prefer to keep `activityid` as the default-name.

src/NLog/Internal/NameNormalizer.cs
304NotModified: The reason is unit testing. A static method inside a generic class is a bit hard to test.

snakefoot: Yes I see, guess that resolves it.

src/NLog/LayoutRenderers/Contexts/TraceActivityIdLayoutRenderer.cs
snakefoot: But I also recognize that it overlapse with https://github.com/NLog/NLog.DiagnosticSource

src/NLog/Internal/NameNormalizer.cs
snakefoot: After further considering, then I guess it would be better with unit-testing of the actual factory-method. Instead of unit-testing an internal-method.

304NotModified: That isn't an unit test but a component test? (Slower, more test code and more maintenance)

snakefoot: For NLog is a component, and a tester doing black-box testing of the NLog.dll would be a component-test. I like unit-test to verify an entire contract (here create an object based on type-name),

src/NLog/LayoutRenderers/Contexts/TraceActivityIdLayoutRenderer.cs
304NotModified: it's called `${activity}` there?

src/NLog/Internal/NameNormalizer.cs
304NotModified: I'm following this: https://martinfowler.com/articles/practical-test-pyramid.html

But I could move the static method I guess

src/NLog/LayoutRenderers/Contexts/TraceActivityIdLayoutRenderer.cs
snakefoot: Yes you are right, then no issue. Guess I was forseen when creating NLog.DiagnosticSource :)

src/NLog/Internal/NameNormalizer.cs
304NotModified: move the method

src/NLog/LayoutRenderers/Contexts/TraceActivityIdLayoutRenderer.cs
304NotModified: (It would be hard if we ${activityid} and ${activity-id}. The difference is hard to communicate when speaking to others :P)

304NotModified: Fixed this. I consistently used the ones without the dash, but move now to the version from https://nlog-project.org/config/?tab=layout-renderers

src/NLog/Internal/NameNormalizer.cs
snakefoot: If speed is your concern that I think it is just as fast to test the factory directly, as to testing the internal-method. And think the correct way is to test the factory-methods, so it is ensured that the factory methods is using the internal-method correctly.

snakefoot (approved):
Looks good. Would have preferred unit-testing of factory-methods, instead of just internal method.

304NotModified:
Oops this one is missing in the news post. See opening message.

PR #4717, opened 2021-12-22
databasetarget: improved parsing of dbtype

Orace:
Reduce the Cognitive Complexity of the `DatabaseParameterInfo.TryParseDbType` method.

Ref: [this sonarcloud issue](https://sonarcloud.io/project/issues?branch=dev&id=nlog2&open=AXokBhP1LhB_xjiWet68).

### DIFF
--- a/src/NLog.Database/DatabaseParameterInfo.cs
+++ b/src/NLog.Database/DatabaseParameterInfo.cs
@@ -49,17 +49,21 @@ namespace NLog.Targets
     [NLogConfigurationItem]
     public class DatabaseParameterInfo
     {
+        /// <summary>
+        /// Mapping of System.Data.DbType, System.Data.SqlDbType and NpgsqlTypes.NpgsqlDbType members names
+        /// to their corresponding CLR types.
+        /// </summary>
         private static readonly Dictionary<string, Type> _typesByDbTypeName =
             new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase)
             {
-                { "Bool", typeof(bool) },
-                { "Timestamp", typeof(DateTime) }, // Npgsql ?
+                // Members from System.Data.DbType
+
                 { nameof(System.Data.DbType.AnsiString), typeof(string) },
-                { nameof(System.Data.DbType.Binary), typeof(byte[]) },
+                // { nameof(System.Data.DbType.Binary), typeof(byte[]) },  // not supported
                 { nameof(System.Data.DbType.Byte), typeof(byte) },
                 { nameof(System.Data.DbType.Boolean), typeof(bool) },
                 { nameof(System.Data.DbType.Currency), typeof(decimal) },
-                { nameof(System.Data.DbType.Date), typeof(DateTime) },
+                { nameof(System.Data.DbType.Date), typeof(DateTime) }, // DateOnly when .net framework will be deprecated
                 { nameof(System.Data.DbType.DateTime), typeof(DateTime) },
                 { nameof(System.Data.DbType.Decimal), typeof(decimal) },
                 { nameof(System.Data.DbType.Double), typeof(double) },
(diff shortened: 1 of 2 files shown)


src/NLog.Database/DatabaseParameterInfo.cs
```diff
@@ -203,61 +238,15 @@ internal bool SetDbType(IDbDataParameter dbParameter)
...
                 return null;
-            if (dbTypeString.IndexOf("Bool", StringComparison.OrdinalIgnoreCase) >= 0)
-                return typeof(bool);
-            if (dbTypeString.IndexOf("Timestamp", StringComparison.OrdinalIgnoreCase) >= 0)
-                return typeof(DateTime);
-            if (dbTypeString.IndexOf(nameof(System.Data.DbType.DateTimeOffset), StringComparison.OrdinalIgnoreCase) >= 0)
-                return typeof(DateTimeOffset);
-            if (dbTypeString.IndexOf(nameof(System.Data.DbType.DateTime), StringComparison.OrdinalIgnoreCase) >= 0)
-                return typeof(DateTime);
-            if (dbTypeString.IndexOf(nameof(System.Data.DbType.Date), StringComparison.OrdinalIgnoreCase) >= 0)
```
snakefoot, line 221: Notice how this matches, when type-string actual only contains the word "Date". This is no longer the case.

Orace: Indeed.

It was to handle (at least) the `DbType.DateTime2` case; it is now an entry of `_typesByDbTypeName`.  
The same apply for `String` variations (ie: `DbType.AnsiString`, `DbType.AnsiStringFixedLength` and `DbType.StringFixedLength`).

To ensure full coverage, it is necessary to know all possible values of `dbTypeString` that need to be handled. 
[The documentation](https://github.com/NLog/NLog/wiki/Database-target#sql-statement) refers to _members_ of the [`DbType`](https://docs.microsoft.com/en-us/dotnet/api/system.data.dbtype), [`SqlDbType`](https://docs.microsoft.com/en-us/dotnet/api/system.data.sqldbtype) and [`NpgsqlDbType`](https://www.npgsql.org/doc/api/NpgsqlTypes.NpgsqlDbType.html) enums.

Current State
---

For now, are handled:
  - Most of the members from `DbType`.
    - Those values are the ones tested.
    - `DbType.Binary` is not handled.
  - Some members from `SqlDbType` (witch fortuitously match `DbType` names) plus `Timestamp`.
    - Some basic values are not handle (like `Bit`, `UniqueIdentifier`, `Text`, ...)
  - Some members from `NpgsqlDbType` (witch fortuitously match `DbType` names) plus `Timestamp`.
    - Some basic values are not handle (like `Text`, `Numeric`, `Uuid`, ...)
    - Some values are badly interpreted (like `TimestampRange` interpreted as `Timestamp`)


With this PR
---

 Are handled:
  - All the members from `DbType`
  - Some members from [`SqlDbType`](https://docs.microsoft.com/en-us/dotnet/api/system.data.sqldbtype) (witch fortuitously match `DbType` names) plus `Timestamp`.
    - Some basic values are not handle (like `Bit`, `UniqueIdentifier`, `Text`, ...)
  - Some values from [`NpgsqlDbType`](https://www.npgsql.org/doc/api/NpgsqlTypes.NpgsqlDbType.html) (witch fortuitously match `DbType` names) plus `Timestamp`.
    - Some basic values are not handle (like `Text`, `Numeric`, `Uuid`, ...)
    - `TimestampRange` is no more badly interpreted as `Timestamp`

src/NLog.Database/DatabaseParameterInfo.cs
```diff
@@ -48,6 +49,160 @@ namespace NLog.Targets
...
+                { nameof(System.Data.DbType.Xml), typeof(string) },
+                { nameof(System.Data.DbType.DateTime2), typeof(DateTime) },
+                { nameof(System.Data.DbType.DateTimeOffset), typeof(DateTimeOffset) },
+
+
+                // Members from System.Data.SqlDbType
+
+                { "BigInt", typeof(long) },
+                // { "Binary", typeof(byte[]) }, // not supported
+                { "Bit", typeof(bool?) },
```
snakefoot, line 94: Why nullable ? Is bit tri-state by default ?

snakefoot:
Not sure it reduces complexity. Before it was just "best effort" to make it convient for users. Now looks like it is intended recognize all DbType for all DbProvider-implementations (along with any future values).

src/NLog.Database/DatabaseParameterInfo.cs
Orace: According to [the doc](https://docs.microsoft.com/en-us/dotnet/api/system.data.sqldbtype) yes. :confused:

Orace:
> Not sure it reduces complexity. Before it was just "best effort" to make it convient for users. Now looks like it is intended to recognize all DbType for all DbProvider-implementations (along with any future values and providers).

Agree on that. 
However, removing commented lines make the code more intelligible. 
Nevertheless current implementation [have issues](https://github.com/NLog/NLog/pull/4717#discussion_r773917997).

src/NLog.Database/DatabaseParameterInfo.cs
snakefoot: I prefer to stay away from very provider specific implementations, espcially when not able to verify if the that specific DbProvider is being used. Not sure I like this direction of trying to recognizes all DbTypes independent of DbProvider. But prefer to to just to recognize the most obvious ones.

src/NLog.Database/DatabaseParameterInfo.cs
snakefoot: Not sure I understand doesn't `DateTime` + `Date` + `DateTime2` map to `typeof(DateTime)` ?

src/NLog.Database/DatabaseParameterInfo.cs
Orace: I put `SqlDbType` and `NpgsqlDbType` because the doc states it's feasible. 

But I think that:

1. The `dbType` parameter should only accept names from `System.Data.DbType`.
2. `contains('Timestamp')` should be removed, it's ambiguous since it matches a `DateTime` for _NpgsqlDbType_ and a `byte[8]` for _SqlDbType_
3. `contains('Bool')` is replaced by `equals('Boolean')`

The last two points break the backward compatibility.

[The doc](https://github.com/NLog/NLog/wiki/Database-target) needs to be updated accordingly.  
By the way the examples are `SqlDbType.NChar` and `NpgsqlDbType.Json` and both are not parsed.

snakefoot: Since the method was only "best effort" and never meant as exact match for all major dbproviders (or lesser known), then I think the docs are very vague. Again it is mostly for convience, so it is easy to setup for the normal dbtypes (not for all dbtypes).

src/NLog.Database/DatabaseParameterInfo.cs
Orace: > Not sure I understand doesn't `DateTime`, `Date` and `DateTime2` map to `typeof(DateTime)` ?

Yes.

> Notice the current implementation is just best-effort to make a guess independent of DbProvider. Details like "Timestamp" for SqlDbType-provide might not be the same for a different DbProvider.

`Timestamp` meaning depends on the DbProvider. Actual implementation make it a `DateTime` choosing _NpgsqlDb_ over _SqlDb_.

I think that should not be the case and `Timestamp` should not be mapped.

Members of `System.Data.DbType` are representative: they map to any useful type.

snakefoot:
Since this is only about reducing code-smells and complexity. Would it be better if it just recognized DbType-enum-values only ? (And any overlap with enum-values from other DbProvider-DbType)

Orace:
> Since this is only about reducing code-smells and complexity. Would it be better if it just recognized DbType-enum-values only ? (And any name-overlap with enum-values from other DbProvider-DbType)

I think so:
 - It's enough (all useful types are covered, none are removed)
 - It's easier to document (just a link to [System.Data.DbType](https://docs.microsoft.com/en-us/dotnet/api/system.data.dbtype))
 - It's _dbProvider-agnostic_

It's already in the last commit of this PR.

src/NLog.Database/DatabaseParameterInfo.cs
```diff
@@ -203,61 +236,15 @@ internal bool SetDbType(IDbDataParameter dbParameter)
             return true;    // DbType not in use
         }
 
-        private static Type TryParseDbType(string dbTypeString)
+        private static Type TryParseDbType(string dbTypeName)
         {
-            string[] dbTypeNames = dbTypeString?.Split(new[] { '.' }, StringSplitOptions.RemoveEmptyEntries);
-            dbTypeString = dbTypeNames?.Length > 0 ? dbTypeNames[dbTypeNames.Length - 1] : dbTypeString;
+            // retrieve the type name if a full name is given
+            dbTypeName = dbTypeName?.Substring(dbTypeName.LastIndexOf('.') + 1);
```
snakefoot, line 242: Can we add `Trim()` at the tail?

snakefoot:
Agree that the code has reduced complexity and more predictable now.

snakefoot (approved):

snakefoot:
@Orace Guess the failing unit-test `ParameterDbTypePropertyNameTest` has to be fixed

snakefoot:
Think you are little too quick. The unit-test verifies that it can map DbType to correct type, by changing the assert, then it is no longer testing that it works.

The problem is that `MockDbType.int32` is no longer recognized as an integer. Which is strange since it should match nicely with `DbType.Int32`. Maybe a bug in the new code ?

snakefoot:
Scratch that. Actually looks completely correct. Sorry about that.

Orace:
I put some details in the comment of 48fadbb, indeed the old code output a string, witch is the default behavior if the dbType parsing fail (it was the case since there was no trimming and `dbType=' MockDbType.int32  '` was nor parsed).

tests/NLog.Database.Tests/DatabaseTargetTests.cs
```diff
@@ -868,7 +868,7 @@ public void ParameterDbTypePropertyNameTest()
 Parameter #1 Direction=Input
 Parameter #1 Name=@level
 Parameter #1 MockDbType=Int32
-Parameter #1 Value=""{0}""
+Parameter #1 Value={0}
```
Orace, line 871: Now `dbType=' MockDbType.int32  '` is correctly interpreted and the value is rendered as an integer instead of a string; hence no more double quotes.

src/NLog.Database/DatabaseParameterInfo.cs
```diff
@@ -203,61 +236,15 @@ internal bool SetDbType(IDbDataParameter dbParameter)
             return true;    // DbType not in use
         }
 
-        private static Type TryParseDbType(string dbTypeString)
+        private static Type TryParseDbType(string dbTypeName)
         {
-            string[] dbTypeNames = dbTypeString?.Split(new[] { '.' }, StringSplitOptions.RemoveEmptyEntries);
-            dbTypeString = dbTypeNames?.Length > 0 ? dbTypeNames[dbTypeNames.Length - 1] : dbTypeString;
+            // retrieve the type name if a full name is given
+            dbTypeName = dbTypeName?.Substring(dbTypeName.LastIndexOf('.') + 1).Trim();
```
304NotModified, line 242: Will this crash of the string ends with a `.`?

304NotModified: Happy new 2022!

Orace: > Will this crash when the string ends with a `.`?

Good catch. Will add a test and fix this.

Orace: Actually, it does not crash, the empty string is returned.  
See [this fiddle](https://dotnetfiddle.net/g5QDwY).
[The documentation](https://docs.microsoft.com/en-us/dotnet/api/system.string.substring?view=net-6.0#System_String_Substring_System_Int32_) agrees too:

> [crash if] `startIndex` is less than zero or greater than the length of this instance.

304NotModified: Yes indeed. Thanks for the test cases :)

304NotModified:
Thanks for the pr!

I like to merge it, but I doubt its a pure fix for code smell. (The behavior has changed) . Maybe we should rename the pr? 
Something like "databasetarget: improved parsing of dbtype"  ?

snakefoot:
Yes now 1 code smell less to worry about in the new year!

PR #4724, opened 2022-01-08
VariableLayoutRendererTests - Changed to isolated LogFactory

snakefoot:


snakefoot:
Looks like `ValueWithVariableMustNotCauseInfiniteRecursion` is not testing what it should.

Think the idea is that it should show that assigning `LogFactory.Configuration = null` should not cause `LogFactory.Configuration` to return null, but automatically reload NLog-config.

PR #4757, opened 2022-01-19
Updated JetBrains Annotations with StructuredMessageTemplateAttribute

snakefoot:
Removed code-regions with ASP + MVC + Razor-attributes, since not relevant for NLog-core-project.


snakefoot:
Guess one also could add `StringFormatMethodAttribute﻿` but it seems it conflicts with using structured-message-templates:

See also: https://github.com/aspnet/Logging/issues/340 + https://youtrack.jetbrains.com/issue/RSRP-483165

snakefoot:
@304NotModified Think this should trigger release of NLog 5.0 RC2. To ensure the new attribute doesn't hurt too much.

PR #4773, opened 2022-01-21
Fix xml-docs and replaced broken link config.html with wiki-link

snakefoot:
See also: https://github.com/NLog/NLog.github.io/issues/22

### DIFF
--- a/src/NLog.Database/DatabaseTarget.cs
+++ b/src/NLog.Database/DatabaseTarget.cs
@@ -59,7 +59,10 @@ namespace NLog.Targets
     /// Writes log messages to the database using an ADO.NET provider.
     /// </summary>
     /// <remarks>
-    /// - NETSTANDARD cannot load connectionstrings from .config
+    /// <para>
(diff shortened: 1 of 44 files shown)


src/NLog.Database/DatabaseTarget.cs
```diff
@@ -78,7 +81,6 @@ namespace NLog.Targets
     /// <code lang="C#" source="examples/targets/Configuration API/Database/MSSQL/Example.cs" height="630" />
     /// </example>
     [Target("Database")]
-    [Target("DB")]
```
304NotModified, line 81: Is this on purpose?

snakefoot: Yes saw it while fixing the xml-docs, and prefer not to promote [acronyms](https://acronyms.thefreedictionary.com/DB)

304NotModified: Its disappointing this in a bit hidden in a large pr. 

(Its not on the title, summary or even related)

snakefoot: Ok. Will restore it, and remove it again in new PR.

PR #4787, opened 2022-01-30
Update Logger-methods with JetBrains StructuredMessageTemplateAttribute (missing V1 legacy)

snakefoot:
Followup to #4757. Looks like Resharper recognize NLog-Logger-methods without help from these attributes, but like consistency.

### DIFF
--- a/src/NLog/Abstractions/ILoggerBase.cs
+++ b/src/NLog/Abstractions/ILoggerBase.cs
@@ -35,6 +35,7 @@ namespace NLog
 {
     using System;
     using System.ComponentModel;
+    using JetBrains.Annotations;
 
     /// <summary>
     /// Logger with only generic methods (passing 'LogLevel' to methods) and core properties.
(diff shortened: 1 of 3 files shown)


TheBigNeo:
Hi
Maybe this problem is only with me and I have not yet found the right settings. 

But since `[Localizable(false)]` was added, the SpellChecker does not work anymore. If a string is passed to a function which has `[Localizable(false)]`, the string is not checked.

Is this just happening to me or can you reproduce it?

TheBigNeo

image

TheBigNeo:
It looks like it is the ReSpeller (JetBrains, Rider) which controls the spelling. 

What is the reason that this attribute has been added?
I have created a ticket at JetBrains to ask why the ReSpeller does not check this.

https://youtrack.jetbrains.com/issue/RIDER-87787/ReSpeller-does-not-check-Localizable-strings

TheBigNeo:
JetBrains has fixed the problem with Rider 2023.3.
There is now a new setting `Check strings explicitly marked as non-localizable`

image

PR #4369, opened 2021-03-22
LoggingConfigurationParser - Recognize LoggingRule.DefaultFilterAction

snakefoot:
Like that the recognized properties matches the name of the actual properties. Ex, [LoggingRule.DefaultFilterResult](https://nlog-project.org/documentation/v4.6.0/html/P_NLog_Config_LoggingRule_DefaultFilterResult.htm) (Though it would be a better match if the property was named `DefaultFilterAction`)

Adds support for doing this in appsetting.json:
```json
        "rules": [
          {
            "logger": "*",
            "minLevel": "Trace",
            "writeTo": "Console",
            "defaultFilterAction": "Log",
            "filters": [
              {
                "type": "when",
                "condition": "contains('${message}','HeartbeatRequest')",
                "action": "Ignore"
              },
              {
                "type": "when",
                "condition": "contains('${message}','HeartbeatResponse')",
                "action": "Ignore"
              }
            ]
          }
        ]
```

See also: https://github.com/NLog/NLog.Extensions.Logging/issues/491

PR #4805, opened 2022-02-24
NLogConfigurationException - Skip string.Format when no args

snakefoot:
Fixed bug introduced with NLog 4.7.9 in #4369 (Bug is only triggered when using invalid NLog.config).

See also https://stackoverflow.com/questions/71254198/nlog-input-string-was-not-in-a-correct-format-using-windows-identity


snakefoot:
Really annoying, that it was not reported 2 days ago :)

PR #4843, opened 2022-03-16
Loading candidate config files with better support for nlog.dll.nlog

snakefoot:


snakefoot:
Starting to think this is not an improvement, but just making it harder to maintain.

Think that finding the location of NLog.dll is not that relevant, especially after Microsoft starting using symbolic-links to the nuget-cache. Then the chance of finding a useful `nlog.dll.nlog`-file is very unlikely.

Issue #4817, opened 2022-03-01
LogFactory: SuspendLogging() does not suspend if ResumeLogging() was called before


TobiasKnauss:
**NLog version**: 5.0.0-RC2 (the GitHub code as of 2022-March-01)

**Platform**: .NET 5

- What is the current result?
Multiple (accidental) calls to `ResumeLogging()` increase the internal `_logsEnabled` counter to values above 0. It should be 0 maximum.
As a consequence, a call to `SuspendLogging()` does not suspend.

- What is the expected result?
`SuspendLogging()` should always suspend, no matter how often `ResumeLogging()` was called before.

- Did you checked the [Internal log](https://github.com/NLog/NLog/wiki/Internal-Logging)?
No, I checked the source code.

- Are there any workarounds?
no

- Is there a version in which it did work? 
Don't know.

- Suggested fix:
```
        public void ResumeLogging()
        {
            lock (_syncRoot)
            {
                _logsEnabled = Math.Min (0, _logsEnabled + 1);
                if (_logsEnabled == 0)
                {
                    ReconfigExistingLoggers();
                }
            }
        }
```

- Can you help us by writing an unit test?
I could, but this would take much more time than if you wrote it your self. Here's the logic for it:
```
private static void NLogUnitTest_ResumeLogging_SuspendLogging_1 ()
{
  // ----- arrange -----
  const int maxSuspendDepth = 100;

  var logFactory = NLog.LogManager.LogFactory;

  // ----- act & assert -----
  // check initial state
  Assert.IsTrue (logFactory.IsLoggingEnabled ());

  // suspend and resume once
  logFactory.SuspendLogging ();
  Assert.IsFalse (logFactory.IsLoggingEnabled ());
  logFactory.ResumeLogging ();
  Assert.IsTrue (logFactory.IsLoggingEnabled ());

  // suspend multiple times and resume by the same number
  for (int counter = 1; counter <= maxSuspendDepth; counter++)
  {
    logFactory.SuspendLogging ();
    Assert.IsFalse (logFactory.IsLoggingEnabled ());
  }
  for (int counter = 1; counter <= maxSuspendDepth; counter++)
  {
    Assert.IsFalse (logFactory.IsLoggingEnabled ());
    logFactory.ResumeLogging ();
  }
  Assert.IsTrue (logFactory.IsLoggingEnabled ());

  // resume without suspend; subsequent SuspendLogging() call must still suspend.
  logFactory.ResumeLogging ();
[...]

PR #4846, opened 2022-03-21
Improvements in XML documentation, LogLevel usage, ResumeLogging() and SuspendLogging()

TobiasKnauss:
Providing enhancements and improvements as discussed in
https://github.com/NLog/NLog/discussions/4816
https://github.com/NLog/NLog/issues/4817

- automatic correction of swapped log levels.
- LogFactory: ResumeLogging(bool) and SuspendLogging(bool) added.
- Some [Obsolete] added.
- XML documentation enhanced

### DIFF
--- a/src/NLog/Config/LoggingConfiguration.cs
+++ b/src/NLog/Config/LoggingConfiguration.cs
@@ -67,7 +67,7 @@ namespace NLog.Config
         public LogFactory LogFactory { get; }
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="LoggingConfiguration" /> class.
+        /// Initializes a new instance of the <see cref="LoggingConfiguration"/> class, which uses the <see cref="LogFactory"/> from the <see cref="LogManager"/>.
         /// </summary>
         public LoggingConfiguration()
             : this(LogManager.LogFactory)
@@ -75,8 +75,12 @@ namespace NLog.Config
         }
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="LoggingConfiguration" /> class.
+        /// Initializes a new instance of the <see cref="LoggingConfiguration"/> class, which uses the given <paramref name="logFactory"/>.
         /// </summary>
+        /// <param name="logFactory">
+        /// The <see cref="LogFactory"/> that will be used in this configuration.
+        /// If <see langword="null"/>, the <see cref="LogFactory"/> from the <see cref="LogManager"/> will be used.
+        /// </param>
         public LoggingConfiguration(LogFactory logFactory)
         {
             LogFactory = logFactory ?? LogManager.LogFactory;
@@ -285,7 +289,9 @@ namespace NLog.Config
         }
 
         /// <summary>
-        /// Add a rule with min- and maxLevel.
+        /// Add a rule with <paramref name="minLevel"/> and <paramref name="maxLevel"/>. The rule will NOT have a name.<br/>
+        /// If a name is required, use <see cref="LoggingRule(string)"/> instead and add the created rule to the
+        /// <see cref="LoggingRules"/> collection of this class.
         /// </summary>
         /// <param name="minLevel">Minimum log level needed to trigger this rule.</param>
         /// <param name="maxLevel">Maximum log level needed to trigger this rule.</param>
@@ -303,7 +309,9 @@ namespace NLog.Config
         }
 
         /// <summary>
-        /// Add a rule with min- and maxLevel.
+        /// Add a rule with <paramref name="minLevel"/> and <paramref name="maxLevel"/>. The rule will NOT have a name.<br/>
+        /// If a name is required, use <see cref="LoggingRule(string)"/> instead and add the created rule to the
+        /// <see cref="LoggingRules"/> collection of this class.
         /// </summary>
         /// <param name="minLevel">Minimum log level needed to trigger this rule.</param>
         /// <param name="maxLevel">Maximum log level needed to trigger this rule.</param>
@@ -316,7 +324,9 @@ namespace NLog.Config
         }
 
         /// <summary>
-        /// Add a rule with min- and maxLevel.
+        /// Add a rule with <paramref name="minLevel"/> and <paramref name="maxLevel"/>. The rule will NOT have a name.<br/>
+        /// If a name is required, use <see cref="LoggingRule(string)"/> instead and add the created rule to the
+        /// <see cref="LoggingRules"/> collection of this class.
         /// </summary>
         /// <param name="minLevel">Minimum log level needed to trigger this rule.</param>
         /// <param name="maxLevel">Maximum log level needed to trigger this rule.</param>
@@ -331,9 +341,11 @@ namespace NLog.Config
         }
 
         /// <summary>
-        /// Add a rule for one loglevel.
+        /// Add a rule for the given <paramref name="level"/>. The rule will NOT have a name.<br/>
+        /// If a name is required, use <see cref="LoggingRule(string)"/> instead and add the created rule to the
+        /// <see cref="LoggingRules"/> collection of this class.
         /// </summary>
-        /// <param name="level">log level needed to trigger this rule. </param>
+        /// <param name="level">The log level needed to trigger this rule.</param>
         /// <param name="targetName">Name of the target to be written when the rule matches.</param>
         /// <param name="loggerNamePattern">Logger name pattern. It may include the '*' wildcard at the beginning, at the end or at both ends.</param>
         public void AddRuleForOneLevel(LogLevel level, string targetName, string loggerNamePattern = "*")
@@ -348,9 +360,11 @@ namespace NLog.Config
         }
 
         /// <summary>
-        /// Add a rule for one loglevel.
+        /// Add a rule for the given <paramref name="level"/>. The rule will NOT have a name.<br/>
+        /// If a name is required, use <see cref="LoggingRule(string)"/> instead and add the created rule to the
+        /// <see cref="LoggingRules"/> collection of this class.
         /// </summary>
-        /// <param name="level">log level needed to trigger this rule. </param>
+        /// <param name="level"> The log level needed to trigger this rule. </param>
         /// <param name="target">Target to be written to when the rule matches.</param>
         /// <param name="loggerNamePattern">Logger name pattern. It may include the '*' wildcard at the beginning, at the end or at both ends.</param>
         public void AddRuleForOneLevel(LogLevel level, Target target, string loggerNamePattern = "*")
@@ -360,9 +374,11 @@ namespace NLog.Config
         }
 
         /// <summary>
-        /// Add a rule for one loglevel.
+        /// Add a rule for the given <paramref name="level"/>. The rule will NOT have a name.<br/>
+        /// If a name is required, use <see cref="LoggingRule(string)"/> instead and add the created rule to the
+        /// <see cref="LoggingRules"/> collection of this class.
         /// </summary>
-        /// <param name="level">log level needed to trigger this rule. </param>
+        /// <param name="level"> The log level needed to trigger this rule. </param>
         /// <param name="target">Target to be written to when the rule matches.</param>
         /// <param name="loggerNamePattern">Logger name pattern. It may include the '*' wildcard at the beginning, at the end or at both ends.</param>
         /// <param name="final">Gets or sets a value indicating whether to quit processing any further rule when this one matches.</param>
@@ -376,7 +392,9 @@ namespace NLog.Config
         }
 
         /// <summary>
-        /// Add a rule for all loglevels.
+        /// Add a rule for all levels. The rule will NOT have a name.<br/>
+        /// If a name is required, use <see cref="LoggingRule(string)"/> instead and add the created rule to the
+        /// <see cref="LoggingRules"/> collection of this class.
         /// </summary>
         /// <param name="targetName">Name of the target to be written when the rule matches.</param>
         /// <param name="loggerNamePattern">Logger name pattern. It may include the '*' wildcard at the beginning, at the end or at both ends.</param>
@@ -392,7 +410,9 @@ namespace NLog.Config
         }
 
         /// <summary>
-        /// Add a rule for all loglevels.
+        /// Add a rule for all levels. The rule will NOT have a name.<br/>
+        /// If a name is required, use <see cref="LoggingRule(string)"/> instead and add the created rule to the
+        /// <see cref="LoggingRules"/> collection of this class.
         /// </summary>
         /// <param name="target">Target to be written to when the rule matches.</param>
         /// <param name="loggerNamePattern">Logger name pattern. It may include the '*' wildcard at the beginning, at the end or at both ends.</param>
@@ -403,7 +423,9 @@ namespace NLog.Config
         }
 
         /// <summary>
-        /// Add a rule for all loglevels.
+        /// Add a rule for all levels. The rule will NOT have a name.<br/>
+        /// If a name is required, use <see cref="LoggingRule(string)"/> instead and add the created rule to the
+        /// <see cref="LoggingRules"/> collection of this class.
         /// </summary>
         /// <param name="target">Target to be written to when the rule matches.</param>
         /// <param name="loggerNamePattern">Logger name pattern. It may include the '*' wildcard at the beginning, at the end or at both ends.</param>
@@ -439,10 +461,12 @@ namespace NLog.Config
         }
 
         /// <summary>
-        /// Removes the specified named logging rule.
+        /// Removes the logging rule with the specified name.
+        /// This method only works if the rule, which shall be removed, has a name.
+        /// If it has no name, edit the <see cref="LoggingRules"/> collection of this class instead.
         /// </summary>
-        /// <param name="ruleName">The name of the logging rule to be removed.</param>
-        /// <returns>Found one or more logging rule to remove, or <see langword="false"/> when not found.</returns>
+        /// <param name="ruleName"> The name of the logging rule to remove. </param>
+        /// <returns><see langword="true"/> if one or more logging rules were removed, or <see langword="false"/> if no rule was found.</returns>
         public bool RemoveRuleByName(string ruleName)
         {
             if (ruleName is null)
--- a/src/NLog/Config/LoggingRule.cs
+++ b/src/NLog/Config/LoggingRule.cs
@@ -63,6 +63,7 @@ namespace NLog.Config
         /// <summary>
         /// Create an empty <see cref="LoggingRule" />.
         /// </summary>
+        /// <param name="ruleName">The name of the logging rule. This name can be used to identify the rule.</param>
         public LoggingRule(string ruleName)
         {
             RuleName = ruleName;
@@ -74,8 +75,8 @@ namespace NLog.Config
         /// Create a new <see cref="LoggingRule" /> with a <paramref name="minLevel"/> and  <paramref name="maxLevel"/> which writes to <paramref name="target"/>.
         /// </summary>
         /// <param name="loggerNamePattern">Logger name pattern used for <see cref="LoggerNamePattern"/>. It may include one or more '*' or '?' wildcards at any position.</param>
-        /// <param name="minLevel">Minimum log level needed to trigger this rule.</param>
-        /// <param name="maxLevel">Maximum log level needed to trigger this rule.</param>
+        /// <param name="minLevel">Minimum log level needed to trigger this rule. Log levels range from <see cref="LogLevel.Trace"/> (lowest) to <see cref="LogLevel.Fatal"/> (highest).</param>
+        /// <param name="maxLevel">Maximum log level needed to trigger this rule. Log levels range from <see cref="LogLevel.Trace"/> (lowest) to <see cref="LogLevel.Fatal"/> (highest).</param>
         /// <param name="target">Target to be written to when the rule matches.</param>
         public LoggingRule(string loggerNamePattern, LogLevel minLevel, LogLevel maxLevel, Target target)
             : this()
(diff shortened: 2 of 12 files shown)


src/NLog/Config/LoggingConfiguration.cs
```diff
@@ -285,7 +289,9 @@ public TTarget FindTargetByName<TTarget>(string name)
         }
 
         /// <summary>
-        /// Add a rule with min- and maxLevel.
+        /// Add a rule with <paramref name="minLevel"/> and <paramref name="maxLevel"/>. The rule will NOT have a name.<br/>
```
snakefoot, line 292: The LoggingRule-name is an exotic feature, that is only relevant for special users that wants to special changes to NLog-configuration at runtime. Think normal users will wonder what is the meaning of rule-name? Is a rule-name needed? What happens if no rule-name is provided? I prefer that xml-comments doesn't mention all the stuff that are not assigned.

src/NLog/Config/LoggingConfiguration.cs
```diff
@@ -303,7 +309,9 @@ public void AddRule(LogLevel minLevel, LogLevel maxLevel, string targetName, str
...
+        /// If a name is required, use <see cref="LoggingRule(string)"/> instead and add the created rule to the
```
snakefoot, line 313: The LoggingRule-name is an exotic feature, that is only relevant for special users that wants to special changes to NLog-configuration at runtime. Think normal users will wonder what is the meaning of rule-name? Is a rule-name needed? What happens if no rule-name is provided? I prefer that xml-comments doesn't mention all the stuff that are not assigned.

src/NLog/Config/LoggingConfiguration.cs
```diff
@@ -316,7 +324,9 @@ public void AddRule(LogLevel minLevel, LogLevel maxLevel, Target target, string
...
```
snakefoot, line 328: The LoggingRule-name is an exotic feature, that is only relevant for special users that wants to special changes to NLog-configuration at runtime. Think normal users will wonder what is the meaning of rule-name? Is a rule-name needed? What happens if no rule-name is provided? I prefer that xml-comments doesn't mention all the stuff that are not assigned.

src/NLog/Config/LoggingConfiguration.cs
```diff
@@ -331,9 +341,11 @@ public void AddRule(LogLevel minLevel, LogLevel maxLevel, Target target, string
         }
 
         /// <summary>
-        /// Add a rule for one loglevel.
+        /// Add a rule for the given <paramref name="level"/>. The rule will NOT have a name.<br/>
+        /// If a name is required, use <see cref="LoggingRule(string)"/> instead and add the created rule to the
```
snakefoot, line 345: The LoggingRule-name is an exotic feature, that is only relevant for special users that wants to special changes to NLog-configuration at runtime. Think normal users will wonder what is the meaning of rule-name? Is a rule-name needed? What happens if no rule-name is provided? I prefer that xml-comments doesn't mention all the stuff that are not assigned.

src/NLog/Config/LoggingConfiguration.cs
```diff
@@ -360,9 +374,11 @@ public void AddRuleForOneLevel(LogLevel level, Target target, string loggerNameP
...
+        /// <see cref="LoggingRules"/> collection of this class.
```
snakefoot, line 379: The LoggingRule-name is an exotic feature, that is only relevant for special users that wants to special changes to NLog-configuration at runtime. Think normal users will wonder what is the meaning of rule-name? Is a rule-name needed? What happens if no rule-name is provided? I prefer that xml-comments doesn't mention all the stuff that are not assigned.

src/NLog/Config/LoggingConfiguration.cs
```diff
@@ -376,7 +392,9 @@ public void AddRuleForOneLevel(LogLevel level, Target target, string loggerNameP
         }
 
         /// <summary>
-        /// Add a rule for all loglevels.
+        /// Add a rule for all levels. The rule will NOT have a name.<br/>
+        /// If a name is required, use <see cref="LoggingRule(string)"/> instead and add the created rule to the
+        /// <see cref="LoggingRules"/> collection of this class.
```
snakefoot, line 397: The LoggingRule-name is an exotic feature, that is only relevant for special users that wants to special changes to NLog-configuration at runtime. Think normal users will wonder what is the meaning of rule-name? Is a rule-name needed? What happens if no rule-name is provided? I prefer that xml-comments doesn't mention all the stuff that are not assigned.

src/NLog/Config/LoggingConfiguration.cs
```diff
@@ -392,7 +410,9 @@ public void AddRuleForAllLevels(string targetName, string loggerNamePattern = "*
...
```
snakefoot, line 415: The LoggingRule-name is an exotic feature, that is only relevant for special users that wants to special changes to NLog-configuration at runtime. Think normal users will wonder what is the meaning of rule-name? Is a rule-name needed? What happens if no rule-name is provided? I prefer that xml-comments doesn't mention all the stuff that are not assigned.

src/NLog/Config/LoggingConfiguration.cs
```diff
@@ -403,7 +423,9 @@ public void AddRuleForAllLevels(Target target, string loggerNamePattern = "*")
...
```
snakefoot, line 428: The LoggingRule-name is an exotic feature, that is only relevant for special users that wants to special changes to NLog-configuration at runtime. Think normal users will wonder what is the meaning of rule-name? Is a rule-name needed? What happens if no rule-name is provided? I prefer that xml-comments doesn't mention all the stuff that are not assigned.

src/NLog/Config/LoggingConfiguration.cs
```diff
@@ -439,10 +461,12 @@ public LoggingRule FindRuleByName(string ruleName)
         }
 
         /// <summary>
-        /// Removes the specified named logging rule.
+        /// Removes the logging rule with the specified name.
+        /// This method only works if the rule, which shall be removed, has a name.
```
snakefoot, line 465: This is pretty much echoes the initial sentence, so doesn't feel needed. But I like the idea of mentioning `<see cref="LoggingRules.RuleName"/>`.

src/NLog/Config/LoggingRule.cs
```diff
@@ -74,8 +75,8 @@ public LoggingRule(string ruleName)
         /// Create a new <see cref="LoggingRule" /> with a <paramref name="minLevel"/> and  <paramref name="maxLevel"/> which writes to <paramref name="target"/>.
         /// </summary>
         /// <param name="loggerNamePattern">Logger name pattern used for <see cref="LoggerNamePattern"/>. It may include one or more '*' or '?' wildcards at any position.</param>
-        /// <param name="minLevel">Minimum log level needed to trigger this rule.</param>
-        /// <param name="maxLevel">Maximum log level needed to trigger this rule.</param>
+        /// <param name="minLevel">Minimum log level needed to trigger this rule. Log levels range from <see cref="NLog.LogLevel.Trace"/> (lowest) to <see cref="NLog.LogLevel.Fatal"/> (highest).</param>
```
snakefoot, line 78: Not sure I like to define details LogLevel all around the code. If people need to know of the available LogLevels, then then it should be explained for the LogLevel-type.

src/NLog/Config/LoggingRule.cs
```diff
@@ -74,8 +75,8 @@ public LoggingRule(string ruleName)
...
+        /// <param name="maxLevel">Maximum log level needed to trigger this rule. Log levels range from <see cref="NLog.LogLevel.Trace"/> (lowest) to <see cref="NLog.LogLevel.Fatal"/> (highest).</param>
```
snakefoot, line 79: Not sure I like to define details LogLevel all around the code. If people need to know of the available LogLevels, then then it should be explained for the LogLevel-type.

src/NLog/Config/LoggingRuleLevelFilter.cs
```diff
@@ -62,9 +62,13 @@ public LoggingRuleLevelFilter GetSimpleFilterForUpdate()
 
         public LoggingRuleLevelFilter SetLoggingLevels(LogLevel minLevel, LogLevel maxLevel, bool enable)
         {
-            for (int i = minLevel.Ordinal; i <= Math.Min(maxLevel.Ordinal, LogLevels.Length - 1); ++i)
+            // If minLevel and maxLevel are swapped, swap again to correct it.
```
snakefoot, line 65: Could this code-change be move to another pull-requests, since it does more than improving xml-comments.

src/NLog/LogLevel.cs
```diff
@@ -38,43 +38,43 @@ namespace NLog
...
     {
         /// <summary>
-        /// Trace log level.
+        /// 'Trace' log level. This is the 1st log level, which is the lowest log level.
         /// </summary>
         public static readonly LogLevel Trace = new LogLevel("Trace", 0);
 
         /// <summary>
-        /// Debug log level.
+        /// 'Debug' log level. This is the 2nd log level (lowest log level = <see cref="Trace"/>).
```
snakefoot, line 52: Think I prefer `Debug log level (Ordinal = 1)`

src/NLog/LogManager.cs
```diff
@@ -45,7 +45,11 @@ namespace NLog
     using NLog.Internal;
 
     /// <summary>
-    /// Creates and manages instances of <see cref="NLog.Logger" /> objects.
+    /// Creates and manages instances of <see cref="NLog.Logger"/> objects.<br/>
+    /// The purpose of LogManager is that one <see cref="LogFactory"/> can be easily used without having to carry around an instance of it.
```
snakefoot, line 49: Not a big fan of very long explainations in xml-comments. Then it is better with a reference to a tutorial-page.

But I think it makes sense to explain that `LogManager wraps a singleton instance of LogFactory.`

snakefoot:
Thank you for the contributition. I know it has been frustrating to make the NLog-source-code work together with your Resharper-configuration.

I hope you will accept that this pull-request is a mixture of many things, and it would be best to handle each item one at a time:
- Xml-comments
- `SetLoggingLevels`
- `SuspendLogging(bool force)`

I hope you will not take my comments as pure critizism, and know that my goal is also to make the NLog-project easy to access.

src/NLog/Config/LoggingConfiguration.cs
TobiasKnauss: In our current implementation of logging in our software, which uses .NET's `TraceSource`, the operator may change the logging levels of all classes (i.e. of each class that contains a `TraceSource`) at any time. Logging has become an essential part in out software, because it's not a "normal" windows desktop application, but an application that controls a complex production machine. We may not simply shut down the application and restart with different logging levels. Also, logging is so important because there are not thousands or millions of copies of our software, but just a few hundred. And each copy has a different configuration because of different machine setup. If a failure occurs in a customer's machine, we take the logs from this machine and solve the problem within hours. These logs must have tons of information in order to find the problem's cause immediately. And this requires a highly configurable logging system, otherwise our log files would reach hundreds of megabytes within minutes.
These are the reasons that make the name of the logging rule very relevant for us. We will configure NLog almost always only programmatically (default logging level `warning` for all loggers) and not via config file, although there may be a config file to set up an initial configuration (to log details if something goes wrong during machine startup).

src/NLog/Config/LoggingConfiguration.cs
TobiasKnauss: Like I answered above, the name of the logging rule is essential for us. Therefore I would call it an exotic feature, but the only available way to identify and change existing logging rules. This should be reflected by the documentation. Maybe most users don't use it, and maybe they will never need it, but maybe they will see a use for it as soon as they know better about the purpose and the possibilities, and when they understand that NLog configuration is not that static as it appears to be from using a config file (which is read once in the beginning).

src/NLog/Config/LoggingRule.cs
TobiasKnauss: I added this part because in my opinion there were no sufficient information about what a log level can be, and especially which log levels are high and which are low. In .NET, it's just the other way round: `Trace` is a high log level there, because it produces most log messages. I do not understand why you named your log levels in the opposite way, but of course I need to accept it. But I would add the extra information to make it clear. And I don't see any disadvantages in enhancing the documentation. The XML doc in NLog is quite short, and I think that my additions will help newcomers to understand NLog better and faster (which is the reason why I made all these proposed changes).

src/NLog/LogLevel.cs
TobiasKnauss: The ordinals have no relevance for users of NLog, at least not for me. The user (in this case, me) wants to know which log levels exist and what they mean. They see the list of members when they type the class name, and they put the cursor or selection at one member. And then they read `'Debug' log level`... What?? I know that it's a log level, because it's in the `LogLevel` class. I know that it's the `Debug` log level because it's named so. The current comment has absolutely no use. But when I tell the user, that it's the second log level and the lowest one is 'Trace' (see above mentioned order of log levels), then they immediately get some useful information.
You might argue against this or other enhancements with a duplication of information, but in documentations you will often find such duplications, where descriptions are repeated on different places. This is not a violation of DRY, because a) it's no code, and b) it helps the user. And if you really don't like duplicate descriptions, you might use XML doc `include` or `inheritdoc` (we use both often) to remove duplicates from the documentation where possible.

src/NLog/LogManager.cs
TobiasKnauss: > Not a big fan of very long explainations in xml-comments.

I have noticed that from the very short XML documentation.

I think that the XML documentation should contain all necessary information to use a library without needing to read tutorials. The tutorial or manual should explain how a software works and how the API looks like in detail. The first part (how it works) could be described in XML comments only in a limited way (e.g. in the description of the classes). The API itself should be properly documented in XML comments, and to reduce your work the API part of the tutorial or manual could partly or fully be built from the XML comments using suitable tools.
Also, the XML documentation often is the only available documentation of a specific version of a library. The tutorials or manuals are usually adapted to reflect the latest version, and documentation of prior versions often is not kept.
Also, there may be situations when no internet is available, and unless you saved a copy of the website or the relevant pages locally, the XML documentation would be the only help you have then.
And the fact about LogManager as a LogFactory wrapper actually was not mentioned anywhere.

TobiasKnauss: I would like to give you an example of a well documented method in our software:

```
  public static partial class CMath
  {
    public static partial class Statistics
    {
      public static partial class DotGridClustering
      {
        /// ------------------------------------------------------------------
        /// <summary>
        ///   Calculate clusters of rings from the given collection of positions of objects.<br /><br />
        ///   The clusters are created by a sequence of operations:<br />
        ///   1) Find connections between the object positions using the DBSCAN method with a
        ///   search distance that is calculated from the maximum ring diameter and
        ///   the maximum object count per ring.<br />
        ///   <i> The following operations are executed twice for each cluster: </i><br />
        ///   2) Calculate the circle from the object positions of the cluster.<br />
        ///   3) Filter the object positions of the cluster based on the deviations
        ///   of [the distances of the objects from the calculated circle center]
        ///   from [the calculated circle radius].<br />
        ///   4) Calculate the circle again.<br />
        ///   5) Re-add object positions to the cluster based on the same condition as in 3).
        /// </summary>
        /// <param name="i_positions"> The positions of objects that are used to calculate the ring clusters. </param>
        /// <param name="i_ringDiameterRange"> The allowed range of the diameters of the rings. </param>
        /// <param name="i_nominalObjectCountPerRing"> The nominal number of objects on a ring. </param>
        /// <param name="i_maximumNeighborGapSize"> The maximum size of a gap between neighboring objects when searching for connections. The gap size is specified as a number of objects. </param>
        /// <param name="i_objectToCenterDistanceTolerances"> The tolerances on [the distances of the objects from the calculated circle center] against [the calculated circle radius]. The lower value must be &lt;= 0, the upper value must be &gt;= 0. </param>
        /// <param name="i_isToleranceRelative"> A flag that specifies whether the given tolerances <paramref name="i_objectToCenterDistanceTolerances" /> contain absolute (if "false") or relative (if "true") values. </param>
        /// <param name="o_clusters"> An outer array, representing the clusters, of an inner array, containing the indices of the objects positions of the cluster. </param>
        /// <returns> <see cref="CResult" /> </returns>
        /// ------------------------------------------------------------------
        public static CResult Calculate01 (IList<CLocation>   i_positions,
                                           CRange<CPhqLength> i_ringDiameterRange,
                                           int                i_nominalObjectCountPerRing,
                                           int                i_maximumNeighborGapSize,
                                           CRange<CPhqLength> i_objectToCenterDistanceTolerances,
                                           bool               i_isToleranceRelative,
                                           out int[][]        o_clusters)
        {

```
which is shown as
grafik

Whenever I look at this documentation, I get all relevant information that I need to use the method somewhere. But this is not the full information; the documentation of this method in our wiki is much longer. If someone needs the details, they will look there. And if someone is not interested in the details or not interested in the XML documentation... Nobody forces you to read it, everybody is free to read or just ignore it. But it should be there.

TobiasKnauss:
> I know it has been frustrating to make the NLog-source-code work together with your Resharper-configuration.

Well, in the end I did not. It was a useless approach. I just edited the relevant parts and then used Beyond Compare to revert all parts that should not have changed.

I may split up the pull request in 3 parts. Please tell me first whether you would accept the PR with the XML comments, because if you don't like it (I have given my replies to your comments), I could avoid that part of extra work. I know that commenting code and writing documentation often is not a nice job, and it sometimes seems to be useless work. But I can assure you that it is not, and I hope you slightly change your mind towards a more comprehensive documentation.

src/NLog/LogManager.cs
snakefoot: I have adopted the school of using method-naming as comments, after seeing so many obsolete comments because methods over time have diverged in functionality. And by default when I need to understand what a method actuel does, then I read the source-code. So I prefer the xml-comment (if any) is a one-liner that explains the intention of the method with little implementation details.

src/NLog/LogLevel.cs
snakefoot: I agree that the current xml-comment `Debug log level.` is not perfect, and more detail could be included. Ex:

- `Debug log level (Ordinal = 1) - Internal events interesting for debugging the behavior of the application`

But like the idea to improve the xml-comment for `class LogLevel` to have a xml-remark-section that lists all public loglevels ordered by severity:

```
- Trace Log Level (Ordinal = 1) - Most verbose level. Example for investigating protocol payloads during development. Almost never enabled in production
- Debug log level (Ordinal = 1) - Internal events interesting for debugging the behavior of the application
- Info log level (Ordinal = 2) - Highlight progress or lifetime events of the application
- Warn log level (Ordinal = 3) - Validation warnings or temporary failures that can be recovered
- Error Log Level (Ordinal = 4) - Exceptions or functionality has failed
- Fatal Log Level (Ordinal = 5) - Most critical level. Application is about to abort
```

src/NLog/Config/LoggingConfiguration.cs
snakefoot: I'm happy that you like the flexibility of the NLog-configuration. And maybe the xml-comments for the `RuleName`-property could be improved to highlight the abilities it provides. But I still prefer that one doesn't try to mention all the stuff that are not assigned by the constructor.

src/NLog/LogManager.cs
snakefoot: One example is #4842 where I changed the internal-logic. Where you nicely highlighted that the xml-comments was completely wrong, so they had to be tweaked #4849 + #4864

snakefoot:
> I know that commenting code and writing documentation often is not a nice job, and it sometimes seems to be useless work. But I can assure you that it is not, and I hope you slightly change your mind towards a more comprehensive documentation.

I volunteered to work as janifor for the NLog-project as an exercise for how to maintain and build a public library. Move fast and break often are not the primary skill, but instead learning the importance of good naming and good documentation, because when it is released, then it is almost "forever".

If having created something nice but with poor documentation and poor naming, then new users will constantly use the wrong door, or find the right door but cannot open it.

But as a janitor, then I cannot take credit for the existing building (and all its quirks). And when it comes to documentation-style, then it is a big topic just like coding-style. Where personal taste comes to play, where I can already see that we have different approaches/styles :)

I like to maintain the style of single-line xml-comments, and I agree with you that there are places where documentation can be improved. Have already done a little with #4849 + #4864

snakefoot:
> I may split up the pull request in 3 parts. Please tell me first whether you would accept the PR with the XML comments,

I think we have only reached the part where we both agree that the xml-documentation can be improved, and we have different suggestions for how it should be solved. But have not yet agreed upon it.

Regarding `SuspendLogging(bool force)` then I think it introduces technical debt, and if you have a need to logging on/off globally without scope-counting, then you should just use `LogManager.GlobalThreshold` (Flip between `LogLevel.Trace` and `LogLevel.Off`).

src/NLog/Config/LoggingRule.cs
snakefoot: Yes it is "funny" with LogLevel in .NET Framework. Where [System.Diagnostics.Debug](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.debug) is low (removed on release-builds) and [System.Diagnostics.Trace](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.trace) is high (stays on release-build).

But NLog has inherited its [LogLevels from Log4j](https://www.tutorialspoint.com/log4j/log4j_logging_levels.htm), and now with Microsoft Extension Logging [LogLevel.Trace](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loglevel) it is lowest level again.

TobiasKnauss:
Let's make it easy: I have suggested some improvements and functionality, and it's up to you to accept or reject them. We have discussed pros and cons, but I don't see a need for more discussion and I don't want to start an argument about whether the improvements are good or bad, required/useful or not. I see that you have already taken some parts of it and I am glad that you see the need for improvement of documentation. If I have more suggestions, I will contact you again.

snakefoot:
Created #4878

snakefoot:
Closing after #4878 + #4849 + #4864 has been committed.

@TobiasKnauss Thank you for pointing out where NLog XML documentation could be improved. If you find more places then please don't hesitate to tell.

TobiasKnauss:
I was going to remove the bookmark to this issue from my list. Before I did so, I read through it again and noticed that you created separate issues only for the changes related to XML comments. Have you also created new issues for the code changes, e.g. for swapping log levels into the correct order, and adding "obsolete"?

snakefoot:
>  Have you also created new issues for the code changes, e.g. for swapping log levels into the correct order, and adding "obsolete"?

No I was lazy. Was thinking an unit-test should be included with any code-changes to loglevels. And obsoleting those exotic condition-logic-classes is not high on my agenda.

PR #4863, opened 2022-04-02
Support SpecialFolder ApplicationDataDir for internalLogFile when parsing nlog.config

### LINKED ISSUE #4861: internalLogFile Appdatadir on Linux & osx
**NLog version**: 4.7.15

**Platform**: .Net 6.0 

**Current NLog config** only as an example, our config has many targets configured for different purposes

```xml
      autoReload="true"
      internalLogLevel="Info"
      internalLogFile="${specialfolder:folder=ApplicationData}/ApplicationName/logs/internal.log">

  <targets>
    <target xsi:type="File" name="allfile" fileName="${specialfolder:folder=ApplicationData}/ApplicationName/logs/all.log"
            layout="${longdate}|${event-properties:item=EventId_Id}|${uppercase:${level}}|${logger}|${message} ${exception:format=tostring}"
            archiveEvery="Day"
            archiveFileName="${specialfolder:folder=ApplicationData}/ApplicationName/logs/all.{#}.log"
            archiveAboveSize="50857600"
            maxArchiveFiles="7"
            />
  </targets>
</nlog>

```

Hello, we are building an application for multiple operating systems.

Every logfile gets saved in ´´´fileName="${specialfolder:folder=ApplicationData}/ApplicationName/logs/*.log"´´´
but the internalLogFile seems to ignore that variable resolution intentionally (like mentioned in a few older issues).

Regarding to the documentation it is possible to use %appdata%, but this does only apply to windows and not to linux or osx.

Also ${currentdir}, ${basedir}, ${tempdir} or ${processdir} are not really options for us since this would also change the location on all the windows machines that are already running.

Do you see any way that an ${appdatadir} (or similar) could be implemented in the future to enable the same logging behavior within all operating systems?
Thanks!

snakefoot:
Trying to resolve #4861

Introduces support for these symbol-types (both for internalLogFile and normal layouts):
- `${applicationDataDir}`
- `${commonApplicationDataDir}`
- `${localApplicationDataDir}`

### DIFF
--- a/src/NLog/Config/LoggingConfigurationParser.cs
+++ b/src/NLog/Config/LoggingConfigurationParser.cs
@@ -254,6 +254,14 @@ namespace NLog.Config
                 if (ContainsSubStringIgnoreCase(internalLogFile, "${processdir}", out string processDirToken))
                     internalLogFile = internalLogFile.Replace(processDirToken, System.IO.Path.GetDirectoryName(LogFactory.CurrentAppEnvironment.CurrentProcessFilePath) + System.IO.Path.DirectorySeparatorChar.ToString());
 #endif
+#if !NETSTANDARD1_3 && !NETSTANDARD1_5
+                if (ContainsSubStringIgnoreCase(internalLogFile, "${applicationDataDir}", out string appDataDirToken))
+                    internalLogFile = internalLogFile.Replace(appDataDirToken, Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData) + System.IO.Path.DirectorySeparatorChar.ToString());
+                if (ContainsSubStringIgnoreCase(internalLogFile, "${commonApplicationDataDir}", out string commonAppDataDirToken))
+                    internalLogFile = internalLogFile.Replace(commonAppDataDirToken, Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData) + System.IO.Path.DirectorySeparatorChar.ToString());
+                if (ContainsSubStringIgnoreCase(internalLogFile, "${localApplicationDataDir}", out string localapplicationdatadir))
+                    internalLogFile = internalLogFile.Replace(localapplicationdatadir, Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData) + System.IO.Path.DirectorySeparatorChar.ToString());
+#endif
                 if (internalLogFile.IndexOf('%') >= 0)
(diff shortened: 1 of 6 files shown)


304NotModified (approved):
Those are long names, but I think those are the best options and it's consistent. So LGTM :)

PS I'm on holiday with limited internet connection.

304NotModified:
Maybe we should document the layout renderer with dashes on https://nlog-project.org/config/?tab=layout-renderers ?

snakefoot:
> Maybe we should document the layout renderer with dashes on https://nlog-project.org/config/?tab=layout-renderers ?

Think it will be confusing with the very simple layout-support for `internalLogFile`-option.

snakefoot:
Decided that ApplicationDataDir was too generic. One might think it would be the directory of the running application. Created #4868

snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/Internal-Logging

cedd82:
> Updated wiki: https://github.com/NLog/NLog/wiki/Internal-Logging

Can the internal-logging wiki be updated to specify where the locations re what was posted above? but also for linux, i cant find what these resolve to in linux online 

${commonApplicationDataDir} - C:\ProgramData
${userApplicationDataDir} - C:\Users\%USERNAME%\AppData\Roaming
${userLocalApplicationDataDir} - C:\Users\%USERNAME%\AppData\Local

snakefoot:
SpecialFolders on Linux is platform dependent. But maybe this page can help:


https://github.com/NLog/NLog/wiki/Special-Folder-Layout-Renderer



Sent from my Galaxy



From: cedd82 ***@***.***>
Date: 24/10/2022 02:27 (GMT+01:00)
To: NLog/NLog ***@***.***>
Cc: Rolf Kristensen ***@***.***>, State change ***@***.***>
Subject: Re: [NLog/NLog] Support SpecialFolder UserApplicationDataDir for internalLogFile when parsing nlog.config (PR #4863)


Updated wiki: https://github.com/NLog/NLog/wiki/Internal-Logging

Can the internal-logging wiki be updated to specify where the locations re what was posted above? but also for linux, i cant find what these resolve to in linux online

${commonApplicationDataDir} - C:\ProgramData
${userApplicationDataDir} - C:\Users%USERNAME%\AppData\Roaming
${userLocalApplicationDataDir} - C:\Users%USERNAME%\AppData\Local

PR #4872, opened 2022-04-08
Changed net35 to net35-client for better fallback from net40-client

snakefoot:
See also  #4223


snakefoot:
Guess client-profile has become obsolete with VS2022, oh well

PR #4891, opened 2022-04-25
Log4JXmlEventLayoutRenderer - IncludeEventProperties default = true

snakefoot:
Better defaults for Log4J-XML where event-properties are included by default in log4net.

But exclude NLogViewer-target since it is special, and has its own standards.


304NotModified (approved):

snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/Log4JXmlEventLayout

Updated wiki: https://github.com/NLog/NLog/wiki/Log4JXMLEvent-Layout-Renderer

PR #4911, opened 2022-05-10
Use the overloading mechanism instead of the optional parameters.

Orace:


snakefoot:
Must admit I like using default-parameters, so do not fully agree with Sonar (I like that one can see what default-value that will be used when not provided). But anyway I'm just the janitor.

KeithLRobertson:
@Orace  What problem does this solve?

snakefoot:
Closing pull-request is it has unresolved conflicts, and it also introduces breaking changes that must wait for NLog 6.0

PR #4924, opened 2022-05-14
ChainsawTarget - IncludeEventProperties = true since IncludeNLogData = false

snakefoot:
Better defaults. Followup to #4891


snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/ChainSaw-target

PR #4934, opened 2022-05-22
SandCastle Docs should always use nuget-packages

snakefoot:
- Changed from building local NLog.dlls to always using dll_to_doc-nuget-project.
- Removed advice about `Install [Sandcastle Help File Builder]` since `NLog.shfbproj` will download it automatically


304NotModified:
Many Thanks!

304NotModified:
updated!

https://nlog-project.org/documentation/v5.0.0/

PR #4935, opened 2022-05-22
DLL to Docs - in csproj SDK

304NotModified:
I don't like old csproj files :P

This also works the wildcards are nice :)

See also: https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#generatepathproperty

Also I think we need a rebuild instead of build - otherwise nice stuff is missing


304NotModified:
PS: noticed this for the first time: https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#restoring-and-building-with-one-msbuild-command

snakefoot (approved):
If BuildDoc.cmd also works with the new project format, then fine with me.

304NotModified:
Yes it works. I already updated the wiki (with that script) for nlog.mailkit v5

snakefoot:
Guess one could move the doc-project out of the src-folder and into the tools-folder.

Since it no longer depends on the NLog-src, but only the relevant nuget-packages.

304NotModified:
good idea :)

snakefoot:
Created #4942

PR #4946, opened 2022-05-29
Replaced netcoreapp2.1 with netcoreapp3.1 to fix obsolete test warnings + Fix vulnerable packages in test projects

304NotModified:


snakefoot:
I don't want that. I want new xunit v3 or xunit v2

304NotModified:
Yeah I also don't like vulnerable packages or .NET Core 2.1 which is EOL.

And I guess xUnit 3 will take some time?

snakefoot:
I'm with leaving it as it is, until xUnit gets v2 fixed, or release v3.

304NotModified:
But we still have the vulnerable packages with .NET Core 2.1? There are only not recognized.

snakefoot:
I'm only worried about the actual nuget-packages. I do not care about any attack vector from NLog-UnitTest-projects.

snakefoot:
Discovered one just had to rollback to xUnit 2.4.0 (Instead of 2.4.1), and then everything was good.

304NotModified:
Nice work!

PR #4965, opened 2022-06-10
Max StackTraceUsage should be computed using bitwise OR

martinzding:
The `StackTraceUsage` of a layout is currently using `Max` to compute the maximum value. This is buggy when individual `StackTraceUsage` values have overlapping bits.


snakefoot:
@martinzding Thank you for spotting this.

snakefoot:
NLog 5.0.1 is now available: https://www.nuget.org/packages/NLog/5.0.1

PR #4981, opened 2022-06-20
AssemblyVersionLayoutRenderer - Support override of Default value

snakefoot:
When specifying Default-value, then it will only attempt to load assembly-version once.


snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/AssemblyVersion-Layout-Renderer

PR #5068, opened 2022-09-24
integrate and refactor ThrowIfNull- fixes #5055

### LINKED ISSUE #5055: Apply ArgumentNullException.ThrowIfNull to NLog source-code
Has been introduced with NET6, but it should be "easy" to implement equal solution in NLog.Internal:

From:

```c#
            if (name is null)
                throw new ArgumentNullException(nameof(name), "Name of logger cannot be null");
```

To

```c#
            ArgumentNullException.ThrowIfNull(loggerName);
```

https://docs.microsoft.com/en-us/dotnet/api/system.argumentnullexception.throwifnull?view=net-6.0

Ex.

```c#
static internal class ArgumentNullException
{
    public static T ThrowIfNull<T>(T argument, [CallerArgumentExpression("argument")] string paramName = null)
        where T : class
    {
        if (argument is null)
        {
            throw new ArgumentNullException(paramName);
        }

        return argument;
    }
}
```

r-52:
This PR introduces a backward compatible way to use `ThrowIfNull` to check if the given argument is null and
throw an exception or simply return the value of the argument. 

This is a draft PR to get initial feedback. It implements a backward compatible attribute that's used on .NET 5 and lower. The checks are an example how it could be used in the Logger.cs class.

An idea that came: maybe the `Guard` class could be extended to also integrate more checks like `ThrowIfNullOrEmpty` for e.g. strings?

**R=**@snakefoot?
**Refs**: #5055

### DIFF
--- a/src/NLog/Internal/Guard.cs
+++ b/src/NLog/Internal/Guard.cs
@@ -1,0 +1,69 @@
+//
+// Copyright (c) 2004-2021 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
+//
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions
+// are met:
+//
+// * Redistributions of source code must retain the above copyright notice,
+//   this list of conditions and the following disclaimer.
+//
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution.
+//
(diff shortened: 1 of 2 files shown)


snakefoot:
Think it looks good. Just need a single unit-test that verifies that `ArgumentNullException` thrown includes the expected input-parameter-name.

snakefoot:
Doesn't sound like a bad idea to also include `ThrowIfNullOrEmpty` for string-values. But maybe for a different pull-request?

eli-entelis:
@r-52 , is there a reason the file is called guard? shouldn't it be something like NullExecptionHandler?

PR #5080, opened 2022-09-30
🚧 Fix ReplaceNewLinesLayoutRendererWrapper handling of new lines in the replacement string. Fixes 5079.

### LINKED ISSUE #5079: ReplaceNewLinesLayoutRendererWrapper fail if replacement contains new line char
**NLog version**: 5.0.4
**Platform**: all

My goal is to align multi line error message.

For example this:
```
2022-09-30 12:57:49.0158|INFO |EMITTER | MESSAGE LINE 1
MESSAGE LINE 2
```

Became this:
```
2022-09-30 12:57:49.0158|INFO |EMITTER | MESSAGE LINE 1
                                       | MESSAGE LINE 2
```

The [replace-newlines](https://github.com/NLog/NLog/wiki/Replace-NewLines-Layout-Renderer) renderer is a good candidate.

I tried:
```xml
<variable name="pretty" value="${replace-newlines:replacement=\r\n                                       | :${layout}}" />
```

- What is the current result?

The replacement method is called twice, hence, inserted newlines are re-replaced.
https://github.com/NLog/NLog/blob/6f246d36ee4b796bed15061d8cae9deb4fdc8f86/src/NLog/LayoutRenderers/Wrappers/ReplaceNewLinesLayoutRendererWrapper.cs#L68-L70

The actual result is:
```
2022-09-30 12:57:49.0158|INFO |EMITTER | MESSAGE LINE 1

                                       |                                        | MESSAGE LINE 2
```

The replacement method should be called once (at least with the replacement string).

- A workaround is to replace Windows new lines to Unix new lines, then replace all new lines by the replacement string:

```C#
string str = builder.ToString(orgLength, builder.Length - orgLength)
                    .Replace(WindowsNewLine, UnixNewLine)
                    .Replace(UnixNewLine, Replacement);
```

Orace:
A fix for #5079, ensure all newlines char are Unix before actually replace them.
_The proof that using any other intermediate char is at least inefficient is let to the reader._

### DIFF
--- a/src/NLog/LayoutRenderers/Wrappers/ReplaceNewLinesLayoutRendererWrapper.cs
+++ b/src/NLog/LayoutRenderers/Wrappers/ReplaceNewLinesLayoutRendererWrapper.cs
@@ -65,8 +65,12 @@ namespace NLog.LayoutRenderers.Wrappers
                 var containsNewLines = builder.IndexOf('\n', orgLength) >= 0;
                 if (containsNewLines)
                 {
-                    builder.Replace(WindowsNewLine, UnixNewLine, orgLength, builder.Length - orgLength);
-                    builder.Replace(UnixNewLine, Replacement, orgLength, builder.Length - orgLength);
+                    string str = builder.ToString(orgLength, builder.Length - orgLength)
+                                        .Replace(WindowsNewLine, UnixNewLine)
+                                        .Replace(UnixNewLine, Replacement);
+
+                    builder.Length = orgLength;
+                    builder.Append(str);
                 }
             }
         }
--- a/src/NLog/LayoutRenderers/Wrappers/ReplaceNewLinesLayoutRendererWrapper.cs
+++ b/src/NLog/LayoutRenderers/Wrappers/ReplaceNewLinesLayoutRendererWrapper.cs
@@ -65,12 +65,8 @@ namespace NLog.LayoutRenderers.Wrappers
                 var containsNewLines = builder.IndexOf('\n', orgLength) >= 0;
                 if (containsNewLines)
                 {
-                    string str = builder.ToString(orgLength, builder.Length - orgLength)
-                                        .Replace(WindowsNewLine, UnixNewLine)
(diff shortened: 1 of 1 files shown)


Orace:
We should use [StringBuilder.Replace](https://learn.microsoft.com/en-us/dotnet/api/system.text.stringbuilder.replace?view=net-6.0#system-text-stringbuilder-replace(system-string-system-string-system-int32-system-int32)) for better performances.

Orace:
Looks like I was wrong 🤔

Orace:
[Help !](https://youtube.com/clip/UgkxP83XT4OaIMZaZJ8DK5IkZ10oerQxQyC-) looks like the failed tests are unrelated. Should we go back to `StringBuilder.Replace`?

snakefoot:
Thank you for the pull-request could you make the logic like this:

```c#
public string Replacement
{
      get => _replacement;
      set
      {
            _replacement = value;
            _replacementWithNewLine = value?.IndexOf('\n') >= 0;
      }
}
private string _replacement;
private bool _replacementWithNewLine 

if (containsNewLines)
{
       string str = builder.ToString(orgLength, builder.Length - orgLength);
       if (_replacementWithNewLine)
              str = str.Replace(WindowsNewLine, UnixNewLine).Replace(UnixNewLine, Replacement);  // Always 2 string-allocations
       else
              str = str.Replace(WindowsNewLine, Replacement).Replace(UnixNewLine, Replacement);  // Avoid 2 string-allocations
}
```

Orace:
@snakefoot is there a good reason to avoid StringBuilder.Replace that doesn't need any string allocation?

snakefoot:
`StringBuilder.Replace` allocates, and is slow.

Orace:
@snakefoot, since we works with strings, we can use a static compiled regex (like `\r?\n`). What do you think?

snakefoot:
RegEx is slower than string.Replace

snakefoot:
Superseeded by #5090

PR #5085, opened 2022-10-22
Introduced TriLetter as LevelFormat

snakefoot:
With following mapping:

- Trace = Trc
- Debug = Dbg
- Info = Inf
- Warn = Wrn
- Error = Err
- Fatal = Ftl


snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/Level-layout-renderer

PR #5101, opened 2022-11-08
Add LogEventDropped handler to the NetworkTarget (#5084)

### LINKED ISSUE #5084: Introduce event handler LogEventDropped for NetworkTarget
I'm using NetworkTarget, and there is no way to have handle backpressure in logger. There is `NetworkTargetQueueOverflowAction` - `Grow` and `Block` will break your applications, so only reasonable selection is `Discard`.

While using `Discard` there is no way to have a notification that my application is discarding logs. I see that I'm dropping lots of log entries on production, but I cannot setup monitoring for that. Only hack I've found out is this:
```
InternalLogger.LogLevel = LogLevel.Debug;
InternalLogger.LogMessageReceived += (sender, eventArgs) =>
{
  if (eventArgs.Message == "NetworkQueue - Discarding single item, because queue is full")
  {
    // Log dropped
  }
};
```
but this solution is flimsy (depends on log message), and need debug logs configured in InternalLogger.

**NLog version**: 5.0.0

**Platform**: .NET Core / AllPlatforms

Not revelant

 - Why do we need it?

Having such notification is very useful, because then I could send a metric to my statsd collector, setup alerts and notifications when my apps are dropping logs.
In current setup I can only compare logs to traffic levels 

 - An example of the XML config, if relevant.

There are no options in config for changing this, so probably not revelant.

ShadowDancer:
Add LogEventDropped handler to the NetworkTarget (#5084)

As proposed in the ticket copied implementation from `AsyncTargetWrapper`

Closes #5084


snakefoot:
@ShadowDancer Thank you for the contribution, and finding all the locations where LogEvents are dropped. Excellent work.

Any reason why the dropped-EventArgs is different from AsyncTargetWrapper ? (Not including the LogEvent)

I recognize that it will give better performance, but curious if the inconsistency is good?

POnakS:
@snakefoot QueuedNetworkSender operates on buffers of bytes, it does not have concept of LogEvent (it operates on buffers instead). I've taken simple approach and unified event across all call sites.

I think LogEvent does not matter for this use case, as I want to know if there is backpressure on logshipper - in that case you probably cannot do anything with these logs anyway.

src/NLog/Internal/NetworkSenders/QueuedNetworkSender.cs
```diff
@@ -78,6 +78,7 @@ protected QueuedNetworkSender(string url)
         public int MaxQueueSize { get; set; }
 
         public NetworkTargetQueueOverflowAction OnQueueOverflow { get; set; }
+        public event EventHandler<LogEventDroppedEventArgs> LogEventDropped;
```
snakefoot, line 81: Using `EventHandler` is dangerous, as you have to reset/unhook them (assign to `null`) or else you will experience "memory leak".

src/NLog/Targets/NetworkTarget.cs
```diff
@@ -567,6 +580,12 @@ private NetworkSender CreateNetworkSender(string address)
         {
             var sender = SenderFactory.Create(address, MaxQueueSize, OnQueueOverflow, MaxMessageSize, SslProtocols, TimeSpan.FromSeconds(KeepAliveTimeSeconds));
             sender.Initialize();
+            
+            if(sender is QueuedNetworkSender queuedNetworkSender)
```
snakefoot, line 584: Would prefer a solution without casting

src/NLog/Internal/NetworkSenders/QueuedNetworkSender.cs
ShadowDancer: From my analysis there is no way for `QueuedNetworkSender`(publisher) to outlive `NetworkTarget`(subscriber), so I can rely on garbage collector to clean up memory.

src/NLog/Targets/NetworkTarget.cs
ShadowDancer: I've updated `SenderFactory` (internal interface) to return `QueuedNetworkSender` instead of `NetworkSender`. This way cast is not required.

It does not have sense to have network sender without buffer, so this is good change which pulls relevant details into the interface.

src/NLog/Targets/NetworkTarget.cs
```diff
@@ -567,6 +580,8 @@ private NetworkSender CreateNetworkSender(string address)
         {
             var sender = SenderFactory.Create(address, MaxQueueSize, OnQueueOverflow, MaxMessageSize, SslProtocols, TimeSpan.FromSeconds(KeepAliveTimeSeconds));
             sender.Initialize();
+            sender.LogEventDropped += OnLogEventDropped;
```
snakefoot, line 583: Can you change so you only attach to `sender.LogEventDropped` when something is observing `NetworkTarget.LogEventDropped`?

src/NLog/Targets/LogEventDroppedEventArgs.cs
```diff
@@ -0,0 +1,49 @@
...
+{
+    /// <summary>
+    /// Arguments for <see cref="NetworkTarget.LogEventDropped"/> events.
+    /// </summary>
+    public class LogEventDroppedEventArgs : EventArgs
+    {
+        /// <summary>
+        /// Provides a value to use with events that do not have event data.
+        /// </summary>
+        public static new readonly LogEventDroppedEventArgs Empty = new LogEventDroppedEventArgs();
```
snakefoot, line 46: Can this become `internal` ?

src/NLog/Internal/NetworkSenders/QueuedNetworkSender.cs
snakefoot: Seems the garbage collector is happily clearning up.

src/NLog/Targets/LogEventDroppedEventArgs.cs
snakefoot, line 46: Would it make sense to have an enum-type to signal the reason for dropping the LogEvent ? (Ex. PayloadSizeOverflow, QueueOverflow, ConnectionOverflow). Then have static-instance for the different reasons)

src/NLog/Targets/NetworkTarget.cs
snakefoot: Like the change. Think it is an improvement.

src/NLog/Targets/LogEventDroppedEventArgs.cs
ShadowDancer: Added enum with reason

ShadowDancer: Done

src/NLog/Targets/NetworkTarget.cs
ShadowDancer: I think it does not make sense, as it saves few bytes on the object which is not created too often (even if you do not cache connections establishing a new one is much heavier operation than 4 bytes).

On the other had it makes the code much more complex, because if you send a log, and then attach handler it won't count dropped logs correctly anymore. You'd have to resubscribe to all cached NetworkSenders, and there are concurrency issues related to that. I think it's much better to leave it as it is.

snakefoot (approved):
I think it looks good. Any last changes before merging?

POnakS:
Not really, go ahead.

PR #5177, opened 2023-03-25
MailTarget - Added support for email message headers

### LINKED ISSUE #5161: Mail target - send to recipient mail server - add message-id header [Feature reqeust]
Hi! Thanks for reporting this feature!

**NLog version**: 4.7.9

**Platform**: .Net 4.5

Currently to send email using a gmail account, we must include the username and password in the config file.

If we are only sending to a gmail user (or other account using google workspace) we could avoid putting the credentials in the config by sending the email directly to the google incoming mx server aspmx.l.google.com, the only problem currently is that nlog does not generate a message-id header and google rejects the email (unless coming from a trusted ip).

If there was an option to indicate that we are sending to the destination server instead of a mta and have nlog generate a message-id it would allow us to deploy nlog without exposing gmail credentials.

snakefoot:
Resolves #5161 (Will also have to update MailKit-Target to re-align)
```xml
<target xsi:type="Mail" name="MailTarget" />
     <mailHeader name="Hello" layout="World" />
</target>
```
Decided on `<mailheader>` since `TargetWithLayoutHeaderAndFooter` already has a concept of `<header>` + `<footer>` for the actual message-body.

### DIFF
--- a/src/NLog/Targets/MailTarget.cs
+++ b/src/NLog/Targets/MailTarget.cs
@@ -308,6 +308,13 @@ namespace NLog.Targets
         /// <docgen category='SMTP Options' order='100' />
         public Layout<int> Timeout { get; set; } = 10000;
 
+        /// <summary>
+        /// Gets the array of email headers that are transmitted with this email message
+        /// </summary>
+        /// <docgen category='Message Options' order='100' />
+        [ArrayParameter(typeof(MethodCallParameter), "mailheader")]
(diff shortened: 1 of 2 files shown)


snakefoot:
Created https://github.com/NLog/NLog.MailKit/pull/154 for NLog.MailKit

snakefoot:
Updated wiki - https://github.com/NLog/NLog/wiki/Mail-target

PR #5200, opened 2023-04-25
JsonLayout - Added IndentJson for pretty format

### LINKED ISSUE #5096: JsonLayout with primitive indented pretty format
Could be nice if JsonLayout could render JSON in indented pretty format:
```
{
  "date": "2018-11-18 14:21:45.0671",
  "level": "INFO",
  "message": "Some message"
}
```
For console output, that should be parsed by humans. By adding new option `IndentJson` (Similar to `IndentXml` for XmlLayout).

Since "primitive" then it is okay with imperfect output when nesting JsonLayout.

See also: https://stackoverflow.com/questions/53361324/nlog-indented-pretty-formatted-json

See also: https://stackoverflow.com/questions/71046880/format-json-object-in-console-or-file-targets-of-nlog

nih0n:
resolves #5096

### DIFF
--- a/src/NLog/Layouts/JSON/JsonLayout.cs
+++ b/src/NLog/Layouts/JSON/JsonLayout.cs
@@ -350,20 +350,17 @@ namespace NLog.Layouts
             if (beginJsonMessage)
             {
                 if (IndentJson)
-                    sb.Append('{');
+                    sb.Append('{').AppendLine().Append(' ', SpacesPerIndent);
                 else
                     sb.Append(SuppressSpaces ? "{" : "{ ");
             }
             else
             {
                 if (IndentJson)
-                    sb.Append(',');
+                    sb.Append(',').AppendLine().Append(' ', SpacesPerIndent);
                 else
                     sb.Append(SuppressSpaces ? "," : ", ");
             }
-
-            if (IndentJson)
-                sb.AppendLine().Append(' ', SpacesPerIndent);
 
             sb.Append('"');
 
--- a/src/NLog/Layouts/JSON/JsonLayout.cs
+++ b/src/NLog/Layouts/JSON/JsonLayout.cs
@@ -350,19 +350,17 @@ namespace NLog.Layouts
             if (beginJsonMessage)
             {
                 if (IndentJson)
-                    sb.Append('{').AppendLine().Append(' ', SpacesPerIndent);
+                    sb.Append('{').AppendLine().Append(' ', SpacesPerIndent).Append('"');
                 else
-                    sb.Append(SuppressSpaces ? "{" : "{ ");
+                    sb.Append(SuppressSpaces ? "{\"" : "{ \"");
             }
             else
             {
                 if (IndentJson)
-                    sb.Append(',').AppendLine().Append(' ', SpacesPerIndent);
+                    sb.Append(',').AppendLine().Append(' ', SpacesPerIndent).Append('"');
                 else
-                    sb.Append(SuppressSpaces ? "," : ", ");
+                    sb.Append(SuppressSpaces ? ",\"" : ", \"");
             }
-
-            sb.Append('"');
 
             if (ensureStringEscape)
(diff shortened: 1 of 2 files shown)


nih0n:
It looks like the code smells are from methods unrelated to this pull request.

tests/NLog.UnitTests/Layouts/JsonLayoutTests.cs
```diff
@@ -68,6 +68,30 @@ public void JsonLayoutRendering()
...
+            };
+
+            var logEventInfo = new LogEventInfo
+            {
+                TimeStamp = new DateTime(2010, 01, 01, 12, 34, 56),
+                Level = LogLevel.Info,
+                Message = "hello, world"
+            };
+
+            Assert.Equal("{\r\n\t\"date\": \"2010-01-01 12:34:56.0000\",\r\n\t\"level\": \"Info\",\r\n\t\"message\": \"hello, world\"\r\n}", jsonLayout.Render(logEventInfo));
```
snakefoot, line 93: To make the unit-test work on non-Windows, then replace `\r\n` with `Environment.Newline`.

src/NLog/Layouts/JSON/JsonLayout.cs
```diff
@@ -341,13 +347,18 @@ private void BeginJsonProperty(StringBuilder sb, string propName, bool beginJson
         {
             if (beginJsonMessage)
             {
-                sb.Append(SuppressSpaces ? "{\"" : "{ \"");
+                sb.Append((IndentJson || SuppressSpaces) ? "{" : "{ ");
             }
             else
             {
-                sb.Append(SuppressSpaces ? ",\"" : ", \"");
+                sb.Append((IndentJson || SuppressSpaces) ? "," : ", ");
```
snakefoot, line 354: Do not think you should check `IdentJson` here also, since `SuppressSpaces` is false by default.

src/NLog/Layouts/JSON/JsonLayout.cs
```diff
@@ -341,13 +347,18 @@ private void BeginJsonProperty(StringBuilder sb, string propName, bool beginJson
...
             }
 
+            if (IndentJson)
+                sb.AppendLine().Append("\t");
```
snakefoot, line 358: Make this a single character `sb.AppendLine().Append('\t');`, instead of string.

src/NLog/Layouts/JSON/JsonLayout.cs
```diff
@@ -358,7 +369,10 @@ private void BeginJsonProperty(StringBuilder sb, string propName, bool beginJson
 
         private void CompleteJsonMessage(StringBuilder sb)
         {
-            sb.Append(SuppressSpaces ? "}" : " }");
+            if (IndentJson)
+                sb.AppendLine();
+            
+            sb.Append((IndentJson || SuppressSpaces) ? "}" : " }");
```
snakefoot, line 375: Do not think you should check IdentJson here also, since SuppressSpaces is false by default.

snakefoot:
> [nih0n](https://github.com/NLog/NLog/pull/5200#issuecomment-1521724407) It looks like the code smells are from methods unrelated to this pull request.

Yes the 2 code-smells can be ignored.

snakefoot:
Would it be possible to change the indent from 1-tab into [2 spaces like Utf8JsonWriter](https://github.com/dotnet/runtime/blob/0c612cf7c36f53651e646d62d3bb362823880c00/src/libraries/System.Text.Json/src/System/Text/Json/JsonConstants.cs#L50) ?

snakefoot:
Updated the PullRequest to reduce the number of calls to StringBuilder.Append() to improve performance (When IndentJson = false)

snakefoot:
NLog v5.1.4 has been released - https://www.nuget.org/packages/NLog/5.1.4

snakefoot:
Updated wiki - https://github.com/NLog/NLog/wiki/JsonLayout

PR #5216, opened 2023-05-11
Refactor - Apply ThrowIfNull and ThrowIfNullOrEmpty

### LINKED ISSUE #5055: Apply ArgumentNullException.ThrowIfNull to NLog source-code

nih0n:
Resolves #5055 

# Changes

Added `Guard` class with two methods
```csharp
static T ThrowIfNull<T>(T arg, [CallerArgumentExpression("arg")] string param = "")
static string ThrowIfNullOrEmpty(string arg, [CallerArgumentExpression("arg")] string param = "")
```
and replaced all occurrences of

```csharp
if (arg is null)
    throw new ArgumentNullException(nameof(arg));

// and

if (string.IsNullOrEmpty(arg))
    throw new ArgumentNullException(nameof(arg));
```

# Remarks

This is the only line that we use the `message` parameter of `ArgumentNullException` so I didn't wanted to include it in the `Guard` methods
https://github.com/NLog/NLog/blob/589ee8b1398665c1810660c5434d0226b0ccf09f/src/NLog/LogFactory.cs#L1074

This check is in a different assembly so I didn't added the refence to the `Guard`. I don't know if it would be ok.
https://github.com/NLog/NLog/blob/589ee8b1398665c1810660c5434d0226b0ccf09f/tests/PackageLoaderTestAssembly/NLogPackageLoaders.cs#L74

### DIFF
--- a/src/NLog/Config/ConfigurationItemFactory.cs
+++ b/src/NLog/Config/ConfigurationItemFactory.cs
@@ -83,7 +83,7 @@ namespace NLog.Config
         internal ConfigurationItemFactory(ServiceRepository serviceRepository, ConfigurationItemFactory globalDefaultFactory, params Assembly[] assemblies)
         {
             CreateInstance = FactoryHelper.CreateInstance;
-            _serviceRepository = serviceRepository ?? throw new ArgumentNullException(nameof(serviceRepository));
+            _serviceRepository = Guard.ThrowIfNull(serviceRepository);
             _targets = new Factory<Target, TargetAttribute>(this, globalDefaultFactory?._targets);
             _filters = new Factory<Filter, FilterAttribute>(this, globalDefaultFactory?._filters);
             _layoutRenderers = new LayoutRendererFactory(this, globalDefaultFactory?._layoutRenderers);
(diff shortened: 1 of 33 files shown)


snakefoot:
Thank you so much, and again sorry about the trouble.

It seems like the obsolete-adventure for NLog v5.2 is almost done. Just need to make a T4-template to replace NLog-Assembly-Scanning.

snakefoot:
Finally able to start preparing the NLog v5.2 release with support for NET Trimming.

PR #5266, opened 2023-06-26
Remove irrelevant StructuredMessageTemplateAttribute

saltukkos:
I found that some overloads in `ILogger` have the `[StructuredMessageTemplate]` but they don't actually treat the input as a structured message, instead treating it like a regular string. This leads to false-positives in R# and irrelevant syntax highlighting. I believe these attributes shouldn't be here, as they are only applied to the `Debug` level ([search link](https://github.com/search?q=repo%3ANLog%2FNLog%20%2F%5C%5BStructuredMessageTemplate%5C%5D%5B%5E%2C%5D*%5C%29%2F%20&type=code)) (and `Fatal` for obsolete version of the method). Note that these methods also don't have `[MessageTemplateFormatMethod]`.

Screenshot 2023-06-26 at 11 44 07


snakefoot (approved):
Thank you for reporting this, and resolving it with this pull-request.

snakefoot:
NLog v5.2.1 has now been released - https://www.nuget.org/packages/NLog

Thank you again for taking the time to report and fix this issues.

PR #5281, opened 2023-07-10
Refactor ConditionBasedFilter when filter to use ternary operator

jokoyoski:
In the refactored code, we directly compare the result of Condition.Evaluate(logEvent) with ConditionExpression.BoxedTrue to determine if the condition is true. Instead of assigning the result to val, we directly use the comparison result to determine the path of execution. This simplifies the code by removing the need for the if statement and condenses it into a single return statement.


snakefoot:
Have now fixed the Azure DevOps-pipeline with #5282, so if you rebase to dev-branch then they should become green.

jokoyoski:
Okay will do that and made the update you advice. Thanks

jokoyoski:
@snakefoot  please check

snakefoot (approved):
Looks good

PR #5297, opened 2023-07-31
Replaced MutableUnsafeAttribute with ThreadAgnosticImmutableAttribute

snakefoot:
- [x] Waiting for NLog v5.3

I don't like the word "unsafe", and since the attribute is just a restriction of the original `ThreadAgnosticAttribute`, then I think `[ThreadAgnosticImmutable]` is better (Only ThreadAgnostic when LogEvent-state is immutable)

### DIFF
--- a/src/NLog/Config/ThreadAgnosticImmutableAttribute.cs
+++ b/src/NLog/Config/ThreadAgnosticImmutableAttribute.cs
@@ -1,4 +1,4 @@
-// 
+﻿// 
 // Copyright (c) 2004-2021 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
 // 
 // All rights reserved.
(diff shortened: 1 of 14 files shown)


snakefoot:
It could be interesting if the layout precalculate was a little smarter. Right now if having a JsonLayout that includes ThreadId, then the entire JsonLayout must be precalculated by the application-thread, and cannot be deferred to async-background-thread.

It could be neat if JsonLayout Precalculate was to recognize that less than 5 SimpleLayout needed to perfrom Thread-context-capture, then instead of rendering the entire Json-Document on the application-thread. Then it would "just" precalculate those 5 SimpleLayout-results.

It could also be interesting if the JsonLayout was able to recognize whether it should precalculate, when it detects that it must IncludeEventProperties = true, but only if the LogEventInfo-properties are "volatile" / "not-immutable". This should be combined with the coming optimization where NLog.Extensions.Logging doesn't provider format-Parameters-array, but only the LogEvent-properties-collection.

PR #5387, opened 2023-10-27
Implemented dynamic archive directory creation

t4m45:
Dynamic archive was only able to create dynamic file names. Directory names were not affected.

For example, with `archiveFileName="Logs/Archive/archived-{#}/file.log"` the created directory's name was `archived-{#}`.

Based on this issue:
https://stackoverflow.com/questions/77366618/subtract-from-date-in-nlog-configuration/77368855#77368855

Note: I did not implement this in FileArchiveModeDynamicSequence, because I'm not familiar with that mode and it looked different to the others.

### DIFF
--- a/src/NLog/Targets/FileArchiveModes/FileArchiveModeDate.cs
+++ b/src/NLog/Targets/FileArchiveModes/FileArchiveModeDate.cs
@@ -35,6 +35,7 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
+using System.Text.RegularExpressions;
 
 namespace NLog.Targets.FileArchiveModes
 {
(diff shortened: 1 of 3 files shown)


snakefoot:
Thank you for the pull-request. I have some requests before merging:
- Please make an unit-test that verifies archive-folder is actually created as expected (Add to `FileTargetTests.cs`)
- Please replace the RegEx-usage with `FileNameTemplate` instead (Maybe rename class to `FilePathTemplate`)

If you have issues with building unit-test-project locally. Then just update the `NLog.UnitTests.csproj` to have `<TargetFrameworks>net461</TargetFrameworks>` (Removing `net6.0`)

snakefoot:
Closing due to inactivity.

PR #5443, opened 2023-12-21
InternalLogger - Obsoleted API to reduce dependencies

snakefoot:
- [x] Waiting for NLog v5.3

Just like LogFactory should be platform-independent (Not require console or filesystem), then the same goes for InternalLogger.

Seems it is enough to just disconnect InternalLogger from the actual console- and file-writing (changed to subscription-logic). So now it is only when directly using `LogToConsole`- or `LogFile`-property, that the penalty comes.

### DIFF
--- a/src/NLog/Common/InternalEventOccurredHandler.cs
+++ b/src/NLog/Common/InternalEventOccurredHandler.cs
@@ -33,54 +33,11 @@
 
 namespace NLog.Common
 {
-    using System;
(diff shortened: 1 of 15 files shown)


snakefoot:
This might not be the best name, as some might think this will be triggered for all LogEvents produced by the application:
```
public static event InternalLogEventPublishedHandler LogEventPublished;
```

PR #5463, opened 2024-01-19
LogManager GetCurrentClassLogger fallback to assembly-name when no namespace

snakefoot:
- [ ] Waiting for NLog v5.3 (Since breaking behavior change)

To improve this use-case with NET6/NET7/NET8 (not specifying any namespace):
```
using NLog;

var logger = LogManager.GetCurrentClassLogger();
```
Fallback to Assembly-Name when caller-method is missing namespace.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,7 +1,7 @@
 version: 5.0.0-{build} # Only change for mayor versions (e.g. 6.0)
 image:
   - Visual Studio 2022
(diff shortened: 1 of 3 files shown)


snakefoot:
Curious if it is better just to dive into `Assembly.FullName` (string-split), or using `new AssemblyName(assembly.FullName).Name`

PR #5480, opened 2024-03-03
Logger LayoutRenderer extended with PrefixName option

snakefoot:


snakefoot:
Should probably find a better name like "BasePath" or "Prefix" or "FamilyName" or "ParentName", since Logger-name can be anything, and not just class-type-name.

Maybe also introduce "RootName" ?

snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/Logger-Layout-Renderer

Issue #5488, opened 2024-03-27
Can't remove [INFO] [Microsoft.EntityFrameworkCore.Database.Command] Executed DbCommand from console logs


Dmdv:
**NLog version**: (e.g. 5.2.8)

**Platform**: ASP.NET Core 8.0

**Current NLog config**

```xml
<nlog>
  <targets>
        <target xsi:type="ColoredConsole"
                name="lifetimeConsole"
                layout="${date:format=yyyy-MM-dd HH\:mm\:ss zzz} [${level:uppercase=true}] [${logger}] ${message} ${exception:format=tostring}" />

  </targets>
    <rules>
        <logger name="System.Net.Http.*" minlevel="Warn" writeTo="lifetimeConsole" final="true" />
        <logger name="Microsoft.EntityFrameworkCore.Database.Command" level="Warn" writeTo="lifetimeConsole" final="true" />
        <logger name="Microsoft.Hosting.Lifetime" minlevel="Info" writeTo="lifetimeConsole" final="true" />
        <logger name="System.*" minlevel="Warn" writeTo="lifetimeConsole" final="true" />
        <logger name="Microsoft.*" minlevel="Warn" writeTo="lifetimeConsole" final="true">
            <filters>
                <when condition="not starts-with('${logger}','Microsoft.EntityFrameworkCore.ChangeTracking')" action="Log" />
                <when condition="not starts-with('${logger}','Microsoft.EntityFrameworkCore.Database.Command')" action="Log" />
                <when condition="contains('${message}','Executed DbCommand')" action="Ignore" />
            </filters>
        </logger>

        <logger name="*" minlevel="Info" writeTo="logfile,lifetimeConsole">
            <filters>
                <when condition="not starts-with('${logger}','Microsoft.EntityFrameworkCore.ChangeTracking')" action="Log" />
                <when condition="not starts-with('${logger}','Microsoft.EntityFrameworkCore.Database.Command')" action="Log" />
                <when condition="contains('${message}','Executed DbCommand')" action="Ignore" />
            </filters>
        </logger>

    </rules>
</nlog>

```

```
// Configure NLog

builder.Logging.ClearProviders();
builder.Host.UseNLog(new NLogAspNetCoreOptions { RemoveLoggerFactoryFilter = true });
```

I've tried everything, nothing works.

Current result:

```
[...]

PR #5489, opened 2024-03-28
Logging Rule with FinalMinLevel should also support dynamic filters

snakefoot:
Ensure that dynamic filters also applies for logging-rules with FinalMinLevel and no targets:
```xml
<nlog throwConfigExceptions="true" internalLogToConsole="true" internalLogLevel="Debug">
  <targets>
    <target type="ColoredConsole"
            name="lifetimeConsole"
            layout="${date:format=yyyy-MM-dd HH\:mm\:ss zzz} [${level:uppercase=true}] [${logger}] ${message} ${exception:format=tostring}" />

  </targets>
  <rules>
    <logger name="System.*" finalMinLevel="Warn" />
    <logger name="System.Net.Http.*" finalMinLevel="Warn" />
    <logger name="Microsoft.*" finalMinLevel="Warn" />
    <logger name="Microsoft.Hosting.Lifetime" finalMinLevel="Info" />
    <logger name="Microsoft.EntityFrameworkCore.*" finalMinLevel="Info" />
    <logger name="Microsoft.EntityFrameworkCore.Database.Command" finalMinLevel="Info">
      <filters defaultAction="Log">
        <when condition="contains('${message}','Executed DbCommand')" action="IgnoreFinal" />
      </filters>
    </logger>
    <logger name="*" minlevel="Info" writeTo="lifetimeConsole" />
  </rules>
</nlog>
```

See also #5488

### DIFF
--- a/src/NLog/Internal/TargetWithFilterChain.cs
+++ b/src/NLog/Internal/TargetWithFilterChain.cs
@@ -132,12 +132,18 @@ namespace NLog.Internal
 
         static private bool GetTargetsByLevelForLogger(string name, List<LoggingRule> loggingRules, LogLevel globalLogLevel, TargetWithFilterChain[] targetsByLevel, TargetWithFilterChain[] lastTargetsByLevel, bool[] suppressedLevels)
         {
+            IList<KeyValuePair<FilterResult?, IList<Filter>>> finalMinLevelWithFilters = null;
             bool targetsFound = false;
             foreach (LoggingRule rule in loggingRules)
             {
                 if (!rule.NameMatches(name))
                 {
                     continue;
+                }
+
+                if (LoggingRuleHasFinalMinLevelFilters(rule))
+                {
+                    CollectFinalMinLevelFiltersFromRule(rule, ref finalMinLevelWithFilters);
                 }
 
                 targetsFound = AddTargetsFromLoggingRule(rule, name, globalLogLevel, targetsByLevel, lastTargetsByLevel, suppressedLevels) || targetsFound;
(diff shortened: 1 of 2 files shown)


src/NLog/Internal/TargetWithFilterChain.cs
```diff
@@ -152,12 +158,59 @@ static private bool GetTargetsByLevelForLogger(string name, List<LoggingRule> lo
...
+        {
+            return LogLevel.Off != rule.FinalMinLevel && rule.Filters.Count != 0 && rule.Targets.Count == 0;
+        }
+
+        private static void CollectFinalMinLevelFiltersFromRule(LoggingRule rule, ref IList<KeyValuePair<FilterResult?, IList<Filter>>> finalMinLevelWithFilters)
+        {
+            finalMinLevelWithFilters = finalMinLevelWithFilters ?? new KeyValuePair<FilterResult?, IList<Filter>>[LogLevel.MaxLevel.Ordinal + 1];
+            for (int i = 0; i <= LogLevel.MaxLevel.Ordinal; ++i)
+            {
+                if (i < rule.FinalMinLevel.Ordinal)
```
BooTeK, line 189: In other places in the codebase FinalMinLevel is either guarded for null or there are prevention to set it to null but it's default appears to be null. When testing 5.3.1 in our codebase (to update) we encountered a rule with a FinalMinLevel unset (null) that causes a null reference to be thrown here.

snakefoot: Thank you for reporting this. I have been able to reproduce the NullReferenceException by creating a LoggingRule with filters but without targets. Will try get a fix out ASAP.

snakefoot: @BooTeK NLog v5.3.2 has now been released - https://www.nuget.org/packages/NLog/5.3.2

PR #5504, opened 2024-04-06
#5500 add net 6 as it has registry included

### LINKED ISSUE #5500: Add additional TFM to reduce dependencies
### Is your feature request related to a problem? Please describe.
I want to minimise dependencies in my project by utilising framework dependencies wherever possible

### Describe the solution you'd like
Packages should not have any dependency which can be provided by the framework and if necessary additional frameworks should be added.

### Describe alternatives you've considered
Accept the additional dependency

### Additional context
n/a

### packages impacted
- NLog.WindowsRegistry

thompson-tomo:
Adds in net 6 as a TFM as it includes Microsoft.Win32.Registry

Closes #5500

### DIFF
--- a/src/NLog.WindowsRegistry/NLog.WindowsRegistry.csproj
+++ b/src/NLog.WindowsRegistry/NLog.WindowsRegistry.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net35;net46;netstandard2.0;net6.0-WINDOWS</TargetFrameworks>
+    <TargetFrameworks>net35;net46;netstandard2.0;net6.0-windows</TargetFrameworks>
 
     <Title>NLog Windows Registry</Title>
     <Company>NLog</Company>
--- a/src/NLog.WindowsRegistry/NLog.WindowsRegistry.csproj
+++ b/src/NLog.WindowsRegistry/NLog.WindowsRegistry.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net35;net46;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net35;net46;netstandard2.0;net6.0-WINDOWS</TargetFrameworks>
 
     <Title>NLog Windows Registry</Title>
     <Company>NLog</Company>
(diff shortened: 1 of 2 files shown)


snakefoot:
Like how you learned that NET5 is obsolete, and moved to NET6. Notice that NET6 will also become obsolete end-of-year.

Curious if you are actually using NLog.WindowsRegistry-nuget-package ? (Or if you are just trying to align things)

thompson-tomo:
The reason I didn't dotnet 5 is that there was no benefit when compared to using net standard 2.0. However Adding net 6 allowed for the removal of the explicit dependency

thompson-tomo:
Just aligning & striving for optimisations in dependencies.

snakefoot:
Then I'm leaning towards just letting the NLog.WindowsRegistry-nuget-package stay unchanged.

It was only created to leave a bridge for exotic users, so they could upgrade to NLog v5 (as it was removed from "core" NLog-nuget-package).

thompson-tomo:
one interesting thing i have just noticed, is that trimming has been enabled in the csrpoj however trimming is only supported from .net6.

snakefoot:
> trimming has been enabled in the csrpoj however trimming is only supported from .net6.

Well Microsoft has been [working on trimming](https://www.hanselman.com/blog/making-a-tiny-net-core-30-entirely-selfcontained-single-executable) for a long time.

Microsoft decided that with .NET7 then all application will have trimming enabled by default, when published. Thus generating warnings for all Nuget-package not enabled for trimming. At that moment NLog v5.2 was released with the trimming-attribute, to ensure all consumers could move forward without build-warnings from NLog.

Microsoft decided that with .NET8 that the trimming analyzer would only fully work, unless project targeted the NET6-API,. As Microsoft believe they have correctly applied the necessary trimming-attributes for the NET6-API (and forward).

The NLog-build-pipelines are constantly upgraded by the wonderful AppVeyor-team, and uses the latest Microsoft-build-tools. Thus the NLog project is constantly battling the latest bugs/features from Microsoft.

Right now I accept that the build-warning just means that the trimming-analyzer might not have detected all exotic issues. The important part is that NLog works with trimming. And hopefully with NLog v6 then it will also work with AOT without build-warnings.

thompson-tomo:
Thanks, note we could be putting the trim attributes in a conditional property group hence remove those warnings which currently appear.

snakefoot:
Closing since on the road to be deprecated. And compiler warnings are being chased here: #5510

PR #5512, opened 2024-04-27
ExceptionLayoutRenderer - Added support for strange conversion from Exception-class to string-value

snakefoot:


snakefoot:
Scaler string value is no longer a key-value-pair-property, but just a simple value without key-name. Doesn't make sense to include in Properties-output.

PR #5546, opened 2024-07-13
Extend TargetWithContext with GetAllPropertiesList #5543

### LINKED ISSUE #5543: Extend TargetWithContext with sister method to GetAllProperties
Alternative to the existing `GetAllProperties()` that allocates dictionary upfront, but instead returns `IEnumerable<KeyValuePair<string, object>>`

But one have to accept the possibility of getting duplicate property-keys (LogEventInfo.Properties + ScopeContext.Properties + Target.ContextProperties)

The goal is to reduce memory-allocation at pretty much all cost.

smnsht:
Trying to resolve #5543

### DIFF
--- a/src/NLog/Targets/TargetWithContext.cs
+++ b/src/NLog/Targets/TargetWithContext.cs
@@ -286,22 +286,9 @@ namespace NLog.Targets
             private IEnumerable<KeyValuePair<string, object>> _scopeContextProperties;
             private ICollection<string> _gdcNames;
 
-            private bool _CountContextPropertiesCalled;
-            private bool _CountEventPropertiesCalled;
-            private bool _CountScopeContextPropertiesCalled;
-            private bool _CountGdcPropertiesCalled;
-
+            private bool _CountPropertiesCalled;
             private int _propertiesCount;
             private bool _checkExcludeProperties;
-
-            private void DoOnce(Action action, ref bool methodHasBeenCalled)
-            {
-                if (methodHasBeenCalled)
-                    throw new InvalidOperationException("The method should be called only once!");
-
-                action();
-                methodHasBeenCalled = true;
-            }
 
             private KeyValuePair<string, object>? BuildPair(string propertyName, object propertyValue)
             {
@@ -346,52 +333,33 @@ namespace NLog.Targets
                 _logEvent = logEvent;
             }
 
-            public GetAllPropertiesCountContext CountContextProperties()
+            public GetAllPropertiesCountContext EstimateCount()
             {
-                DoOnce(() =>
-                {
-                    _propertiesCount += _targetContext.ContextProperties?.Count ?? 0;
-                }, ref _CountContextPropertiesCalled);
-                return this;
-            }
+                if (_CountPropertiesCalled)
+                    throw new InvalidOperationException("The method should be called only once!");
+                _CountPropertiesCalled = true;
 
-            public GetAllPropertiesCountContext CountEventProperties()
-            {
-                DoOnce(() =>
-                {
-                    if (_targetContext.IncludeEventProperties && _logEvent.HasProperties)
-                    {
-                        _logEventProperties = _logEvent.Properties;
-                        _propertiesCount += _logEventProperties.Count;
-                    }
-                }, ref _CountEventPropertiesCalled);
-                return this;
-            }
+                _propertiesCount += _targetContext.ContextProperties?.Count ?? 0;
 
-            public GetAllPropertiesCountContext CountScopeContextProperties()
-            {
-                DoOnce(() =>
+                if (_targetContext.IncludeEventProperties && _logEvent.HasProperties)
                 {
-                    if (_targetContext.IncludeScopeProperties)
-                    {
-                        // TODO: && !logEvent.TryGetCachedLayoutValue(_contextLayout.ScopeContextPropertiesLayout, out object value)
-                        _scopeContextProperties = ScopeContext.GetAllProperties();
-                        _propertiesCount += _scopeContextProperties.Count();
-                    }
-                }, ref _CountScopeContextPropertiesCalled);
-                return this;
-            }
+                    _logEventProperties = _logEvent.Properties;
+                    _propertiesCount += _logEventProperties.Count;
+                }
 
-            public GetAllPropertiesCountContext CountGdcProperties()
-            {
-                DoOnce(() =>
+                if (_targetContext.IncludeScopeProperties)
                 {
-                    if (_targetContext.IncludeGdc)
-                    {
-                        _gdcNames = GlobalDiagnosticsContext.GetNames();
-                        _propertiesCount += _gdcNames.Count;
-                    }
-                }, ref _CountGdcPropertiesCalled);
+                    // TODO: && !logEvent.TryGetCachedLayoutValue(_contextLayout.ScopeContextPropertiesLayout, out object value)
+                    _scopeContextProperties = ScopeContext.GetAllProperties();
+                    _propertiesCount += _scopeContextProperties.Count();
+                }
+
+                if (_targetContext.IncludeGdc)
+                {
+                    _gdcNames = GlobalDiagnosticsContext.GetNames();
+                    _propertiesCount += _gdcNames.Count;
+                }
+
                 return this;
             }
 
@@ -467,10 +435,7 @@ namespace NLog.Targets
         protected IEnumerable<KeyValuePair<string, object>> GetAllPropertiesList(LogEventInfo logEvent)
         {
             var ctx = new GetAllPropertiesCountContext(this, logEvent)
-                .CountContextProperties()
-                .CountEventProperties()
-                .CountScopeContextProperties()
-                .CountGdcProperties();
+                .EstimateCount();
 
             return ctx.PropertiesCount > 0
                 ? ctx.BuildList()
--- a/src/NLog/Targets/TargetWithContext.cs
+++ b/src/NLog/Targets/TargetWithContext.cs
@@ -286,7 +286,6 @@ namespace NLog.Targets
             private IEnumerable<KeyValuePair<string, object>> _scopeContextProperties;
             private ICollection<string> _gdcNames;
 
-            private bool _CountPropertiesCalled;
             private int _propertiesCount;
             private bool _checkExcludeProperties;
 
@@ -320,8 +319,6 @@ namespace NLog.Targets
                 get => _propertiesCount;
             }
 
-            private GetAllPropertiesCountContext() { }
-
             public GetAllPropertiesCountContext(TargetWithContext targetContext, LogEventInfo logEvent)
             {
                 if (targetContext is null)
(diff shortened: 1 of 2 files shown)


smnsht:
@snakefoot 
Hi,
Seems like SonarCloud not picking up unit tests for TargetWithContext.cs
Can you please direct me in this regard?

snakefoot:
Think I was a little too quick with the issue-creation. Instead of saying `GetContextPropertyList` then should be a sister-method to `GetAllProperties`. Ex. `GetAllPropertyList()` (so some better name).

The goal of the new `GetAllPropertyList` is to reduce allocation at all cost, also the yield-state-machine.

Thus if no ContextProperties and no LogEventInfo.Properties and no ScopeContext.Properties, then it would just return empty-array (zero allocation).

Also the new `GetAllPropetyList` should not depend on caling the existing `CaptureContextProperties` + `CaptureScopeContextProperties` + `CaptureContextGdc` that allocates `Dictionary`-object. 

Might also consider changing the internal `PropertiesDictionary` to implement `IEnumerabl<KeyValuePair<string, object>>` so if only using `IncludeEventProperties = true` (and empty ContextProperties  + empty ScopeContext.Properties), then it would just return LogEvenInfo.Properties directly (zero allocation).

Was thinking the new `GetAllPropertyList` should only allocate the yield-state-machine, when necessary.

src/NLog/Targets/TargetWithContext.cs
```diff
@@ -276,6 +277,104 @@ protected bool ShouldIncludeProperties(LogEventInfo logEvent)
...
+            }
+
+            if (IncludeScopeProperties && !logEvent.TryGetCachedLayoutValue(_contextLayout.ScopeContextPropertiesLayout, out object value))
+            {
+                scopeContextProperties = ScopeContext.GetAllProperties();
+                propertiesCount += scopeContextProperties.Count();
+            }
+            else
+            {
+                scopeContextProperties = null;
```
snakefoot, line 361: Believe you are missing logic from `CombineProperties` about casting `out object value` into `IDictionary` and extracting the properties.

src/NLog/Targets/TargetWithContext.cs
```diff
@@ -276,6 +277,104 @@ protected bool ShouldIncludeProperties(LogEventInfo logEvent)
...
+        /// <summary>
+        ///
+        /// </summary>
+        /// <param name="logEvent"></param>
+        /// <returns></returns>
+        protected IEnumerable<KeyValuePair<string, object>> GetAllPropertiesList(LogEventInfo logEvent)
+        {
+            if (EstimateAllPropertiesCount(logEvent, out var logEventProperties, out var scopeContextProperties, out var gdcNames, out int propertiesCount) > 0)
+            {
+                List<KeyValuePair<string, object>> combinedProperties = new List<KeyValuePair<string, object>>(propertiesCount);
```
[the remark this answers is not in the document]
smnsht: @snakefoot 
1) _Any reason for allocating List, instead of using yield-engine ?_
   I've rewritten method to use yield return

2) _Seems we are missing the ContextProperties ?_
   Yep... now fixed

3) _Believe you are missing logic from CombineProperties..._
   I will look at this tomorrow

src/NLog/Targets/TargetWithContext.cs
```diff
@@ -286,33 +286,48 @@ protected bool ShouldIncludeProperties(LogEventInfo logEvent)
         {
             if (EstimateAllPropertiesCount(logEvent, out var logEventProperties, out var scopeContextProperties, out var gdcNames, out int propertiesCount) > 0)
             {
-                List<KeyValuePair<string, object>> combinedProperties = new List<KeyValuePair<string, object>>(propertiesCount);
                 bool checkExcludeProperties = ExcludeProperties.Count > 0;
-
-                Action<string, object> tryAddProperty = (string propertyName, object propertyValue) =>
+                Func<string, object, bool, KeyValuePair<string, object> ?> buildPair = (string propertyName, object propertyValue, bool serialize) =>
```
snakefoot, line 290: Possible this can be moved into a private-method, instead of local-method ?

src/NLog/Targets/TargetWithContext.cs
```diff
@@ -286,33 +286,48 @@ protected bool ShouldIncludeProperties(LogEventInfo logEvent)
...
 
-                if(logEventProperties?.Count > 0)
+                if (logEventProperties?.Count > 0)
                 {
                     foreach (var property in logEventProperties)
-                        tryAddProperty(property.Key.ToString(), property.Value);
+                    {
+                        var pair = buildPair(property.Key.ToString(), property.Value, false);
+                        if (pair != null)
+                            yield return pair.Value;
```
snakefoot, line 319: Can the yield-engine be moved into its own method, so the yield-engine is only allocated when something to yield. Move the yield-engine into a private method, and ensure `GetAllPropertiesList` only calls the private method when `EstimateAllPropertiesCount` > 0.

Right now the yield-engine will be allocated when entering `GetAllPropertiesList`, because the method now contains the `yield`-keyword.

src/NLog/Targets/TargetWithContext.cs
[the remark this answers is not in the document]
snakefoot: Notice I will be on vacation next week, so there will be less feedback.

src/NLog/Targets/TargetWithContext.cs
smnsht: moved to `GetAllPropertiesCountContext` class  (explanation below)

src/NLog/Targets/TargetWithContext.cs
smnsht: I've removed `EstimateAllPropertiesCount()` method, and introduced `GetAllPropertiesCountContext` class.

Prev. `EstimateAllPropertiesCount` is now `ctx.PropertiesCount`

yield-engine moved into `GetAllPropertiesCountContext.BuildList()`

IMHO, now the code looks more elegant.

src/NLog/Targets/TargetWithContext.cs
smnsht: I think I don't have to destructure the CombineProperties() method, just call it with reference to null:
(new code)

```
if (_targetContext.IncludeScopeProperties)
{
    IDictionary<string, object> combinedProperties = null;
    var contextLayout = _targetContext._contextLayout.ScopeContextPropertiesLayout;
    if (!_targetContext.CombineProperties(_logEvent, contextLayout, ref combinedProperties))
    {
        _scopeContextProperties = ScopeContext.GetAllProperties();
        _propertiesCount += _scopeContextProperties.Count();
    }
}

```

src/NLog/Targets/TargetWithContext.cs
```diff
@@ -276,6 +277,175 @@ protected bool ShouldIncludeProperties(LogEventInfo logEvent)
...
+                {
+                    _logEventProperties = _logEvent.Properties;
+                    _propertiesCount += _logEventProperties.Count;
+                }
+
+                if (_targetContext.IncludeScopeProperties)
+                {
+                    IDictionary<string, object> combinedProperties = null;
+                    var contextLayout = _targetContext._contextLayout.ScopeContextPropertiesLayout;
+                    if (!_targetContext.CombineProperties(_logEvent, contextLayout, ref combinedProperties))
```
snakefoot, line 354: When `CombineProperties` returns `true` and have assigned `ref combinedProperties` then it should be used as `_scopeContextProperties`.

src/NLog/Targets/TargetWithContext.cs
```diff
@@ -276,6 +277,175 @@ protected bool ShouldIncludeProperties(LogEventInfo logEvent)
...
+                if (targetContext is null)
+                    throw new ArgumentNullException(nameof(targetContext));
+                if (logEvent is null)
+                    throw new ArgumentNullException(nameof(logEvent));
+
+                _targetContext = targetContext;
+                _logEvent = logEvent;
+            }
+
+            public GetAllPropertiesCountContext EstimateCount()
```
snakefoot, line 336: Maybe just change into a private-method, that is called by the constructor. Then you don't need the `_CountPropertiesCalled`-validation.,

src/NLog/Targets/TargetWithContext.cs
```diff
@@ -276,6 +277,175 @@ protected bool ShouldIncludeProperties(LogEventInfo logEvent)
...
+
+                return new KeyValuePair<string, object>(propertyName, serializedValue);
+            }
+
+            public int PropertiesCount
+            {
+                get => _propertiesCount;
+            }
+
+            private GetAllPropertiesCountContext() { }
```
snakefoot, line 323: Any reason for the private constructor ?

smnsht: Extra caution means ensuring that required objects (TargetWithContext  &  LogEventInfo) are passed to the constructor.

After second thought, this is an internal class, not a part of any API. Really "**extra caution**".

Deleted.

src/NLog/Targets/TargetWithContext.cs
smnsht: Done

src/NLog/Targets/TargetWithContext.cs
smnsht: Sorry, not sure if I followed you quite well... Do you mean something like that?

```
if (_targetContext.IncludeScopeProperties)
{
    IDictionary<string, object> combinedProperties = null;
    var contextLayout = _targetContext._contextLayout.ScopeContextPropertiesLayout;
    if (!_targetContext.CombineProperties(_logEvent, contextLayout, ref combinedProperties))
    {
        _scopeContextProperties = ScopeContext.GetAllProperties();
        _propertiesCount += _scopeContextProperties.Count();
    }
    // NEW
    else if(combinedProperties?.Count > 0)
    {
        _scopeContextProperties = combinedProperties;
        _propertiesCount += combinedProperties.Count;
    }
}

```

​Assuming it is true, I made my latest tentative commit

snakefoot: Again the goal is to reduce allocation at all costs, so adding `ToDictionary` is not optimal. The same with the allocation of the new private class.

But now the unit-test are green, so maybe I can merge, and adjust the code myself afterwards.

smnsht: Thank you for your patience

snakefoot:
I have now spent some time on benchmarking the new `GetAllPropertiesList()` and the existing `GetAllProperties()`. See also #5563

After several performance tweaks of the existing `PropertiesDictionary`, then the overhead of the existing `GetAllProperties()` has been minimized, so the added complexity of implementing Lazy-evaluation of LogEvent-Properties doesn't seem to pay off.

Seems my suggested idea #5543 doesn't provide the bonus, that I was expecting. The overhead from allocating a new Dictionary compared to lazy-allocated yield-engine doesn't seem that big.

Thank you for taking the time to implement proof-of-concept of how lazy-allocated-list-implementation could be done, but currently it doesn't seem a good direction after benchmark-testing.

Issue #4558, opened 2021-09-13
Can I decrease MemoryStream capacity in FileTarget somehow?


Crashhp:
Hi!

Here is my target configuration
```
<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
    <targets>
      <target name="logfile" xsi:type="AsyncWrapper" queueLimit="5000" overflowAction="Discard">
        <target xsi:type="File" layout="${message}" keepFileOpen="true" openFileCacheTimeout="30" encoding="UTF-8" createDirs="true" fileName="${appsetting:name=LOG_PATH}${appsetting:name=INSTANCE_NAME}/Logs/${shortdate:universalTime=True}/External/${appsetting:name=INSTANCE_NAME}_API/${logger}_${level}.txt"/>
      </target>
    </targets>
    <rules>
      <logger name="*" minlevel="Trace" writeTo="logfile">
      </logger>
    </rules>
    <extensions>
      <add assembly="NLog.Extended"/>
    </extensions>
  </nlog>
```

FileTarget uses ReusableStreamCreator that internally working with MemoryStream. Turns out that when you log big message, capacity of MemoryStream gets increased and that capacity is never being recalculated as far as I can tell, so there will be wasted memory space after unusually big message.

We could always try slicing up long messages to smaller ones on our side of things so that MemoryStream never becomes big at all. So my question, is there another way to free up memory space taken by MemoryStream with configuration or something?

Sorry for the grammar!

snakefoot:
NLog FileTarget attempts to perform the file-write in single operation as this works with atomtic-append.

I guess NLog could change its logic to throw away "very large" file-streams and string-builders. This is already implemented in `StringBuilderPool`.

But usually the idea is to adapt the reusable-objects to match the actual work-load. What is considered a "big message" for you ? (Notice that #3857 has been implemented to avoid huge object, but it will not protect against a single message)

snakefoot:
I guess you could "cheat" and create your own [custom Target](https://github.com/NLog/NLog/wiki/How-to-write-a-custom-target) that did the following:

```c#
[Target("MyFirst")] 
public sealed class MyFirstTarget: TargetWithLayout 
{
        [RequiredParameter] 
        public string FileName { get; set; }

        protected override void Write(LogEventInfo logEvent) 
        { 
                var logFile = RenderLayout(logEvent, FileName);
[...]

PR #5548, opened 2024-07-16
FileTarget: Reset reusable MemoryStream when above max capacity

RomanSoloweow:
Implementation for old feature request #4558.

Context:
We have rest logging (we log all requests and responses) and sometimes size of response can reach 4 megabytes or more. We came across the fact that logging requires about 400 megabytes of RAM. Here's what we found:

1. For some reason, the size of the StringBuilder used to write to memoryStream is 2-3 times larger than the original LogEvent message
2. If current MemoryStream.Capacity < StringBuilder.Length - capacity increases by 3 times from StringBuilder.Length
3. In come cases MemoryStream.Capacity still increases, despite that MemoryStream.Capacity > StringBuilder.Length
4. As a result of all these increases, a 3.5 megabyte log record results in stream capacity exceeding 80 megabytes.

After testing this fix, the amount of memory consumed was reduced by 5 times: 750 megabytes vs 150. 
I think this fix will reduce memory consumption for users logging large messages. Those who log large messages constantly can increase capacity


src/NLog/Targets/FileTarget.cs
```diff
@@ -510,6 +511,29 @@ public bool ArchiveOldFileOnStartup
...
         public long ArchiveOldFileOnStartupAboveSize { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value of the minimal number of bytes allocated for write to file stream 
+        /// </summary>
+        /// <remarks>
+        /// Default value is 4096 which means that after writing message to file stream capacity
+        /// will be changed on this value
+        /// </remarks>
+        public int WriteStreamCapacity
```
snakefoot, line 521: I prefer not having the internal recycle logic exposed through the API.

src/NLog/Targets/FileTarget.cs
```diff
@@ -971,8 +995,8 @@ private void ResetFileAppenders(string reason)
             }
         }
 
-        private readonly ReusableStreamCreator _reusableFileWriteStream = new ReusableStreamCreator(4096);
-        private readonly ReusableStreamCreator _reusableAsyncFileWriteStream = new ReusableStreamCreator(4096);
+        private ReusableStreamCreator _reusableFileWriteStream;
```
snakefoot, line 998: Like the default-capacity of 4K.

src/NLog/Targets/FileTarget.cs
```diff
@@ -971,8 +995,8 @@ private void ResetFileAppenders(string reason)
...
+        private ReusableStreamCreator _reusableAsyncFileWriteStream;
```
snakefoot, line 999: Like the default-capacity of 4K.

snakefoot:
I like the idea of not exploding into huge memory-usage by handling a single crazy LogEvent.

But I prefer the idea of just discarding crazy-large objects (Larger than ex. 100 KByte), instead of trying to re-assign their capacity.

src/NLog/Targets/FileTarget.cs
RomanSoloweow: It's still 4K by default. We set it in contructor

src/NLog/Targets/FileTarget.cs
RomanSoloweow: It's still 4K by default. We set it in contructor

RomanSoloweow:
We quite improved our research:

1. We log big response ~ 14.5 MB
2. Stream capacity increased 3X (~ 43.5 MB). Because Encoding.GetMaxByteSize return 3 time size for UTF8 case.

After that

1. We log new big response ~ 14.7 MB
2.  Stream capacity increased 3X  (~ 44.1 MB)
3. Stream capacity increased 2X  (~ 88.2 MB) because MemoryStream allocate 2 time more size by default.


It's mean that finally we got 6X overhead

RomanSoloweow:
@snakefoot I understand your idea of ​​avoiding logging large messages, but a 6x increase just sounds crazy 🤯

snakefoot:
Notice "discarding crazy large objects" doesn't mean "discarding large logevents", but that reusable-objects (StringBuilder / MemoryStream) that have grown crazy large should be discarded (start from fresh, which is already done by the NLog `StringBuilderPool`).

RomanSoloweow:
I think your idea close to [RecyclableMemoryStream](https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream). What are you think about it?

snakefoot:
> @RomanSoloweow I think your idea close to `RecyclableMemoryStream`. What are you think about it?

Like that NLog has no dependencies, so would prefer improving the existing `ReusableObjectCreator` (Have the ability to start from fresh, similar to how it is done by NLog StringBuilderPool).

snakefoot:
Notice when I say "similar to NLog StringBuilderPool", then it doesn't mean that `ReusableBuilderCreator` should be changed into a full-blown-objectpool. I'm just saying that one should take the logic for resetting an object, and extract into a seperate method, which then can be called by both NLog `StringBuilderPool` and `ReusableBuilderCreator`.

RomanSoloweow:
Already started implementing Pool based stream 😅

snakefoot:
Yes sorry about the confusion. I have tendency to expect people to be mind-readers, and whatever that is crystal clear to me, should also be crystal clear to others :)

RomanSoloweow:
Please review BufferPool before i will start inroduce it. I would like to write a couple of tests, but it seems that writing tests on Pools is not very common

RomanSoloweow:
Some Benchmarks

Original NLog 
image


With BufferPool
image

snakefoot:
I'm surprised that it is possible to build something that is faster than the standard MemoryStream. Need to test this myself.

The object-pool in NLog has been about reusing the existing .NET building-blocks, and not trying to create our own to maintain.

So I'm a little reluctant going down this road of building our own Stream-class.

Notice I will be going on vacation now, so expect less feedback

RomanSoloweow:
In this case, we will reuse existing buffers as long as we do not exceed the expected size. If exceeded, we will release excess memory. I think this will help us avoid endless memory growth due to one large message


> So I'm a little reluctant going down this road of building our own Stream-class.

RomanSoloweow:
No problem. Enjoy your vacation 🙏

> Notice I will be going on vacation now, so expect less feedback

RomanSoloweow:
You were right. In the previous implementation, I missed the GetBuffer method, which is used in all write operations. 

Implementing the GetBuffer method means implementing a method that creates a new array containing a copy of all buffers. 
Obviously this is a bad idea, so I passed the implementation to use just Stream, this led to a lot of changes, but it seems to allow us to optimize the memory allocated

new Benchmarks:

MemoryStream:

image


BufferPool:
image

snakefoot:
Think we are diverting from the initial task, of changing `ReusableBuilderCreator` to restart with fresh StringBuilder when it has grown too large.

Now there are breaking API-changes, which will require major version bump. And you have also sneaked in a `StreamWriter`-allocation in the `CopyToStream`-method.

Ofcourse it is interesting to re-think the FileTarget-class, and I have plans for NLog v6 to make a "light" FileTarget (Remove ConcurrentWrites-support and the Archive-support), and extract the original FileTarget-class into its own nuget-package.

RomanSoloweow:
Could you clarify your idea regarding ReusableBuilderCreator and StringBuilderPool? Are you offering multiple MemoryStreams and putting them in a Pool?

snakefoot:
I see 2 tasks:
- Fix ReusableBuilderCreator so it resets its StringBuilder, when it has grown too big (Extract method from `StringBuilderPool`)
- Fix ReusableStreamCreator so it resets its MemoyStream, when it has grown too big (Maybe by assigning Capacity, when having detected it has grown too big).

RomanSoloweow:
@snakefoot can you check it now please?

RomanSoloweow:
new tests:

Original:
image

new: 
image

src/NLog/Internal/ObjectPools/ReusableBuilderCreator.cs
```diff
@@ -41,7 +41,13 @@ namespace NLog.Internal
     internal sealed class ReusableBuilderCreator : ReusableObjectCreator<StringBuilder>
     {
         public ReusableBuilderCreator()
-            : base(128, (cap) => new StringBuilder(cap), (sb) => { sb.ClearBuilder(); })
+            : base(128, (cap) => new StringBuilder(cap), (capacity, sb) =>
```
snakefoot, line 44: Can `capacity` be renamed to `maxCapacity` ?

src/NLog/Internal/ObjectPools/ReusableStreamCreator.cs
```diff
@@ -41,7 +41,14 @@ namespace NLog.Internal
     internal sealed class ReusableStreamCreator : ReusableObjectCreator<System.IO.MemoryStream>, IDisposable
     {
         public ReusableStreamCreator(int capacity)
-            :base(capacity, (cap) => new System.IO.MemoryStream(cap), (m) => { m.Position = 0; m.SetLength(0); })
+            :base(capacity, (cap) => new System.IO.MemoryStream(cap), (cap, m) =>
```
snakefoot, line 44: Can `cap` be renamed to `maxCapacity` ?

src/NLog/Internal/ObjectPools/ReusableBuilderCreator.cs
```diff
@@ -41,7 +41,13 @@ namespace NLog.Internal
...
+            {
+                if (sb.Capacity > capacity)
```
snakefoot, line 46: Can this be the logic from `StringBuilderPool` extracted into a method:
```c#
                if (sb.Length > maxCapacity)
                {
                    sb.Remove(0, stringBuilder.Length - 1);  // Attempt soft clear that skips re-allocation
                    if (sb.Capacity > maxCapacity)
                    {
                        sb.Capacity = maxCapacity / 2;
                    }
                }
```

src/NLog/Internal/ObjectPools/ReusableObjectCreator.cs
```diff
@@ -41,11 +41,11 @@ namespace NLog.Internal
     internal class ReusableObjectCreator<T> where T : class
     {
         protected T _reusableObject;
-        private readonly Action<T> _clearObject;
+        private readonly Action<int, T> _clearObject;
         private readonly Func<int, T> _createObject;
         private readonly int _initialCapacity;
```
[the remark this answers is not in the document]
RomanSoloweow: 4096  * 50 * 1024 = 4 * 50 * 1024 * 1024 = 200 MB. 

I think this is a bad idea. What is the reason for these numbers?

src/NLog/Internal/ObjectPools/ReusableBuilderCreator.cs
RomanSoloweow: i'm not sure that it's maxCapacity. Because at the moment the capacity can be as large as desired until we reset this value

RomanSoloweow: for stringBuilder maxCapacity means that you really can't exceed this size
image

src/NLog/Internal/ObjectPools/ReusableBuilderCreator.cs
RomanSoloweow: Can you share what remove all characters except one, better than doing Clean and setting Capacity? I couldn't find any confirmation of this.

src/NLog/Internal/ObjectPools/ReusableBuilderCreator.cs
snakefoot: I only requested the input-parameter to be renamed. The StringBuilder should still be allowed to grow (to ensure log-events are not thrown away), but when StringBuilder has grown beyond the maxCapacity, then Capacity should be reset.

src/NLog/Internal/ObjectPools/ReusableObjectCreator.cs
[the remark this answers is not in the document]
snakefoot: Yes 200 MByte was a little to high:
- StringBuilder MaxCapacity = 1024 * 5000 = 5M
- MemoryStream = 4096 * 5000 = 20M

Thus reducing from 400 MByte to 30Mbyte. The reason for these numbers is that for applications that are frequently logging many (or very large logevents) should not be punished by constant allocation by capacity-assignment. So finding a middle ground between not using obscene amount of memory, but still support high-throughput.

src/NLog/Internal/ObjectPools/ReusableBuilderCreator.cs
snakefoot: The StringBuilder works kind of like your `BufferPoolStream` where it chains together multiple array into a stream. The StringBuilder Capacity comes from the total size of its internal arrays. When removing all characters except one, then  StringBuilder should attempt to hold on to its initial array, and lets go of the other char-array for the garbage collector to eat.

src/NLog/Internal/ObjectPools/ReusableStreamCreator.cs
RomanSoloweow: fixed

src/NLog/Internal/ObjectPools/ReusableBuilderCreator.cs
RomanSoloweow: fixed

src/NLog/Internal/ObjectPools/ReusableObjectCreator.cs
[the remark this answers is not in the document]
RomanSoloweow: fixed

src/NLog/Internal/ObjectPools/ReusableBuilderCreator.cs
RomanSoloweow: for me really many tests failed. I had this before my changes. I don't know what the reason is.
image

RomanSoloweow:
new tests. 

Original:

image

new:
image

RomanSoloweow:
I think the max values ​​are too big to be useful. Also, we still haven't solved the problem of a single large message causing a spike.

I think the best choice would be to reset the value each time to something really small or allow these values ​​to be configured

src/NLog/Internal/ObjectPools/ReusableStreamCreator.cs
```diff
@@ -40,8 +40,15 @@ namespace NLog.Internal
     /// </summary>
     internal sealed class ReusableStreamCreator : ReusableObjectCreator<System.IO.MemoryStream>, IDisposable
     {
-        public ReusableStreamCreator(int capacity)
-            :base(capacity, (cap) => new System.IO.MemoryStream(cap), (m) => { m.Position = 0; m.SetLength(0); })
+        public ReusableStreamCreator(int initialCap, int maxCap)
+            :base(initialCap, maxCap, (initialCapacity, _) => new System.IO.MemoryStream(initialCapacity), (initialCapacity, maxCapacity, m) =>
+            {
+                if (m.Capacity > maxCapacity)
+                    m.Capacity = maxCapacity;
```
snakefoot, line 47: Can the assignment of Capacity happen after having called `SetLength(0)`?

src/NLog/Internal/ObjectPools/ReusableObjectCreator.cs
```diff
@@ -41,16 +41,18 @@ namespace NLog.Internal
...
 
-        protected ReusableObjectCreator(int initialCapacity, Func<int, T> createObject, Action<T> clearObject)
+        protected ReusableObjectCreator(int initialCapacity, int maxCapacity, Func<int, int, T> createObject, Action<int, int, T> clearObject)
         {
-            _reusableObject = createObject(initialCapacity);
+            _reusableObject = createObject(initialCapacity, maxCapacity);
             _clearObject = clearObject;
             _createObject = createObject;
             _initialCapacity = initialCapacity;
+            _maxCapacity = maxCapacity;
```
snakefoot, line 55: Maybe logic about handling `_initialCapacity` + `_maxCapacity` should just be kept inside the delegate-methods `createObject` and `clearObject`.

Thus the base-class `ReusableObjectCreator` doesn't have `_initialCapacity` + `_maxCapacity` as members.

src/NLog/Internal/ObjectPools/ReusableBuilderCreator.cs
snakefoot: Maybe your TestRunner are running the tests in parallel? The original NLog unit-tests are depending on the global singleton `LogManager`, which causes side-effects when running in parallel. I have tried to fix many of the unit-tests with NLog v5, but still a long road ahead. Hopefully with NLog v6, then then all unit-test will be fixed.

Any help with improving the NLog unit-test is most wellcome.

snakefoot:
> I think the max values ​​are too big to be useful. Also, we still haven't solved the problem of a single large message causing a spike.

Thought the original goal was to fix the situation where application-memory-usage grew to 400-500 MByte. Have this not been fixed by having max-capacity of 20 MByte?

RomanSoloweow:
As I said before we have about 7 fileTargets (there may be more than 30), each of which may have a rare huge message. Now we see that when logging 10 MB, 1 FileTarget allocates up to 70 MB (the same as before). This means that 7 FileTargets will allocate the same 500 MB.

snakefoot:
If you start with fixing the broken unit-test on the build-server. Then I guess we can work on a MaxCapacity-property for the FileTarget, so people can shoot themselves in the foot if they like.

RomanSoloweow:
As I said for many tests failed out of the box. I use Rider. I have difficulty understanding which tests exactly were affected by my changes. So I need your help here

snakefoot:
When making a pull-request for the NLog project, then it will trigger build on both AzureDevOps and AppVeyor.

You can view the results of these builds to see what tests are failing. Just scroll down the bottom of the pull-request page where you will find links to build details.

src/NLog/Internal/ObjectPools/ReusableBuilderCreator.cs
```diff
@@ -40,8 +40,20 @@ namespace NLog.Internal
...
-            : base(128, (cap) => new StringBuilder(cap), (sb) => { sb.ClearBuilder(); })
+            : base(initialCapacity: InitialCapacity, maxCapacity: MaxCapacity, (initialCapacity, _) => new StringBuilder(initialCapacity), (initialCapacity, maxCapacity, sb)  =>
+            {
+                if (sb.Length > maxCapacity)
+                {
+                    sb.Remove(0, sb.Length - 1);  // Attempt soft clear that skips re-allocation
+                    
+                    if (sb.Capacity > maxCapacity)
+                        sb.Capacity = maxCapacity;
+                }
```
snakefoot, line 55: You have now remove the call to `sb.ClearBuilder()`

RomanSoloweow: Fixed

RomanSoloweow:
@snakefoot hi, any news? What's the next?

snakefoot:
I have created #5561 that tries to improve the Large-Object-Heap-handling for NLog. Each FileTarget should now have max-memory-usage of 10 MByte. Thus 7 FileTargets = 70 MByte.

Either you can enroll my suggestions in your pull-request, or I can merge my pull-request and o

RomanSoloweow:
I fill enroll your suggestions

RomanSoloweow:
new tests:

Original:
image


With changes:
image

snakefoot:
And do you approve of the new memory behavior, after handling a huge LogEvent?

RomanSoloweow:
Summarizing:

StringBuilder is initialized with a size of 40 KB. After exceeding the maximum size of 400 KB, capacity will be reset to 40 KB.
MemoryStream is initialized with a size of 4 KB. After exceeding the maximum size of 976 KB, capacity will be reset to 80 KB.

RomanSoloweow:
I believe this is a strong step forward, and we will be able to improve the values in the future. So yes, I agree and approve new behavior.

snakefoot:
Notice that a single unicode char is 2 bytes. So your Stringbuilder values has to be multiplied by 2.

Bur happy you approve, also like that the capacity reset now uses a memory size below 85 KByte to avoid the LOH.

Sent from my Galaxy



From: GMIKE ***@***.***>
Date: 08/08/2024 22:45 (GMT+01:00)
To: NLog/NLog ***@***.***>
Cc: Rolf Kristensen ***@***.***>, Mention ***@***.***>
Subject: Re: [NLog/NLog] FileTarget: Add option to configure write stream default capacity (PR #5548)


Summarizing:

StringBuilder is initialized with a size of 40 KB. After exceeding the maximum size of 400 KB, capacity will be reset to 40 KB.
MemoryStream is initialized with a size of 4 KB. After exceeding the maximum size of 976 KB, capacity will be reset to 80 KB.

snakefoot:
@RomanSoloweow Thank you for the contribution, and effort to improve the NLog code base.

RomanSoloweow:
When is the new version planned to be released?

snakefoot:
I'm guessing within a week or two. Trying to see if `TargetWithContext` can be improved, after being made aware that it allocates a little too much.

RomanSoloweow:
Thank you for your review and help in solving the problem!

snakefoot:
NLog v5.3.3 has been released:
- https://www.nuget.org/packages/NLog/5.3.3

PR #5563, opened 2024-08-09
TargetWithContext - Added GetAllPropertiesList for reduced allocation

snakefoot:
My suggestion for resolving #5543 

Reduces allocation with 20-50 pct. and improves speed by 20 pct.

Still not sure if I like the added complexity to the API (too many holes in the cheese already). Alternative approach could be a custom IDictionary.

### DIFF
--- a/src/NLog/Targets/TargetWithContext.cs
+++ b/src/NLog/Targets/TargetWithContext.cs
@@ -34,6 +34,7 @@
 namespace NLog.Targets
 {
     using System;
+    using System.Collections;
     using System.Collections.Generic;
     using System.ComponentModel;
     using System.Text;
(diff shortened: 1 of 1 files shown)


snakefoot:
I have now spent some time on benchmarking the new `GetAllPropertiesList()` and the existing `GetAllProperties()`. See also #5546

After several performance tweaks of the existing `PropertiesDictionary`, then the overhead of the existing `GetAllProperties()` has been minimized, so the added complexity of implementing Lazy-evaluation of LogEvent-Properties doesn't seem to pay off.

Seems my suggested idea #5543 doesn't provide the bonus, that I was expecting. The overhead from allocating a new Dictionary compared to lazy-allocated yield-engine doesn't seem that big.

Thank you for taking the time to implement proof-of-concept of how lazy-allocated-list-implementation could be done, but currently it doesn't seem a good direction after benchmark-testing.

PR #5572, opened 2024-08-17
 Layout.FromMethod that supports typed Layout #5532 

### LINKED ISSUE #5532: Layout.FromMethod that supports typed Layout
Implement helper to return type `Layout<T>`

```c#
public static Layout<T> FromMethod<T>(Func<LogEventInfo, T> layoutMethod, LayoutRenderOptions options = LayoutRenderOptions.None)
{
   ...
}
```

smnsht:
Trying to resolve #5532

### DIFF
--- a/tests/NLog.UnitTests/Layouts/LayoutTypedTests.cs
+++ b/tests/NLog.UnitTests/Layouts/LayoutTypedTests.cs
@@ -1,41 +1,39 @@
-﻿//
+﻿// 
 // Copyright (c) 2004-2021 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
-//
+// 
 // All rights reserved.
-//
-// Redistribution and use in source and binary forms, with or without
-// modification, are permitted provided that the following conditions
+// 
+// Redistribution and use in source and binary forms, with or without 
+// modification, are permitted provided that the following conditions 
 // are met:
-//
-// * Redistributions of source code must retain the above copyright notice,
-//   this list of conditions and the following disclaimer.
-//
+// 
+// * Redistributions of source code must retain the above copyright notice, 
+//   this list of conditions and the following disclaimer. 
+// 
 // * Redistributions in binary form must reproduce the above copyright notice,
 //   this list of conditions and the following disclaimer in the documentation
-//   and/or other materials provided with the distribution.
-//
-// * Neither the name of Jaroslaw Kowalski nor the names of its
+//   and/or other materials provided with the distribution. 
+// 
+// * Neither the name of Jaroslaw Kowalski nor the names of its 
 //   contributors may be used to endorse or promote products derived from this
-//   software without specific prior written permission.
-//
+//   software without specific prior written permission. 
+// 
 // THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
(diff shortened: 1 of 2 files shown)


tests/NLog.UnitTests/Layouts/LayoutTypedTests.cs
```diff
@@ -42,11 +44,11 @@ public class LayoutTypedTests : NLogTestBase
         [Fact]
         public void LayoutFixedIntValueTest()
         {
-            // Arrange
+            // Arrange
```
snakefoot, line 47: Possible that you can remove these extra white-spaces?

snakefoot:
Can you also copy the `Url`-value-tests, and verify they will work with `Layout.FromMethod` ?

tests/NLog.UnitTests/Layouts/LayoutTypedTests.cs
smnsht: Done.

I've noticed file headers retain trailing spaces, so, I turn off an extension (trailing white spaces) in Visual Studio when working on NLog.

What would be the correct way to handle the situation?

smnsht:
> Can you also copy the `Url`-value-tests, and verify they will work with `Layout.FromMethod` ?

Done.

`LayoutFixedEmptyUrlTest()` should be skipped, because It is not relevant for typed `FromMethod()`

Is that correct?

smnsht:
For more involved changes I will be available on the weekend (Fri-Sat)

snakefoot:
Something similar to ``LayoutFixedEmptyUrlTest` could be relevant to verify what happens if the Func-method returns `null`

tests/NLog.UnitTests/Layouts/LayoutTypedTests.cs
snakefoot: Yes I have notice that other developers has extensions included that does different bonus formatting on save-file.

Guess one could "optimize" the NLog-repo to be more friendly towards bonus-formatting-developers.

smnsht:
We have this situation covered in **Layout_FromMethod_FixedNullUrlValueTest()** (line 288)

```
[Fact]
public void Layout_FromMethod_FixedNullUrlValueTest()
{
    // Arrange
    Uri uri = null;
    Layout<Uri> layout = Layout<Uri>.FromMethod(l => uri);

    // Act
    var result = layout.RenderValue(LogEventInfo.CreateNullEvent());

    // Assert
    Assert.Equal(uri, result);
    Assert.Same(result, layout.RenderValue(LogEventInfo.CreateNullEvent()));
    Assert.Equal("", layout.Render(LogEventInfo.CreateNullEvent()));
    Assert.False(layout == uri);
    Assert.True(layout != uri);
    Assert.False(layout.IsFixed);
    Assert.NotEqual("null", layout.ToString());
    Assert.NotEqual(new Uri("//other"), layout);
}

```

snakefoot:
The it seems covered already.

smnsht:
Could it be AppVeyor build failing because of stripped-off trailing white spaces in the file headers?

```
Missing headers (copy them form other another file).
2 errors: 
 -------- 
-C:\projects\nlog\src/NLog\Layouts\Typed\Layout.cs
- C:\projects\nlog\tests/NLog.UnitTests\Layouts\LayoutTypedTests.cs 
Command exited with code 1
```

snakefoot:
I think this initial API implementation looks good. Excellent job on the unit-tests to ensure the API behaves as expected.

Then future pull-request can work towards making it non-boxing.

snakefoot:
Thank you for the contribution and getting the ball rolling. Created #5575 to continue the work.

PR #5587, opened 2024-09-01
NetworkTarget - Added SendTimeoutSeconds to assign TCP Socket SendTimeout

snakefoot:
TCP Socket will automatically block when send-buffer becomes full, to throttle when receiver cannot keep up.

By default the TCP Socket SendTimeout is zero, which means block forever, which can happen when network-cable is unplugged and send-buffer gets full.

The new setting `SendTimeoutSeconds` will enable timer, when send-buffer becomes full. Consider using a timeout-value of `100` seconds.


snakefoot:
Updated wiki: https://github.com/nlog/nlog/wiki/Network-target

PR #5635, opened 2024-10-06
Move MailTarget to its own nuget-package

snakefoot:


snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/Mail-target

PR #5641, opened 2024-10-10
Fix XML docs for ICreateFileParameters options

Toby222:


snakefoot:
Very nicely spotted, and thank you also offering a pull-request.

PR #5646, opened 2024-10-13
ColoredConsoleTarget without direct RegEx support for word highlight

snakefoot:
Also removed RegEx support for `${replace}`-layoutrenderer. Instead introduced new `${regex-replace}` in NLog.RegEx-package.

Working towards removing RegEx as dependency for NLog to prepare for AOT


snakefoot:
Updated wiki - https://github.com/NLog/NLog/wiki/ColoredConsole-target
Updated wiki - https://github.com/NLog/NLog/wiki/Replace-Layout-Renderer
Created new wiki - https://github.com/NLog/NLog/wiki/RegEx-Replace-Layout-Renderer

PR #5647, opened 2024-10-13
Regex-Replace - Removed RegEx-option since implicit by type name

snakefoot:
Followup to #5646


snakefoot:
Updated wiki - https://github.com/NLog/NLog/wiki/Replace-Layout-Renderer
Created new wiki - https://github.com/NLog/NLog/wiki/RegEx-Replace-Layout-Renderer

PR #5673, opened 2025-01-23
FileTarget without ConcurrentWrites and Mutex and FileSystemWatcher

snakefoot:
- [x] Make git-clone of `FileTarget.cs` and `FileTargetTests.cs` for `ConcurrentFileTarget`-nuget-package before merging

New archive-logic that rolls to the next file-sequence-no, instead of doing `File.Move` of the active-logging-file. This should reduce conflicts with other applications holding file-locks. It is still possible to use `ArchiveFileName` that will activate the original archive-logic with `File.Move`.

Initially the idea was to remove all archive-logic around `ArchiveFileName`. But it felt like too big a breaking-change, as it would break many NLog tutorials/examples, and also because many unit-tests were based upon `ArchiveFileName`. But still there are many breaking changes, and `ArchiveFileName` is now mostly kept around for legacy reason:
- `ConcurrentWrites`-property has been removed.
- `EnableArchiveFileCompression`-property has been removed (Zip-compression removed)
- `CleanupFileName`-property has been removed.
- `FileAttributes`-property has been removed.
- `ForceManaged`-property has been removed.
- `ConcurrentWriteAttempts`-property has been removed.
- `ConcurrentWriteAttemptDelay`-property has been removed.
- `ForceMutexConcurrentWrites`-property has been removed.
- `NetworkWrites`-property has been removed.
- `FileNameKind`-property has been removed.
- `ArchiveFileKind`-property has been removed.
- `ArchiveOldFileOnStartupAboveSize`-property has been removed.
- `ArchiveDateFormat`-property marked as obsolete. Instead use new `ArchiveSuffixFormat`
- `ArchiveNumbering`-property marked as obsolete. Instead use new `ArchiveSuffixFormat` (Rolling is unsupported).

The `ConcurrentWrites` is a very big hammer with lots of dependencies (Mutex + FileSystemWatcher), and still it is not perfect as it has issues with Mutex-Security-Rights, and if also enabling file-archive-logic, then there will be issues with file-locking. If multi-process writing to the same file is important, then one can still use `KeepFileOpen=false` (slow so make sure to include `async="true"`).

The original FileTarget has been extracted into its own nuget-package `NLog.Targets.ConcurrentFile`. So users depending on `ConcurrentWrites=true` or `EnableArchiveFileCompression=true` or `ArchiveNumbering=Rolling` can still upgrade to NLog v6, by using that nuget-package.

Other things needed:
- Consider introducing AtomicFileTarget (FileTarget with ConcurrentWrites but without mutex)
- Consider adding virtual method on FileTarget so others can react to archive-roll-events (ex. perform file-compression)


snakefoot:
Once again hit in the face with `Failed to convert the binary code coverage reports to XML. No code coverage information will be uploaded to the server (SonarQube/SonarCloud)`.

Guess it is trial-on-Error again until finding the perfectly fine code, that Sonar-Qube doesn't like, but fail to explain why.

snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/FileTarget-Archive-Examples
Updated wiki: https://github.com/NLog/NLog/wiki/Atomic-File-Target

Still need to write some about the new option `archiveSuffixFormat`

PR #5690, opened 2025-02-01
SimpleLayout - Marked Text-setter as obsolete

snakefoot:
Already now NLog v5 expects SimpleLayout to be immutable, so make the contract clear.


snakefoot:
Closing as impossible.

PR #5712, opened 2025-02-22
XmlParser to replace XmlReader to support AOT

snakefoot:
XmlReader doesn't introduce AOT-warnings, but does introduce a lot of dependencies, because XmlReader has the ability to load from network-http-url / codegen / etc.


snakefoot:
Discovered that with #5715 that System.Xml doesn't generate AOT-warnings, but do introduce a 3 MByte App-Size-Increase.

snakefoot:
With this custom XML parser, then it shaves off 4 MByte, and removes dependencies on System.Net. Because XmlReader has the ability to load XML document using web-url.

snakefoot:
Maybe it should be a hybrid, so Layout- and LayoutRenderer-parsing always works but full-configuration-file-parsing with targets / filters / etc. is something that must be enabled explicitly. Thus one will only pay for the targets / configuration-options being used.

Would prefer that NLog with AOT-support would have a minimal build and startup-overhead, when not depending on dynamic loading from `NLog.config`-file.

snakefoot:
Tested basic hybrid version where AOT-build only included NLog Layouts + NLog Targets used by Fluent API. And simple console-application changed from 7 MByte to 4 Mbyte. I guess further optimizations can be made, but it seems disabling automatic loading of `NLog.config` makes it easier to trim NLog.

snakefoot:
Not sure what has happened but now published application is now 5 MByte with everything included. Still more than 2 MByte of a simple Console.WriteLine-application.

Starting to consider if this is enough, and thus skipping the huge breaking change of not automatically loading `NLog.config`-file

PR #5715, opened 2025-03-01
Fix AOT warnings by not using MakeGenericType

snakefoot:
Stop using obsolete Assembly-CodeBase (original source-location), but instead always using Assembly-Location (can be nuget-cache-folder, because of NTFS-links)


snakefoot:
Create new pull-request with Android-detection:
```
            if (System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform(System.Runtime.InteropServices.OSPlatform.Create("ANDROID")))
                return RuntimeOS.Android;
``

PR #5722, opened 2025-03-05
Extracted NLogTraceListener into NLog.Targets.Trace

snakefoot:
TraceSource + TraceListener + TraceActivity all seem legacy now, so moving out of NLog-core.


snakefoot:
Starting to wonder if `System.Diagnostics.Trace.WriteLine` is still relevant to keep in NLog-core, or whether it is okay to throw it into its own nuget-package.

Maybe one should also throw `System.Diagnostics.Debug.WriteLine` into the mix, and call the nuget-package something else.

snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/Trace-target
Updated wiki: https://github.com/NLog/NLog/wiki/NLog-Trace-Listener-for-System-Diagnostics-Trace
Updated wiki: https://github.com/NLog/NLog/wiki/Trace-Activity-Id-Layout-Renderer

PR #5724, opened 2025-03-09
ConfigurationItemFactory - Lazy initialize factory types until requested by config

snakefoot:
Makes it easier for AOT-builds to trim NLog-types that are not used. But still waiting for automatic configuration-loading to be disabled by default.


snakefoot:
Still need to resolve 2 issues:
- Disable loading NLog.config by default (avoid hard dependency on XmlLoggingConfiguration)
- Avoid unsafe reflection when validating NLog LoggingConfiguration built using fluent API

PR #5732, opened 2025-03-17
Removed FileContentsLayoutRenderer as too exotic

snakefoot:
Can be replaced by Func-LayoutRenderer if needed. Reduce AOT-build dependencies when only need console output.


snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/FileContents-layout-renderer

PR #5740, opened 2025-03-23
ValueFormatter - Skip string-quotes by default

### LINKED ISSUE #5738: Skip string quotes when structured formatting
Decision stolen from Serilog with the initial implementation of message-template-formatting with NLog v4.5 - #2208

But Microsoft Extension Logging decided against this behavior, and I think NLog should follow. Thus avoiding surprises when using NLog Logger directly instead as LoggingProvider for Microsoft Extension Logging.

snakefoot:
Resolves #5738

To revert to old behavior:
```csharp
LogManager.Setup().SetupSerialization(s => s.RegisterValueFormatterWithStringQuotes());
```


snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/How-to-use-structured-logging

PR #5749, opened 2025-03-24
ReplaceLayoutRendererWrapper - SearchFor and ReplaceWith with basic layout support

### LINKED ISSUE #4042: Dynamic template within replace layout renderer
There is an unexpected behavior of [Replace Layout Renderer] when [Var Layout Renderer] is used within `searchFor` option. 

**NLog version**: 4.7.2  

**Platform**: .NET Core 3  

**Current NLog config**:

```xml
<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd" xsi:schemaLocation="NLog NLog.xsd"
      autoReload="true"
      internalLogLevel="Off"
      internalLogFile="${basedir}/Logs/nlog.log">
  <variable name="secret" value="" />
  <targets>
    <target name="Console"
            xsi:type="Console"
            encoding="utf-8"
            layout="${pad:padding=5:inner=${uppercase:${level}}} ${replace:inner=${message}:searchFor=${var:secret}:replaceWith=****}" />
  </targets>
  <rules>
    <logger name="*" minlevel="Trace" writeTo="Console" />
  </rules>
</nlog>
```

**Code snippet to reproduce**:

```c#
public class Program
{
    private static readonly Logger logger = LogManager.GetCurrentClassLogger();

    public static void Main(string[] args)
    {
        LogManager.Configuration.Variables["secret"] = "MYSECRET"; // In real life the value is loaded from application's configuration
        logger.Info("MYSECRET should be hidden anywhere.");
    }
}
```

It displays following line:
> INFO secret:replaceWith=****}

But expected message is:
> INFO **** should be hidden.

If I set `secret` value within **nlog.config** and use `${secret}` instead of `${var:secret}` it works as expected. But I need the variable will be updated through API after apptlication's configuration binding.

[Replace Layout Renderer]: https://github.com/NLog/NLog/wiki/Replace-Layout-Renderer
[Var Layout Renderer]: https://github.com/nlog/nlog/wiki/Var-Layout-Renderer

snakefoot:
Resolves #4042 for NLog v5.5

- [ ] Wait for NLog v5.5

Also applies the same technique for these:
- AllEventPropertiesLayoutRenderer - Separator
- ExceptionLayoutRenderer - Separator + ExceptionDataSeparator
- ScopeContextNestedStatesLayoutRenderer - Separator
- StackTraceLayoutRenderer  - Separator


snakefoot:
Closing due to side-effects, so postponed for NLog v6 with #5748

PR #5752, opened 2025-03-30
NetworkTarget - Added support for loading SSL certificate from file

### LINKED ISSUE #3033: Log to use SSL with tcp and/or http to logstash targets.
- Why do we need it?

1) When using NLog from .NET Core it is not possible to use the Windows Certificate Store.
NLog could effectively use a new "certificate" attribute for <target xsi:type="Network" ...>
2) The public key x509 certificate would then be available via the solution itself in a file with a relative path to nlog.config itself.
3) Rolf Kristensen is referred to through this link https://github.com/NLog/NLog/issues/2677
but the code doesnt seem available.

-- how?

4) More concretely in the XML taken from NLog.config - a solution e.g. that allows the following: 
 <target name="logstash"
             xsi:type="Network" 
             layout="${structuredlogging.json}" 
             address="tcp://es01.monitoring-production.aweu.cloud-xxxxx.com:10201"
             certificate="<relative path to the nlog.config file of certificate>" /> 
5) i think a solution where the new certificate attribute is available for Network Targets expecially Targets with a tcp based adress (i saw internally in NLog they use SslStream.   but although SslStream is there - it seems always to be null when i debug through the code. 
6) Network-Targets Support also https ... so the new Certificate attribute could also be used there.
  
**NLog version**: 
4.5.12

**Platform**:   / .Net 4.0 / .Net 4.5 / Mono 4 / Silverlight 4 / Silverlight 5 / Xamarin Android / Xamarin iOs  / .NET Core 2

snakefoot:
Resolves #3033 by introducing 2 new options for NetworkTarget:
- public Layout SslCertificateFile { get; set; }
- public Layout SslCertificatePassword { get; set; }

### DIFF
--- a/src/NLog.Targets.Network/Internal/PlatformDetector.cs
+++ b/src/NLog.Targets.Network/Internal/PlatformDetector.cs
@@ -62,6 +62,8 @@ namespace NLog.Targets.Internal
                 return PlatformOS.MacOSX;
             if (System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform(System.Runtime.InteropServices.OSPlatform.Windows))
                 return PlatformOS.Windows;
+            if (System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform(System.Runtime.InteropServices.OSPlatform.Create("ANDROID")))
+                return PlatformOS.Linux;
 #endif
             return PlatformOS.Unknown;
         }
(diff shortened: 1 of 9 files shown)


snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/Log4jxml-target
Updated wiki: https://github.com/NLog/NLog/wiki/Syslog-Target
Updated wiki: https://github.com/NLog/NLog/wiki/GELF-Target

PR #5754, opened 2025-04-01
NLog.Targets.Network - Introduced SyslogTarget and SyslogLayout

### LINKED ISSUE #5745: Syslog Layout for NLog.Targets.Network
Allow the NetworkTarget to support Syslog-server out of the box.

Syslog-format is a fixed-length format, so maybe try using CompoundLayout as wrapper around several fixed-length-layouts.

Support rfc3164 and rfc5424 (default: rfc5424)

snakefoot:
Resolves #5745


snakefoot:
@jasells + @RomanLeonB + @JayToltTech + @dpmm99 

NLog v6 Preview1 has now been published, and introduces a new nuget-package [NLog.Targets.Network](https://www.nuget.org/packages/NLog.Targets.Network) that includes the following NLog Targets:
- NLog NetworkTarget (moved out of the NLog-core-package)
- NLog GelfTarget
- NLog SyslogTarget

Would love to get input on whether the new NLog SyslogTarget will able to support your scenario, before reaching NLog v6 Final. Example of NLog.config:
```xml
<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" >
  <extensions>
    <add assembly="NLog.Targets.Network" />
  </extensions>
  
  <targets async="true">
	<target xsi:type="SysLog" name="SyslogTcp" address="tcp://localhost:514">
            <Rfc3164>false</Rfc3164>
            <Rfc5424>true</Rfc5424>
            <StructuredDataParam name="MyPropertyName" layout="MyPropertyValue" />
	</target>
  </targets>

  <rules>
    <logger name="*" minlevel="Debug" writeTo="SyslogTcp" />
  </rules>
</nlog>
``` 
Available options that can be adjusted:
- SyslogAppName
- SyslogHostName
- SyslogMessage
- SyslogLevel
- SyslogFacility
- StructuredDataId
- IncludeEventProperties

NLog SyslogTarget also inherits options from NLog NetworkTarget - https://github.com/nlog/nlog/wiki/Network-target where NLog v6 introduces the new options:
- SslCertificateFile
- SslCertificatePassword
- SendTimeoutSeconds (Default 100)
- NoDelay (Default true)

See also: https://nlog-project.org/2025/04/29/nlog-6-0-major-changes.html

jasells:
🤔Maybe....  The only issue I had was incorrect identification of Android as linux in .Net8 Maui.  This fork is where Ifixed that issue for my current purposes.

https://github.com/jasells/NLog.Targets.Syslog/commit/263c51978147ff43b7b88f383a01345bbca52068#diff-bb92c1c4cfb695048f9bf0b3ab3b519c4d0e05e7b1cf4bfa49c4c47b6be06232

See changes to identify Android as linux OS in .Net8 (probably .Net6+?):
src/NLog.Targets.Syslog/MessageSend/SocketInitialization.cs


The root of the issue was the os-specific socket init code in that lib.  If this stack doesn't get overly complicated in setting up the sockets, it may well work without the little hack I added to id the os in the other lib.

snakefoot:
@RyanGaudion @davidgeary @sanderweltje @JayBeavers

The SyslogTarget included in NLog v6 is a rewrite that is based on NLog NetworkTarget. It should also work with Maui Android.

Any feedback on the new SyslogTarget from NLog v6 [NLog.Targets.Network](https://www.nuget.org/packages/NLog.Targets.Network) is most welcome.

PR #5762, opened 2025-04-03
Removed static constructor from LogFactory

Pavan8374:
This PR introduces a test to ensure that no new explicit static constructors are introduced into the codebase while allowing the existing ones.

Changes Introduced:

- Implemented a unit test that scans all .cs files in the solution for explicit static constructors.
- Allowed existing static constructors using a whitelist of class names to avoid breaking changes.
- The test will fail only if a new static constructor is added.

Why This is Needed:

- Prevents unnecessary static initialization, improving performance and maintainability.
- Avoids accidental introduction of new static constructors.
- Reduces long-term maintenance effort by tracking only class names, not specific lines.

### DIFF
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -1245,7 +1245,7 @@ namespace NLog
         {
             try
             {
-                _loggerShutdown?.Invoke(sender, args);
+                _loggerShutdown?.Invoke(null, args);
             }
             catch (Exception ex)
             {
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -95,46 +95,24 @@ namespace NLog
         /// <summary>
         /// Event that is raised when the logging system is shutting down.
         /// </summary>
-        internal static event EventHandler<EventArgs> LoggerShutdown
+        private static event EventHandler<EventArgs> LoggerShutdown
         {
             add
             {
-                if (_loggerShutdown == null && DefaultAppEnvironment != null)
+                if (_loggerShutdown is null && DefaultAppEnvironment != null)
                 {
-                    try
-                    {
-                        InternalLogger.Debug("Registered shutdown event handler for ProcessExit.");
-                        DefaultAppEnvironment.ProcessExit += OnLoggerShutdown;
-                    }
-                    catch (Exception exception)
-                    {
-                        InternalLogger.Warn(exception, "Error registering ProcessExit event handler.");
-                        if (exception.MustBeRethrown())
-                        {
-                            throw;
-                        }
-                    }
+                    InternalLogger.Debug("Registered shutdown event handler for ProcessExit.");
+                    DefaultAppEnvironment.ProcessExit += OnLoggerShutdown;
                 }
                 _loggerShutdown += value;
             }
             remove
             {
                 _loggerShutdown -= value;
-                if (_loggerShutdown == null && DefaultAppEnvironment != null)
+                if (_loggerShutdown is null && DefaultAppEnvironment != null)
                 {
-                    try
-                    {
-                        InternalLogger.Debug("Unregistered shutdown event handler for ProcessExit.");
-                        DefaultAppEnvironment.ProcessExit -= OnLoggerShutdown;
-                    }
-                    catch (Exception exception)
-                    {
-                        InternalLogger.Warn(exception, "Error unregistering ProcessExit event handler.");
-                        if (exception.MustBeRethrown())
-                        {
-                            throw;
-                        }
-                    }
+                    InternalLogger.Debug("Unregistered shutdown event handler for ProcessExit.");
+                    DefaultAppEnvironment.ProcessExit -= OnLoggerShutdown;
                 }
             }
         }
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -1254,6 +1254,14 @@ namespace NLog
                     throw;
                 InternalLogger.Error(ex, "LogFactory failed to shutdown properly.");
             }
+            finally
+            {
+                _loggerShutdown = null;
+                if (DefaultAppEnvironment != null)
+                {
+                    DefaultAppEnvironment.ProcessExit -= OnLoggerShutdown;
+                }
+            }
         }
 
         private void OnStopLogging(object sender, EventArgs args)
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -90,14 +90,53 @@ namespace NLog
         /// </remarks>
         public event EventHandler<LoggingConfigurationChangedEventArgs> ConfigurationChanged;
 
-        private static event EventHandler<EventArgs> LoggerShutdown;
+        private static event EventHandler<EventArgs> _loggerShutdown;
 
         /// <summary>
-        /// Initializes static members of the LogManager class.
+        /// Event that is raised when the logging system is shutting down.
         /// </summary>
-        static LogFactory()
+        public static event EventHandler<EventArgs> LoggerShutdown
         {
-            RegisterEvents(DefaultAppEnvironment);
+            add
+            {
+                if (_loggerShutdown == null && DefaultAppEnvironment != null)
+                {
+                    try
+                    {
+                        DefaultAppEnvironment.ProcessExit += OnLoggerShutdown;
+                        InternalLogger.Debug("Registered shutdown event handler for ProcessExit.");
+                    }
+                    catch (Exception exception)
+                    {
+                        InternalLogger.Warn(exception, "Error registering ProcessExit event handler.");
+                        if (exception.MustBeRethrown())
+                        {
+                            throw;
+                        }
+                    }
+                }
+                _loggerShutdown += value;
+            }
+            remove
+            {
+                _loggerShutdown -= value;
+                if (_loggerShutdown == null && DefaultAppEnvironment != null)
+                {
+                    try
+                    {
+                        DefaultAppEnvironment.ProcessExit -= OnLoggerShutdown;
+                        InternalLogger.Debug("Unregistered shutdown event handler for ProcessExit.");
+                    }
+                    catch (Exception exception)
+                    {
+                        InternalLogger.Warn(exception, "Error unregistering ProcessExit event handler.");
+                        if (exception.MustBeRethrown())
+                        {
+                            throw;
+                        }
+                    }
+                }
+            }
         }
 
         /// <summary>
@@ -1225,44 +1264,17 @@ namespace NLog
             }
         }
 
-        private static void RegisterEvents(IAppEnvironment appEnvironment)
-        {
-            if (appEnvironment is null) return;
-
-            try
-            {
-                appEnvironment.ProcessExit += OnLoggerShutdown;
-            }
-            catch (Exception exception)
-            {
-                InternalLogger.Warn(exception, "Error setting up termination events.");
-
-                if (exception.MustBeRethrown())
-                {
-                    throw;
-                }
-            }
-        }
-
         private static void OnLoggerShutdown(object sender, EventArgs args)
         {
             try
             {
-                LoggerShutdown?.Invoke(sender, args);
+                _loggerShutdown?.Invoke(sender, args);
             }
             catch (Exception ex)
             {
                 if (ex.MustBeRethrownImmediately())
                     throw;
                 InternalLogger.Error(ex, "LogFactory failed to shutdown properly.");
-            }
-            finally
-            {
-                LoggerShutdown = null;
-                if (defaultAppEnvironment != null)
-                {
-                    defaultAppEnvironment.ProcessExit -= OnLoggerShutdown;  // Unregister from AppDomain
-                }
             }
         }
 
--- a/src/NLog/LogFactory.cs
+++ b/src/NLog/LogFactory.cs
@@ -95,7 +95,7 @@ namespace NLog
         /// <summary>
         /// Event that is raised when the logging system is shutting down.
         /// </summary>
-        public static event EventHandler<EventArgs> LoggerShutdown
+        internal static event EventHandler<EventArgs> LoggerShutdown
         {
             add
             {
@@ -103,8 +103,8 @@ namespace NLog
                 {
                     try
                     {
-                        DefaultAppEnvironment.ProcessExit += OnLoggerShutdown;
                         InternalLogger.Debug("Registered shutdown event handler for ProcessExit.");
+                        DefaultAppEnvironment.ProcessExit += OnLoggerShutdown;
                     }
                     catch (Exception exception)
                     {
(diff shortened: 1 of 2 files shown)


snakefoot:
Any reason for doing file scanning instead of doing reflection on the class-types in the NLog-assembly like the other API-tests?

Pavan8374:
You're right—I initially overlooked the reflection-based approach you suggested. The reason I went with file scanning is that GetType() might pick up unintended constructors, including compiler-generated ones, which could lead to false positives. By scanning source files, we ensure that only explicitly written static constructors in the project's code are detected. Let me know if you see any concerns with this approach!

snakefoot:
Why not just exclude compiler generated constructor-methods ? (If you haven't guessed then I'm not a fan of file-scanning)

Pavan8374:
Got it. I’ll update the approach to exclude compiler-generated constructors using reflection. Thanks for the input

Pavan8374:
Hey @snakefoot, I tried excluding compiler-generated static constructors using typeof(CompilerGeneratedAttribute), but the results are still not as expected. Here’s the approach I took:

`
    // Known allowed explicit static constructors (full type names)
    var knownStaticConstructors = new HashSet<string>
    {
        "NLog.LogFactory",
        "NLog.Targets.ConcurrentFileTarget",
        "NLog.UnitTests.DatabaseTargetTests",
        "NLog.Targets.ConcurrentFile.Tests.ProcessRunner",
        "NLog.UnitTests.LogManagerTests",
    };

    var assembly = typeof(NLog.LogFactory).Assembly;
    var typesWithStaticConstructors = assembly.GetTypes()
        // Exclude compiler-generated types (e.g., lambdas, nested <>c classes)
        .Where(type => !type.IsDefined(typeof(CompilerGeneratedAttribute), inherit: false))
        .Select(type => new
        {
            TypeName = type.FullName,
            StaticConstructor = type.TypeInitializer
        })
        // Check for non-compiler-generated static constructors
        .Where(t => t.StaticConstructor != null
                     && !t.StaticConstructor.IsDefined(typeof(CompilerGeneratedAttribute), inherit: false))
        .Select(t => t.TypeName)
        .Distinct()
        .ToList();

    var newConstructors = typesWithStaticConstructors
        .Where(type => !knownStaticConstructors.Contains(type))
        .ToList();
`

This correctly filters out compiler-generated types, but the test still flags several classes that don't have explicit static constructors but seem to have static initializers, which IL treats as static constructors.

Here is the result of above testcase after executed:

` NLog.UnitTests.ApiTests.ShouldNotHaveExplicitStaticConstructors
   Source: ApiTests.cs line 475
   Duration: 9.3 sec

  Message: 
Found new explicit static constructors in:
NLog.GlobalDiagnosticsContext
NLog.LogEventInfo
NLog.Logger
NLog.LogLevel
NLog.LogManager, etc.
`

Also, I have verified the above classes donot have explicit static constructors, but may have static initializers. Any thoughts on how to refine this further?

snakefoot:
The goal is not to build the perfect unit-test. The unit-test was mostly to provide a TODO-list for classes to look at. If there are some false-positives, because we are not experts in the dotnet-compiler, then it is fine with me. Ofcourse a bonus part of the unit-test is to prevent new code from introducing static-constructors, so it could be nice that it was perfect, but it is not primary concern for me.

The important part is to look at the discovered classes, and see if something can be done to remove them from the TODO-list.

Pavan8374:
Hey @snakefoot, currently, the list of detected static constructors includes 50+ types. I can whitelist them all by adding to knownStaticConstructors, with comments separating the explicit ones from likely compiler-generated ones.

Just wanted to check:
Would you prefer adding them all to the known list, or is it okay if the test fails (and serves as a TODO list)?

Just wanted to confirm the direction you’d prefer before updating the PR

snakefoot:
Think you should add them all to the known list of past mistakes, so the test passes, and then they can be eaten one at a time (either being fixed or verified as false positive).

Pavan8374:
Hi @snakefoot, I've Updated the test case to include all detected types in the known static constructors list, as discussed.
Grouped known explicit and likely compiler-generated ones separately for clarity.
Test now passes and ready for further review or gradual cleanup. Let me know if any adjustments are needed

snakefoot:
Think it looks good. Surprised so many false-positives, and also surprised about these:
```
                "NLog.Targets.ConcurrentFileTarget",
                "NLog.UnitTests.DatabaseTargetTests",
                "NLog.Targets.ConcurrentFile.Tests.ProcessRunner",
                "NLog.UnitTests.LogManagerTests",
```
Since they should not be part of the NLog-assembly. Think we can start the gradual cleanup

Pavan8374:
You're right — those types aren’t part of the NLog assembly. I had parsed them from the overall source tree and missed separating them clearly in the comment. I’ll update the test to better reflect that distinction.

Also, Shall we start planning the gradual cleanup now?

Pavan8374:
Got it, thanks! Since the root issue stems from LogManager, I'll continue the LogFactory cleanup as part of this PR.

If things get too large or need broader refactoring, I might consider moving it to a separate PR — but for now, I’ll focus on getting the cleanup done here.

snakefoot:
Notice that I will go on vacation next week, so do not expect much feedback from me.

tests/NLog.UnitTests/ApiTests.cs
```diff
@@ -466,5 +467,97 @@ public void ValidateConfigurationItemFactory()
...
+                "NLog.Internal.ObjectReflectionCache+ObjectPropertyInfos",
+                "NLog.Internal.ObjectReflectionCache+EmptyDictionaryEnumerator",
+                "NLog.Internal.PropertiesDictionary+PropertyKeyComparer",
+                "NLog.Internal.SingleItemOptimizedHashSet`1+ReferenceEqualityComparer",
+                "NLog.Config.LoggerNameMatcher+NoneLoggerNameMatcher",
+                "NLog.Config.LoggerNameMatcher+AllLoggerNameMatcher",
+                "NLog.Config.LoggingConfigurationParser+ValidatedConfigurationElement"
+            };     
+            var assembly = typeof(NLog.LogFactory).Assembly;
+            var typesWithStaticConstructors = assembly.GetTypes()
```
snakefoot, line 537: Can we use the class-member `allTypes` instead of calling `assembly.GetTypes()` ?

snakefoot: I'm close to having completed the other remaining tasks before creating the first NLog v6 Preview-Build.

So I might fix the LogFactory static-constructor myself, as I would like to get some early feedback for NLog v6.

Pavan8374: That’s great to hear. Appreciate the update and the direction. On my end, I’ve almost completed the LogFactory static constructor fix using the lazy subscription approach you suggested — just one commit away.

Quick question: what’s the best way to test this change locally before committing, to ensure everything works as expected and doesn’t break any existing behavior?

snakefoot: Usually the testing depends on the implementation, and I guess you have already noticed that LogFactory tries to hookup to the AppDomain ProcessExit to ensure flush happens automatically on appdomain-shutdown. Not sure if there any unit-tests to verify this in NLog (you can take a look), but I guess with minimal effort you can make a test-application and verify flush correctly happens on process-exit.

Pavan8374: ---

Hey @snakefoot,

I’ve pushed the changes for the `LogFactory` shutdown event using the lazy subscription pattern. Now the `ProcessExit` subscription only kicks in when someone actually subscribes to `LoggerShutdown`.

Here’s a quick rundown of what I changed:

- Switched `LoggerShutdown` to a property with custom `add`/`remove` accessors to manage the `ProcessExit` hook.
- Removed the static constructor and `RegisterEvents` method — they’re no longer needed.
- Cleaned up `OnLoggerShutdown` by dropping the forced unregistration from the `finally` block.

I tested it out with a console app and things looked good:
- The event handler got called on app exit.
- Logs flushed properly to the file.
- Adding and removing multiple handlers behaved as expected.

Let me know what you think, or if there’s anything I should adjust!

---

snakefoot:
Seems CodeRabbit gets confused by your git-merge with dev-branch. Think you should either change your configuration to perform git-merge-with-rebase-and-force-push, or you should stop merging with dev-branch.

image

Pavan8374:
Thanks for pointing that out! I had some merge commits earlier and did a rebase recently, which probably caused the confusion. I'll clean up the branch with a proper rebase and force push to maintain a linear history.

src/NLog/LogFactory.cs
```diff
@@ -90,14 +90,53 @@ public class LogFactory : IDisposable
...
 
-        private static event EventHandler<EventArgs> LoggerShutdown;
+        private static event EventHandler<EventArgs> _loggerShutdown;
 
         /// <summary>
-        /// Initializes static members of the LogManager class.
+        /// Event that is raised when the logging system is shutting down.
         /// </summary>
-        static LogFactory()
+        public static event EventHandler<EventArgs> LoggerShutdown
```
snakefoot, line 98: Can it become `internal` instead of `public`, since the public API should remain unchanged.

src/NLog/LogFactory.cs
```diff
@@ -90,14 +90,53 @@ public class LogFactory : IDisposable
...
         {
-            RegisterEvents(DefaultAppEnvironment);
+            add
+            {
+                if (_loggerShutdown == null && DefaultAppEnvironment != null)
+                {
+                    try
+                    {
+                        DefaultAppEnvironment.ProcessExit += OnLoggerShutdown;
```
snakefoot, line 106: Possible that all the exception-handling could be performed inside `DefaultAppEnvironment.ProcessExit` ?

src/NLog/LogFactory.cs
```diff
@@ -90,14 +90,53 @@ public class LogFactory : IDisposable
...
+                _loggerShutdown += value;
+            }
+            remove
+            {
+                _loggerShutdown -= value;
+                if (_loggerShutdown == null && DefaultAppEnvironment != null)
+                {
+                    try
+                    {
+                        DefaultAppEnvironment.ProcessExit -= OnLoggerShutdown;
```
snakefoot, line 127: Possible that all the exception-handling could be performed inside `DefaultAppEnvironment.ProcessExit` ?

src/NLog/LogFactory.cs
```diff
@@ -90,14 +90,53 @@ public class LogFactory : IDisposable
...
         {
-            RegisterEvents(DefaultAppEnvironment);
+            add
+            {
+                if (_loggerShutdown == null && DefaultAppEnvironment != null)
+                {
+                    try
+                    {
+                        DefaultAppEnvironment.ProcessExit += OnLoggerShutdown;
+                        InternalLogger.Debug("Registered shutdown event handler for ProcessExit.");
```
snakefoot, line 107: Can `InternalLogger.Debug` be moved up so it happens before the subscription to ProcessExit ?

src/NLog/LogFactory.cs
```diff
@@ -90,14 +90,53 @@ public class LogFactory : IDisposable
...
+            }
+            remove
+            {
+                _loggerShutdown -= value;
+                if (_loggerShutdown == null && DefaultAppEnvironment != null)
+                {
+                    try
+                    {
+                        DefaultAppEnvironment.ProcessExit -= OnLoggerShutdown;
+                        InternalLogger.Debug("Unregistered shutdown event handler for ProcessExit.");
```
snakefoot, line 128: Can `InternalLogger.Debug` be moved up so it happens before the unsubscription to ProcessExit ?

snakefoot:
Have some minor comments, but I think it looks great.

The AppVeyor-build has a complaint:
```
  Failed NLog.UnitTests.ApiTests.ShouldNotHaveExplicitStaticConstructors [11 ms]
  Error Message:
   Found new explicit static constructors in:
NLog.Internal.ArrayHelper+EmptyArray`1
  Stack Trace:
     at NLog.UnitTests.ApiTests.ShouldNotHaveExplicitStaticConstructors() in C:\projects\nlog\tests\NLog.UnitTests\ApiTests.cs:line 560
```

Guess `NLog.Internal.ArrayHelper+EmptyArray`1` should be added to the exclude-list.

Pavan8374:
Thanks for the review and the positive feedback! I've noted the minor comments. I also saw the test failure regarding  "NLog.Internal.ArrayHelper+EmptyArray`1"; interestingly, it passed on my local setup, but I will add it to the exclude list as you suggested. I'll look into all the comments and make the necessary adjustments.

src/NLog/LogFactory.cs
Pavan8374: I'm not entirely sure about the consequences of moving all the exception handling into DefaultAppEnvironment.ProcessExit. Since it's an abstraction over AppDomain.CurrentDomain.ProcessExit and DomainUnload, pushing error handling down to that level might obscure potential issues at the caller side or make it harder to trace failures during shutdown.

Would love to hear your thoughts on whether that abstraction layer is the right place to absorb and handle those exceptions, or if keeping them where they're currently handled makes more sense for clarity and debugging.

snakefoot: Already now there is error-handling inside `AppEnvironmentWrapper`, so I don't see any issue with moving this error-handling over there.

snakefoot:
Have now merged #5790 and #5791, so now `LogManager` and `FactoryHelper` should no longer have static-constructors (can be removed from the exclude list, after merge-with-rebase)

snakefoot:
Closing and reopening the pull-request to tickle the build-servers.

snakefoot:
Should I update the pull-request to remove the additional exception-logic in `LogFactory.LoggerShutdown`-eventhandler ?

Pavan8374:
Thank you so much for your review and guidance.
I apologize for the delay — I intended to update it earlier, but my schedule has been unexpectedly busy. I really appreciate your understanding.

If you prefer to merge it as is, I’m absolutely fine with that.
Otherwise, I will surely update the PR on Monday as per your suggestion.

Thank you again for your patience and support!

Pavan8374:
Hey @snakefoot, I think the unit test needs to be updated since we replaced some of the static constructors and need to remove them from the exclude list. Can I open an issue and update the test?

snakefoot:
Pull request for improvement are always welcome. You don't have to open an issue first.



Sent from my Galaxy



From: Pavan Kumar ***@***.***>
Date: 28/04/2025 07:26 (GMT+01:00)
To: NLog/NLog ***@***.***>
Cc: Rolf Kristensen ***@***.***>, Mention ***@***.***>
Subject: Re: [NLog/NLog] Removed static constructor from LogFactory (PR #5762)

[https://avatars.githubusercontent.com/u/97884852?s=20&v=4]Pavan8374 left a comment (NLog/NLog#5762)<https://github.com/NLog/NLog/pull/5762#issuecomment-2834017142>

Hey @snakefoot<https://github.com/snakefoot>, I think the unit test needs to be updated since we replaced some of the static constructors and need to remove them from the exclude list. Can I open an issue and update the test?

PR #5767, opened 2025-04-05
Marked JsonLayout EscapeForwardSlash as obsolete and without any effect

snakefoot:
Followup to #5695 for NLog v6, where EscapeForwardSlash no longer has any effect.


snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/JsonLayout

PR #5770, opened 2025-04-06
NLog.Targets.AtomicFile for atomic file append

### LINKED ISSUE #4244: ConcurrentWrites with atomic file writes on .Net5 platform
.Net5 will automatically use the NetStandard2-dependency, so ConcurrentWrites will depend on global mutex (And the global mutex can only be used between applications using the [same service-account](https://github.com/NLog/NLog/issues/4242#issuecomment-762425470)).

Could be nice if support for atomic-file-appending could be enabled for the .Net5-platform. And the global-mutex could be used across service-accounts just like on .NetFramework.

snakefoot:
Trying to resolve #4244 by supporting O_APPEND for Linux with help from `Mono.Posix.NETStandard`


snakefoot:
Created wiki: https://github.com/NLog/NLog/wiki/Atomic-File-Target

Issue #5747, opened 2025-03-24
Log4j Xml Layout for NLog.Targets.Network


snakefoot:
Change the existing Log4j Layout to rely on NLog XmlLayout, instead of using System.Xml.

ana1250:
Hi @snakefoot ,
I’d like to take this up. Could you confirm if this applies to NLog.Targets.Network only? Also, any key points I should keep in mind while implementing this?
Thanks!

snakefoot:
Make sure to take latest from `dev`-branch, as #5754 has been merged, that moved files around in NLog.Target.Network-project.

ana1250:
Hey @snakefoot ,

Since we're now fully shifting to the Log4JXmlEventLayout for generating Log4J-compatible XML logs, and all the layout responsibilities (attributes, elements, conditions, etc.) are handled within this layout class:

Would it be okay to deprecate and remove the Log4JXmlEventLayoutRenderer class?

We’d also update any references to use the layout directly where applicable, to avoid maintaining duplicate paths.

Let me know if that makes sense or if you’d prefer we keep the renderer for backward compatibility.

Thanks!

snakefoot:
Think it makes good sense to completely remove `Log4JXmlEventLayoutRenderer` is that makes your task easier for NLog v6

ana1250:
Hey @snakefoot,

I have hit a bit of a roadblock. I’ve tried multiple approaches — including pure StringBuilder, using Append, and directly calling Renderer.Render() — but I’m still running into issues with either the output not matching expectations or the custom layout (Log4JXmlEventLayout) not being used properly.

Here’s my current implementation:
https://github.com/ana1250/NLog/pull/1

I was able to get a near-working solution using StringBuilder, but it bypasses the custom layout we’re trying to support, which defeats the purpose.

At this point, I’m unsure how best to proceed. Would really appreciate your guidance on what direction I should take or if I might be missing something obvious.

Thanks in advance!

snakefoot:
Not sure how to guide you, without just doing it myself, and then use my solution as an example :)

ana1250:
Thanks for the detailed suggestions! I’ve started integrating the changes — switched over to using InnerXml, dropped the old renderer, and added support for WrapInCData (hooked this into the existing EncodeWrapper, though I am thinking about renaming it since having two wrappers might feel a bit heavy).

[...]

PR #5774, opened 2025-04-12
Log4JXmlEventLayout using Nlog XmlLayout

ana1250:
This PR replaces the usage of System.Xml with NLog’s native XML layout infrastructure (XmlLayout, XmlElement, etc.) to construct Log4j-compatible XML logs.  #5747 

System.Xml Removal: Removed System.Xml dependencies from both the Log4JXmlLayout and Log4JXmlEventLayoutParameter classes.

Renderer Updates:

- Removed the Log4jRenderer and mapped Chainsaw logs to use the new layout.
- Updated Log4JXmlEventLayout with improved rendering logic and formatting.

New Features:

- Added CDataEncode property to optionally wrap values in CDATA.
- Introduced RemoveInvalidXmlChar to strip out illegal XML characters.
- Added ThrowableEncode for escaping stack trace output. -- Removed as per review

Tests:

-  Added tests for CDATA encoding and invalid character removal.
-  Fixed and updated existing tests (Log4JXmlEventLayoutParameter spacing issue).
-  Marked outdated bad XML test as obsolete.
-  Temporarily modified layout tests for ongoing validation.
- Fixed the appinfo switch case

TODO:

Fix Scopenested rendering
Remove XML helper from Network.Target and dependent obsolete unit test


src/NLog/Internal/XmlHelper.cs
```diff
@@ -445,5 +445,19 @@ private static string EnsureDecimalPlace(string text)
...
         private static readonly char[] DecimalScientificExponent = new[] { 'e', 'E' };
+
+        public static void RemoveInvalidXmlIfNeeded(StringBuilder builder, int orgLength)
+        {
+                var cleanedText = RemoveInvalidXmlChars(builder.ToString(orgLength, builder.Length - orgLength));
+                builder.Length = orgLength;
+                builder.Append(cleanedText);
+        }
+
+        public static void WrapInCData(StringBuilder builder, int orgLength)
```
snakefoot, line 456: It doesn't look optimal to perform `builder.ToString()` if you know upfront that you will wrap using `<![CDATA`.

src/NLog/Internal/XmlHelper.cs
```diff
@@ -445,5 +445,19 @@ private static string EnsureDecimalPlace(string text)
         }
 
         private static readonly char[] DecimalScientificExponent = new[] { 'e', 'E' };
+
+        public static void RemoveInvalidXmlIfNeeded(StringBuilder builder, int orgLength)
+        {
+                var cleanedText = RemoveInvalidXmlChars(builder.ToString(orgLength, builder.Length - orgLength));
```
snakefoot, line 451: It doesn't look optimal to perform `builder.ToString()`, unless you know the builder contains invalid characters.

src/NLog.Targets.Network/Layouts/Log4JXmlEventLayout.cs
```diff
@@ -142,88 +249,67 @@ public bool IncludeScopeNested
...
-        /// Gets or sets a value indicating whether to include source info (file name and line number) in the information sent over the network.
+        /// Disable XML escaping for the log4j:throwable value. Use with caution!
         /// </summary>
-        /// <docgen category='Layout Options' order='100' />
-        public bool IncludeSourceInfo
-        {
-            get => Renderer.IncludeSourceInfo;
-            set => Renderer.IncludeSourceInfo = value;
-        }
+        public bool ThrowableEncode { get; set; } = true;
```
snakefoot, line 299: Any reason for the rename from `WriteThrowableCData` to `ThrowableEncode` ?

src/NLog/Internal/XmlHelper.cs
ana1250: public static void WrapInCData(StringBuilder builder, int orgLength)
{
    builder.Insert(orgLength, "<![CDATA[");
    builder.Append("]]>");
}
We can do this but to incorporate extra ]]> suggested by rabbit it wont be possible without ToString()
here.

But we can do this in transform section as suggested

src/NLog.Targets.Network/Layouts/Log4JXmlEventLayout.cs
snakefoot: This is the code:
```
            InnerXml.Elements.Add(new XmlElement("log4j:throwable", "${exception:format=ToString}")
            {
                CDataEncode  = WriteThrowableCData,
                Encode = ThrowableEncode,
            });
```
I think `CDataEncode = true` should win over `Encode`, so when having assigned `CDataEncode = true` then it will always wrap in CDATA. Thus there is no need to assign `Encode` from `ThrowableEncode`.

src/NLog/LayoutRenderers/Wrappers/XmlEncodeLayoutRendererWrapper.cs
```diff
@@ -67,7 +78,18 @@ public sealed class XmlEncodeLayoutRendererWrapper : WrapperLayoutRendererBase
         protected override void RenderInnerAndTransform(LogEventInfo logEvent, StringBuilder builder, int orgLength)
         {
             Inner.Render(logEvent, builder);
-            if (XmlEncode)
+
+            if (RemoveInvalidXml)
```
snakefoot, line 82: Think this should be implicit when `XmlEncode = true`, without needing an extra option.

src/NLog/LayoutRenderers/Wrappers/XmlEncodeLayoutRendererWrapper.cs
```diff
@@ -67,7 +78,18 @@ public sealed class XmlEncodeLayoutRendererWrapper : WrapperLayoutRendererBase
...
+            {
+                XmlHelper.RemoveInvalidXmlIfNeeded(builder, orgLength);
+            }
+
+            if (CDataEncode)
+            {
+                XmlHelper.WrapInCData(builder, orgLength);
```
snakefoot, line 89: Think the logic should be changed, so you do this:
```
protected override void RenderInnerAndTransform(LogEventInfo logEvent, StringBuilder builder, int orgLength)
{
      if (CDataEncode)
      {
           builder.Append("<![CDATA[");
      }

      Inner.Render(logEvent, builder);
 
      if (CDataEncode)
      {
           builder.Append("]]>");
      }
      else if (XmlEncode)
      {
           XmlHelper.RemoveInvalidXmlIfNeeded(builder, orgLength);
      }
}
```

src/NLog/Internal/XmlHelper.cs
snakefoot: I have tried to make a suggestion for how to implement the logic in `XmlEncodeLayoutRendererWrapper.cs`

src/NLog/Internal/XmlHelper.cs
ana1250: public static void RemoveInvalidXmlIfNeeded(StringBuilder builder, int orgLength)
        {
            var text = builder.ToString(orgLength, builder.Length - orgLength);

#if !NET35
            if (ContainsInvalidXmlChars(text))
            {
                var cleanedText = RemoveInvalidXmlChars(text);
                builder.Length = orgLength;
                builder.Append(cleanedText);
            }
#else
    var cleanedText = RemoveInvalidXmlChars(text);
    builder.Length = orgLength;
    builder.Append(cleanedText);
#endif
        }
        
                private static bool ContainsInvalidXmlChars(string text)
        {
#if !NET35
            for (int i = 0; i < text.Length; ++i)
            {
                if (!XmlConvert.IsXmlChar(text[i]))
                {
                    return true;
                }
            }
            return false;
#else
    return InvalidXmlChars.IsMatch(text);
#endif
        }
        
 At first i thought in the RemoveInvalidXmlChars method we are already checking once so used directly but checking first for invalid then manipulating like above implementation would work?

src/NLog.Targets.Network/Layouts/Log4JXmlEventLayout.cs
ana1250: YEs makes sense i added it for testing purpose ,we can remove it.

src/NLog/Internal/XmlHelper.cs
ana1250: Yes will check that thanks!

src/NLog/Internal/XmlHelper.cs
snakefoot: I was hoping to avoid the `var text = builder.ToString(orgLength, builder.Length - orgLength);` unless having confirmed that the `builder` contains invalid-xml-chars.

src/NLog/LayoutRenderers/Wrappers/XmlEncodeLayoutRendererWrapper.cs
ana1250: Yea i see ,but if we have cdata as true and then we wont be able to clear the invalid xml ,
```
protected override void RenderInnerAndTransform(LogEventInfo logEvent, StringBuilder builder, int orgLength)
{
    if (CDataEncode)
    {
        builder.Append("<![CDATA[");
    }

    Inner.Render(logEvent, builder);

    if (CDataEncode)
    {
        builder.Append("]]>");  // TODO: Handle case where inner text contains "]]>"
    }

    // Always ensure the XML is valid
    XmlHelper.RemoveInvalidXmlIfNeeded(builder, orgLength);

    // Only perform XML escaping when CDATA is not used
    if (!CDataEncode && XmlEncode)
    {
        XmlHelper.XmlEscapeIfNeeded(builder, orgLength);
    }
}
```
Updated the method to always call RemoveInvalidXmlIfNeeded, even when CDataEncode is enabled.

Although CDATA skips XML escaping, it does not skip XML validity (e.g., invalid control characters are still not allowed in XML even inside CDATA).

Therefore, we now first wrap with CDATA if needed, render the inner content, always remove invalid XML characters after, and finally perform XML escaping if required when CDATA is not used.

src/NLog/LayoutRenderers/Wrappers/XmlEncodeLayoutRendererWrapper.cs
ana1250: Added comment in support to this in review comment of same block.

src/NLog/Internal/XmlHelper.cs
ana1250: public static void RemoveInvalidXmlIfNeeded(StringBuilder builder, int orgLength)
{
#if !NET35
    bool containsInvalid = false;
    for (int i = orgLength; i < builder.Length; ++i)
    {
        if (!XmlConvert.IsXmlChar(builder[i]))
        {
            containsInvalid = true;
            break;
        }
    }

    if (containsInvalid)
    {
        var text = builder.ToString(orgLength, builder.Length - orgLength);
        var cleanedText = RemoveInvalidXmlChars(text);
        builder.Length = orgLength;
        builder.Append(cleanedText);
    }
#else
    var text = builder.ToString(orgLength, builder.Length - orgLength);
    var cleanedText = RemoveInvalidXmlChars(text);
    builder.Length = orgLength;
    builder.Append(cleanedText);
#endif
}

maybe like this?

src/NLog/LayoutRenderers/Wrappers/XmlEncodeLayoutRendererWrapper.cs
snakefoot: Think your method looks good.

src/NLog/Internal/XmlHelper.cs
snakefoot: Think your suggested method-implementation looks good.

src/NLog/LayoutRenderers/Wrappers/XmlEncodeLayoutRendererWrapper.cs
ana1250: If we want to corporate rabbit suggestion
public static void WrapInCData(StringBuilder builder, int orgLength)
{
    var content = builder.ToString(orgLength, builder.Length - orgLength)
                         .Replace("]]>", "]]]]><![CDATA[>");
    builder.Length = orgLength;
    builder.Append("<![CDATA[")
           .Append(content)
           .Append("]]>");
}

builder to string would be required

like we doing in transform
 public static string WrapInCData(string text)
 {
     if (string.IsNullOrEmpty(text))
         return "<![CDATA[]]>";

     if (text.Contains("]]>"))
         text = text.Replace("]]>", "]]]]><![CDATA[>");

     return $"<![CDATA[{text}]]>";
 }
 
 Existing test cases for log4j are passing as of now (without the rare condition of rabbit one)

snakefoot: Maybe something like this:
```c#
protected override void RenderInnerAndTransform(LogEventInfo logEvent, StringBuilder builder, int orgLength)
{
    if (CDataEncode)
    {
        builder.Append("<![CDATA[");
        orgLength = builder.Length;  // Notice
    }

    Inner.Render(logEvent, builder);

    if (CDataEncode)
    {
        // TODO Move this into a method with a good name
        for (int i = orgLength; i < builder.Length; ++i)
        {
           if (builder[i] == ']`)
           {
              var content = builder.ToString(orgLength, builder.Length - orgLength)
                         .Replace("]]>", "]]]]><![CDATA[>");
              builder.Length = orgLength;
              builder.Append(content);
              break;
           }
        }
        builder.Append("]]>");
    }

    // Always ensure the XML is valid
    XmlHelper.RemoveInvalidXmlIfNeeded(builder, orgLength);

    // Only perform XML escaping when CDATA is not used
    if (!CDataEncode && XmlEncode)
    {
        XmlHelper.XmlEscapeIfNeeded(builder, orgLength);
    }
}
```

ana1250: Yes applied thanks for suggestion!

src/NLog/LayoutRenderers/Wrappers/XmlEncodeLayoutRendererWrapper.cs
```diff
@@ -76,6 +98,11 @@ protected override void RenderInnerAndTransform(LogEventInfo logEvent, StringBui
         /// <inheritdoc/>
         protected override string Transform(string text)
         {
+            if (CDataEncode)
+            {
+                return $"<![CDATA[{text}]]>";
```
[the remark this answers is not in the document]
ana1250: Fixed it in latest commit

src/NLog/Internal/XmlHelper.cs
```diff
@@ -445,5 +445,73 @@ private static string EnsureDecimalPlace(string text)
...
+                }
+            }
+
+            if (needsFix)
+            {
+                for (int i = orgLength; i + 2 < builder.Length; ++i)
+                {
+                    if (builder[i] == ']' && builder[i + 1] == ']' && builder[i + 2] == '>')
+                    {
+                        builder.Remove(i, 3);
```
snakefoot, line 496: `StringBuilder.Remove` hurts performance and it is actually better to just do this:
```c#
        public static void EnsureValidCDataContent(StringBuilder builder, int orgLength)
        {
            for (int i = orgLength; i + 2 < builder.Length; ++i)
            {
                if (builder[i] == ']' && builder[i + 1] == ']' && builder[i + 2] == '>')
                {
                   var content = builder.ToString(i, builder.Length - i)
                         .Replace("]]>", "]]]]><![CDATA[>");
                   builder.Length = orgLength;
                   builder.Append(content);
                   break;
                }
            }

            builder.Append("]]>");
        }
```

src/NLog.Targets.Network/Layouts/Log4JXmlEventLayout.cs
```diff
@@ -55,58 +52,165 @@ namespace NLog.Layouts
...
-            set => Renderer.IncludeScopeProperties = value;
-        }
+        public bool IncludeScopeProperties { get => _includeScopeProperties ?? (_includeMdlc == true || _includeMdc == true); set => _includeScopeProperties = value; }
+        private bool? _includeScopeProperties;
+
+        /// <summary>
+        ///  Gets or sets whether the log4j:throwable xml-element should be written as CDATA
+        /// </summary>
+        /// <docgen category='Layout Options' order='50' />
+        public bool WriteThrowableCData { get; set; }
```
snakefoot, line 176: Could this be moved down just below the `Layout AppInfo`-property, where it was before?

src/NLog/LayoutRenderers/Wrappers/XmlEncodeLayoutRendererWrapper.cs
```diff
@@ -66,16 +71,36 @@ public sealed class XmlEncodeLayoutRendererWrapper : WrapperLayoutRendererBase
...
-            if (XmlEncode)
+
+            if (CDataEncode)
+            {
+                XmlHelper.EnsureValidCDataContent(builder, orgLength);
+            }
+
+            XmlHelper.RemoveInvalidXmlIfNeeded(builder, orgLength);
+
+            if (!CDataEncode && XmlEncode)
```
snakefoot, line 89: Could the method be changed into:
```c#
        protected override void RenderInnerAndTransform(LogEventInfo logEvent, StringBuilder builder, int orgLength)
        {
            if (CDataEncode)
            {
                builder.Append("<![CDATA[");
                orgLength = builder.Length; 
            }

            Inner.Render(logEvent, builder);

            XmlHelper.RemoveInvalidXmlIfNeeded(builder, orgLength);

            if (CDataEncode)
            {
                XmlHelper.EnsureValidCDataContent(builder, orgLength);
            }
            else if (XmlEncode)
            {
                XmlHelper.PerformXmlEscapeWhenNeeded(builder, orgLength, XmlEncodeNewlines);
            }
        }
```

snakefoot:
Think the pull-request is super close to being ready for merging.

Thank you for being patient with all my suggestions. I hope my last comments are acceptable.

src/NLog/Internal/XmlHelper.cs
ana1250: Yea makes sense ,will do that.

src/NLog.Targets.Network/Layouts/Log4JXmlEventLayout.cs
ana1250: Yes definitely.

src/NLog/LayoutRenderers/Wrappers/XmlEncodeLayoutRendererWrapper.cs
ana1250: Yea i thought that too,will do that.

ana1250:
> Think the pull-request is super close to be ready for merge.
> 
> Thank you for being patient with all my comments. I hope my last suggestions are acceptable.

No worries at all! Thanks a ton for all the guidance — your feedback really helped me improve the code a lot. Appreciate you being patient too!

snakefoot:
I think the code looks great, and super happy that the NLog XmlLayout was able to lift the task. And thank for adding the the new feature `CDataEncode`.

Do you have any more changes planned? Or should I just merge when the build-servers approves? (Then I can fiddle with getting the separator-option working again)

ana1250:
Yup, nothing more from my side! Merge away once the build is happy.

snakefoot:
Closing and opening pull-request to tickle the build-servers

PR #5779, opened 2025-04-12
NetworkTarget - Introduced option NoDelay to disable delayed ACK

### LINKED ISSUE #5778: NetworkTarget with Socket.NoDelay to disable TCP_NODELAY
TCP_NODELAY disables the nagle-algorithm and delayed ACK, which removes a delay of 200ms for the first message to reduce ACKS.

Nagle-algorithm is important for high-latency / low-bandwith network-connections, but now most have fiber-connection or better.

Believe most people will not notice the overhead of extra ACKS, but will notice the 200ms delay.

Socket.NoDelay should be assigned to `true` by default, but should of course be configurable.

The feature should also be implemented for NLog v5.5 but with default value `false`.

snakefoot:
Resolves #5778


snakefoot:

PR #5780, opened 2025-04-12
NetworkTarget - Introduced option NoDelay to disable delayed ACK

snakefoot:
See also #5778

- [x] Wait for NLog v5.5


snakefoot:

PR #5783, opened 2025-04-24
NLog.Targets.GZipFile for writing GZip Log Files

snakefoot:
Using dotnet `GZipStream` for writing directly to GZip File. Alternative to the old EnableArchiveFileCompression = true.

The `EnableArchiveFileCompression` had the side-effect that it would stall other file-logging, until file-compression of old file had completed (slow when large log-file)


snakefoot:
Created wiki: https://github.com/NLog/NLog/wiki/GZip-File-Target

PR #5794, opened 2025-04-26
Layout Render without caching, as handled by Precalculate

snakefoot:
Avoid the pitfall of constantly having to avoid cache-capture with with NLog v6, since it is only the NLog Precalculation-Logic that should perform capture, and everywhere else capture should be avoided. If there are situations where caching in `Render` was needed to make async Layout-Precalculate work, then it should be fixed by improving the Layout-Precalculate-implementation.

Follow up to #5758

Custom Layout-classes that override `Precalculate`-method to perform conditional-capture, must now call `base.Precalculate()` to perform layout-value-capture, and not depend on `Render()`.


snakefoot:
Maybe one should completely remove this pitfall with NLog v6, since it is only the NLog Precalculation-Logic that should perform capture, and everywhere else it should be no capture by default.

PR #5796, opened 2025-04-26
Log4JXmlEventLayout - Inherit from CompoundLayout to optimize for async

snakefoot:
Followup to #5774

Allow to offload most of the rendering to Async-Background-Thread, and only use the application-thread for minimal capture.


ana1250:
I had tried several approaches earlier, but kept running into issues like 'Separator assigned twice' or 'no default property' errors. I noticed you were able to fix the ScopeNestedSeparator handling cleanly in this PR. Could you briefly explain what the key change or trick was that made it work?

snakefoot:
It is because when parsing NLog-Layouts that contains LayoutRenderers, then there is special syntax where one use `:` as option-separator. Thus if wanting to assign a value that contains the special-value `:`, then one must escape the special-value first.

See also: https://github.com/NLog/NLog/wiki/SimpleLayout

Issue #5537, opened 2024-06-05
Support params ReadOnlySpan with NetStandard2.1 and lang version 13


snakefoot:
Avoid params-array-allocation upfront when LogLevel is not enabled for logging.

PR #5800, opened 2025-05-02
Logger with support for params ReadOnlySpan

snakefoot:
Skip params `object[]`-allocation when LogLevel is not enabled.

First step for resolving #5537 (Still missing the logic to skip `object[]`-allocation when message-template with properties like with Microsoft Extension Logging)

Changed NETSTANDARD2.1 to use LangVersion=13 to support `params ReadOnlySpan` and `OverloadResolutionPriority`:

- https://devblogs.microsoft.com/dotnet/csharp13-calling-methods-is-easier-and-faster/
- https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-13.0/overload-resolution-priority

Maybe the `[OverloadResolutionPriority(-1)]` can be a long-game approach for cleaning up the V1 Logger-API with less users affected by the breaking changes (Can cleanup when NET8 has reached end-of-support, some time in 2027)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -57,6 +57,7 @@ for:
         - image: Ubuntu2204
     environment:
       DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
+      MSBUILDTERMINALLOGGER: off
       FrameworkPathOverride: /usr/lib/mono/4.6.1-api/
     build_script:
     - ps: dotnet --version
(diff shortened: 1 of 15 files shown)


snakefoot:
LangVersion=13 is only supported with NET9, so had to update AppVeyor from Ubuntu2004 to Ubuntu2204 for NET9-support.

But the console-output from xUnit-tests looks garbled after the change to Ubuntu2204.

PR #5831, opened 2025-05-23
DumpApiXml - Scan multiple assemblies

### LINKED ISSUE #1168: XSD only contains targets from NLog package
The XSD currently only contains the targets from the NLog package and not extension packages, like NLog.Web and NLog.Windows.Forms.

Maybe we need more XSD files for the other extensions?

See https://github.com/NLog/NLog/issues/1166

snakefoot:
With NLog v6 then it becomes more relevant to merge XML-docs from multiple NLog-extension-assemblies.

Resolves #1168

### DIFF
--- a/src/NLog.proj
+++ b/src/NLog.proj
@@ -103,6 +103,7 @@
   </Target>
 
   <Target Name="BuildApiDumpTool">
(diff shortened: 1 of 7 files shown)


snakefoot:
Seems `NLog.Windows.Forms` exposes unwanted public properties, that need to put under protection.

- https://github.com/NLog/NLog.Windows.Forms/pull/190

PR #5834, opened 2025-05-24
LogEventInfo SequenceID marked obsolete, instead use CounterLayoutRenderer

snakefoot:
Another step towards resolving #4159, where `${sequenceid}` can be replaced by `${counter:sequence=global}`

And added a little sugar by optimizing `${counter}` to use Interlocked and ConcurrentDictionary


snakefoot:
Updated wiki with alternatives: https://github.com/NLog/NLog/wiki/SequenceId-layout-renderer

PR #5850, opened 2025-05-30
LogEventInfo - New constructor with ReadOnlySpan MessageTemplateParameter

snakefoot:


snakefoot:
Reverted again with #5851, since stackalloc only supports unsafe-struct.

PR #5915, opened 2025-07-03
ReplaceNewLinesLayoutRendererWrapper - Replace all line ending characters

oikku:
Windows uses \r\n, Linux and new Macs use \n. Older Macs use \r. Unicode adds more line endings. Old implementation didn't even replace single \r characters. Text editors like notepad++ treat single \r as new line character and log files aren't formatted as expected when using newline replacements.

Line ending characters replaced are those defined in unicode 16.0.0 specification section 5.8 Newline guidelines in table 5-1.

### DIFF
--- a/src/NLog/LayoutRenderers/Wrappers/ReplaceNewLinesLayoutRendererWrapper.cs
+++ b/src/NLog/LayoutRenderers/Wrappers/ReplaceNewLinesLayoutRendererWrapper.cs
@@ -84,29 +84,41 @@ namespace NLog.LayoutRenderers.Wrappers
                     orgLength = lineEndIndex > orgLength ? lineEndIndex - 1 : orgLength;
                     string str = builder.ToString(orgLength, builder.Length - orgLength);
                     builder.Length = orgLength;
-                    int current = 0;
-                    int index = IndexOfLineEndCharacters(str, 0);
-                    while (index > -1)
+
+                    bool ignoreNewLine = false;
+                    foreach (var chr in str)
                     {
-                        builder.Append(str.Substring(current, index - current));
-                        builder.Append(Replacement);
-                        if (index < str.Length - 1 && str[index] == '\r' && str[index + 1] == '\n')
+                        if (IsLineEndCharacter(chr))  // switches on chr, and allow method-reuse
                         {
-                            index += 2;
+                            if (!ignoreNewLine || chr != '\n')
+                            {
+                                builder.Append(Replacement);
+                            }
+                            ignoreNewLine = chr == '\r';
                         }
                         else
                         {
-                            index++;
+                            builder.Append(chr);
+                            ignoreNewLine = false;
                         }
-                        current = index;
-                        index = IndexOfLineEndCharacters(str, index);
-                    }
-                    if (current < str.Length - 1)
-                    {
-                        builder.Append(str.Substring(current));
                     }
                 }
             }
+        }
+
+        private static bool IsLineEndCharacter(char ch)
+        {
+            switch (ch)
+            {
+                case '\r':
+                case '\n':
+                case '\u0085':
+                case '\u2028':
+                case '\u000C':
+                case '\u2029':
+                    return true;
+            }
+            return false;
         }
 
         private static int IndexOfLineEndCharacters(string builder, int startPosition)
--- a/src/NLog/LayoutRenderers/Wrappers/ReplaceNewLinesLayoutRendererWrapper.cs
+++ b/src/NLog/LayoutRenderers/Wrappers/ReplaceNewLinesLayoutRendererWrapper.cs
@@ -102,10 +102,10 @@ namespace NLog.LayoutRenderers.Wrappers
             {
                 case '\r':
                 case '\n':
-                case '\u0085':
-                case '\u2028':
-                case '\u000C':
-                case '\u2029':
+                case '\u000C':  // FORM FEED (FF)
+                case '\u0085':  // NEXT LINE (NEL)
+                case '\u2028':  // LINE SEPARATOR
+                case '\u2029': // PARAGRAPH SEPARATOR
                     return true;
             }
             return false;
--- a/src/NLog/LayoutRenderers/Wrappers/ReplaceNewLinesLayoutRendererWrapper.cs
+++ b/src/NLog/LayoutRenderers/Wrappers/ReplaceNewLinesLayoutRendererWrapper.cs
@@ -100,12 +100,12 @@ namespace NLog.LayoutRenderers.Wrappers
         {
             switch (ch)
             {
-                case '\r':
-                case '\n':
-                case '\u000C':  // FORM FEED (FF)
+                case '\r':      // CARRIAGE RETURN
+                case '\n':      // LINE FEED
+                case '\f':      // FORM FEED (FF)
                 case '\u0085':  // NEXT LINE (NEL)
                 case '\u2028':  // LINE SEPARATOR
-                case '\u2029': // PARAGRAPH SEPARATOR
+                case '\u2029':  // PARAGRAPH SEPARATOR
                     return true;
             }
             return false;
--- a/src/NLog/LayoutRenderers/Wrappers/ReplaceNewLinesLayoutRendererWrapper.cs
+++ b/src/NLog/LayoutRenderers/Wrappers/ReplaceNewLinesLayoutRendererWrapper.cs
@@ -34,6 +34,7 @@
 namespace NLog.LayoutRenderers.Wrappers
 {
     using System;
+    using System.Linq;
     using System.Text;
     using NLog.Config;
     using NLog.Internal;
@@ -51,9 +52,6 @@ namespace NLog.LayoutRenderers.Wrappers
     [ThreadAgnostic]
     public sealed class ReplaceNewLinesLayoutRendererWrapper : WrapperLayoutRendererBase
     {
-        private const string WindowsNewLine = "\r\n";
-        private const string UnixNewLine = "\n";
-
         /// <summary>
         /// Gets or sets a value indicating the string that should be used to replace newlines.
         /// </summary>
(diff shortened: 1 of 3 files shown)


src/NLog/LayoutRenderers/Wrappers/ReplaceNewLinesLayoutRendererWrapper.cs
```diff
@@ -64,33 +62,51 @@ public string Replacement
...
         private string _replacement = " ";
-        private bool _replaceWithNewLines;
 
         /// <summary>
         /// Gets or sets a value indicating the string that should be used to replace newlines.
         /// </summary>
         /// <docgen category='Layout Options' order='50' />
         public string ReplaceNewLines { get => Replacement; set => Replacement = value; }
 
+        private static readonly char[] LineEndCharacters = new char[] { '\u000D', '\u000A', '\u0085', '\u2028', '\u000C', '\u2029' };
```
snakefoot: Instead of using unicode-numbers directly, then I prefer the more human-readable standard chars when possible. Ex `\n` + `\r` + etc.

src/NLog/LayoutRenderers/Wrappers/ReplaceNewLinesLayoutRendererWrapper.cs
```diff
@@ -64,33 +62,51 @@ public string Replacement
...
+                var containsNewLines = builder.IndexOfAny(LineEndCharacters, orgLength) >= 0;
                 if (containsNewLines)
                 {
-                    string str = builder.ToString(orgLength, builder.Length - orgLength)
-                                        .Replace(WindowsNewLine, _replaceWithNewLines ? UnixNewLine : Replacement)
-                                        .Replace(UnixNewLine, Replacement);
-
+                    string str = builder.ToString(orgLength, builder.Length - orgLength);
+                    // If replacement is longer than single character then reserve some extra capacity in StringBuilder to avoid some resize operations.
+                    var sb = new StringBuilder(str.Length + (Replacement.Length <= 1 ? 0 : 10 * Replacement.Length));
```
snakefoot: Would it be possible that you could write directly to the existing `builder` instead of creating a new StringBuilder ?
```csharp
string str = builder.ToString(orgLength, builder.Length - orgLength);
builder.Length = orgLength;
bool ignoreNewLine = false;
foreach (var chr in str)
{
      switch (chr)
      {
          case '\r':
               builder.Append(Replacement);
               ignoreNewLine = true;
               break;
          case '\n':
               if (!ignoreNewLine)
                   builder.Append(Replacement);
               ignoreNewLine = false;
               break;
          default:
               builder.Append(chr);
               ignoreNewLine = false;
               break;
      }
}
```
Would think that switch-statement is faster than Linq-Contains.

Would think that foreach is faster than enumerating with for-loop.

snakefoot: Maybe one could make it even faster by using this StringBuilder-method
```csharp
public System.Text.StringBuilder Append(string value, int startIndex, int count);
```
Instead of adding single char at a time using `builder.Append(chr)`.

src/NLog/LayoutRenderers/Wrappers/ReplaceNewLinesLayoutRendererWrapper.cs
```diff
@@ -82,19 +78,72 @@ protected override void RenderInnerAndTransform(LogEventInfo logEvent, StringBui
...
-                                        .Replace(UnixNewLine, Replacement);
-
+                    string str = builder.ToString(orgLength, builder.Length - orgLength);
                     builder.Length = orgLength;
-                    builder.Append(str);
+                    int current = 0;
+                    int index = IndexOfLineEndChar(str, 0);
+                    while (index > -1)
+                    {
+                        builder.Append(str.Substring(current, index - current));
```
snakefoot: `str.SubString` will allocate a string, and will hurt performance. My suggestion about maybe using `StringBuilder.Append(string value, int startIndex, int count)` was about improving performance (by reducing the number of StringBuilder-Append-calls).

Maybe it is simpler to just do this:
```csharp
string str = builder.ToString(orgLength, builder.Length - orgLength);
builder.Length = orgLength;
bool ignoreNewLine = false;
foreach (var chr in str)
{
      if (IsLineEndChar(chr))  // switches on chr, and allow method-reuse
      {
           if (!ignoreNewLine || chr != '\n')
           {
              builder.Append(Replacement);
           }
           ignoreNewLine = chr == '\r';
      }
      else
      {
           builder.Append(chr);
           ignoreNewLine = false;
      }
}
```

src/NLog/LayoutRenderers/Wrappers/ReplaceNewLinesLayoutRendererWrapper.cs
```diff
@@ -82,19 +78,72 @@ protected override void RenderInnerAndTransform(LogEventInfo logEvent, StringBui
             Inner?.Render(logEvent, builder);
             if (builder.Length > orgLength)
             {
-                var containsNewLines = builder.IndexOf('\n', orgLength) >= 0;
-                if (containsNewLines)
+                if (ContainsLineEndCharacters(builder, orgLength))
```
snakefoot: Maybe `ContainsLineEndCharacters` could be renamed to `IndexOfLineEndCharacters` ?
```charp
var lineEndIndex = IndexOfLineEndCharacters(builder, orgLength);
if (lineEndIndex >= 0)
{
       orgLength = lineEndIndex > orgLength ? lineEndIndex - 1 : orgLength;
       string str = builder.ToString(orgLength, builder.Length - orgLength);
}
```
Thus only allocate a string that contains the part containing the newline-char.

oikku: I ran some benchmarks. Latest version seems to be fastest and allocates only one string instance from original builder at the start.

| Method                                    | Mean     | Error    | StdDev   | Median   |
|------------------------------------------ |---------:|---------:|---------:|---------:|
| FirstVersion                              | 52.11 us | 2.358 us | 6.727 us | 50.80 us |
| SubstringAppend                           | 24.58 us | 0.794 us | 2.201 us | 23.60 us |
| InlineReplaceInOriginalBuilder            | 32.22 us | 1.331 us | 3.818 us | 30.70 us |
| AppendWithIsLineEndCharacterIgnoreNewLine | 23.05 us | 1.051 us | 2.983 us | 21.80 us |

src/NLog/LayoutRenderers/Wrappers/ReplaceNewLinesLayoutRendererWrapper.cs
```diff
@@ -82,19 +78,86 @@ protected override void RenderInnerAndTransform(LogEventInfo logEvent, StringBui
...
+                case '\u0085':
+                case '\u2028':
+                case '\u000C':
+                case '\u2029':
+                    return true;
+            }
+            return false;
+        }
+
+        private static int IndexOfLineEndCharacters(string builder, int startPosition)
```
snakefoot: Please remove this unused method.

src/NLog/LayoutRenderers/Wrappers/ReplaceNewLinesLayoutRendererWrapper.cs
```diff
@@ -82,19 +78,86 @@ protected override void RenderInnerAndTransform(LogEventInfo logEvent, StringBui
...
+                }
+            }
+            return -1;
+        }
+
+        private static int IndexOfLineEndCharacters(StringBuilder builder, int startPosition)
+        {
+            for (int i = startPosition; i < builder.Length; i++)
+            {
+                switch (builder[i])
```
snakefoot: Please update to call `IsLineEndCharacter`.

src/NLog/LayoutRenderers/Wrappers/ReplaceNewLinesLayoutRendererWrapper.cs
snakefoot: Thank you for doing the benchmark-test. Actual numbers are better than me guessing.

src/NLog/LayoutRenderers/Wrappers/ReplaceNewLinesLayoutRendererWrapper.cs
```diff
@@ -64,11 +60,9 @@ public string Replacement
             set
             {
                 _replacement = value;
-                _replaceWithNewLines = value?.IndexOf('\n') >= 0;
```
snakefoot, line 67: Since special logic has been removed, then it can be changed to normal property:
```
public string Replacement { get; set; } = " ";
```
And remove the old backing-field `_replacement`.

src/NLog/LayoutRenderers/Wrappers/ReplaceNewLinesLayoutRendererWrapper.cs
```diff
@@ -82,19 +68,62 @@ protected override void RenderInnerAndTransform(LogEventInfo logEvent, StringBui
...
+                        {
+                            builder.Append(chr);
+                            ignoreNewLine = false;
+                        }
+                    }
+                }
+            }
+        }
+
+        private static bool IsLineEndCharacter(char ch)
```
snakefoot, line 99:
```csharp
        private static bool IsLineEndCharacter(char ch)
        {
            switch (ch)
            {
                case '\r':      // CARRIAGE RETURN
                case '\n':      // LINE FEED
                case '\f':      // FORM FEED (FF)
                case '\u0085':  // NEXT LINE (NEL)
                case '\u2028':  // LINE SEPARATOR
                case '\u2029':  // PARAGRAPH SEPARATOR
                    return true;
            }
            return false;
        }
```
Believe this will be my last nitpick. Think the updated code is a good improvement, and thank you again for the benchmark.

snakefoot (approved):
Thank you so much for being patient with me and all my comments.

Very happy about the result, and think it is a good improvement together with the new unit-tests.

snakefoot:
Closing an re-opening the pull-request to tickle the build-servers.

snakefoot:
NLog v6.0.2 has been released:

- https://www.nuget.org/packages/NLog/6.0.2#releasenotes-body-tab

Thank you again for the contribution

Issue #5920, opened 2025-07-05
FileTarget with ArchiveSuffixFormat doesn't format the date as expected


13xforever:
**NLog version**: 6.0.1

**Platform**: .NET9 on Windows 11 24H2 and Ubuntu 24.04

**Current NLog config** (xml or C#, if relevant)

```csharp
var fileTarget = new FileTarget("logfile") {
            FileName = Path.GetFullPath(Path.Combine(LogPath, "bot.log")),
            ArchiveEvery = FileArchivePeriod.Day,
            ArchiveSuffixFormat = ".{1:yyyyMMdd}.{0:0}",
            MaxArchiveDays = 30,
            KeepFileOpen = true,
            AutoFlush = false,
            OpenFileFlushTimeout = 1,
            Layout = "…",
};
```

- What is the current result?
   - On Windows: `bot..1.log`
   - On Linux: `bot..250705.log` (???)
- What is the expected result?
   I'd expect to see `bot.20250705.0.log` on all platforms
- Did you check the [Internal log](https://github.com/NLog/NLog/wiki/Internal-Logging)?
   No
- Please post full exception details (message, stacktrace, inner exceptions)
- Are there any workarounds? yes/no
- Is there a version in which it worked? 
  I just want consistent naming after upgrading from v5
- Can you help us by writing a unit test?

snakefoot:
Yes that sounds like a bug on Linux. `ArchiveSuffixFormat` provides 2 placeholders `{0}` and `{1}` where the later place-holder should only work when also specifying the legacy `ArchiveFileName`.

Maybe try this:
```c#
var fileTarget = new FileTarget("logfile") {
            FileName = Path.GetFullPath(Path.Combine(LogPath, "bot.log")),
            ArchiveFileName = Path.GetFullPath(Path.Combine(LogPath, "bot.log")), // Activate legacy `File.Move`-logic
            ArchiveEvery = FileArchivePeriod.Day,
            ArchiveSuffixFormat = ".{1:yyyyMMdd}.{0:0}",
            MaxArchiveDays = 30,
            KeepFileOpen = true,
            AutoFlush = false,
            OpenFileFlushTimeout = 1,
            Layout = "…",
};
```
See also: [NLog FileTarget and ArchiveSuffixFormat](https://nlog-project.org/2025/04/29/nlog-6-0-major-changes.html#nlog-filetarget-and-archivesuffixformat)

Will try to figure out why Linux does not behave as intended.

[...]

PR #5921, opened 2025-07-06
FileTarget - Activate legacy ArchiveFileName when ArchiveSuffixFormat contains legacy placeholder

snakefoot:
See also: #5920

### DIFF
--- a/src/NLog/Targets/FileTarget.cs
+++ b/src/NLog/Targets/FileTarget.cs
@@ -296,7 +296,7 @@ namespace NLog.Targets
         /// <docgen category='Archival Options' order='50' />
         public Layout? ArchiveFileName
         {
-            get => _archiveFileName;
+            get => _archiveFileName ?? (_archiveSuffixFormat?.IndexOf("{1") >= 0 ? FileName : null);
             set
             {
                 var archiveSuffixFormat = _archiveSuffixFormat;
@@ -422,6 +422,7 @@ namespace NLog.Targets
             {
                 if (!string.IsNullOrEmpty(value) && ArchiveFileName is SimpleLayout simpleLayout)
                 {
+                    // When legacy ArchiveFileName only contains file-extension, then strip away leading underscore from suffix
                     var fileName = Path.GetFileNameWithoutExtension(simpleLayout.OriginalText);
                     if (StringHelpers.IsNullOrWhiteSpace(fileName) && value.IndexOf('_') == 0)
                     {


hhyyrylainen:
Was this released already? I just got the update to NLog 6.0 and I had to immediately go on GitHub to find out why the XML documentation example on `ArchiveSuffixFormat` does not work correctly. Turns out that despite `ArchiveFileName` saying it shouldn't be used (`Avoid using ArchiveFileName when possible, and instead rely on only using FileName and ArchiveSuffixFormat.`) I need to specify that to be the exact same string as `FileName` for the date suffix example to work...

If I understood correctly this change would have saved me in this situation.

snakefoot:
@hhyyrylainen Yes sorry about the confusion. I would like to warn people that legacy `ArchiveFileName` with `File.Move` can sometimes fail because of file-locks, but I also recognize a lot of critical applications depend on the existing behavior from NLog v4/v5. It is perfectly fine to use legacy/old archive-mode when accepting the risk that NLog can fail to perform archive-roll because of file-locks, and even if archive-roll fails then NLog will continue to be working/logging.

This pull-request will automatically enable legacy archive-mode, when seeing `ArchiveSuffixFormat` contains `{1}`-place-holder. As other NLog user was confused that using legacy `{1}`-place-holder didn't automatically activate legacy-archive-mode. This skips the requirement of explicit assigning `ArchiveFileName`.

Any feedback is most welcome about the major rewrite of the NLog FileTarget in NLog v6. Both in behavior and in documentation. Again the goal is to make the transition from NLog v5 to NLog v6 as smooth as possible.

hhyyrylainen:
That's good to hear.

I'd personally love to see a migration guide to NLog 6, and how to do it properly. In my use case I just want old log files (or if the current log is too big) to be named with a date when they are rotated out, but it is not clear to me what I'd need to do to achieve that without using any of the legacy or marked obsolete options.

For reference this is the new config I ended up with (and it seems to now work as it did before):
```c#
var fileTarget = new AtomicFileTarget
{
    FileName = $"{basePath}/thrive-launcher-log.txt",
    ArchiveAboveSize = GlobalConstants.MEBIBYTE * 2,
    ArchiveEvery = FileArchivePeriod.Month,
    ArchiveFileName = $"{basePath}/thrive-launcher-log.txt",
    ArchiveSuffixFormat = ".{1:yyyy-MM-dd}",
    MaxArchiveFiles = 4,
    Encoding = Encoding.UTF8,
    KeepFileOpen = true,
    ConcurrentWrites = true,
    LineEnding = LineEndingMode.Default,
};
```

snakefoot:
Think your FileTarget-configuration looks valid. Maybe consider removing properties that just assigns default-values in NLog v5 / v6:
```csharp
var fileTarget = new AtomicFileTarget
{
    FileName = $"{basePath}/thrive-launcher-log.txt",
    ArchiveFileName = $"{basePath}/thrive-launcher-log.txt",
    ArchiveAboveSize = GlobalConstants.MEBIBYTE * 2,
    ArchiveEvery = FileArchivePeriod.Month,
    ArchiveSuffixFormat = ".{1:yyyy-MM-dd}",
    MaxArchiveFiles = 4,
    ConcurrentWrites = true,
};
```
I'm not sure how to provide a simple guide that works for everyone. Right now have the very simple example shown in [NLog FileTarget and ArchiveSuffixFormat](https://nlog-project.org/2025/04/29/nlog-6-0-major-changes.html#nlog-filetarget-and-archivesuffixformat)

hhyyrylainen:
Doesn't that your example also still use the legacy mode due to having `ArchiveFileName`? I was curious about what is the replacement for that log date naming feature in NLog 6 that wouldn't cause legacy mode to be enabled?

Edit: I read the linked documentation page and that was very nice and clear on the reasoning. The reason why I'd want the newest log file to have the "plain" name is that I think it is much easier to get users to attach the right log file to bug reports if I can just always tell them to just look for "thrive-launcher-log.txt" and attach that as it is always the latest.

snakefoot:
Yes the static-log-filename is useful when wanting predictable log-filename. But I'm working with services where Elastic-agents uploads the logfiles to ElasticSearch, and it is not important to have predictable log-filenames. Instead I had issues with background applications like Windows Defender suddenly causing file-lock-problems.

This is how it would look for me:
```c#
var fileTarget = new FileTarget
{
    FileName = System.IO.Path.Combine(basePath, "thrive-launcher-log-${date:format=yyy-MM-dd}.txt"),
    ArchiveAboveSize = GlobalConstants.MEBIBYTE * 2,
    MaxArchiveFiles = 4,
};
```

I recommend not using `AtomicFileTarget` unless really necessary to have multiple processes on the same machine writing to the same logfile.

snakefoot:
@hhyyrylainen NLog v6.0.2 has been released:

- https://www.nuget.org/packages/NLog/6.0.2#releasenotes-body-tab

Thank you again for the feedback

PR #5954, opened 2025-08-08
ConsoleWordHighlightingRule - Support List of Words for simpler configuration

snakefoot:
With RegEx one could do the following:
```
regex="abc|def|ghi"
```
Now one can do this (Remember to remove any leading/trailing white-spaces):
```
words="abc,def,ghi"
```


snakefoot:

PR #5964, opened 2025-08-13
ColoredConsoleTarget - WholeWords should only highlight the word

snakefoot:
Thus one can specify `text="|Warn|" wholeWords="true"` and it will only highlight `Warn` when inside `|`


snakefoot:

PR #5883, opened 2025-06-16
AsyncTargetWrapper - LogEventDropped and EventQueueGrow events fixes

dance:
Problems fixed:
- if one was already subscribed for `EventQueueGrow` event, subscribing for `LogEventDropped` had no effect (wrong field was being checked for null)
- `EventQueueGrow` reported old queue size despite parameter name `LogEventQueueGrowEventArgs.NewQueueSize`. Fixed that by doubling `RequestLimit` first, then invoking the event
- `ConcurrentRequestQueue` could erroneously multiply its `RequestLimit` more than once when hitting current limit under concurrent load because of a race condition

Improvements:
- unified `OnLogEventDropped` and `OnLogEventQueueGrows` unit-tests for `ConcurrentRequestQueue` and `AsyncRequestQueue` - they should behave the same
- added unit test to illustrate `RequestLimit` race in `ConcurrentRequestQueue`

PR #5967, opened 2025-08-20
Add Benchmark project and fix boxing issue in ConsoleRowHighlightingRule

karpinsn:
The new benchmark measures the performance of using `ColoredConsoleTarget` to log to a console. The performance fix is to remove the boxing that is occurring in the condition passed to `ConsoleRowHighlightingRule`. It now checks the condition of the expression and returns a cached boxed boolean value from `ConditionExpression`. The current benchmarks results in saving ~8% allocations (from 1.33Mb to 1.22Mb). This doesn't result in a noticable CPU performance saving but should help with GC overhead.

# Before - 1.33Mb
Before
image

# After - 1.22Mb
After
image

### DIFF
--- a/src/NLog.Benchmarks/ColoredConsoleTargetBenchmark.cs
+++ b/src/NLog.Benchmarks/ColoredConsoleTargetBenchmark.cs
@@ -1,0 +1,61 @@
+using System;
+using System.IO;
+
+using BenchmarkDotNet.Attributes;
+
+using NLog.Config;
+using NLog.Targets;
+
+namespace NLog.Benchmarks
+{
+    public class ColoredConsoleTargetBenchmark
+    {
+        private static readonly Logger logger = LogManager.GetCurrentClassLogger();
+        private static readonly string message = "The cat sat at the bar.";
+        private MemoryStream memoryStream = new MemoryStream();
+        private StreamWriter streamWriter;
+        private TextWriter oldConsoleOutWriter;
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            var config = new LoggingConfiguration();
+            var target = new ColoredConsoleTarget("ColoredConsole")
+                { Layout = "${logger} ${message}" };
+            target.WordHighlightingRules.Add(
+                new ConsoleWordHighlightingRule
+                {
+                    ForegroundColor = ConsoleOutputColor.Red,
+                    Text = "at",
+                    IgnoreCase = true
+                });
+
+            config.AddTarget(target);
+            config.AddRule(LogLevel.Debug, LogLevel.Fatal, target);
+
+            LogManager.Configuration = config;
+
+            streamWriter = new StreamWriter(memoryStream);
+            oldConsoleOutWriter = Console.Out;
+            Console.SetOut(streamWriter);
+        }
+
+        [GlobalCleanup]
+        public void Cleanup()
+        {
+            Console.SetOut(oldConsoleOutWriter);
+            streamWriter.Dispose();
+            memoryStream.Dispose();
+        }
+
+        [Benchmark]
+        public void WordHighlightingTextIgnoreCase()
+        {
+            memoryStream.SetLength(0); // Clear out the previous run
+            for (int i = 0; i < 1000; ++i)
+            {
+                logger.Debug(message);
+            }
+        }
+    }
+}
--- a/src/NLog.Benchmarks/NLog.Benchmarks.csproj
+++ b/src/NLog.Benchmarks/NLog.Benchmarks.csproj
@@ -1,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net462</TargetFramework>
+    <OutputType>Exe</OutputType>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="BenchmarkDotNet" Version="0.15.2" />
+    <PackageReference Include="Microsoft.VisualStudio.DiagnosticsHub.BenchmarkDotNetDiagnosers" Version="18.0.36328.1" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\tests\NLog.UnitTests\NLog.UnitTests.csproj" />
+  </ItemGroup>
+</Project>
--- a/src/NLog.Benchmarks/NLog.Benchmarks.csproj
+++ b/src/NLog.Benchmarks/NLog.Benchmarks.csproj
@@ -3,6 +3,8 @@
   <PropertyGroup>
     <TargetFramework>net462</TargetFramework>
     <OutputType>Exe</OutputType>
+    <SonarQubeTestProject>true</SonarQubeTestProject>
+    <IsPackable>false</IsPackable>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/NLog.Benchmarks/NLog.Benchmarks.csproj
+++ b/src/NLog.Benchmarks/NLog.Benchmarks.csproj
@@ -7,10 +7,10 @@
 
   <ItemGroup>
     <PackageReference Include="BenchmarkDotNet" Version="0.15.2" />
-    <PackageReference Include="Microsoft.VisualStudio.DiagnosticsHub.BenchmarkDotNetDiagnosers" Version="18.0.36328.1" />
+    <PackageReference Include="Microsoft.VisualStudio.DiagnosticsHub.BenchmarkDotNetDiagnosers" Version="18.0.36422.2" />
   </ItemGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\..\tests\NLog.UnitTests\NLog.UnitTests.csproj" />
+    <ProjectReference Include="..\NLog\NLog.csproj" />
   </ItemGroup>
 </Project>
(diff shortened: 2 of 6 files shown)


karpinsn:
I'll fix the unreachable code nit, I'm just waiting for a new fix in our profiling agent that found this. If folks are interested this talk shows the agent that found this fix: https://www.youtube.com/watch?v=zhXhUMkrJDQ

snakefoot:
Thank you for this pull-request, very nice find.

Some observations:
- When the NLog maintainer fails to do it correctly, then there is probably a pitfall to be removed. Maybe one should create a method that supports boolean-condition, that automatically does the mapping to non-boxing.
- Really like the idea of having a benchmark-project that exercise the different features on NLog. But for the Console-Target, then I prefer "mocking" so instead of modifying the global-console then Console-Target should get a constructor where one can provide the output-stream.
- Saw the Youtube-video and it is almost magical. Amazing how it was able to recognize that multiple expression-compiles could be merged into a single expression-compile. But yes the code-style of the CoPilot-changes is a little weird, and fails to follow the project standards.
- Sad to see the recently added unit-test `Enqueue_WhenGrowBehaviourAndHighlyConcurrent_GrowOnce` is unstable and causing builds to fail. Will probably revert the unit-test and wait for the original author to fix it. #5883

snakefoot:
Possible you can make co-pilot rewrite this pull-request to match #5968 ?

I like the goal of having an official NLog-Benchmark-project depending on Benchmark.net, which makes it easy to profile common usage-patterns. But not sure it should be mixed into this pull-request. Possible that you can update this pull-request to only include the code-fix ?

karpinsn:
Yep, happy to reformat to #5968. Would you like me to split this PR into 2 separate PRs, one for the benchmarks and one for the fix? I did it as 2 separate commits with the first being the benchmark addition and the second being the perf fix.

>Saw the Youtube-video and it is almost magical. Amazing how it was able to recognize that multiple expression-compiles could be merged into a single expression-compile. But yes the code-style of the CoPilot-changes is a little weird, and fails to follow the project standards.

Yeah, its pretty incredible for being able to run a whole bunch of different experiments really quickly and finding clever optimizations, I can't wait to release it! It doesn't always hit the standards right and sometimes creates / extends public interfaces breaking things so you need to keep it in line / touch things up afterwards. I tried to do that with this PR, feel free to point out anything you are not happy with, I'm more than happy to change stuff. Just trying to find successes with the project to make sure it really helps folks and do some good in the community as well :)

karpinsn:
>But for the Console-Target, then I prefer "mocking" so instead of modifying the global-console then Console-Target should get a constructor where one can provide the output-stream.

Happy to change this around, was just trying to leave existing APIs as is and this seemed to be the best way to do that. I can add another constructor to `ColoredConsoleTarget` that takes in two `TextWriter` objects (one for error and one for out) then change `ColoredConsoleTarget.GetOutput` from static to be an instance method that returns either the passed in objects or the normal console ones. Is this the sort of change you are looking for? Are there any requirements on changing public classes I should be aware of?

snakefoot:
> Yep, happy to reformat to #5968. Would you like me to split this PR into 2 separate PRs, one for the benchmarks and one for the fix?

After further consideration, then just keep this in a single pull-request (Code-fix + new benchmark-project). But please update code-fix to match #5968

>  I can add another constructor to ColoredConsoleTarget that takes in two TextWriter objects

Maybe easier that you just commit the NLog.Benchmark-project in its current state. Then I can adjust to my liking (Think it will become an internal-constructor only visible to the unit-test-projects, that takes a method-delegate that returns the TextWriter, where the default delegate will be a method that returns StdOut / StdErr depending on target-configuration).

snakefoot:
> I tried to do that with this PR, feel free to point out anything you are not happy with, I'm more than happy to change stuff. Just trying to find successes with the project to make sure it really helps folks and do some good in the community as well :)

Maybe the "Profiler Agent" should rely on a "Code Improvement"-agent that have been prompted differently (Like the "Code Review" agent follows project-guidelines-md). Instead of making the surgical code-fix directly (based on profiler-red-path-discovery), then give the task of implementing the surgical code-fix to a "Code Improvement"-agent. The "Code Improvement"-agent can then consider how the initial change would affect the overall code-base, and matches the project-code-style, and whether other caller-locations should also be considered.

But maybe that is only for enterprise-customers, while free-riders just have to do their own code-review of the surgical code-fix.

karpinsn:
>Maybe the "Profiler Agent" should rely on a "Code Improvement"-agent

It's a really interesting thought. I looked at the change myself and touched things up as I don't want to be inundating maintainers with bad AI contributions and I missed bringing this up level. It's been fascinating to see it try different perf experiments and what it comes up with. I'm guessing we could have a "Code Improvement"-agent do a few different versions of a performance fix so the dev can chose which one they want (or which is the closest).

>But maybe that is only for enterprise-customers, while free-riders just have to do their own code-review of the surgical code-fix.

Well the Profiler Agent will be in all SKUs of Visual Studio, and if we do a "Code Improvement"-agent hopefully that one can be as well.

karpinsn:
I fixed up the code to align with #5968, let me know if there are any other changes you would like.

I'm going to try and see if I can find other fixes, for future PRs do you want each benchmark separate from the fix or new benchmark and fix in the same PR? I typically think of benchmarks as unit tests for perf which is why I include them with the fix. Also, any areas / APIs in particular you want me to have the Profiler Agent try and review?

snakefoot:
> for future PRs do you want each benchmark separate from the fix or new benchmark and fix in the same PR? I typically think of benchmarks as unit tests for perf which is why I include them with the fix. 

I'm happy with benchmark being added together with code-fix (Just needed to adjust my world view, since few NLog contributors focus on isolated benchmark, and if creating benchmarks then making them non-generic and closely matching their own local environment/config). Was thinking to have a dedicated benchmark for the different NLog-features / NLog-use-case.

> Also, any areas / APIs in particular you want me to have the Profiler Agent try and review?

I was thinking that one should have at least one benchmark for each NLog-target / NLog-Target-wrapper / NLog-Layout / NLog-LayoutRenderer.  But maybe also a benchmark for the common NLog Logger use-cases:
- NLog Logger where LogLevel is below minimum-loglevel (no-op)
- NLog Logger where LogLevel is below minimum-loglevel, and includes structured properties (no-op)
- NLog Logger where LogLevel is below minimum-loglevel, and includes positional parameters (no-op)
- NLog Logger where LogLevel matches a single target
- NLog Logger where LogLevel matches a single target and includes structured properties.
- NLog Logger where LogLevel matches a single target and includes positional parameters.
- NLog Logger where LogLevel matches two targets

snakefoot:
Think you should mark the NLog.Benchmark-csproj, as "test"-project, then Sonar will ignore the missing code-coverage:
```xml
<SonarQubeTestProject>true</SonarQubeTestProject>
<IsPackable>false</IsPackable>
```
https://docs.sonarsource.com/sonarqube-server/10.8/analyzing-source-code/dotnet-environments/specify-test-project-analysis/

karpinsn:
Looks like most of the other projects have:
`<SonarQubeExclude>true</SonarQubeExclude>`
Preference between this and `SonarQubeTestProject`?

snakefoot:
> Preference between this and SonarQubeTestProject?

I prefer `<SonarQubeTestProject>` since I usually like the Sonar-Code-Style-NitPicks. Not wanting to disable Sonar completely.

But I guess at some point then dotnet-analyzers / dotnet-coverage will take over everything :)

snakefoot:
@karpinsn Just curious. Why did you choose net462 for the benchmark-project instead of net8 / net9 ? I'm currently treating .NET Framework as mostly legacy. Is it because the JIT-compiler is not so good with net462, while the JIT-compiler in net8 / net9 can hide/fix less optimal code? (Along with more aggressive GC because of DATAS)

karpinsn:
@snakefoot I based it off the NLog.RegEx.Tests project since the `ColoredConsoleTargetTests` is what I used to have Copilot generate the benchmark. I saw you target both net462 and net8.0 so I opted for the min version since then Copilot won't generate anything that is net8.0 specific. I can specify multiple frameworks, BenchmarkDotNet supports running across versions so you can see the performance enhancement. Want me to do that here or generate a new PR for that?

snakefoot:
>  I saw you target both net462 and net8.0 so I opted for the min version since then Copilot won't generate anything that is net8.0 specific.

Ok so a signal to CoPilot that it should not use code-features that doesn't work with all target-platforms.

> Want me to do that here or generate a new PR for that?

No it is fine. Was just surprised that net462 was chosen as the "future". But I will probably change it to net8 / net9 some time later.

snakefoot:
Closing and reopening the pull-request to tickle the build-servers and the unstable unit-test (See #5949)

snakefoot:
Thank you for the contribution. I have made some adjustments to the new NLog.Benchmarks-project:
-  Moved NLog.Benchmarks into the Tests-project-folder #5974 
-  ColoredConsoleTargetBenchmark - Removed global side-effects #5975

Guess I should start working on extracting the many nuget-packages into their own isolated repositories, since NLog v6 is looking stable. Hopefully it will bring down the build-time and improve the feedback-loop.

karpinsn:
Awesome, thanks for helping me get this in! I'm starting to poke at the different layout renderers and seeing if the Profiler Agent can find anything.

PR #6064, opened 2025-12-17
FileTarget - Reduce code complexity for ArchiveNumbering property

snakefoot:


src/NLog/Targets/FileTarget.cs
```diff
@@ -390,16 +390,14 @@ public string ArchiveNumbering
             get => _archiveNumbering ?? "Sequence";
             set
             {
-                if (string.Equals(value, _archiveNumbering))
-                    return;
-
-                _archiveNumbering = string.IsNullOrEmpty(value) ? null : value.Trim();
-                if (_archiveNumbering is null || string.IsNullOrEmpty(_archiveNumbering))
+                var archiveNumbering = value?.Trim() ?? string.Empty;
+                if (string.IsNullOrEmpty(archiveNumbering) || string.Equals(_archiveNumbering, archiveNumbering, StringComparison.OrdinalIgnoreCase))
```
lechu445, line 394:
```suggestion
                if (string.IsNullOrEmpty(archiveNumbering) || archiveNumbering.Equals(_archiveNumbering, StringComparison.OrdinalIgnoreCase))
```
this will reduce one null check that is called in the static string.Equals method

src/NLog/Targets/FileTarget.cs
```diff
@@ -390,16 +390,14 @@ public string ArchiveNumbering
...
                     return;
 
+                _archiveNumbering = archiveNumbering;
                 if (_archiveSuffixFormat is null || ReferenceEquals(_archiveSuffixFormat, _legacyDateArchiveSuffixFormat) || ReferenceEquals(_archiveSuffixFormat, _legacySequenceArchiveSuffixFormat))
                 {
-                    ArchiveSuffixFormat = _archiveNumbering.IndexOf("date", StringComparison.OrdinalIgnoreCase) >= 0 ? _legacyDateArchiveSuffixFormat : _legacySequenceArchiveSuffixFormat;
+                    ArchiveSuffixFormat = archiveNumbering.IndexOf("date", StringComparison.OrdinalIgnoreCase) >= 0 ? _legacyDateArchiveSuffixFormat : _legacySequenceArchiveSuffixFormat;
```
lechu445, line 400:
```suggestion
                    ArchiveSuffixFormat = archiveNumbering.Contains("date", StringComparison.OrdinalIgnoreCase) ? _legacyDateArchiveSuffixFormat : _legacySequenceArchiveSuffixFormat;
```
this can be simplified

src/NLog/Targets/FileTarget.cs
snakefoot: Old habit of using `string.Equals` not that worried about a bonus null-check for an obsolete property. But thank you for noticing.

src/NLog/Targets/FileTarget.cs
snakefoot: Recall that Contains with StringComparison is not available on NET35, which is why using IndexOf. But thank you for noticing.

PR #6085, opened 2026-01-31
PropertiesDictionary - Initial 3 properties without dictionary

snakefoot:


snakefoot:
Think the performance gained is not worth the price in complexity and the new dual behavior of message-template-properties-array.

PR #6087, opened 2026-02-01
Target - Reduce code complexity of WriteAsyncLogEvents

snakefoot:
Failing to initialize NLog targets is an exotic situation, so changing behavior to simplify logic should be acceptable.

- [x] Wait for NLog v6.2


snakefoot:
@coderabbitai full review

PR #6113, opened 2026-03-08
FileTarget - Upfront validation of ArchiveSuffixFormat property

snakefoot:
Detect invalid use of `{#}` or `${date}` at startup, instead of later when archive is performed.


snakefoot:
@coderabbitai resume

PR #6136, opened 2026-03-29
MemoryTarget: add INotifyCollectionChanged support via ObservableLogs

### LINKED ISSUE #6124: MemoryTarget with support for INotifyCollectionChanged
Making it easier to hookup a GUI to NLog MemoryTarget. Ex. for displaying the last X error-events.

See also:
- https://learn.microsoft.com/en-us/archive/blogs/xtof/making-sense-of-notifycollectionchangedeventargs
- https://baumbartsjourney.wordpress.com/2009/06/01/an-alternative-to-observablecollection/
- https://blog.stephencleary.com/2009/07/interpreting-notifycollectionchangedeve.html

Maybe declare new interface, and add property to MemoryTarget that returns this interface:
```
public interface IObservableCollection<T> : IReadOnlyList<T>, INotifyCollectionChanged { }
```

Not an expert on `INotifyCollectionChanged` so unsure how much is required for it to be useful / performant. Maybe some verification is needed to see if it works as expected with Winforms / WPF / MAUI / etc.

reabr:
Closes #6124
Added INotifyCollectionChanged support to MemoryTarget via a new ObservableLogs property, making it easier to bind NLog output to a GUI (WPF, MAUI, etc).

Changes:

ThreadSafeList<T> now implements INotifyCollectionChanged (guarded with #if !NET35)
New ObservableLogs property exposed on MemoryTarget
Two unit tests added covering Add and Reset events

### DIFF
--- a/src/NLog/Targets/MemoryTarget.cs
+++ b/src/NLog/Targets/MemoryTarget.cs
@@ -88,6 +88,14 @@ namespace NLog.Targets
         /// </remarks>
         public IList<string> Logs => _logs;
 
+#if !NET35
+        /// <summary>
+        /// Gets <see cref="Logs"/> as <see cref="System.Collections.Specialized.INotifyCollectionChanged"/>
+        /// to support GUI data binding (e.g. WPF, MAUI).
+        /// </summary>
+        public System.Collections.Specialized.INotifyCollectionChanged ObservableLogs => _logs;
+#endif
+
         /// <summary>
         /// Gets or sets the max number of items to have in memory. Zero or Negative means no limit.
         /// </summary>
@@ -127,9 +135,20 @@ namespace NLog.Targets
         }
 
         private sealed class ThreadSafeList<T> : IList<T>
+#if !NET35
+            , System.Collections.Specialized.INotifyCollectionChanged
+#endif
         {
             private readonly List<T> _list = new List<T>();
+#if !NET35
 
+            public event System.Collections.Specialized.NotifyCollectionChangedEventHandler? CollectionChanged;
+            private void OnCollectionChanged(System.Collections.Specialized.NotifyCollectionChangedEventArgs e)
+            {
+            CollectionChanged?.Invoke(this, e);
+            }
+
+#endif
             public T this[int index]
             {
                 get
@@ -157,19 +176,45 @@ namespace NLog.Targets
                 {
                     _list.Add(item);
                 }
+#if !NET35
+                OnCollectionChanged(new System.Collections.Specialized.NotifyCollectionChangedEventArgs(
+                    System.Collections.Specialized.NotifyCollectionChangedAction.Add, item));
+#endif
+
+
             }
 
             public void Add(T item, int maxListCount)
             {
+#if !NET35
+                bool itemsRemoved = false;
+#endif
                 lock (_list)
                 {
                     if (maxListCount > 0)
                     {
                         while (_list.Count >= maxListCount)
+                        {
                             _list.RemoveAt(0);
+#if !NET35
+                 itemsRemoved = true;
+#endif
+                        }
                     }
                     _list.Add(item);
                 }
+#if !NET35
+               if (itemsRemoved)
+               {
+                   OnCollectionChanged(new System.Collections.Specialized.NotifyCollectionChangedEventArgs(
+                       System.Collections.Specialized.NotifyCollectionChangedAction.Reset));
+               }
+               else
+               {
+                   OnCollectionChanged(new System.Collections.Specialized.NotifyCollectionChangedEventArgs(
+                       System.Collections.Specialized.NotifyCollectionChangedAction.Add, item));
+               }
+#endif
             }
 
             void ICollection<T>.Clear()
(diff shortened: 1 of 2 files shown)


src/NLog/Targets/MemoryTarget.cs
```diff
@@ -157,19 +176,45 @@ public void Add(T item)
...
+#if !NET35
+               if (itemsRemoved)
+               {
+                   OnCollectionChanged(new System.Collections.Specialized.NotifyCollectionChangedEventArgs(
+                       System.Collections.Specialized.NotifyCollectionChangedAction.Reset));
+               }
+               else
+               {
+                   OnCollectionChanged(new System.Collections.Specialized.NotifyCollectionChangedEventArgs(
+                       System.Collections.Specialized.NotifyCollectionChangedAction.Add, item));
```
snakefoot: I'm not an expert in `INotifyCollectionChanged` but should it not include the index of the insert ?
>[NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction, Object, Int32)](https://learn.microsoft.com/en-us/dotnet/api/system.collections.specialized.notifycollectionchangedeventargs.-ctor?view=net-10.0#system-collections-specialized-notifycollectionchangedeventargs-ctor(system-collections-specialized-notifycollectionchangedaction-system-object-system-int32))

Possible that we can skip the allocation of `NotifyCollectionChangedEventArgs` when no subscribers?

src/NLog/Targets/MemoryTarget.cs
```diff
@@ -157,19 +176,45 @@ public void Add(T item)
...
+                 itemsRemoved = true;
+#endif
+                        }
                     }
                     _list.Add(item);
                 }
+#if !NET35
+               if (itemsRemoved)
+               {
+                   OnCollectionChanged(new System.Collections.Specialized.NotifyCollectionChangedEventArgs(
```
snakefoot: I'm not an expert in `INotifyCollectionChanged`, but is `Reset` the most optimal way of notifying that an element was removed ?

Possible that we can skip the allocation of `NotifyCollectionChangedEventArgs` when no subscribers?

src/NLog/Targets/MemoryTarget.cs
```diff
@@ -157,19 +176,45 @@ public void Add(T item)
                 {
                     _list.Add(item);
                 }
+#if !NET35
+                OnCollectionChanged(new System.Collections.Specialized.NotifyCollectionChangedEventArgs(
+                    System.Collections.Specialized.NotifyCollectionChangedAction.Add, item));
```
snakefoot: I'm not an expert in `INotifyCollectionChanged` but should it not include the index of the insert ?
>[NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction, Object, Int32)](https://learn.microsoft.com/en-us/dotnet/api/system.collections.specialized.notifycollectionchangedeventargs.-ctor?view=net-10.0#system-collections-specialized-notifycollectionchangedeventargs-ctor(system-collections-specialized-notifycollectionchangedaction-system-object-system-int32))


Possible that we can skip the allocation of `NotifyCollectionChangedEventArgs` when no subscribers?

src/NLog/Targets/MemoryTarget.cs
snakefoot: Should `INotifyCollectionChanged` not be implemented for the other methods that can modify the list?

src/NLog/Targets/MemoryTarget.cs
reabr: Good catch added the index to both Add overloads. Also added a CollectionChanged != null guard before allocating NotifyCollectionChangedEventArgs to skip the allocation when there are no subscribers.

src/NLog/Targets/MemoryTarget.cs
reabr: Yeah Reset felt right here since we might drop multiple items at once

src/NLog/Targets/MemoryTarget.cs
reabr: Done
Clear → Reset, Remove/RemoveAt → Remove with index, Insert → Add with index, indexer setter → Replace

src/NLog/Targets/MemoryTarget.cs
snakefoot: 99.9 pct of the time, it will only remove a single item, unless `MaxLogsCount` is constantly being modified.

snakefoot:
The hard part for me was not to implement the interface, but ensuring that the extra complexity would be useful for anyone. Doing the actual proof-of-concept-testing for the various platforms (Winforms / WPF / MAUI / etc.).

But now you have forced me to really consider if my half-baked idea actually makes any sense. And I actually think the best solution for GUI-developers would be to setup a NLog [MethodCall-target](https://github.com/NLog/NLog/wiki/MethodCall-target) and insert into their own instance of [ObservableList<T>](https://learn.microsoft.com/dotnet/api/microsoft.visualstudio.extensibility.ui.observablelist-1). This will allow them to insert their own basic DTO instead of the complex/noisy NLog LogEventInfo. 

So unless you have an urgent need for `INotifyCollectionChanged`-support, then I prefer to close my original-issue along with this pull-request. I'm sorry about the effort/time wasted on this pull-request, but most of my issues are reminders for myself about half-baked ideas. In the future I suggest you write a comment about being interested in working on the issue and your ideas, before starting the actual work.

reabr:
@snakefoot,

I understand your concerns about complexity and the preferred MethodCall target approach.

I'm fine with closing this PR if you feel the MethodCall target pattern is the better direction. The implementation is here if you or others want to reference it later, but I won't push back on closing it.

Thanks for reviewing!

snakefoot:
Closing as not needed, but maybe others can find ideas from it.

PR #6138, opened 2026-03-29
SimpleLayout skip Precalculate when INoAllocationStringValueRenderer available

### LINKED ISSUE #6111: INoAllocationStringValueRenderer should skip precalculate Layout
There is a lot of memory-allocation when needing to perform thread-context-capture. When NLog sees too many Properties on LogEventInfo, then it skips checking if all properties are immutable, and automatically assumes the entire LogEventInfo as mutable/dangerous where async-logging requires capture of thread-context.

But when the NLog SimpleLayout sees that `INoAllocationStringValueRenderer` is available, and returning not-null then it can actually skip the precalculation-logic with its memory-allocations. This will optimize async-logging of non-existing LogEventInfo-Properties or when of string-type.

Bonus trouble comes when using `INoAllocationStringValueRenderer`-capable Layout as nested Layout, and need to evaluate if the upper Layout should attempt to precalculate itself entirely or more optimal to precalcuate all nested Layouts individually.

reabr:
## Summary
Optimize async logging by skipping `Precalculate` when `INoAllocationStringValueRenderer` is available.

## Problem
When async logging is used, NLog calls `Precalculate` on the app thread to capture thread-context before handing the log event to a background thread. This causes memory allocations even when the layout renderers are capable of serving the value without allocations and without needing thread-context.

## Changes
- `SimpleLayout.PrecalculateMustRenderLayoutValue`: Skip precalculation when all renderers implement `INoAllocationStringValueRenderer` and return non-null for the current log event
- `SimpleLayout.AllRenderersAreNoAllocation`: New per-event runtime check
- `SimpleLayout.IsNoAllocationLayout`: New static structural check used by parent layouts
- `Layout.ResolveLayoutPrecalculation`: Skip no-alloc capable nested layouts when building the precalculate list

Fixes #6111

### DIFF
--- a/src/NLog/Layouts/Layout.cs
+++ b/src/NLog/Layouts/Layout.cs
@@ -407,17 +407,10 @@ namespace NLog.Layouts
                 ++layoutCount;
                 if (layout?.ThreadAgnostic == false || layout?.ThreadAgnosticImmutable == true)
                 {
-                    if (layout is SimpleLayout simpleLayout && simpleLayout.IsNoAllocationLayout)
+                    precalculateLayoutCount++;
+                    if (layout is SimpleLayout)
                     {
-                        // No-alloc layout can skip precalculation entirely — no need to count it
-                    }
-                    else
-                    {
-                        precalculateLayoutCount++;
-                        if (layout is SimpleLayout)
-                        {
-                            precalculateSimpleLayoutCount++;
-                        }
+                        precalculateSimpleLayoutCount++;
                     }
                 }
             }
@@ -428,8 +421,7 @@ namespace NLog.Layouts
             }
             else
             {
-                return allLayouts.Where(layout => (layout?.ThreadAgnostic == false || layout?.ThreadAgnosticImmutable == true) &&
-                                      !(layout is SimpleLayout sl && sl.IsNoAllocationLayout)).ToArray();
+                return allLayouts.Where(layout => layout?.ThreadAgnostic == false || layout?.ThreadAgnosticImmutable == true).ToArray();
             }
         }
 
--- a/src/NLog/Layouts/Layout.cs
+++ b/src/NLog/Layouts/Layout.cs
@@ -93,7 +93,7 @@ namespace NLog.Layouts
         {
             if (text is null)
                 return new Layout<string>(null);
-            else 
+            else
                 return FromString(text, ConfigurationItemFactory.Default);
         }
 
@@ -407,10 +407,17 @@ namespace NLog.Layouts
                 ++layoutCount;
                 if (layout?.ThreadAgnostic == false || layout?.ThreadAgnosticImmutable == true)
                 {
-                    precalculateLayoutCount++;
-                    if (layout is SimpleLayout)
+                    if (layout is SimpleLayout simpleLayout && simpleLayout.IsNoAllocationLayout)
                     {
-                        precalculateSimpleLayoutCount++;
+                        // No-alloc layout can skip precalculation entirely — no need to count it
+                    }
+                    else
+                    {
+                        precalculateLayoutCount++;
+                        if (layout is SimpleLayout)
+                        {
+                            precalculateSimpleLayoutCount++;
+                        }
                     }
                 }
             }
@@ -421,7 +428,8 @@ namespace NLog.Layouts
             }
             else
             {
-                return allLayouts.Where(layout => layout?.ThreadAgnostic == false || layout?.ThreadAgnosticImmutable == true).ToArray();
+                return allLayouts.Where(layout => (layout?.ThreadAgnostic == false || layout?.ThreadAgnosticImmutable == true) &&
+                                      !(layout is SimpleLayout sl && sl.IsNoAllocationLayout)).ToArray();
             }
         }
 
--- a/src/NLog/Layouts/SimpleLayout.cs
+++ b/src/NLog/Layouts/SimpleLayout.cs
@@ -337,7 +337,7 @@ namespace NLog.Layouts
                 return false;
 
             if (_rawValueRenderer != null && TryGetRawValue(logEvent, out var rawValue) && IsRawValueImmutable(rawValue))
-                return false; // If raw value is immutable, then we can skip precalculate-caching
+                return false;   // If raw value is immutable, then we can skip precalculate-caching
 
             if (logEvent.TryGetCachedLayoutValue(this, out var _))
                 return false;
--- a/src/NLog/Layouts/SimpleLayout.cs
+++ b/src/NLog/Layouts/SimpleLayout.cs
@@ -126,14 +126,11 @@ namespace NLog.Layouts
                 else if (_layoutRenderers[0] is IStringValueRenderer stringValueRenderer)
                 {
                     _stringValueRenderer = stringValueRenderer;
+                    _noAllocRenderer = stringValueRenderer as INoAllocationStringValueRenderer;
                 }
                 if (_layoutRenderers[0] is IRawValue rawValueRenderer)
                 {
                     _rawValueRenderer = rawValueRenderer;
-                }
-                if (_layoutRenderers[0] is INoAllocationStringValueRenderer noAllocRenderer)
-                {
-                    _noAllocRenderer = noAllocRenderer;
                 }
             }
         }
@@ -344,13 +341,12 @@ namespace NLog.Layouts
             if (logEvent.TryGetCachedLayoutValue(this, out var _))
                 return false;
 
-            // No-allocation optimization takes priority over IStringValueRenderer fast-path
-            // since it can avoid both StringBuilder and AddCachedLayoutValue
-            if (_noAllocRenderer?.GetFormattedStringNoAllocation(logEvent) is not null)
-                return false;  
-
             if (IsSimpleStringText)
             {
+                // No-allocation optimization means it can skip both StringBuilder and precalculate-caching
+                if (_noAllocRenderer?.GetFormattedStringNoAllocation(logEvent) is not null)
+                    return false;
+            
                 var cachedLayout = GetFormattedMessage(logEvent);
                 logEvent.AddCachedLayoutValue(this, cachedLayout);
                 return false;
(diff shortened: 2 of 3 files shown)


reabr:
@coderabbitai I have already added unit tests in this PR (in SimpleLayoutOutputTests.cs). Please mark the "Create PR with unit tests" task as complete.

reabr:
@coderabbitai I have already added unit tests in this PR (in SimpleLayoutOutputTests.cs). Please mark the "Create PR with unit tests" task as complete.

snakefoot:
Again you have started on one of my half-baked ideas, which is more a discussion with myself, than something that is ready to be worked on.

I'm still considering if I want to add more complexity/logic around `INoAllocationStringValueRenderer`, since it is already somewhat exotic. It gives a huge performance bonus for my application, that writes large blobs of text that must be encoded as Json.

The current implementation you have made doesn't recognize the special behavior of `INoAllocationStringValueRenderer` that it can return `null` when the optimization is not possible. It is not possible during initialization to decide that one can completely skip the precalculation. One must check each LogEventInfo to see if optimization is possible, and if so then skip the precalculation-capture.

In the future I suggest you write a comment about being interested in working on the issue and your ideas, before starting the actual work.

reabr:
@snakefoot,

Thanks for the feedback. I completely understand your frustration regarding the process  I apologize for jumping into implementation before discussing the design on the issue. I'll make sure to coordinate first in the future.

Regarding the technical concern about `INoAllocationStringValueRenderer` returning `null`:

I might not have made this clear enough in the code structure, but I did attempt to handle the per-event check. In `SimpleLayout.cs`, I added a method that checks each `LogEventInfo` at runtime:

```csharp
private bool AllRenderersAreNoAllocation(LogEventInfo logEvent)
{
    // ...
    if (noAllocRenderer.GetFormattedStringNoAllocation(logEvent) is null)
        return false; // Optimization not possible for this specific event
    // ...
}
```


This method is called from PrecalculateMustRenderLayoutValue(logEvent), so the decision to skip precalculation is evaluated dynamically for each LogEventInfo, not determined at layout initialization.
Could you clarify your concern?

Is it that calling GetFormattedStringNoAllocation(logEvent) just to check for null introduces too much overhead (rendering twice)?
    Or is the issue with how the static IsNoAllocationLayout hint interacts with the per event fallback?
    Or is there another aspect of the design you'd prefer to handle differently?

If the design still feels too experimental to merge right now, I'm fine closing the PR and revisiting later. Just let me know what you'd prefer. 👍

src/NLog/Layouts/Layout.cs
```diff
@@ -421,7 +428,8 @@ internal void PerformObjectScanning()
             }
             else
             {
-                return allLayouts.Where(layout => layout?.ThreadAgnostic == false || layout?.ThreadAgnosticImmutable == true).ToArray();
+                return allLayouts.Where(layout => (layout?.ThreadAgnostic == false || layout?.ThreadAgnosticImmutable == true) &&
+                                      !(layout is SimpleLayout sl && sl.IsNoAllocationLayout)).ToArray();
```
snakefoot: Seems this logic will cause a JsonLayout to skip precalculation of inner SimpleLayout that only includes INoAllocationStringValueRenderer. So when the NLog Target precalculates the JsonLayout, then it will not check the SimpleLayout.

src/NLog/Layouts/SimpleLayout.cs
```diff
@@ -348,9 +347,49 @@ private bool PrecalculateMustRenderLayoutValue(LogEventInfo logEvent)
                 return false;
             }
 
+            //  if every renderer can serve this event without allocation,
+            // there is no thread-context to capture — skip precalculation
+            if (AllRenderersAreNoAllocation(logEvent))
```
snakefoot: Instead of enumerating all `_layoutRenderers` then maybe only support single item, similar to `_rawValueRenderer` ? Ex. `${message}` (So skipping enumeration and casting)

src/NLog/Layouts/Layout.cs
reabr: you're right. Filtering at the ResolveLayoutPrecalculation level is too aggressive and would break nested scenarios (like JsonLayout containing a SimpleLayout). The inner layout still needs to be invoked so it can make the per-event decision.
Proposed Fix:
Remove the filtering logic from Layout.cs (ResolveLayoutPrecalculation).
Override Precalculate in SimpleLayout to handle the short-circuit internally:
```
protected internal override void Precalculate(LogEventInfo logEvent)
{
    if (AllRenderersAreNoAllocation(logEvent))
        return; // Skip work for this event
    
    base.Precalculate(logEvent); // Proceed normally
}
```
 Remove IsNoAllocationLayout since it's no longer needed for filtering.

This ensures nested layouts still get called correctly, but we still avoid allocations when possible on a per-event basis.
If you're okay with this approach, I'll push the fix shortly.

src/NLog/Layouts/SimpleLayout.cs
reabr: Makes sense to fast path the single renderer case and skip the loop entirely.
I can add a _noAllocRenderer field similar to that and use it in AllRenderersAreNoAllocation.
If you're okay with that direction, I'll push the fix (including the nested-layout correction from earlier) in the next commit.

src/NLog/Layouts/Layout.cs
snakefoot: I think it makes sense to be "optimistic" about `INoAllocationStringValueRenderer`, so when a SimpleLayout only include one LayoutRenderer that implements `INoAllocationStringValueRenderer` then it doesn't "count". But it should be always included in the list of Layouts to be precalculated.

src/NLog/Layouts/SimpleLayout.cs
snakefoot: I think it make sense to only perform the optimization when detecting a single layoutrenderer, that implements `INoAllocationStringValueRenderer`, which means it can completely skip the use of StringBuilder when "lucky".

reabr:
@snakefoot 
Addressed your feedback:
 Scoped `INoAllocationStringValueRenderer` to single-renderer `SimpleLayout` only (no enumeration overhead)
Runtime feasibility check per `LogEventInfo` with graceful fallback when `null` is returned
Prioritized over `IStringValueRenderer` fast path to prevent premature caching
Layouts remain in precalculation lists =>  nested `JsonLayout` behavior preserved
Added unit tests for both optimization & fallback paths + fixed `_rawValueRenderer` warning

Let me know if this aligns with your design vision!

tests/NLog.UnitTests/Layouts/SimpleLayoutOutputTests.cs
```diff
@@ -288,5 +288,60 @@ protected override string GetFormattedMessage(LogEventInfo logEvent)
...
+
+            // Act
+            layout.Precalculate(logEvent);
+
+            // Assert - value should be cached because renderer returned null (optimization not possible)
+            Assert.True(logEvent.TryGetCachedLayoutValue(layout, out _));
+        }
+
+        [LayoutRenderer("noalloc")]
+        private sealed class NoAllocTestRenderer : LayoutRenderer, INoAllocationStringValueRenderer
```
snakefoot: Instead of creating your own custom `LayoutRenderer` for the uni-test, why not verify the logic using the existing `MessageLayoutRenderer` ?

snakefoot:
> @snakefoot Let me know if this aligns with your design vision!

Really like the now surgical fix, and thank you for the extra effort of implementing unit-test that ensures the optimization actually works.

I think it looks great, and have added a single comment around changing the unit-test to use an existing NLog LayoutRenderer for verification that the optimization works.

tests/NLog.UnitTests/Layouts/SimpleLayoutOutputTests.cs
reabr: Replaced the custom `NoAllocTestRenderer` with the existing `${message}` renderer.

src/NLog/Layouts/SimpleLayout.cs
```diff
@@ -336,11 +339,16 @@ private bool PrecalculateMustRenderLayoutValue(LogEventInfo logEvent)
...
 
             if (_rawValueRenderer != null && TryGetRawValue(logEvent, out var rawValue) && IsRawValueImmutable(rawValue))
-                return false;   // If raw value is immutable, then we can skip precalculate-caching
+                return false; // If raw value is immutable, then we can skip precalculate-caching
 
             if (logEvent.TryGetCachedLayoutValue(this, out var _))
                 return false;
 
+            // No-allocation optimization takes priority over IStringValueRenderer fast-path
+            // since it avoids both StringBuilder AND string allocation during preca
```
snakefoot: `since it avoids both StringBuilder AND string allocation during preca` -> `since it can avoid both StringBuilder and AddCachedLayoutValue`

reabr: @snakefoot Unit tests have been added and committed. The "Create PR with unit tests" task shows as incomplete is this a manual checkbox I need to mark, or is there an automated check that needs to pass?

snakefoot (approved):
I think it looks great. Do you have any pending changes, that you want to add?

src/NLog/Layouts/SimpleLayout.cs
snakefoot: When the builds are green, and no pending sonar-issues then it is ready for review. After that then it is just me, that needs to press the "Merge"-button.

The CodeRabbit is a little noisy, and its use of check-boxes doesn't work well with "task-list-completed". Could be nice if "task-list-completed" only reacted to checkboxes in the initial-post.

reabr:
> I think it looks great. Do you have any pending changes, that you want to add?

No pending changes from my side .everything is ready.

snakefoot:
@reabr Thank you for the contribution to the NLog-library, and being understanding about the state of my half baked ideas :)

reabr:
Thanks! Happy to help, Glad it's merged! 🙌

snakefoot:
NLog v6.1.3 has been released:
- https://www.nuget.org/packages/NLog/6.1.3#releasenotes-body-tab

Thank you again for the contribution.

reabr:
Thanks for the update! Happy to help. 😊

PR #6155, opened 2026-04-17
Improve TypedGetLoggerTest coverage for different logger names

MateusMo:
### Summary

Improves unit test coverage for `TypedGetLoggerTest` by adding validation for different logger names.

### Changes

- Added assertion to verify that loggers with different names do not share the same instance
- Added assertion to validate correct logger name assignment

### Motivation

Ensures that the `LogFactory` caching mechanism correctly distinguishes loggers by name.

### Notes

- No functional changes
- Test-only improvement

### Checklist

- [x] Create PR with unit tests

### DIFF
--- a/tests/NLog.UnitTests/GetLoggerTests.cs
+++ b/tests/NLog.UnitTests/GetLoggerTests.cs
@@ -64,15 +64,18 @@ namespace NLog.UnitTests
             Logger l3 = lf.GetLogger("AAA", typeof(Logger));
             Logger l5 = lf.GetLogger("AAA");
             Logger l6 = lf.GetLogger("AAA");
+            Logger otherLogger = lf.GetLogger("BBB");
 
             Assert.Same(l1, l2);
             Assert.Same(l5, l6);
             Assert.Same(l3, l5);
 
             Assert.NotSame(l1, l3);
+            Assert.NotSame(l5, otherLogger);
 
             Assert.Equal("AAA", l1.Name);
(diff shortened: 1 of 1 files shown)


MateusMo:
The task-list-completed check seems to be counting a duplicate task.

The checklist is completed and tests are included, but the action still reports one incomplete item.

Could you please take a look?

MateusMo:
Friendly ping — would love to know if there's anything blocking review or if any changes are needed.

snakefoot:
You have used an existing test that verifies handling of User-Defined Logger-Types, and extended this with validation of Logger-Names for the Standard-Logger-Type.

I have not looked through all the use-cases for the `LogFactory.GetLogger("...")`, but are you sure that adding extra validation to `TypedGetLoggerTest` is the right place?

MateusMo:
Hi @snakefoot, fair point!

I extended `TypedGetLoggerTest` because it was the closest test I found for `LogFactory.GetLogger`, but I can see now that mixing Standard-Logger-Type name validation into a test focused on User-Defined Logger-Types makes it less cohesive.

Would it make more sense to move these cases to a dedicated test class for `LogFactory.GetLogger("...")`, or is there an existing test file you'd point me to as the right home for this?

snakefoot:
Nice of you to offer that I solve the problem for you. However, I think you should reflect on the feedback I've already provided.

snakefoot:
Closing due to inactivity.

PR #6156, opened 2026-04-18
Improve StringHelpers.Join test coverage (empty, single and multiple …

MateusMo:
This PR adds additional unit tests for StringHelpers.Join.

It covers:
- empty collection case
- single element case
- multiple elements concatenation

These tests improve coverage and ensure consistent behavior across edge cases for string joining logic.

All tests pass on net8.0.

### Checklist

- [x] Create PR with unit tests

### DIFF
--- a/tests/NLog.UnitTests/Internal/StringHelpersTests.cs
+++ b/tests/NLog.UnitTests/Internal/StringHelpersTests.cs
@@ -33,6 +33,7 @@
 
 
 using System;
+using System.Collections.Generic;
 using NLog.Internal;
 using Xunit;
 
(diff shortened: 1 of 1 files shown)


MateusMo:
Friendly ping — would love to know if there's anything blocking review or if any changes are needed.

snakefoot:
StringHelpers.Join is a pure forward to string.Join, and is only there because NET35 has API-limitations.

I'm not a big fan of testing the .NET library (should happen in the dotnet-runtime-repo). What is driving this pull-request?

MateusMo:
Hi @snakefoot, thanks for the context!

You're right — I wasn't aware that `StringHelpers.Join` is just a thin wrapper over `string.Join` kept for NET35 compatibility. The motivation was simply to improve code coverage, but I agree that testing the .NET runtime behavior isn't the right goal here.

I'll close this PR. If there's a more meaningful area where test coverage would actually help the project, I'm happy to look into that instead.

PR #6162, opened 2026-04-22
Add `${datetimeoffset}` layout renderer (fixes #6161)

### LINKED ISSUE #6161: Add `DateTimeOffset` raw-value support to `${date}` / `${longdate}` (or add a new `${datetimeoffset}` renderer)
## Background

Issue #6115 / PR #6116 added typed `DateTimeOffset` flow through `DatabaseTarget` when `DatabaseParameterInfo.ParameterType = typeof(DateTimeOffset)` and the layout's raw value is already a `DateTimeOffset` object.

That fix works great when the user populates `LogEventInfo.Properties["Timestamp"] = new DateTimeOffset(e.TimeStamp)` and uses `${event-properties:item=Timestamp}` as the layout — the typed object flows through to the `SqlParameter` without a string round-trip.

## Problem

`LogEventInfo.TimeStamp` is a `DateTime`, and `${date}` / `${longdate}` only expose it as a `DateTime` raw value. To target a SQL `DATETIMEOFFSET` column from `e.TimeStamp`, callers must:

1. Manually populate an event property in **every** log call (or via a wrapper):

   ```csharp
   e.Properties["Timestamp"] = new DateTimeOffset(e.TimeStamp);
   ```

2. Reference it via `${event-properties:item=Timestamp}` instead of the built-in `${date}`.

This is boilerplate that exists solely to bridge `DateTime` → `DateTimeOffset` for the typed parameter pipeline. Using `${date:format=o}` defeats the raw-value path (forces string rendering), and relying on `ParameterType = typeof(DateTimeOffset)` to convert from a `DateTime` raw value loses fidelity for users whose `TimeSource` is configured for a specific kind.

## Proposed Enhancement

Either of the following would eliminate the workaround:

### Option A (preferred — minimal surface area)

Add an `asDateTimeOffset` (bool) property to `DateLayoutRenderer` (and `LongDateLayoutRenderer`). When `true`, the renderer's `IRawValue.TryGetRawValue` returns a `DateTimeOffset` constructed from `LogEventInfo.TimeStamp` (using `TimeZoneInfo.Local` for `Kind.Local` / `Kind.Unspecified` and `TimeSpan.Zero` for `Kind.Utc`). String rendering remains unchanged.

```xml
<parameter name="@Timestamp"
           layout="${date:asDateTimeOffset=true}"
           parameterType="System.DateTimeOffset" />
```

### Option B (purely additive)

Add a dedicated `${datetimeoffset}` layout renderer that always exposes `e.TimeStamp` as a typed `DateTimeOffset` raw value, with an optional `format` for string rendering.

## Why this fits NLog's model

- It follows the same pattern PR #6116 established: typed values flow end-to-end without string round-trips.
- It uses the existing `IRawValue` plumbing — no new abstraction needed.
- It keeps file/console targets working (string rendering is unchanged) while enabling SQL `DATETIMEOFFSET` columns to receive a native typed value.
- It removes the only remaining reason to push timestamps through `Properties`, which is a documented pain point for SQL `DatabaseTarget` users.

## Alternatives considered

- Setting `e.Properties["Timestamp"] = new DateTimeOffset(e.TimeStamp)` — works today but is per-call boilerplate.
- `${date:format=o}` + `ParameterType=DateTimeOffset` — round-trips through string, defeats the benefit of #6116.
- Custom `LayoutRenderer` subclass — every consumer would have to ship one; better suited as a built-in.

## Environment

- NLog: 6.1.2
- NLog.Database: 6.0.3
- Target: SQL Server `DATETIMEOFFSET` column via `Microsoft.Data.SqlClient`

LorneCash:
Fixes #6161

Introduces a dedicated `DateTimeOffsetLayoutRenderer` (`${datetimeoffset}`) so that `DatabaseTarget` with `parameterType="System.DateTimeOffset"` can flow timestamps directly to a SQL `DATETIMEOFFSET` column without a string round-trip or workaround via `LogEventInfo.Properties`.

**Changes**

- New `DateTimeOffsetLayoutRenderer` (`${datetimeoffset}`) — `TryGetRawValue` always returns a typed `DateTimeOffset`; `Format`, `Culture` and `UniversalTime` properties match the existing date renderers; default format `yyyy/MM/dd HH:mm:ss.fff` keeps string output identical to `${date}` out of the box.
- `AssemblyExtensionTypes.cs` — registered `"datetimeoffset"` so the renderer is available in XML config and `SimpleLayout` without manual registration.
- `ShortDateLayoutRenderer` — fixed `GetValue`: when `UniversalTime=false`, `ToLocalTime()` is now only called when `timestamp.Kind == Utc`. Calling it unconditionally on `DateTimeKind.Unspecified` caused an off-by-one date on UTC-offset machines.
- Tests updated in `DateTests` and `LongDateTests` to cover the new renderer.

### DIFF
--- a/src/NLog/Config/AssemblyExtensionTypes.cs
+++ b/src/NLog/Config/AssemblyExtensionTypes.cs
@@ -115,6 +115,7 @@ namespace NLog.Config
             layoutRendererFactory.RegisterType<NLog.LayoutRenderers.CounterLayoutRenderer>("counter", checkTypeExists);
             layoutRendererFactory.RegisterType<NLog.LayoutRenderers.CurrentDirLayoutRenderer>("currentdir", checkTypeExists);
             layoutRendererFactory.RegisterType<NLog.LayoutRenderers.DateLayoutRenderer>("date", checkTypeExists);
+            layoutRendererFactory.RegisterType<NLog.LayoutRenderers.DateTimeOffsetLayoutRenderer>("datetimeoffset", checkTypeExists);
             layoutRendererFactory.RegisterType<NLog.LayoutRenderers.DbNullLayoutRenderer>("dbnull", checkTypeExists);
             layoutRendererFactory.RegisterType<NLog.LayoutRenderers.DirectorySeparatorLayoutRenderer>("dirseparator", checkTypeExists);
             layoutRendererFactory.RegisterType<NLog.LayoutRenderers.EnvironmentLayoutRenderer>("environment", checkTypeExists);
--- a/src/NLog/LayoutRenderers/DateTimeOffsetLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/DateTimeOffsetLayoutRenderer.cs
@@ -1,0 +1,131 @@
+//
+// Copyright (c) 2004-2024 Jaroslaw Kowalski <jaak@jkowalski.net>, Kim Christensen, Julian Verdurmen
+//
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions
+// are met:
+//
+// * Redistributions of source code must retain the above copyright notice,
+//   this list of conditions and the following disclaimer.
+//
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution.
+//
+// * Neither the name of Jaroslaw Kowalski nor the names of its
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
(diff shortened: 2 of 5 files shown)


snakefoot:
Curious why the conversion introduced with NLog v6.1.2 as part of #6116 is not working for you?
```csharp
if (typeCode == TypeCode.DateTime && typeof(DateTimeOffset) == propertyType)
   return new DateTimeOffset(convertibleValue.ToDateTime(formatProvider));
```
See also https://github.com/NLog/NLog.Targets.Database/pull/66 where `${date}` converts using `DbType.DateTimeOffset` (without string-conversion)

LorneCash:
I had tried to update my code before and it wasn't working but now it works fine with just "${date:format=O}" so I don't know what I did before but please disregard this PR.  I'm fine with how it is currently. I'm sorry for the trouble.

snakefoot:
It should also work if you do this `"${date}"` (without specifying `format=O`, so skipping the string-parsing)

LorneCash:
It does, thanks.

PR #6167, opened 2026-05-03
TemplateEnumerator - MoveNext without upfront allocation of Hole-name

snakefoot:
MoveNext() is used to probe the message-template. Reduce the overhead of the probe-logic.


snakefoot:
NET10 code-path shows no improvement from these changes.

PR #6175, opened 2026-05-16
Enum parse from string maps empty string to default value (no flags)

snakefoot:


snakefoot:
Think it will lead to more confusion than user-friendliness.

PR #6180, opened 2026-05-17
feat(LimitingWrapper): expose MessagesDroppedCount

### LINKED ISSUE #6173: Expose a property that is a count of dropped messages from a LimitingWrapper
Add a count of dropped messages from limiters.  This would be very useful to expose a variable that can be used in the body content so that we can get an idea of how many messages were dropped by the limiter since the last email was sent.

It can be helpful for email messages where you must limit how many that can be sent but still get a better idea of how many errors are occurring by just looking at the email.

wyf027:
## Summary

- Adds `MessagesDroppedCount` on `LimitingTargetWrapper`, incremented when a log event is discarded after `MessageLimit` is reached.
- Resets together with `MessagesWrittenCount` when the current interval expires.
- Adds unit tests for discard counting and interval reset.

Useful for monitoring how many events were suppressed between emails (e.g. read the property from code or target diagnostics).

## Test plan

- [x] `WriteMoreMessagesThanMessageLimitDiscardsExcessMessages` asserts `MessagesDroppedCount == 1`
- [x] `MessagesDroppedCountResetsWhenIntervalExpires`
- [ ] CI (no local .NET SDK in contributor environment)

Fixes #6173

### DIFF
--- a/src/NLog/Targets/Wrappers/LimitingTargetWrapper.cs
+++ b/src/NLog/Targets/Wrappers/LimitingTargetWrapper.cs
@@ -113,7 +113,13 @@ namespace NLog.Targets.Wrappers
         public int MessagesWrittenCount { get; private set; }
 
         /// <summary>
-        /// Initializes the target and resets the current Interval and <see cref="MessagesWrittenCount"/>.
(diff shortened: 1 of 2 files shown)


snakefoot:
Are you sure this will resolve #6173 ? (Sounds like they want to include the counter in the email-body)

Maybe review the code from the AI-bot, before submitting pull-request.

PR #6210, opened 2026-06-07
ScopeContext - Fix List capacity for GetAllNestedStates

jnyrup:
Parenthesize `??` as `+` binds stronger


snakefoot:
Nicely spotted. Thank you for the contribution.

PR #6215, opened 2026-06-11
FileTarget - Added FileLifecycleHooks for extending archive logic

### LINKED ISSUE #6213: Make FileTarget extendable for further cleanup logic.
I've got a new requirement to compress completed/closed logs as zip archive and store them for a number of days before deleting the zip archives. Looking at the `FileTarget` class I cannot see a way to extend with the additional cleanup logic. But with very minimal changes it would be possible to extend the cleanup. Some possibilities would be:
* Change `FileTarget` so that it is possible to inherit from it and change its behavior, for example by making `CloseFile` virtual and overwrite it 
* Extend at the level of `IFileArchiveHandler`; It could be an idea to make `
.DeleteOldArchiveFile` virtual and overwrite it, in addition to this `FileTarget` would need to change so that it is possible to set/create a custom `IFileArchiveHandler`
* Do something similar to Serilog; Serilog has `FileLifecycleHooks` on its `FileSink`. `FileLifecycleHooks` is an optional argument for `FileSink`, if provided `FileSink` calls methods on it when some specific actions happen for example `OnFileDeleting`.  

If you think it would be a good feature for NLog to make something like this happen I would like to help implement this in what ever version you think would make the most sense.

Dave-Senn:
PR that fixes https://github.com/NLog/NLog/issues/6213

Introduces `FileLifecycleHooks` for the NLog FileTarget, providing extension points for custom logic around the lifecycle of log files and the FileTarget itself.

Lifecycle hooks can be used to:

- **Customize initialization** - Run custom logic when the FileTarget is initialized.
- **Customize file streams** - Wrap the underlying stream to add encryption, compression, buffering, or other processing.
- **Process closed files** - Perform custom actions when NLog closes an individual log file.
- **Process files before deletion** - Copy, compress, upload, or otherwise process a file before NLog deletes it.
- **Perform target cleanup** - Execute custom cleanup logic when the FileTarget is shut down.

For example, `OnFileDeleting` can be used to create a ZIP archive of an old log file before NLog deletes it, while `OnFileOpened` can wrap the output stream to transparently encrypt log data.


src/NLog.Targets.AtomicFile/AtomicFileTarget.cs
```diff
@@ -100,34 +115,80 @@ private Stream CreateAtomicFileStream(string filePath)
...
                 FileMode.Append,
                 System.Security.AccessControl.FileSystemRights.AppendData | System.Security.AccessControl.FileSystemRights.Synchronize, // <- Atomic append
                 fileShare,
                 bufferSize: 1,  // No internal buffer, write directly from user-buffer
                 FileOptions.None);
+
+            if (Hooks is not null)
+                try
+                {
+                    netFxStream = Hooks.OnFileOpened(filePath, netFxStream);
```
snakefoot: Can the changes be isolated to FileTarget only, and just have this code where FileTarget calls `CreateFileStream` ?

src/NLog/Targets/FileTarget.cs
```diff
@@ -464,6 +463,11 @@ public string ArchiveSuffixFormat
         /// <docgen category='Archival Options' order='50' />
         public bool WriteFooterOnArchivingOnly { get; set; }
 
+        /// <summary>
+        /// Optional file lifecycle hooks.
+        /// </summary>
+        protected internal FileLifecycleHooks? Hooks { get; }
```
snakefoot: Thinking it should be a private-member only

src/NLog/Targets/FileTarget.cs
```diff
@@ -582,6 +608,16 @@ protected override void InitializeTarget()
         /// <inheritdoc />
         protected override void CloseTarget()
         {
+            if (Hooks is not null)
+                try
+                {
+                    Hooks.OnTargetClose(this);
```
snakefoot: Can we skip the explicit null-check and just do `Hooks?.OnTargetClose(this);`

src/NLog/Targets/FileTarget.cs
```diff
@@ -995,6 +1031,15 @@ private void CloseFile(string filename, OpenFileAppender openFile)
             finally
             {
                 openFile.FileAppender.Dispose();
+                if (Hooks is not null)
+                    try
+                    {
+                        Hooks.OnFileClosed(openFile.FileAppender.FilePath);
```
snakefoot: Can we skip the explicit null-check and just do `Hooks?.OnFileClosed(openFile.FileAppender.FilePath);`

src/NLog/Targets/FileTarget.cs
```diff
@@ -1359,7 +1404,18 @@ protected virtual Stream CreateFileStream(string filePath, int bufferSize)
                 fileMode = FileMode.Create; // Create or truncate
             }
 
-            return new FileStream(filePath, fileMode, FileAccess.Write, fileShare, bufferSize);
+            Stream stream = new FileStream(filePath, fileMode, FileAccess.Write, fileShare, bufferSize);
+            if (Hooks is not null)
+                try
+                {
+                    stream = Hooks.OnFileOpened(filePath, stream);
```
snakefoot: Can this logic be moved into the method that calls CreateFileStream ?

src/NLog/Targets/FileLifecycleHooks.cs
```diff
@@ -0,0 +1,95 @@
...
+        public virtual void OnTargetClose(FileTarget target)
+        {
+        }
+
+        /// <summary>
+        /// Creates a linked hook combining the current and given <see cref="FileLifecycleHooks"/>.
+        /// </summary>
+        /// <param name="hooks">Hooks to link with current <see cref="FileLifecycleHooks"/>. Methods on <paramref name="hooks"/> will be invoked after the current <see cref="FileLifecycleHooks"/> methods.</param>
+        /// <returns>Returns a <see cref="FileLifecycleHooks"/> chaining this <see cref="FileLifecycleHooks"/> and <paramref name="hooks"/>. Methods of this instance will be invoked before methods of <paramref name="hooks"/>.</returns>
+        public FileLifecycleHooks CreateLinkedHooks(FileLifecycleHooks hooks) =>
```
snakefoot: Think this should be an extension method. And maybe leave it out in the initial version, unless you have an urgent need to chain things together?

src/NLog/Targets/FileTarget.cs
Dave-Senn: done

src/NLog/Targets/FileTarget.cs
Dave-Senn: done

src/NLog/Targets/FileTarget.cs
Dave-Senn: I don’t think private would work. Hooks need to be accessible from BaseFileArchiveHandler for the OnFileDeleting method.

src/NLog/Targets/FileTarget.cs
Dave-Senn: Yes, this is much better! => Done.

src/NLog/Targets/FileLifecycleHooks.cs
Dave-Senn: I don't personally have a use case for this. I only added it because I thought this was what you meant by:

"But I see the potential of being able to chain multiple FileLifecycleHooks together and imbue them with several properties."

Would you prefer that I remove it, or refactor it into an extension method instead?

src/NLog.Targets.AtomicFile/AtomicFileTarget.cs
Dave-Senn: Done

src/NLog/Targets/FileLifecycleHooks.cs
snakefoot: Think it should be removed, until a need actually shows up.

src/NLog/Targets/FileTarget.cs
snakefoot: Yes Then I guess it should be changed to be "internal" only.

src/NLog.Targets.AtomicFile/AtomicFileTarget.cs
```diff
@@ -54,6 +54,21 @@ public class AtomicFileTarget : FileTarget
...
+        /// </summary>
+        public AtomicFileTarget()
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AtomicFileTarget" /> class.
+        /// </summary>
+        /// <param name="hooks">Hooks to perform actions on certain life cycle events.</param>
+        public AtomicFileTarget(FileLifecycleHooks hooks) : base(hooks)
```
snakefoot: Unless you really need support for AtomicFileTarget, then I think FileLifecycleHooks should only be supported by pure FileTarget.

src/NLog.Targets.GZipFile/GZipFileTarget.cs
```diff
@@ -62,6 +62,15 @@ public GZipFileTarget(string name) : this()
             Name = name;
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="GZipFileTarget" /> class.
+        /// </summary>
+        /// <param name="hooks">Hooks to perform actions on certain life cycle events.</param>
+        public GZipFileTarget(FileLifecycleHooks hooks) : base(hooks)
```
snakefoot: Unless you really need support for GZipFileTarget, then I think FileLifecycleHooks should only be supported by pure FileTarget. Since FileLifecycleHooks is now an alternative way of adding compression for the file-stream.

src/NLog/Targets/FileTarget.cs
Dave-Senn: Changed to internal

src/NLog/Targets/FileLifecycleHooks.cs
Dave-Senn: Removed CreateLinkedHooks and linked class

src/NLog.Targets.AtomicFile/AtomicFileTarget.cs
Dave-Senn: Removed hooks from AtomicFileTarget

src/NLog.Targets.GZipFile/GZipFileTarget.cs
Dave-Senn: Removed hooks from GZipFileTarget

src/NLog/Targets/FileTarget.cs
```diff
@@ -1359,7 +1413,8 @@ protected virtual Stream CreateFileStream(string filePath, int bufferSize)
                 fileMode = FileMode.Create; // Create or truncate
             }
 
-            return new FileStream(filePath, fileMode, FileAccess.Write, fileShare, bufferSize);
+            var stream = new FileStream(filePath, fileMode, FileAccess.Write, fileShare, bufferSize);
```
snakefoot: Can this be returned to its original way:
```
return new FileStream(filePath, fileMode, FileAccess.Write, fileShare, bufferSize);
```

tests/NLog.UnitTests/Targets/FileTargetTests.cs
```diff
@@ -4568,5 +4654,56 @@ private static void AssertFileContentsEndsWith(string fileName, string contents,
...
+                _callRecording.Add($"{_name}_{nameof(OnFileOpened)}_{filePath}_{underlyingStream.GetType().Name}");
+                return _throwOnCall ? throw new DivideByZeroException("Test error") : underlyingStream;
+            }
+
+            public override void OnTargetInitialize(FileTarget target)
+            {
+                _callRecording.Add($"{_name}_{nameof(OnTargetInitialize)}_{target.Name}");
+                if (_throwOnCall)
+                    throw new DivideByZeroException("Test error");
+            }
```
snakefoot, line 4714: Can the mock-method-overrides be updated to call base. Ex:
```c#
            public override void OnTargetInitialize(FileTarget target)
            {
                _callRecording.Add($"{_name}_{nameof(OnTargetInitialize)}_{target.Name}");
                if (_throwOnCall)
                    throw new DivideByZeroException("Test error");
                 base.OnTargetInitialize(target);
            }
```
So Sonar Code Coverage will become happy?

snakefoot:
I think the changes looks good. Have made 2 additional comments, that should be easy to fix.

Do you have anything else that you want to add? 

Notice you can play with the pre-release nuget-package here from the build-server:
- https://ci.appveyor.com/project/nlog/nlog/builds/54218867/job/586q0nb41712ecxw/artifacts

Just download to local folder. Rename from `.zip` to `.nuget` and then in Visual Studio setup nuget-source from local-folder.

src/NLog/Targets/FileLifecycleHooks.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// Called before a log file gets deleted.
+        /// This can be used to copy old logs to an archive location or send to a backup server.
+        /// </summary>
+        /// <param name="filePath">The full path to the file being deleted.</param>
+        public virtual void OnFileDeleting(string filePath)
+        {
+        }
+
+        /// <summary>
+        /// Called after an open file was closed.
```
snakefoot: Maybe provide some examples of when this will be called. Ex. part of file-archive-event, part of file-open-limit-exceeded, target-closed, etc.

And one can consider to override OnTargetClose if one need detect what OnFileClosed-events comes from target-closing.

Maybe also explain that by default then FileTarget will not always close the old files, before moving to the next file. Unless having configured `OpenFileCacheSize = 1`

src/NLog/Targets/FileLifecycleHooks.cs
```diff
@@ -0,0 +1,87 @@
...
+namespace NLog.Targets
+{
+    /// <summary>
+    /// Enables hooking into log file lifecycle events.
+    /// Hooks run synchronously and therefore may affect responsiveness of the application if long operations are performed.
+    /// </summary>
+    public abstract class FileLifecycleHooks
+    {
+        /// <summary>
+        /// Called before a log file gets deleted.
```
snakefoot: Maybe explain what is required for files to be deleted.

Since you are trying to open a door into a dark room of the NLog FileTarget, then I guess some light should be made for those trying to enter the room.

src/NLog/Targets/FileTarget.cs
```diff
@@ -565,6 +581,15 @@ protected override void InitializeTarget()
             if (Layout.IsNullOrEmpty(FileName))
                 throw new NLogConfigurationException("FileTarget FileName-property must be assigned. FileName is needed for file writing.");
 
+            try
+            {
+                Hooks?.OnTargetInitialize(this);
```
snakefoot: Maybe any exceptions from OnTargetInitialize should not be caught, since the configured Hooks are having issues? Instead the FileTarget should go into error-state, and be disabled for output.

src/NLog/Targets/FileTarget.cs
```diff
@@ -464,6 +463,11 @@ public string ArchiveSuffixFormat
         /// <docgen category='Archival Options' order='50' />
         public bool WriteFooterOnArchivingOnly { get; set; }
 
+        /// <summary>
+        /// Optional file lifecycle hooks.
+        /// </summary>
+        internal FileLifecycleHooks? Hooks { get; }
```
snakefoot: Maybe rename the `Hooks`-property to `FileLifeCycle` ?

src/NLog/Targets/FileTarget.cs
Dave-Senn: Done

tests/NLog.UnitTests/Targets/FileTargetTests.cs
Dave-Senn: Sure, we all like when our code analyzers are happy; done.

src/NLog/Targets/FileTarget.cs
Dave-Senn: Done

src/NLog/Targets/FileTarget.cs
Dave-Senn: Yes, that makes sense. What is the correct way to put the target into an error state and disable further output?

src/NLog/Targets/FileLifecycleHooks.cs
Dave-Senn: I've improved the documentation for OnFileDeleting

src/NLog/Targets/FileTarget.cs
snakefoot: By not catching the exception, then the NLog engine will register Target initialization as failed.

src/NLog/Targets/FileLifecycleHooks.cs
```diff
@@ -0,0 +1,90 @@
...
+        /// <param name="underlyingStream">The underlying <see cref="Stream"/> opened on the log file.</param>
+        /// <returns>The <see cref="Stream"/> NLog should use when writing events to the log file.</returns>
+        public virtual Stream OnFileOpened(string filePath, Stream underlyingStream)
+            => underlyingStream;
+
+        /// <summary>
+        /// Called when the target gets initialized.
+        /// </summary>
+        /// <param name="target">Initialized target.</param>
+        public virtual void OnTargetInitialize(FileTarget target)
```
snakefoot: Maybe add a remark to XML-docs that if Exception is thrown then FileTarget will be put in error-state and be disabled for output.

Maybe also update the XML-docs Summary, that overriding this method can be useful for overriding/re-assigning FileTarget-properties.

src/NLog/Targets/FileArchiveHandlers/BaseFileArchiveHandler.cs
```diff
@@ -438,6 +438,15 @@ protected bool DeleteOldArchiveFile(string filepath, string archiveCleanupReason
                 {
                     InternalLogger.Info("{0}: Cleanup file archive {1}. Delete file: '{2}'", _fileTarget, archiveCleanupReason, filepath);
                     _fileTarget.CloseOpenFileBeforeArchiveCleanup(filepath);
+                    try
+                    {
+                        _fileTarget.FileLifeCycle?.OnFileDeleting(filepath);
```
snakefoot: Maybe move this logic into FileTarget internal-method, so all logic calling FileLifeCycle is inside the FileTarget-class ? (Can make FileLifeCycle private)

src/NLog/Targets/FileLifecycleHooks.cs
Dave-Senn: Done.

src/NLog/Targets/FileArchiveHandlers/BaseFileArchiveHandler.cs
Dave-Senn: Do you think `FileLifeCycle?.OnFileDeleting(filepath);` should be moved to `CloseOpenFileBeforeArchiveCleanup`? To me, it looks like this should work.

snakefoot: Yes that would also work. Maybe rename `CloseOpenFileBeforeArchiveCleanup` to `CloseOpenFileBeforeArchiveFileDelete`

src/NLog/Targets/FileLifecycleHooks.cs
Dave-Senn: Done

src/NLog/Targets/FileTarget.cs
Dave-Senn: Thanks for the info; Done

src/NLog/Targets/FileArchiveHandlers/BaseFileArchiveHandler.cs
Dave-Senn: Done

src/NLog/Targets/FileTarget.cs
```diff
@@ -954,8 +981,17 @@ private void PruneOpenFileCache()
             }
         }
 
-        internal void CloseOpenFileBeforeArchiveCleanup(string filepath)
+        internal void CloseOpenFileBeforeArchiveFileDelete(string filepath)
         {
+            try
+            {
+                FileLifeCycle?.OnFileDeleting(filepath);
```
snakefoot: Think this logic should be moved down, and executed after the FileTarget has closed its File-Handles.

tests/NLog.UnitTests/Targets/FileTargetTests.cs
```diff
@@ -111,6 +110,93 @@ public void SimpleFileTest(bool keepFileOpen)
             }
         }
 
+        [Fact]
+        public void HooksTest()
```
snakefoot: Can it be renamed from `HooksTest` to `FileLifeCycleTest` ?

tests/NLog.UnitTests/Targets/FileTargetTests.cs
```diff
@@ -111,6 +110,93 @@ public void SimpleFileTest(bool keepFileOpen)
...
+            finally
+            {
+                if (File.Exists(logFile))
+                    File.Delete(logFile);
+            }
+        }
+
+
+        [Fact]
+        public void Hooks_ErrorHandlingTest()
```
snakefoot: Can it be renamed from `Hooks_ErrorHandlingTest` to `FileLifeCycle_ErrorHandlingTest` ?

Dave-Senn: Done

tests/NLog.UnitTests/Targets/FileTargetTests.cs
Dave-Senn: Done

src/NLog/Targets/FileTarget.cs
Dave-Senn: Done

src/NLog/Targets/FileLifecycleHooks.cs
```diff
@@ -0,0 +1,107 @@
...
+        /// </summary>
+        /// <remarks>
+        ///     Note that<see cref="OnFileClosed" /> is not guaranteed to be invoked before <see cref = "FileTarget" /> advances to the next file.
+        ///     You may override <see cref="OnTargetClose" /> to determine whether a file-closed event was caused by <see cref="FileTarget" /> shutting down or by <see cref="FileTarget" /> switching to a different file.
+        ///     <see cref="FileTarget"/> may close files in the following scenarios:
+        ///     - <see cref="FileTarget"/> archives the current file because of an archive constraint and switches to the next file.
+        ///     - An open file is deleted from storage and no longer exists.
+        ///     - If <see cref="FileTarget.DeleteOldFileOnStartup"/> is enabled, <see cref="FileTarget"/> may close open files.
+        ///     - <see cref="FileTarget"/> closes all open files when the target itself is shutting down.
+        ///     - If <see cref="FileTarget.OpenFileMonitorTimerInterval"/> is greater than 0, <see cref="FileTarget"/> may close files that have not received log events for a period of time.
```
snakefoot: Maybe better to mention using FileTarget.OpenFileCacheSize = 1 to prevent having multiple file open, than the exotic OpenFileMonitorTimerInterval

snakefoot:
I guess those 2 additional comments suddenly exploded into a lot more. But thank you for your quick resolve. I'm happy about the result, and the Sonar-Report also looks good. 

Do you have anything else that you want to add? And have you verified that the new FileTarget API provides what you need?
- Nuget-Package PreRelease: https://ci.appveyor.com/project/nlog/nlog/builds/54244123/job/nc1ankfowj447nls/artifacts

src/NLog/Targets/FileLifecycleHooks.cs
Dave-Senn: Comment updated.

tests/NLog.UnitTests/Targets/FileTargetTests.cs
```diff
@@ -111,6 +110,93 @@ public void SimpleFileTest(bool keepFileOpen)
...
+            try
+            {
+                var callRecording = new List<string>();
+                var fileTarget = new FileTarget(new MockFileLifecycleHooks("hooks", callRecording))
+                {
+                    FileName = Layout.FromLiteral(logFile),
+                    Name = "my-file",
+                    LineEnding = LineEndingMode.LF,
+                    Layout = "${level} ${message}",
+                    OpenFileCacheTimeout = 0
```
snakefoot: Any reason for assigning OpenFileCacheTimeout  ? Maybe more interesting to assign `ArchiveAboveSize = 5`  and MaxArchiveFiles = 1 ?

tests/NLog.UnitTests/Targets/FileTargetTests.cs
```diff
@@ -111,6 +110,93 @@ public void SimpleFileTest(bool keepFileOpen)
...
+            try
+            {
+                var callRecording = new List<string>();
+                var fileTarget = new FileTarget(new MockFileLifecycleHooks("hooks", callRecording, true))
+                {
+                    FileName = Layout.FromLiteral(logFile),
+                    Name = "my-file",
+                    LineEnding = LineEndingMode.LF,
+                    Layout = "${level} ${message}",
+                    OpenFileCacheTimeout = 0
```
snakefoot, line 177: Any reason for assigning OpenFileCacheTimeout  ? Maybe more interesting to assign `ArchiveAboveSize = 5`  and MaxArchiveFiles = 1 ?

tests/NLog.UnitTests/Targets/FileTargetTests.cs
Dave-Senn: I've updated the test

tests/NLog.UnitTests/Targets/FileTargetTests.cs
Dave-Senn: I've updated the test

Dave-Senn:
> I guess those 2 additional comments suddenly exploded into a lot more. But thank you for your quick resolve. I'm happy about the result, and the Sonar-Report also looks good.
> 
> Do you have anything else that you want to add? And have you verified that the new FileTarget API provides what you need?
> 
>     * Nuget-Package PreRelease: https://ci.appveyor.com/project/nlog/nlog/builds/54244123/job/nc1ankfowj447nls/artifacts

Thank you very much for your time and frequent reviews. I've tested the new behavior with the package from the build and it works for my use case. I don't have anything I would like to add to this PR.

snakefoot:
I usually make a NLog patch-releases once a month (like now), and postpone bigger features for the minor-releases once a year (like 5-7 month from now)

How urgent is this feature for you?

tests/NLog.UnitTests/Targets/FileTargetTests.cs
```diff
@@ -4568,5 +4661,58 @@ private static void AssertFileContentsEndsWith(string fileName, string contents,
...
+            {
+                _callRecording.Add($"{_name}_{nameof(OnFileDeleting)}_{filePath}");
+                if (_throwOnCall)
+                    throw new DivideByZeroException("Test error");
+                base.OnFileDeleting(filePath);
+            }
+
+            public override Stream OnFileOpened(String filePath, Stream underlyingStream)
+            {
+                _callRecording.Add($"{_name}_{nameof(OnFileOpened)}_{filePath}_{underlyingStream.GetType().Name}");
```
snakefoot, line 4705: Can we add call to `base.OnFileOpened(filePath, underlyingStream);` for perfect code coverage ?

Dave-Senn: Done

Dave-Senn:
> I usually make a NLog patch-releases once a month (like now), and postpone bigger features for the minor-releases once a year (like 5-7 month from now)
> 
> How urgent is this feature for you?

Of course, I would like to have access to this feature as fast as possible, but it’s not that urgent. 5-7 Months work for me 😊

snakefoot:
> @Dave-Senn  Of course, I would like to have access to this feature as fast as possible, but it’s not that urgent. 5-7 Months work for me 😊

Then I will release NLog v6.2 after my summer holiday around 1st August.

snakefoot:
NLog v6.2 has now been released, that includes FileLifecycleHooks for the FileTarget:
- https://www.nuget.org/packages/NLog/6.2.0

Thank you for the contribution

PR #6218, opened 2026-06-13
ScopeContext - Optimize collection of properties in reverse order

snakefoot:
- [ ] Wait for NLog v6.2

NLog v5 introduced the ScopeContext using a "lazy" stack, compared to NLog v4 that maintained property-dictionary upfront. The "lazy" stack is much faster, when not all the scopes produce log-output with scope-properties (Ex. only logging scope on Error)

The initial NLog v5.0 had an optimization where it tried to guess the number of properties in a scope, so it could allocate a collection with the expected initial capacity. But this optimization was implemented so it enumerated the "lazy" stack using recursion. This caused NLog v5 to explode with stack-overflow when the object-graph grew too long. The recursion-logic was fixed with NLog v5.1.3 but to avoid rocking the boat, then it was decided to keep the reverse-stack-ordering (oldest first). 

The oldest first ordering introduces a performance hit. Because the newest property is at the top of the "lazy" stack. When NLog collects scope-properties, then it must perform list-insert-first and move all existing items in the list (expensive).

Spending time of on ensuring special ordering for something that operates as Dictionary-collection seems like a waste of time. Instead of having oldest first, then it is more optimal to have newest first.

This includes a bonus-optimization in the merge/collapse logic of the "lazy"-stack (reducing the object-graph), where it would first allocate List (for ordering) and then convert into Dictionary. Now it will just add directly to the Dictionary, so skipping extra List-allocation and double-insert.

### DIFF
--- a/src/NLog/Internal/ScopeContextAsyncState.cs
+++ b/src/NLog/Internal/ScopeContextAsyncState.cs
@@ -113,7 +113,7 @@ namespace NLog.Internal
                         return _allProperties = MergeUniqueProperties(_allProperties, properties);
                     }
 
-                    AddProperties(properties);
+                    CollectProperties(properties, _propertyCollector);
                 }
 
                 return _allProperties = EnsureUniqueProperties(_allProperties);
--- a/src/NLog/Internal/ScopeContextAsyncState.cs
+++ b/src/NLog/Internal/ScopeContextAsyncState.cs
@@ -36,6 +36,7 @@
 namespace NLog.Internal
 {
     using System;
+    using System.Collections;
     using System.Collections.Generic;
 
     /// <summary>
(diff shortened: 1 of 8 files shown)


snakefoot:
Performed a benchmark, and it was hard to see cost-reduction from removing list-insert-first, when staying at the normal 2-5 scope-properties. The real win was the new object-graph-merge-logic, that wrote directly into the Dictionary (and skipped list-allocation). Maybe more than 10 scope-properties are needed to see the cost of list-insert-first with memmove. Or maybe the reduction in cost was eaten by the new virtual-function-call using ICollection-interface (to allow writing directly to dictionary), instead of non-virtual List-Add.

But the removal of the List-Insert-First is actually the door that allows the new object-graph-merge-logic. Since ICollection-Add can be used on both List and Dictionary. And Dictionary always adds to the end without supporting re-ordering like List.

PR #6219, opened 2026-06-14
ScopeContext - CaptureCompleted  now calls CollectProperties

snakefoot:


snakefoot:
Seems AddProperties is needed, since ordering is important

PR #6231, opened 2026-07-10
Replacing ProcessInfoLayoutRenderer with ProcessStartLayoutRenderer

snakefoot:
- [x] Wait for NLog v6.2

NET11 reduces the AOT-filesize when not using exotic features from the Process-object.

Changed `${processinfo}` to only support Id + Name + WorkingSet, and everything else are either hardcoded values or compatibility values.

Instead introduced `${processstart}` with the following options:
- Format
- UniversalTime
- Culture

`${processstart}` outputs the same value during the entire AppDomain-lifetime, which allows NLog to convert into literal for optimal performance.

And there is also the existing `${processid}` + `${processname}` + `${processtime}` + `${gc}`. And have now introduced support for `${gc:workingset}`

**This PR with NET10 (Guess NET11 will be better)**
`ProcessPath: 'C:\projects\nlog\Tests\TestTrimPublish\bin\release\net10.0\win-X64\publish\TestTrimPublish.exe' with FileSize: 5349376`

**NLog v6.1.4**
`ProcessPath: 'C:\projects\nlog\Tests\TestTrimPublish\bin\release\net10.0\win-X64\publish\TestTrimPublish.exe' with FileSize: 5384192`


snakefoot:
Updated the wiki:
- https://github.com/NLog/NLog/wiki/ProcessStart-Layout-Renderer
- https://github.com/NLog/NLog/wiki/ProcessInfo-Layout-Renderer

Issue #6235, opened 2026-07-17
New immutable LogEvent struct for non-allocation logging


snakefoot:
Introduce a new logging pipeline in NLog that uses an immutable LogEvent struct instead of the existing mutable LogEventInfo class.

The goal is to provide a safer, immutable logging pipeline that prevents mutation of log events, while enabling an allocation-efficient execution path where you only pay for the execution-context capture you need. Backwards compatibility for the old LogEventInfo pipeline will still be maintained.

**Immutable log events**

LogEventInfo is currently mutable, which means targets, layouts, and wrappers can accidentally modify the same log event while it travels through the logging pipeline.

This can lead to subtle bugs where one target changes data that is later observed by another target.

An immutable LogEvent would eliminate these side effects, making the logging pipeline easier to reason about and safer for target authors.

**Non-allocation logging**

A LogEvent implemented as a struct could be passed through the logging pipeline without heap allocation for the common case.

The struct would contain storage for a small number of built-in properties and event properties (Ex. 5), allowing many log events to remain allocation-free. Only when that capacity is exceeded would it fall back to allocating a Array / Dictionary (or similar).

**Mutate by clone**
NLog Target Wrappers often want to imbue with additional behavior, or make other adjustments of the LogEvent. There will be extension methods that can clone into a local/temporary "builder" which can be mutated, and finally build a new updated LogEvent.

**Compatibility**

The new LogEvent-struct will get a new, optimized pipeline. But leaving the old legacy pipeline with LogEventInfo.

Legacy targets continue receiving LogEventInfo. But new targets will receive the immutable LogEvent.

Thinking that LogEvent will have the legacy LogEventInfo as part of the Layout-Precalculate-Cache, when detecting legacy Target / Layout in the logging-output. Thus the LogEvent-struct can provide legacy LogEventInfo-object, when required by the ouput-target-chain.

Because the LogEvent struct will likely be relatively large, the best performance will be achieved on platforms that support passing structs by in (ref readonly) to reduce copy-by-value-overhead.

snakefoot:
Maybe it could be the driver for other LogEvent-changes in NLog v7:
- Use UTC-time by default (instead of local-time by default)
- Culture Invariant Formatting by default (instead of culture-aware by default)

[...]

PR #6252, opened 2026-08-18
DateLayoutRenderer - Changed default Format to yyyy-MM-dd HH:mm:ss.ffff

### LINKED ISSUE #6232: DateLayoutRenderer default Format should be culture invariant
Little weird that DateLayoutRenderer has the default Format:
- `yyyy/MM/dd HH:mm:ss.fff`

Why not follow the default-format found every where else in NLog:
- `yyyy-MM-dd HH:mm:ss.ffff`

Interesting that `${date}` "originally" was born with default `Format = G`, that would follow the local culture. Then later hard-coded to one culture with explicit Format-string (with `ffff`). Think that Logging library should stay culture invariant unless requested otherwise.

Also `$[processtime}` should match the following format (Instead of `fff`):
- `HH:mm:ss.ffff`

HarnageaGabriel:
Fixes #6232.

## Background / breaking-change tradeoff

This issue is tagged `breaking change` / `breaking behavior change`. Investigation showed `DateLayoutRenderer.Culture` already defaults to `CultureInfo.InvariantCulture` (via #5740, already on `dev`), so `${date}` output was already stable across thread/system culture — there was no live culture-dependency bug to fix.

I posted a scoping comment on the issue asking whether the maintainer wanted (a) a straight default change, (b) a non-breaking opt-in flag, or (c) something else, since the actual remaining gap is just the default `Format` *string* itself being inconsistent with the rest of NLog. @snakefoot responded that there's no bug — it's "just a funny default value" — and that this targets the **NLog v7 milestone**, so a straight default change is fine there.

This PR implements that straight change:

## Changes

- `DateLayoutRenderer`'s default `Format` changed from `yyyy/MM/dd HH:mm:ss.fff` to `yyyy-MM-dd HH:mm:ss.ffff`, matching the convention used elsewhere in NLog (`${longdate}`, `InternalLogger`'s internal timestamp format).
- Updated `DefaultDateTest` in `DateTests.cs` to assert the new default output.
- Added a `CHANGELOG.md` entry under a new "Version 7.0 (Unreleased)" / "Breaking Changes" section.

This is a **breaking behavior change**: anyone using `${date}` without an explicit `Format` will see different default output. Users who already set `Format` explicitly are unaffected.

## Testing

- `dotnet build src/NLog/NLog.csproj -c Debug` — succeeds.
- `dotnet test tests/NLog.UnitTests/NLog.UnitTests.csproj --filter "FullyQualifiedName~DateTests"` — 27/27 pass (net10.0 and net462).
- Full `NLog.UnitTests` suite (net10.0) — 2587/2589 pass (2 skipped, unrelated to this change); the run's final "Out of memory" test-host crash is from an unrelated memory-heavy test, not exercised by or related to this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Date format: (year/month/day)
 
 ## Change Log
 
+### Version 7.0 (Unreleased)
+**Breaking Changes**
+- [#6232](https://github.com/NLog/NLog/issues/6232) DateLayoutRenderer - Changed default Format to `yyyy-MM-dd HH:mm:ss.ffff` for consistency with rest of NLog (breaking change, was `yyyy/MM/dd HH:mm:ss.fff`).
+
 ### Version 6.2 (2026/08/16)
 **Improvements**
 - [#6215](https://github.com/NLog/NLog/pull/6215) FileTarget - Added FileLifecycleHooks for extending archive logic. (@Dave-Senn)
--- a/src/NLog/LayoutRenderers/DateLayoutRenderer.cs
+++ b/src/NLog/LayoutRenderers/DateLayoutRenderer.cs
@@ -55,7 +55,7 @@ namespace NLog.LayoutRenderers
         /// </summary>
         public DateLayoutRenderer()
         {
-            _format = Format = "yyyy/MM/dd HH:mm:ss.fff";
+            _format = Format = "yyyy-MM-dd HH:mm:ss.ffff";
         }
 
         /// <summary>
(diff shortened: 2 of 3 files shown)


snakefoot:
You should probably not expect this to be merged before NLog v7 roadmap is ready. In a year or more (no deadline yet).

Also think this will be down-prioritized if start working on #6235, which require rewrite of everything to support a new logging pipeline, where all layoutrenderers has be rewritten, including this.

PR #6254, opened 2026-08-20
Replace System.Random usage with custom ScatterGenerator

jobvk:
## Summary

- Add a cross-target `ScatterGenerator` helper.
- Replace `System.Random` in file-open retry jitter and randomized target selection.

## Reason for change

[Datadog Runtime Code Analysis (IAST)](https://docs.datadoghq.com/security/code_security/iast/#supported-vulnerability-types)
reports uses of `System.Random` as **Weak Randomness**. This finding is associated
with [CWE-338: Use of Cryptographically Weak PRNG](https://cwe.mitre.org/data/definitions/338.html).

CWE-338 specifically concerns non-cryptographic random-number generators used in
a security context, such as generating keys, tokens, or session identifiers.
NLog currently uses random values only for retry jitter and target selection, so
these values do not protect a security boundary.

However, runtime security scanners cannot always determine that context and
report the use of `System.Random` in NLog-consuming services. Using
`ScatterGenerator` removes these findings while preserving the existing
randomized behavior and support for older target frameworks.


snakefoot:
Think removing/replacing Random-usage in NLog is an awesome idea (Both to remove false-positive-security-warning but also AOT-build-size). NLog does not need "perfect" random, as merely for simple jitter and the random-target that seldom have more than 10 as max-value (most likely 2 or 4), where the standard Random also produces repeated values.

Would this implementation work for you:
```c#
using System;

    /// <summary>
    /// Lightweight, non-cryptographic pseudo-random-number-generator (PRNG).
    /// Intended for jitter, load balancing, and other non-security-sensitive uses.
    /// </summary>
    /// <remarks>
    /// The output is predictable and provides no uniqueness guarantees.
    /// Must not be used for cryptographic purposes, security tokens, or unique identifiers.
    /// </remarks>
    internal sealed class ScatterGenerator
    {
        // Weyl-sequence increment derived from the golden ratio. Being odd makes it
        // coprime with 2^32, giving the Weyl sequence a full 2^32-period.
        private const uint Gamma = 0x9E3779B9u;

        private uint _state;

        public ScatterGenerator()
        {
            unchecked
            {
                // Mix the current timestamp with an identity-based hash to vary the initial
                // state across instances. Neither source provides uniqueness or entropy.
                _state = (uint)System.Diagnostics.Stopwatch.GetTimestamp()
                    ^ (uint)System.Runtime.CompilerServices.RuntimeHelpers.GetHashCode(this);
            }
        }

        /// <summary>
        /// Returns the next pseudo-random unsigned 32-bit value.
        /// </summary>
        public uint NextUInt()
        {
            unchecked
            {
                // Weyl sequence step: cheap deterministic state progression with a full 2^32 period.
                uint x = _state += Gamma;

                // MurmurHash3 fmix32: invertible avalanche transformation that
                // scatters the structured Weyl sequence across the output bits.
                x ^= x >> 16;
                x *= 0x85EBCA6Bu;
                x ^= x >> 13;
                x *= 0xC2B2AE35u;
                x ^= x >> 16;

                return x;
            }
        }

        /// <summary>
        /// Returns a pseudo-random value in the range [0, maxValue).
        /// </summary>
        public int Next(int maxValue)
        {
            if (maxValue < 0)
                throw new ArgumentOutOfRangeException(nameof(maxValue), maxValue, "MaxValue must be non-negative");

            if (maxValue == 0)
                return 0;

            // Multiply-high reduction maps the 32-bit output into [0, maxValue).
            // Bucket populations differ by at most one, avoiding the systematic
            // low-bucket bias introduced by naive modulo reduction.
            return (int)(((ulong)NextUInt() * (uint)maxValue) >> 32);
        }

        /// <summary>
        /// Returns a pseudo-random value in the range [minValue, maxValue).
        /// </summary>
        public int Next(int minValue, int maxValue)
        {
            if (minValue > maxValue)
                throw new ArgumentOutOfRangeException(nameof(maxValue), maxValue, $"MinValue={minValue} > MaxValue={maxValue}");

            if (minValue == maxValue)
                return minValue;

            unchecked
            {
                // Calculate the range as uint so the full signed Int32 range can be
                // represented, including int.MinValue to int.MaxValue.
                uint range = (uint)((long)maxValue - minValue);
                // Multiply-high reduction produces an offset in [0, range).
                uint offset = (uint)(((ulong)NextUInt() * range) >> 32);
                return minValue + (int)offset;
            }
        }
    }
```
I don't mind having multiple instances (Just replace `new Random` with this), than paying the price of ThreadStatic or Interlocked, that hurts the Cpu-Caches.

tests/SampleExtensions/MyConditionMethods.cs
```diff
@@ -33,17 +33,15 @@
...
     [ConditionMethods]
     public static class MyConditionMethods
     {
-        [ConditionMethod("myrandom")]
-        public static int Random(int max)
+        [ConditionMethod("myvalue")]
+        public static int Value(int value)
         {
-            return new Random().Next(max);  // NOSONAR
+            return value;
```
snakefoot: Maybe keep calling it "myrandom" but make it sloppy:
```csharp
        [ConditionMethod("myrandom")]
        public static int Random(int max)
        {
           return (int)((ulong)Stopwatch.GetTimestamp() % (uint)max); // Sloppy by default
        }
```

snakefoot:
Thank you for adjusting your pull-request to my proposal.

After some considerations then I think "ScatterGenerator" is a better class-name than my other suggestions:  "VariationGenerator" / "PseudoRandom" / "NonCryptoRandom". Prefer not to make any promises around being "random".

Possible that you can update the pull-request and rename the class-names?
- VariationGenerator.cs -> ScatterGenerator.cs
- VariationGeneratorTests.cs -> ScatterGeneratorTests.cs

Then it will be ready for merge.

snakefoot:
@jobvk Thank you for the nice contribution, and driving this. Believe NLog v6.2.1 will be released in 20 days (more or less)

snakefoot:
NLog v6.2.1 has now been released, and should fix the issue with NLog reported using unsafe `Random`:
- https://www.nuget.org/packages/NLog/6.2.1#releasenotes-body-tab

Thank you again for the contribution, and improving the NLog library.

PR #6267, opened 2026-09-03
JsonLayout - Escape lone surrogates so output stays valid UTF8

dylanpulver:
### Problem

An **unpaired surrogate** is written into JSON as-is, so the line is not UTF8-encodable — `RequiresJsonEscape` returns `EscapeUnicode` for every char > 127, and a lone surrogate has no UTF8 form. Easy to hit: truncating a message that ends mid-emoji leaves one.

```csharp
var layout = new JsonLayout { Attributes = { new JsonAttribute("msg", "${message}") } };
layout.Render(new LogEventInfo(LogLevel.Info, "l", "\ud83d"));
```

| | before | after |
|---|---|---|
| rendered | raw surrogate | `{"msg":"\ud83d"}` |
| strict UTF8 | **throws** | ok |
| lenient UTF8 | silently `�` | preserved |

### Why only JSON

`XmlHelper` already checks `XmlConvertIsXmlSurrogatePair` in three places. The JSON serializer was the one encoder without the check — same input and defaults, `XmlLayout` stays encodable, `JsonLayout` does not.

### Change

Escape lone surrogates as `\uXXXX`. **Valid pairs are still written as-is**, so existing output is byte-identical; only the broken case changes.

### Verified

- 4 new theory cases: fail 3/4 on `dev`, pass with the change.
- `dotnet test -f net8.0`, filtered to Layouts/Json/Xml/Csv/LayoutRenderers: **1259 passed, 0 failed**.
- Changing only the predicate (no pair passthrough) also stops the corruption, but escapes valid emoji too and changes existing output — hence the pair check.

Test builds surrogates from char-codes: a lone one cannot survive an `InlineData` literal.

### DIFF
--- a/src/NLog/Targets/DefaultJsonSerializer.cs
+++ b/src/NLog/Targets/DefaultJsonSerializer.cs
@@ -620,15 +620,6 @@ namespace NLog.Targets
                     continue;
                 }
 
-                if (!escapeUnicode && char.IsHighSurrogate(ch) && i + 1 < text.Length && char.IsLowSurrogate(text[i + 1]))
-                {
-                    // Valid surrogate-pair can be written as-is. Only lone surrogates need escaping,
-                    // since they cannot be encoded as UTF8 (and would corrupt the output)
-                    destination.Append(ch);
-                    destination.Append(text[++i]);
-                    continue;
-                }
-
                 switch (ch)
                 {
                     case '"':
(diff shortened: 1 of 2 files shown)


src/NLog/Targets/DefaultJsonSerializer.cs
```diff
@@ -620,6 +620,15 @@ private static void AppendStringEscape(StringBuilder destination, string text, b
                     continue;
                 }
 
+                if (!escapeUnicode && char.IsHighSurrogate(ch) && i + 1 < text.Length && char.IsLowSurrogate(text[i + 1]))
```
snakefoot: For sake of readability (and performance) can this logic be extracted into a method that is only called in the switch-default-case below?

snakefoot:
Thank you for the contribution, and the code update. Think it looks good, and also nice with the unit-test.

Will be released with NLog v6.2.1 (in 2 weeks time)

snakefoot:
NLog v6.2.1 has now been released, and includes your improvement for surrogate handling:
- https://www.nuget.org/packages/NLog/6.2.1#releasenotes-body-tab

Thank you again for the contribution, and improving the NLog library.

Issue #5917, opened 2025-07-03
Include property default value in XML docs


snakefoot:
To improve the Sandcastle-Docs, and normal API navigation then default-value should be more visible:
```csharp
        /// <summary>
        /// Gets or sets the layout that renders event ID.
        /// </summary>
        /// <value>Default: ${event-properties:item=EventId}</value>
        public Layout<int> EventId { get; set; } = "${event-properties:item=EventId}";
```
Maybe also hint whether property-value is required (and not optional).

PR #6275, opened 2026-09-05
docs: document default values for Filters properties in XML docs

dualfroz:
## Problem

Base: `NLog:dev` <- `dualfroz:feat/filters-default-value-xml-docs-5917`

Refs #5917

## Summary

#5917 asked for property default values to be visible in the generated XML
docs (Sandcastle docs / IntelliSense), not just in the source code. Three
PRs already landed for this (#5919 Targets and Layouts, #5922
LayoutRenderers, #5925 Target Wrappers), but the issue itself is still
open and the `NLog.Filters` namespace was never covered - none of its
properties document a default value.

This PR applies the same documented convention (`<remarks>Default: ...`,
and `<b>[Required]</b>` for properties a filter cannot function without)
to every settable property in `NLog.Filters`:

- `Filter.Action`
- `ConditionBasedFilter.Condition`
- `LayoutBasedFilter.Layout`
- `WhenContainsFilter.IgnoreCase`, `.Substring`
- `WhenEqualFilter.IgnoreCase`, `.CompareTo`
- `WhenNotContainsFilter.Substring`, `.IgnoreCase`
- `WhenNotEqualFilter.CompareTo`, `.IgnoreCase`
- `WhenRepeatedFilter.TimeoutSeconds`, `.MaxLength`, `.IncludeFirst`,
  `.MaxFilterCacheSize`, `.DefaultFilterCacheSize`,
  `.FilterCountPropertyName`, `.FilterCountMessageAppendFormat`,
  `.OptimizeBufferDefaultLength`

`WhenRepeatedFilter.OptimizeBufferReuse` is intentionally left alone -
it's `[Obsolete]` since NLog 5.0 and always returns `true`, so there is
no meaningful default left to document.

## Regression test

`FilterDefaultValueDocsTests.SettablePropertiesDocumentDefaultValue`
loads the XML doc file generated alongside `NLog.dll`
(`GenerateDocumentationFile` is already on for the project), then
reflects over every public settable property under `NLog.Filters` and
asserts its doc entry contains `Default:`. It fails for any current or
future property in that namespace that skips documenting its default,
not just the ones touched here.

## How to verify

```
dotnet test tests/NLog.UnitTests/NLog.UnitTests.csproj -f net8.0 -c Debug --filter "FullyQualifiedName~FilterDefaultValueDocsTests"
```

SDK: 9.0.317 (net8.0 runtime 8.0.30 installed alongside for the test host).

Result: 1/1 passed with the fix in place. Reverting the `src/NLog/Filters/*.cs`
changes (test file kept) makes the same test fail with 19 missing
properties reported, confirming it actually catches the regression.

### DIFF
--- a/src/NLog/Filters/ConditionBasedFilter.cs
+++ b/src/NLog/Filters/ConditionBasedFilter.cs
@@ -50,6 +50,7 @@ namespace NLog.Filters
         /// <summary>
         /// Gets or sets the condition expression.
         /// </summary>
+        /// <remarks><b>[Required]</b> Default: <see langword="null"/></remarks>
         /// <docgen category='Filtering Options' order='10' />
         public ConditionExpression Condition { get; set; } = ConditionLiteralExpression.Null;
 
--- a/src/NLog/Filters/Filter.cs
+++ b/src/NLog/Filters/Filter.cs
@@ -53,6 +53,7 @@ namespace NLog.Filters
         /// <summary>
         /// Gets or sets the action to be taken when filter matches.
         /// </summary>
+        /// <remarks>Default: <see cref="FilterResult.Neutral"/></remarks>
         /// <docgen category='Filtering Options' order='10' />
         public FilterResult Action { get; set; }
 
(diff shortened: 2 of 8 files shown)


snakefoot:
Thank you so much for the contribution, very much appreciated.

Like your idea about having an unit-test to ensure future developments also includes minimum documentation.

PR #6285, opened 2026-09-18
MemoryTarget - Simplify GetEnumerator to always enumerate entry count

snakefoot:


snakefoot:
Think complex is better, so it doesn't wrap around on clear.

